PR #7, opened 2023-09-19
revert: "perf: reduce Module size"

hyf0:
Reverts rolldown-rs/rolldown#6

`Module` would be stored in `Vec`, which is already in the heap. Wrapping with `Box` will cause extra dereferencing on every visiting, which `Module` would be visited frequently a lot.

`ExternalModule` are few, you don't need to worry about the memory usages now.

### DIFF
--- a/crates/rolldown/src/bundler/module/module.rs
+++ b/crates/rolldown/src/bundler/module/module.rs
@@ -10,8 +10,8 @@ use super::{external_module::ExternalModule, render::RenderModuleContext, Normal
 
 #[derive(Debug)]
 pub enum Module {
-  Normal(Box<NormalModule>),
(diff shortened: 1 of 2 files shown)


hyf0:
@underfin  [Reducing type size](https://nnethercote.github.io/perf-book/type-sizes.html) is a good way to improve performance, but it also has it's own restrictions.

Replacing `Box<str>` with [`thin_str`](https://docs.rs/thin_str/latest/thin_str/) is a win, but it's not only about reducing the size. `Box<Box<str>>`(1 * usize) is also smaller than `Box<str>`(2 * usize), but accessing `Box<Box<str>>` would be much slower than `Box<str>`.

PR #9, opened 2023-09-20
 feat: link external module

underfin:
The pr is a implement for external module linker. It includes external module imports and exports.

## Create external module symbols

Because the rust ownership i moved the logic of  create external module symbols to prepare linking prase.

## Link external module imports(Namespace/Default/Named import)

``` ts
// main.js
import * as ns from 'external'
import { a } from 'external'
console.log(ns, a)
```

The codegen will preserve the import external module decls and deconflict symbols.

## Link external module exports(Namespace/Named Reexport)

``` ts
// foo.js
export { a } from 'external'
// main.js
import * as ns from 'external'
import { a } from 'foo.js'
console.log(ns, a)
```
The codegen will preserve the exports external module decls and rewrite the ast with `import` decl.
- Namespace Reexport
``` .ts
export * as ns from 'external'
// rewrite the ast
import * as ns from 'external'
```
- Named Reexport
``` .ts
export { a } from 'external'
// rewrite the ast
import { a } from 'external'
```

# TODO

- Exports Satr Reexport  external module 
- entry exports external module


hyf0 (commented):
hold for CR

crates/rolldown/src/bundler/module/external_module.rs
```diff
@@ -26,31 +27,73 @@ impl ExternalModule {
...
+    &mut self,
+    symbols: &mut Symbols,
+    exported: Atom,
+    is_star: bool,
+  ) -> SymbolRef {
+    if is_star {
+      if let Some((_, symbol_ref)) = &self.symbol_for_namespace {
+        return *symbol_ref;
+      }
+      self.is_symbol_for_namespace_referenced = true;
```
hyf0: You don't need this, it should be resolved already by https://github.com/rolldown-rs/rolldown/blob/d7de04fdd5787a2733691b39050101596139eb48/crates/rolldown/src/bundler/graph/linker.rs#L50.

crates/rolldown/src/bundler/visitors/scanner.rs
```diff
@@ -78,12 +78,12 @@ impl<'a> Scanner<'a> {
...
   }
 
-  fn add_star_import(&mut self, local: SymbolId, record_id: ImportRecordId) {
+  fn add_star_import(&mut self, local: SymbolId, imported: &Atom, record_id: ImportRecordId) {
     self.result.import_records[record_id].is_import_namespace = true;
     self.result.named_imports.insert(
       local,
       NamedImport {
-        imported: Atom::new_inline("#STAR#"),
+        imported: imported.clone(),
```
hyf0: For `import * as foo from 'foo'`, there is no `imported` and `foo` is already stored in the `imported_as`. So the change is not needed.

crates/rolldown/src/bundler/module/external_module.rs
```diff
@@ -26,31 +27,73 @@ impl ExternalModule {
...
   pub fn finalize(&mut self, _ctx: ModuleFinalizeContext) {}
 
-  pub fn render(&self, _ctx: RenderModuleContext) -> Option<MagicString<'static>> {
-    let mut rendered = MagicString::new(format!("import \"{}\"", self.resource_id.as_ref()));
+  pub fn render(&self, ctx: RenderModuleContext) -> Option<MagicString<'static>> {
+    let mut rendered = if let Some((symbol, symbol_ref)) = &self.symbol_for_namespace {
+      let value = ctx
+        .final_names
+        .get(symbol_ref)
+        .unwrap_or(symbol)
```
hyf0: Using `unwrap` would enough. if the name not exist, it's a bug and should panic.

crates/rolldown/src/bundler/visitors/scanner.rs
underfin: The `imported_as` type is `SymbolRef` and the `foo` is not stored, so I use the `imported` to store it.

hyf0: `named_imports` is a `HashMap`. The `foo` is already stored as the key in the `HashMap`.

hyf0: You could add new field if needed.

Secondly, if you need to generate some name, you could do something like this

https://github.com/rolldown-rs/rolldown/blob/d7de04fdd5787a2733691b39050101596139eb48/crates/rolldown/src/bundler/module/module_builder.rs#L34

See more comments at https://github.com/rolldown-rs/rolldown/pull/9#discussion_r1331651540

crates/rolldown/src/bundler/module/external_module.rs
```diff
@@ -26,31 +27,73 @@ impl ExternalModule {
...
+    is_star: bool,
+  ) -> SymbolRef {
+    if is_star {
+      if let Some((_, symbol_ref)) = &self.symbol_for_namespace {
+        return *symbol_ref;
+      }
+      self.is_symbol_for_namespace_referenced = true;
+      let symbol_ref: SymbolRef = (
+        self.id,
+        symbols.tables[self.id].create_symbol(exported.clone()),
```
hyf0: For `is_alias_star: true`, just create a name using `self.path.generate_unique_name() + "ns"`. Don't misuse the `exported` variable.

If you really want to do this way, add a new param called `suggested_name: &Atom` for this use case.

crates/rolldown/src/bundler/module/external_module.rs
```diff
@@ -16,41 +13,50 @@ pub struct ExternalModule {
...
           self.id,
-          symbols.tables[self.id].create_symbol(exported.clone()),
+          symbols.tables[self.id].create_symbol(symbol.clone()),
         )
           .into()
-      })
+      });
+  }
+
+  pub fn resolve_export(&self, exported: Atom, is_star: bool) -> SymbolRef {
```
hyf0: `exported: Atom` should be `exported: &Atom`.

crates/rolldown/src/bundler/visitors/mod.rs
```diff
@@ -139,10 +249,23 @@ impl<'ast, 'p> VisitMut<'ast, 'p> for Finalizer<'ast> {
             }
           }
           oxc::ast::ast::ModuleDeclaration::ExportNamedDeclaration(named_decl) => {
+            // external: export { a } from 'a' => import { a } from 'a'
```
hyf0, line 252: What will happen for case
```js
export { a } from 'a'
const a = 1;
console.log(a)
```

crates/rolldown/src/bundler/visitors/mod.rs
```diff
@@ -51,10 +146,25 @@ impl<'ast> Finalizer<'ast> {
         }
         // FIXME: ugly hack
         oxc::ast::ast::ModuleDeclaration::ExportNamedDeclaration(decl) if decl.span.size() != 0 => {
+          if let Some(source) = &decl.source {
```
hyf0, line 149: IMO, we need to remove these statements related to external modules, otherwise, we can't deal with this [input](https://rollupjs.org/repl/?version=3.28.1&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMCdtMSclNUNuaW1wb3J0JTIwJTdCJTIwYSUyMCU3RCUyMGZyb20lMjAnLiUyRnNoYXJlZCclNUNuaW1wb3J0JTIwJ20yJyU1Q25jb25zb2xlLmxvZyhhKSUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTJDJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlN0QlMkMlN0IlMjJjb2RlJTIyJTNBJTIyZXhwb3J0JTIwY29uc3QlMjBhJTIwJTNEJTIwJ2EnJTVDbmV4cG9ydCUyMGNvbnN0JTIwYiUyMCUzRCUyMCdiJyUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTJDJTIybmFtZSUyMiUzQSUyMnNoYXJlZC5qcyUyMiU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJvdXRwdXQlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJlcyUyMiU3RCUyQyUyMnRyZWVzaGFrZSUyMiUzQXRydWUlN0QlN0Q=) with correct execution order.

Please add a test case of [this](https://rollupjs.org/repl/?version=3.28.1&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMCdtMSclNUNuaW1wb3J0JTIwJTdCJTIwYSUyMCU3RCUyMGZyb20lMjAnLiUyRnNoYXJlZCclNUNuaW1wb3J0JTIwJ20yJyU1Q25jb25zb2xlLmxvZyhhKSUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTJDJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlN0QlMkMlN0IlMjJjb2RlJTIyJTNBJTIyZXhwb3J0JTIwY29uc3QlMjBhJTIwJTNEJTIwJ2EnJTVDbmV4cG9ydCUyMGNvbnN0JTIwYiUyMCUzRCUyMCdiJyUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTJDJTIybmFtZSUyMiUzQSUyMnNoYXJlZC5qcyUyMiU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJvdXRwdXQlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJlcyUyMiU3RCUyQyUyMnRyZWVzaGFrZSUyMiUzQXRydWUlN0QlN0Q=)

crates/rolldown/src/bundler/visitors/mod.rs
underfin: I added to test to `export_external` test

crates/rolldown/src/bundler/visitors/mod.rs
underfin: It already been added at `import_external`. The import order issue look like is complex, i will work it at next pr.

hyf0 (approved):

crates/rolldown/src/bundler/visitors/mod.rs
hyf0: I gave a look and the `import_external` test is not like [this](https://rollupjs.org/repl/?version=3.28.1&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMCdtMSclNUNuaW1wb3J0JTIwJTdCJTIwYSUyMCU3RCUyMGZyb20lMjAnLiUyRnNoYXJlZCclNUNuaW1wb3J0JTIwJ20yJyU1Q25jb25zb2xlLmxvZyhhKSUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTJDJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlN0QlMkMlN0IlMjJjb2RlJTIyJTNBJTIyZXhwb3J0JTIwY29uc3QlMjBhJTIwJTNEJTIwJ2EnJTVDbmV4cG9ydCUyMGNvbnN0JTIwYiUyMCUzRCUyMCdiJyUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTJDJTIybmFtZSUyMiUzQSUyMnNoYXJlZC5qcyUyMiU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJvdXRwdXQlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJlcyUyMiU3RCUyQyUyMnRyZWVzaGFrZSUyMiUzQXRydWUlN0QlN0Q=)

PR #12, opened 2023-09-22
feat: add plugin driver

underfin:
### Description

Add plugin driver.

### Test Plan

---


hyf0:
I currently don't see the need to add `PluginDriver` while the main features haven't finished.

crates/rolldown/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 mod bundler;
-
+#[allow(dead_code)]
+pub mod plugin;
```
hyf0, line 3: Don't use `pub` here. Please, make sure every exported item are explicitly marked in lib.rs just like those one.

https://github.com/rolldown-rs/rolldown/blob/e94964bec0786e4621cc787f4b9c272d148b37db/crates/rolldown/src/lib.rs#L11-L18

crates/rolldown/src/plugin/resolve_id.rs
```diff
@@ -17,23 +17,14 @@ pub struct ResolveIdArgs<'a> {
...
-      },
-      _ => false,
-    }
-  }
+pub struct ResolveIdResult {
+  pub id: String,
+  pub external: Option<ResolveIdExternal>,
+  //   pub assertions: Option<FxHashMap<String, String>>,
+  //   pub meta: Option<FxHashMap<String, String>>,
+  pub module_side_effects: Option<ModuleSideEffects>,
```
hyf0, line 25: I think again, I don't think introduce those fields which we are not going to used for a not short time.

These unused code add complexity to codebase and I don't think this is a  good idea.

crates/rolldown/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 mod bundler;
-
+#[allow(dead_code)]
```
hyf0, line 2: `dead code` adds useless complexity, I don't recommend it.

crates/rolldown/src/plugin/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+  pub fn with_context(context: T) -> Self {
+    Self { context }
+  }
+}
+
+#[async_trait::async_trait]
+pub trait Plugin: Debug + Send + Sync {
+  fn name(&self) -> &'static str;
+
+  async fn options(&self, _options: &mut InputOptions) -> rolldown_error::Result<()> {
```
hyf0, line 39: IMO, it's not a good idea to add those hooks rushly without deep thinking  and practicing.

PR #14, opened 2023-09-25
fix: external module import order

underfin:
### Description

The pr fix external module import order(`import_external` test).

#### Module Ast Part
The pr implement `Ast Part`, a part is composed of some ast stmts. 

``` .ts
// main.js
import 'external'  // execute first
import 'a.js'
```
See the above module the external module should execute first, but before implement will render to the following code.

``` .ts
// a.js
xxx
// main.js
import 'external'
```

The part is used to generate correct code execute order, see the following.

``` .ts
// main.js
import 'external'
// a.js
xxx
```

The `main.js` has `external` part and `a.js` part, it will scanned ast first scan ast.

#### Chunk Part Codegen

Before chunk codegen is implemented by module ast codegen, the pr will collect chunk parts by visite chunk module part. The chunk render is also used ast part to gen some ast stmts. 

### Test Plan

Update snapshot.

---

### DIFF
--- a/crates/rolldown/src/bundler/graph/graph.rs
+++ b/crates/rolldown/src/bundler/graph/graph.rs
@@ -57,6 +57,8 @@ impl Graph {
       match action {
         Action::Enter => {
           if !entered_ids.contains(&id) {
+            *module.exec_order_mut() = next_exec_order;
+            next_exec_order += 1;
             entered_ids.insert(id);
             stack.push((Action::Exit, id));
             stack.extend(
@@ -74,8 +76,6 @@ impl Graph {
           }
         }
         Action::Exit => {
-          *module.exec_order_mut() = next_exec_order;
-          next_exec_order += 1;
           sorted_modules.push(id);
         }
       }
--- a/crates/rolldown/src/bundler/visitors/scanner.rs
+++ b/crates/rolldown/src/bundler/visitors/scanner.rs
@@ -11,14 +11,15 @@ use oxc::{
   span::Atom,
 };
 use rolldown_common::{
-  ImportRecord, ImportRecordId, LocalExport, LocalOrReExport, ModuleId, NamedImport, ReExport,
-  StmtInfo, StmtInfoId,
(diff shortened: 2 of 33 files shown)


crates/rolldown/src/bundler/graph/graph.rs
```diff
@@ -57,6 +57,8 @@ impl Graph {
       match action {
         Action::Enter => {
           if !entered_ids.contains(&id) {
+            *module.exec_order_mut() = next_exec_order;
+            next_exec_order += 1;
```
underfin, line 61: Before only keep execute module order, but shouln't keep module execute stmts order.

crates/rolldown/src/bundler/visitors/scanner.rs
```diff
@@ -11,14 +11,15 @@ use oxc::{
...
 };
 use rolldown_oxc::BindingIdentifierExt;
 use rustc_hash::FxHashMap;
 
 #[derive(Debug, Default)]
 pub struct ScanResult {
+  pub parts: IndexVec<PartId, Part>,
   pub named_imports: FxHashMap<SymbolId, NamedImport>,
   pub named_exports: FxHashMap<Atom, LocalOrReExport>,
   pub stmt_infos: IndexVec<StmtInfoId, StmtInfo>,
```
hyf0, line 25: We already have the similar struct like `Part`, which is called `StmtInfo`.

crates/rolldown/tests/esbuild/default/export_forms_es6/artifacts.snap
```diff
@@ -27,5 +16,16 @@ function Fn () {
 class Class {
 }
 
+// a.js
```
hyf0, line 19: The old snapshot show a wrong order, which is a bug, but the new snapshot seems also wrong about the order between `a.js` and `main.js`.

hyf0:
I currently don't prefer to use `Part` or `StmtInfo` to solve the order problem, it adds unnecessary complexity for solving the problem.

crates/rolldown/src/bundler/graph/graph.rs
hyf0: I don't get your words.

PR #20, opened 2023-09-28
refactor: module render with source mutation

### LINKED ISSUE #17: render module with source mutation instead of ast mutaition
## Description

Module render implementation has two ways.

###  Module Render With Ast Mutation

Replaced the ast nodes then codegen ast.eg Babel transform, esbuild bundling.

#### Advantages

The chunk render can be implemented by concat the modules ast, and also chunk minify/format can be implemented by one ast parse. The cold start performance is very fast.

#### Drawbacks

Do ast mutation causes the ast to be dirty at rebuild, we must parse the code again and cause rebuild performance regression.

Other the developer experience is bad because we need to manually create ast nodes, and the code snip is huge and difficult to review.

### Module Render With Source Mutation

Replace the original source using the ast span position. eg Rollup using MagicString, webpack using ReplaceSource.

#### Drawbacks

The chunk render can be implemented by concat the modules mutation source and the minify/format needs to be parse again.

## Rolldown main branch implement

- Module render with Ast Mutation
> Here has walk ast tree twice, one is scan ast to launch module graph,  one is walk ast tress to run ast mutation
- Chunk render with concat modules ast codegen result.

## Launch stage

Should be finished at early stage, it will block tree shaking codegen and other things.

## Launch Source Mutation Step

Add source Mutation action abstract and scan the mutation actions with walking ast tree once.

underfin:
### Description

close https://github.com/rolldown-rs/rolldown/issues/17.
The external module render will refactor at chunk split, i will skip the external module test.

### Test Plan

Not need

---

### DIFF
--- a/crates/rolldown/src/bundler/module/source_mutation.rs
+++ b/crates/rolldown/src/bundler/module/source_mutation.rs
@@ -1,0 +1,37 @@
+use oxc::{
+  semantic::{ReferenceId, SymbolId},
+  span::{Atom, Span},
+};
+use rolldown_common::{ModuleId, SymbolRef};
+use rustc_hash::FxHashMap;
+
+use crate::bundler::graph::symbols::Symbols;
+
+#[derive(Debug)]
+pub enum SourceMutation {
+  RenameSymbol(Box<(Span, Atom)>),
+  Remove(Box<Span>),
+  AddExportDefaultBindingIdentifier(Box<Span>),
+  //   AddNamespaceExport(),
+}
+
+pub fn get_symbol_final_name<'a>(
+  module_id: ModuleId,
+  symbol_id: SymbolId,
+  symbols: &'a Symbols,
+  final_names: &'a FxHashMap<SymbolRef, Atom>,
+) -> Option<&'a Atom> {
+  let symbol_ref = (module_id, symbol_id).into();
+  let final_ref = symbols.par_get_canonical_ref(symbol_ref);
+  final_names.get(&final_ref)
+}
+
+pub fn get_reference_final_name<'a>(
+  module_id: ModuleId,
+  reference_id: ReferenceId,
+  symbols: &'a Symbols,
+  final_names: &'a FxHashMap<SymbolRef, Atom>,
+) -> Option<&'a Atom> {
+  symbols.tables[module_id].references[reference_id]
+    .and_then(|symbol| get_symbol_final_name(module_id, symbol, symbols, final_names))
+}
--- a/crates/rolldown/src/bundler/module/source_mutation.rs
+++ b/crates/rolldown/src/bundler/module/source_mutation.rs
@@ -1,37 +1,9 @@
-use oxc::{
-  semantic::{ReferenceId, SymbolId},
-  span::{Atom, Span},
-};
-use rolldown_common::{ModuleId, SymbolRef};
-use rustc_hash::FxHashMap;
-
-use crate::bundler::graph::symbols::Symbols;
+use oxc::span::{Atom, Span};
 
 #[derive(Debug)]
 pub enum SourceMutation {
   RenameSymbol(Box<(Span, Atom)>),
   Remove(Box<Span>),
   AddExportDefaultBindingIdentifier(Box<Span>),
-  //   AddNamespaceExport(),
-}
-
-pub fn get_symbol_final_name<'a>(
-  module_id: ModuleId,
-  symbol_id: SymbolId,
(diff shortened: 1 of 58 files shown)


underfin:
!bench

crates/rolldown/tests/fixtures/import_external/artifacts.snap
```diff
@@ -7,12 +7,12 @@ input_file: crates/rolldown/tests/fixtures/import_external
...
+
+console.log(ext, a, b)
 // main.js
-import * as external_ns from 'external';
-import { a, b } from 'external';
-console.log(external_ns, a, b);
+
+
+
+console.log(ext, a, b)
```
IWANABETHATGUY: This snap looks wrong.

underfin: Yeah. The external module render will refactor at chunk split, i will skip the external module test.

crates/rolldown_oxc/src/compiler.rs
```diff
@@ -12,7 +12,7 @@ pub struct OxcCompiler;
 #[allow(clippy::box_collection)]
 pub struct OxcProgram {
   program: ast::Program<'static>,
-  source: Pin<Box<String>>,
+  pub source: Pin<Box<String>>,
```
hyf0: Adding a new method `source` which returns `&str` would be much proper/safer.

crates/rolldown/src/bundler/module_loader/module_task.rs
```diff
@@ -79,6 +80,7 @@ impl ModuleTask {
     builder.import_records = Some(import_records);
     builder.star_exports = Some(star_exports);
     builder.default_export_symbol = export_default_symbol_id;
+    builder.source_mutation = source_mutation;
```
hyf0: I have a different thought now. We don't want mutate AST, but we will keep holding the AST. So there's no need to force to create all source mutations in the scan pass.

crates/rolldown/src/bundler/visitors/scanner.rs
```diff
@@ -281,13 +318,34 @@ impl<'ast, 'p> VisitMut<'ast, 'p> for Scanner<'ast> {
...
   }
 
+  /// avoid visit_binding_identifier for import decl
+  fn visit_import_specifier(&mut self, _: &'p mut oxc::ast::ast::ImportSpecifier) {}
+  fn visit_import_default_specifier(&mut self, _: &'p mut oxc::ast::ast::ImportDefaultSpecifier) {}
+  fn visit_import_name_specifier(&mut self, _: &'p mut oxc::ast::ast::ImportNamespaceSpecifier) {}
+
   fn visit_binding_identifier(&mut self, ident: &'p mut oxc::ast::ast::BindingIdentifier) {
     let symbol_id = ident.symbol_id.get().unwrap();
     if self.scope.root_scope_id() == self.symbol_table.get_scope_id(symbol_id) {
```
hyf0: This is wrong. All symbol is possibly going to be renamed, including scoped symbols.

crates/rolldown/src/bundler/visitors/mod.rs
```diff
@@ -1,276 +1 @@
...
-  semantic::{ScopeTree, SymbolId},
-  span::Atom,
-};
-use rolldown_common::{ModuleId, ResolvedExport, SymbolRef};
-use rolldown_oxc::{BindingIdentifierExt, TakeIn};
-use rustc_hash::{FxHashMap, FxHashSet};
-
-use super::graph::symbols::Symbols;
-
-pub struct FinalizeContext<'ast> {
```
hyf0, line 23: I think we gonna still have the `finalize` pass, but we just don't mutate the AST anymore.

underfin:
!bench

underfin:
!bench

crates/rolldown/src/bundler/module/source_mutation.rs
```diff
@@ -0,0 +1,9 @@
+use oxc::span::{Atom, Span};
+
+#[derive(Debug)]
+pub enum SourceMutation {
```
hyf0, line 4: I think `trait` is good now, since we are going to box everything. But we could refactor it in another PR.

hyf0 (approved):

crates/rolldown/src/bundler/module/source_mutation.rs
underfin: Ok.

PR #21, opened 2023-09-28
refactor: graph sorted module

IWANABETHATGUY:
### Test Plan

---

### DIFF
--- a/crates/rolldown/src/bundler/graph/graph.rs
+++ b/crates/rolldown/src/bundler/graph/graph.rs
@@ -34,17 +34,11 @@ impl Graph {
   }
 
   pub fn sort_modules(&mut self) {
-    #[derive(PartialEq, Eq, PartialOrd, Ord, Hash)]
-    enum Action {
-      Enter,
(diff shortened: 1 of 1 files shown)


crates/rolldown/src/bundler/graph/graph.rs
```diff
@@ -95,3 +89,19 @@ impl Graph {
...
+
+#[derive(PartialEq, Eq, PartialOrd, Ord, Hash)]
+enum Action {
+  Enter(ModuleId),
+  Exit(ModuleId),
+}
+
+impl Action {
+  #[inline]
+  fn get_module_id(&self) -> ModuleId {
```
hyf0: LGTM. But I think `module_id` is a more rust-style name than `get_module_id`.

hyf0 (approved):

crates/rolldown/src/bundler/graph/graph.rs
IWANABETHATGUY: Done

PR #23, opened 2023-10-04
feat: basic code splitting

underfin:
### Description

related https://github.com/rolldown-rs/rolldown/issues/22.

- [x] generate chunks and chunk own modules
- [ ] cross chunks symbol imports

### Test Plan

---


underfin:
!bench

PR #27, opened 2023-10-07
feat: code spltting for dynamic import

underfin:
### Description

related https://github.com/rolldown-rs/rolldown/issues/22

### Test Plan

Update simple case

---

PR #37, opened 2023-10-08
feat: prepare supporting runtime

underfin:
### Description

related https://github.com/rolldown-rs/rolldown/issues/35

### Test Plan

no need.

---

### DIFF
--- a/crates/rolldown/src/bundler/graph/graph.rs
+++ b/crates/rolldown/src/bundler/graph/graph.rs
@@ -1,7 +1,7 @@
 use super::{linker::Linker, symbols::Symbols};
 use crate::bundler::{
   module::module_id::ModuleVec, module_loader::ModuleLoader,
-  options::normalized_input_options::NormalizedInputOptions,
+  options::normalized_input_options::NormalizedInputOptions, runtime::Runtime,
 };
 use rolldown_common::ModuleId;
 use rustc_hash::FxHashSet;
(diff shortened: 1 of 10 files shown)


crates/rolldown/src/bundler/graph/graph.rs
```diff
@@ -13,6 +13,7 @@ pub struct Graph {
   pub entries: Vec<(Option<String>, ModuleId)>,
   pub sorted_modules: Vec<ModuleId>,
   pub symbols: Symbols,
+  pub runtime: ModuleId,
```
underfin: The runtime module put into `modules` of graph and use `runtime` to refrence it, but maybe is good to put it to `runtime`.

underfin:
!bench

hyf0:
Would you mind holding this PR before #27 being merged? I need to apply some changes to this PR after #27.

underfin:
ok

crates/rolldown/src/bundler/module_loader/module_task.rs
```diff
@@ -50,7 +51,11 @@ impl ModuleTask {
     let mut builder = ModuleBuilder::default();
     tracing::trace!("process {:?}", self.path);
     // load
-    let source = tokio::fs::read_to_string(self.path.as_ref()).await?;
+    let source = if RUNTIME_PATH == self.path.as_ref() {
```
hyf0, line 54: I'm intending to not include `runtime` module as the part of making module graph. In this way, we don't need to write such code.

Another way is to have a special `ModuleTaskForRuntime` to deal with the special logic of runtime module.

I haven't think through which way is better. Let's change it later.

PR #38, opened 2023-10-09
feat: basic commonjs support

underfin:
### Description

related https://github.com/rolldown-rs/rolldown/issues/36

### Test Plan

Add basic.

---


crates/rolldown/src/bundler/visitors/commonjs_source_render.rs
```diff
@@ -0,0 +1,66 @@
+use oxc::ast::Visit;
+use rolldown_common::ModuleResolution;
+
+use crate::bundler::module::module::Module;
+
+use super::RendererContext;
+
+pub struct CommonJsSourceRender<'ast> {
```
hyf0, line 8: Any reason why we can not merge those renderer into one?

crates/rolldown/tests/fixtures/basic_commonjs/artifacts.snap
```diff
@@ -0,0 +1,42 @@
...
+var esm_named_var;
+
+var esm_ns = {
+  get esm_named_var() { return esm_named_var },
+  get default() { return esm_default },
+  get esm_named_fn() { return esm_named_fn },
+  get esm_named_class() { return esm_named_class }
+};
+var init_esm = __esm({
+'esm.js'() {
```
hyf0, line 23: We need this https://github.com/Rich-Harris/magic-string/blob/cf554e7b9384fa09695d541cb84dc5118b8374e0/README.md?plain=1#L125.

crates/rolldown/src/bundler/visitors/commonjs_source_render.rs
underfin: Merging three kinds of source render into one visitor, will cause code confusion to understand and difficulty to review.

crates/rolldown/tests/fixtures/basic_commonjs/artifacts.snap
underfin: Yes.

crates/rolldown/src/bundler/visitors/mod.rs
```diff
@@ -102,6 +107,15 @@ impl<'ast> RendererContext<'ast> {
     get_reference_final_name(module_id, reference_id, self.symbols, self.final_names)
   }
 
+  pub fn get_runtime_symbol_final_name(&self, name: &str) -> Atom {
+    let symbol = self.runtime.resolve_symbol(&name.into());
+    self
+      .get_symbol_final_name(self.runtime.id, symbol)
```
hyf0: Would you mind using `SymbolRef` rather than `(self.runtime.id, symbol)`?

crates/rolldown/src/bundler/runtime/mod.rs
```diff
@@ -24,8 +24,7 @@ impl Runtime {
       .collect()
   }
 
-  pub fn _resolve_symbol(&self, name: &Atom) -> SymbolRef {
-    let symbol_id = self.name_to_symbol[name];
-    (self.id, symbol_id).into()
+  pub fn resolve_symbol(&self, name: &Atom) -> SymbolId {
```
hyf0: Returning `SymbolRef` would be more convenient for linking or other functions.

crates/rolldown/src/bundler/visitors/mod.rs
hyf0: It would be also good if you could change signature of `get_symbol_final_name` to

```rust
pub fn get_canonical_name<'a>(
  symbol_ref: SymbolRef,
  symbols: &'a Symbols,
  canonical_names: &'a FxHashMap<SymbolRef, Atom>,
) -> Option<&'a Atom> {
  let symbol_ref = (module_id, symbol_id).into();
  let final_ref = symbols.par_get_canonical_ref(symbol_ref);
  canonical_names.get(&final_ref)
}
```

crates/rolldown/tests/fixtures/basic_commonjs/artifacts.snap
underfin: I list it to task.

hyf0 (approved):

underfin:
!bench

crates/rolldown/src/bundler/visitors/scanner.rs
```diff
@@ -25,7 +25,8 @@ pub struct ScanResult {
   pub import_records: IndexVec<ImportRecordId, ImportRecord>,
   pub star_exports: Vec<ImportRecordId>,
   pub export_default_symbol_id: Option<SymbolId>,
-  pub dynamic_imports: FxHashMap<Span, ImportRecordId>,
+  pub imports: FxHashMap<Span, ImportRecordId>,
+  pub module_resolution: Option<ModuleResolution>,
```
hyf0, line 29: What's `ModuleResolution`?

Looking into the code, I thought it  was marking a module whether is cjs or esm?

crates/rolldown/src/bundler/visitors/scanner.rs
```diff
@@ -54,6 +55,16 @@ impl<'a> Scanner<'a> {
     }
   }
 
+  fn set_module_resolution(&mut self, module_resolution: ModuleResolution) {
+    if let Some(resolution) = &self.result.module_resolution {
+      if resolution != &module_resolution {
+        // TODO shouldn't mix esm syntax and cjs syntax
```
hyf0, line 61: Is `mix esm syntax and cjs syntax` not allowed or not implemented?

crates/rolldown/src/bundler/visitors/mod.rs
```diff
@@ -48,98 +83,67 @@ impl<'ast> SourceRenderer<'ast> {
...
+    get_reference_final_name(module_id, reference_id, self.symbols, self.final_names)
+  }
+
+  pub fn get_runtime_symbol_final_name(&self, name: &str) -> Atom {
+    let symbol = self.runtime.resolve_symbol(&name.into());
+    self
+      .get_symbol_final_name(symbol)
+      .cloned()
+      .unwrap_or_else(|| name.into())
+    // .expect(&format!("runtime symbol {name} not found"))
```
hyf0, line 111: Why this was comment out? panic should be expected here.

underfin: Because runtime symbol deconfict is not implemented.

crates/rolldown/src/bundler/visitors/scanner.rs
underfin: I remember it shouldn't support rolldown at previous rolldown meetings

crates/rolldown/src/bundler/visitors/scanner.rs
underfin: Yes.

crates/rolldown/src/bundler/visitors/mod.rs
hyf0: It's better to add a `TODO` comment here or raise a issue, otherwise, this is easily to be forgotten.

underfin: I already added to runtime task.

crates/rolldown/src/bundler/visitors/scanner.rs
hyf0: As I remember correctly, the meeting said it should be supported, but it could be enable or disable in module level. https://github.com/rolldown-rs/meeting-notes/blob/3fe04946dd858b4f5844763e0f476318b17d7854/09-18-2023.md?plain=1#L30

crates/rolldown/src/bundler/visitors/mod.rs
hyf0: Have you also add that person who implementing the `runtime symbol deconflicting` should un-comment out this line of code? 

If you haven't, I think it's just like I said above that this is easily to be forgotten.

crates/rolldown/src/bundler/visitors/scanner.rs
hyf0: I think it could be replaced with `module_type` that you gonna to add in https://github.com/rolldown-rs/rolldown/pull/56.

PR #51, opened 2023-10-12
test: add Rollup test

### LINKED ISSUE #46: Setup rollup test suite
https://github.com/rolldown-rs/rolldown_legacy/tree/main/packages/rollup-tests

underfin:
### Description
close https://github.com/rolldown-rs/rolldown/issues/46.
You can use `yarn test:update` if run some test success, The `packages/rollup-tests/src/failed-tests.json` will remove some success tests.

Current test status

┌────────────┬────────┐
│  (index)   │ Values │
├────────────┼────────┤
│   total    │  869   │
│   failed   │   0    │
│ skipFailed │  629   │
│  ignored   │   8    │
│  skipped   │   0    │
│   passed   │  232   │
└────────────┴────────┘

### Test Plan

---


hyf0 (approved):

Boshen:
Rollup is a moving target, perhaps we should submodule it instead of adding so many files.

underfin:
I have a try for as rollup as git submodules, here has some problems.
- rollup test runner(utils.js)  and test case depend on the runner(require some method from runner), so I meet crash at run rollup test using custom runner. It means we need port test cases and using custom runner.
- rollup test runner need a lot of node_modules dependencies, which has some post-install package, eg esbuild/puppeteer,etc.So it look like isn't a clean way if we use it as test runner.

hyf0 (approved):

PR #53, opened 2023-10-13
feat: add plugin adapter

underfin:
### Description

related https://github.com/rolldown-rs/rolldown/issues/48

### Test Plan

-

---


crates/rolldown/src/bundler/bundler.rs
```diff
@@ -12,14 +13,16 @@ use crate::{bundler::bundle::bundle::Bundle, InputOptions};
 
 pub struct Bundler {
   input_options: NormalizedInputOptions,
+  _plugins: Vec<BoxPlugin>,
 }
 
 impl Bundler {
-  pub fn new(input_options: InputOptions) -> Self {
+  pub fn new(input_options: InputOptions, plugins: Vec<BoxPlugin>) -> Self {
```
hyf0: Could you add a new method `with_plugins` and keep the `new` as the original one.

underfin: Yeah. Ok.

hyf0 (approved):

PR #54, opened 2023-10-14
feat: runtime symbol deconflict

underfin:
### Description

related https://github.com/rolldown-rs/rolldown/issues/36

### Test Plan

---

### DIFF
--- a/crates/rolldown/src/bundler/graph/linker.rs
+++ b/crates/rolldown/src/bundler/graph/linker.rs
@@ -61,7 +61,7 @@ impl<'graph> Linker<'graph> {
       match module {
         Module::Normal(module) => {
           if module.module_resolution == ModuleResolution::CommonJs {
-            mark_module_wrap_and_reference(self.graph, module.id, &mut modules_wrap);
+            mark_module_wrap(self.graph, module.id, &mut modules_wrap);
           }
         }
         Module::External(_) => {}
@@ -136,7 +136,7 @@ impl<'graph> Linker<'graph> {
       }
     }
 
-    fn mark_module_wrap_and_reference(
+    fn mark_module_wrap(
       graph: &Graph,
       module_id: ModuleId,
       modules_wrap: &mut IndexVec<ModuleId, bool>,
@@ -148,7 +148,7 @@ impl<'graph> Linker<'graph> {
           }
           modules_wrap[module_id] = true;
           module.import_records.iter().for_each(|record| {
-            mark_module_wrap_and_reference(graph, record.resolved_module, modules_wrap);
+            mark_module_wrap(graph, record.resolved_module, modules_wrap);
           });
         }
         Module::External(_) => {}
--- a/crates/rolldown/src/bundler/graph/linker.rs
+++ b/crates/rolldown/src/bundler/graph/linker.rs
@@ -19,7 +19,7 @@ impl<'graph> Linker<'graph> {
   }
 
   pub fn link(&mut self) {
-    self.create_module_wrap_symbol_and_runtime_symbol();
+    self.wrap_modules_if_needed();
 
     // propagate star exports
     for id in &self.graph.sorted_modules {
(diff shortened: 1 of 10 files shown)


crates/rolldown/src/bundler/graph/linker.rs
```diff
@@ -53,52 +55,121 @@ impl<'graph> Linker<'graph> {
...
     }
+
+    fn create_runtime_warp_symbol_reference(
+      module: &NormalModule,
+      symbols: &mut Symbols,
+      runtime_esm_symbol: SymbolId,
+      runtime_commonjs_symbol: SymbolId,
+    ) {
+      if module.module_resolution == ModuleResolution::CommonJs {
+        symbols.tables[module.id].create_reference(Some(runtime_commonjs_symbol));
```
hyf0: The return of `create_reference` was not used. What's the function for?

crates/rolldown/src/bundler/graph/linker.rs
```diff
@@ -53,52 +55,121 @@ impl<'graph> Linker<'graph> {
...
+
+    for (module_id, importers) in modules_wrap.into_iter_enumerated() {
+      if !importers.is_empty() {
         match &mut self.graph.modules[module_id] {
-          Module::Normal(m) => {
-            m.add_wrap_symbol(&mut self.graph.symbols);
+          Module::Normal(importee) => {
+            importee.create_wrap_symbol(&mut self.graph.symbols);
+
+            create_runtime_warp_symbol_reference(
```
hyf0: Since you are going to iterate the all module no matter how, you could initialize all runtime symbols in all `NormalModule`. This will be much cleaner and it's not a  heavy operation.

crates/rolldown/src/bundler/module/normal_module.rs
```diff
@@ -298,7 +298,7 @@ impl NormalModule {
     }
   }
```
hyf0, line 300: Add a `initialize_runtime_symbols` method to create all runtime symbols for every `NormalModule`.

crates/rolldown/src/bundler/graph/linker.rs
```diff
@@ -53,52 +55,121 @@ impl<'graph> Linker<'graph> {
...
+            }
+          });
         }
         Module::External(_) => {}
       }
     }
 
-    for (module_id, wrap) in modules_wrap.into_iter_enumerated() {
-      if wrap {
+    let runtime_esm_symbol = self.graph.runtime.resolve_symbol_id(&"__esm".into());
```
hyf0: `SymbolId` is not crossing module, you should use `SymbolRef` here.

crates/rolldown/src/bundler/graph/linker.rs
```diff
@@ -53,52 +55,121 @@ impl<'graph> Linker<'graph> {
...
+      importer: &NormalModule,
+      importee: &NormalModule,
+      symbols: &mut Symbols,
+      runtime_to_esm_symbol: SymbolId,
+      runtime_to_commonjs_symbol: SymbolId,
+    ) {
+      if importer.module_resolution == ModuleResolution::Esm
+        && importee.module_resolution == ModuleResolution::CommonJs
+      {
+        symbols.tables[importer.id].create_reference(Some(runtime_to_esm_symbol));
```
hyf0: The `runtime_to_esm_symbol` is belong to the runtime module. You should create a local symbol id just belong to the `importer` and union them together, otherwise, it might get conflicted.

crates/rolldown/src/bundler/runtime/mod.rs
```diff
@@ -29,4 +29,11 @@ impl Runtime {
       .unwrap_or_else(|| panic!("Failed to resolve symbol: {}", name));
     (self.id, *symbol_id).into()
   }
+
+  pub fn resolve_symbol_id(&self, name: &Atom) -> SymbolId {
```
hyf0: As the upper comment said, you don't need to add this method.

crates/rolldown/src/bundler/graph/linker.rs
underfin: Here need to initialize specific runtime symbol, because we also need to create wrap symbol by condition, It will be do at same time.

hyf0: initializing all runtime symbols in all NormalModule won't affects the creating wrap symbol by condition.

crates/rolldown/src/bundler/graph/linker.rs
hyf0: @underfin Why resolve this comment? This statement seems meaningless, what's usage of this?

crates/rolldown/src/bundler/module/normal_module.rs
```diff
@@ -47,7 +47,9 @@ pub struct NormalModule {
   pub namespace_symbol: (SymbolRef, ReferenceId),
   pub is_symbol_for_namespace_referenced: bool,
   pub cjs_symbols: FxHashMap<Atom, SymbolRef>,
-  pub wrap_symbol: Option<SymbolId>,
+  pub wrap_symbol: Option<SymbolRef>,
+  // Imported symbols that are used in the module, including runtime symbols, wrap symbols
+  pub import_symbols: FxHashMap<SymbolRef, SymbolRef>,
```
hyf0: I thought `StmtInfo` + `ImportRecords` are enough to store the used symbols from other modules. Why not keep using them? Is this how esbuild did the similar thing?

PR #56, opened 2023-10-15
feat: compat node module commonjs interop

### LINKED ISSUE #55: esm & cjs interop support node.js compat
Interop with Esm & commonjs has two mode. 
- babel mode will detect `__esModule` to identify the original commonjs or esm transformed commonjs and do symbol binding
- node mode don't do this, node.js always binds the default export to the value of module.exports. It will enabled at `.mjs` file and `type: module` at package.json.

ref: https://babeljs.io/docs/babel-plugin-transform-modules-commonjs#importinterop

underfin:
### Description

close https://github.com/rolldown-rs/rolldown/issues/55

Why need this, please refer to https://github.com/rolldown-rs/rolldown/issues/55. Because we follow esbuild implemention, the pr implement has two parts.
- detect `ModuleType` at resolver
- If importer module is ".mjs" or "type:module" at package.json,  codegen result will render to  `__toEsm(xx, 1/*node mode*/)` 

### Test Plan

Added.

---

### DIFF
--- a/crates/rolldown/src/bundler/module/module_builder.rs
+++ b/crates/rolldown/src/bundler/module/module_builder.rs
@@ -58,9 +58,9 @@ impl NormalModuleBuilder {
       default_export_symbol: self.default_export_symbol,
       namespace_symbol: self.namespace_symbol.unwrap(),
       is_symbol_for_namespace_referenced: false,
-      source_mutations: Default::default(),
+      source_mutations: Vec::default(),
       exports_kind: self.exports_kind.unwrap_or(ExportsKind::Esm),
-      cjs_symbols: Default::default(),
+      cjs_symbols: FxHashMap::default(),
       wrap_symbol: None,
       module_type: self.module_type,
     }
--- a/crates/rolldown/src/bundler/module/module_builder.rs
+++ b/crates/rolldown/src/bundler/module/module_builder.rs
@@ -4,7 +4,7 @@ use oxc::{
   span::{Atom, Span},
 };
 use rolldown_common::{
-  ImportRecord, ImportRecordId, LocalOrReExport, ModuleId, ModuleResolution, NamedImport,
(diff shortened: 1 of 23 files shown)


crates/rolldown/src/bundler/module/module_builder.rs
```diff
@@ -29,6 +29,7 @@ pub struct NormalModuleBuilder {
   pub default_export_symbol: Option<SymbolId>,
   pub namespace_symbol: Option<(SymbolRef, ReferenceId)>,
   pub module_resolution: Option<ModuleResolution>,
+  pub type_module: bool,
```
hyf0: The name is really confusing. Could you come up with a proper name for it?

What's the relationship between `type_module: true` and module_resolution field?

crates/rolldown_resolver/src/lib.rs
```diff
@@ -74,6 +76,7 @@ impl Resolver {
     match resolved {
       Ok(info) => Ok(ResolveRet {
         resolved: info.path().to_string_lossy().to_string().into(),
+        type_module: is_type_module(&info),
```
hyf0: https://github.com/evanw/esbuild/blob/b94c7a05b302f0eed76e54ef69dba98888c3acab/internal/js_ast/js_ast.go#L1440-L1454

underfin: It is unnecessary at now, we use `type_module`(or other name) is enough.

hyf0: I suggest using `enum` with only one variant. It has the same ability as you use `bool` but open to extension for the future changes.

crates/rolldown/src/bundler/module/module_builder.rs
hyf0: For the naming, I think we could just call it `module_type`.

hyf0 (approved):

PR #69, opened 2023-10-17
chore: 🤖 use rust 1.73.0

### LINKED ISSUE #65: Rust v1.73.0

IWANABETHATGUY:
### Description
1. Closed https://github.com/rolldown-rs/rolldown/issues/65
2. Use toolchain rust 1.73.0
3. The modifications are just cargo clippy fixing .

### Test Plan

---

### DIFF
--- a/crates/rolldown/src/bundler/module_loader/module_loader.rs
+++ b/crates/rolldown/src/bundler/module_loader/module_loader.rs
@@ -49,7 +49,7 @@ impl<'a> ModuleLoader<'a> {
       return Err(anyhow::format_err!("You must supply options.input to rolldown"));
     }
 
-    let resolved_entries = self.resolve_entries().await?;
+    let resolved_entries = self.resolve_entries();
 
     let mut intermediate_modules: IndexVec<ModuleId, Option<Module>> =
       IndexVec::with_capacity(resolved_entries.len() + 1 /* runtime */);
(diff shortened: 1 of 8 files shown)


crates/rolldown/src/bundler/module_loader/module_loader.rs
```diff
@@ -124,9 +126,7 @@ impl<'a> ModuleLoader<'a> {
   }
 
   #[allow(clippy::collection_is_never_read)]
-  async fn resolve_entries(
```
IWANABETHATGUY, line 127: image

crates/rolldown/tests/esbuild/import_star/import_star_export_star_omit_ambiguous/artifacts.snap
```diff
@@ -19,8 +19,8 @@ const y = 2
 
 
 var common_ns = {
-  get x() { return x },
-  get z() { return z }
+  get z() { return z },
```
IWANABETHATGUY, line 22: Not sure if the order matters here ,it seems that the fxhash map hash order is different between 1.73.0 and 1.71.0
## 1.73.0: 
image

## 1.71.0
image

crates/rolldown/src/bundler/module_loader/module_loader.rs
hyf0: Just use `#[allow] here.

crates/rolldown/tests/esbuild/import_star/import_star_export_star_omit_ambiguous/artifacts.snap
hyf0: Ignore this currenly.

hyf0 (approved):

hyf0:
Please use `squash and merge` to merge the PR.

PR #76, opened 2023-10-19
feat: support mix cjs and esm

underfin:
#### Implemention

1. The module will detect esm export syntax and cjs exports to make module syntax. 
2. The esm import and cjs export will be supported at both cjs & esm module.
3. If a module mix esm export and cjs exports, here will use esm export, the cjs export will be ignored and it will throw error at runtime.

Here has a different implement with esbuild, if a module hasn't esm export syntax and cjs exports, esbuild will mark it to cjs, but  fllowing the behavior will caused rollup behavior error, so rolldown will mark it to esm.

You can see detail at https://github.com/rolldown-rs/rolldown/issues/74. The warning i will add at next pr.

### Test Plan

Added.

---


crates/rolldown/src/bundler/visitors/scanner.rs
```diff
@@ -31,11 +31,14 @@ pub struct ScanResult {
 
 pub struct Scanner<'a> {
   pub idx: ModuleId,
+  pub module_type: ModuleType,
   pub scope: &'a mut ScopeTree,
   pub symbol_table: &'a mut SymbolTable,
   pub current_stmt_info: StmtInfo,
   pub result: ScanResult,
   pub unique_name: &'a str,
+  pub esm_export_key_word: Option<Span>,
```
hyf0: `keyword`

crates/rolldown/src/bundler/visitors/scanner.rs
```diff
@@ -52,16 +56,35 @@ impl<'a> Scanner<'a> {
...
+      module_type,
     }
   }
 
-  fn set_exports_kind(&mut self, exports_kind: ExportsKind) {
-    if let Some(resolution) = &self.result.exports_kind {
-      if resolution != &exports_kind {
-        // TODO shouldn't mix esm syntax and cjs syntax
-      }
+  fn set_esm_export_key_word(&mut self, span: Span) {
```
hyf0: `keyword`

hyf0 (approved):

PR #81, opened 2023-10-19
fix: add identifier for export default function without identifer

underfin:
### Description

The pr fixed some error with esm wrapped codegen and enable some esbuild passed test.
- add identifier for export default function without identifier
- avoid source mutation caused syntax error
```.ts
export default class Foo {} Foo.prop = 123
// before
var Foo = class Foo {} Foo.prop = 123  // syntax error
// pr
var Foo = class Foo {}\n Foo.prop = 123 // add `\n` to avoid error
```

### Test Plan
added.

---


hyf0:
on hold for CR

hyf0:
> ```ts
> export default class Foo {} Foo.prop = 123
> var Foo class Foo {}; Foo.prop = 123 // add `;` to avoid error
> ```

This looks confusing and invalid. What it means?

underfin:
I updated at description

hyf0:
@underfin  Are `export default class Foo`  being renamed to `var Foo = class Foo` expected? or just a temporary workaround?

underfin:
Esbuild will codegen `class Foo{}`  to `var Foo = class {}`, it is equals to `var Foo = class Foo{}`, so here not removed `Foo`.

hyf0 (approved):

PR #94, opened 2023-10-25
fix: rewrite unresolved symbol reference callee to indirect call

underfin:
### Description

Because the unresolved symbol reference is rewritten to property access, if it is a callee expression, here may change the function internally `this`, so need to rewrite it to an indirect call.

Also fixed the commonjs reference render.

### Test Plan
Updated.

---


hyf0 (approved):
cool

PR #98, opened 2023-10-25
chore: auto generate esbuild test scripts.

### LINKED ISSUE #78: Auto generated esbuild tests.
1. A tool automatically converts esbuild tests to snapshot test

IWANABETHATGUY:
### Description
1. the auto scripts look basically available, I regenerated the tests in `default` dir, the only difference is the entry name, but there are still some issues that need to be addressed.
-  Not all esbuild inline snapshots are coded as literal. e.g., the` files`  field referenced a top-level binding.
image
But luckily, only a few tests are written like this.

We could use this for now, and improve in the future.
2. Closed https://github.com/rolldown-rs/rolldown/issues/78

### Test Plan

---


hyf0:
We might need to config `test.config.json` to change the behaviors of rolldown to make the test passed. So `test.config.json` should only be created while there isn't one.

I'm not sure if we need to reference esbuild tests using git submodules, let's have a discussion of it.

Above all, well done👍

Boshen:
Nice job!

> I'm not sure if we need to reference esbuild tests using git submodules, let's have a discussion of it.

I don't think submodule is required because this is based on a script, let's just run it periodically.

yarn.lock
Boshen, line 1: Why are we using yarn?

hyf0 (approved):

scripts/bundler_default.go
Boshen, line 1: This is huge! shall we just download this from github? https://raw.githubusercontent.com/evanw/esbuild/main/internal/bundler_tests/bundler_default_test.go

scripts/genESbuildTest.js
Boshen, line 1: Can you add a top level documentation on what this does, and also how to run this?

yarn.lock
hyf0: yarn3 could works well with docker while pnpm doesnt for publishing packages. This is the experience from @Brooooooklyn.

scripts/genESbuildTest.js
IWANABETHATGUY: Sure

scripts/bundler_default.go
IWANABETHATGUY: yes,  I would remove this golang file, after I add some comments about how to use this scripts.

scripts/genESbuildTest.js
IWANABETHATGUY: Done

Boshen:
You may merge if CI is running these tests.

underfin (approved):

PR #104, opened 2023-10-26
refactor: use `Symbol` to store namespace info

hyf0:
### Description

Storing in `Symbol` is more fit into the current arch. The pr is also a preparation of https://github.com/rolldown-rs/rolldown/issues/103.

### Test Plan

---


underfin:
!bench

PR #128, opened 2023-10-30
chore: remove unnecessary println

hyf0:
### Test Plan

---


Boshen:
Shall we add `#![deny(clippy::print_stdout)]` to the crate?

PR #131, opened 2023-10-31
refactor: linking imports and exports

underfin:
### Description

Refactor linking imports and exports.

The new implement for linking imports and exports is following esbuild, it has some following steps.
1. Add the export named declaration to `ResolvedExports`, If there is a reexport named declaration `export { a} from 'a'`, it will be converted to `import { a } from 'a'; export { a }`.
2. Mark module has dynamic export, it due to export star from commonjs need to generate resolve symbol at runtime.
3. Add the export star declaration to `ResolvedExports`. Here should process symbol shadowing and potentially ambiguous export.
4. Linking imports to `ResolvedExports`.
5. Collect `exclude_ambiguous_sorted_resolved_exports`. It is excluding the ambiguous export  from`ResolvedExports`.

### Test Plan

Not need.

---

### DIFF
--- a/crates/rolldown/src/bundler/chunk/render_chunk_exports.rs
+++ b/crates/rolldown/src/bundler/chunk/render_chunk_exports.rs
@@ -19,7 +19,7 @@ impl Chunk {
         linking_info
           .exclude_ambiguous_resolved_exports
           .iter()
-          .map(|name| (name, linking_info.resolved_exports.get(name).unwrap().symbol_ref))
+          .map(|name| (name, linking_info.resolved_exports[name].symbol_ref))
           .collect::<Vec<_>>()
       },
     );
--- a/crates/rolldown/src/bundler/chunk/render_chunk_exports.rs
+++ b/crates/rolldown/src/bundler/chunk/render_chunk_exports.rs
@@ -1,6 +1,7 @@
+use oxc::span::Atom;
 use string_wizard::MagicString;
 
-use crate::bundler::graph::graph::Graph;
+use crate::bundler::graph::{graph::Graph, linker::is_ambiguous_export};
 
 use super::chunk::Chunk;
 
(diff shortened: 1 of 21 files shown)


crates/rolldown/tests/esbuild/import_star/import_export_star_ambiguous_warning/artifacts.snap
```diff
@@ -14,8 +14,9 @@ const x = 1
 const y = 2
 // common.js
 var common_ns = {
+  get x() { return x },
   get z() { return z },
-  get x() { return x }
+  get y() { return y }
```
hyf0: `y` should be shadowed.

underfin: wip status

crates/rolldown/src/bundler/chunk/render_chunk_exports.rs
```diff
@@ -6,29 +6,28 @@ use super::chunk::Chunk;
...
           .collect::<Vec<_>>()
       },
       |entry_module_id| {
         let linking_info = &graph.linking_infos[entry_module_id];
-        linking_info.resolved_exports.iter().collect::<Vec<_>>()
+        linking_info.exports().map(|(name, export)| (name, export.symbol_ref)).collect::<Vec<_>>()
       },
     );
 
-    export_items.sort_by_key(|(exported_name, _)| exported_name.as_str());
```
hyf0, line 23: Should we keep sorting?

underfin: It already sorted at `exclude_ambiguous_sorted_resolved_exports`.

underfin:
!bench

hyf0 (approved):

PR #141, opened 2023-11-01
fix: generate local symbol for esm import cjs instead of reference cjs namespace symbol

### LINKED ISSUE #127: Should not generate namespace binding for `ExportsKind::CommonJs`
Founded in https://github.com/rolldown-rs/rolldown/pull/126. See https://github.com/rolldown-rs/rolldown/blob/c18181625c3aae63bf2fe7bda0dd66ff33a283d5/crates/rolldown/tests/esbuild/splitting/shared-commonjs-into-es6/artifacts.snap for example.

Some modules try to import the namespace binding of a cjs module, which is not exist. This make `test_tests__esbuild__splitting__shared_commonjs_into_es6__test_config_json` failed. To make the test works,  https://github.com/rolldown-rs/rolldown/pull/126 make the cjs module generate namespace binding.

This is a temp workaround. I will investigate why some modules try to import the namespace binding of a cjs module.

underfin:
… namespace symbol

### Description
close https://github.com/rolldown-rs/rolldown/issues/127.
The pr fixed some bugs.
- fix: generate local symbol for esm import cjs instead of reference cjs namespace symbol
- fix: reExport stmt should reference self namespace symbol
- fix: hoisted module declaration for export all declaration and export name declaration which has source imported

### Test Plan
Updated.

---

### DIFF
--- a/crates/rolldown/src/bundler/chunk/render_chunk_exports.rs
+++ b/crates/rolldown/src/bundler/chunk/render_chunk_exports.rs
@@ -34,10 +34,9 @@ impl Chunk {
         let symbol = graph.symbols.get(canonical_ref);
         let canonical_name = &self.canonical_names[&canonical_ref];
         if let Some(ns_alias) = &symbol.namespace_alias {
-          if let Some(property_name) = &ns_alias.property_name {
-            let canonical_ns_name = &self.canonical_names[&ns_alias.namespace_ref];
-            s.append(format!("var {canonical_name} = {canonical_ns_name}.{property_name};\n"));
-          }
+          let canonical_ns_name = &self.canonical_names[&ns_alias.namespace_ref];
+          let property_name = &ns_alias.property_name;
+          s.append(format!("var {canonical_name} = {canonical_ns_name}.{property_name};\n"));
         }
         if canonical_name == exported_name {
           format!("{canonical_name}")
--- a/crates/rolldown/src/bundler/chunk/render_chunk_exports.rs
+++ b/crates/rolldown/src/bundler/chunk/render_chunk_exports.rs
@@ -34,11 +34,10 @@ impl Chunk {
         let symbol = graph.symbols.get(canonical_ref);
         let canonical_name = &self.canonical_names[&canonical_ref];
         if let Some(ns_alias) = &symbol.namespace_alias {
-          let canonical_ns_name = &self.canonical_names[&ns_alias.namespace_ref];
-          s.append(format!(
-            "var {canonical_name} = {canonical_ns_name}.{};\n",
-            ns_alias.property_name
-          ));
+          if let Some(property_name) = &ns_alias.property_name {
+            let canonical_ns_name = &self.canonical_names[&ns_alias.namespace_ref];
+            s.append(format!("var {canonical_name} = {canonical_ns_name}.{property_name};\n"));
+          }
         }
         if canonical_name == exported_name {
           format!("{canonical_name}")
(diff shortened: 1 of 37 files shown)


crates/rolldown/src/bundler/graph/linker.rs
```diff
@@ -408,12 +401,15 @@ impl<'graph> Linker<'graph> {
                   MatchImportKind::Found(symbol_ref) => {
                     symbols.union(info.imported_as, symbol_ref);
                   }
-                  MatchImportKind::NameSpace => {
+                  MatchImportKind::NameSpace(symbol_ref) => {
                     symbols.get_mut(info.imported_as).namespace_alias = Some(NamespaceAlias {
-                      property_name: info.imported.clone(),
-                      namespace_ref: importee.namespace_symbol,
+                      property_name: if info.is_imported_star {
+                        None
```
hyf0: Is this situation unreachable? If it's, there's no need to make `property_name` optional.

crates/rolldown/src/bundler/module/normal_module.rs
```diff
@@ -243,21 +244,18 @@ impl NormalModule {
     symbol_ref
   }
 
-  pub fn reference_symbol_in_facade_stmt_infos(
+  #[allow(clippy::map_entry, clippy::use_self)]
```
hyf0, line 247: It's suggested to not ignore lint rules for new added code.

crates/rolldown/src/bundler/graph/linker.rs
underfin: It can be reached. Before `import * as ns from 'cjs'` is used cjs namespace symbol, so `property_name` is unused at codegen. But the pr fixed this case, because here should refercend the local symbol `import_cjs`, and the correspond `property_name` should be `None`.

crates/rolldown/src/bundler/module/normal_module.rs
underfin: The `map_entry` should be ignored because the `self_linking_info` has owner ship problem, and the `use_self` is wired because it warning for `importee: &NormalModule`, it shouldn't be self.

crates/rolldown/src/bundler/chunk/render_chunk_exports.rs
```diff
@@ -34,11 +34,10 @@ impl Chunk {
...
         let canonical_name = &self.canonical_names[&canonical_ref];
         if let Some(ns_alias) = &symbol.namespace_alias {
-          let canonical_ns_name = &self.canonical_names[&ns_alias.namespace_ref];
-          s.append(format!(
-            "var {canonical_name} = {canonical_ns_name}.{};\n",
-            ns_alias.property_name
-          ));
+          if let Some(property_name) = &ns_alias.property_name {
+            let canonical_ns_name = &self.canonical_names[&ns_alias.namespace_ref];
+            s.append(format!("var {canonical_name} = {canonical_ns_name}.{property_name};\n"));
```
underfin: Here also has issue because local symbol not created, i will fix it at next pr

crates/rolldown/src/bundler/graph/linker.rs
hyf0: If that's the case, De we still need to use `NamespaceAlias`?

Could we just do?

```rust
if info.is_imported_star {
  symbols.union(info.imported_as, symbol_ref);
} else {
  symbols.get_mut(info.imported_as).namespace_alias = ...
}

underfin: Yeah. It is good for me.

hyf0: Would you mind reverting the change of `NamespaceAlias.property_ame` back to non-optional using the way I mentioned above? Other than that, the PR LGTM.

hyf0 (approved):

crates/rolldown/src/bundler/graph/linker.rs
underfin: Yeah. I do it.

PR #142, opened 2023-11-01
fix: create symbol for cjs export at entry

underfin:
### Description

Fix: create symbol for cjs export at entry

### Test Plan
Updated.

---

### DIFF
--- a/crates/rolldown/src/bundler/chunk/render_chunk_exports.rs
+++ b/crates/rolldown/src/bundler/chunk/render_chunk_exports.rs
@@ -11,14 +11,14 @@ impl Chunk {
         self
           .exports_to_other_chunks
           .iter()
-          .map(|(export_ref, alias)| (alias, *export_ref))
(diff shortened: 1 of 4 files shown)


hyf0 (approved):

crates/rolldown/src/bundler/chunk/render_chunk_exports.rs
```diff
@@ -29,15 +29,21 @@ impl Chunk {
     let mut s = MagicString::new("");
     let rendered_items = export_items
       .into_iter()
-      .map(|(exported_name, export_ref)| {
+      .map(|(exported_name, export_ref, entry_module_id)| {
         let canonical_ref = graph.symbols.par_canonical_ref_for(export_ref);
         let symbol = graph.symbols.get(canonical_ref);
-        let canonical_name = &self.canonical_names[&canonical_ref];
```
underfin, line 35: The name from `symbol_ref` is already correct.

PR #147, opened 2023-11-02
fix: module execute order

underfin:
### Description

Before the module children's modules execute order is mistaken. 

https://esbuild.github.io/try/#YgAwLjE5LjMALS1idW5kbGUgLS1mb3JtYXQ9Y2pzIC0tZXh0ZXJuYWw6YQBlAGVudHJ5LmpzAGltcG9ydCAnLi9hJwppbXBvcnQgJy4vYycAAGEuanMAY29uc29sZS5sb2coMSkAAGMuanMAY29uc29sZS5sb2coMik

Here has a esbuild example. 

``` .ts
// main.js
import 'a.js'
import 'b.js'
```

Here execute order should be `a.js -> b.js -> main.js`，The `a.js` should be executed before `b.js`.

Other using recursion is clearer than stack, so I rewrite it.

### Test Plan

Added

---


crates/rolldown/src/bundler/graph/graph.rs
```diff
@@ -32,62 +32,65 @@ impl Graph {
...
-            );
-          }
-        }
-        Action::Exit(id) => {
-          *module.exec_order_mut() = next_exec_order;
-          next_exec_order += 1;
-          sorted_modules.push(id);
-        }
-      }
+    let mut visited_modules = FxHashSet::default();
```
hyf0, line 39: No need to rewrite this function, just one line change would be enough. https://github.com/rolldown-rs/rolldown/pull/148/files

PR #178, opened 2023-11-06
feat: wasm

### LINKED ISSUE #63: Wasm Build
Let's check for wasm build early so we don't get caught up with technical debt when we start shipping wasm.

See https://github.com/web-infra-dev/oxc/blob/3605f7c0fc2421496a869c2b8fb470340903a21f/.github/workflows/ci.yml#L57

houyunlu:
### Description
1. Closed https://github.com/rolldown-rs/rolldown/issues/63
2. FInished wasm32 target compilation

### Test Plan

---


crates/rolldown-wasm/Cargo.toml
hyf0, line 1: should be underscope.

houyunlu: ok

crates/rolldown_utils/src/lib.rs
```diff
@@ -1,20 +1,5 @@
 mod magic_string_ext;
-use std::future::Future;
 
-use async_scoped::TokioScope;
 pub mod reserved_word;
 
-pub fn block_on_spawn_all<Iter, Out>(iter: Iter) -> Vec<Out>
```
houyunlu, line 7: explained before.

crates/rolldown/src/bundler/module_loader/module_loader.rs
```diff
@@ -113,11 +112,11 @@ impl<'a, T: FileSystemExt + 'static> ModuleLoader<'a, T> {
   }
 
   #[allow(clippy::collection_is_never_read)]
-  fn resolve_entries(&mut self) -> BatchedResult<Vec<(Option<String>, ResolvedRequestInfo)>> {
+  async fn resolve_entries(&mut self) -> BatchedResult<Vec<(Option<String>, ResolvedRequestInfo)>> {
     let resolver = &self.resolver;
 
     let resolved_ids =
-      block_on_spawn_all(self.input_options.input.iter().map(|input_item| async move {
+      futures::future::join_all(self.input_options.input.iter().map(|input_item| async move {
```
houyunlu, line 197: Use `futures::join_all` to avoid introducing another crate, they have same effect.

hyf0:
Would you mind on hold this pr before https://github.com/rolldown-rs/rolldown/pull/182 being merged?

crates/rolldown/src/bundler/module_loader/module_loader.rs
hyf0: I'm not sure. `block_on_spawn_all` would run tasks in parallel while `join_all` seems to run them concurrently.

houyunlu:
!bench

hyf0 (approved):

PR #192, opened 2023-11-07
chore: use vite to build playground

houyunlu:
### Test Plan

---


hyf0 (approved):

web/playground/index.js
```diff
@@ -1,5 +1,5 @@
-import { bundle, FileItem, AssetItem } from 'rolldown-wasm'
-
+import init, { bundle, FileItem, AssetItem } from '../wasm'
+await init();
```
houyunlu, line 2: Use relative path instead of workspace package because of vite limitation, for more details you could refer to https://github.com/vitejs/vite/issues/8427

PR #210, opened 2023-11-09
fix: async_scope_block_on

### LINKED ISSUE #183: Use `block_on_spawn_all` instead of `join_all` to resolve entries
I'm not sure. `block_on_spawn_all` would run tasks in parallel while `join_all` seems to run them concurrently.

_Originally posted by @hyf0 in https://github.com/rolldown-rs/rolldown/pull/178#discussion_r1383294824_

houyunlu:
### Description
closed https://github.com/rolldown-rs/rolldown/issues/183

### Test Plan

---


hyf0 (approved):
Cool

PR #225, opened 2023-11-11
refactor: directly convert binding option to bundler option

underfin:
### Test Plan

---


underfin:
* main
  * **PR #225** <a href="https://app.graphite.dev/github/pr/rolldown-rs/rolldown/225?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/rolldown-rs/rolldown/225?utm_source=stack-comment).

crates/rolldown/src/bundler/options/input_options.rs
```diff
@@ -18,6 +18,6 @@ impl From<String> for InputItem {
 #[derive(Derivative)]
 #[derivative(Debug)]
 pub struct InputOptions {
```
hyf0, line 20: Hi, nice work.

Let me add some contexts about the old code.

- Rolldown allow to be used as a rust crate, so we need care about the dx of using rolldown directly in rust.
- Rolldown has a lot of options. In most cases, we don't want users to bother them. So, we want to users could write something like

```rust
let build = rolldown(InputOptions {
  main: "..",
  ..Default::default()
})
```
- Those optional fields make it easily to use `derive(Default)` on `InputOptions`.
- The reason why don't use `impl Default` is that the `default` function signature are limited, which could not accept other params other than `self`.
  - This disable us to generate default value depend on outside value. For example, we may generate different default `OutputOptions` depend on the different `InputOptions`. But the `default` function in `trait Default` does not accept other params.

May I ask the reason of this refactor? I may miss your point, let's see how we could achieve your goal.

underfin: You can understand this for the bellowing reasons.
1. The `input` of `InputOptions`  shouldn't be `Option` for rust usage, maybe other fields can be `Option` or `impl Defalut`, here the `InputOptions` better usage should be like 
``` .rs
let build = rolldown(InputOptions::new(input, // other fileds))
```
The `OutputOptions` is also same as `InputOptions`, but look like it can be `Default` at now.
2. It avoids the rust crate has some logics of some options to convert, it is better than before.

underfin: Other I add `impl Default for InputOptions` at now, but I'm not sure either api regression at future. But I always don't think as a required option to `Option` value is a good point, because you always need `unwrap` it and give a default value to it. It is directly for move the logic to `impl Default`.

hyf0: > The input of InputOptions shouldn't be Option for rust usage

I agree with you. `Vec<InputItem>` is better than `Option<Vec<InputItem>>` and this should be the only exception.

> here the `InputOptions` better usage should be like 
> ``` .rs
> let build = rolldown(InputOptions::new(input, // other fileds))
> ```

`InputOptions` is meant to be a plain struct. If users need to create options from a function, I would said that I haven't seen this pattern from rollup, webpack or other bundlers. Especially, considering we would have many fields in `InputOptions`.  `rolldown(InputOptions::new(input, a, b, c, d, e, f))` might not be better option compared to

```rust
let build = rolldown(InputOptions {
  input: vec![...],
  ..Default::default(),
});
```

> The `OutputOptions` is also same as `InputOptions`, but look like it can be `Default` at now.

From what said above:

This disable us to generate default value depend on outside value. For example, we may generate different default `OutputOptions` depend on the different `InputOptions`. But the `default` function in `trait Default` does not accept other params.

For example, we might need to resolve to different output format depend on if user enable compatible mode.

In this case, user would write code like 

```rust
let build = rolldown(InputOptions {
  input: vec![...],
  compatible: true,
  ..Default::default(),
});

build.write(OutputOptions {
  dir: Some("..."),
  ..Default::default()
});

```

and we will normalize it in `normalize_output_options` with `compatible: true` or `false`. And resolve to different output format(`cjs` or `esm`) if user not set `format` option.

If user explicit set `format` option, such as

```rust
build.write(OutputOptions {
  dir: Some("..."),
  format: Some("esm")
  ..Default::default()
});
```

Then, rolldown could be told that user explicitly set `format` option for `format` is `Some` instead of `None`.

The PR currently removes those `Option`, which make detecting  if user explicitly set some options impossible.

underfin: > The PR currently removes those Option, which make detecting if user explicitly set some options impossible.

If a required option could be `Option`, you always need it and give it a default value. Like this 
``` .rs
OutputOptions {
  format: Some("esm")
  ..Default::default()
}
NormaliedOutputOptions {
  format: options.format.unwrap_or("esm")
}
```
The Better idea should be 
``` .rs
OutputOptions {
  format: "esm",
  ..Default::default()
}
impl  Defalut for OutputOptions {
   fn default ()-> Self {
       format: "esm"
   }
}
```
It just moved the convert options logic to `impl Default`, it is better than convert options.

crates/rolldown/src/bundler/options/output_options.rs
```diff
@@ -1,15 +1,28 @@
...
+  pub format: OutputFormat,
+}
+
+impl Default for OutputOptions {
+  fn default() -> Self {
+    Self {
+      entry_file_names: FileNameTemplate::from("[name].js".to_string()),
+      chunk_file_names: FileNameTemplate::from("[name]-[hash].js".to_string()),
+      dir: "dist".into(),
+      format: OutputFormat::Esm,
```
hyf0, line 25: We might need to resolve to different `format` depend of if user set `InputOptions.compatable` to `true`. How do we do this using `fn default`

crates/rolldown/src/bundler/options/input_options.rs
hyf0: Let me try to be more clear about this. https://github.com/rolldown-rs/rolldown/pull/225/files#r1390204921

crates/rolldown/src/bundler/options/output_options.rs
underfin: Yeah. It can't be done by default. But in other words, if has this situation, the `format` value should be removed at  `OutputOptions`. Because the result controlled by many options, we need to avoid the situation.

underfin: Can you give a actually option like this at rollup options?

underfin: Actually I do it because the rust crate option normalized option is extra from binding option convert. The binding option convert result should be final option result. It is worked for 99% users, I think we should make code is easy to review for this situation. 
And then the user from rust side, we may be consider export some wrapped options at wrapped crate. The rolldown core crate should be clear for most situation.

hyf0: > Can you give a actually option like this at rollup options?

Rollup now normalize the output options depending on the input options. You could find this logic from https://github.com/rollup/rollup/blob/52c55bb1e17154ae6d01fb40e0e4a3589bc20a8f/src/rollup/rollup.ts#L232-L255.

And `rolldown` is the final crate I'm thinking to expose to rust users. I'm sorry I don't tell you this in the first place.

Anyway, thanks for your great work.

hyf0 (approved):

crates/rolldown/src/bundler/options/output_options.rs
underfin: > You could find this logic from https://github.com/rollup/rollup/blob/52c55bb1e17154ae6d01fb40e0e4a3589bc20a8f/src/rollup/rollup.ts#L232-L255.

OK. It can be done at binding crate, not at `OutputOptions Defalut`. The core should export actually options which is buildler used. The options convert should be at binding crate for node/wasm, rust side API also should follow it.

PR #234, opened 2023-11-12
feat:  use @rolldown/core to bench with esbuild

### LINKED ISSUE #174: Benchmark with esbuild using `@rolldown/core`

houyunlu:
### Description
1. Closed https://github.com/rolldown-rs/rolldown/issues/174

### Test Plan

---

### DIFF
--- a/bench/esbuild.js
+++ b/bench/esbuild.js
@@ -1,0 +1,14 @@
+const esbuild = require('esbuild')
+const path = require('path')
+
+async function esbuildBundle(entryPoints) {
+  await esbuild.build({
+    entryPoints: entryPoints,
+    bundle: true,
+    outfile: 'out.js',
+  })
+}
+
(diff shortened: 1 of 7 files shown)


bench/esbuild.js
```diff
@@ -0,0 +1,14 @@
+const esbuild = require('esbuild')
+const path = require('path')
+
+async function esbuildBundle(entryPoints) {
+  await esbuild.build({
+    entryPoints: entryPoints,
+    bundle: true,
+    outfile: 'out.js',
```
underfin, line 8: Esbuild here write result to file, but rollldown not.

hyf0 (approved):

bench/esbuild.js
hyf0: Let's merge it first, and solve it later.

PR #261, opened 2023-11-14
feat: add ViteScannerPlugin

underfin:
### Test Plan

---

PR #262, opened 2023-11-14
feat: add native plugin example in rust

hyf0:
### Description

This PR is the first part of how to write a native plugin for rolldown and expose it in node binding.

### Test Plan

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1588,6 +1588,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "rolldown_plugin_hello"
+version = "0.1.0"
+dependencies = [
+ "async-trait",
+ "rolldown",
+]
+
+[[package]]
 name = "rolldown_plugin_vite_scanner"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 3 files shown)


hyf0:
* main
  * **PR #262** <a href="https://app.graphite.dev/github/pr/rolldown-rs/rolldown/262?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/rolldown-rs/rolldown/262?utm_source=stack-comment).

underfin:
https://github.com/rolldown-rs/rolldown/pull/261 I think the plugin could be a good example?

hyf0:
> #261 I think the plugin could be a good example?

Yeah. I look into it and found out that it's too unique to be an general purpose example.

Since it's for vite only, the way users use it and how we publish it is most likely different from the native plugins for general purpose that rolldown provide.

underfin:
It is a empty crate and unused, maybe adding it to development-native-plugin-guide is better?

hyf0:
> It is a empty crate and unused, maybe adding it to development-native-plugin-guide is better?

Thanks for the suggestion. Sorry I don't add many contexts, it's going to be used in the future.

It's part of how to write a native plugin for rolldown and expose it in node binding.

underfin (approved):

PR #263, opened 2023-11-14
fix: playground panic

houyunlu:
### Description
I don't know why, but with main branch code, playground panic when try to build.
image

### Test Plan

---


crates/rolldown_binding_wasm/src/lib.rs
```diff
@@ -61,7 +61,7 @@ pub fn bundle(file_list: Vec<FileItem>) -> Vec<AssetItem> {
         })
         .collect::<Vec<_>>();
       let mut bundler = Bundler::with_plugins_and_fs(
-        InputOptions { input, cwd: "/".into(), ..Default::default() },
```
hyf0, line 64: I'm ok with this fix. But I wonder does the default external option panic?

houyunlu: yeah

hyf0: That's wired. Let's create an issue for future investigating.

hyf0 (approved):

crates/rolldown_binding_wasm/src/lib.rs
JoviDeCroock: I would assume this is because the underlying default impl calls `std::env::current_dir()` which indeed isn't supported in i.e. the browser

hyf0: @JoviDeCroock  Good catch! Highly possible.

houyunlu: > I would assume this is because the underlying default impl calls `std::env::current_dir()` which indeed isn't supported in i.e. the browser

yeah, you are right.

PR #281, opened 2023-11-15
feat(rust): add `print` in `OxcCompiler`

hyf0:
### Test Plan

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1114,6 +1114,7 @@ checksum = "2a72360715a0bac81660a30a00c402448706ccc09abe102b625f0d4bb66cc526"
 dependencies = [
  "oxc_allocator",
  "oxc_ast",
+ "oxc_codegen",
  "oxc_diagnostics",
  "oxc_index",
  "oxc_parser",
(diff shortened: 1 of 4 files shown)


hyf0:
* main
  * **PR #281** <a href="https://app.graphite.dev/github/pr/rolldown-rs/rolldown/281?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/rolldown-rs/rolldown/281?utm_source=stack-comment).

Boshen:
Why do we need codegen again?

hyf0:
> Why do we need codegen again?

We plan to add some pre-transpile operations for some thing like contestants folding. After transpiling, we need to transform the ast to string, and use the string as the input for bundling. This make sure the ast rolldown scans have no inconsistency between the ast and the source code.

Boshen:
> > Why do we need codegen again?
> 
> We plan to add some pre-transpile operations for some thing like contestants folding. After transpiling, we need to transform the ast to string, and use the string as the input for bundling. This make sure the ast rolldown scans have no inconsistency between the ast and the source code.

Thanks! So this is preparation for the transformer.

hyf0:
> > > Why do we need codegen again?
> > 
> > 
> > We plan to add some pre-transpile operations for some thing like contestants folding. After transpiling, we need to transform the ast to string, and use the string as the input for bundling. This make sure the ast rolldown scans have no inconsistency between the ast and the source code.
> 
> Thanks! So this is preparation for the transformer.

Yeah. Many transform functionality require this. TypeScript, JSX and so on. Also for some internal plugins using oxc, rolldown  only support passing around string between transform hooks, which also requires this.

underfin (approved):

PR #286, opened 2023-11-15
chore: Replace anyhow/ariadne with miette for diagnostic errors

### LINKED ISSUE #154: Diagnostics for errors/warnings
I had a prototype for this.

For unresolved entry error, we would have a `Dianostic` to show

image

rather than the current

https://github.com/rolldown-rs/rolldown/blob/0c9ab02bd89da358a0d9fb9719e85cd63d0f4aad/crates/rolldown/tests/fixtures/errors/unresolved_entry/artifacts.snap#L8-L12

milesj:
### Description

Integrates miette for errors + diagnostics: https://docs.rs/miette/5.10.0/miette/

Fixes: https://github.com/rolldown-rs/rolldown/issues/154

### Test Plan

---


crates/rolldown/src/bundler/bundler.rs
```diff
@@ -97,7 +97,7 @@ impl<T: FileSystem + Default + 'static> Bundler<T> {
         .plugin_driver
         .build_end(Some(&HookBuildEndArgs {
           // TODO(hyf0): 1.Need a better way to expose the error
-          error: format!("{:?}\n{:?}", error.code(), error.to_diagnostic().print_to_string()),
+          error: error.to_string(),
```
milesj, line 100: Need to verify this but should work since diagnostics just use `Display`

crates/rolldown_error/src/error/external_entry.rs
```diff
@@ -1,9 +1,11 @@
+use crate::PathExt;
+use miette::Diagnostic;
 use std::path::PathBuf;
-
 use thiserror::Error;
 
-#[derive(Error, Debug)]
-#[error("Entry module \"{:?}\" cannot be external.", id)]
+#[derive(Error, Debug, Diagnostic)]
+#[diagnostic(code = "UNRESOLVED_ENTRY")]
```
milesj, line 7: Tried to use `error_codes` but miette only supports literal strings

crates/rolldown_error/src/error/mod.rs
```diff
@@ -4,46 +4,45 @@ use std::{
...
   Io(Box<std::io::Error>),
+
   // TODO: probably should remove this error
+  #[diagnostic(code = "NAPI_ERROR")]
   #[error("Napi error: {status}: {reason}")]
   Napi { status: String, reason: String },
 }
 
 impl BuildError {
-  pub fn code(&self) -> &'static str {
```
milesj, line 38: `Diagnostic` defines the same method, so using that instead

crates/rolldown_fs/src/file_system.rs
```diff
@@ -16,17 +16,17 @@ pub trait FileSystem: Send + Sync + OxcResolverFileSystem {
   /// # Errors
   ///
   /// * See [std::fs::remove_dir_all]
-  fn remove_dir_all(&self, path: &Path) -> io::Result<()>;
+  fn remove_dir_all(&self, path: &Path) -> miette::Result<()>;
```
milesj: Unsure of this change. Its nice to use miette everywhere however.

Cargo.toml
```diff
@@ -29,8 +28,8 @@ async-trait    = "0.1.74"
 futures        = "0.3.29"
 thiserror      = "1.0.50"
 smallvec       = "1.11.1"
-ariadne        = "0.3.0"
 async-scoped   = { version = "0.7.1" }
 regex          = "1.10.2"
 once_cell      = "1.18.0"
 dashmap        = "5.5.3"
+miette         = { version = "5.10.0", features = ["fancy"] }
```
Boshen: I'd use `fancy-no-backtrace`. See context: `https://github.com/zkat/miette/issues/132`

Boshen (commented):
I see some commented out code, everything else is 👍

crates/rolldown_fs/src/file_system.rs
hyf0: Yeah. I think we should keep this part. The file system is just a wrapper of real filesystem. It would be better that they keep the same function signature.

crates/rolldown_error/src/error/external_entry.rs
hyf0: Yeah. I met the same problem :). But I think is fine now.

hyf0 (approved):
Cool

PR #295, opened 2023-11-16
chore: Upgrade to Yarn v4.

milesj:
### Description

Was going through the testing doc and noticed we're still on v3.

### Test Plan

Just that build commands still work (they do locally).

---


.yarnrc.yml
```diff
@@ -1,7 +1,7 @@
-nodeLinker: node-modules
+compressionLevel: mixed
+
+enableGlobalCache: false
 
-plugins:
```
milesj, line 3: built into v4 now

hyf0:
Cool. LGTM. 

---

IMHO, I notice we almost add `--all` in every `yarn workspaces foreach` command. Not sure why yarn don't make it by deault.

package.json
```diff
@@ -12,11 +12,11 @@
   "scripts": {
     "lint": "oxlint --ignore-path=./.oxlintignore --import-plugin",
     "lint-filename": "ls-lint",
-    "build": "yarn workspaces foreach --topological-dev run build",
+    "build": "yarn workspaces foreach --all --parallel --topological-dev run build",
```
hyf0, line 15: I'm not familiar with yarn4. Is `--parallel`conflicted with `--topological-dev`?

milesj: I'm not entirely sure, but since we only have 2 packages right now, it works fine. Was curious if it did anything special.

hyf0: Yeah. `build` requires `topological-dev` for buiding `@rolldown/node` require building `@rolldown/node-binding` correctly.

If things are run in parallel, you might get building `@rolldown/node` successfully with `@rolldown/node-binding` building failed.

This cause `@rolldown/node` rely on the old version of `@rolldown/node-binding` and people might hard to notice it. Things could happen to run but with unexpected behaviors. This also apply for testing, which is a package rely on `@rolldown/node`.

milesj: Looks like it works then because node-binding always finished first.

hyf0: Yeah. It might happen to works. The key is that if `@rolldown/node-binding` builds failed, `@rolldown/node` should also stop the build process. `--parallel` breaks the constrains.

hyf0: I just tested locally. `--topological-dev` overwrites the `--parallel`😂.

milesj: Good to know! hah

PR #304, opened 2023-11-17
chore: ignore custom justfile

houyunlu:
### Description
1. Allow users to use custom just file extension,  this can avoid violating project scripts convention.

### Test Plan

---


hyf0 (approved):
Cool. So it's has local level config.

PR #318, opened 2023-11-17
ci: Enable cache warmup.

milesj:
### Description

Context: https://github.com/rolldown-rs/rolldown/issues/71

https://github.com/moonrepo/setup-rust/pull/16

### Test Plan

---

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -63,7 +63,7 @@ jobs:
       - name: Install Rust
         uses: moonrepo/setup-rust@v1
         with:
-          cache: false
+          cache-base: main
           components: clippy
 
       - name: Run Clippy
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -63,6 +63,7 @@ jobs:
       - name: Install Rust
         uses: moonrepo/setup-rust@v1
         with:
+          cache: false
```
milesj: Feel like we dont need caching for linting?

Boshen: It seems like crates are compiled for linting.

milesj: yeah wanted to see how long it took, seems to be lengthy. added cache back

hyf0 (approved):

PR #342, opened 2023-11-20
chore(wasm): add missing wasm-pack dep

sxzz:
### Description

Add the missing `wasm-pack` dependency in `@rolldown/wasm-binding` package. If I didn't install `wasm_pack` as global dep, the execution fails.

```sh
❯ nr build:wasm
command not found: wasm-pack
ERROR: "build:wasm" exited with 127.
```

### Test Plan

---

### DIFF
--- a/crates/rolldown_binding_wasm/package.json
+++ b/crates/rolldown_binding_wasm/package.json
@@ -26,6 +26,7 @@
     "./snippets/*"
   ],
   "devDependencies": {
-    "npm-run-all": "^4.1.5"
+    "npm-run-all": "^4.1.5",
+    "wasm-pack": "^0.12.1"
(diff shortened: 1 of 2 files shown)


hyf0 (approved):
We have this actually in https://github.com/rolldown-rs/rolldown/blob/8bfdcc99af9ad8c0d3503b5098571d316654f3ef/justfile#L7.

But I didn't thought `wasm-pack` could be installed by npm. This should be more friendly than using binstall. Cool.

sxzz:
I installed it before, but the older version didn't support workspace inheritance. Causes compilation errors. So I just uninstalled the global version. I think the version should be maintained by the project itself.

PR #346, opened 2023-11-21
Add a script to improve node/binding builds

milesj:
### Description

Was looking into how we can improve the constant rebuilding of the node package, and the node/wasm binding crates. This is kind of a hard problem to solve, and I don't think I actually solved it, but instead added a script to reduce the friction.

The problem is that any small change to either the rust or js code would require a new build, so we simply couldn't cache the output of anything, and reuse that cache in subsequent builds.

Instead, I opted to go with a hashing pattern that would _only build or re-build_ if there are changes since the last build. This will help with unnecessarily running cargo builds (which can take some time even for no changes), or re-running the entire node package build.

This also adds a `--watch` mode that you can run in the background when running tests/flows.

### Test Plan

Pull down the branch and test?


crates/rolldown_binding/Cargo.toml
```diff
@@ -2,12 +2,10 @@
 name    = "rolldown_binding"
 version = "0.0.1"
 
-
 homepage.workspace   = true
 license.workspace    = true
 edition.workspace    = true
 repository.workspace = true
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
milesj, line 10: me just testing the watcher

scripts/gen-esbuild-test.js
```diff
@@ -1,53 +1,53 @@
-const Parser = require("tree-sitter");
-const Go = require("tree-sitter-go");
-const fs = require("fs-extra");
-const path = require("path");
-const changeCase = require("change-case");
-const chalk = require("chalk");
-const dedent = require("dedent");
+const Parser = require('tree-sitter')
```
milesj, line 1: just from removing the prettier ignore

hyf0:
Awesome. Could you also add descriptions about how to use this in https://github.com/rolldown-rs/rolldown/tree/main/CONTRIBUTING.md.

milesj:
Yah can do!

hyf0 (approved):

PR #395, opened 2023-12-13
fix: support browser filed false value

underfin:
### Description

You can simply prevent a module or file from being loaded into a bundle by specifying a value of ```false``` for any of the keys. This is useful if you know certain codepaths will not be executed client side but find it awkward to split up or change the code structure.

```javascript
"browser": {
    "module-a": false,
    "./server/only.js": "./shims/server-only.js"
}
```

The above will cause the following to return an empty object into `a`.
```javascript
var a = require('module-a');
```

ref https://github.com/defunctzombie/package-browser-field-spec?tab=readme-ov-file#ignore-a-module

### Test Plan

Added.

---

### DIFF
--- a/crates/rolldown_common/src/file_path.rs
+++ b/crates/rolldown_common/src/file_path.rs
@@ -9,21 +9,29 @@ use regex::Regex;
 use sugar_path::{AsPath, SugarPath};
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
-pub struct FilePath(Arc<str>);
+pub struct FilePath((Arc<str>, bool /* ignored (used to browser filed "false" value) */));
 
 impl FilePath {
   pub fn new(value: impl Into<String>) -> Self {
-    Self(value.into().into())
+    Self((value.into().into(), false))
+  }
+
+  pub fn with_ignored(value: impl Into<String>) -> Self {
+    Self((value.into().into(), true))
   }
 
   pub fn as_str(&self) -> &str {
-    &self.0
+    &self.0 .0
+  }
+
+  pub fn is_ignored(&self) -> bool {
+    self.0 .1
   }
 }
 
 impl AsRef<str> for FilePath {
   fn as_ref(&self) -> &str {
-    &self.0
+    &self.0 .0
   }
 }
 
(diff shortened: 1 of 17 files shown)


underfin:
* `main`
  * **PR #395** <a href="https://app.graphite.dev/github/pr/rolldown-rs/rolldown/395?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/rolldown-rs/rolldown/395?utm_source=stack-comment).

crates/rolldown_common/src/file_path.rs
```diff
@@ -9,29 +9,37 @@ use regex::Regex;
 use sugar_path::{AsPath, SugarPath};
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
-pub struct FilePath(Arc<str>);
+pub struct FilePath((Arc<str>, bool /* ignored (used to browser filed "false" value) */));
```
hyf0: `FilePath` is just simple new-type for hiding internal data strcutrues. It's not the place to add this extra `bool`.

underfin: Here need a struct to include real `path` and path flags, add the path flags to here to avoid add a new-type. Esbuild also do it,see https://github.com/evanw/esbuild/blob/main/internal/logger/logger.go#L258. The flag is same `bool`, making the file path is an ignored path.

hyf0: Simply "esbuild do it too" is not a solid reason.

Boshen: Let's add a new struct

```rust

struct ResolvedPath {
  path: FilePath,
  ignored: bool,
  // package_json: ... you'll need this later for side effects
}

```

Remember ... Rust's 精髓 is 包一层 🤡

underfin: Here I add a new `ResolvedPath` struct, but actually, the feature is the same as `FilePath`, here code is wrapped as you said...

hyf0:
Allow me to get more time to figure out what `xxx: false` is and how other bundlers handle it.

Boshen:
> Allow me to get more time to figure out what `xxx: false` is and how other bundlers handle it.

https://github.com/microsoft/TypeScript/blob/2e3b8af85e1f782634349139819151881369e3ea/package.json#L100-L110

It tells the bundler to ignore resolving these node modules.

hyf0 (approved):

PR #399, opened 2023-12-27
feat: plugin context binding

underfin:
### Description

The rollup has tree kind of plugin context, here has some description.

- `PluginContext`, used at `buildStart/load/resolveId...` hooks
- `TransformPluginContext`, used at `transform` hook
- `MminimalPluginContext`, used at `options` hook

The derived relationship is `MminimalPluginContext` -> `PluginContext` -> `TransformPluginContext`, the `MminimalPluginContext` does not need implemented at the rust side because `options` hook is at js side. So the rust side only implement `PluginContext`/`TransformPluginContext` and binding.

The `PluginContext` is created for use in one the plugin's different hooks, because it's `cache` filed is isolate, ref https://github.com/rollup/rollup/blob/master/src/utils/PluginDriver.ts#L108.

The `TransformPluginContext` is derived from  `PluginContext`, it extends some methods and also overwrite some methods, you can see https://github.com/rollup/rollup/blob/master/src/utils/transform.ts#L108.

### Implement

The rust side is also implements `PluginContext`  and `TransformPluginContext`, using `PluginContext` to create `TransformPluginContext` and custom some methods. 

Here move `PluginContext` to immutable because its `cache` need to make sure it thread-safe.

### Test Plan

Not need at now.

---

### DIFF
--- a/crates/rolldown/src/bundler/plugin_driver/mod.rs
+++ b/crates/rolldown/src/bundler/plugin_driver/mod.rs
@@ -12,24 +12,26 @@ use crate::{
 pub type SharedPluginDriver = Arc<PluginDriver>;
 
 pub struct PluginDriver {
-  plugins: Vec<BoxPlugin>,
+  plugins: Vec<(BoxPlugin, PluginContext)>,
 }
 
 impl PluginDriver {
   pub fn new(plugins: Vec<BoxPlugin>) -> Self {
-    Self { plugins }
+    Self {
+      plugins: plugins.into_iter().map(|plugin| (plugin, PluginContext::new())).collect::<Vec<_>>(),
+    }
   }
 
   pub async fn build_start(&self) -> HookNoopReturn {
-    for plugin in &self.plugins {
-      plugin.build_start(&mut PluginContext::new()).await?;
+    for (plugin, ctx) in &self.plugins {
+      plugin.build_start(ctx).await?;
     }
     Ok(())
   }
 
   pub async fn resolve_id(&self, args: &HookResolveIdArgs<'_>) -> HookResolveIdReturn {
-    for plugin in &self.plugins {
-      if let Some(r) = plugin.resolve_id(&mut PluginContext::new(), args).await? {
+    for (plugin, ctx) in &self.plugins {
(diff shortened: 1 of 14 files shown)


crates/rolldown/src/bundler/plugin_driver/mod.rs
```diff
@@ -12,51 +12,53 @@ use crate::{
 pub type SharedPluginDriver = Arc<PluginDriver>;
 
 pub struct PluginDriver {
-  plugins: Vec<BoxPlugin>,
+  plugins: Vec<(BoxPlugin, PluginContext)>,
```
milesj, line 16: Is context different per plugin? Shouldn't context be the same across all plugins?

underfin: Yeah. The plugin context is different, you can see here https://github.com/rollup/rollup/blob/master/src/utils/PluginDriver.ts#L108.

underfin:
* **#400** <a href="https://app.graphite.dev/github/pr/rolldown-rs/rolldown/400?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#399** <a href="https://app.graphite.dev/github/pr/rolldown-rs/rolldown/399?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

hyf0:
I'm going to on hold thid PR until we have real-world usages to verify if the changes are meaningful. Especially, some code-deleting are not understandable to me.

milesj (approved):
@underfin Changes here look good, is there any other pieces missing before this lands?

underfin:
> @underfin Changes here look good, is there any other pieces missing before this lands?

Need review from @hyf0 .

hyf0:
I just looked around all PRs related to plugin. They are ok under the current plugin system code. But there are some problems in the current plugin system code and these PRs are just patch the bad code and add features on top of it. That's bad for having a solid ground.

So I will remove the bad code and come up with new one. So I'm gonna keep holding up these PRs before the new plugin system code.

PR #406, opened 2024-01-10
[WIP] Implement hook filtering

milesj:
### Description

RFC: https://insider.affine.pro/workspace/4cb4fc09-57fa-4be8-8b17-f849c060beab/Cedt2CZvM_QJ6yy5WomcP

### Test Plan

TBD

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1558,6 +1558,7 @@ dependencies = [
  "napi",
  "napi-build",
  "napi-derive",
+ "once_cell",
  "rolldown",
  "rolldown_error",
  "rolldown_fs",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1604,6 +1604,7 @@ dependencies = [
  "napi-build",
  "napi-derive",
  "once_cell",
+ "regex",
  "rolldown",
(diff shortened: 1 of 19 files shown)


hyf0:
Cool. So much inspiration. I also add some comments in that RFC doc and it would be great to hear you thoughts.

milesj:
@hyf0 Continuing the convo here since Affine doesn't have comments. Makes it too hard.

1. Regex or Glob?
For perf, globs use regex under the hood anyways, so with either approach, it will still be the same perf. But also depends on how complicated the glob -> regex pattern looks like under the hood. 

I would say regex is more powerful than glob as a whole, but for file paths, I believe globs hit like 99% of use cases, and are just an overall better dev ex.

3. Pattern?
I originally thought of using that approach, or something similar, but after auditing many rollup/vite plugins, the same filter was used for all plugin hooks. We don't really need per-hook filtering, it's mainly per-plugin as a whole.

Open to another name/impl. I went with `filterId` since it's explicitly, it's filtering by ID (path).

2/4. Name/implementation
Both of these depend on which pattern we go with.

PR #407, opened 2024-01-12
chore(binding): separate plugin

underfin:
### Description

Because the `OutputPlugin` is also derived from `Plugin`, so here move out `inputOptions` dir.

### Test Plan

Not need

---

PR #408, opened 2024-01-12
feat: support output options plugin

underfin:
#### OutputPlugin binding

The pr intend to find a solution to support `OutputPlugin`, but here has some issue need to resolve.

The `write/generate` is async function, it isn't allow has `PluginOptions` argument, because the `JsFunction` isn't `Send`, here has some solution to resolve it.

 - We can using `ThreadsafeFunction` to replace `JsFunction` for `PluginOptions`, see this issue https://github.com/napi-rs/napi-rs/issues/1644, but the `ThreadsafeFunction` isn't `ToNapiValue`, we only can be using it directly as argument at the async function, it is ugly for many plugin hook.
 - Replace `write/generate` to sync function, but it will block node main thread work. Here I also try using `AsyncTask` to rewrite `write/generate` function, but it also has issue with the `JsFunction` isn't `Send`.
 
 So here is a solution to using a sync function `set_output_plugins`  to pass `PluginOptions` to rust side before call `write/generate`. Here use a map to store `plugin` using the sync function and take it to the `write/generate` call.

#### Rust side OutputPluginDriver  and `OutputPlugin`

Here not create new a struct/trait to do that, but can be do for rust API clear. I'm free to implement it if anyone think it should do.

### Test Plan

Not need at now.

---


underfin:
* **#408** <a href="https://app.graphite.dev/github/pr/rolldown-rs/rolldown/408?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#407** <a href="https://app.graphite.dev/github/pr/rolldown-rs/rolldown/407?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

milesj (dismissed):

PR #409, opened 2024-01-16
feat(ci): add release workflow

underfin:
### Description

Here port release workflow from https://github.com/rolldown-rs/rolldown_legacy/pull/116. Here remove some platform test, only keep x86_64-apple-darwin/x86_64-pc-windows-msvc/x86_64-unknown-linux-gnu test.

The workflow is some steps at now: 
1. Run `yarn verison` at local,it will create new version for packages,  and push it to one pr
2. Run `Release` workflow action at github `Action` tab, choose your branch. It will trigger release build and test and publish it.

### Test Plan

Here has a windows test crash, i will fix it at next days.

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -275,7 +275,7 @@ jobs:
           find ./packages/ -type d -maxdepth 1 -exec cp LICENSE {} \;
           find ./packages/ -type d -maxdepth 1 -exec cp THIRD-PARTY-LICENSE {} \;
           echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
-          yarn lerna publish from-package --no-verify-access --yes
+          yarn lerna publish from-package --no-private --no-verify-access --yes
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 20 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,171 @@
...
+            target: aarch64-unknown-linux-gnu
+            code-target: linux-arm64-gnu
+
+          - os: macos-latest
+            target: x86_64-apple-darwin
+            code-target: darwin-x64
+
+          - os: macos-latest
+            target: aarch64-apple-darwin
+            code-target: darwin-arm64
```
Boshen: We need to change these to support all the platforms such as `arm64-musl` etc

underfin: Yeah

.github/workflows/release.yml
```diff
@@ -0,0 +1,281 @@
...
+          name: node-artifact
+
+      - name: Publish
+        run: |
+          find ./packages/ -type d -maxdepth 1 -exec cp LICENSE {} \;
+          find ./packages/ -type d -maxdepth 1 -exec cp THIRD-PARTY-LICENSE {} \;
+          echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
+          yarn lerna publish from-package --no-verify-access --yes
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
underfin: Here need to add npm secret to repo setting. I'm not permission with it and npm binding packages, look like @Brooooooklyn should have it, could you add it?

Brooooooklyn: `ROLLDOWN_NPM_TOKEN` added, in case we will have more others NPM_TOKEN

hyf0 (approved):
Feel free to merge this if it's ready.

PR #410, opened 2024-01-16
feat: ast-based manipulation

hyf0:
### Test Plan

---


crates/rolldown_oxc/src/ast_snippet.rs
```diff
@@ -0,0 +1,30 @@
+use oxc::{allocator::Allocator, ast::ast, span::Atom};
+
+use crate::{Dummy, IntoIn};
+
+// `AstBuilder` is more suitable name, but it's already used in oxc.
+pub struct AstSnippet<'me> {
```
underfin, line 6: Maybe here we should use oxc `AstBuilder` instead of it? cc @Boshen. If it can be do, we can do it at next pr.

underfin (approved):

PR #435, opened 2024-03-06
chore: add changelog

underfin:
### Description

Add changelog at bump version. The `lerna version --conventional-commits` can't custom version, so here using `conventional-changelog` to generate changelog.

### Test Plan
Not need.

---


package.json
```diff
@@ -35,7 +35,8 @@
     "TODO(hyf0): #need to investigate following commands": "_",
     "build:ci:release": "run-s build:node:ci build:binding:release",
     "build:node:ci": "yarn workspace @rolldown/node run build",
-    "version": "yarn lerna version --no-private"
+    "version": "yarn lerna version --no-private && node scripts/update-root-package-version.js && yarn run changelog",
```
hyf0, line 38: It's better to move these commands up before `TODO(hyf0): #need to investigate following command`.

CHANGELOG.md
```diff
@@ -0,0 +1,188 @@
+## 0.0.4 (2024-03-04)
```
hyf0, line 1: If the version is related to the root package, it seems meaningless. If `CHANGELOG.md` is for end users, I don't think they could understand the changelog from the content.

PR #438, opened 2024-03-06
refactor: remove MagicString at chunk render

underfin:
### Description

Because `render_normal_module` using `MagicString` to add module path to debugger, it will need to calculate sourcemap for it, It is unnecessary at now so here remove it using `String` instead of it. 

### Test Plan

Not need.

---


hyf0 (changes requested):
`MagicString` is necessary. Don't remove it.

crates/rolldown_sourcemap/src/concat_sourcemap.rs
```diff
@@ -4,18 +4,13 @@ use rolldown_error::BuildError;
 use crate::SourceMap;
 
 pub fn concat_sourcemaps(
-  content_and_sourcemaps: &[(String, Option<SourceMap>)],
+  content_and_sourcemaps: &[(String, Option<SourceMap>, Option<String>)],
 ) -> Result<(String, SourceMap), BuildError> {
   let mut s = String::new();
   let mut map = ParcelSourcemap::new("");
   let mut line_offset = 0;
```
hyf0, line 11: I would not expect we need to calculate the sourcemap manually. This is buggy.

underfin: Here only calculate the lineoffset, other things is done at pacel-sourcemap.

underfin:
> `MagicString` is necessary. Don't remove it.

I add it to root, if you want to keep `MagicString` at some code, please point it.

hyf0:
I believe we are not hurry to support sourcemap, so it's better to do it in a practical way.

I'm not sure the steps of sourcemap part, but this is the gist:

1. rendering normal module will emit `MagicString`
2. rendering chunk will concat `MagicString`  using `Joiner`, then emit string and chunk sourcemap.
3. remapping sourcemap chain with the chunk sourcemap to get the final sourcemap

underfin:
> 1. rendering normal module will emit MagicString
> 2. rendering chunk will concat MagicString using Joiner, then emit string and chunk sourcemap.

If using MagicString, we need to implement its sourcemap.
Other point is the oxc codegen result is `String`, if here using `MagicString` only to add a comment, i think it is unnecessary.
Why do you want to keep `MagicString` at here, can you explain it?

hyf0:
> > 1. rendering normal module will emit MagicString
> > 2. rendering chunk will concat MagicString using Joiner, then emit string and chunk sourcemap.
> 
> If using MagicString, we need to implement its sourcemap. Other point is the oxc codegen result is `String`, if here using `MagicString` only to add a comment, i think it is unnecessary. Why do you want to keep `MagicString` at here, can you explain it?

1. `string_wizard` is already support generating sourcemap
2. adding `comment`s is a great example to show how to manipulate the generated code without considering things related to sourcemap
3. We might need to manipulate the generated code or chunk code for other requirements. Things like generating sourcemap should be hidden under some abstractions, so people wouldn't need to care about it while replacing some pieces in the generated code.
4. `MagicString` is a pre-defined format and organized way to deal with sourcemap, rather than the current hard code way. 
5. remapping with chunk sourcemap is needed anyway when `render_chunk` hook supports sourcemap.

underfin:
> string_wizard is already support generating sourcemap

That's cool. If you support `Joiner` sourcemap, it will be instead of `concat_sourcemaps`, original code at https://github.com/Rich-Harris/magic-string/blob/master/src/Bundle.js#L85.

hyf0:
@underfin, I created an issue to make things clear. #441

PR #469, opened 2024-03-08
chore: update deps

ghost:
### Test Plan

---


ghost:
update examples deps esbuild version

yyx990803:
Thanks for the PR. Turns out the example doesn't actually need `esbuild` as a dep and it has been removed in 664c55c.

Please also note that bumping deps should include lockfile changes, and in the future, this will likely be automated so manual dep-bumping PRs are in general not needed.

PR #471, opened 2024-03-08
chore: use better type

liuseen-l:


MinisculeGirraffe:
never would actually be the correct type for this function. 

It doesn't ever return a value since it always throws in every condition.

packages/node/src/utils/index.ts
```diff
@@ -9,11 +9,11 @@ export function arraify<T>(value: T | T[]): T[] {
   return Array.isArray(value) ? value : [value]
 }
 
-export function unimplemented(info?: string): never {
```
MinisculeGirraffe, line 12: never would actually be the correct type for this function.

It doesn't ever return a value since it always throws in every condition.

yyx990803: This.

PR #473, opened 2024-03-08
fix: correct id in transform hook

sxzz:
### Description

Correct the `id` passed in the `transform` hook; it should be the resolved path, not `source`.

### Test Plan

---

The unit test has been refined for greater precision to ensure that the transform result is included in the final bundle.

P.S This is my first PR in Rust language :P

### DIFF
--- a/crates/rolldown/src/module_loader/normal_module_task.rs
+++ b/crates/rolldown/src/module_loader/normal_module_task.rs
@@ -63,7 +63,9 @@ impl<'task, T: FileSystem + Default + 'static> NormalModuleTask<'task, T> {
 
     // Run plugin transform.
     let source: Arc<str> =
-      transform_source(&self.ctx.plugin_driver, source, &mut sourcemap_chain).await?.into();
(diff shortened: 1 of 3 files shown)


hyf0 (approved):

sxzz:
@hyf0 This is nearly the most critical plugin bug blocking the use of unplugin. Could you please release a patch for it ASAP?

hyf0:
> @hyf0 This is nearly the most critical plugin bug blocking the use of unplugin. Could you please release a patch for it ASAP?

Try the local compiled version as a workaround. We currently don't release any version officially. We haven't decide the final package name, so I tend not to do more release before we decide the final name.(Releasing would make it hard to remove package from npm)

hyf0:
Once we decide the final package name, we will set up an automatic release process for nightly version to keep everthing up to date.

sxzz:
It works on my local, but I want to make my unplugin support it and pass CI on GitHub actions...

Alright, let's wait for the final name.

hyf0:
@sxzz  Nightly release works now. Installing`rolldown@nightly` would be enough. There are something wrong with semver version, so far you could use nightly version all the time.

PR #476, opened 2024-03-08
docs: docs scripts at root and docs guide

patak-cat:
### Description

Add docs related scripts to the root package.json (like we have in Vite, but using `:` to align with rolldown's convention).

Add a simple docs guide in contrib.


package.json
```diff
@@ -74,4 +77,4 @@
   "engines": {
     "node": ">=18.18.0"
   }
-}
\ No newline at end of file
+}
```
patak-dev, line 80:
```suggestion
}```

hyf0 (approved):

package.json
patak-cat, line 80: My editor keeps adding this change. I think there should be an extra line here given that there is one at the end of other files in the project 🤔

PR #478, opened 2024-03-08
fix: enforce non-empty chunks array and add type assertion

zclsx:
## Overview

This PR addresses a TypeScript compile-time error related to the `transformToRollupOutput` function's return type. The TypeScript compiler could not infer that the returned `output` array always meets the expected type `[OutputChunk, ...(OutputChunk | OutputAsset)[]]`, due to its dynamic construction.

## Problem

The TypeScript error was twofold:
1. There was a potential for the `chunks` array to be empty, which would violate the expected type by not providing at least one `OutputChunk`.
2. TypeScript could not statically verify that the constructed array met the specific tuple type expected.

## Solution

To resolve these issues, the following changes were made:
1. **Runtime Check**: Introduced a runtime check that throws an error if the `chunks` array is found to be empty. This ensures that the scenario of violating the expected type by having no `OutputChunk` is avoided.
   
   ```typescript
   if (chunks.length === 0) {
     throw new Error("No output chunks found. At least one OutputChunk is required.");
   }

### DIFF
--- a/packages/node/src/utils/transform-to-rollup-output.ts
+++ b/packages/node/src/utils/transform-to-rollup-output.ts
@@ -69,12 +69,15 @@ function transformToRollupOutputAsset(asset: OutputAsset): RollupOutputAsset {
 export function transformToRollupOutput(output: Outputs): RollupOutput {
   const { chunks, assets } = output
 
+  if (chunks.length === 0) {
+    throw new Error("No output chunks found. At least one OutputChunk is required.");
+  }
+  
   return {
-    // @ts-expect-error here chunks.length > 0
     output: [
       ...chunks.map(transformToRollupOutputChunk),
       ...assets.map(transformToRollupOutputAsset),
-    ],
+    ] as [OutputChunk, ...(OutputChunk | OutputAsset)[]],
   }
 }
 


yyx990803:
I'm not sure if the runtime error is warranted here. We essentially know the the chunks will have a non-zero length, it's just the non-zero length is not expressed in the [types generated by napi-rs in the node bindings](https://github.com/rolldown-rs/rolldown/blob/main/crates/rolldown_binding/index.d.ts).

So in a way, the "right fix" would be adjusting the [binding rust src](https://github.com/rolldown-rs/rolldown/blob/main/crates/rolldown_binding/src/output.rs#L82-L85) so that the generated types also express the non-zero length.

underfin:
Thank

underfin:
Thank you contrcontribute, it already fixed at https://github.com/rolldown-rs/rolldown/pull/503

zclsx:
> I'm not sure if the runtime error is warranted here. We essentially know the the chunks will have a non-zero length, it's just the non-zero length is not expressed in the [types generated by napi-rs in the node bindings](https://github.com/rolldown-rs/rolldown/blob/main/crates/rolldown_binding/index.d.ts).
> 
> So in a way, the "right fix" would be adjusting the [binding rust src](https://github.com/rolldown-rs/rolldown/blob/main/crates/rolldown_binding/src/output.rs#L82-L85) so that the generated types also express the non-zero length.

zclsx:
> I'm not sure if the runtime error is warranted here. We essentially know the the chunks will have a non-zero length, it's just the non-zero length is not expressed in the [types generated by napi-rs in the node bindings](https://github.com/rolldown-rs/rolldown/blob/main/crates/rolldown_binding/index.d.ts).
> 
> So in a way, the "right fix" would be adjusting the [binding rust src](https://github.com/rolldown-rs/rolldown/blob/main/crates/rolldown_binding/src/output.rs#L82-L85) so that the generated types also express the non-zero length.



> I'm not sure if the runtime error is warranted here. We essentially know the the chunks will have a non-zero length, it's just the non-zero length is not expressed in the [types generated by napi-rs in the node bindings](https://github.com/rolldown-rs/rolldown/blob/main/crates/rolldown_binding/index.d.ts).
> 
> So in a way, the "right fix" would be adjusting the [binding rust src](https://github.com/rolldown-rs/rolldown/blob/main/crates/rolldown_binding/src/output.rs#L82-L85) so that the generated types also express the non-zero length.

So happy to communicate with my idol! I was ill yesterday and did not reply in time. Sorry. I see this has been solved in 503, which is great!

zclsx:
> Thank you contrcontribute, it already fixed at #503

It's great and hopefully I can contribute to the community in the future

PR #495, opened 2024-03-09
chore: remove chalk deps

ErKeLost:
### Description

Maybe we can reduce the problem of project dependency by customizing the color sequence and reduce the usage volume (temporarily placed in the scripts directory, which may not be appropriate, if other packages can also be used)

### Test Plan

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
     "@ls-lint/ls-lint": "^2.2.2",
     "@parcel/watcher": "^2.3.0",
     "@taplo/cli": "^0.7.0",
-    "chalk": "^4",
     "change-case": "^4",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -46,7 +46,6 @@
     "@ls-lint/ls-lint": "^2.2.2",
     "@parcel/watcher": "^2.3.0",
     "@taplo/cli": "^0.7.0",
-    "chalk": "^4",
```
hyf0, line 49: Thanks. But we think using `chalk` is alright. It's only for devs not for users.

ErKeLost: > Thanks. But we think using `chalk` is alright. It's only for devs not for users.

Thank you for your reply.

PR #503, opened 2024-03-09
feat: restrict typings and refactor `RenderedModule`

hyf0:
### Description

Types in `@rolldown/node` is kind of over-complex and mix up with rollup's types. I think we need make types reliable to our users and use other ways to keep the compatibility with rollup's types.

Types should make user aware what features are available and what aren't.

---

https://github.com/rolldown-rs/rolldown/blob/f81f87dc4a67f18b8de26f3084d20e003aec3d68/crates/rolldown_common/src/types/rendered_module.rs#L4-L5

Above length is calculated based on utf-8 while rollup emits length based utf-16. Remove these currently, until we find a way to get correct result.

---

### Test Plan

---

### DIFF
--- a/crates/rolldown/src/chunk/mod.rs
+++ b/crates/rolldown/src/chunk/mod.rs
@@ -2,6 +2,7 @@ mod de_conflict;
 pub mod render_chunk;
 mod render_chunk_exports;
 mod render_chunk_imports;
+
 use index_vec::IndexVec;
 use rolldown_common::ChunkId;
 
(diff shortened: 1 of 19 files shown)


hyf0:
* **#503** <a href="https://app.graphite.dev/github/pr/rolldown-rs/rolldown/503?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #508, opened 2024-03-09
feat: preserve class names when renaming declarations

Demivan:
### Description

Preserve class names when fixing naming collisions.

When variable name clashes:
```ts
const a = class {}
```

becomes:
```ts
const a$1 = class a {}
```

When class name clashes:

```ts
class A {}
```

becomes:

```ts
var A$1 = class A {} 
```

### Test Plan

Enabled `rollup@function@class-name-conflict` test.

This is my first actual Rust PR, so there could be issues.

---

### DIFF
--- a/crates/rolldown/src/finalizer/impl_visit_mut_for_finalizer.rs
+++ b/crates/rolldown/src/finalizer/impl_visit_mut_for_finalizer.rs
@@ -330,6 +330,83 @@ impl<'ast, 'me: 'ast> VisitMut<'ast> for Finalizer<'me, 'ast> {
     self.visit_class_body(&mut class.body);
   }
 
+  fn visit_variable_declarator(&mut self, declarator: &mut ast::VariableDeclarator<'ast>) {
+    // If we rename a variable declaration, we should preserve the class name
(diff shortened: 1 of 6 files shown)


hyf0:
Thank for your contribution.

Yes, this is an important behavior and we plan to follow the [esbuild's behavior](https://github.com/evanw/esbuild/issues/2149) and support both function and class. It's not an easy task, since we need to integrate the feature with tree shaking and expose options to users. We don't want simply add this behavior so far.

Demivan:
Makes sense. I tried to match the Rollup implementation instead.

I will create a separate pr with c5e1779796c3ba9875598fe004e86eddc092a73a - adds more info about failure in Rollup tests.

What can be worked on then?

hyf0:
> Makes sense. I tried to match the Rollup implementation instead.
> 
> I will create a separate pr with [c5e1779](https://github.com/rolldown-rs/rolldown/commit/c5e1779796c3ba9875598fe004e86eddc092a73a) - adds more info about failure in Rollup tests.
> 
> What can be worked on then?

Glad you are interested in it. Just create a good first issue: https://github.com/rolldown-rs/rolldown/issues/511

Issue #120, opened 2023-10-29
Fix tests that are executed failed


hyf0:
Those tests are failed to be executed, founded in #119.

- [x] test_tests__esbuild__default__arrow_fn_scope__test_config_json
- [x] https://github.com/rolldown-rs/rolldown/issues/135
- [x] test_tests__esbuild__default__const_with_let__test_config_json
- [x] #136
- [ ] test_tests__esbuild__default__forbid_const_assign_when_bundling__test_config_json
- [x] test_tests__esbuild__default__forbid_const_assign_when_lowering_using__test_config_json
- [ ] test_tests__esbuild__default__import_missing_common_js__test_config_json
- [ ] test_tests__esbuild__default__import_then_catch__test_config_json
- [ ] test_tests__esbuild__default__nested_require_without_call__test_config_json
- [ ] test_tests__esbuild__default__preserve_key_comment__test_config_json
- [ ] test_tests__esbuild__default__require_property_access_common_js__test_config_json
- [ ] test_tests__esbuild__default__require_resolve__test_config_json
- [ ] test_tests__esbuild__default__require_without_call__test_config_json
- [ ] test_tests__esbuild__default__this_inside_function__test_config_json
- [ ] test_tests__esbuild__default__var_relocating_bundle__test_config_json
- [ ] test_tests__esbuild__import_star__import_namespace_undefined_property_side_effect_free_file__test_config_json
- [x] test_tests__esbuild__import_star__import_of_export_star__test_config_json
- [x] test_tests__esbuild__import_star__import_of_export_star_of_import__test_config_json
- [x] test_tests__esbuild__splitting__shared_commonjs_into_es6__test_config_json

## Enhancements

Not bug.

- [ ] test_tests__esbuild__default__dynamic_import_with_template_iife__test_config_json

Demivan:
`test_tests__esbuild__default__preserve_key_comment__test_config_json` is blocked by https://github.com/oxc-project/oxc/issues/1046. Oxc does not support printing comments. But I have fixed the snapshot - it had incorrect syntax: #492

AliceLanniste:
`tests__esbuild__default__var_relocating_bundle__test_config_json` is improper order because of
https://github.com/rolldown/rolldown/blob/06cf72d80e4439e1539ddac1a6ddaceacb13a0c5/crates/rolldown/tests/common/case.rs#L69.
[...]

PR #538, opened 2024-03-10
feat: emit warning for const assign code

kazupon:
### Description

fix `test_tests__esbuild__default__forbid_const_assign_when_bundling__test_config_json` in #120

### Test Plan

---


crates/rolldown/tests/esbuild/default/forbid_const_assign_when_bundling/entry.js
```diff
@@ -3,4 +3,5 @@ x = 2
 function foo() {
 	const y = 1
 	y = 2
-}
\ No newline at end of file
+}
+foo()
```
kazupon, line 21: If the function is not called, it tree-shaked, so I’ve added it.

crates/rolldown/src/ast_scanner/impl_visit.rs
```diff
@@ -45,8 +45,11 @@ impl<'ast> Visit<'ast> for AstScanner<'ast> {
   fn visit_identifier_reference(&mut self, ident: &IdentifierReference) {
     let symbol_id = self.resolve_symbol_from_reference(ident);
     match symbol_id {
-      Some(symbol_id) if self.is_top_level(symbol_id) => {
-        self.add_referenced_symbol(symbol_id);
+      Some(symbol_id) => {
+        self.try_diagnostic_forbid_const_assign(symbol_id);
```
hyf0: Since you just need the `Symbold`, you could call this method in `visit_binding_identifier`. Using `visit_identifier_reference` may cause calling `try_diagnostic_forbid_const_assign` multiple times on the same `SymbolId`.

hyf0:
Thanks. This PR is amazing.

crates/rolldown/tests/esbuild/default/forbid_const_assign_when_bundling/test.config.json
```diff
@@ -6,5 +6,6 @@
         "import": "entry.js"
       }
     ]
-  }
-}
\ No newline at end of file
+  },
+  "expectExecuted": false
```
hyf0: Is this error catchable? If it is, you could do


```
let err;

try { ... } catch (e) { err = e }

assert(typeof err !== 'undefined`)
```

hyf0:
You could run `just smoke` locally to find style errors more conveniently.

crates/rolldown/tests/esbuild/default/forbid_const_assign_when_bundling/test.config.json
kazupon: test_script cannot catch re-assignment errors (`TypeError: Assignment to constant variable.`), so I've changed it to catch them in entry.js.
https://github.com/rolldown/rolldown/pull/538/commits/4e50ce2d03426700aed0e575e425afa08fa418ef

PR #552, opened 2024-03-11
feat: detect side effects in ConditionalExpression

Demivan:
### Description

Add trivial detection for side effects in ConditionalExpression

### Test Plan

Added test

### DIFF
--- a/crates/rolldown/src/ast_scanner/side_effect_detector.rs
+++ b/crates/rolldown/src/ast_scanner/side_effect_detector.rs
@@ -152,7 +152,6 @@ impl<'a> SideEffectDetector<'a> {
         seq_expr.expressions.iter().any(|expr| self.detect_side_effect_of_expr(expr))
       }
       Expression::ConditionalExpression(cond_expr) => {
-        // TODO: If `test` can be statically evaluated, we can skip evaluating `consequent` or `alternate`.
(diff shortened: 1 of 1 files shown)


crates/rolldown/src/ast_scanner/side_effect_detector.rs
```diff
@@ -151,6 +151,12 @@ impl<'a> SideEffectDetector<'a> {
       Expression::SequenceExpression(seq_expr) => {
         seq_expr.expressions.iter().any(|expr| self.detect_side_effect_of_expr(expr))
       }
+      Expression::ConditionalExpression(cond_expr) => {
+        // TODO: If `test` can be statically evaluated, we can skip evaluating `consequent` or `alternate`.
```
hyf0: You could remove these comments. We haven't done this. But in the future, we would have a transform pass that done partial evaluation and `true ? true : bar` would be transformed to `true`.

This way we don't need to add complex analysis code in `SideEffectDetector`.

hyf0 (changes requested):

crates/rolldown/src/ast_scanner/side_effect_detector.rs
Demivan: Done

hyf0 (approved):

hyf0:
Thanks

PR #553, opened 2024-03-11
chore: Prepare gen-esbuild-test.js script for JSDoc type integration

shonya3:
### Description

1. Changed CJS to ESM and changed imports accordingly
2. Added links to .go test files for both test cases and declared the testcases array as 'const'
3. Added tree-sitter-go.d.ts and included to tsconfig.json to prevent TypeScript errors

### Test Plan

---


hyf0 (approved):
Thank you!

PR #555, opened 2024-03-11
feat: add side effect detection for most of the statements

Demivan:
### Description

Implemented side effect detection for:
* BlockStatement
* DoWhileStatement
* WhileStatement
* IfStatement
* ReturnStatement
* LabeledStatement
* TryStatement
* EmptyStatement
* ContinueStatement
* BreakStatement
* SwitchStatement

### Test Plan

Rust tests added.

Actually disabled one failing Rollup test:

```js
if ( false ) {
	var foo;
}

assert.equal( foo, undefined );
```

After implementing side effect detection for `if` test is now failing. This is because `VariableDeclaration` without `init` is considered side effect free right now.
`VariableDeclaration` needs a better detection. It is not side effect free if it is being referenced.


hyf0 (approved):
Thank you!

PR #556, opened 2024-03-11
chore: Start adding JSDoc comments and types to gen-esbuild-test.js

shonya3:
### Description

1. Improve doc comments about testcases constant object. Derive TestCaseName type from it.
2. Create readTestCaseSource, that attempts to read .go source or panics with helpful error message. 
3. Add ReturnType annotation to the getTopLevelBinding function.

### Test Plan

---


hyf0 (approved):
Thank you!

PR #564, opened 2024-03-12
feat: add issues template

ErKeLost:
### Description

I'm not sure if we need to add a restriction on `reproduction link` right now, and if `rolldown playground` can run stably in the future, we can add the option of `reproduction link`.

### Test Plan

---


hyf0:
Thanks. We haven't got the chance to think about this. So I'm gonna on hold this PR for now and will review it in the future.

IWANABETHATGUY (approved):
Thanks

PR #566, opened 2024-03-12
chore: Rename cases array to suites to prevent further confusion

shonya3:
### Description

Renamed constant array from 'cases' to 'suites' and derived the 'TestCaseName' type.

Why? In the main loop of this script, the variable 'testCaseName' represents individual test cases, while "default" or "importstar" represent sets of test cases, which are more appropriately termed as 'suites'. This terminology is consistent with the .go source files, where they are also referred to as 'suites'.
```go
var default_suite = suite{
	name: "default",
}
```

### Test Plan

---


hyf0 (approved):
Thank you!

Issue #543, opened 2024-03-11
Migrate from Yarn / Lerna to pnpm


yyx990803:
Lerna was used for legacy reasons but is actually no longer necessary.

Yarn was also used for legacy reasons - old versions of pnpm had issues in Docker, but should no longer be an issue according to @Brooooooklyn. We should investigate whether moving to pnpm would improve CI time.

hyf0:
It's better do this after I finish merging the `@rolldown/node-binding` and `rolldown` together. I will cue you in this issue @underfin after finishing.

Brooooooklyn:
As a reference, oxc-resolver is already using pnpm and it seems to have no problem: https://github.com/oxc-project/oxc-resolver/blob/main/.github/workflows/release_napi.yml

milesj:
It pnpm's publish/release flow anywhere near as good as lerna's? That's pretty much why everyone still uses lerna.

trivikr:
Most package managers have similar features, and the choice should mostly depend on what most contributors are comfortable with.

Personally, I use a yarn modern and changesets in my projects, and have been happy with.
Since rolldown is aimed at consumption in vite which uses pnpm, would that be the best option?

hyf0:
> It's better do this after I finish merging the `@rolldown/node-binding` and `rolldown` together. I will cue you in this issue @underfin after finishing.

@underfin Done in #545. I update the napi-rs cli to `^3.0.0-alpha.43`. See documents at https://github.com/napi-rs/napi-rs/tree/main/cli/docs.

Commands related to releasing should be no problems, but you might want to checkout out them again.

jonkoops:
Let me know if some assistance is required with this, I've ported a couple of projects from Yarn to PNPM, I'd be willing to take this on if that helps.

> It pnpm's publish/release flow anywhere near as good as lerna's? That's pretty much why everyone still uses lerna.

Should be pretty decent, are there any specific Lerna features you need that are not in PNPM?

milesj:
I'm looking into this.

hyf0:
The migration is finished in #567. But there are some other problems left:

- [x] Make release flow work with using pnpm
- [ ] Install `tree-sitter-go` with pnpm cause errors in my m2pro mac

ghiscoding:
[...]

PR #567, opened 2024-03-12
feat: migrate from yarn to pnpm

milesj:
### Description

Part of #543

This PR does a bunch of things:

- Migrates from yarn to pnpm. I ran some small benchmarks and pnpm was about 3x faster on cold installs, but the same on warm installs.

|      | yarn 4.1.1 | pnpm 8.15.4 |
| ---- | ---------- | ----------- |
| Cold | 14.42s     | 4.5s       |
| Warm | 1.45s      | 1.3s        |

- Audited dependencies of every `package.json`, and removed deps that weren't used, or should be in the root.
- 
- Tested all scripts, and fixed any issues I encountered.

### Test Plan

---


package.json
```diff
@@ -3,54 +3,47 @@
...
   },
-  "workspaces": [
-    "crates/rolldown_binding_wasm",
-    "packages/**",
-    "web/**",
-    "examples/**",
-    "scripts"
-  ],
   "scripts": {
+    "nuke": "rm -rf node_modules && rm -rf packages/*/node_modules && rm -rf web/*/node_modules && rm -rf examples/*/node_modules && rm -rf crates/rolldown_binding_wasm/node_modules && rm -rf scripts/node_modules",
```
milesj: For testing...

packages/bench/index.js
```diff
@@ -105,6 +106,15 @@ async function runRollup(item) {
 for (const suite of suites) {
   const bench = new Bench({ time: 100, iterations: suite.benchIteration ?? 10 })
 
+  // Check if inputs have been initialized
+  for (const input of suite.inputs) {
```
milesj, line 110: When testing, this would crash hard if `just setup-bench` wasn't ran, so added this little check.

packages/rolldown/package.json
```diff
@@ -63,11 +63,10 @@
   },
   "devDependencies": {
     "@napi-rs/cli": "^3.0.0-alpha.43",
-    "@types/node": "^20.11.25",
-    "npm-run-all2": "^6.1.2",
-    "prettier": "^3.2.5",
+    "colorette": "^2.0.20",
```
milesj, line 66: There are used in tests, but weren't listed. Because pnpm scopes them, they're now required.

packages/rolldown/test/cases/external/array/config.ts
```diff
@@ -1,4 +1,4 @@
-import type { RollupOptions, RollupOutput } from 'rolldown'
+import type { RollupOptions } from '../../../../src'
```
milesj, line 1: Because of pnpm, rolldown was no longer able to reference itself, so had to use relative paths.

.github/workflows/ci.yml
```diff
@@ -99,19 +99,24 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 20
-          cache: yarn
+
+      - name: Install pnpm
+        run: corepack enable
+
+      - name: Install pnpm
+        run: corepack enable
```
Demivan: This is duplicated.
`actions/cache` needed to cache pnpm packages, so they are not downloaded every time.

milesj:
Dependencies install in less than 45 seconds, typically around 15-20. Not really worth adding caching at this point.

Demivan:
> Dependencies install in less than 45 seconds, typically around 15-20. Not really worth adding caching at this point.

Why not make it even faster? It is just 4–5 lines of CI code.

hyf0:
@milesj Do you have time to clean up this PR? So we could merge the migration first. We could solve release problems in another PR.

---

I saw the c-spell job failed. You could add the issue file to `cspell.json#ignorePaths` to ignore it.

hyf0:
Let me take it from here. I will comment out the release ci for temp.

hyf0 (approved):
Great work!

hyf0 (approved):
Great work!

milesj:
Downloading and unpacking an archive over the network will still take similar times. Caching isn't really necessary.

PR #568, opened 2024-03-13
chore: improve paths configuration and update "how to use" section

shonya3:
### Description

1. Added the resolveTestCaseDir function to simplify the main loop code
2. Created the TESTS_ROOT_DIR constant and placed it at the top of the file for easy configuration from one place.
3. Updated the 'How to use this script' section.

Note: I want to point out, that first version of script had. But was removed later.
```js
// TODO Add a `.` prefix instead of skipping rest of the code.
```
image
https://github.com/rolldown/rolldown/pull/105/commits/4a4a3f149cc3caf7a436ebe3a6b9b28019044dba

That script, as i understand, never generated prefixed with `.` directories for skipped tests.
But there are prefixed with dot skipped tests in  /crates/rolldown/tests/esbuild/
Maybe, it  was generated some other way.
Just want to point, maybe we still need to generate dot-prefixed directories for skipped tests. Maybe the deletion of the TODO comment was accidental.

### Test Plan

---

### DIFF
--- a/scripts/gen-esbuild-test.js
+++ b/scripts/gen-esbuild-test.js
@@ -9,12 +9,12 @@ import chalk from 'chalk'
 import * as dedent from 'dedent'
 
 // How to use this script
-// 1. Set the root directory (relative to scripts dir)
+// 1. Set the tests root directory
 
-/** Relative to scripts dir. Expected values: '../crates/rolldown' or './', or your variant.  */
-const TESTS_ROOT_DIR = './'
+const TESTS_ROOT_DIR = import.meta.dirname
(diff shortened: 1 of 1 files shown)


hyf0 (changes requested):
Thank for your contribution. But I think the current implementation is more understandable than new one. This is an easy task and hard code is better than abstraction in this situation.

shonya3:
Got it. It is about the function as I understand. Is the "How to use this script" part fine?

shonya3:
Now, this is good and clear, i like this one personally!

hyf0 (approved):
Good!

PR #570, opened 2024-03-13
chore: optimize sourcemap 

ErKeLost:
### Description

1. Optimize concat_sourcemaps to calculate the offset in advance at the end of the judgment if you prevent repeated calculation in the loop

2. Fix the wrong test sourcemap type

### Test Plan

---


ErKeLost:
hi @underfin  I see that the sourcemap has started refactoring, but I see that the line offset calculation is still within the loop. Do you think moving the offset calculation out of the loop can improve the code performance?

underfin:
> Do you think moving the offset calculation out of the loop can improve the code performance?

@ErKeLost I think it does not affect performance, but that's ok. Because I'm refactoring sourcemap implement, so the pr will be closed to avoid conflict. Thank you.

ErKeLost:
Okay, thank you for your reply.

PR #574, opened 2024-03-13
refactor(rust): limit usages of oxc `Atom`

hyf0:
### Test Plan

---

### DIFF
--- a/crates/rolldown/src/finalizer/impl_visit_mut_for_finalizer.rs
+++ b/crates/rolldown/src/finalizer/impl_visit_mut_for_finalizer.rs
@@ -64,7 +64,7 @@ impl<'ast, 'me: 'ast> VisitMut<'ast> for Finalizer<'me, 'ast> {
             if matches!(importee_linking_info.wrap_kind, WrapKind::Esm) {
               let wrapper_ref_name =
                 self.canonical_name_for(importee_linking_info.wrapper_ref.unwrap());
-              program.body.push(self.snippet.call_expr_stmt(wrapper_ref_name.to_oxc_atom()));
+              program.body.push(self.snippet.call_expr_stmt(wrapper_ref_name));
             }
 
             match importee.exports_kind {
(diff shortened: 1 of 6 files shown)


hyf0:
* **#574** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/574?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
@hyf0 Would it be helpful if we added a method in Oxc `Atom<'a>::new_in` which takes a `&str` + an `oxc_allocator::Allocator`?

If you're planning on doing more refactoring, then I think it could be a good idea. Once the inline optimization is implemented in `Atom`, for short strings `Atom::new` / `Atom::new_in` will copy the source string directly into its inline bytes, whereas building an `oxc_allocator::String` first will result in 2 copies - copy into the arena, then copy again into inline representation.

By the way `Atom::Arena(s)` no longer exists in latest Oxc, replaced by `Atom::new`. Sorry for all the churn! We think we've finally got there now with the API...

hyf0:
This PR is only for integrating with `Atom` must have bound to a lifetime. So where to construct it become important. We need limit the constructing to a certain scope, so it won't affect other code.

> If you're planning on doing more refactoring

Yeah. Prabably no big one. But change `Atom::Arena(s)` to `Atom::new` will definitely do.

> , for short strings Atom::new / Atom::new_in will copy the source string directly into its inline bytes

Yeah. This would be helpful.

> By the way Atom::Arena(s) no longer exists in latest Oxc,

Got it. That's what this PR for. This PR makes that updating to latest oxc only need to change few code, otherwise we need do what this PR do in the updating PR too.

overlookmotel:
Thanks for coming back. OK cool, seems like no hurry then. We can leave it until inline optimization is implemented. I just didn't want you to have to make loads of changes now, and then make them all over again later. But sounds like that's not going to happen.

PR #576, opened 2024-03-13
feat: use pnpm changeset replace yarn lerna

joinmouse:
use pnpm changeset replace yarn lerna version flow work


hyf0:
Thank for your contribution. Bumping version is a key process during release, so it's probably better to do it with setting up releasing together. we did this in https://github.com/rolldown/rolldown/pull/577.

PR #582, opened 2024-03-15
fix: rolldown example root is empty in debug mode

zhazhazhu:
The root variable of the file path crates/rolldown/examples/basic.rs is empty in debug mode


hyf0 (approved):
Thanks!

zhazhazhu:
> Thanks!

😀

PR #583, opened 2024-03-15
feat(package): yarn lerna version => pnpm changeset version

joinmouse:


web/playground/index.html
hyf0, line 1: I don't think this change is related to this PR. Could you keep it original?

packages/rolldown/src/options/input-options.ts
hyf0, line 1: I don't think this change is related to this PR. Could you keep it original?

hyf0 (changes requested):

package.json
```diff
@@ -29,7 +29,7 @@
     "docs": "pnpm --filter rolldown-docs run dev",
     "docs:build": "pnpm --filter rolldown-docs run build",
     "docs:preview": "pnpm --filter rolldown-docs run preview",
-    "version": "yarn lerna version --force-publish --no-private --no-git-tag-version && node scripts/post-version.cjs",
+    "version": "pnpm changeset version",
```
milesj: we should maybe remove this since we have it in the `justfile`.

packages/rolldown/src/options/input-options.ts
joinmouse: I pnpm run prettier , fmt this file, now rolled back this file.

package.json
joinmouse: @hyf0 What do you think？

hyf0: Yeah. I agree with @milesj. He knows this better.

joinmouse: Ok, Now I remove this script about version.

hyf0:
Run `just fmt` to fix format errors.

joinmouse:
> Run `just fmt` to fix format errors.

Now, fix fmt.

hyf0 (approved):
Thank you!

PR #584, opened 2024-03-15
ci: move node-validation to seprate file

Dunqing:
### Description

`pnpm format:ci` will format Markdown files. but the ci action will ignore markdown changes. So I think we should do a lint check in the new action.
### Test Plan

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,36 +116,6 @@ jobs:
       - name: Run Test
         run: pnpm test
 
-  node-validation:
-    name: Node Validation
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-
-      - name: Install pnpm
-        uses: pnpm/action-setup@v3
-
-      - name: Install node
-        uses: actions/setup-node@v4
-        with:
-          node-version: 20
-          cache: pnpm
-
-      - name: Install dependencies
-        run: pnpm install
-
-      - name: Oxc Lint
-        run: pnpm lint-code
-
-      - name: Lint Filename
-        run: pnpm lint-filename
-
-      - name: Format Check
-        run: pnpm format:ci
-
-      - name: Lint Spell
-        run: pnpm lint-spell
-
   # changesets:
   #   name: Changesets Status
   #   runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


Dunqing:
* **#584** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/584?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @Dunqing and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

hyf0:
I get you. But the better solution would be removing `paths-ignore:` in `CI.yaml`. Due to the "Status check" of github is static and running these actions are required. We need to bare with useless action running for some PRs.

hyf0:
> I get you. But the better solution would be removing `paths-ignore:` in `CI.yaml`. Due to the "Status check" of github is static and running these actions are required. We need to bare with useless action running for some PRs.

Or should we say the `path ignore `is not smart enough. It should treat skipped actions as completed successfully rather than skipped.

Dunqing:
> I get you. But the better solution would be removing `paths-ignore:` in `CI.yaml`. Due to the "Status check" of github is static and running these actions are required. We need to bare with useless action running for some PRs.

CI action has many jobs, As far as I know only `node-validation` needs to remove path_ignores. but this causes all other jobs to be executed, so I moved it to a new action to avoid useless CI check

hyf0:
> > I get you. But the better solution would be removing `paths-ignore:` in `CI.yaml`. Due to the "Status check" of github is static and running these actions are required. We need to bare with useless action running for some PRs.
> 
> CI action has many jobs, As far as I know only `node-validation` needs to remove path_ignores. but this causes all other jobs to be executed, so I moved it to a new action to avoid useless CI check

As I was saying, it's useless and we have to bare with it. You will notice that currently your PR doesn't not run `cargo test` action and it's required to merge this PR. The problem is that github can't set what actions are required dynamically, so we need to set needed actions required. That's where we hit the "path ignore" problem.

Skipped actions by "path ignore" won't pass the checks of github.

hyf0 (approved):
Let's merge this one first. I will refactor them using https://github.com/dorny/paths-filter later.

PR #598, opened 2024-03-17
fix: Windows run the test case error

liulinboyi:
In order for more people to contribute code on Windows, the error of test cases on Windows has been fixed. You can now pass all test cases on Windows.

## pass all test cases on Windows
https://github.com/rolldown/rolldown/assets/41336612/850b9f7a-921b-4ebb-b1d7-7b5714c70289

### snapshot error
image

image

image

### Bundler compile cmd path error
image

### Windows exec node error
image

in windows the compiler entry will be `\\\\?\\D:\\rolldown\\crates\\rolldown\\tests\\fixtures\\resolve\\browser-filed-false\\dist\\package.mjs`, the `node` can't read the path currently.

#### Error
image
image
#### Success
image

image


Dunqing:
We also need to run ci in Windows

liulinboyi:
> We also need to run ci in Windows

Okay the workflows add new os: Windows done.

hyf0:
Well done!

hyf0 (approved):

PR #607, opened 2024-03-18
feat: generate sourcemap

underfin:
### Description

The pr intend to generate sourcemap.Here add more `Span` position to make sure sourcempa generate correctly.

### Test Plan

Added simple test, i will add more test at next.


hyf0 (approved):
Awesome.

PR #610, opened 2024-03-19
feat: rolldown cli first implementation

### LINKED ISSUE #599: Implement CLI in `rolldown`
For reference: https://github.com/rolldown/rolldown/discussions/592.

Let's have a CLI for start and keep improving it.

Some requirements:

- The CLI should be written in JS with JSDoc.

We want to avoid extra build steps and JS with JSDoc is perfect for such a situation.

kazupon:
### Description

related #599 

This PR is support rolldown cli.
The first rolldown cli supports the `--config` option.
The following is usage:

```
rolldown --config rolldown.config.js
```

- The spec of config file is almost the same as rolloup: https://rollupjs.org/command-line-interface/#configuration-files
- Support config formats are `.js`, `.mjs`, and `.ts` (support esm only).
- TS format, rollup cli require `--configPlugin`. rolldown cli does not use plugin, rollodown cli resolve it.

### Test Plan

---


yarn.lock
hyf0, line 1: We have switched to `pnpm` from `yarn`

milesj:
This should also be written in TypeScript.

hyf0:
> This should also be written in TypeScript.

This is expected. We plan to move this to `rolldown` package instead of a new standalone package. Thus, we decide to use js + jsdoc to get type safety and get rid of build steps for cli. This way we won't be needed to modify the build steps.

hyf0:
> Support config formats are .js, .mjs, and .ts (support esm only).

@kazupon  You could delay the support for `.ts` files. It's kind of tricky and have some traps. let's have a fully research before we doing it.

packages/cli/lib/main.js
```diff
@@ -0,0 +1,37 @@
...
+import path from 'node:path'
+import { URL } from 'node:url'
+import { getPackageJSON } from './utils.js'
+
+const __dirname = path.dirname(new URL(import.meta.url).pathname)
+const { version, description } = getPackageJSON(path.resolve(__dirname, '..'))
+
+/**
+ * NOTE:
+ *  currenctly, It's hard to customize usage with citty `renderUsage`.
```
pi0: Please let me know or feel free to open an issue for what you like to be customizable or better rendered in citty usage renderer 👍🏼 :)

packages/cli/package.json
hyf0, line 1: We don't want to introduce a new package just for a CLI. Please move code to the `rolldown` folder.

packages/cli/bin/cli.js
hyf0, line 1: Please move this file to `/packages/rolldown/bin/cli.js`.

packages/cli/lib/main.js
hyf0, line 1: Please move files of `packages/cli/lib` to `/packages/rolldown/cli`.

packages/cli/test/config.test.js
```diff
@@ -0,0 +1,52 @@
+import assert from 'node:assert'
+import { describe, test } from 'node:test'
```
hyf0: I remember `node:test` has some problems on croess-platform. You could use `vitest` under the `packages/rolldown`.

hyf0 (changes requested):

packages/cli/lib/main.js
```diff
@@ -0,0 +1,37 @@
+import { defineCommand, runMain, showUsage } from 'citty'
+import { colors } from 'consola/utils'
+import path from 'node:path'
+import { URL } from 'node:url'
+import { getPackageJSON } from './utils.js'
+
+const __dirname = path.dirname(new URL(import.meta.url).pathname)
+const { version, description } = getPackageJSON(path.resolve(__dirname, '..'))
```
hyf0: You should be able to import `.json` file directly while using node.

Just write `import { version, description } from '../package.json'` and it should works.

kazupon:
> @kazupon You could delay the support for `.ts` files. It's kind of tricky and have some traps. let's have a fully research before we doing it.

I use jiti for TS config loading.
https://github.com/unjs/jiti
jiti can also resolve typescript. In fact, it is also used as a config loader for c12 in unjs.
https://github.com/unjs/c12

another show case, jiti also used in tailwind config loading.
https://github.com/tailwindlabs/tailwindcss/blob/3.4/src/lib/load-config.ts

So I understand that I may use it to resolve TS configs.

@pi0 
If you have any comments to add, please feel free to comment. :)

packages/cli/lib/main.js
kazupon: ~I have noticed that when some js modules from cli are moved to `packages/rolldown`, it is hard to distinguish them from the binding src if they are not organized.~

~I think it should be moved to `packages/rolldown/src/cli`.~

kazupon: After thinking about it, I realized that I could also organize the cli code in `packages/rolldown/cli`.
Don't mind my above comment. 😅

packages/cli/lib/main.js
kazupon: I am implementing cli based on es module. So the rolldown package must have `"type": "module"`.

And in that case, node throw the following error when importing json:
```sh
TypeError [ERR_IMPORT_ASSERTION_TYPE_MISSING]: Module "file:///path/to/rolldown/packages/rolldown/package.json" needs an import attribute of type "json
```
We must use import assertion as follow:

```js
import pkgJson from '../package.json' assert { type: 'json' }
```

hyf0:
Something wrong with the filename conventions.

You could add `**/*.config.{js,mjs,cjs,ts,cts,mts}` to `.ls-lint.yml`.

kazupon:
Hmm, 🤔   I've done basic implementation of build with config, and rolldown `write` seems to work, but the file is not generated. In my environment...

`examples/basic-vue` is building, but I don't know what's causing it...

You can try that:
```sh
CONSOLA_LEVEL=4 node ./bin/cli.js -c test/fixtures/basic/rolldown.config.js
```

hyf0:
I think this PR is good to go. Thank your wonderful work! @kazupon 

Let me takes this PR from here. With some thoughts come to my head, I will do some refactor.

- We want to `rolldown` to have almost zero dependencies.
  - If we apply this rule on CLI, it means either we implement the CLI from the ground or we introduce build steps to bundle these dependencies. I prefer the bundle solution, because it's simple to do.
- Since build steps can't be avoided, using ts to implement CLI is fine now.(Sorry to @kazupon 
- I'm gonna remove support for `.ts`. As I said before, supporting `.ts` config file could introduce a lot of issues, such as https://github.com/vitejs/vite/issues/5370. Let's have a fully research on this to get the best solution for using `.ts` config file. In the meanwhile the command `some-ts-transpiler ./rolldown.config.ts --out ./rolldown.config.js && rolldown --config ./rolldown.config.js` could be a workaround.

pi0:
@hyf0 Regarding the `.ts` extension, I totally respect your judgment if you prefer to delay or not support `.ts` extension. But I guess it won't have edge cases like https://github.com/vitejs/vite/issues/5370. jiti even allows mixed syntaxes and is heavily tested over years -- we might even replace it's transpile step with rolldown natives! (probably the only limitation is top-level await support that will be added in the future versions of it)

hyf0:
> @hyf0 Regarding the `.ts` extension, I totally respect your judgment if you prefer to delay or not support `.ts` extension. But I guess it won't have edge cases like [vitejs/vite#5370](https://github.com/vitejs/vite/issues/5370). jiti even allows mixed syntaxes and is heavily tested over years -- we might even replace it's transpile step with rolldown natives! (probably the only limitation is top-level await support that will be added in the future versions of it)

Glod to know that. Will look into `jiti` for sure.

kazupon:
@hyf0 
> Let me takes this PR from here. With some thoughts come to my head, I will do some refactor.

okay! 👌 
If you need any help with the CLI, let me know. I am always happy to help rolldown. :)

kazupon:
> We want to `rolldown` to have almost zero dependencies.
> * If we apply this rule on CLI, it means either we implement the CLI from the ground or we introduce build steps to bundle these dependencies. I prefer the bundle solution, because it's simple to do.

if you will use cli library, you can bundle with unbuild.
In the current version of unbuild, deps under node_modules cannot be bundled without a little hook adjustment.
Fortunately, @pi0  is also the author of unbuild. He may be able to provide some support. :)

hyf0 (approved):
Again. Thank @kazupon for his great work.

PR #629, opened 2024-03-21
chore: isSupportedFormat ts ext name

ErKeLost:
### Test Plan

---


packages/rolldown/src/cli/utils.ts
```diff
@@ -23,5 +23,5 @@ export async function loadConfig(
  */
 function isSupportedFormat(configPath: string): boolean {
   const ext = path.extname(configPath)
-  return /\.(js|mjs)$/.test(ext)
+  return /\.(js|mjs|ts)$/.test(ext)
```
kazupon: As commented [here](https://github.com/rolldown/rolldown/pull/610#issuecomment-2009485012), we are currently not supporting the ts extension.

ErKeLost: Thank you for your reply. I see that error's prompt supports ts format, so do you think other logic optimizations can be merged? I can remove the rules of this ts type.

kazupon (approved):
Your PR looks a good refactoring for me. :)

hyf0 (approved):
Thanks!

PR #634, opened 2024-03-21
feat: add banner output option

PengBoUESTC:
### Description
add `banner` option for config
**WIP**

### Test Plan

---


crates/rolldown/src/chunk/mod.rs
```diff
@@ -155,7 +155,10 @@ impl Chunk {
           Ok(())
         },
       )?;
-
+    // add banner
+    if let Some(banner) = &output_options.banner {
+      concat_source.prepend_source(Box::new(RawSource::new(banner.to_string())))
+    }
```
underfin, line 163: Here `prepend_source` is unnecessary, you can move it to the first of chunk `concat_source` and using `addSource`.

PengBoUESTC: I'm not good at `rust` at all, so maybe U should review carefully +_+

underfin (approved):

underfin:
Thank you.

crates/rolldown_binding/src/options/binding_output_options/mod.rs
```diff
@@ -15,7 +18,10 @@ pub struct BindingOutputOptions {
 
   // amd: NormalizedAmdOptions;
   // assetFileNames: string | ((chunkInfo: PreRenderedAsset) => string);
-  // banner: () => string | Promise<string>;
+  #[derivative(Debug = "ignore")]
+  #[serde(skip_deserializing)]
+  #[napi(ts_type = "undefined | string | ((chunk: RenderedChunk) => string | Promise<String>)")]
```
hyf0, line 23: This should be
```
#[napi(ts_type = "string | ((chunk: RenderedChunk) => MaybePromise<VoidNullable<string>>")]
```

PengBoUESTC:
> Thank you.

Thank U :);

crates/rolldown_binding/src/options/binding_output_options/mod.rs
PengBoUESTC: Thank U :);

PR #639, opened 2024-03-22
feat: add react example

js-0127:
### Add React Template

I've added a React example for the rolldown feature and updated the lock file.


IWANABETHATGUY:
Closed in favor of stale, we already support transform `jsx, tsx` by default, please reopen another pr.

PR #640, opened 2024-03-22
fix(rust): should not hard code `output.dir`

PengBoUESTC:
### Description

out `dir` is hard code in `normalize_options`

### Test Plan

---


hyf0 (approved):
Good catch!

PR #642, opened 2024-03-22
fix: rolldown test config priority

PengBoUESTC:
### Description

`config` will overwrite the `input` & `output` config in test runner

### Test Plan

---


hyf0 (approved):
Thank you!

PengBoUESTC:
> Thank you!

+_+

PR #644, opened 2024-03-23
chore: add repository.directory in package.json

trivikr:
### Description

Adds repository.directory in package.json which specifies the monorepo workspace directory in which source code lives.

This will allow users browsing a specific `@rolldown/*` package on npmjs to directly jump to it's source code.
Docs: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#repository

### Test Plan

This will be verified on npm publish.

---


hyf0 (approved):
Thank you!

PR #658, opened 2024-03-25
fix(cli): normalize provided config path to satisfy esm-loader in windows 

woai3c:
### Description
resolved issue with ESM loader on Windows by converting file paths to file:// URLs

#### before
image

#### now
image

### Test Plan

---

### DIFF
--- a/packages/rolldown/src/cli/utils.ts
+++ b/packages/rolldown/src/cli/utils.ts
@@ -1,4 +1,5 @@
 import path from 'node:path'
+import { pathToFileURL } from 'url'
 import { ERR_UNSUPPORTED_CONFIG_FORMAT } from './errors.js'
 import { RolldownConfigExport } from '../types/rolldown-config-export.js'
 
@@ -15,7 +16,9 @@ export async function loadConfig(
   if (!isSupportedFormat(configPath)) {
     throw new Error(ERR_UNSUPPORTED_CONFIG_FORMAT)
   }
-  return import(configPath).then((config) => config.default)
+  return import(pathToFileURL(configPath).toString()).then(
+    (config) => config.default,
+  )
 }
 
 /**


hyf0:
Good catch. Could you add a test for this case?

woai3c:
> Good catch. Could you add a test for this case?

This seems difficult to write tests for because it is not like a function or a feature where you can control the inputs and outputs; changing the input of the function would change the output. The error is that it can't recognize the path, you need to use nodejs's pathToFileURL() to convert it into a path that is compatible and recognizable across different operating systems.

woai3c:
> Good catch. Could you add a test for this case?

This seems difficult to write tests for because it is not like a function or a feature where you can control the inputs and outputs; changing the input of the function would change the output. The error is that it can't recognize the path, you need to use nodejs's pathToFileURL() to convert it into a path that is compatible and recognizable across different operating systems.

I tried writing some test cases to see what the test cases look like before the code is modified. It turns out that running the tests works fine, but directly executing the rolldown --config ./rolldown.config.js build task results in an error. Since I don't know much about vitest, I'm not sure if there is special handling inside, or if it is caused by other circumstances.

image

image

woai3c:
> Good catch. Could you add a test for this case?

`pathToFileURL()` This function converts a file path into a file:// URL, which is valid on all platforms. On Linux and macOS, the file path is already a valid file:// URL, so the pathToFileURL function doesn't actually change the path. However, on Windows, because the path format is different from the file:// URL, the pathToFileURL function converts the path into a valid file:// URL.
```js
import { pathToFileURL } from 'url'
import path from 'path'

// 在 Linux 和 macOS 上，这将输出 'file:///path/to/file'
// 在 Windows 上，这将输出 'file:///C:/path/to/file'
console.log(pathToFileURL(path.resolve('/path/to/file')))
```

hyf0 (approved):
Thank you!

PR #664, opened 2024-03-26
feat: add filter for rolldown runner script

PengBoUESTC:
### Description
add `TEST_FILTER` for runner script, so that we can use test like that
`TEST_FILTER=input pnpm test`;  *just run for input cases*
`TEST_FILTER=input/string pnpm test`; *just run for input/string case*
`TEST_FILTER=string pnpm test`; *just run for string cases*

**But I'm not sure does this worked for windows system**

### Test Plan

---

### DIFF
--- a/packages/rolldown/test/runner.ts
+++ b/packages/rolldown/test/runner.ts
@@ -4,18 +4,21 @@ import { InputOptions, OutputOptions, RollupOptions, rolldown } from '../src'
 import path from 'node:path'
 import fs from 'node:fs'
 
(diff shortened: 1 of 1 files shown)


hyf0:
I guess https://rolldown.rs/contrib-guide/test#run-the-specific-test will meet your need.

PengBoUESTC:
> I guess https://rolldown.rs/contrib-guide/test#run-the-specific-test will meet your need.

My bad :( .

PR #684, opened 2024-03-28
chore: improve rolldown test

Dunqing:
### Description

Closer to the implementation of vitest

### Test Plan

---

### DIFF
--- a/packages/rolldown/tests/fixture.test.ts
+++ b/packages/rolldown/tests/fixture.test.ts
@@ -25,8 +25,6 @@ function main() {
         return
       }
 
-      // FIXME: This empty log is here to make vitest shows stdout/stderr content made from rust. Wonder why.
-      console.log()
       try {
         const output = await compileFixture(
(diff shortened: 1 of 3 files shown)


packages/rolldown/tests/fixture.test.ts
```diff
@@ -18,15 +18,11 @@ function main() {
...
       nodePath.dirname(testConfigPath),
     )
-    test(dirPath, async (ctx) => {
-      const testConfig = await loadTestConfig(testConfigPath)
-      if (testConfig.skip) {
-        ctx.skip()
-        return
-      }
+
+    const testConfig = await loadTestConfig(testConfigPath)
```
hyf0: I didn't do this because this will slow down spawning tests.

hyf0: And using `Promise.all()` will causes the tests created unorderly,

hyf0 (approved):
Well done!

PR #685, opened 2024-03-28
chore(test): pre-import test config files by`import.meta.glob`

Dunqing:
### Description

Get all configs at compile time. It won't slow down spawning tests.

One downside is that we can't get the type of import.meta.glob from `vitest`

### Test Plan

---

### DIFF
--- a/packages/rolldown/tests/fixture.test.ts
+++ b/packages/rolldown/tests/fixture.test.ts
@@ -26,7 +26,7 @@ function main() {
           testConfig.afterTest(output)
         }
       } catch (err) {
-        throw new Error(`Failed in ${testConfigPath}`)
+        throw new Error(`Failed in ${testConfigPath}`, { cause: err })
       }
     })
   }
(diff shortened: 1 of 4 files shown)


packages/rolldown/tests/fixture.test.ts
```diff
@@ -1,33 +1,25 @@
+/// <reference types="vite/client" />
```
hyf0: Feel this not ideal. Maybe vitest could exposed these from https://github.com/vitest-dev/vitest/blob/main/packages/vitest/importMeta.d.ts.

hyf0 (approved):

hyf0:
Adding `vite` in devDependencies should be fine.

packages/rolldown/tests/fixture.test.ts
Dunqing: Yes, I will report this issue to `vitest`

Dunqing:
> Adding `vite` in devDependencies should be fine.

OK, we can't avoid installing `vite` unless `vitest` reexported `vite/client`

PR #678, opened 2024-03-27
chore: Debug with Javascript and Rust

moushicheng:
# Description
I have updated the documentation and tutorials，

this is documentation content:⬇️
## Debug with Javascript and Rust
1. Open Javascript Debug Termial
2. Execute any program that works with javascript and rust in Javascript Debug Termial
3. Start Attach:Rust  in Tab `Run and Debug`
4. select javascript process.
5. enjoy mixed debug!

I think you'll still be confused after you read this steps.so you can see this video to learn about this process

[bilibili.com: Debug with Javascript and Rust](https://www.bilibili.com/video/BV1Rm421n79f/)

PR #686, opened 2024-03-28
chore: Optimize launch.json configuration to enable simultaneous debu…

woai3c:
### Description
Optimize launch.json configuration to enable simultaneous debugging of JS and Rust.

demo

### Test Plan

---

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,21 +1,23 @@
 {
-  // 使用 IntelliSense 了解相关属性。
-  // 悬停以查看现有属性的描述。
-  // 欲了解更多信息，请访问: https://go.microsoft.com/fwlink/?linkid=830387
-  "version": "0.2.0",
   "configurations": [
     {
       "name": "debug-js-basic",
-      "program": "./node_modules/rolldown/bin/cli.js",
-      "args": ["--config", "./rolldown.config.js"],
-      "request": "launch",
+      "port": 9229,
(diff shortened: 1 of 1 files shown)


hyf0:
Nice done. However I'm not gonna merge this PR due to the same reason I said in https://github.com/rolldown/rolldown/pull/678. This is a longer version. Debugging both rust and js in the same time is very useful, you could share you experiences and config in https://github.com/rolldown/rolldown/discussions/categories/contribution.

Currently the `launch.json` of main branch isn't kind of useful without good-written documents to cooperate. And it's not well maintained and people has their own needs for `launch.json`.

I'm gonna remove it from the main branch and reject updates for it. `launch.json` will be added to `.gitignore`, so people could have their local `launch.json` for their own usages.

woai3c:
> Nice done. However I'm not gonna merge this PR due to the same reason I said in #678. This is a longer version. Debugging both rust and js in the same time is very useful, you could share you experiences and config in https://github.com/rolldown/rolldown/discussions/categories/contribution.
> 
> Currently the `launch.json` of main branch isn't kind of useful without good-written documents to cooperate. And it's not well maintained and people has their own needs for `launch.json`.
> 
> I'm gonna remove it from the main branch and reject updates for it. `launch.json` will be added to `.gitignore`, so people could have their local `launch.json` for their own usages.

Ok,Got it!

PR #689, opened 2024-03-28
fix: correct the TS type of the banner ts-binding

### LINKED ISSUE #682: Correct type of `banner`
This should be

```rust
#[napi(ts_type = "string | ((chunk: RenderedChunk) => MaybePromise<VoidNullable<string>>")]
```

_Originally posted by @hyf0 in https://github.com/rolldown/rolldown/pull/634#discussion_r1542477756_

---

For field `pub banner: Option<JsAsyncCallback<RenderedChunk, Option<String>>>`,

```rust
#[napi(ts_type = "undefined | string | ((chunk: RenderedChunk) => string | Promise<String>)")]
```

is wrong.

- For `xxx: Optioon<xxx>` NAPI-RS will generate type `xxx?: xxx`, so `undefined` is unnecessary.
- The return value in rust is `Option<String>`, so it should be marked as `string | undefined | null | void | Promise<string> | Promise<undefined> | Promise<null> | Promise<void>` in TS. However we have a `VoidNullable` to simplify it, we could just mark it as `MaybePromise<VoidNullable<string>>`

JoviDeCroock:
### Description

This changes the type to be `VoidNullable<string> | ((chunk: RenderedChunk) => MaybePromise<VoidNullable<string>>")`.

Resolves #682

### DIFF
--- a/crates/rolldown_binding/src/options/binding_output_options/mod.rs
+++ b/crates/rolldown_binding/src/options/binding_output_options/mod.rs
@@ -20,7 +20,7 @@ pub struct BindingOutputOptions {
   // assetFileNames: string | ((chunkInfo: PreRenderedAsset) => string);
   #[derivative(Debug = "ignore")]
   #[serde(skip_deserializing)]
-  #[napi(ts_type = "string | ((chunk: RenderedChunk) => MaybePromise<VoidNullable<string>>)")]
+  #[napi(ts_type = "VoidNullable<string> | ((chunk: RenderedChunk) => MaybePromise<VoidNullable<string>>)")]
   pub banner: Option<JsAsyncCallback<RenderedChunk, Option<String>>>,
   // chunkFileNames: string | ((chunkInfo: PreRenderedChunk) => string);
   // compact: boolean;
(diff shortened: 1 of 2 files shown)


crates/rolldown_binding/src/options/binding_output_options/mod.rs
```diff
@@ -20,7 +20,7 @@ pub struct BindingOutputOptions {
   // assetFileNames: string | ((chunkInfo: PreRenderedAsset) => string);
   #[derivative(Debug = "ignore")]
   #[serde(skip_deserializing)]
-  #[napi(ts_type = "undefined | string | ((chunk: RenderedChunk) => string | Promise<String>)")]
+  #[napi(ts_type = "VoidNullable<string> | ((chunk: RenderedChunk) => MaybePromise<VoidNullable<string>>)")]
```
hyf0:
```suggestion
  #[napi(ts_type = "Nullable<string> | ((chunk: RenderedChunk) => MaybePromise<VoidNullable<string>>)")]
```

You also need to run `pnpm build` to update generate types. You could run `just roll-node` to check if it's correct.

hyf0: `just fmt` will fix almost every lint errors.

hyf0 (approved):
Thank you!

PR #692, opened 2024-03-29
feat: expose `cwd` in input options

### LINKED ISSUE #687: Allow to pass `cwd` option via `InputOptions`
The rolldown in rust is already support passing `cwd` option.

https://github.com/rolldown/rolldown/blob/31926bd88f50765b9da113ee545ec7e4c0b439a1/packages/rolldown/src/options/input-options-adapter.ts#L17

But we don't exposed it in the `InputOptions`.

https://github.com/rolldown/rolldown/blob/31926bd88f50765b9da113ee545ec7e4c0b439a1/packages/rolldown/src/options/input-options.ts#L10-L15

---

Allowing to change the current workspace by passing `cwd` option is designed on purpose. This allow users to create multiple bundler instances with different `cwd` in the single thread.

So far it's very useful in testing, which enable us to compile and test multiple fixtures concurrently in the same thread.

jwnx:
Exposes `cwd` prop in `InputOptions` and updates rolldown tests to
leverage it. Closes #687

### DIFF
--- a/packages/rolldown/tests/cli/fixtures/basic/rolldown.config.js
+++ b/packages/rolldown/tests/cli/fixtures/basic/rolldown.config.js
@@ -1,24 +1,15 @@
-import path from 'node:path'
-import { fileURLToPath } from 'node:url'
-
-const __dirname = path.dirname(fileURLToPath(import.meta.url))
-
 export default {
-  // input: 'src/index.js',
-  input: path.resolve(__dirname, 'src/index.js'),
+  input: 'src/index.js',
   output: [
     {
-      // dir: 'build',
-      // file: 'build/bundle.js',
-      dir: path.resolve(__dirname, 'build'),
-      file: path.resolve(__dirname, 'build/bundle.js'),
+      dir: 'build',
+      file: 'build/bundle.js',
     },
   ],
   resolve: {
     conditionNames: ['import'],
     alias: {
-      // modules: 'src/modules',
-      modules: path.resolve(__dirname, 'src/modules'),
+      modules: 'src/modules',
     },
   },
 }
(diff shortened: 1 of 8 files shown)


jwnx:
@hyf0 I refactored rolldown tests to use `cwd` for testing purposes, but if you find it unnecessary I can remove this change.

packages/rolldown/tests/fixtures/input/array/_config.ts
```diff
@@ -4,7 +4,8 @@ import { expect } from 'vitest'
 import { getOutputChunkNames } from '@tests/utils'
 
 const config: RollupOptions = {
-  input: [path.join(__dirname, 'main.js'), path.join(__dirname, 'entry.js')],
+  input: ['main.js', 'entry.js'],
+  cwd: __dirname,
```
hyf0: Since you already pass `cwd` in `packages/rolldown/tests/fixture.test.ts`, this is unnessary to pass it again.

packages/rolldown/tests/fixture.test.ts
hyf0, line 1: You could also remove this line

https://github.com/rolldown/rolldown/blob/31926bd88f50765b9da113ee545ec7e4c0b439a1/packages/rolldown/tests/fixture.test.ts#L37

Rolldown will emit to `<CWD>/dist` by default.

packages/rolldown/tests/fixtures/input/array/_config.ts
jwnx: Done!

packages/rolldown/tests/fixture.test.ts
jwnx: Removed

packages/rolldown/tests/cli/fixtures/basic/rolldown.config.js
jwnx, line 1: @hyf0 I was able to remove a lot of path manipulation from this file and tests are passing just fine. Let me know if you think we should keep the old code.

hyf0 (approved):
Well done!

PR #693, opened 2024-03-29
feat: add footer output option

PengBoUESTC:
### Description

add `footer` output option based on `banner` option

### Test Plan

---


underfin (approved):

underfin:
Thank you.

PengBoUESTC:
> Thank you.

thanks!

packages/rolldown/tests/fixtures/output/footer/_config.ts
```diff
@@ -0,0 +1,29 @@
+import type {
+  RollupOptions,
+  RolldownOutput,
+  RolldownOutputChunk,
+} from '../../../../src'
+import { expect } from 'vitest'
```
hyf0, line 6: This should be `import { ... } from 'rolldown'` directly. And you don't need to import these types with using `defineTest`. Types will be inferred automatically.

PengBoUESTC: Ok, I will fix these .

hyf0: Nice. I create an [issue](https://github.com/rolldown/rolldown/issues/695) for this. Not just the banner test, some other tests also have this issue.

PengBoUESTC: Ok

PR #694, opened 2024-03-29
perf: avoid rebuild sourcemap at remapping one sourcemap

underfin:
### Test Plan

---


hyf0 (approved):
Wired bnechmark results.

---

See it wrong, ignore me.

PR #697, opened 2024-03-29
refactor: replace rolldown type import in tests with defineTest helper function

### LINKED ISSUE #695: Refactor current tests using `defineTest`
This should be `import { ... } from 'rolldown'` directly. And you don't need to import these types with using `defineTest`. Types will be inferred automatically.

_Originally posted by @hyf0 in https://github.com/rolldown/rolldown/pull/693#discussion_r1544207134_

---

Bad case:

Take https://github.com/rolldown/rolldown/pull/693#discussion_r1544207134_  as a bad case.

- `import ... from '../../../src'` should be avoided. Use `import ... from 'rolldown' instead.

- Use `defineTest` to get auto type hint, these's no need to do it manually.

---

Try refactor tests to make them look like this one

https://github.com/rolldown/rolldown/blob/main/packages/rolldown/tests/fixtures/plugin/build-start/_config.ts

PengBoUESTC:
### Description
replace rolldown type import in tests with defineTest helper function 
fix for https://github.com/rolldown/rolldown/issues/695

### Test Plan

---

### DIFF
--- a/packages/rolldown/tests/fixtures/external/array/_config.ts
+++ b/packages/rolldown/tests/fixtures/external/array/_config.ts
@@ -1,4 +1,4 @@
-import { defineTest } from '@tests/index'
+import { defineTest } from '@tests'
 
 export default defineTest({
   config: {
--- a/packages/rolldown/tests/fixtures/external/array/_config.ts
+++ b/packages/rolldown/tests/fixtures/external/array/_config.ts
@@ -1,4 +1,4 @@
-import { defineTest } from '@tests'
+import { defineTest } from '@tests/index'
 
 export default defineTest({
   config: {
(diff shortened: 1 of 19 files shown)


packages/rolldown/tests/fixtures/external/array/_config.ts
```diff
@@ -1,4 +1,4 @@
-import { defineTest } from '@tests'
+import { defineTest } from '@tests/index'
```
hyf0:
```suggestion
import { defineTest } from '@tests'
```

PengBoUESTC: Done

hyf0 (approved):
Thank you!

PR #699, opened 2024-03-29
chore(ci/bench): tracking benchmarks of js instead of rust

hyf0:
### Test Plan

---

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -1,4 +1,4 @@
-name: codspeed-benchmarks
+name: Codspeed Benchmarks
 
 on:
   push:
@@ -9,6 +9,10 @@ on:
       - '**/*.toml'
       - 'Cargo.lock'
       - 'rust-toolchain.toml'
+      - 'packages/rolldown/**'
+      - 'packages/bench/**'
+      - 'package.json'
+      - 'pnpm-lock.yaml'
   pull_request:
     types: [opened, synchronize]
     paths:
@@ -16,6 +20,10 @@ on:
       - '**/*.toml'
       - 'Cargo.lock'
       - 'rust-toolchain.toml'
+      - 'packages/rolldown/**'
+      - 'packages/bench/**'
+      - 'package.json'
+      - 'pnpm-lock.yaml'
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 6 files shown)


hyf0:
* **#699** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/699?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
@hyf0 Just to give you a heads-up: I've found CodSpeed's measures of code which is a combination of JS and Rust via NAPI to be very inaccurate. https://github.com/CodSpeedHQ/action/issues/96

In Oxc I've implemented a very hacky workaround for this. It runs the benchmarks *outside* of CodSpeed's instrumentation and records the results to a file - "bench X took N ms". Then a 2nd pseudo-benchmark runs *within* CodSpeed's instrumentation but it just reads the results of the real benchmark from the file and executes random simple instructions in a loop for the length of time required to consume N ms. It's the 2nd pseudo-benchmark which CodSpeed actually measures, and uploads to their back end.

https://github.com/oxc-project/oxc/blob/c1a2958a5a8a221b7f1412e4c3f0402001b885b4/.github/workflows/benchmark.yml#L110-L187

This is a totally horrible hack, and I'm working on a better solution, but just wanted to point out that it seems CodSpeed's measures aren't currently reliable for JS+Rust mixed code.

hyf0:
@overlookmotel  Thanks for sharing this. This is interesting. You could see the pure rust benchmark `threejs-bundle` using codspeed takes `834.1 ms` and js benchmark `threejs` takes `848.1 ms`. rust `threejs10x-bundle` takes `8.6 s` and js `threejs10x` takes `	9 s`.

It seems that the codspeed doesn't affect the performance much. Maybe it's because that `parser` is so fast that the affection of codspeed become important. However bundler already takes too much time than codspeed itself, so it become ignoreable.

overlookmotel:
I don't *think* it had anything to do with the speed of the parser. The big problem I was finding is that the *relative* speed of JS vs Rust shown on CodSpeed is completely wrong.

Basically, it seems to think Rust is faster than it really is, or JS is slower than it really is.

I noticed this problem when testing AST transfer. AST transfer removes a load of code from Rust side (no serde serialization), and adds more JS code instead (JS deserializer reading from buffer and creating JS objects).

When comparing before/after benchmarks outside of CodSpeed (i.e. real world usage), AST transfer produces a ~5x speed-up. But under CodSpeed's instrumentation it says it's a performance regression - completely inaccurate.

So where you might find you hit this problem is if (a) you make changes to how data is transferred between Rust and JS, or (b) if you e.g. replace a JS plugin with one written in Rust. In these cases, when CodSpeed tells you something is faster/slower, it seems it can't be trusted. It may be advisable to double-check by running the benchmarks locally to confirm before merging any changes like this.

Hope this is useful info.

hyf0:
Hit the same problem in https://github.com/rolldown/rolldown/pull/703. I'm using the your hack solution :).

PR #703, opened 2024-03-31
chore: bump oxc 0.11.0

underfin:
### Test Plan

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -129,9 +129,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.4.2"
+version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ed570934406eb16438a4e976b1b4500774099c13b8cb96eec99f620f05090ddf"
+checksum = "cf4b9d6a944f767f8e5e0db018570623c85f3d925ac718db4e06d0187adb21c1"
 
 [[package]]
 name = "bumpalo"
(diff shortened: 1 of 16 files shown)


crates/rolldown_resolver/Cargo.toml
```diff
@@ -13,7 +13,7 @@ repository.workspace = true
 workspace = true
 
 [dependencies]
-oxc_resolver   = { workspace = true }
+oxc_resolver   = { workspace = true, features = ['package_json_raw_json_api'] }
```
Boshen: What's needed from `package_json_raw_json_api`? Let's try and avoid this for performance reasons.

crates/rolldown_error/Cargo.toml
```diff
@@ -16,5 +16,5 @@ workspace = true
 [dependencies]
 ariadne    = { workspace = true }
 napi       = { workspace = true, optional = true }
-oxc        = { workspace = true }
+oxc        = { workspace = true, features = ["sourcemap"] }
```
Boshen: You should enable the feature on the workspace level

underfin: Done

crates/rolldown_resolver/Cargo.toml
underfin: I upgrade the version, already fix it.

underfin:
Look like the js side ci time is not correct at now. I run threejs10x at local using debug build, the pr is 800ms, the main branch is 1.6s.

hyf0:
> Look like the js side ci time is not correct at now. I run threejs10x at local using debug build, the pr is 800ms, the main branch is 1.6s.

Prabably due to https://github.com/rolldown/rolldown/pull/699#issuecomment-2028027139.

underfin:
Could you add the pervious rust benchmark to ci. Let us see the actual performance.

hyf0:
> Could you add the pervious rust benchmark to ci. Let us see the actual performance.

I'm fixing the current one. It should shows the actual performance after being fixed.

hyf0 (approved):

PR #704, opened 2024-03-31
chore: reserved keywords need to be considered as used variables.

woai3c:
### Description
Test code:
```js
// index.js
import { interface } from './answer'
import { interface as interface2 } from './lib'

console.log(interface + interface2)

// answer.js
export const interface = 100
console.log(interface)

// lib.js
export const interface = 200
console.log(interface)

```
Before output(has reserved keyword: `interface`):
```js
// answer.js
const interface = 100; // System reserved keywords cannot be used as variables.
console.log(interface);

// lib.js
const interface$1 = 200;
console.log(interface$1);

// index.js
console.log(interface + interface$1);
```
After output(no reserved keyword):
```js

// answer.js
const interface$1 = 100;
console.log(interface$1);

// lib.js
const interface$2 = 200;
console.log(interface$2);

// index.js
console.log(interface$1 + interface$2);

```

### Test Plan

---


underfin (approved):

crates/rolldown/src/utils/reserved_names.rs
```diff
@@ -0,0 +1,65 @@
...
+  "float",
+  "goto",
+  "int",
+  "long",
+  "native",
+  "short",
+  "synchronized",
+  "throws",
+  "transient",
+  "volatile",
```
Boshen: Are these ... java keywords?

underfin: Yeah. It is es3 RESERVED WORDS.

Boshen: They are not part of the current ecma spec https://tc39.es/ecma262/#prod-ReservedWord

woai3c: > Are these ... java keywords?

```
abstract, boolean, byte, char, double, final, float, goto, int, long, native, short, synchronized, throws, transient, volatile
```
These are reserved words in ES3, but they can be used safely in ES6 and later versions. Do they need to be removed?

woai3c: > They are not part of the current ecma spec https://tc39.es/ecma262/#prod-ReservedWord

Ok, I will remove these keywords.

underfin: Yeah. esbuild https://github.com/evanw/esbuild/blob/main/internal/js_lexer/js_lexer.go#L162-L212 and rollup https://github.com/rollup/rollup/blob/master/src/utils/RESERVED_NAMES.ts also not include them.

woai3c: @Boshen @underfin keep keywords consistent with rollup

PR #706, opened 2024-03-31
fix(node/bench): avoid affection of codspeed plugin in CI

hyf0:
### Description

codspeed has unkown problems for benchmarking program mixed with rust and js. For details:
- https://github.com/rolldown/rolldown/pull/699
- https://github.com/CodSpeedHQ/action/issues/96

We now use a hacky solution come up from @overlookmotel. We will first run a real benchmark to collect data, then run facade benchmark of vitest using the previous real benchmark data to pretend doing the bundling work.

### Test Plan

---


hyf0:
* **#706** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/706?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

hyf0:
Give up so far. I can't make codspeed work right for benchmarking

```js
function sleep(ms: number) {
  return new Promise((resolve) => setTimeout(resolve, ms))
}
```

overlookmotel:
From my (incomplete) understanding, CodSpeed does not actually measure execution time (wall clock). It measures number of instructions executed and then converts that to milliseconds via some heuristic.

I imagine this is why `setTimeout` doesn't work as expected. No instructions execute, so no "time" is taken.

This is why in my hack on Oxc I have it execute meaningless instructions in a loop. That *does* get measured.

If you like, I can submit a PR applying the same technique?

hyf0:
> From my (incomplete) understanding, CodSpeed does not actually measure execution time (wall clock). It measures number of instructions executed and then converts that to milliseconds via some heuristic.
> 
> I imagine this is why `setTimeout` doesn't work as expected. No instructions execute, so no "time" is taken.
> 
> This is why in my hack on Oxc I have it execute meaningless instructions in a loop. That _does_ get measured.
> 
> If you like, I can submit a PR applying the same technique?

Free to do it. Just looked your PR again. Nice solution, with this hack we could tracking benchmark rust and nodejs together. Solved my another requirement.

hyf0:
@overlookmotel please be on hold. I'm thinking dropping codspeed using https://github.com/benchmark-action/github-action-benchmark.

overlookmotel:
Sure. Personally, I really like Codspeed. It's just this one problem that is annoying. For benchmarking pure Rust code, it's excellent.

PR #707, opened 2024-03-31
fix: output hook ts type

PengBoUESTC:
### Description
output hook ts type Outputs -> BindingOutputs

### Test Plan

---


PengBoUESTC:
close in https://github.com/rolldown/rolldown/pull/708

PR #708, opened 2024-03-31
feat: add bindingify output hook

PengBoUESTC:
### Description
add bindingify output hook for packages/rolldown

### Test Plan

---


hyf0 (approved):
Thank you!

PR #722, opened 2024-04-02
feat: use regex to format file name template

PengBoUESTC:
### Description

use `regex` to format file name template, and support `hash` option for name template

### Test Plan

---


PengBoUESTC:
@hyf0,  I need some suggestions about this PR =_=.

hyf0:
Closed due to there is better solution on this: https://github.com/rolldown/rolldown/pull/1496.

PR #696, opened 2024-03-29
refactor: passing data of `BindingOutputs` lazily

underfin:
### Description

Lazy get BindingOutputs  fields to improve performance.

### Test Plan

Not need.

---

PR #727, opened 2024-04-02
fix(cli): user-friendly console output of chunk information

kazupon:
### Description

The rolldown CLI only outputs the following exit message and time when a build is executed with the command.

This PR user-friendly output to console like vite.

- before
rolldown-cli-friendly-1

- after
rolldown-cli-friendly-2

### Test Plan

---

### DIFF
--- a/packages/rolldown/src/cli/commands/bundle.ts
+++ b/packages/rolldown/src/cli/commands/bundle.ts
@@ -1,14 +1,9 @@
 import { performance } from 'node:perf_hooks'
-import { gzip } from 'node:zlib'
-import { promisify } from 'node:util'
 import consola from 'consola'
 import { colors, ColorFunction } from 'consola/utils'
-import { RolldownOptions, rolldown } from '../../index'
+import { RolldownOptions, RollupOutput, rolldown } from '../../index'
 import { RolldownConfigExport } from '../../types/rolldown-config-export'
 import { arraify } from '../../utils'
-import { RollupOutput } from 'dist/index.mjs'
-
-const compress = promisify(gzip)
 
 export async function bundle(configExport: RolldownConfigExport) {
   const options = arraify(configExport)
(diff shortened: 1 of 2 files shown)


packages/rolldown/src/cli/commands/bundle.ts
```diff
@@ -14,13 +19,133 @@ export async function bundle(configExport: RolldownConfigExport) {
 }
 
 async function bundleInner(options: RolldownOptions) {
-  const start = performance.now()
+  const dir = options.output?.dir ?? 'dist'
```
kazupon, line 17: The default for `dir` is held by the Rust side.
https://github.com/rolldown/rolldown/blob/667c7df335e998cd52275da2225827b0d7e17201/crates/rolldown/src/utils/normalize_options.rs#L83

I think we need a mechanism to share `dir` defaults between JS side and Rust side.
If we change this value on either side, it may not work.

packages/rolldown/src/cli/commands/bundle.ts
```diff
@@ -14,13 +19,133 @@ export async function bundle(configExport: RolldownConfigExport) {
...
+        ),
+      }
+    }),
+  )
+}
+
+async function calculateCompressedSize(
+  code: string | Uint8Array,
+): Promise<number | null> {
+  const compressed = await compress(
```
hyf0: This would really slow down the speed of cli for bigger project with doing real compressing action. Nt sure if this is good idea.

packages/rolldown/src/cli/commands/bundle.ts
hyf0: For non-gziped size, you should able to get them in the returned value of `write` or `generate`. However, this might still slow down the performance potentially if we didn't do the optimization in #696. 

I suggest we now only print non-gized sizes. It should solves
- the performance problem of large project
- not knowing `default for dir` problem

kazupon:
I have dropped the display gzip size feature.

hyf0 (approved):
Thanks!

PR #729, opened 2024-04-03
chore: remove unused crates

Boshen:
I improved https://github.com/Boshen/cargo-shear even more.


hyf0:
Does `cargo shear`  support exit with 1 if found any unused deps. Might want to add it in CI.

hyf0:
Saw it. Cool I will add it in CI.

hyf0 (approved):

PR #733, opened 2024-04-03
feat(rust): don't generate sourcemap for virtual modules

underfin:
### Description

The runtime module and module which path starts with `\0` shouldn't generate sourcemap. Ref see https://github.com/rollup/rollup/blob/master/src/Module.ts#L279.

### Test Plan

Updated test.

---

### DIFF
--- a/crates/rolldown/src/chunk/mod.rs
+++ b/crates/rolldown/src/chunk/mod.rs
@@ -113,7 +113,7 @@ impl Chunk {
                 .expect_file()
                 .relative_path(&input_options.cwd)
                 .to_string_lossy().as_ref(),
-           !output_options.sourcemap.is_hidden() && m.sourcemap
+           !output_options.sourcemap.is_hidden() && !m.is_virtual
         );
         Some((
           m.resource_id.expect_file().to_string(),
--- a/crates/rolldown/src/chunk/mod.rs
+++ b/crates/rolldown/src/chunk/mod.rs
@@ -113,7 +113,7 @@ impl Chunk {
                 .expect_file()
                 .relative_path(&input_options.cwd)
                 .to_string_lossy().as_ref(),
-           !output_options.sourcemap.is_hidden()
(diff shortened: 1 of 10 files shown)


crates/rolldown/src/chunk/mod.rs
```diff
@@ -113,7 +113,7 @@ impl Chunk {
                 .expect_file()
                 .relative_path(&input_options.cwd)
                 .to_string_lossy().as_ref(),
-           !output_options.sourcemap.is_hidden()
+           !output_options.sourcemap.is_hidden() && m.sourcemap
```
hyf0:
```suggestion
           !output_options.sourcemap.is_hidden() && m.is_virtual()
```

crates/rolldown_common/src/module/normal_module.rs
```diff
@@ -37,6 +37,8 @@ pub struct NormalModule {
   pub default_export_ref: SymbolRef,
   pub sourcemap_chain: Vec<rolldown_sourcemap::SourceMap>,
   pub is_included: bool,
+  // The runtime module and module which path starts with `\0` shouldn't generate sourcemap. Ref see https://github.com/rollup/rollup/blob/master/src/Module.ts#L279.
+  pub sourcemap: bool,
```
hyf0: Let's add `is_virtual` method in `NormalModule` directly and use it. It makes eaiser to understand why this modules are excluded from generating sourcemap instead of wondering why `sourcemap` is `false` and who control this value.

hyf0:
Shocked. How come this PR pass the windows ci.

crates/rolldown_common/src/module/normal_module.rs
underfin: DOne

underfin:
> Shocked. How come this PR pass the windows ci.

The error is flashing...

hyf0 (approved):

PR #754, opened 2024-04-05
fix(cli): Improvements to more brandaly CLI

kazupon:
### Description

I have worked the following improvements to make the CLI a bit more brandaly

#### usage

- before
rolldown-usage-before

- after
rolldown-usage-after

#### build

- before
rolldown-build-before

- after
rolldown-build-after


packages/rolldown/src/cli/env.ts
```diff
@@ -0,0 +1,28 @@
+import tty from 'node:tty'
+
+/**
+ * The following flags are based on conosola
+ * https://github.com/unjs/consola/blob/24c98ceb90c269a170fd116134d91803a89f2c9d/src/utils/color.ts#L7-L26
```
kazupon, line 5: @pi0
I would like to use `std-env` for these flags, but it seemed that tree-shaking was not very optimized for it. If we use `std-env` and can bundle only the necessary flags code, I would like to know.

If `std-env` is not optimized for tree-shaking, I can contribute to it.

packages/rolldown/src/cli/index.ts
```diff
@@ -31,16 +31,12 @@ const main = defineCommand({
     },
   },
   async run(ctx) {
-    if (ctx.args.help) {
-      await showUsage(ctx.cmd)
-      return
-    }
```
kazupon, line 37: This code is not necessary because `citty` side calls `showUsage`.

packages/rolldown/src/cli/index.ts
```diff
@@ -58,4 +54,4 @@ function parseArgs(args: Record<string, any>) {
   return { configPath }
 }
 
-runMain(main)
+runMain(main, { showUsage })
```
kazupon, line 57: @pi0
when we pass a custom `showUsage` to `runMain`, the `showUsage` and `renderUsage` that `citty` has are not tree-shaken. 😞

packages/rolldown/src/cli/utils.ts
```diff
@@ -1,11 +1,17 @@
...
 import nodePath from 'node:path'
+import { createConsola } from 'consola'
 import { ERR_UNSUPPORTED_CONFIG_FORMAT } from './errors.js'
 import { RolldownConfigExport } from '../types/rolldown-config-export.js'
 
 /**
- * @typedef {import('../rollup').RollupOptions} RollupOptions
+ * Console logger
  */
+export const logger = createConsola({
```
kazupon, line 10: The default consola instance has time. rolldown cli does not need it, so we introduced a custom consola.

packages/rolldown/src/cli/env.ts
pi0: std-env has ESM exports, is ~3.58kB in total and I tried to make the dependency graph as less coupled as possible.

Still, if you believe more tree-shaking is possible, please open an upstream issue explaining what exact import names you believe are not necessary bits and are not tree-shaken. 🙏🏼

packages/rolldown/src/cli/index.ts
pi0: `runMain` is a relatively simple wrapper for convenience of mainly built-in usage. Requiring to always pass a `showUage` method makes citty DX harder. you can copy it entirely if you like but as mentioned before, you can instead contribute to upsteam's showUsage also to make it better instead of forking it.

packages/rolldown/src/cli/env.ts
kazupon: Thank you for your comment! 😄 
I'll try to open as issue and share the my ideas to resolve it.

packages/rolldown/src/cli/index.ts
kazupon: Okay, I'll try to contribute it!

packages/rolldown/src/cli/usage.ts
hyf0, line 1: Nice work. I'm a little confused why we write these things manually. I was expected there are already frameworks that does these things.

kazupon: citty can take `showUsage` as an option to `runMain` to support customization by user.
You are correct, `showUsage` in this PR is almost same to citty except for the CLI title.
like oxc, `citty` is still a young CLI framework that has not yet reached major version 1.0. I will send some feedback. :)

kazupon: @hyf0 
Customization of `showUsage` is currently being worked on in this issue.
https://github.com/unjs/citty/issues/137
But, it seems to take a bit of time, so if you are in a hurry, you can merge it.
I will open another refactoring PR.

hyf0:
CI failed. Would you mind fixing it and let's merge this PR. I'm planning to do some refactor after this PR to support more out-of-the-box features.

hyf0 (approved):
Thanks

Issue #624, opened 2024-03-21
Rolldown doesn't support `output.file` option


kazupon:
## reproduction repo
https://github.com/kazupon/rolldown-output-repro1

## background
I noticed on #610

## reproduction step

```sh
npm install

npm run build

tree -L 2
.
├── README.md
├── app.js
├── dist
│   ├── app.js
│   └── app.js.map
├── node_modules
│   ├── @rolldown
│   ├── @rollup
│   ├── @types
│   ├── fsevents
│   └── rollup
├── package-lock.json
├── package.json
├── rolldown.js
├── rollup.js
└── rollup_build
    ├── rollup.js
    └── rollup.js.map

9 directories, 10 files
```

rolldown `write` output option:

```js
import { rolldown } from '@rolldown/node'

const bundle = await rolldown({
  input: 'app.js',
})

await bundle.write({
  format: 'esm',
  sourcemap: true,
  file: 'rolldown_build/rolldown.js'
})
```

rolldown cannot output to `'rolldown_build/rolldown.js'` with `file` option.

rollup can ouput with `file` option.

rolldown `write` should have the same result as rollup.

kazupon:
Ah, I see, the Rust side `write` does not receive output otpions from the JS side, it writes with the outputoptions passed to the Bundler when it initializes.
https://github.com/rolldown/rolldown/blob/ec85051013b58b31abbb92e50472a9b1e476e321/crates/rolldown_binding/src/bundler.rs#L19-L41

If we pass output options to write, then we have to test the performance ((de)serialize).

hyf0:
@kazupon 

> the Rust side write does not receive output otpions from the JS side

Nope. It does take the output options from js side, which is the options that passed from `write` or `generate`. I'm not sure the problem. But it probably due to we haven't support `output.file` option.

You should be able to use `output.entryFileName: 'rolldown_build/rolldown.js'` to do this. Or `output.dir: 'rolldown_build'` with `output.entryFileName: 'rolldown.js'`.

Notice that this only works out in single entry.

You see the functionality of `output.file` could be done with other options. I personally feel it's kind of overengineering. Let me talk with team about how we handle these kind of options.

sapphi-red:
FWIW Vite does not support `output.file` as we generate multiple files. https://github.com/vitejs/vite/issues/11096#issuecomment-1333878325, https://github.com/vitejs/vite/pull/14909

PR #758, opened 2024-04-05
fix: sourcemap sources path  relative to output file dirname

underfin:


packages/rollup-tests/test/sourcemaps/index.js
```diff
@@ -33,7 +33,8 @@ runTestSuiteWithSamples(
 						};
 						const outputOptions = {
 							exports: 'auto',
-							file: directory + '/_actual/bundle.' + format + '.js',
+							dir: directory + '/dist/',
```
hyf0: to get the same result, you could use workaround in #624.

```suggestion
							entryFileName:  directory + '/_actual/bundle.' + format + '.js',
```

underfin: The `entryFileName` hasn't add into binding. So here just using `dir`.

hyf0: Ok. If there isn't any downside, I'm ok with both.

hyf0 (approved):

PR #765, opened 2024-04-06
fix: normalize original sourcemap sources

underfin:

### DIFF
--- a/packages/rolldown/src/plugin/bindingify-build-hooks.ts
+++ b/packages/rolldown/src/plugin/bindingify-build-hooks.ts
@@ -3,7 +3,7 @@ import type { BindingPluginOptions } from '../binding'
 
 import type { Plugin } from './index'
 import { RolldownNormalizedInputOptions } from '../options/input-options'
-import { isEmptySourcemapFiled, transformSourcemap } from '../utils'
+import { isEmptySourcemapFiled } from '../utils'
 import path from 'path'
 import { SourceMapInputObject } from '../types/sourcemap'
 
@@ -144,7 +144,7 @@ export function bindingifyLoad(
 
     return {
       code: ret.code,
-      map: transformSourcemap(ret.map),
+      map: JSON.stringify(map),
     }
   }
 }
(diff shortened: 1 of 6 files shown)


packages/rolldown/src/plugin/bindingify-build-hooks.ts
```diff
@@ -119,6 +124,24 @@ export function bindingifyLoad(
...
+      return { code: ret.code }
+    }
+
+    let map =
+      typeof ret.map === 'object'
+        ? ret.map
+        : (JSON.parse(ret.map) as SourceMapInputObject)
+    if (!isEmptySourcemapFiled(map.sources)) {
+      // normalize original sourcemap sources
+      // Port form https://github.com/rollup/rollup/blob/master/src/utils/collapseSourcemaps.ts#L180-L188.
```
hyf0, line 137: Is this a temporary workaround? Is the practical way of doing this in rust's `collapseSourcemaps` function?

underfin: The rust `collapseSourcemaps` is a stable sourcemap remapping.  Here normalize sources path to support rollup cases, it is unrelevant to remapping.

hyf0: I saw this is for test https://github.com/rollup/rollup/blob/64870c31062983a741832d8cd2d4bcdf55e4de55/test/sourcemaps/samples/loaders/_config.js.

What happen if we have the same logic in rust plugins? 

I saw the test is to ensure the absolute path become relative. So I guess this is the normalization this PR does. Is my understanding right?

image

image

hyf0:
I still have questions, but let's merge it first.

---

https://github.com/rolldown/rolldown/pull/765#discussion_r1554556939

hyf0 (approved):

packages/rolldown/src/plugin/bindingify-build-hooks.ts
underfin: Here it is intend to test `sourceRoot` feature. You should see here https://github.com/rollup/rollup/blob/64870c31062983a741832d8cd2d4bcdf55e4de55/test/sourcemaps/samples/loaders/_config.js#L27-L30.

underfin: > What happen if we have the same logic in rust plugins?

The logic maybe should be put into rust side, but I'm not sure at now. It need to loader `sourcemap` is stable, the plugin should normalize `sourceRoot`.

PR #769, opened 2024-04-06
feat: support plugin.order 

moushicheng:
### Description
⚠️This pr is under construction

support plugin attr order:
[rollup doc](https://rollupjs.org/plugin-development/#build-hooks)
> order: "pre" | "post" | null
> If there are several plugins implementing this hook, either run this plugin first ("pre"), last ("post"), or in the user-specified > position (no value or null).
```js
export default function resolveFirst() {
	return {
		name: 'resolve-first',
		resolveId: {
			order: 'pre',
			handler(source) {
				if (source === 'external') {
					return { id: source, external: true };
				}
				return null;
			}
		}
	};
}
```
If several plugins use "pre" or "post", Rollup runs them in the user-specified order. This option can be used for all plugin hooks. For parallel hooks, it changes the order in which the synchronous part of the hook is run.

## confusion
I want to sort the plugins , i think it should implementation in PluginDriver.
site: crates/rolldown_plugin/src/plugin_driver/build_hooks.rs
```js
impl PluginDriver {
  pub async fn build_start(&self) -> HookNoopReturn {
    // TODO should call `build_start` of all plugins in parallel
    for (plugin:&Box<dyn Plugin>, ctx) in &self.plugins {
      plugin.build_start(ctx).await?;
    }
    Ok(())
  }
}
```
but beacause of plugin type is  ```&Box<dyn Plugin>```,i cant get buildstart hook options although i delivery attribute order.
i want to change plugin type to JsPlugin,but JsPlugin is define in crates rolldown_binding.This does not seem to be a dependency.
so how can i get completly plugin.build_start? I admit I'm not very familiar with rust...

### DIFF
--- a/crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
+++ b/crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
@@ -17,14 +17,31 @@ use super::{
 };
 
 #[napi_derive::napi(object, object_to_js = false)]
+#[derive(Default)]
+pub struct HookOption {
+  #[napi(ts_type = "'pre'|'post'|null")]
+  pub order: Option<String>,
+  pub sequential: Option<bool>,
+}
+
+#[napi_derive::napi(object, object_to_js = false)]
+pub struct BuildStartHookOption {
+  #[napi(ts_type = "(ctx: BindingPluginContext) => MaybePromise<VoidNullable>")]
+  pub handler: MaybeAsyncJsCallback<BindingPluginContext, ()>,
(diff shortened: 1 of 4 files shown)


hyf0:
Thanks for your work. However it's better to create an issue to discuss about the feature rather than just throwing a PR.

Let's create an issue about supporting plugin order and have a fully investigation on it.

underfin:
Close it by https://github.com/rolldown/rolldown/pull/1929, it avoid binding `order` for each hooks.

PR #798, opened 2024-04-08
fix: passing `BindingResolveOptions#alias` from js to rust with maintaining order

### LINKED ISSUE #790: Passing `BindingResolveOptions#alias` from js to rust with maintaining order
https://github.com/rolldown/rolldown/blob/c304a0dcde43234654626f8695cd745f37de4d33/crates/rolldown_binding/src/options/binding_input_options/binding_resolve_options.rs#L8-L11

### Use `Option<Vec<(String, Vec<String>)>>>`

The problem is that NAPI-RS so far doesn't support tuples

### Define a temporary struct like

```rust
pub struct AliasItem {
  pub name: String,
  pub paths: Vec<String>,
}
```

and write `Option<Vec<AliasItem>>>`.

---

I personally prefer using tuples on this case. However this requires [discussing with NAPI-RS if this feature is really needed](https://github.com/napi-rs/napi-rs/issues/1372).

PengBoUESTC:
### Description

convert `alias` to `Vec<AliasItem>`

for issue https://github.com/rolldown/rolldown/issues/790

### DIFF
--- a/crates/rolldown_binding/src/types/binding_resolve_alias.rs
+++ b/crates/rolldown_binding/src/types/binding_resolve_alias.rs
@@ -1,0 +1,9 @@
+use serde::Deserialize;
+
+#[napi_derive::napi(object)]
+#[derive(Deserialize, Debug, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct AliasItem {
+  pub name: String,
+  pub paths: Vec<String>,
+}
--- a/crates/rolldown_binding/src/types/binding_resolve_alias.rs
+++ b/crates/rolldown_binding/src/types/binding_resolve_alias.rs
@@ -4,6 +4,6 @@ use serde::Deserialize;
 #[derive(Deserialize, Debug, Default)]
 #[serde(rename_all = "camelCase")]
 pub struct AliasItem {
-  pub name: String,
-  pub paths: Vec<String>,
+  pub find: String,
+  pub replacement: Vec<String>,
 }
--- a/packages/rolldown/src/options/input-options.ts
+++ b/packages/rolldown/src/options/input-options.ts
@@ -3,7 +3,11 @@ import {
   InputOptions as RollupInputOptions,
 } from '../rollup-types'
 import { ensureArray, normalizePluginOption } from '../utils'
-import { BindingInputOptions, BindingResolveOptions } from '../binding'
+import {
+  BindingInputOptions,
+  BindingResolveOptions,
+  AliasItem,
+} from '../binding'
 import { Plugin } from '../plugin'
 
 // TODO export compat plugin type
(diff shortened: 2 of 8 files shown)


crates/rolldown_binding/src/types/binding_resolve_alias.rs
```diff
@@ -0,0 +1,9 @@
+use serde::Deserialize;
+
+#[napi_derive::napi(object)]
+#[derive(Deserialize, Debug, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct AliasItem {
+  pub find: String,
+  pub replacement: Vec<String>,
```
hyf0:
```suggestion
  pub replacements: Vec<String>,
```

crates/rolldown_binding/src/types/binding_resolve_alias.rs
hyf0, line 1: filename should be `alias_item.rs`.

packages/rolldown/src/options/input-options.ts
```diff
@@ -17,8 +21,12 @@ export interface InputOptions {
   shimMissingExports?: BindingInputOptions['shimMissingExports']
 }
 
+export type Alias = Omit<AliasItem, 'replacement'> & {
+  replacement: string
+}
+
 export type RolldownResolveOptions = Omit<BindingResolveOptions, 'alias'> & {
-  alias?: Record<string, string>
+  alias?: Record<string, string> | Alias[]
```
hyf0: Please don't change the type of `alias`. For users, they still pass alias using object.

hyf0 (changes requested):

packages/rolldown/src/options/input-options.ts
PengBoUESTC: But I notice that, in `vite` and `rollup`, config `alias` is support for Array and object

hyf0: Yeah. Vite use https://github.com/rollup/plugins/tree/master/packages/alias#entries under the hood. The API it provided has different usages for `array` or `object`, which means you could not use `object` to [use `customResolver` API](https://github.com/rollup/plugins/blob/9e7f576f33e26d65e9f2221d248a2e000923e03f/packages/alias/README.md?plain=1#L162).

However, we don't have this need. In our cases, `object` and `array` does the same thing. To keep it simple, supporting `object` only would be fine.

PengBoUESTC: Ok, get it

crates/rolldown_binding/src/types/binding_resolve_alias.rs
PengBoUESTC: Done

crates/rolldown_binding/src/types/binding_resolve_alias.rs
PengBoUESTC: Done

hyf0 (approved):
Thanks

PR #807, opened 2024-04-09
feat: implement some expr side effect detector

woai3c:
### Description
I have enhanced some of the code for side-effect detection in expressions, but there are still parts that I'm currently unsure how to implement.

### DIFF
--- a/crates/rolldown/src/ast_scanner/side_effect_detector.rs
+++ b/crates/rolldown/src/ast_scanner/side_effect_detector.rs
@@ -110,6 +110,8 @@ impl<'a> SideEffectDetector<'a> {
       | Expression::RegExpLiteral(_)
       | Expression::FunctionExpression(_)
       | Expression::ArrowFunctionExpression(_)
+      | Expression::MetaProperty(_)
+      | Expression::ThisExpression(_)
       | Expression::StringLiteral(_) => false,
       Expression::ObjectExpression(obj_expr) => {
         obj_expr.properties.iter().any(|obj_prop| match obj_prop {
@@ -172,7 +174,6 @@ impl<'a> SideEffectDetector<'a> {
       Expression::PrivateInExpression(private_in_expr) => {
         self.detect_side_effect_of_expr(&private_in_expr.right)
       }
-      Expression::MetaProperty(_) | Expression::ThisExpression(_) => false,
       Expression::AwaitExpression(await_expr) => {
         self.detect_side_effect_of_expr(&await_expr.argument)
       }
--- a/crates/rolldown/src/ast_scanner/side_effect_detector.rs
+++ b/crates/rolldown/src/ast_scanner/side_effect_detector.rs
@@ -574,6 +574,7 @@ mod test {
   #[test]
   fn test_meta_property_expression() {
     assert!(!get_statements_side_effect("import.meta"));
+    assert!(!get_statements_side_effect("const meta = import.meta"));
     assert!(get_statements_side_effect("import.meta.url"));
     assert!(get_statements_side_effect("const { url } = import.meta"));
     assert!(get_statements_side_effect("import.meta.url = 'test'"));
--- a/crates/rolldown/src/ast_scanner/side_effect_detector.rs
+++ b/crates/rolldown/src/ast_scanner/side_effect_detector.rs
@@ -1,5 +1,5 @@
 use once_cell::sync::Lazy;
-use oxc::ast::ast::{IdentifierReference, MemberExpression};
+use oxc::ast::ast::{BindingPatternKind, Expression, IdentifierReference, MemberExpression};
 use rolldown_common::AstScope;
 use rustc_hash::FxHashSet;
 
@@ -70,6 +70,7 @@ impl<'a> SideEffectDetector<'a> {
     let MemberExpression::StaticMemberExpression(member_expr) = expr else {
       return true;
     };
+
     let prop_name = &member_expr.property.name;
     match &member_expr.object {
       oxc::ast::ast::Expression::Identifier(ident) => {
@@ -93,15 +94,11 @@ impl<'a> SideEffectDetector<'a> {
           prop_name.as_str(),
         ))
       }
-      oxc::ast::ast::Expression::ThisExpression(_) | oxc::ast::ast::Expression::MetaProperty(_) => {
-        false
-      }
       _ => true,
     }
   }
 
   fn detect_side_effect_of_expr(&self, expr: &oxc::ast::ast::Expression) -> bool {
-    use oxc::ast::ast::Expression;
     match expr {
       Expression::BooleanLiteral(_)
       | Expression::NullLiteral(_)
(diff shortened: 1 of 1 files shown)


crates/rolldown/src/ast_scanner/side_effect_detector.rs
```diff
@@ -161,23 +164,29 @@ impl<'a> SideEffectDetector<'a> {
...
-
+      Expression::BinaryExpression(binary_expr) => {
+        // For binary expressions, both sides could potentially have side effects
+        self.detect_side_effect_of_expr(&binary_expr.left)
+          || self.detect_side_effect_of_expr(&binary_expr.right)
+      }
+      Expression::PrivateInExpression(private_in_expr) => {
+        self.detect_side_effect_of_expr(&private_in_expr.right)
+      }
+      Expression::MetaProperty(_) | Expression::ThisExpression(_) => false,
```
hyf0: I think `this.xxx` is considering may have side effect.

We could consider`import.meta` doesn't have side effect, but `import.meta.xxx` is tricky and I would like to consider it have side effect so far.

crates/rolldown/src/ast_scanner/side_effect_detector.rs
```diff
@@ -161,23 +164,29 @@ impl<'a> SideEffectDetector<'a> {
...
+      Expression::AwaitExpression(await_expr) => {
+        self.detect_side_effect_of_expr(&await_expr.argument)
```
hyf0: Consider case like
```js
let thenable = {
  then() {
     console.log('hello')
   }
}

await thenable
```

hyf0 (changes requested):

crates/rolldown/src/ast_scanner/side_effect_detector.rs
```diff
@@ -526,4 +536,47 @@ mod test {
...
+    assert!(!get_statements_side_effect("const a = 1; const b = 2; a + b"));
+    // accessing global variable may have side effect
+    assert!(get_statements_side_effect("1 + foo"));
+    assert!(get_statements_side_effect("2 + bar"));
+  }
+
+  #[test]
+  fn test_private_in_expression() {
+    assert!(!get_statements_side_effect("#privateField in this"));
+    assert!(!get_statements_side_effect("#privateField in this.x"));
```
hyf0:
```suggestion
    assert!(get_statements_side_effect("#privateField in this.x"));
```

crates/rolldown/src/ast_scanner/side_effect_detector.rs
```diff
@@ -526,4 +536,47 @@ mod test {
...
+    // accessing global variable may have side effect
+    assert!(get_statements_side_effect("#privateField in bar"));
+    assert!(get_statements_side_effect("#privateField in foo"));
+  }
+
+  #[test]
+  fn test_this_expression() {
+    assert!(!get_statements_side_effect("this"));
+    assert!(!get_statements_side_effect("this.a"));
```
hyf0:
```suggestion
    assert!(get_statements_side_effect("this.a"));
```

crates/rolldown/src/ast_scanner/side_effect_detector.rs
```diff
@@ -526,4 +536,47 @@ mod test {
...
+    assert!(!get_statements_side_effect("this.a + this.b"));
```
hyf0:
```suggestion
    assert!(get_statements_side_effect("this.a + this.b"));
```

crates/rolldown/src/ast_scanner/side_effect_detector.rs
```diff
@@ -526,4 +536,47 @@ mod test {
...
+    assert!(get_statements_side_effect("this.a = 10"));
+  }
+
+  #[test]
+  fn test_meta_property_expression() {
+    assert!(!get_statements_side_effect("import.meta"));
+    assert!(!get_statements_side_effect("import.meta.url"));
```
hyf0:
```suggestion
    assert!(get_statements_side_effect("import.meta.url"));
```

crates/rolldown/src/ast_scanner/side_effect_detector.rs
woai3c: ok

crates/rolldown/src/ast_scanner/side_effect_detector.rs
```diff
@@ -526,4 +536,47 @@ mod test {
...
+    assert!(!get_statements_side_effect("{ url } = import.meta"));
```
hyf0:
```suggestion
    assert!(get_statements_side_effect("{ url } = import.meta"));
```

crates/rolldown/src/ast_scanner/side_effect_detector.rs
woai3c: ok

crates/rolldown/src/ast_scanner/side_effect_detector.rs
```diff
@@ -526,4 +536,47 @@ mod test {
...
+  }
+
+  #[test]
+  fn test_await_expression() {
+    assert!(!get_statements_side_effect("await 1 + 1"));
+    assert!(!get_statements_side_effect("const a = 1; const b = 2; await a + b"));
+    assert!(get_statements_side_effect("await foo + bar"));
+    assert!(get_statements_side_effect("await foo"));
+    assert!(get_statements_side_effect("await fetch('https://example.com')"));
+  }
```
hyf0:
```suggestion
     assert!(get_statements_side_effect("let foo; await foo;"));
  }
```

hyf0 (changes requested):

crates/rolldown/src/ast_scanner/side_effect_detector.rs
woai3c: The modification code for 'AwaitExpression' has been removed.

woai3c: > Consider case like
> 
> ```js
> let thenable = {
>   then() {
>      console.log('hello')
>    }
> }
> 
> await thenable
> ```

 The modification code for 'AwaitExpression' has been removed.

crates/rolldown/src/ast_scanner/side_effect_detector.rs
woai3c: > I think `this.xxx` is considering may have side effect.
> 
> We could consider`import.meta` doesn't have side effect, but `import.meta.xxx` is tricky and I would like to consider it have side effect so far.

I have made modifications to the code based on your feedback. Please take a look again to see if there are any issues with the code.

crates/rolldown/src/ast_scanner/side_effect_detector.rs
woai3c: > Consider case like
> 
> ```js
> let thenable = {
>   then() {
>      console.log('hello')
>    }
> }
> 
> await thenable
> ```

this case always be false.

hyf0: Yeah. But it has side effect, so it should be true.

woai3c:
@hyf0  The code for `AwaitExpression` has been reset. Both `ThisExpression` and `MetaProperty` have been modified as per your suggestions. Please review the updated code. Thank you.

By the way, I'd like to contribute more to Rolldown. Currently, I usually look for TODO comments in the code and then implement them. However, some TODOs seem quite challenging to implement. Could you give me some advice on what aspects of Rolldown I should focus on to contribute more effectively?

hyf0 (approved):
Thank you!

PR #824, opened 2024-04-10
refactor: distinguish recoverable error and diagnostic error 

underfin:
### Description

- using `anyhow::Error` as recoverable error 
- using `Build::Error` as diagnostic error 
- The `InterError` is a enum to wrap the recoverable error and diagnostic error.


hyf0:
I think about this before. The problem here is that rolldown might return multiple errors in any place and this behavior of returning multiple errors is infectious just like async function.

To not end up with situation like that, `BatchedErrors` was introduced. `BatchedErrors` could be used to return single or multiple errors. So if there isn't very good reason, we probably should keep this way.

crates/rolldown_error/src/lib.rs
```diff
@@ -10,6 +10,8 @@ use sugar_path::SugarPath;
 
 pub use crate::{build_error::BuildError, event_kind::EventKind};
 
+pub type Result<T> = std::result::Result<T, BuildError>;
```
hyf0:
```suggestion
pub type BuildResult<T> = std::result::Result<T, BuildError>;
```

I prefer don't be conflicted with official `Result` of rust std. That's why `BuildError` was not called `Error`.

hyf0:
Think again we do have a few cases that we are sure that there will be single error only. These functions should be edge of the program to avoid causing infectious changes.

crates/rolldown/src/chunk/mod.rs
```diff
@@ -79,7 +77,7 @@ impl Chunk {
     options: &SharedOptions,
     graph: &LinkStageOutput,
     chunk_graph: &ChunkGraph,
-  ) -> BatchedResult<ChunkRenderReturn> {
+  ) -> Result<ChunkRenderReturn> {
```
hyf0:
```suggestion
  ) -> BatchedResult <ChunkRenderReturn> {
```

Not in the edge of the program, should keep `BatchedResult`.

crates/rolldown/src/stages/bundle_stage/mod.rs
```diff
@@ -92,7 +92,7 @@ impl<'a> BundleStage<'a> {
     let mut assets = vec![];
 
     render_chunks(self.plugin_driver, chunks).await?.into_iter().try_for_each(
-      |chunk| -> Result<(), BuildError> {
+      |chunk| -> rolldown_error::Result<()> {
```
hyf0: Not in the edge of the program, should changed to `BatchedResult`.

crates/rolldown_plugin/src/plugin_driver/build_hooks.rs
```diff
@@ -62,7 +59,7 @@ impl PluginDriver {
   pub async fn render_chunk(
     &self,
     mut args: HookRenderChunkArgs<'_>,
-  ) -> Result<(String, Vec<SourceMap>), BuildError> {
+  ) -> Result<(String, Vec<SourceMap>)> {
```
hyf0: Not in the edge of the program, should change to `BatchedResult`.

hyf0 (changes requested):

hyf0:
One small thing. I'm ok with PR without description, but please try to make the title more meaningful like "Introduce `type Result<T> = Result<T, BuildError>`" or "New type alias for `Result<T, BuildError>`".

crates/rolldown_error/src/lib.rs
underfin: I don't think is correct. Like rustc has `std::result::Result`, `std::io::Result`, etc. It is better has `rolldown:result`.

underfin:
> I think about this before. The problem here is that rolldown might return multiple errors in any place and this behavior of returning multiple errors is infectious just like async function.

> To not end up with situation like that, BatchedErrors was introduced. BatchedErrors could be used to return single or multiple errors. So if there isn't very good reason, we probably should keep this way.

Actually, i think we could make `BuildError` handle multiple errors, it will avoid having different `Error` and `Result` at rolldown.

crates/rolldown_error/src/lib.rs
hyf0: I think this not a question of correct or wrong. `std::result::Result` is included in the scope by default. You could use it without import statements. Using `BuildResult` to avoid conflict make you could write official `Result` and `BuildResult` together. Otherwise you need write `rolldown_error::Result<>` explicitly.

As I said above, this is not a not a question of correct or wrong. It's about programing style and I think it's a better style due to above reasons. I'm open to other styles, if they prove to be more beneficial, but I hardly, in this cause, think "rustc did it this way" as a strong argument.

hyf0:
> > I think about this before. The problem here is that rolldown might return multiple errors in any place and this behavior of returning multiple errors is infectious just like async function.
> 
> > To not end up with situation like that, BatchedErrors was introduced. BatchedErrors could be used to return single or multiple errors. So if there isn't very good reason, we probably should keep this way.
> 
> Actually, i think we could make `BuildError` handle multiple errors, it will avoid having different `Error` and `Result` at rolldown.

I think about this too. That will make the error define like 
```
enum Error {
 A,
 B,
 Many<Vec<Error>>
}
```

which could lead us to mutiple level `Error::Many`, then we end up with a `flat_all` method to recursively flat them. I want to avoid this situation.
Error is error. Mutiple errors is not a special form of error and should not be consider as the same thing.

underfin:
> Mutiple errors is not a special form of error and should not be consider as the same thing.

Yeah. It makes sense. But the rolldown `Error` `BatchedErrors` is complex, it will be caused some place could be to `Error` or `BatchedErrors`.It's important to write code for developers. One `Error` `Result` type makes the code clear. 
We could realize it as rolldown maintainer, avoid it to spread other peoples.

hyf0:
1.
> > Mutiple errors is not a special form of error and should not be consider as the same thing.
> 
> Yeah. It makes sense. But the rolldown `Error` `BatchedErrors` is complex

When you are saying that the current `BatchedErrors` implementation

```rust
pub struct BatchedErrors(Vec<BuildError>);
```

 is complex, I saw you added

``` rust
#[derive(Debug)]
pub enum Error {
  Single(BuildError),
  Vec(Vec<BuildError>),
}
```
I do not judge which one is simple, but I don't see how `BatchedErrors` is complex.

2.
> it will be caused some place could be to `Error` or `BatchedErrors`.

Yeah. I said above already:

> To not end up with situation like that, BatchedErrors was introduced. BatchedErrors could be used to return single or multiple errors. So if there isn't very good reason, we probably should keep this way.
> https://github.com/rolldown/rolldown/pull/824#issuecomment-2046831984

3.
> It's important to write code for developers. One `Error` `Result` type makes the code clear.

I reviewed this PR deeply and let's how it makes code cleaner.

- Almost all changes are renaming `-> BatchedResult<T>` to `-> Result<T>`

There are literal no changes on how people write rust code. No `Error::Single` or `Error::Many` was called. If nothing changes, how this make code cleaner in the first place?

- Let's see how this affect rust usages of rolldown

```rust
let output = Bundler::new().generate();
// To see the result, you need to import `rolldown::{Result, Error}` first.
match output {
  Ok(_) => ...,
  Err(err) => match err {
    Single(e) => {
      // some code
    },
   Many(errs) => {
     // errors.iter().for_each(|| => {
        // some code
    })
   }
  }
}

// Wait we could write it like
match output {
  Ok(_) => ...,
  Err(single_or_many) =>  single_or_many.into_vec().iter().for_each(|| {
      // some code
    })
  }
}

// Wait again, we could already do that using `BatchedErrors`
match output {
  Ok(_) => ...,
  Err(batched_errors) =>  batched_errors.iter().for_each(|| {
      // some code
    })
  }
}
```

If you think above one is more dx friendly, I could tell you it's not. If you think both two are fine, what's the point of rewriting code that already be fine.

image

4.
> We could realize it as rolldown maintainer, avoid it to spread other peoples.

`BatchedError` and `BatchedResult` are intentional to named a little twisty to force dev and rust users of rolldown to know that there might be multiple errors here.

underfin:
> Almost all changes are renaming -> BatchedResult<T> to -> Result<T>

Yeah. But i want to you know as a core and plugin developer, you only write `Result<T>`, don't worry to write `BatchedResult` or `Result`. It is the point I want say.

Here also has a change is `IntoBatchedResult`, using the pattern is easy to implement, we can use `collect::<Result<>>`. 

> Let's see how this affect rust usages of rolldown

Yeah. You list some usage patterns, but I think it should be better to use `std::Error::Fmt` to display error, the pr not implement this because make code changes smaller.

hyf0:
> Here also has a change is `IntoBatchedResult`, using the pattern is easy to implement, we can use `collect::<Result<>>`.

Thanks for letting me notice another problem. I hope you use `collect::<Result<>>` not because it's pretty. I'm not sure if you notice there is a behavior change. `IntoBatchedResult` wold collect all errors, while `collect::<Result<>>` would only return the first error.

---

I'm fine with change `trait IntoBatchedResult { fn into_batched_result() }` to `trait IntoBatchedResult { fn collect() }` if that satisfy your requirement.

underfin:
Thank you point to me, i had a mistake.

I rethink this, the pr implementation is also not better. I think we could use `BundleOutput#errors` to avoid `BatchedResult` returning multiple errors. What do you think about it?

hyf0:
CI failed. I'm working on it.

underfin:
Here has a unexpected behavior https://github.com/rolldown/rolldown/pull/824/commits/a5ff1fb394a43a1e65482b9035676fbac202a612#diff-67b9bbb80e81f2457380c93953dd8aaed8e403246493d0cbc70c4378832928dfL54

hyf0:
>  I'm working on it.

crates/rolldown/src/stages/generate_stage/mod.rs
```diff
@@ -44,7 +43,7 @@ impl<'a> GenerateStage<'a> {
   }
 
   #[tracing::instrument(skip_all)]
-  pub async fn generate(&mut self) -> BatchedResult<Vec<Output>> {
+  pub async fn generate(&mut self) -> Result<Vec<Output>> {
```
hyf0, line 46: This `anyhow::Result` aslo eats the internal source map error, which is not expected.

hyf0: Fixed it.

hyf0 (approved):

PR #825, opened 2024-04-10
feat: support running Js plugin in parallel via Worker

sapphi-red:
### Description

This PR implements parallel js plugin prototype.

context: https://github.com/sapphi-red/parallel-js-plugin-experiment/blob/main/benchmark.md, https://github.com/sapphi-red/parallel-js-plugin-experiment/blob/main/research-for-design.md

<details>
<summary>benchmark result on my machine</summary>

```
esbuild: 132.52ms
rolldown (single, sucrase): 184.33ms
rolldown (single, swc, transform): 201.12ms
rolldown (parallel, swc, transformSync): 211.14ms
rolldown (single, esbuild, transform): 221.13ms
rolldown (parallel, swc, transform): 261.34ms
rolldown (parallel, sucrase): 354.53ms
rolldown (single, swc, transformSync): 645.19ms
rolldown (parallel, babel, transformSync): 1996.61ms
rolldown (parallel, babel, transformAsync): 2011.78ms
rolldown (single, babel, transformSync): 2924.19ms
rolldown (single, babel, transformAsync): 2968.43ms
```

</details>

- remaining TODOs
  - [x] use a consistent naming (thread-safe-plugin or parallel-plugin)
  - [x] ~~resolve the TODOs written in the code~~
  - [x] find out why `rolldown (parallel, sucrase)` is 200ms slower than `rolldown (single, sucrase)` (100ms is caused by the worker instantiation, but not sure the rest 100ms is caused by) -> https://github.com/rolldown/rolldown/pull/825#issuecomment-2047295007
  - [x] resolve conflicts


hyf0:
`thread-safe` is probably not suitable to be the final name. IMO, everything is thread-safe in JavaScript because the js engine doesn't provide unsafe way to deal with thread.

crates/rolldown_plugin/src/plugin.rs
```diff
@@ -85,3 +87,52 @@ pub trait Plugin: Any + Debug + Send + Sync + 'static {
 }
 
 pub type BoxPlugin = Box<dyn Plugin>;
+
+#[derive(Debug)]
+pub enum PluginOrThreadSafePlugin {
```
hyf0: Good jobs. But there a few suggestions i'd like to make. All things related to js should probably be limited to `rolldown_binding`. With this principle, `rolldown_plugin` and `PluginDriver` should not get affected by if the plugin is a js/worker plugin or a rust plugin.

You should probably  move `PluginOrThreadSafePlugin` to `rolldown_binding` and redefine it like

```rust
impl Plugin for ThreadSafePlugin {

  fn transfoem(&self) {
    if self.options.single  {
      let permit = worker_manager.unwrap().acquire().await;
        let plugin = &plugins[permit.worker_index() as usize];
        f(plugin).await
    } else {
      let _permit = worker_manager.unwrap().acquire_all().await;
       future::join_all(plugins.iter().map(f)).await
    }
  }
  ... // other delegate hooks

}
```

So `ThreadSafePlugin` would be treated as a normal `Box<dyn Plugin>`, there's nothing special from the view of rolldown rust.

crates/rolldown_plugin/src/plugin_driver/build_hooks.rs
```diff
@@ -9,14 +9,19 @@ impl PluginDriver {
   pub async fn build_start(&self) -> HookNoopReturn {
     // TODO should call `build_start` of all plugins in parallel
     for (plugin, ctx) in &self.plugins {
-      plugin.build_start(ctx).await?;
```
hyf0: Said above, these changes to `PluginDriver` is unnecessary. They should be hide into the `ThreadSafePlugin`.

crates/rolldown_plugin/src/plugin_driver/mod.rs
```diff
@@ -1,25 +1,35 @@
...
 
 mod build_hooks;
 mod output_hooks;
 
 pub type SharedPluginDriver = Arc<PluginDriver>;
 
 pub struct PluginDriver {
-  plugins: Vec<(BoxPlugin, SharedPluginContext)>,
+  plugins: Vec<(PluginOrThreadSafePlugin, SharedPluginContext)>,
+  worker_manager: Option<WorkerManager>,
```
hyf0: Should store this in `rolldown_binding`, since it's a javescript thing.

crates/rolldown_plugin/src/worker_manager.rs
```diff
@@ -0,0 +1,69 @@
+use async_channel::{Receiver, Sender};
+
+pub(crate) struct WorkerManager {
+  free_workers_sender: Sender<u16>,
```
hyf0: This is probably enough as a prototype. However, since you could pass the value around, you could actually think of a way to pass `PluginOptions` or mutiple `PluginOptions` directly to the bundler. That's would be more friednly to interact with `ThreadSafePlugin`.

sapphi-red: Would you expand this a bit more?

hyf0: Yeah. This is actually a rough thought. Let me see if I could make it clear.

1. For ever  js `rolldown(...)` call, there is a uuid called `BundlerId` generated in js size or other values that's unique.
2. In js, in the calling of `write` or `generate`, when you normalize the plugins in js side, initialize these worker plugin with passing the `BundlerId` and a unique generated `WorkerId` for each worker.
3. Create the rust bundler with options and `BundlerId`
4. While the rust bundler is creating, follwing rust code would be executed
```rust
// Not runnable but you get the idea.
// Use channel would also works, just need to write more code.
static BundlerToWorkerMap: DashMap<BundlerId, DashMap<WokerId, Vec<PluginOptions>>>


struct WokerPlugin(BundlerId, WorkerId, Option<Vec<PluginOptions>>);

impl Plugin for WokerPlugin {
  async fn build_start() {
    // worker plugin won't block the bundler creating until they have to.
     let plugin_options = BundlerToWorkerMap.get().remove().expect("Failed to get worker plugins")
     if pluin_options.len == 1 {
       // single mode
       self.2 = Some(plugin_options)
     } else {
       plugin_options = map(option -> Mutex::new(option))
     }
  }

 async fn transform() {
    for p in self.2 {
      p.transform()
     }
  }
}

fn new_bundler(plugins: Vec<PluginOptionsOrBundlerIdWithWorkerId>) {
  // ...
let plugins: Vec<dyn Plugin>  plugins.map(|item| if item is PluginOptions { JsPlugn::new(item) else { WorkerPlugin(item.0, item1) } }).
}

#[napi]
pub fn registerWorkerPlugin(bundlerId, workerId, plugin_options: Vec<PluginOptions>) {
  BundlerToWorkerMap.get_or_insert_default(bundlerId).get_or_insert_default().insert(WorkerId)
}
```

5. in js's worker code, do
```js
import { registerWokerPlugin } from 'rolldown'

// single mode
registerWokerPlugin(bundlerId, workerId, {
 name: 'signle',
 transform: () => {
  }
})
// mutiple
registerWokerPlugin(bundlerId, workerId, 
[
{
 name: 'signle',
 transform: () => {
  }
},
{
 name: 'signle2',
 transform: () => {
  }
},
])

```

hyf0: How we free these resource if something worng would be a issue. However, I haven't tackle them, but I guess it won't be a big problem.

sapphi-red:
> * find out why `rolldown (parallel, sucrase)` is 200ms slower than `rolldown (single, sucrase)` (100ms is caused by the worker instantiation, but not sure the rest 100ms is caused by)

This seems to be caused by the unfairness of the benchmark. If I run `hyperfine 'node src/foo.js' 'node src/foo.js p'`, I get around 20ms difference that I expect. But if I run `hyperfine 'node src/foo.js l' 'node src/foo.js pl'`, I get 150ms difference for each build. I guess this 130ms difference comes from the fact that code evaluated in workers are parsed/compiled by v8 for each build. Non-worker code is only parsed/compiled at the first build.

<details>
<summary>src/foo.js</summary>

```js
import * as rolldown from 'rolldown'
import { REPO_ROOT, PROJECT_ROOT } from './utils.js'
import parallelBabelPlugin from './parallel-babel-plugin/index.js'
// NOTE: swapped babel with sucrase
import { babelPlugin } from './parallel-babel-plugin/impl.js'
import { builtinModules } from 'node:module'

const isParallel = !!process.argv[2]?.includes('p')
const loop = !!process.argv[2]?.includes('l')

const options = {
  input: [nodePath.join(REPO_ROOT, './tmp/bench/rome/src/entry.ts')],
  external: builtinModules,
  // Need this due rome is not written with `isolatedModules: true`
  shimMissingExports: true,
  plugins: [isParallel ? parallelBabelPlugin() : babelPlugin()],
  resolve: {
    extensions: ['.ts'],
    tsconfigFilename: nodePath.join(
      REPO_ROOT,
      './tmp/bench/rome/src/tsconfig.json',
    ),
  },
}

for (let i = 0; i < (loop ? 10 : 1); i++) {
  const build = await rolldown.rolldown({
    platform: 'node',
    ...options,
  })
  await build.write({
    dir: nodePath.join(PROJECT_ROOT, `./dist/rolldown/foo`),
  })
  await build.destroy()
}
process.exit(0)
```
</details>

crates/rolldown_plugin/src/plugin.rs
sapphi-red: Done in 8dfccdfbc95d3bc57f0b603e547e7bfe344f8c8d 👍

.gitignore
hyf0, line 1: Seem the change it's not used. It would be good if could revert it this one.

Cargo.toml
```diff
@@ -69,7 +69,8 @@ single_match = "allow"
 [workspace.dependencies]
 anyhow                    = "1.0.82"
 ariadne                   = "0.4.0"
-async-trait               = "0.1.80"
+async-channel             = "2.2.0"
+async-trait               = "0.1.79"
```
hyf0: Seems the version is bumping down.

crates/rolldown/src/bundler_builder.rs
hyf0, line 1: Seem the change it's not used. It would be good if could revert it this one.

crates/rolldown_binding/src/options/plugin/js_plugin.rs
```diff
@@ -21,6 +21,10 @@ impl JsPlugin {
   pub(crate) fn new_boxed(inner: BindingPluginOptions) -> Box<dyn Plugin> {
     Box::new(Self { inner })
   }
+
+  pub(super) fn new_raw(inner: BindingPluginOptions) -> Self {
```
hyf0: `boxed` is the special form, se we add `boxed` after `new`. While this one return `Self` directly, using `new` would be just fine.

crates/rolldown_binding/src/parallel_js_plugin_registry.rs
```diff
@@ -0,0 +1,96 @@
...
+    atomic::{self, AtomicU16},
+    Mutex,
+  },
+};
+
+type PluginsInSingleWorker = Vec<BindingPluginWithIndex>;
+type PluginsList = Vec<PluginsInSingleWorker>;
+pub(crate) type PluginValues = FxHashMap<usize, Vec<BindingPluginOptions>>;
+
+static PLUGINS_MAP: Lazy<Mutex<FxHashMap<u16, PluginsList>>> = Lazy::new(Mutex::default);
```
hyf0: You could just use `DashMap`, it has better performance and better dx.

```rust
static PLUGINS_MAP: Lazy<DashMap<u16, PluginsList>> = Lazy::new(DashMap::default);
```


https://docs.rs/dashmap/latest/dashmap/struct.DashMap.html

crates/rolldown_plugin/src/worker_manager.rs
sapphi-red: I think I got what you mean. What would be the upside of doing this?

crates/rolldown_binding/src/parallel_js_plugin_registry.rs
```diff
@@ -0,0 +1,96 @@
...
+static NEXT_ID: AtomicU16 = AtomicU16::new(1);
+
+#[napi(custom_finalize)]
```
hyf0, line 22: Why `custom_finalize`? I thought `ParallelJsPluginRegistry` was just a simple object,

hyf0 (changes requested):
Seem some changes aren't used. It would be good if could revert them.

sapphi-red:
For reference, I got this result with the script above (https://github.com/rolldown/rolldown/pull/825#issuecomment-2047295007) that removes some "unfairness".
```
$ hyperfine 'node src/foo.js' 'node src/foo.js p'

Benchmark 1: node src/foo.js
  Time (mean ± σ):      4.020 s ±  0.036 s    [User: 6.871 s, System: 0.700 s]
  Range (min … max):    3.974 s …  4.100 s    10 runs

Benchmark 2: node src/foo.js p
  Time (mean ± σ):      2.199 s ±  0.020 s    [User: 17.130 s, System: 2.876 s]
  Range (min … max):    2.165 s …  2.226 s    10 runs

Summary
  node src/foo.js p ran
    1.83 ± 0.02 times faster than node src/foo.js
```

hyf0:
Except the above comments, I'm fine with other changes. Since we are exploring it, it's not necessary we make it perfect in any place.

So let's merge the PR once the above comments being solved.

crates/rolldown_binding/src/options/plugin/js_plugin.rs
sapphi-red: I wasn't sure if I can just name it `new`. I was thinking that maybe making non-boxed JsPlugin is discouraged. But it seems it's not the case, will update.

crates/rolldown_plugin/src/worker_manager.rs
hyf0: I'm seeing the new changes and you probably already did what I mean.

crates/rolldown_binding/src/parallel_js_plugin_registry.rs
sapphi-red: If an error is thrown between https://github.com/rolldown/rolldown/blob/568038e3f0e92cd37a6eb0b3f603451bc5ee8f7b/packages/rolldown/src/utils/create-bundler.ts#L20-L22 and https://github.com/rolldown/rolldown/blob/568038e3f0e92cd37a6eb0b3f603451bc5ee8f7b/packages/rolldown/src/utils/create-bundler.ts#L33-L37 , `ParallelJsPluginRegistry` will be dropped when it's GCed on JS side, but the values in `PLUGINS_MAP` won't be dropped. I think we need to manually clear them to avoid memory leak.
https://github.com/rolldown/rolldown/blob/568038e3f0e92cd37a6eb0b3f603451bc5ee8f7b/crates/rolldown_binding/src/parallel_js_plugin_registry.rs#L65-L70

hyf0: I guess that's I want to have a `BundlerId` to link all resources to a bundler and drop everything while something is wrong. However, I also can't be sure how this really works out. Let's use the current implementation for now.

hyf0 (approved):
Nice. This is a big step.

PR #833, opened 2024-04-11
feat(node/plugin): support `renderStart` hook

### LINKED ISSUE #821: Implement `renderStart` hook.
The `renderStart` hook is run at before generate chunks, [here](https://github.com/rolldown/rolldown/blob/main/crates/rolldown/src/bundler.rs#L132) ia s good place to call it.

For plugin implement, you can follow `generate_bundle` hook, add it to [here](https://github.com/rolldown/rolldown/blob/main/crates/rolldown_plugin/src/plugin_driver/output_hooks.rs#L6).

The typing is `(outputOptions: OutputOptions, inputOptions: InputOptions) => void`, you should use js side normalized options.

Also, if you add a test at [here](https://github.com/rolldown/rolldown/tree/main/packages/rolldown/tests/fixtures/plugin) would be nice.

PengBoUESTC:
### Description

implement render start hook : https://github.com/rolldown/rolldown/issues/821

### DIFF
--- a/crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
+++ b/crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
@@ -59,9 +59,7 @@ pub struct BindingPluginOptions {
     Option<MaybeAsyncJsCallback<(String, RenderedChunk), Option<BindingHookRenderChunkOutput>>>,
 
   #[serde(skip_deserializing)]
-  #[napi(
-    ts_type = "(outputOptions: BindingOutputOptions, inputOptions: BindingInputOptions) => void"
-  )]
+  #[napi(ts_type = "() => void")]
   pub render_start: Option<MaybeAsyncJsCallback<(), ()>>,
 
   #[serde(skip_deserializing)]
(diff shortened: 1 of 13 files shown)


crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
```diff
@@ -58,6 +58,12 @@ pub struct BindingPluginOptions {
   pub render_chunk:
     Option<MaybeAsyncJsCallback<(String, RenderedChunk), Option<BindingHookRenderChunkOutput>>>,
 
+  #[serde(skip_deserializing)]
+  #[napi(
+    ts_type = "(outputOptions: BindingOutputOptions, inputOptions: BindingInputOptions) => void"
```
hyf0: Wrong type. 
```suggestion
    ts_type = "() => void"
```

packages/rolldown/src/plugin/bindingify-output-hooks.ts
```diff
@@ -1,8 +1,24 @@
...
+  options: RolldownNormalizedInputOptions,
+  hook?: Plugin['renderStart'],
+): BindingPluginOptions['renderStart'] {
+  if (!hook) {
+    return undefined
+  }
+  const [handler, _optionsIgnoredSofar] = normalizeHook(hook)
+
+  return async () => {
+    // @ts-expect-error
```
hyf0:
```suggestion
```

It's not suggested to write `@ts-expect-error` if there isn't a good reason.

hyf0 (approved):
Thanks

hyf0 (approved):

PR #842, opened 2024-04-12
perf: fast drop LinkStageOutput

underfin:


hyf0:
1. What's the optimization for? Rust or Node performance?

- If this is only for rust performance, we don't need this due to rust users could have their own `fast-drop` and drop our bundler
  - The reason we don't need this is that I think it's not our responsibility to do this optimization, since it could be done in user-land.
  - Second, this create threads implicitly in `drop` method
    - Implicitly here is bad. You couldn't predict creating thread behaviors by looking into the control flow.
    - it also can't be turned off. If users don't want this behavior, they need to either give up rolldown or raise a PR. So it doesn't make scene in the first place, especially user could this optimization on their own.

- If this only for node performance
  - You need a benchmark data to show that this is worth it for such a big change.
  - The changed code should be limited to `rolldown_binding`

2. How this optimization works?

Dropping data to another thread may show advantage on multiple time build but bad for one-time usage such as cli.

--- 

Saw you making this a draft, not worth for me to continue writing this.

underfin:
I'm finding why the ci benchmark is not changed. The js benchmark makes it difficult to see changes at local, but the local rust benchmark will see the result. The pr branch is 180 ms compared to main branch 200 ms for threejs10x.

> If this is only for rust performance, we don't need this due to rust users could have their own fast-drop and drop our bundler

It is a Rust problem because we never drop memory manually, it depends on Rust compiler. The problem is `Module/AsT` will be used at rolldown everywhere, the drop action happened at the final process, but the data is larger, it take 20% times to drop it, and it will block the main thread.

> How this optimization works

The performance is due to has something to do before quit the main thread(drop data),  move the action of drop data to other thread will run it and other something in parallel. [Here](https://news.ycombinator.com/item?id=23362518) has some detail for it.

Boshen:
The technique is legit. But maybe don't call it `fast_drop`, call it `drop_in_another_thread`?

Boshen:
To show evidence, you should use [Mac Xcode Instruments](https://web-infra-dev.github.io/rspack-dev-guide/profiling/intro.html#mac-xcode-instruments) and see if your CPUs are occupied with drops in the end.

crates/rolldown/src/drop_in_another_thread.rs
```diff
@@ -0,0 +1,36 @@
+use std::{mem, thread};
+
+use crate::types::{ast_table::AstTable, module_table::ModuleTable, symbols::Symbols};
+
+fn drop_in_another_thread<T>(src: T)
```
hyf0, line 5: Move this to `rolldown_utils` or `rolldown_binding`.

crates/rolldown/src/drop_in_another_thread.rs
```diff
@@ -0,0 +1,36 @@
...
+
+/// Here using the pattern of impl `Drop` trait to it,
+/// it avoid drop them manually at `scan_stage` or `generate_stage`.
+impl Drop for ModuleTable {
+  fn drop(&mut self) {
+    drop_in_another_thread(std::mem::take(self));
+  }
+}
+
+impl Drop for Symbols {
```
hyf0, line 26: Apply this on `Bundler` in `rolldown_binding`. Revert all changes in `rolldown`.

hyf0 (changes requested):

hyf0:
> Dropping data to another thread may show advantage on multiple time build but bad for one-time usage such as cli.

With that being said, we probably could add a flag to show we are running in `cli` mode and we could call `std::mem:forget` instead of `drop_in_another_thread` on the bundler. Probably would be more faster.

crates/rolldown/src/drop_in_another_thread.rs
hyf0: Thinking back again. Guess it would be ok stay in `rolldown` crate.

crates/rolldown/src/drop_in_another_thread.rs
hyf0: However, I still think should do this explicitly rather do it in `drop` method.

Brooooooklyn:
Using `std::mem::forget` is absolutely a footgun. If you don't perform any memory release operations during a large project build, it will lead to OOM issues more frequently, especially in memory-constrained environments like GitHub Actions and Vercel build. Introducing more complex memory management mechanisms, such as GC algorithms or LRU, would significantly increase the complexity of memory management and the likelihood of bugs.

I also oppose manually spawning memory to another thread for release internally; on some platforms, like wasm32, this approach would be counterproductive.

crates/rolldown/src/drop_in_another_thread.rs
```diff
@@ -0,0 +1,36 @@
...
+  fn drop(&mut self) {
+    drop_in_another_thread(std::mem::take(self));
```
hyf0, line 28: I was so against implicity that I didn't notce this in the first place. Does this cause infinity loop? Every `drop` on `Self` will create a new `Self` and when the new `Self` is dropped another `Self` will be created again.

We end up create infinity threads on infinity `Self`.

hyf0:
Overall, I think it's valueable. rspack use this optimization and has some improment. However as @Brooooooklyn said, it's not a best way in any situation, so what we need here is only apply this optimization in some conditions.

PR #846, opened 2024-04-13
feat(rolldown): introduce rolldown_rayon

Brooooooklyn:
### Description

Introduce `rolldown_rayon` crate to make it easier to write rayon code on the non-rayon supported platforms.


Brooooooklyn:
* **#846** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/846?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @Brooooooklyn and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

hyf0:
hhh. I was writing macros about this. I guess they are the same.

hyf0 (approved):

Brooooooklyn:
macro can always produce better performance because it outputs the optimized codes in the compile stage.
traits don't affect the compile performance if the `lto` is off. And it can produce the same optimized codes during `lto`.
IMO, utilizing traits generally makes our code more readable.

hyf0:
Let's merge it first. I want to do some changes after mering.

1. move code to `rolldown_utils`. We might release rust crates in the futurue. It's kind of not worth to have a single crate to do it.
2. Rename `par_iter` to `maybe_par_iter` to show the they might not run in parallel under the hood. Especially we might introduce some logic to decide if we want to eanble parallel even in `not(wasm)` platform.

Something like swc did in

```rust
  ($name:ident.into_iter().$operator:ident($($rest:expr)*), $threshold:expr) => {
      if $name.len() >= $threshold {
          use rayon::prelude::*;
          $name.into_par_iter().$operator($($rest)*)
      } else {
          $name.into_iter().$operator($($rest)*)
      }
  };
```

hyf0:
Reminds of `oxc_sourcemap` enable rayon by default. Just a notice @Brooooooklyn if these's problem of it, you could disable sourcemap generatetion as workaround.

Issue #206, opened 2023-11-09
Support `output.generatedCode` option


hyf0:
Yes. You are right. We might need to transform it to es5. I'm thinking use https://rollupjs.org/configuration-options/#output-generatedcode to control this behavior. Let's create an issue for discussions and not touch the runtime module currently.

_Originally posted by @hyf0 in https://github.com/rolldown-rs/rolldown/pull/202#discussion_r1387840958_

hyf0:
From my point of view, using https://rollupjs.org/configuration-options/#output-generatedcode to control this behavior is more proper.

hyf0:
Related esbuild's code: https://github.com/evanw/esbuild/blob/ab9b0021f1b0b878cd746eaba07f45d4786ec461/internal/runtime/runtime.go#L18.

7086cmd:
In Rollup, the `output.generatedCode` option is utilized to control the code generation process. However, in Rolldown, we can potentially achieve the similar functionality using the `oxc_transformer` for syntax lowering without the need for additional control over whether `es5` mode is enabled.

> Which language features Rollup can safely use in generated code. This will not transpile any user code but only change the code Rollup uses in wrappers and helpers. You may choose one of several presets:

Occasionally, we require the control of behavior, such as `constBindings`, `objectShorthand`, `reservedNamesAsProps`, `symbols`, while some of these can be performed during syntax lowering.

In my opinion, the `generatedCode` is utilized for controlling the code generated by Rolldown, particularly in this context - runtime code generation. For syntax lowering, I believe it would be more appropriate to use the `target` option in esbuild or a `transformer` field with manual transformer configuration.

TheAlexLichter:
To add additional context, `output.generateCode` is used at the moment to circumvent a [Vue.js bug](https://github.com/vuejs/core/issues/8351), e.g. [implemented in Nuxt.js](https://github.com/nuxt/nuxt/blob/4edb14cf5e78ecd97fa354eaca65b50cc4c57d56/packages/vite/src/server.ts#L97)

IWANABETHATGUY:
> To add additional context, `output.generateCode` is used at the moment to circumvent a [Vue.js bug](https://github.com/vuejs/core/issues/8351), e.g. [implemented in Nuxt.js](https://github.com/nuxt/nuxt/blob/4edb14cf5e78ecd97fa354eaca65b50cc4c57d56/packages/vite/src/server.ts#L97)

Ok, let's add it into backlog

[...]

PR #874, opened 2024-04-16
refactor(rolldown): simplify the runtime module creation

Brooooooklyn:
### Description

Avoid to create an async task for runtime module creation.

### DIFF
--- a/crates/rolldown/src/module_loader/module_loader.rs
+++ b/crates/rolldown/src/module_loader/module_loader.rs
@@ -46,7 +46,7 @@ pub struct ModuleLoader {
   shared_context: Arc<TaskContext>,
   rx: tokio::sync::mpsc::UnboundedReceiver<Msg>,
   visited: FxHashMap<Arc<str>, ModuleId>,
-  runtime_id: Option<NormalModuleId>,
+  runtime_id: NormalModuleId,
   remaining: u32,
   intermediate_normal_modules: IntermediateNormalModules,
   external_modules: ExternalModuleVec,
@@ -74,6 +74,8 @@ impl ModuleLoader {
   ) -> Self {
     let (tx, rx) = tokio::sync::mpsc::unbounded_channel::<Msg>();
 
+    let tx_to_runtime_module = tx.clone();
+
     let common_data = Arc::new(TaskContext {
       input_options: Arc::clone(&input_options),
       tx,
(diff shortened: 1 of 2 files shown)


Brooooooklyn:
* **#874** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/874?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @Brooooooklyn and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

hyf0:
Purpose of original code is we don't want parsing runtime module to block the main thread logic. We could simplify them, but what's the goal here?

---
edited

We also don't think through whether to expose the runtime module via hooks. However, if we gonna do this way, it means we might need execture js code here. If we make it sync, it will just block the whole bundling process.

Brooooooklyn:
> Purpose of original code is we don't want parsing runtime module to block the main thread logic.

The `ModuleLoader::new` is already in a nested async call stack, which means It doesn't necessarily happen in the main thread. Spawning merely moves computation from one tokio worker to another tokio worker, which I believe is an unnecessary overhead.

Regarding the optimization of Tokio scheduling, I think it can wait until there is a large real-world project that can be connected, and then fine-tune through [tokio_console](https://github.com/tokio-rs/console). Before that, let's prioritize simplifying the code complexity and making the code control flow clearer.

hyf0:
> > Purpose of original code is we don't want parsing runtime module to block the main thread logic.
> 
> The `ModuleLoader::new` is already in a nested async call stack, which means It doesn't necessarily happen in the main thread. Spawning merely moves computation from one tokio worker to another tokio worker, which I believe is an unnecessary overhead.
> 
> Regarding the optimization of Tokio scheduling, I think it can wait until there is a large real-world project that can be connected, and then fine-tune through [tokio_console](https://github.com/tokio-rs/console). Before that, let's prioritize simplifying the code complexity and making the code control flow clearer.

Changes you made here https://github.com/rolldown/rolldown/pull/874/files#diff-79e9098941ee0e7d5a23b2dba8b426dd3b85aafcad407c9606f3788c2858d819L56-L64 means that creating and scanning runtime module will definitely run after `resolve_user_defined_entries` while the previous is running in parallel if tokio workers is running in multiple thread mode.

Tokio runtime will do the best effort while it can with the original code. However, with this PR changed, it would never be running in parallel.

Brooooooklyn:
I tested the execution time of create_runtime_module, and it's about **560.917µs** on my computer.
image

I feel it's not worth it to schedule an extra task for this level of computational workload. 

And I found that `RuntimeNormalModuleTaskResult` is almost a constant; maybe we can make it static in the future (by computing in advance during compilation through `build.rs`). Before this, I think it is worthwhile to simplify the logic and program control flow used to create it.

hyf0:
> And I found that `RuntimeNormalModuleTaskResult` is almost a constant; maybe we can make it static in the future

Yeah. Runtime module needs to be generated dynamically depends on user configuration. If user want to generated code targeting es5 and es6 or others.

https://rollupjs.org/configuration-options/#output-generatedcode

~Notice that this option allows to accept a function. As mush as I don't want to support accepting a function, but~ all of that show that the complexity to deal with runtime module is enough to encapsulate it in a single task and do everything in it.

---

The principle here is always to run parallel if possible because we want to as soon as possible to warm up the parsing process of user's code. This part is the key step of bundling, simplicity isn't over performance here. I also know that current execution time of runtime module task is very short, but we still spawned it. Because we know we would add more code here.

We used to parse runtime module in sync and refactor it to make it ran in async due to the above reasons. However, if we already know what would happen most likely, guess it would be good to make the code more satisfy the future

Related issues:
- https://github.com/rolldown/rolldown/issues/206

hyf0:
https://github.com/evanw/esbuild/blob/7a225fff7233a8de275c016e2215a6f9ea356030/internal/bundler/bundler.go#L1283 does it here. I'm not saying what esbuild does is definitely correct. But in this case, it doesn't show much negative, so it could be sign to infer us how the code is gonna like to be in the future.

---edited

I just found that esbuild does this more early. It parses runtime module even before call `build_start` hook.

Brooooooklyn:
@hyf0 made slight modifications to optimize the program flow, keeping the logic for task run spawn.

crates/rolldown/src/module_loader/module_loader.rs
```diff
@@ -82,16 +84,38 @@ impl ModuleLoader {
...
+    let mut intermediate_normal_modules = IntermediateNormalModules::new();
+    let mut symbols = Symbols::default();
+    let runtime_id = intermediate_normal_modules.alloc_module_id(&mut symbols);
+
+    let task = RuntimeNormalModuleTask::new(runtime_id, tx_to_runtime_module);
+    let handle = tokio::runtime::Handle::current();
+
+    #[cfg(target_family = "wasm")]
+    {
+      // could not block_on/spawn the main thread in WASI
```
hyf0, line 96: Changes are fine. But this comment seems a little confused with `could not block_on` and the following `handle.block_on(`.

hyf0 (approved):

PR #879, opened 2024-04-16
chore: remove unnecessary condition checks

zscumt123:
### Description
I found some duplicate conditional statements, and I'm not sure if they can be removed
image


hyf0 (approved):
Thanks

PR #762, opened 2024-04-05
refactor(rust/testing): don't repeat `bundlerOptions` twice

hyf0:
### Description

We used to maintain a copy of `BundlerOptions` for allowing us to writing test config in json. This cause a lot of duplicate works for writing the same options twice.

Now we use `BundlerOptions` directly, it solve the problem.

PR #880, opened 2024-04-16
feat(rust):  only derive `deserialize` on `BundlerOptions` in testing

### LINKED ISSUE #786: Reduce binary sizes by removing derived macros which is unnecessary in production
To support writing test config in json format we add `Deserialize, JsonSchema` on `BundlerOptions`. So they are only for rust testing. Users of rolldown doesn't rely on these features.

https://github.com/rolldown/rolldown/blob/e39ddd845ac5ade81cdf1a2b8e0588c3eaf70f9b/crates/rolldown_common/src/inner_bundler_options/mod.rs#L14-L16

However, these test-only feature are compiled and increased the binary size. We don't want that. So we need hide these features using [feature gate](https://doc.rust-lang.org/cargo/reference/features.html).

## Guidance

See how oxc does this

- https://github.com/oxc-project/oxc/blob/6de1b77b1b9bbace4661ffb59850e8fbe167dd6a/crates/oxc_ast/src/ast/js.rs#L39-L49
- https://github.com/oxc-project/oxc/blob/main/crates/oxc_ast/Cargo.toml

oddguan:
Added a feature gate called "serialize" for enabling converting between JSON and `BundlerOptions` in test envs.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -43,6 +43,5 @@
     "temp": true,
     "tmp": true
   },
-  "editor.wordWrap": "on",
-  "rust-analyzer.cargo.features": "all"
+  "editor.wordWrap": "on"
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -43,5 +43,6 @@
     "temp": true,
     "tmp": true
   },
-  "editor.wordWrap": "on"
+  "editor.wordWrap": "on",
+  "rust-analyzer.cargo.features": "all"
 }
--- a/crates/rolldown_common/Cargo.toml
+++ b/crates/rolldown_common/Cargo.toml
@@ -29,3 +29,6 @@ schemars           = { workspace = true }
 serde              = { workspace = true, features = ["derive"] }
 serde_json         = { workspace = true }
 sugar_path         = { workspace = true }
+
+[features]
+serialize = []
--- a/crates/rolldown_common/Cargo.toml
+++ b/crates/rolldown_common/Cargo.toml
@@ -31,4 +31,4 @@ serde_json         = { workspace = true }
 sugar_path         = { workspace = true }
 
 [features]
-serialize = []
+deserialize_bundler_options = []
--- a/crates/rolldown_common/src/inner_bundler_options/mod.rs
+++ b/crates/rolldown_common/src/inner_bundler_options/mod.rs
@@ -1,8 +1,8 @@
 use std::{fmt::Debug, path::PathBuf};
 
-#[cfg(feature = "serialize")]
+#[cfg(feature = "deserialize_bundler_options")]
 use schemars::JsonSchema;
-#[cfg(feature = "serialize")]
+#[cfg(feature = "deserialize_bundler_options")]
 use serde::{Deserialize, Deserializer};
 
 use self::types::{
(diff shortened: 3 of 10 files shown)


oddguan:
This PR is for issue #786

crates/rolldown_common/src/inner_bundler_options/mod.rs
```diff
@@ -11,14 +13,21 @@ use self::types::{
 
 pub mod types;
 
-#[derive(Default, Debug, Deserialize, JsonSchema)]
-#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[derive(Default, Debug)]
+#[cfg_attr(
+  feature = "serialize",
```
hyf0: please rename to `deserialize_bundler_options`

crates/rolldown_common/Cargo.toml
```diff
@@ -29,3 +29,6 @@ schemars           = { workspace = true }
 serde              = { workspace = true, features = ["derive"] }
 serde_json         = { workspace = true }
 sugar_path         = { workspace = true }
+
+[features]
+serialize = []
```
hyf0: please rename to `deserialize_bundler_options`

.vscode/settings.json
hyf0, line 1: Unexpected changes.

justfile
hyf0, line 1: Unexpected changes.

hyf0 (changes requested):

justfile
oddguan: feature gates are disabled by default. Therefore when running tests, code won't compile because deserialize functionalities are disabled by default. Is there a way to enable it in test envs without modifying justfile?

hyf0: See how we did it to tokio

https://github.com/rolldown/rolldown/blob/93dc6ce52262d65dfc37a63eca44900981d48864/crates/rolldown/Cargo.toml#L35-L42

oddguan: done

hyf0:
You forgot about this one and others:

https://github.com/rolldown/rolldown/blob/07a175bbb30396af762ad7bc6d6009a84c43d96d/crates/rolldown_common/src/inner_bundler_options/types/resolve_options.rs#L6

...

Please make sure all depended types by `BundleOptions` are changed too.

oddguan:
> You forgot about this one and others:
> 
> https://github.com/rolldown/rolldown/blob/07a175bbb30396af762ad7bc6d6009a84c43d96d/crates/rolldown_common/src/inner_bundler_options/types/resolve_options.rs#L6
> 
> ...
> 
> Please make sure all depended types by `BundleOptions` are changed too.

Looked into the original PR #762 and found some leftover files. Not sure if I covered all. Please check thanks

oddguan:
@hyf0 needs approval on workflows

crates/rolldown_common/src/inner_bundler_options/mod.rs
```diff
@@ -29,16 +38,23 @@ pub struct BundlerOptions {
   pub dir: Option<String>,
   pub format: Option<OutputFormat>,
```
hyf0, line 39: https://github.com/rolldown-rs/rolldown/blob/ee3739d65dfed173dbc582270f7cc89017dc285a/crates/rolldown_common/src/inner_bundler_options/types/output_format.rs

Missed this one

hyf0 (changes requested):

crates/rolldown_common/src/inner_bundler_options/mod.rs
oddguan: done

hyf0 (approved):
Thanks!

PR #881, opened 2024-04-16
ci: add cargo bench as extra rust benchmark

### LINKED ISSUE #869: Reliable way for benchmark rolldown in rust on CI
The old way https://github.com/rolldown/rolldown/blob/6f01071b31f5255ce0a0427a3a67c4757f96504d/.github/workflows/bench.yaml is more reliable than current one, we should bring back it and combine it with https://github.com/benchmark-action/github-action-benchmark.

underfin:
### Description

close https://github.com/rolldown/rolldown/issues/869

### DIFF
--- a/.github/workflows/benchmark-rust.yml
+++ b/.github/workflows/benchmark-rust.yml
@@ -23,8 +23,8 @@ concurrency:
   cancel-in-progress: ${{ github.ref_name != 'main' }}
 
 jobs:
-  benchmark:
-    name: Benchmark
+  codspeed-benchmark:
+    name: Codspeed Benchmark
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Branch
@@ -40,18 +40,6 @@ jobs:
           cache-base: main
           cache-target: release
 
-      - name: Install pnpm
-        uses: pnpm/action-setup@v3
-
-      - name: Install node
-        uses: actions/setup-node@v4
-        with:
-          node-version: 20
-          cache: pnpm
-
-      - name: Install dependencies
-        run: pnpm install
-
       - name: Setup benchmark
         run: just setup-bench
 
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark-rust.yml
```diff
@@ -40,18 +40,6 @@ jobs:
...
-        uses: actions/setup-node@v4
-        with:
-          node-version: 20
-          cache: pnpm
-
-      - name: Install dependencies
-        run: pnpm install
-
       - name: Setup benchmark
         run: just setup-bench
```
hyf0: setup-bench requires `zx` installed.

underfin: ok

.github/workflows/benchmark-rust.yml
```diff
@@ -63,3 +63,79 @@ jobs:
...
+      - name: Show/Update benchmark result
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: 'Rust Benchmark'
+          # What benchmark tool the output.txt came from
+          tool: 'cargo'
+          # Where the output from the benchmark tool is stored
+          output-file-path: tmp/new-benchmark-rust-output.txt
+          # Where the previous data file is stored
+          external-data-json-path: tmp/benchmark-rust-output.txt
```
hyf0: I was hoping we are doing this way: https://github.com/rolldown/rolldown/blob/6f01071b31f5255ce0a0427a3a67c4757f96504d/.github/workflows/bench.yaml#L37-L59

From experences of nodejs benchmark CI, using previous data is most likely inaccurate due to they are running in different github machines.

We could just use github-benchmark-action to store data and draw chart.

underfin: > From experences of nodejs benchmark CI, using previous data is most likely inaccurate due to they are running in different github machines.

Yeah. We need to run benchmark at one machine. The store data and draw chart is unnecessary, because it has some noise.

hyf0 (approved):

PR #897, opened 2024-04-17
ci: add `cargo-shear`; run `cargo shear --fix`

### LINKED ISSUE #732: Use `cargo shear` to check if there are unused deps in CI
Does `cargo shear`  support exit with 1 if found any unused deps. Might want to add it in CI.

_Originally posted by @hyf0 in https://github.com/rolldown/rolldown/issues/729#issuecomment-2033746207_

Boshen:
closes #732

### Description

`cargo shear` is performing very well in oxc and Rspack, let's add it to Rolldown as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,12 +83,15 @@ jobs:
       - name: Install Rust
         uses: moonrepo/setup-rust@v1
         with:
-          bins: just, taplo-cli
+          bins: just, taplo-cli, cargo-shear
           cache-base: main
           components: clippy,rustfmt
 
       - name: Lint
         run: just lint-rust
+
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -89,6 +89,12 @@ jobs:
       - name: Lint
         run: just lint-rust
 
+      - name: Install Cargo Shear
+        run: cargo binstall --no-confirm cargo-shear
```
hyf0: Do we need an extra step? Could we reuse the upper `bins: just, taplo-cli` field of `moonrepo/setup-rust@v1`?

hyf0:
Need to resolve conflicts

hyf0 (approved):

PR #900, opened 2024-04-17
feat: add render_error hook

underfin:

### DIFF
--- a/crates/rolldown_binding/src/options/plugin/js_plugin.rs
+++ b/crates/rolldown_binding/src/options/plugin/js_plugin.rs
@@ -106,6 +106,18 @@ impl Plugin for JsPlugin {
     Ok(())
   }
 
+  // --- Output hooks ---
+
+  async fn render_start(
+    &self,
+    _ctx: &rolldown_plugin::SharedPluginContext,
+  ) -> rolldown_plugin::HookNoopReturn {
+    if let Some(cb) = &self.render_start {
+      cb.await_call(()).await?;
+    }
+    Ok(())
+  }
+
   async fn render_chunk(
     &self,
     _ctx: &rolldown_plugin::SharedPluginContext,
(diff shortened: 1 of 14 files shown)


crates/rolldown_binding/src/options/plugin/js_plugin.rs
```diff
@@ -106,6 +106,18 @@ impl Plugin for JsPlugin {
     Ok(())
   }
 
+  // --- Output hooks ---
```
hyf0:
```suggestion
  // --- Generate hooks ---
```

packages/rolldown/src/plugin/index.ts
```diff
@@ -76,6 +76,16 @@ export interface Plugin {
     >
   >
 
+  buildEnd?: Hook<(this: null, err?: string) => MaybePromise<NullValue>>
+  // --- Output hooks ---
```
hyf0:
```suggestion
  // --- Generate hooks ---
```

packages/rolldown/tests/plugin/plugin.test.ts
```diff
@@ -0,0 +1,33 @@
+import { expect, test, vi } from 'vitest'
+import { rolldown, Plugin } from 'rolldown'
+
+async function buildWithPlugin(plugin: Plugin) {
```
hyf0, line 4: I think `fixture` based testing is enough to do the similar things, but let's evolve a little bit and see how this goes.

hyf0 (approved):
Good work!

packages/rolldown/tests/plugin/plugin.test.ts
underfin: Yeah. Here separate it because the `error` hook needs to catch the error, i want to avoid increasing fixture test complex.

PR #902, opened 2024-04-17
fix: buildEnd hook handle error

underfin:


crates/rolldown/src/module_loader/module_loader.rs
```diff
@@ -254,14 +252,12 @@ impl ModuleLoader {
...
         }
         Msg::Panics(err) => {
-          panic_errors.push(err);
+          return Err(err);
         }
       }
       self.remaining -= 1;
     }
 
-    assert!(panic_errors.is_empty(), "Panics occurred during module loading: {panic_errors:?}");
```
underfin, line 263: I removed this using `Result` instead of it to avoid panic at here.

hyf0 (approved):

PR #908, opened 2024-04-18
feat: support basic moduleParsed hook

underfin:

### DIFF
--- a/crates/rolldown/src/module_loader/normal_module_task.rs
+++ b/crates/rolldown/src/module_loader/normal_module_task.rs
@@ -5,8 +5,8 @@ use futures::future::join_all;
 use index_vec::IndexVec;
 use oxc::span::SourceType;
 use rolldown_common::{
-  AstScope, FilePath, ImportRecordId, ModuleType, NormalModuleId, RawImportRecord, ResolvedPath,
-  ResourceId, SymbolRef,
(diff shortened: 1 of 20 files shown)


crates/rolldown/src/module_loader/normal_module_task.rs
```diff
@@ -75,6 +75,15 @@ impl NormalModuleTask {
     let (ast, scope, scan_result, ast_symbol, namespace_symbol) = self.scan(&source);
     tracing::trace!("scan {:?}", self.resolved_path);
 
+    self
+      .ctx
+      .plugin_driver
+      .module_parsed(Arc::new(ModuleInfo {
+        code: Some(Arc::clone(&source)),
+        id: Arc::clone(&self.resolved_path.path),
+      }))
```
underfin, line 84: Here `ModuleInfo` if comes from `NormalModule` is better, but it need to construct `NormalModule` at task, it needs to use `NormalModule` instead of `NormalModuleBuilder` to send main thread.

hyf0: Yeah. I feel you. In short, the problem here is if we want design a struct with a lot of `Option` or No `Option` but with default value. Either way has their advantages and downsides.

hyf0 (approved):

PR #918, opened 2024-04-19
chore(benchmark): use criterion2

Boshen:
I forked criterion to keep things updated:

[criterion] is [passively-maintained](https://github.com/bheisler/criterion.rs/blob/f1ea31a92ff919a455f36b13c9a45fd74559d0fe/Cargo.toml#L63C27-L63C48) with outdated dependencies.

This is fork is updated with:

* renovate bot dependency update
* builtin [codspeed](https://codspeed.io) feature
* `clap` replaced with [`bpaf`](https://github.com/pacak/bpaf) to reduce binary size and compilation time
* merged the `criterion-plot` crate into `criterion2`

As you can see, a lot of stuff is removed from Cargo.lock

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,15 +34,9 @@ checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
 
 [[package]]
 name = "anes"
-version = "0.1.6"
+version = "0.2.0"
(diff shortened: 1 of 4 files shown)


hyf0:
Sounds good. I'm fine as long as it doesn't affect workflows.

PR #924, opened 2024-04-20
refactor: mark Program as covariant and borrow it directly

branchseer:
### Description

`Program` is now covariant over the allocator lifetime because of https://github.com/oxc-project/oxc/pull/2943. We can borrow it directly from `OxcAst` without `with_dependent`.


hyf0 (approved):
Thanks!

Boshen:
Very nice improvement!

PR #931, opened 2024-04-21
chore(rolldown): clean module-export-import stuff

Cecil0o0:
### Description

Do some cleanings.
1. clean unused symbols declared in import specifiers.
2. mark type imports, distinguish them with non-type imports.
3. simplify grammar, change code from `import ...; export ...;` to `export ... from ...;`


hyf0 (approved):
Thanks

PR #950, opened 2024-04-22
fix(node): should call hooks with await

zscumt123:
### Description
When building with Rolldown, add two `buildStart` hook，The plugin execution result is incorrect
```javascript

async function sleep(duration) {
  return new Promise((resolve) => {
    setTimeout(() => {
      //here console  sleep
      console.log('sleep')
      resolve(duration)
    }, duration)
  })
}

export default defineConfig({
  input: './index.js',
  resolve: {
    conditionNames: ['import'],
  },
  plugins: [
    {
      name: 'build_start1',
      async buildStart(options) {
        console.log('1')
      //  here sleep 4000ms
        await sleep(4000)
      },
    },
    {
      name: 'build_start2',
      async buildStart(options) {
        console.log(2)
      },
    },
  ],
})

```
the result , `console.log(sleep)` execute after build end
image

### DIFF
--- a/packages/rolldown/src/plugin/bindingify-build-hooks.ts
+++ b/packages/rolldown/src/plugin/bindingify-build-hooks.ts
@@ -17,7 +17,7 @@ export function bindingifyBuildStart(
   const [handler, _optionsIgnoredSofar] = normalizeHook(hook)
 
   return async (ctx) => {
-    handler.call(ctx, options)
+    await handler.call(ctx, options)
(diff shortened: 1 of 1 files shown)


hyf0 (approved):
Thanks for catching it!

hyf0:
Ci failed.

zscumt123:
image
I don't know how to resolve it , can you help me?

PR #951, opened 2024-04-22
feat: warmup files

sapphi-red:
### Description

Implemented `warmupFiles` (that works like [`server.warmup`](https://vitejs.dev/config/server-options.html#server-warmup) in Vite) to utilize parallelism when processing a single file is slow.

This does reduce ~150ms for the benchmark with parallel js plugins, but it still doesn't utilize the maximum parallelism.

### DIFF
--- a/crates/rolldown/Cargo.toml
+++ b/crates/rolldown/Cargo.toml
@@ -15,6 +15,7 @@ workspace = true
 anyhow             = { workspace = true }
 dunce              = { workspace = true }
 futures            = { workspace = true }
+glob               = { workspace = true }
 index_vec          = { workspace = true }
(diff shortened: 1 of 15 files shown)


hyf0:
Great idea. Glad to see this reducing ~150ms time in the benchmark. However, we don't want to introduce this kind of complexity/optimization at this moment. We plan to have a period that when almost everything is done, and we will focus on performance and try any kind of optimization. We have listed some of them in https://github.com/rolldown/rolldown/discussions/875.

Anyway, this is a great idea, I'll make a note of it.

sapphi-red:
This might only be worth in some cases.

results with rome:

- without
image
- with
image

PR #967, opened 2024-04-24
refactor: compat NormalizedOutputOptions to rollup

underfin:

### DIFF
--- a/crates/rolldown_binding/src/options/binding_output_options/mod.rs
+++ b/crates/rolldown_binding/src/options/binding_output_options/mod.rs
@@ -1,3 +1,5 @@
+use crate::types::js_callback::MaybeAsyncJsCallback;
+
 use super::super::types::binding_rendered_chunk::RenderedChunk;
 use super::plugin::BindingPluginOrParallelJsPluginPlaceholder;
 use derivative::Derivative;
@@ -5,7 +7,7 @@ use napi::threadsafe_function::ThreadsafeFunction;
 use napi_derive::napi;
 use serde::Deserialize;
 
-pub type AddonOutputOption = ThreadsafeFunction<RenderedChunk, Option<String>, false>;
+pub type AddonOutputOption = MaybeAsyncJsCallback<RenderedChunk, Option<String>>;
 
 #[napi(object, object_to_js = false)]
 #[derive(Deserialize, Derivative)]
@@ -22,9 +24,7 @@ pub struct BindingOutputOptions {
   // assetFileNames: string | ((chunkInfo: PreRenderedAsset) => string);
   #[derivative(Debug = "ignore")]
   #[serde(skip_deserializing)]
-  #[napi(
-    ts_type = "Nullable<string> | ((chunk: RenderedChunk) => MaybePromise<VoidNullable<string>>)"
-  )]
+  #[napi(ts_type = "(chunk: RenderedChunk) => MaybePromise<VoidNullable<string>>")]
   pub banner: Option<AddonOutputOption>,
   // chunkFileNames: string | ((chunkInfo: PreRenderedChunk) => string);
   // compact: boolean;
@@ -38,9 +38,7 @@ pub struct BindingOutputOptions {
   // footer: () => string | Promise<string>;
   #[derivative(Debug = "ignore")]
   #[serde(skip_deserializing)]
-  #[napi(
-    ts_type = "Nullable<string> | ((chunk: RenderedChunk) => MaybePromise<VoidNullable<string>>)"
-  )]
+  #[napi(ts_type = "(chunk: RenderedChunk) => MaybePromise<VoidNullable<string>>")]
   pub footer: Option<AddonOutputOption>,
   #[napi(ts_type = "'es' | 'cjs'")]
   pub format: Option<String>,
--- a/crates/rolldown_binding/src/options/binding_output_options/mod.rs
+++ b/crates/rolldown_binding/src/options/binding_output_options/mod.rs
@@ -1,12 +1,11 @@
 use super::super::types::binding_rendered_chunk::RenderedChunk;
 use super::plugin::BindingPluginOrParallelJsPluginPlaceholder;
-use crate::types::js_callback::MaybeAsyncJsCallback;
 use derivative::Derivative;
 use napi::threadsafe_function::ThreadsafeFunction;
 use napi_derive::napi;
 use serde::Deserialize;
 
-pub type AddonOutputOption = MaybeAsyncJsCallback<RenderedChunk, Option<String>>;
+pub type AddonOutputOption = ThreadsafeFunction<RenderedChunk, Option<String>, false>;
 
 #[napi(object, object_to_js = false)]
 #[derive(Deserialize, Derivative)]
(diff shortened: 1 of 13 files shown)


crates/rolldown_binding/src/options/binding_output_options/mod.rs
```diff
@@ -1,12 +1,11 @@
 use super::super::types::binding_rendered_chunk::RenderedChunk;
 use super::plugin::BindingPluginOrParallelJsPluginPlaceholder;
-use crate::types::js_callback::MaybeAsyncJsCallback;
 use derivative::Derivative;
 use napi::threadsafe_function::ThreadsafeFunction;
 use napi_derive::napi;
 use serde::Deserialize;
 
-pub type AddonOutputOption = MaybeAsyncJsCallback<RenderedChunk, Option<String>>;
+pub type AddonOutputOption = ThreadsafeFunction<RenderedChunk, Option<String>, false>;
```
hyf0:
```suggestion
pub type AddonOutputOption = JsCallback<RenderedChunk, Option<String>, false>;
```

hyf0: Use `JsCallback` here. Could you also add the reason that why change it from async to sync. I know there's problem of footer and banner. Would you mind describing how this fix it?

underfin: Here need to be sync at first. Other reason here change it because the result of `MaybeAsyncJsCallback` need to be a `Promise<T>`, if the result is `String` will error at here, The `Either<Either<Promise<Ret>, Ret>` look like is not correct. 
The `JsCallback` need to be implement `await_call` method, it is unnecessary. Actually, I think here `JsCallback` using `ThreadsafeFunction<Args, Ret, false>` is enough, the `Either<Ret, UnknownReturnValue>` is unnecessary.

hyf0: > Here need to be sync at first

If rollup provide sync function in `banner` and `footer`, it's fine we align with it.

> The JsCallback need to be implement await_call method

I don't understand what you are saying here.`JsCallback` is just a type alias. It has nothing to do with `implement await_call method`.

The reason we use type alias is to make devs understand easily these functions/rust types are connected with JavaScript while `threadafe_function` is a little bit difficult to understand in the first look.

To keep the code style and be friendly to contributors, I suggest you to use `JsCallback` here. It has a [full page comment](https://github.com/rolldown/rolldown/blob/a43d3b5f5c3c405404934080124e7ac929e5150a/crates/rolldown_binding/src/types/js_callback.rs#L10-L68) to describe how to understand it for contributors.

> if the result is String will error at here, The Either<Either<Promise<Ret>, Ret> look like is not correct.

If the `ThreadsafeFunction<(),  Either<Either<Promise<String>, String>>` couldn't handle `returning 'hello'` from js side, it might be an unexpected behavior of NAPI-RS.

It's better that you point out this in the PR description and we notify NAPI-RS. Then we fix it by creating an issue that this fix is a workaround, we should revert the change if NAPI-RS fix the behavior. @Brooooooklyn cc

> Actually, I think here JsCallback using ThreadsafeFunction<Args, Ret, false> is enough, the Either<Ret, UnknownReturnValue> is unnecessary.

1. All code in this repo is so far written with purpose. If you think something is unnecessary, please provide a reason.
2. Old NAPI-RS doesn't throw friendly errors like it does now. So this `Either<Ret, UnknownReturnValue>` could help us to quickly find out what's wrong.
3. Even though NAPI-RS now emit the error like ` Failed to convert JavaScript value `Number 1 ` into rust type `String`, still the `Either<Ret, UnknownReturnValue>` here could help us quickly locate the convertion error is happed while returning something from js, not args or struct fields.

hyf0: Since you change the type here and I didn't see the related changes of `BindingOutputOptions`, I guess you forget to change the `ts_type` of `BindingOutputOptions#footer` and `BindingOutputOptions#banner`.

hyf0: Keep the `Either<..., Unkowwn>` . The best error message we can get for debugging unexpected passed js value is

```
Caused by: Error: Rolldown internal error: InvalidArg, Unknown return value. Cannot convert to `Option<String>` in `ThreadsafeFunction<RenderedChunk, Either<Either<Promise<Option<String>>, Option<String>>, UnknownReturnValue>, false>`.
```

This is much friendly that the raw error message that NAPI-RS prodvided.

underfin: Updated.

underfin: > This is much friendly that the raw error message that NAPI-RS provided.

OK. I get it.

> Here need to be sync at first.

I find i had a mistake for it after changed typing. Oirginal is correct.

> The JsCallback need to be implement await_call method

I will descript it at new pr.

packages/rollup-tests/src/ignored-tests.js
```diff
@@ -27,8 +28,10 @@ const ignoreTests = [
   'rollup@function@preserve-symlink: follows symlinks',
   'rollup@function@symlink: follows symlinks',
   // The rolldown output chunk including `module comment` caused line offset, the rollup provider the fake sourcemap can't remapping.
-  // cSpell:disable
   "rollup@sourcemaps@render-chunk-babili: generates valid sourcemap when source could not be determined@generates es",
+  // Here has unexpected error `Error: nul byte found in provided data at position: 0` from rust due to #967.
+  // It crashed at call `banner` function at rust. 
+  "rollup@sourcemaps@excludes-plugin-helpers: excludes plugin helpers from sources@generates es",
```
underfin, line 34: Here the error is wired, cc @Brooooooklyn  could you meet it ?

hyf0: What's the error message here?

hyf0 (approved):

crates/rolldown_binding/src/options/binding_output_options/mod.rs
hyf0: > The JsCallback need to be implement await_call method

I think I get it. Yeah you need to implemement the logic that handle `UnkownReturnValue` as `await_call` did it.

underfin: > Caused by: Error: Rolldown internal error: InvalidArg, Unknown return value. Cannot convert to `core::option::Option<alloc::string::String>`.

The error is also difficult to debugger. I meet it at today, i can't find anywhere crashed. I think we can add fields  `name` to make it easy.

packages/rollup-tests/src/ignored-tests.js
underfin: image

crates/rolldown_binding/src/options/binding_output_options/mod.rs
hyf0: Yeah. It's better than the old NAPI-RS error message, but indeed still confusing. So that's what https://github.com/rolldown/rolldown/pull/970 try to solve.

PR #970, opened 2024-04-24
feat(rust): improve error message of unknown return value of `JsCallback`

hyf0:
### Description

Now we get error message

```
Caused by: Error: Rolldown internal error: InvalidArg, UNKNOWN_RETURN_VALUE. Cannot convert unknown to `Option<String>` in ThreadsafeFunction<RenderedChunk, Either<Either<Promise<Option<String>>, Option<String>>, UnknownReturnValue>, false>.
```

instead of

```
Caused by: Error: Rolldown internal error: InvalidArg, Unknown return value. Cannot convert to core::option::Option<alloc::string::String>.
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1364,6 +1364,7 @@ dependencies = [
  "rolldown_plugin",
  "rolldown_sourcemap",
  "rolldown_tracing",
+ "rolldown_utils",
  "rustc-hash",
  "serde",
  "tracing",
@@ -1490,7 +1491,9 @@ dependencies = [
  "async-scoped",
  "base64",
  "futures",
+ "once_cell",
  "rayon",
+ "regex",
  "xxhash-rust",
 ]
 
--- a/crates/rolldown_binding/Cargo.toml
+++ b/crates/rolldown_binding/Cargo.toml
@@ -29,6 +29,7 @@ rolldown_error     = { workspace = true, features = ["napi"] }
 rolldown_plugin    = { workspace = true }
 rolldown_sourcemap = { workspace = true }
 rolldown_tracing   = { workspace = true }
+rolldown_utils     = { workspace = true }
 rustc-hash         = { workspace = true }
 serde              = { workspace = true }
 tracing            = { workspace = true }
--- a/crates/rolldown_binding/src/types/js_callback.rs
+++ b/crates/rolldown_binding/src/types/js_callback.rs
@@ -1,11 +1,10 @@
-use std::any::type_name;
-
 use futures::Future;
 use napi::{
   bindgen_prelude::{FromNapiValue, Promise},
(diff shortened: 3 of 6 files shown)


hyf0:
* **#970** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/970?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

underfin:
The `ThreadsafeFunction<RenderedChunk, Either<Either<Promise<Option<String>>, Option<String>>, UnknownReturnValue>, false>` is better than previous one.  But it also has some complex, I'm think add a unique `id` to make it easy, like `JSCallback<'id', args, ret>`

hyf0:
> The `ThreadsafeFunction<RenderedChunk, Either<Either<Promise<Option<String>>, Option<String>>, UnknownReturnValue>, false>` is better than previous one. But it also has some complex, I'm think add a unique `id` to make it easy.

New-type `struct Footer(...)` could does the trick, but using new-type cause other problems due new-type itself. Rust doesn't have type constructor, so you need to define new-type manually for every js callback. we could introduce `js_callback!(Footer, ...)` to simplify the process. But I'm not really think through if it's a good idea.

So while the current solution is still complex, it's usable enough. We could gradually improve it.

hyf0:
Another solution we could emit more detailed error message by improving `pretty_type_name`

```
ThreadsafeFunction:
- args: RenderedChunk,
- return: Promise<Option<String>> | Option<String> | UnknownReturnValue
```
.

But still, it doesn't convince me that it's a solution that is better enough to be the final one.


---edited

Might be

```
ThreadsafeFunction: (RenderedChunk) -> Promise<Option<String>> | Option<String> | UnknownReturnValue
```

It's indeed more readable, but still it doesn't convince me that it's a solution that is better enough to be the final one.

hyf0:
> I'm think add a unique `id` to make it easy, like `JSCallback<'id', args, ret>`

Too bad it's impossible due to `JsCallback` is only a type alias. Though we could make `JsCallback` a new-type and ignore the following problems:
- New-type `JsCallback` isn't recognized by `#[napi]`, so we need to deal with this first

The problem here is that rust currently doesn't support `&'static str` as the const generic. So `JsCallback<"Footer", ...>` is not impossible. The alternative solution is using meaningless struct like
```
struct JsCallback<T, ...> {
  _label: std::marker::PhantomData<T>,
}
struct Footer;
footer: JsCallback<Footer, ...>
```

But you should be able to see how this bad dx this could be.

---

We could also try to store the label string in the new-type `JsCallback`, but `type_name` doesn't reconize this, which requires us to implement another mechanism like `type_name`. I doubt it's worth it.

PR #973, opened 2024-04-25
feat: add ModuleInfo some fileds

underfin:

### DIFF
--- a/crates/rolldown/src/module_loader/module_loader.rs
+++ b/crates/rolldown/src/module_loader/module_loader.rs
@@ -1,7 +1,7 @@
 use index_vec::IndexVec;
 use rolldown_common::{
   EntryPoint, EntryPointKind, ExternalModule, ExternalModuleVec, ImportKind, ImportRecordId,
-  ModuleId, ModuleTable, NormalModule, NormalModuleId,
+  ImporterRecord, ModuleId, ModuleTable, NormalModule, NormalModuleId,
 };
 use rolldown_error::BuildError;
 use rolldown_fs::OsFileSystem;
@@ -120,19 +120,19 @@ impl ModuleLoader {
   fn try_spawn_new_task(
     &mut self,
     info: &ResolvedRequestInfo,
-    importer: Option<(NormalModuleId, ImportKind)>,
+    importer: Option<ImporterRecord>,
   ) -> ModuleId {
     match self.visited.entry(Arc::<str>::clone(&info.path.path)) {
       std::collections::hash_map::Entry::Occupied(visited) => {
         let id = *visited.get();
         // If dependency already created, here need to update the `importers/dynamic_importers`
-        if let Some((importer, kind)) = importer {
+        if let Some(importer) = importer {
           if let ModuleId::Normal(id) = id {
             if let Some(m) = self.intermediate_normal_modules.modules[id].as_mut() {
-              if kind.is_static() {
-                m.importers.push(importer);
+              if importer.kind.is_static() {
+                m.importers.push(importer.importer_path);
               } else {
-                m.dynamic_importers.push(importer);
+                m.dynamic_importers.push(importer.importer_path);
               }
             }
           }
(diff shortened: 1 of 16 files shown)


underfin:
* **#973** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/973?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#972** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/972?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/rolldown/src/module_loader/runtime_normal_module_task.rs
```diff
@@ -80,6 +80,9 @@ impl RuntimeNormalModuleTask {
       import_records: IndexVec::default(),
       is_included: false,
       sourcemap_chain: vec![],
+      // The internal runtime module `importers/dynamic_importers` should be skip.
+      importers: vec![],
```
hyf0, line 82: Would it be better if we collect these datas in `module_loader` instead of worker? Since we are running in parallel, these's big chance that those values is not stable cross builds.

hyf0:
It would be good to describe what's the purpose of this PR.  If I understand correctly, adding `importers` and `dynamic_importers` is just for rollup compatibility. No features or internal code of rolldown rely on these data.

I wonder what requirements of user-land need these data. I'm fine with adding these data for improving rollup compatibility. However, if there are no strong requirements on this and it show some downsides to the performance or other features, there's a highly chance these data would not be maintained by rolldown in the future.

underfin:
> It would be good to describe what's the purpose of this PR. If I understand correctly, adding `importers` and `dynamic_importers` is just for rollup compatibility. No features or internal code of rolldown rely on these data.
> 
> I wonder what requirements of user-land need these data. I'm fine with adding these data for improving rollup compatibility. However, if there are no strong requirements on this and it show some downsides to the performance or other features, there's a highly chance these data would not be maintained by rolldown in the future.

See here https://github.com/vitejs/vite/blob/main/packages/vite/src/node/plugins/reporter.ts#L121-L133, https://github.com/withastro/astro/blob/22d00865cf84138facb3aa1f1f5d0252ece1caa1/packages/astro/src/core/build/graph.ts#L67

crates/rolldown/src/module_loader/runtime_normal_module_task.rs
underfin: Move it to `module_loader` couldn't make it stable, i add `sort` at getter.

crates/rolldown/src/module_loader/module_loader.rs
```diff
@@ -118,9 +117,28 @@ impl ModuleLoader {
...
+  fn try_spawn_new_task(
+    &mut self,
+    info: &ResolvedRequestInfo,
+    importer: Option<ImporterRecord>,
+  ) -> ModuleId {
     match self.visited.entry(Arc::<str>::clone(&info.path.path)) {
-      std::collections::hash_map::Entry::Occupied(visited) => *visited.get(),
+      std::collections::hash_map::Entry::Occupied(visited) => {
+        let id = *visited.get();
+        // If dependency already created, here need to update the `importers/dynamic_importers`
```
hyf0: What if the dependency is not created. Are we gonna missing some importers?

crates/rolldown/src/module_loader/module_loader.rs
```diff
@@ -220,7 +239,13 @@ impl ModuleLoader {
             .into_iter()
             .zip(resolved_deps)
             .map(|(raw_rec, info)| {
-              let id = self.try_spawn_new_task(&info);
+              let id = self.try_spawn_new_task(
+                &info,
+                Some(ImporterRecord {
+                  kind: raw_rec.kind,
+                  importer_path: module.resource_id.expect_file().clone(),
```
hyf0: A module could have multiple importers. Why we do we want to use this one?

crates/rolldown/src/module_loader/module_loader.rs
underfin: No. Here has two case for `importer`.
1. Creating module with init `importer` at `normal_module_task`.
2. If the module already crated, here only need to update them.

crates/rolldown/src/module_loader/module_loader.rs
underfin: See above.

crates/rolldown/src/module_loader/module_loader.rs
hyf0: Assume the `init importer at normal_module_task` is still running and there are other depedencies that imports that running module. So, the module is not created currently. How we update it?

underfin: I get it, here because the `dynamic_importers/importers` is invalid at `module_parsed` hook, it can be move to `module_loader`.

hyf0 (approved):

PR #982, opened 2024-04-26
chore(rust/tracing): use async style by default and support `chrome-json-threaded`

sapphi-red:
### Description

If you set `RD_LOG_OUTPUT_STYLE=async` and add `#[tracing::instrument(level = "trace", skip_all)]` to https://github.com/rolldown/rolldown/blob/84dc41968d1952ce329478bfb382d5ea8e26df11/crates/rolldown_binding/src/options/plugin/js_plugin.rs#L84, you'll get the following result on perfetto.dev.

With single thread babel:
image

With parallel babel:
image

### DIFF
--- a/crates/rolldown_tracing/src/lib.rs
+++ b/crates/rolldown_tracing/src/lib.rs
@@ -16,7 +16,6 @@ use tracing_subscriber::EnvFilter;
 
 static LOG_ENV_NAME: &str = "RD_LOG";
 static LOG_OUTPUT_ENV_NAME: &str = "RD_LOG_OUTPUT";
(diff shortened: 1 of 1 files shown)


hyf0 (dismissed):

hyf0:
async style is more clearly than threaed, I make it as the default style. In the meantime, I don't want to introduce another env variable. So, I reused the `RD_LOG_OUTPUT` and devs could use `chrome-json` or `chrome-json-threaded` values to choose the trace style.

sapphi-red:
@hyf0 I guess you forgot to approve?

hyf0 (approved):

hyf0:
> @hyf0 I guess you forgot to approve?

Haha thanks.

Issue #938, opened 2024-04-22
Migrate all esbuild's tests


hyf0:
- ~bundler_css_test~
- [x] #989
- [x] bundler_default_test
- [x] bundler_glob_test
- [x] bundler_importstar_test
- [x] bundler_importstar_ts_test
- [x] bundler_loader_test
- [x] bundler_lower_test
- [x] bundler_packagejson_test #783 
- [x] bundler_splitting_test
- ~bundler_test~
- [x] bundler_ts_test

---

We need to try to run all the tests, and keep a diff of passed and failed in a snapshot. We need to make sure nothing hangs or crashes.

hyf0:
`bundler_css_test` is not needed so far.

Tests related to typescript could be migrated but not passed due to we don't have builtn typescript transpile support. Others should be fine.

chiehw:
I'm confused about two place: 
1. why do some test folders start with dot?
2. why do some test folders not have artifacts.snap file?

For example: 
https://github.com/rolldown/rolldown/pull/984/files
![image](https://github.com/rolldown/rolldown/assets/99897242/f21066c9-3ebc-4932-a6cb-fd6f6e0fd855)

hyf0:
@chiehw  https://rustdoc.swc.rs/testing_macros/attr.fixture.html#ignoring-a-test

chiehw:
ok, thanks. Why ignore these tests? a little curious. It seems that some test can be passed.

hyf0:
> ok, thanks. Why ignore these tests? a little curious. It seems that some test can be passed.

Mainly due to there aren't passed. If you find passed tests that are ignored, it's welcome to open a PR to enable them.

zhusjfaker:
https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_loader_test.go

i will work it @hyf0  @IWANABETHATGUY

hyf0:
> https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_loader_test.go
> 
> i will work it @hyf0 @IWANABETHATGUY

Thanks. The loader feature is complete yet. But you could start migrating these cases.

IWANABETHATGUY:
> https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_loader_test.go
> 
> i will work it @hyf0 @IWANABETHATGUY

[...]

PR #984, opened 2024-04-27
test(esbuild/glob): migrate tests cases of glob

PerfectPan:
### Description

part of https://github.com/rolldown/rolldown/issues/938

- move the ts related test into tests.txt(or should just ignored?)
- not quite sure what should be defined in _config.json for https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_glob_test.go#L32

### DIFF
--- a/crates/rolldown/tests/esbuild/glob/tests.txt
+++ b/crates/rolldown/tests/esbuild/glob/tests.txt
@@ -1,0 +1,2 @@
+TestTSGlobBasicNoSplitting
+TestTSGlobBasicSplitting
--- a/crates/rolldown/tests/esbuild/glob/.basic_no_bundle/_config.json
+++ b/crates/rolldown/tests/esbuild/glob/.basic_no_bundle/_config.json
@@ -1,0 +1,13 @@
+{
+  "_comment": "Input of this test are invalid, so the compiling should works but the execution should fail",
+  "expectExecuted": false,
+  "config": {
+    "input": [
+      {
+        "name": "entry_js",
+        "import": "entry.js"
+      }
+    ],
+    "format": "Cjs"
+  }
+}
--- a/crates/rolldown/tests/esbuild/glob/.basic_no_bundle/entry.js
+++ b/crates/rolldown/tests/esbuild/glob/.basic_no_bundle/entry.js
@@ -1,0 +1,11 @@
+const ab = Math.random() < 0.5 ? 'a.js' : 'b.js'
+console.log({
+  concat: {
+    require: require('./src/' + ab),
+    import: import('./src/' + ab),
+  },
+  template: {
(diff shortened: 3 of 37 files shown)


hyf0:
@IWANABETHATGUY cc

crates/rolldown/tests/esbuild/glob/tests.txt
```diff
@@ -0,0 +1,2 @@
+TestTSGlobBasicNoSplitting
```
IWANABETHATGUY, line 1: Can you explain what is this file used for?

hyf0: It's used to record what tests aren't migrated. I guess it's following what I did in https://github.com/rolldown/rolldown/pull/782.

PerfectPan: Yes, I followed this pr https://github.com/rolldown/rolldown/pull/782 to ensure the missing test cases would be added in the future since I saw rolldown currently should not consider ts related test cases.

IWANABETHATGUY: I noticed that we used an `ignored-text.md` to record this info, but used a `tests.txt` in `glob` tests, can we unify them?

hyf0: I didn't know file. I just take a look, it's created long time ago. I'm fine with unifying them. @IWANABETHATGUY Feel free to decide how to handle `ignored-test.md` and `tests.md`.

---edited

But some notices:

- Ignored tests could be misunderstanding. Currently we ignore some fixtures by adding `.` in the folder name.
- `ignored-test.md` doesn't mean the same thing that said above

IWANABETHATGUY: Looks like the file is a little redundant? Since the `.` prefix is already means it is a ignored tests

IWANABETHATGUY:
Overall, LGTM, could you please resolve the failed test cases, and let the CI pass?

crates/rolldown/tests/esbuild/glob/tests.txt
hyf0: Nah. tests in `tests.md` means tests we have not copied/migrated from esbuild.

hyf0: It's a little bit chaos since now we have `tests.md`, `ignored-tests.txt` and adding `.`. That's why I support unifying them.

IWANABETHATGUY: got it

IWANABETHATGUY (approved):

PR #985, opened 2024-04-27
fix: exit process when build finishes

sapphi-red:
### Description

The cli didn't exit process immediately after the build finished. This PR fixes that.

But I'm not sure why it did exit the process after some seconds even if the thread safe functions are not "weak".

We can add a test by using [`why-is-node-running`](https://www.npmjs.com/package/why-is-node-running). I didn't add it though.

### DIFF
--- a/crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
+++ b/crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
@@ -92,7 +92,7 @@ impl<const WEAK: bool> Debug for BindingPluginOptions<WEAK> {
 }
 
 impl FromNapiValue for BindingPluginOptions<true> {
-  #[allow(deprecated)] // cb.unref is deprecated, fix napi_derive to support const generics with
+  #[allow(deprecated)] // cb.unref is deprecated, fix napi_derive to support const generics
   unsafe fn from_napi_value(
     env: napi::sys::napi_env,
     napi_val: napi::sys::napi_value,
--- a/crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
+++ b/crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
@@ -28,7 +28,7 @@ pub struct BindingPluginOptions<const WEAK: bool = false> {
 
   #[serde(skip_deserializing)]
   #[napi(ts_type = "(ctx: BindingPluginContext) => MaybePromise<VoidNullable>")]
-  pub build_start: Option<MaybeAsyncJsCallback<BindingPluginContext, (), false>>,
+  pub build_start: Option<MaybeAsyncJsCallback<BindingPluginContext, (), WEAK>>,
 
   #[serde(skip_deserializing)]
   #[napi(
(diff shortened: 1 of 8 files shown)


crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
```diff
@@ -84,12 +85,72 @@ pub struct BindingPluginOptions {
...
 }
 
-impl Debug for BindingPluginOptions {
+impl<const WEAK: bool> Debug for BindingPluginOptions<WEAK> {
   fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
     f.debug_struct("BindingPluginOptions").field("name", &self.name).finish_non_exhaustive()
   }
 }
 
+impl FromNapiValue for BindingPluginOptions<true> {
```
sapphi-red: napi_derive doesn't allow us to write it like:
```rust
#[napi_derive::napi(object, object_to_js = false)]
#[derive(Deserialize, Default)]
#[serde(rename_all = "camelCase")]
pub struct BindingPluginOptions<const WEAK: bool = false> {
  #[serde(skip_deserializing)]
  #[napi(ts_type = "(ctx: BindingPluginContext) => MaybePromise<VoidNullable>")]
  pub build_start: Option<MaybeAsyncJsCallback<BindingPluginContext, (), WEAK>>,
}
```
Is there a way to make this work in napi_derive?

crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
```diff
@@ -84,12 +85,72 @@ pub struct BindingPluginOptions {
...
+  #[allow(deprecated)] // cb.unref is deprecated, fix napi_derive to support const generics with
```
sapphi-red:
```suggestion
  #[allow(deprecated)] // cb.unref is deprecated, fix napi_derive to support const generics
```

crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
hyf0: @Brooooooklyn Does threadsafe_function `unref` by default?

hyf0: Do we still need customization if we already make these option with `WEAK: true` like you above code?

Brooooooklyn: Yes if `WEAK` is `true`

Brooooooklyn: > Is there a way to make this work in napi_derive?

hmmm,  I think I can support the const generic in `napi_derive`

crates/rolldown_binding/src/types/js_callback.rs
```diff
@@ -65,18 +65,20 @@ use rolldown_utils::debug::pretty_type_name;
 /// - Rust(Simplified): `MaybeAsyncJsCallback<(Option<String>, i32), Option<i32>>`
 /// - Js: `(a: string | null | undefined, b: number) => Promise<number | null | undefined | void> | number | null | undefined | void`
 /// - Js(Simplified): `(a: Nullable<string>, b: number) => MaybePromise<VoidNullable<number>>`
-pub type JsCallback<Args, Ret> = ThreadsafeFunction<Args, Either<Ret, UnknownReturnValue>, false>;
+pub type JsCallback<Args, Ret, const WEAK: bool = false> =
```
hyf0:
```suggestion
pub type JsCallback<Args, Ret, const WEAK: bool = true> =
```

I think we could make it as the default.

crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
```diff
@@ -84,12 +85,72 @@ pub struct BindingPluginOptions {
...
+  #[allow(deprecated)] // cb.unref is deprecated, fix napi_derive to support const generics
+  unsafe fn from_napi_value(
+    env: napi::sys::napi_env,
+    napi_val: napi::sys::napi_value,
+  ) -> napi::Result<Self> {
+    let mut false_options = BindingPluginOptions::<false>::from_napi_value(env, napi_val)?;
+    let env = &napi::Env::from_raw(env);
+
+    if let Some(cb) = &mut false_options.build_start {
+      cb.unref(env)?;
```
Brooooooklyn: no need if `WEAK` is `true` https://github.com/napi-rs/napi-rs/blob/main/crates/napi/src/threadsafe_function.rs#L296

crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
sapphi-red: > Do we still need customization if we already make these option with `WEAK: true` like you above code?

Yes, because we need `BindingPluginOptions<true>` for main thread plugins and `BindingPluginOptions<false>` for plugins in worker threads.

If `pub struct BindingPluginOptions<const WEAK: bool = false>` is flipped to `pub struct BindingPluginOptions<const WEAK: bool = true>`, then we need `impl FromNapiValue for BindingPluginOptions<false>` and call `cb.refer(env)` instead.

hyf0: Nah. It's ok we mark main thread plugins `WEAK: true`.

hyf0:
So let's make `WEAK: true` by default in `JsCallback` and remove the customization, since they are all `WEAK: true` now.

sapphi-red:
> If `pub struct BindingPluginOptions<const WEAK: bool = false>` is flipped to `pub struct BindingPluginOptions<const WEAK: bool = true>`, then we need `impl FromNapiValue for BindingPluginOptions<false>` and call `cb.refer(env)` instead.

Actually, this wasn't correct. If I call a WEAK=true threadsafe-function in worker threads, it throws "Rolldown internal error: Closing" error, even if I call `cb.refer(env)`.

sapphi-red:
I guess this happens because [`WEAK=true` calls `napi_unref_threadsafe_function`](https://github.com/napi-rs/napi-rs/blob/a54b759bb1767d0b45968ea4acf795de78af268a/crates/napi/src/threadsafe_function.rs#L295-L301) but the docs says it doesn't allow to call it from non-main thread.
https://nodejs.org/api/n-api.html#napi_unref_threadsafe_function:~:text=is%20also%20idempotent.-,This%20API%20may%20only%20be%20called%20from%20the%20main%20thread.,-Miscellaneous%20utilities

sapphi-red:
This is the code I wanted to write but reverted before I made this PR due to compile errors: https://github.com/rolldown/rolldown/pull/985/commits/76f912b6db1b65a46708b1f3fa3e8c139a558527

hyf0:
I looked into it. The current workaround would be use different struct for main and worker threads. Making NAPI-RS supporting generic could reduce the maintain overhead. @Brooooooklyn

hyf0:
let's on hold this pr before #1003.

PR #991, opened 2024-04-28
chore: esbuild dce tests

### LINKED ISSUE #989: bundler_dce_test

IWANABETHATGUY:
### Description
Closed https://github.com/rolldown/rolldown/issues/989


hyf0 (approved):

.gitignore
```diff
@@ -160,4 +160,8 @@ crates/rolldown_binding/artifacts/
 packages/**/THIRD-PARTY-LICENSE
 
 # Opinionated configuration
-.vscode/launch.json
\ No newline at end of file
+.vscode/launch.json
```
hyf0: Could you delete this one?

hyf0 (changes requested):

.gitignore
```diff
@@ -160,4 +160,8 @@ crates/rolldown_binding/artifacts/
...
+
+# node_modules in tests
+!/crates/rolldown/tests/**/node_modules
```
hyf0, line 164: I'm not sure. You might want to limit the scope of this glob. We might rely on `node_modules` that created by `npm`, we just don't have the case currently.

hyf0: In the future, we probably will ensure the correctness of mainstream libraries, such as vue, react and antd etc. It's better we install them by `pnpm` instead of having a copy in rolldown's repo.

IWANABETHATGUY: what about 
`!/crates/rolldown/tests/**/esbuild/node_modules`, limit the scope under **esbuild** tests.

.gitignore
IWANABETHATGUY: image, 
Ok,but it is not added by me.

.gitignore
hyf0: Yeah. That should work.

hyf0 (approved):

Issue #122, opened 2023-10-29
Ensure chunk execution order are correct


hyf0:

hyf0:
It's interesting that I found esbuild only ensure chunk order is deterministic, the order is calculated by some deterministic keys rather execution order.

Rollup calculated the imported chunk order per module rather than follow the global module execution order calculated by `sortModules`.

Example

[esbuild](https://hyrious.me/esbuild-repl/?version=0.21.2&b=e%00a.js%00import+%22.%2F1.js%22%0Aimport+%22.%2F2.js%22%0Aconsole.log%28%27a%27%29&b=e%00b.js%00import+%22.%2F1.js%22%0Aimport+%22.%2F2.js%22%0Aconsole.log%28%27a%27%29&b=e%00c.js%00import+%22.%2F1.js%22%0Aconsole.log%28%27c%27%29&b=%001.js%00console.log%281%29&b=%002.js%00console.log%282%29&o=--bundle+--format%3Desm+--splitting+--outdir%3D.+--tree-shaking%3Dfalse)

![image](https://github.com/rolldown/rolldown/assets/49502170/2f6d1594-ae58-4a4b-80bc-42654f428e79)

[...]

PR #1002, opened 2024-04-29
feat: support OutputChunk#dynamicImports and RenderedChunk#dynamicImp…

underfin:
…orts


hyf0 (approved):

crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
```diff
@@ -234,47 +236,29 @@ impl<'a> GenerateStage<'a> {
...
-            .map(|id| {
-              self.link_output.module_table.normal_modules[*id].resource_id.expect_file().as_str()
-            })
-            .collect::<Vec<_>>();
-          resource_ids.sort_unstable();
-          resource_ids
-        });
-        cross_chunk_imports
-      })
-      .collect::<Vec<_>>();
```
underfin: The `cross_chunk_dynamic_imports` is collected by chunk modules `import_records`, the chunk modules is sorted, so the order is stable.
The `cross_chunk_imports` is from `chunk_meta_imports_vec`, it is collected by chunk modules `referenced_symbols`, the order is also stable.

hyf0: Didn't notice this part. Could you keep the original sorting logic. It have better stability on calculating chunk hash. https://github.com/rolldown/rolldown/pull/1002/files#r1583073224

---

Just notice you change it from hashset to vec.

hyf0 (changes requested):

crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
```diff
@@ -9,14 +9,14 @@ use rolldown_common::{
...
-use rolldown_utils::rayon::{IntoParallelIterator, ParallelBridge, ParallelIterator};
+use rolldown_utils::rayon::{ParallelBridge, ParallelIterator};
 use rustc_hash::{FxHashMap, FxHashSet};
 
 type ChunkMetaImports = IndexVec<ChunkId, FxHashSet<SymbolRef>>;
 type ChunkMetaImportsForExternalModules =
   IndexVec<ChunkId, FxHashMap<ExternalModuleId, Vec<NamedImport>>>;
 type ChunkMetaExports = IndexVec<ChunkId, FxHashSet<SymbolRef>>;
-type IndexCrossChunkImports = IndexVec<ChunkId, FxHashSet<ChunkId>>;
+type IndexCrossChunkImports = IndexVec<ChunkId, Vec<ChunkId>>;
```
hyf0: How `Vec<ChunkId>` avoid duplicate imports to the same chunk?

underfin: Yeah. Here should be `IndexSet`. The `chunk_meta_imports` is also has same problem.

crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
```diff
@@ -234,47 +236,29 @@ impl<'a> GenerateStage<'a> {
...
         ),
       )
       .par_bridge()
       .for_each(
         |(
           chunk,
-          (imports_from_other_chunks, (imports_from_external_modules, cross_chunk_imports)),
+          (
+            imports_from_other_chunks,
+            (imports_from_external_modules, (cross_chunk_imports, cross_chunk_dynamic_imports)),
```
hyf0, line 281: Could you keep the `index` prefix above. The above `cross_chunk_imports: IndexVec<ChunkId, Vec<ChunkId>>` with this `cross_chunk_imports: Vec<ChunkId>>` could be really confusing,

crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
hyf0: I don't think so. If the order relies on the order of traversing symbols, we might break the order due to adjust the generated runtime symbols by ourselves.

Depending the order calculated by `sorting with modules` is much reliable than this.

underfin: Ok. Let me revert it.

crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
underfin: Updated

underfin:
Here i revert the `index_sorted_cross_chunk_imports` back, but the order  is not correct, it seems to be the  execute order of chunk first module .

crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
hyf0: Saw you were saying revert it back in https://github.com/rolldown/rolldown/pull/1002#issuecomment-2082796466. Did you mean rollup ensure the order of `imports` is the execution order of chunk?

underfin: Yeah. The rollup using js `Set/Map`, the Set and Map have the  insertion order, the insertion order is stabled.

hyf0: >  The rollup using js Set/Map, the Set and Map have the insertion order, the insertion order is stabled.

I don't get it. How this is the reason you are using `Vec<ChunkId>` now?

crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
```diff
@@ -1,22 +1,26 @@
...
+use rolldown_utils::rayon::IntoParallelIterator;
+use rolldown_utils::rayon::{ParallelBridge, ParallelIterator};
+use rustc_hash::{FxHashMap, FxHashSet, FxHasher};
 
 type ChunkMetaImports = IndexVec<ChunkId, FxHashSet<SymbolRef>>;
 type ChunkMetaImportsForExternalModules =
   IndexVec<ChunkId, FxHashMap<ExternalModuleId, Vec<NamedImport>>>;
 type ChunkMetaExports = IndexVec<ChunkId, FxHashSet<SymbolRef>>;
-type IndexCrossChunkImports = IndexVec<ChunkId, FxHashSet<ChunkId>>;
+type IndexCrossChunkImports = IndexVec<ChunkId, Vec<ChunkId>>;
```
hyf0: Let me try to make things clear again:

Do not change `IndexCrossChunkImports` from `IndexVec<ChunkId, FxHashSet<ChunkId>>` to `IndexVec<ChunkId, Vec<ChunkId>>`.

- `Vec<ChunkId>` may contains duplicated `ChunkId`
- Don't use `IndexSet` on `IndexCrossChunkImports`. https://github.com/rolldown/rolldown/pull/1002/files#r1583073224

hyf0 (changes requested):

crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
underfin: I missing to revert it.

crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
underfin: I missing revert it.

underfin:
I revert it back, but the code need to rewrite with stable order.
Look like change `FxHashSet` to `IndexSet` at `type ChunkMetaImports = IndexVec<ChunkId, FxHashSet<SymbolRef>>;` `type IndexCrossChunkImports = IndexVec<ChunkId, FxHashSet<ChunkId>>;` is expected. 
The `imports_from_other_chunks_vec` is also need to using `IndexMap` instead of `HashMap`

hyf0:
> I revert it back, but the code need to rewrite with stable order.

No need unless we relies on this kind of [order](https://github.com/rolldown/rolldown/pull/1002#issuecomment-2082796466) and we don't.

Relying on execution order will also make hash changed frequently than using order from `sorting module resourcesId`.

hyf0 (approved):

underfin:
> > I revert it back, but the code need to rewrite with stable order.
> 
> No need unless we relies on this kind of [order](https://github.com/rolldown/rolldown/pull/1002#issuecomment-2082796466) and we don't.
> 
> Relying on execution order will also make hash changed frequently than using order from `sorting module resourcesId`.

It will cause the chunk order is not correct. let me give a example for you later.

hyf0:
> > > I revert it back, but the code need to rewrite with stable order.
> > 
> > 
> > No need unless we relies on this kind of [order](https://github.com/rolldown/rolldown/pull/1002#issuecomment-2082796466) and we don't.
> > Relying on execution order will also make hash changed frequently than using order from `sorting module resourcesId`.
> 
> It will cause the chunk order is not correct. let me give a example for you later.

It's known by me and that's why https://github.com/rolldown/rolldown/issues/122 remains open. I'm still solving it with considering the stability of hash and correctness of execution order rather than only one aspect. Leave it to me.

PR #891, opened 2024-04-16
chore(rust/test): generate `_test.schema.json` automatically

kainzc:
### Description

This pull request addresses the issue opened in #785, aiming to automate the generation of the `_test.schema.json` file in response to modifications to `BundlerOptions` or `TestConfig` in the Rolldown project.

### Implementation Details

- **Creation of `rolldown_testing_config` Crate**: As part of the solution, `TestConfig` was extracted into a new, separate crate named `rolldown_testing_config`. This structural change was essential for enabling direct access to `TestConfig` from the `build.rs` file within the `rolldown_testing` crate, without creating circular dependencies.
- **Integration into Build Process**: The schema generation logic, which utilizes the `schemars` library, has been integrated directly into the `build.rs` script of `rolldown_testing`. This script now automatically regenerates `_test.schema.json` whenever `TestConfig` or related dependencies are updated.

### What has been done

- Integrated the `rolldown_testing_config` crate to centralize configuration definitions.
- Automated the generation of `_test.schema.json` through adjustments in the build process.

### How to Test

1. Make changes to `BundlerOptions` or `TestConfig` within the `rolldown_testing_config` crate.
2. Build the project to trigger the automated schema generation.
3. Check the updated `_test.schema.json` to confirm that it reflects the changes.

PR #1004, opened 2024-04-29
chore: move testConfig to rolldown_common

chiehw:
### Description

refs: https://github.com/rolldown/rolldown/pull/891/files#r1567748277

I try to solve this problem, I am not sure if it is appropriate to move testConfig to rolldown_common, I run `just roll` locally, everything worked fine.


hyf0:
It's not appropriate. `rolldown_common` should not contains code only for testing.

chiehw:
ok

PR #1013, opened 2024-04-30
chore(ci): re-enable codspeed

adriencaccia:
### Description

Re-enable [CodSpeed](https://codspeed.io/), as it was disabled in https://github.com/rolldown/rolldown/pull/926.
The workflow is now fixed and works well.

### Small gotcha
In order for the benchmarks to run properly, the following env variable needs to be set when running the benchmarks:
https://github.com/adriencaccia/rolldown/blob/9b286443607aee45e4f52d0ef939d8d162adddf0/.github/workflows/benchmark-rust.yml#L69
Without it, https://github.com/rolldown/rolldown/blob/388db906355819d70d65c5b23b258414c4735955/crates/rolldown_testing/src/workspace.rs#L32-L33 will fail.
This line is now needed because `cargo-codspeed` now use the directory of the bench's package to run it ( https://github.com/CodSpeedHQ/codspeed-rust/blob/3e11dedb090c75be7d5cc42a3460edba439033f8/crates/cargo-codspeed/src/run.rs#L96)

### DIFF
--- a/.github/workflows/benchmark-rust.yml
+++ b/.github/workflows/benchmark-rust.yml
@@ -25,48 +25,50 @@ concurrency:
   cancel-in-progress: ${{ github.ref_name != 'main' }}
 
 jobs:
-  # https://github.com/CodSpeedHQ/codspeed-rust/issues/29
-  # codspeed-benchmark:
-  #   name: Codspeed Benchmark
-  #   if: false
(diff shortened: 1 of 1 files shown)


hyf0:
Thanks

hyf0 (approved):

PR #1023, opened 2024-05-05
fix: use `citty` to detect if an argument is unspecified

Cecil0o0:
### Description

FIXME: `citty` doesn't support detecting if an argument is unspecified

### DIFF
--- a/packages/rolldown/package.json
+++ b/packages/rolldown/package.json
@@ -98,7 +98,6 @@
     "emnapi": "^1.1.1",
     "execa": "^8.0.1",
     "glob": "^10.3.10",
-    "mri": "^1.2.0",
     "npm-run-all2": "^6.1.2",
     "rolldown": "workspace:*",
     "rollup": "^4.12.1",
--- a/packages/rolldown/src/cli/index.ts
+++ b/packages/rolldown/src/cli/index.ts
@@ -1,21 +1,10 @@
-import process from 'node:process'
-import parseArgs from 'mri'
 import { defineCommand, runMain, showUsage } from 'citty'
-import { ensureConfig, logger } from './utils.js'
 import { bundle } from './commands/bundle.js'
 import {
   version,
   description,
 } from '../../package.json' assert { type: 'json' }
 import { DEFAULT_CONFIG_FILENAME } from './constants.js'
-
-interface ParsedArgs {
-  config?: string | true
-  c?: string | true
-  // `citty` intercept the help option, so we don't need to deal with it
-  // help?: boolean
-  // h?: boolean
-}
 
 const main = defineCommand({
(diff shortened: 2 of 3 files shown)


hyf0:
I'm confused how this PR fix the problem that motioned in the title. From my understanding, in this PR, there's no way to distinguish if user input `rolldown -c ""`  or `rolldown -c`.

Cecil0o0:
@hyf0 Sorry I mis-understood this problem, for now, it still couldn't distinguish if user input `rolldown -c ""` or `rolldown -c`.

Let me try another solution

Cecil0o0:
I have found a way to solve this problem, if we remove the `type` property of  `config` ArgsDef, `citty` would parse it to oneOf `string`, `boolean` and `number`, if so, we could distinguish that.

when user input `rolldown -c`, this arg would be `true`, and `rolldown -c ""` could be number `0`. Back to this situation, handling type `string` or `boolean` is enough.

could it be ok? @hyf0

hyf0:
> I have found a way to solve this problem, if we remove the `type` property of `config` ArgsDef, `citty` would parse it to oneOf `string`, `boolean` and `number`, if so, we could distinguish that.
> 
> when user input `rolldown -c`, this arg would be `true`, and `rolldown -c ""` could be number `0`. Back to this situation, handling type `string` or `boolean` is enough.
> 
> could it be ok? @hyf0

I don't think it's better than the current solution. The best way to solve the problem is solving it in citty's repo. https://github.com/unjs/citty/issues/142

Cecil0o0:
> > I have found a way to solve this problem, if we remove the `type` property of `config` ArgsDef, `citty` would parse it to oneOf `string`, `boolean` and `number`, if so, we could distinguish that.
> > when user input `rolldown -c`, this arg would be `true`, and `rolldown -c ""` could be number `0`. Back to this situation, handling type `string` or `boolean` is enough.
> > could it be ok? @hyf0
> 
> I don't think it's better than the current solution. The best way to solve the problem is solving it in citty's repo. [unjs/citty#142](https://github.com/unjs/citty/issues/142)

first code block:
https://github.com/unjs/citty/blob/0ea32b6c033839957744938f51dcf5b867d425e9/src/args.ts#L10-L43

second code block:
https://github.com/rolldown/rolldown/blob/8645e7782bd179ca1a261f5109ed32d7605ab8bd/packages/rolldown/src/cli/index.ts#L41

After I looked carefully at the two code blocks above, compare their logic that they describe. I thought that your code which uses `parseArgs` exported by `mri` without second param, may be equal to *not* to set `type` for `config` argument.

And maybe, it could just meet `mri`'s design, please look at the option description below:

image


by the way, have a good night!

Issue #939, opened 2024-04-22
Validate `_config.json` using `_config.schema.json`


hyf0:
See https://rolldown.rs/contrib-guide/testing#how-test-cases-are-defined first.

So we have #785 now. Next we want is to validate the `_config.json` using `_config.schema.json` to avoid invalid input.

Validate the input, before parsing it. Panic if it's invalid.

https://github.com/rolldown/rolldown/blob/7b64957632f0942ff69dac7899f74f8d29141b6d/crates/rolldown_testing_config/src/lib.rs#L35-L38

katungi:
Hey @hyf0 , Can I give this a shot?

hyf0:
> shot?

Sure. Thanks.

hyf0:
@katungi Hi. Are you still working on this?

katungi:
Hey @hyf0 yes I am. Will be done by tomorrow.

chiehw:
> Validate the input, before parsing it. Panic if it's invalid.

Does this mean checking `_config.json` in build.rs? @hyf0

IWANABETHATGUY:
Closed in favor of #1042

PR #1042, opened 2024-05-06
chore(rust/test): validate `_config.json`

katungi:
### Description

This pull request fixes the issue #939, which will validate the `_config.json` using `_config.schema.json` to avoid invalid input.

### Implementation Details
- Pass an additional param to `from_config_path` function. The additional param is the path to the schema.
- Read the schema file and compile it using `jsonschema` then cache the file.
- Read and parse the `config.json` file.
- Validate the schema and only generate

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -35,7 +35,7 @@
   "json.schemas": [
     {
       "fileMatch": ["_config.json"],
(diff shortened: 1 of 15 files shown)


hyf0 (approved):

hyf0:
Thanks

PR #1058, opened 2024-05-08
chore(renovate): group updates for patch and minor

hyf0:
### Description

- patch and minor updates now would be grouped together except the `rustc` bumping. `rustc` is still updated per PR not grouped.
- renovate doesn't support built-in autoApprove for github. So the related config is removed. We gonna use https://github.com/apps/renovate-approve instead.

We might need to only group patch updates, since minor updates may have a big chance to break CI.  However, let's see how it goes now.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -12,24 +12,34 @@
   ],
   "packageRules": [
     {
+      "automerge": true,
+      "matchUpdateTypes": ["patch", "minor", "lockFileMaintenance", "pin"]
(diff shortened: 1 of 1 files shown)


hyf0:
* **#1058** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1058?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #1081, opened 2024-05-10
chore(renovate): fix schedule

jackw:
### Description

Updates the renovate config to run after 12:00pm every weekday which I believe is correct syntax and inline with when the maintainers of this repo would like renovate to run.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -8,7 +8,7 @@
   "assignees": ["hyf0", "IWANABETHATGUY", "underfin"],
   "assigneesSampleSize": 1,
   "timezone": "Asia/Shanghai",
-  "schedule": ["after 12:00pm every weekday"],
+  "schedule": ["after 12:00pm and before 4:00am"],
   "ignorePaths": [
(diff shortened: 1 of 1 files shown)


hyf0:
Good suggestions.

What I intend to do is to update deps every day. And the updates should be grouped in a certain time of a day not a range, if some deps miss the time, it has to wait for the next day. Using range might cause multiple updates in multiple PRs, but we want to group them together.

Using weekday is nice, but I think again, it's fine that we update deps in non-weekday.

jackw:
@hyf0 thanks for the feedback and suggestions. Based on what you've described maybe the schedule would be better suited to:

```
"schedule": [
  "after 12:00pm and before 4:00am",
],
```
According to the [renovate docs](https://docs.renovatebot.com/known-limitations/#the-mend-renovate-app-and-scheduled-jobs) it's best to give at least 3 or 4 hrs for the bot to check the repo.

hyf0 (approved):
Thanks

PR #1082, opened 2024-05-10
feat: mutate `Output` and `OutputChunk`

underfin:
### Description

The pr intends to make the the item of `Output`  is mutable.

First, it should be mutable at rust side and js side, i changed  the `bundle` argument to `Vec<Output>` at `Plugin#generate_bundle/write_bundle`, before it using `&Vec<Output>`, it will let the hook can mutate `bundle`, so it need to return an new `bundle`, if you not changed it, just return the original value. It also remove `Arc` at `Output`, using `Arc` is difficult to mutate it.

For rust side data pass to js side, it need to serialize and deserialize, it has some overhead. Here is some way to do it at the binding side, here only discuss mutate the item of `Output`, not including delete item and add item.
A. serialize the `Output` to js side, and mutate it, then serialize it to rust side.  It has some unnecessary overhead if we only  mutate one field of dada.
B. serialize the `Output` to js side but hold on the `Output` mutable reference and export the setter function of fileds, if some fields mutate, it only need to serialize corresponding data to rust side.

The pr implement `B`, it is better than `A`. If you have some good ideas, please let me know.

I will support the delete the item of `Output` at next pr. It only need to export a `deleteItem` API and using `Proxy` to intercept it at js side. 

For add item to `Output`, because the data comes from js side, we need to serialize it to rust side and push to `Output`. But it hasn't the situation at now, it will be delay util somebody need the feature.


crates/rolldown_binding/src/options/plugin/parallel_js_plugin.rs
```diff
@@ -139,23 +139,25 @@ impl Plugin for ParallelJsPlugin {
...
     }
-    Ok(())
   }
 
   async fn write_bundle(
     &self,
     ctx: &rolldown_plugin::SharedPluginContext,
-    bundle: &Vec<rolldown_common::Output>,
-  ) -> rolldown_plugin::HookNoopReturn {
+    bundle: Vec<rolldown_common::Output>,
```
hyf0: Want's wrong with using `&mut Vec<rolldown_common::Output>`?

hyf0:
> For add item to Output, because the data comes from js side, we need to serialize it to rust side and push to Output. But it hasn't the situation at now, it will be delay util somebody need the feature.

Be hesitate to support such behavior. Users should use `this.emitFile` to add item in output. https://rollupjs.org/plugin-development/#generatebundle:~:text=Do%20not%20directly%20add%20assets%20to%20the%20bundle.%20This%20circumvents%20internal%20mechanisms%20that%20Rollup%20has%20for%20tracking%20assets

crates/rolldown_binding/src/options/plugin/parallel_js_plugin.rs
underfin: It should be correct. Let me update code.

hyf0 (approved):

PR #1083, opened 2024-05-10
feat: support delete OutputBundle item

underfin:


packages/rolldown/tests/fixtures/plugin/generate-bundle/_config.ts
```diff
@@ -31,13 +31,16 @@ export default defineTest({
           expect(isWrite).toBe(true)
           // Mutate chunk
           chunk.code = 'console.error()'
+          // Delete chunk
+          delete bundle['index.js']
```
hyf0, line 35: How rollup deletes assets? Are we introducing new API?

underfin: > How rollup deletes assets? Are we introducing new API?

Rollup just using `detele`, the new API is unnessary.

hyf0 (approved):

PR #1103, opened 2024-05-12
feat: derive side effects from package json

IWANABETHATGUY:
### Description
1.  derive side effects from package json
2. the overhead look unavoidable, I will try to optimize it when I have time.


crates/rolldown/src/module_loader/module_loader.rs
```diff
@@ -24,13 +27,19 @@ use crate::{SharedOptions, SharedResolver};
 
 pub struct IntermediateNormalModules {
   pub modules: IndexVec<NormalModuleId, Option<NormalModule>>,
+  pub module_side_effects: IndexVec<NormalModuleId, Option<Arc<PackageJson>>>,
```
hyf0: We probably don't need this to get package.json.

IWANABETHATGUY: OK, is there another place we already store the related `PackageJson`, I can't find the related logic in our codebase.

hyf0: The code is placed in resolver. Let me create a PR to show it. I'm almost done it.

hyf0: https://github.com/rolldown/rolldown/pull/1170

crates/rolldown/src/stages/link_stage/tree_shaking.rs
```diff
@@ -10,21 +10,25 @@ struct Context<'a> {
   symbols: &'a Symbols,
   is_included_vec: &'a mut IndexVec<NormalModuleId, IndexVec<StmtInfoId, bool>>,
   is_module_included_vec: &'a mut IndexVec<NormalModuleId, bool>,
+  has_export_used: &'a mut IndexVec<NormalModuleId, bool>,
```
hyf0: What does `has_export_used` field do?

hyf0 (dismissed):

hyf0 (approved):

PR #1107, opened 2024-05-13
fix: test_package_json_browser_over_main_node

zhusjfaker:
### Description

https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_packagejson_test.go#L507

Fix bugs in past migration code


zhusjfaker:
@IWANABETHATGUY

hyf0 (approved):
Thanks

PR #1109, opened 2024-05-13
chore(rust/esbuild-tests): enable more test cases.

Cecil0o0:
### Description

`define_optional_chain_lowered` totally equal with `define_optional_chain`, I leave the latter?


hyf0:
Neither define or TLA are supported. We aren't gonna enable these tests.

Cecil0o0:
> Neither define or TLA are supported. We aren't gonna enable these tests.

Could I ask for reason Why we not to support TLA?

hyf0:
TLA is not supported yet. We aren't saying not support TLA.

Cecil0o0:
> TLA is not supported yet. We aren't saying not support TLA.

Got. Try to find other cases

PR #1121, opened 2024-05-14
feat: support TransformPluginContext#getCombinedSourcemap

underfin:


hyf0 (approved):
Looks like a very high advanced features. Wonder it's real-world usages.

underfin:
It be used at vite `cssPostPlugin`, but I'm not notice it used at dev server..

PR #1124, opened 2024-05-14
fix test package json browser cases

zhusjfaker:
### Description

Fix bugs in past migration code

https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_packagejson_test.go#L640

https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_packagejson_test.go#L181

https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_packagejson_test.go#L260


hyf0 (approved):
Thanks a lot.

PR #1134, opened 2024-05-15
chore(rust/tests): enable esbuild tests.

Cecil0o0:


Cecil0o0:
[cargo-test (windows-latest)](https://github.com/rolldown/rolldown/actions/runs/9098024438/job/25007305635?pr=1134#logs) runs failed, because of the '/esbuild/default/source_map' case.

I don't know why there are two different source map between Windows and Mac, and how to deal with it?

PR #1142, opened 2024-05-16
chore: use web-tree-sitter instead node for better compatibility

IWANABETHATGUY:
### Description
1. tree-sitter was written by `c`, and has many compatible issues when using node binding, this PR changed node binding into **wasm** to reduce some compatible issues.


scripts/tree-sitter-wasm/tree-sitter-go.wasm
Boshen, line 1: adding a 235KB wasm to a github repo will blow up git :-(

Boshen:
> to reduce some compatible issues.

Can you state the actual issues?

hyf0:
> > to reduce some compatible issues.
> 
> Can you state the actual issues?

pnpm install `tree-sitter-go` failed in my mac. I guess its postinstall scripts has some problems. Using wasm would be much proper than compiling while pnpm install.

IWANABETHATGUY:
> > to reduce some compatible issues.
> 
> Can you state the actual issues?

The npm package "tree-sitter" is compiled using [node-gyp](https://github.com/tree-sitter/node-tree-sitter/blob/master/binding.gyp). You need to recompile it when using a different Node.js version. If you have experience with using "sass," you may already understand this compatibility issue.

scripts/tree-sitter-wasm/tree-sitter-go.wasm
IWANABETHATGUY: +235KB for `tree-sitter-go.wasm` but reduce `tree-sitter-go` npm package, overall the size of the git repo should not blow up

Boshen: https://blog.logrocket.com/efficiently-manage-large-files-git-with-git-lfs/#storing-large-files-bad-idea

hyf0: Solved. Wonder if we need to delete the commit that adding the file in this PR.

hyf0 (approved):

PR #1144, opened 2024-05-16
refactor: shared module table at PluginContext

underfin:
### Description

The pr intend to support `PluginContext#getModuleInfo`, before i already support `NormalModule` cast to `ModuleInfo`, it only work at `moduleParesd` hook. But the `PluginContext#getModuleInfo` need to valid at plugin lifetime. I consider two ways to support it.

A. Add a `moduleInfos`(Dashmap) to owner `ModuleInfo`at `PluginContext`, but it need to set the module info when the `ModuleInfo` fields change, it need to re-set many times, due to https://rollupjs.org/plugin-development/#this-getmoduleinfo. It also has some overhead if the plugin not used it.
B. Shared `ModuleTable` at `PluginContext`, the typing is `Arc<RwLock<ModuleTable>>`, we only need to muate it at module loader and sort modules, it always do it at single thread, so we are using `RwLock` to avoid caused performance regression if lock conflict. Other, The`ModuleInfo` will be lazy, if you used it, it will cast from `NormalModule`.

The pr intend to implement `B`, it is sound good to me. If you have some good ideas, please let know.

The mutate `SharedModuleTable` is not finished at module loader, it look like is complex, i intend to do it at next pr.

### DIFF
--- a/crates/rolldown/src/bundler.rs
+++ b/crates/rolldown/src/bundler.rs
@@ -11,6 +11,7 @@ use crate::{
   BundlerOptions, SharedOptions, SharedResolver,
 };
 use anyhow::Result;
+use rolldown_common::SharedModuleTable;
 use rolldown_error::BuildError;
 use rolldown_fs::{FileSystem, OsFileSystem};
 use rolldown_plugin::{BoxPlugin, HookBuildEndArgs, HookRenderErrorArgs, SharedPluginDriver};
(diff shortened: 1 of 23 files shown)


hyf0:
I give a rough look, it's not worth to wrap `ModuleTable` in `Arc<Mutex<...>>` just for support `this.getModuleInfo`. 

Compatibility not comes first. If we can't implement a compatible feature in reasonable way, we either try to find other ways or just give up the feature temperately. The compatible feature should not deeply affects the internal core code of rolldown

underfin:
> Compatibility not comes first. If we can't implement a compatible feature in reasonable way, we either try to find other ways or just give up the feature temperately. The compatible feature should not deeply affects the internal core code of rolldown

I'm also agree it, I'm not sure for the compatibility, let us discuss it at the meeting.

PR #1149, opened 2024-05-17
chore: don't ignore `node_modules` under `crates/rolldown/esbuild/`

IWANABETHATGUY:
### Description
1. I pulled the latest tests from `esbuild` repo, so some tests  are updated.
2. update `gitignore` to make sure the `node_modules` under `tests/esbuild` are not ignored.


scripts/misc/gen-esbuild-test.mjs
```diff
@@ -15,7 +15,7 @@ import * as fsExtra from 'fs-extra'
 
 const TREE_SITTER_WASM_GO_FILENAME = path.resolve(
   import.meta.dirname,
-  '../../tmp/tree-sitter-go.wasm',
+  '../../temp/tree-sitter-go.wasm',
```
hyf0: It's `tmp`. `temp` is gonna be removed in `.gitignore` . Benchmark input are also placed in `tmp` not `temp`.

hyf0:
```suggestion
  '../../tmp/tree-sitter-go.wasm',
```

IWANABETHATGUY: wait, if I change it to `tmp` then it would be stored in the git repository, does that violate what you said to download on demand?

IWANABETHATGUY: Or, I need to add `tmp` dir to `.gitignore`?

hyf0: https://github.com/rolldown/rolldown/blob/d73d334564a19bf72ff0ece84d44766417f0ada7/.gitignore#L3

IWANABETHATGUY: ok

scripts/misc/gen-esbuild-test.mjs
```diff
@@ -517,7 +517,7 @@ function ensureTreeSitterWasmGo() {
   fsExtra.ensureDirSync(path.dirname(TREE_SITTER_WASM_GO_FILENAME))
   return new Promise((rsl, rej) => {
     nodeHttps.get(
-      'https://github.com/tree-sitter/tree-sitter.github.io/raw/master/tree-sitter-go.wasm',
+      'https://tree-sitter.github.io/tree-sitter-go.wasm',
```
IWANABETHATGUY, line 520: I tried `wasm` from the previous url but got some runtime error, so I used `wasm` URL from the `tree-sitter` playground.

hyf0 (approved):

PR #1150, opened 2024-05-17
fix: tree shaking multi decl

IWANABETHATGUY:
### Description
for case: 
```
// index.js
import {a,} from './share'
console.log(a)
// share.js
export const a = 1000, b = 1000;
```
#### Before
```
// share.js
const a = 100, b = 1000;
// main1.js
console.log(a);
```

#### After
```
// share.js
const a = 100;
// main1.js
console.log(a);
```


crates/rolldown/tests/esbuild/default/ambiguous_reexport_msg/artifacts.snap
```diff
@@ -9,13 +9,13 @@ input_file: crates/rolldown/tests/esbuild/default/ambiguous_reexport_msg
 
 ```js
 // a.js
-let a = 1, x$1 = 2;
+let a = 1;
```
IWANABETHATGUY, line 12: align with esbuild

IWANABETHATGUY: https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/snapshots/snapshots_default.txt#L1-L16

crates/rolldown_common/src/types/stmt_info.rs
```diff
@@ -77,19 +86,48 @@ pub struct StmtInfo {
...
   pub fn to_debug_stmt_info_for_tree_shaking(&self) -> DebugStmtInfoForTreeShaking {
     DebugStmtInfoForTreeShaking {
-      is_included: self.is_included,
+      is_included: self.partial_included(),
       side_effect: self.side_effect,
       source: self.debug_label.clone().unwrap_or_else(|| "<Noop>".into()),
     }
   }
+
+  pub fn partial_included(&self) -> bool {
```
hyf0, line 103: Could we convert
```js
var a, x;
```

to

```js
var a;
var x;
```

before scanning since we already need to mutate ast to hoist import statements.

Introducing `partial_included` and `fully_included` just for solving the above case is kind of over complex.

hyf0: I would consider `pre-transpile` is a more genernal and universal way to solve many tree-shaking edge cases.

For example

we could deal with the following case by transpile
```js
{
  var a;
  var b;
}

export { a, b }
```

to

```
{
var a;
var b;

export {a, b }
}
```

without changing how tree_shaking works currently.

hyf0: The key problem is that the current esbuild's treeshaking implementation works on statements level.

I understand that introducing `partial_included` and `fully_included` could solve the multiple declaration case, but it requires changes on many places, linking, code splitting, deconflicting and finalizing. It's kind of unworthy.

The `pre-transpile` way solve the problem by only need to add extra code in transpile step without touching any other places.

hyf0: However, I would consider `pre-transpile` way is kind of workaround. If we could design a treeshaking implementation that's works on the level that smaller than statements, I would not suggest using `pre-transpile`.

In short, I would suggest using `pre-transpile` way on this case for patching esbuid's treeshaking implementation.

If we gonna design a new treeshaking implementation, we could introducing new concepts to solve the problem.

hyf0: We might consider the `pre-transpile` is a another representation of `AST -> HIR`, while we are `AST -> (Tree shaking friendly) AST`.

IWANABETHATGUY: Yeah, but transpile would introduce another source map remapping, which should be another overhead, I will try to find another way to finish this with minimal modification.

hyf0:
It's suggested that:

- You could place the transform logic on multiple declarations in https://github.com/rolldown/rolldown/blob/4dc323cf46681d57d56592ee0b7ebdefdaa96268/crates/rolldown/src/utils/tweak_ast_for_scanning.rs
- Create a `VariableDeclarationExt` trait in https://github.com/rolldown/rolldown/blob/4dc323cf46681d57d56592ee0b7ebdefdaa96268/crates/rolldown_oxc_utils/src/ext/mod.rs
- Create a method `fn unfold_declarators(&mut self, alloc: &Allocator) -> Vec<VariableDeclaration>` on `VariableDeclarationExt`. Feel free to choose a more suitable function name or signature.

IWANABETHATGUY:
This PR is blocked by the source map modification revert,

IWANABETHATGUY:
closed in favor of #1564

PR #1179, opened 2024-05-20
chore: CR for #1103

hyf0:
### Description

CR for #1103


crates/rolldown/src/module_loader/normal_module_task.rs
```diff
@@ -114,12 +112,20 @@ impl NormalModuleTask {
     }
 
     let resource_id = ResourceId::new(Arc::clone(&self.resolved_path.path));
+    let stable_resource_id = resource_id.stabilize(&self.ctx.input_options.cwd);
+
+    let side_effects = self
+      .package_json
+      .as_ref()
+      .and_then(|p| p.check_side_effects_for(&stable_resource_id))
```
hyf0, line 120: Though I used `stable_resource_id`, which is a relatived path to cwd, I think it's also ok we use absolute path `resource_id`.

PR #1183, opened 2024-05-21
fix: esbuild case test_common_js_variable_in_esm_type_module

zhusjfaker:
### Description

https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_packagejson_test.go#L2757

The original case threw a warning, which is ignored here.


hyf0 (approved):
Thanks

PR #1184, opened 2024-05-21
fix: test_package_json_type_should_be_types

zhusjfaker:
### Description

https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_packagejson_test.go#L2516

The original case threw a warning, which is ignored here.


hyf0 (approved):
Thanks

Issue #1185, opened 2024-05-21
When rolldown handles external modules,it should replace the source in the ImportDeclaration with oxc-resolver to return the result in the error


zhusjfaker:
in  esbuild repo 

https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_packagejson_test.go#L2878

in rolldown esbuild test package_json case

https://github.com/rolldown/rolldown/tree/main/crates/rolldown/tests/esbuild/packagejson/.test_package_json_subpath_import_node_builtin_issue3485

oxc-resolver test unit case
```rust
   #[test]
    fn test_resolver_package_json_bad_main() {
        let options = ResolveOptions {
            alias_fields: vec![vec!["browser".into()]],
            exports_fields: vec![vec!["exports".into()]],
            condition_names: vec!["node".into(), "import".into(), "require".into()],
            main_fields: vec!["main".into()],
            // alias: vec![("#fs".into(),vec!["fs".into()] ), ("#http".into(),vec![ "node:http".into()])],
            ..ResolveOptions::default()
        };
        let dir = Path::new("/Users/jason.zhu/Desktop/github/rolldown/crates/rolldown/tests/esbuild/packagejson/test_package_json_subpath_import_node_builtin_issue3485");
        let resolver = Resolver::new(options);
        let res = resolver.resolve(dir, "#http").unwrap();
        println!("res-> \n {:#?}", res);
    }
```

terminal output
```sh
failures:

---- tests::test_resolver_package_json_bad_main stdout ----
thread 'tests::test_resolver_package_json_bad_main' panicked at tests/mod.rs:90:50:
called `Result::unwrap()` on an `Err` value: NotFound("node:http")
stack backtrace:
   0: rust_begin_unwind
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/result.rs:1649:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/result.rs:1073:23
   4: mod::tests::test_resolver_package_json_bad_main
[...]

PR #1186, opened 2024-05-21
test(esbuild/packgaejson): enable test_package_json_subpath_import_node_builtin_issue3485 

zhusjfaker:
### Description

https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_packagejson_test.go#L2878

Please check the result of the #1185  issue before merging in this MR.

https://github.com/rolldown/rolldown/issues/1185

#1185 

Confirm that the pack is what the rolldown user wants.

```js

import { default as fs } from "#fs";
import { default as http } from "#http";

// entry.js
fs.readFileSync();
http.createServer();
```

even "#fs"  will cause the js to crash


hyf0:
Please rename the title to `test(esbuild/packgaejson): enable test_package_json_subpath_import_node_builtin_issue3485 `

hyf0:
The result is not right. However, it's ok we merge them first and solve #1185 later.

zhusjfaker:
> The result is not right. However, it's ok we merge them first and solve #1185 later.

I'll make an issue of the progress of all esbuild packagejson cases, so that we can track the progress of all esbuild packagejson cases by table, and the reasons for the obstacles and related merge-ins in the MR and other related issue discussions.

zhusjfaker:
As of this MR, I've finished migrating all the cases that were successfully packaged by esbuild, but there are still a few cases that throw exceptions, so I'll reflect the progress of the migration in the overall issue.

hyf0 (dismissed):

crates/rolldown/tests/esbuild/packagejson/test_package_json_subpath_import_node_builtin_issue3485/_config.json
```diff
@@ -7,5 +7,6 @@
       }
     ],
     "platform": "Node"
```
hyf0:
```suggestion
    "platform": "node"
```

See #1188.

zhusjfaker:
> The result is not right. However, it's ok we merge them first and solve #1185 later.

done

crates/rolldown/tests/esbuild/packagejson/test_package_json_subpath_import_node_builtin_issue3485/_config.json
zhusjfaker: ok

zhusjfaker: done

hyf0 (approved):
Thanks

PR #1208, opened 2024-05-22
feat: bump oxc to v0.13.1

Boshen:
cc @underfin do you need to update the sourcemap change?


Boshen:
🤔 my node.js in nix is stuck in `v18.14.1` and I can't update snapshits, it fails with `node: bad option: --import`.

Can someone update the snapshots for me 😅 Thanks.

crates/rolldown/tests/snapshots/fixtures__filename_with_hash.snap
hyf0, line 1: Interesting. Many fixtures's hash is changed, but the content emits to snapshot doesn't change.

hyf0 (approved):

underfin:
cc @IWANABETHATGUY the sourcemap is supported for the ast has unorder span.

PR #1213, opened 2024-05-22
fix(tree_shaking): determine indirect side effects

IWANABETHATGUY:
### Description
1. derived side effects should have higher priority, we should keep the side effects that derived from `package.json` as is whatever the side effects of the dependency is.
2. You could refer esbuild as a example https://esbuild.github.io/try/#YgAwLjIwLjIALS1idW5kbGUgLS1taW5pZnk9ZmFsc2UgLS1mb3JtYXQ9ZXNtAGUAZW50cnkuanMAaW1wb3J0IHtifSBmcm9tICIuL3BhY2thZ2VzL2xpYi5qcyI7CmNvbnNvbGUubG9nKGIpCgAAcGFja2FnZXMvbGliLmpzAGV4cG9ydCAqIGZyb20gJy4uL2Fub3RoZXIvYS9pbmRleCcKZXhwb3J0ICogZnJvbSAnLi4vYW5vdGhlci9iL2luZGV4JwoAAGZpbGUuanMAAABhbm90aGVyL2EvaW5kZXguanMAZXhwb3J0IGNvbnN0IGEgPSAxMDAwMDsKY29uc29sZS5sb2coJ3Jlc3VsdCcpCgAAYW5vdGhlci9iL2luZGV4LmpzAGV4cG9ydCBjb25zdCBiID0gMTAwOwoAAHBhY2thZ2VzL3BhY2thZ2UuanNvbgB7CiAgInNpZGVFZmZlY3RzIjogZmFsc2UKfQo


hyf0 (dismissed):

crates/rolldown/src/stages/link_stage/tree_shaking.rs
```diff
@@ -41,7 +42,7 @@ fn include_module(ctx: &mut Context, module: &NormalModule) {
   module.import_records.iter().for_each(|import_record| match import_record.resolved_module {
     rolldown_common::ModuleId::Normal(importee_id) => {
       let importee = &ctx.modules[importee_id];
-      if !ctx.tree_shaking || importee.side_effects {
+      if !ctx.tree_shaking || importee.side_effects.inner() {
```
hyf0: Let's create a method called has_side_effect on NormalModule.

hyf0 (approved):

PR #1223, opened 2024-05-23
feat: handle `proposal-json-superset` chars with iteration of superset-contained string

7086cmd:
### Description

The crate `rolldown_lang_json` has a `TODO` related to the `JSON superset` proposal. It may solve this handling task with the mention of iterating the string.

Since I am not a professional Rust developer but a high-school student, I may not sync the PR process immediately. I just read a little of the project, and I hope my efforts will improve the repository a little bit :)

This PR creates a function named `escape_json_superset`, returns a `Cow::<str>`, and checks if the JSON string contains specific characters. If so, the function will iterate the string and replace the UTF-8 char, and use `\u2028` or `\u2029` instead.

### DIFF
--- a/crates/rolldown_lang_json/src/lib.rs
+++ b/crates/rolldown_lang_json/src/lib.rs
@@ -1,11 +1,14 @@
+use std::borrow::Cow;
+
 use anyhow::Ok;
 use rolldown_utils::ecma_script::is_validate_identifier_name;
 use serde_json::Value;
-// TODO: handling https://github.com/tc39/proposal-json-superset
+// TODO: handling https://github.com/tc39/proposal-json-superset // Fixing by Ethan Goh (@7086cmd) in this issue.
 
 pub fn json_to_esm(json: &str) -> anyhow::Result<String> {
+  let json = escape_json_superset(json);
   // TODO: use zero-copy deserialization
-  let json_value: Value = serde_json::from_str(json)?;
+  let json_value: Value = serde_json::from_str(&json)?;
 
   match json_value {
     Value::Object(map) => {
@@ -37,5 +40,21 @@ pub fn json_to_esm(json: &str) -> anyhow::Result<String> {
       let json_str = serde_json::to_string(&json_value)?;
       Ok(format!("export default {json_str}"))
     }
+  }
+}
+
+fn escape_json_superset(json: &str) -> Cow<str> {
+  if json.contains('\u{2028}') || json.contains('\u{2029}') {
+    let mut escaped = String::with_capacity(json.len());
+    for c in json.chars() {
+      match c {
+        '\u{2028}' => escaped.push_str("\\u2028"),
+        '\u{2029}' => escaped.push_str("\\u2029"),
+        _ => escaped.push(c),
+      }
+    }
+    Cow::Owned(escaped)
+  } else {
+    Cow::Borrowed(json)
   }
 }


7086cmd:
I am sorry for leaving `Initial Commit` due to the transformation of the codespace.

> Since I am on campus, I can't directly clone the repo to localhost due to the performance of the develop computer, and apologize for my rudeness :)

7086cmd:
By the way, tests weren't written to test its correctness (including the crate). I don't have time to write tests tomorrow. If necessary, I will write it the day after tomorrow (May 25).

crates/rolldown_lang_json/src/lib.rs
```diff
@@ -39,3 +42,19 @@ pub fn json_to_esm(json: &str) -> anyhow::Result<String> {
     }
   }
 }
+
+fn escape_json_superset(json: &str) -> Cow<str> {
+  if json.contains('\u{2028}') || json.contains('\u{2029}') {
```
Boshen, line 47: This is a performance regression. I think we need some clever way to detect this, which doesn't happen 99.99999% of the case.

7086cmd: So it mean that `.contains` may lead to a lack of performance? I may consider other ways.
But in my idea, maybe iterating the string causes it more ().

7086cmd: Sorry, I misunderstood your meaning. It does be a VERY small case.

I wonder what `serde` would do after parsing it?

7086cmd: I have asked ChatGPT, and it responded the following code:

```rust
use anyhow::Result;
use serde_json::Value;
use std::borrow::Cow;

pub fn json_to_esm(json: &str) -> Result<String> {
    // Preprocess the JSON to handle the JSON superset by escaping U+2028 and U+2029
    let json = escape_json_superset(json);

    // Deserialize the preprocessed JSON string
    let json_value: Value = serde_json::from_str(&json)?;

    // Transform the JSON value to ESM format
    let esm_string = json_to_esm_transform(&json_value)?;
    Ok(esm_string)
}

fn escape_json_superset(json: &str) -> Cow<str> {
    let mut escaped = String::with_capacity(json.len());
    let mut last_index = 0;
    let mut found_escape = false;

    for (index, ch) in json.char_indices() {
        match ch {
            '\u{2028}' | '\u{2029}' => {
                if !found_escape {
                    found_escape = true;
                    escaped.push_str(&json[last_index..index]);
                }
                match ch {
                    '\u{2028}' => escaped.push_str("\\u2028"),
                    '\u{2029}' => escaped.push_str("\\u2029"),
                    _ => unreachable!(),
                }
                last_index = index + ch.len_utf8();
            }
            _ => {}
        }
    }

    if found_escape {
        escaped.push_str(&json[last_index..]);
        Cow::Owned(escaped)
    } else {
        Cow::Borrowed(json)
    }
}

fn json_to_esm_transform(value: &Value) -> Result<String> {
    // Implement the logic to convert the JSON value to ESM format.
    // For demonstration, let's assume you want to convert a simple JSON object to a JS module export.

    let mut esm_string = String::new();
    if let Some(obj) = value.as_object() {
        esm_string.push_str("export default ");
        esm_string.push_str(&serde_json::to_string(obj)?);
        esm_string.push(';');
    } else {
        return Err(anyhow::anyhow!("Expected a JSON object"));
    }
    Ok(esm_string)
}
```

If it is a alternative and faster method, I will take the modification into consideration the day after tomorrow.

IWANABETHATGUY: FYI, https://github.com/IWANABETHATGUY/escape-json, I did a benchmark before. But it may have better way to do this.

7086cmd: Thanks for benchmarking! I wonder how `serde` will behavior. Otherwise we are just "talking military strategies on paper (纸上谈兵)".

7086cmd: It returns it as well:
```bash
Object {"x": String("\u{2028}"), "y": String("\u{2029}")}
{"x":" ","y":" "}
```

and `assert_eq!` passed in:
```rust
assert_eq!(data, stringified)
```

7086cmd:
How about taking the `JSON` library in Python as an example? It transforms UTF-8 characters into `\u` when dumping. In bundling JSON, may be a good choice not only for encoding conflicts but also can elegantly solving this problem.

I tested the `escape` char in Python, and it returns `\u2028`, etc. in the way we excepted in the `TODO`.

IWANABETHATGUY:
We should respect mainstream bundler behavior.

IWANABETHATGUY:
Keeping `\u2028`  and `\u2029` as is should be fine,
## Example1
```js
//index.js
import json from './answer.json';
console.log('outer:',json);

// answer.json

{
  "a": "'\u2028'"
}

```
**webpack**
```js
/******/ 	var __webpack_modules__ = ({

/***/ "./a/answer.json":
/***/ ((module) => {

module.exports = {"a":"'\u2028'"};

/***/ })

/******/ 	});
... snip . ...
```

**esbuild**
```js
// answer.json
var answer_default = {
  a: "'\u2028'"
};

// entry.js
console.log("outer:", answer_default);
```

IWANABETHATGUY:
webpack will optimized the big json string, wrap the json into `JSON.parse`: 
```js
//index.js
import json from './answer.json';
console.log('outer:',json);

// answer.json

{
  "a": "'\u2028'jfieawijfiawjiofjiewaojio"
}


```

**webpack**
```js
/***/ "./a/answer.json":
/***/ ((module) => {

module.exports = /*#__PURE__*/JSON.parse('{"a":"\'\\u2028\'jfieawijfiawjiofjiewaojio"}');

/***/ })
// .... snip ..
```

IWANABETHATGUY:
Since we don't have json optimization, keep the character as is should be fine(remove the TODO)? @hyf0 what do you think?

7086cmd:
I agree. Since things won't be unexcepted when calling the module with `\u2028`, etc., I think simply closing the pr is ok.

But I don't know how the ES2016 environment behaviors. If it is successful, I will close it.

7086cmd:
ESLint when env is ES2017 doesn't report errors. I will close it first due to it is unnecessary.

PR #1224, opened 2024-05-23
test(esbuild/packgaejson): test_package_json_syntax_error

zhusjfaker:
### Description

https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_packagejson_test.go#L65

Fixed the 2 cases mentioned above

And add the ability to capture the main thread's uncaptured panic in order to pass the appropriate test cases.


zhusjfaker:
Sorry for the non-negotiable change to one part of the Case module code, but I think it's test code that shouldn't be too affected, and has fully passed the "just roll" test command.

crates/rolldown_testing_config/src/lib.rs
```diff
@@ -22,6 +22,9 @@ pub struct TestConfig {
   #[serde(default)]
   /// If `true`, the sourcemap visualizer will be snapshot.
   pub visualize_sourcemap: bool,
+  #[serde(default)]
+  /// If `true`, the fixture throw an unknown error that can be caught in the final execution phase.
+  pub expect_uncaught_error: bool,
```
hyf0, line 27: I don't really quite get the idea of `expect_uncaught_error`.

 If some fixtures panics, it means there are problems in rolldown's code and need to be fixed. Marking it with `expect_uncaught_error: true` helps nothing.

zhusjfaker: Currently, the case parsing for the two packagejsons are not included in vec[buildError], esbuild is trying to test that if there is any writing error in the packagejsons, the program can be terminated normally and throw an exception, so far there is no problem with the implementation of rolldown. However, if you want to continue testing in all cases without interruption, you need to catch those errors that are not in the context of vec[buildError].

zhusjfaker: Based on these issues, so I added these codes

hyf0:
Panic fixtures need to either being fixed or ignored. Adding `expect_uncaught_error: true` to make the test pass and pretend everything is fine is not expected.

zhusjfaker:
> Panic fixtures need to either being fixed or ignored. Adding `expect_uncaught_error: true` to make the test pass and pretend everything is fine is not expected.

Sorry I wanted to emphasize that the standard testing process for esbuild includes these panic fixtures, which are also not fixed in the esbuild testing process. I'm not sure if that's a clear explanation.

zhusjfaker:
And you don't add this field, if there is a problem with the fixtures and it needs to be fixed it can be perceived and also reflected now

zhusjfaker:
> anic fixtures need to either being fixed or ignored. Adding `expect_uncaught_error: true` to make the test pass and pretend everything is fine is not expected.

maybe i will change the logic and make sure every oxc-resovler panic will be caughted

zhusjfaker:
If I'm done with the code, I'll let you review it again.

hyf0:
> Sorry I wanted to emphasize that the standard testing process for esbuild includes these panic fixtures

This is already done by `expected_error` field. If some errors should be caught by `expected_error` but don't, it's a bug of rolldown and need to be fixed, which is usually transform the panic to returning a `BuildError`.

Panics is usually program errors and can't not be handled, so it's always unexpected. If some errors are wrongly considered as programs errors, we should fix the behavior rather than thinking programs errors are expected.

Boshen:
Rolldown should never panic. Is this test case panicking right now? We need to fix the panic and let it return an error if this is the case.

zhusjfaker:
> Rolldown should never panic. Is this test case panicking right now? We need to fix the panic and let it return an error if this is the case.

i will fix it , don't worry

PR #1225, opened 2024-05-23
feat(tree_shaking):  pure annotation

IWANABETHATGUY:


crates/rolldown/src/ast_scanner/side_effect_detector.rs
```diff
@@ -194,10 +206,64 @@ impl<'a> SideEffectDetector<'a> {
...
       | Expression::JSXFragment(_) => true,
+      Expression::CallExpression(expr) => {
+        let leading_comment_index: Option<usize> = if self.attached_comment_vecmap.is_empty() {
+          None
+        } else {
+          // because the span is the content of comment, so the two span should never overlapped,
+          // the result Variant should be `Err(usize)`
+          let insert_index = self
+            .attached_comment_vecmap
+            .binary_search_by(|probe| probe.0.end.cmp(&expr.span.start));
```
hyf0, line 217: I'm fine with this implementation.

However, this feel not right about how to get comments of ast nodes. I was expecting we could get comment of ast node by simply using API like `trivias.get_comments(expr.span)`.

Once oxc support emit comments in codegen, we could see how oxc get comments from oxc ast nodes.

@Boshen cc

crates/rolldown/src/module_loader/runtime_normal_module_task.rs
```diff
@@ -112,6 +112,7 @@ impl RuntimeNormalModuleTask {
     );
     let mut symbol_for_module = AstSymbols::from_symbol_table(symbol_table);
     let facade_path = ResourceId::new("runtime");
+    let trivias = ast.with_mut(|fields| std::mem::take(fields.trivias));
```
hyf0, line 115: Probably, it's better to use it directly rather take out `trivias` field, it might be needed in the future. But this doesn't block merging PR.

hyf0 (approved):
Good work.

PR #1226, opened 2024-05-23
test(rust/esbuild/default): enable re_export_default_external.

Cecil0o0:


crates/rolldown/tests/esbuild/default/re_export_default_external_common_js/artifacts.snap
```diff
@@ -0,0 +1,31 @@
...
+## entry_js.cjs
+
+```js
+const { __toESM } = require("./$runtime$.cjs");
+const { default: foo } = __toESM(require("foo"));
+const { default: bar } = __toESM(require("bar"));
+
+exports.bar = bar;
+exports.foo = foo;
+```
```
IWANABETHATGUY, line 17: Please recover this test case, 
the esbuild snapshot is
```js
================================================================================
TestReExportDefaultExternalCommonJS
---------- /out.js ----------
// entry.js
var entry_exports = {};
__export(entry_exports, {
  bar: () => import_bar.default,
  foo: () => import_foo.default
});
module.exports = __toCommonJS(entry_exports);
var import_foo = __toESM(require("foo"));

// bar.js
var import_bar = __toESM(require("bar"));

```

crates/rolldown/tests/esbuild/default/re_export_default_external_es6/artifacts.snap
```diff
@@ -0,0 +1,29 @@
...
+# Assets
+
+## entry_js.mjs
+
+```js
+import { default as foo } from "foo";
+import { default as bar } from "bar";
+
+export { bar, foo };
+```
```
IWANABETHATGUY, line 15: Could you solve the `unresolved` error? esbuild snapshot does not have these errors.

Cecil0o0: Ok, will be resolved tomorrow

crates/rolldown/tests/esbuild/default/re_export_default_external_common_js/artifacts.snap
Cecil0o0: could I ask, "recover this test case" means not to enable this case until snapshot be same with esbuild's?

hyf0: No need to revert. I will take a note on this case and fix it later. It's fine we merge them first. But still could you solve the `UNRESOLVED` warning.

Cecil0o0: Got it, I solved them.

crates/rolldown/tests/esbuild/default/re_export_default_external_common_js/_config.json
```diff
@@ -0,0 +1,13 @@
+{
+  "expectExecuted": false,
```
IWANABETHATGUY, line 2: Why do you mark this test case as none executed?

hyf0 (approved):
Thanks

PR #1228, opened 2024-05-24
test(esbuild/packgaejson): test_package_json_syntax_error_trailing_comma

zhusjfaker:
### Description

in mr https://github.com/rolldown/rolldown/pull/1224#event-12918095610

base the Based on the discussion results of the above #1224 

i change the code of the fixtures again

at this time.  i have impl anyhow::error  for rolldown:error   -> buildevent

and  use config fields "expectError"

please review it again  thank you every one


crates/rolldown_error/src/event_kind.rs
```diff
@@ -14,6 +14,7 @@ pub enum EventKind {
   // TODO remove following kinds
   NapiError,
   IoError,
+  AnyhowError,
```
hyf0, line 17: We use anyhow to represent program errors and `BuildError` as errors related to bundling.

So anyhow Error should not be encapsulated by BuildError.

crates/rolldown/tests/esbuild/packagejson/test_package_json_syntax_error_trailing_comma/artifacts.snap
```diff
@@ -0,0 +1,37 @@
...
+source: crates/rolldown/tests/common/case.rs
+expression: content
+input_file: crates/rolldown/tests/esbuild/packagejson/test_package_json_syntax_error_trailing_comma
+---
+# Errors
+
+## ANYHOW_Error
+
+```text
+[ANYHOW_Error] Error: Anyhow error: Unexpectedly failed to resolve dependencies of /src/entry.js. Got errors [
```
hyf0, line 11: Passed tests should not return any `anyhow::Error`.

> Panics/`anyhow::Error` are usually program errors and can't not be handled, so it's always unexpected. If some errors are wrongly considered as programs errors, we should fix the behavior rather than thinking programs errors are expected. https://github.com/rolldown/rolldown/pull/1224#issuecomment-2126882955

hyf0 (changes requested):

PR #1229, opened 2024-05-24
feat: support hook module side effects

underfin:
### Description

The pr intend to support `resolveId/load/transfrom` return `moduleSideEffects`, the filed see detail at https://rollupjs.org/plugin-development/#resolveid.

### DIFF
--- a/crates/rolldown/src/module_loader/normal_module_task.rs
+++ b/crates/rolldown/src/module_loader/normal_module_task.rs
@@ -4,9 +4,9 @@ use anyhow::Result;
 use futures::future::join_all;
 use oxc_index::IndexVec;
 use rolldown_common::{
-  side_effects::DeterminedSideEffects, AstScope, ImportRecordId, ModuleType, NormalModule,
-  NormalModuleId, PackageJson, RawImportRecord, ResolvedPath, ResolvedRequestInfo, ResourceId,
-  SymbolRef,
+  side_effects::{DeterminedSideEffects, HookSideEffects},
+  AstScope, ImportRecordId, ModuleType, NormalModule, NormalModuleId, PackageJson, RawImportRecord,
+  ResolvedPath, ResolvedRequestInfo, ResourceId, SymbolRef,
 };
 use rolldown_error::BuildError;
 use rolldown_oxc_utils::OxcAst;
@@ -33,6 +33,7 @@ pub struct NormalModuleTask {
   module_type: ModuleType,
   errors: Vec<BuildError>,
   is_user_defined_entry: bool,
+  side_effects: Option<HookSideEffects>,
 }
 
 impl NormalModuleTask {
@@ -43,6 +44,7 @@ impl NormalModuleTask {
     module_type: ModuleType,
     is_user_defined_entry: bool,
     package_json: Option<Arc<PackageJson>>,
+    side_effects: Option<HookSideEffects>,
   ) -> Self {
     Self {
       ctx,
@@ -52,6 +54,7 @@ impl NormalModuleTask {
       errors: vec![],
       is_user_defined_entry,
       package_json,
+      side_effects,
     }
   }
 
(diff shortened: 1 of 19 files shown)


crates/rolldown/src/module_loader/normal_module_task.rs
```diff
@@ -127,6 +142,26 @@ impl NormalModuleTask {
...
+      .unwrap_or_else(|| {
+        let analyzed_side_effects = stmt_infos.iter().any(|stmt_info| stmt_info.side_effect);
+        DeterminedSideEffects::Analyzed(analyzed_side_effects)
+      });
+
+    let side_effects = match hook_side_effects {
+      Some(side_effects) => match side_effects {
+        HookSideEffects::True => default_side_effects,
+        HookSideEffects::False => DeterminedSideEffects::PackageJson(false),
+        HookSideEffects::NoTreeshake => unimplemented!(),
```
underfin, line 161: @IWANABETHATGUY Here need to disable treeshake for the module, look like it need to implement it.

hyf0: I looked into the feature 'no-treeshake'. If it's only for compatibility and no solid reasons, there's no need to implement the feature for adding extra complexity.

hyf0: I searched around and found the only real-world use case if for `lottie-web` which used `eval`. https://github.com/rollup/rollup/issues/2963

crates/rolldown/src/module_loader/normal_module_task.rs
```diff
@@ -127,6 +142,26 @@ impl NormalModuleTask {
...
+      })
+      .unwrap_or_else(|| {
+        let analyzed_side_effects = stmt_infos.iter().any(|stmt_info| stmt_info.side_effect);
+        DeterminedSideEffects::Analyzed(analyzed_side_effects)
+      });
+
+    let side_effects = match hook_side_effects {
+      Some(side_effects) => match side_effects {
+        HookSideEffects::True => default_side_effects,
+        HookSideEffects::False => DeterminedSideEffects::PackageJson(false),
```
hyf0, line 160: Guess it's more suitable to treat returned `side_effects: true/false` as `DeterminedSideEffects::PackageJson` rather than `DeterminedSideEffects::Analyzed`.

Maybe explicitly returned `side_effects: true` should not be overwritten by later `determine_side_effects`.

@IWANABETHATGUY cc need your confirmation.

---

Personally, I think that maybe `DeterminedSideEffects::PackageJson` should  be renamed to `DeterminedSideEffects:UserDefined`. But this is just a thought, no need to worry about it in this PR.

crates/rolldown/src/module_loader/normal_module_task.rs
hyf0: It's not hard to implement this feature, we just need to mark all `StmtInfo` with `side_effect: true`. But still, if it is a overengineering feature, it's not worth for us to take actions on it.

packages/rolldown/src/utils/transform-side-effects.ts
```diff
@@ -0,0 +1,30 @@
+import { ModuleSideEffects } from '../plugin'
+
+// TODO The typing should import from binding, but const enum is disabled by `isolatedModules`.
```
hyf0, line 3: @Brooooooklyn  It there a way to generate non-const enum in NAPI-RS for enum?

crates/rolldown/src/module_loader/normal_module_task.rs
underfin: Yeah.It not be hurry to implement it. Here common case is `false` or `undefined`

hyf0: Saw that `unimplemented!` macro, let's keep it for a while.

hyf0:
I'm all fine with this PR except some comments, which need IWANABETHATGUY's confirmation. Assigned to him.

crates/rolldown/src/module_loader/normal_module_task.rs
```diff
@@ -127,6 +142,26 @@ impl NormalModuleTask {
         let analyzed_side_effects = stmt_infos.iter().any(|stmt_info| stmt_info.side_effect);
         DeterminedSideEffects::Analyzed(analyzed_side_effects)
       });
```
IWANABETHATGUY, line 144: looks like this is unused

crates/rolldown/src/module_loader/normal_module_task.rs
```diff
@@ -127,6 +142,26 @@ impl NormalModuleTask {
...
+
+    let default_side_effects = self
```
IWANABETHATGUY, line 146: What is the priority between `hook_side_effects`, `packagejson` and `analyzed`, could you elaborate it?

crates/rolldown/src/module_loader/normal_module_task.rs
IWANABETHATGUY: If no other places need to worry about the source of side effects, I am ok with it.

crates/rolldown/src/module_loader/normal_module_task.rs
underfin: The `hook_side_effects` is set to the module by user manual, so the priority is high. After it should be `packagejson`, and then `analyzed`. You can see here https://github.com/rollup/rollup/blob/master/src/Module.ts#L694 and https://github.com/rollup/plugins/blob/master/packages/node-resolve/src/index.js#L254

IWANABETHATGUY: got it

crates/rolldown/src/module_loader/normal_module_task.rs
underfin: It is the default behavior for side effects, it used at next lines, please read it.

IWANABETHATGUY: image
Am I missing something? the code was duplicated twice

IWANABETHATGUY (approved):

packages/rolldown/src/utils/transform-side-effects.ts
Brooooooklyn: `napi build --no-const-enum`

PR #1230, opened 2024-05-25
chore(node/rolldown): import type.

Cecil0o0:


packages/rolldown/src/cli/utils.ts
```diff
@@ -1,7 +1,7 @@
 import { pathToFileURL } from 'node:url'
 import nodePath from 'node:path'
 import { createConsola } from 'consola'
-import { RolldownConfigExport } from '../types/rolldown-config-export.js'
+import type { RolldownConfigExport } from '../types/rolldown-config-export.js'
```
IWANABETHATGUY: Is `import type` from a js module meaningful?

IWANABETHATGUY:
Can you split the PR into two parts?
one for solving the circular dependency, and another for changing `import` to `import type`?

IWANABETHATGUY:
Could you share the tools that check **circular dependency**, so I could verify if there are more **circular dependency** in our codebase?

packages/rolldown/src/cli/utils.ts
Cecil0o0: It's not. At that time, I didn't see this module specifier which ends with `.js`, and now I find this module actually is a ts module, aha.

would you think remove suffix, be a appropriate way for making meaningful?

Cecil0o0:
> Can you split the PR into two parts? one for solving the circular dependency, and another for changing `import` to `import type`?

of course could, I'd like to do this.

Cecil0o0:
> Could you share the tools that check **circular dependency**, so I could verify if there are more **circular dependency** in our codebase?

I would try to do this in another PR. Maybe a change serves as a supplement, because rolldown has done it natively, `packages/rolldown` had been checked natively by rolldown ourself. #1231

IWANABETHATGUY (approved):

PR #1232, opened 2024-05-25
fix(node/rolldown): CIRCULAR_DEPENDENCY warnings.

### LINKED ISSUE #1231: could `CIRCULAR_DEPENDENCY` be an issue which should be solved in current stage?
https://github.com/rolldown/rolldown/blob/c476ba10a86d0e9d579b9c62587e3c9ef760f487/packages/rolldown/src/utils/index.ts#L10-L32

code block above may involve `CIRCULAR_DEPENDENCY` warnings below:

image

could we move them into a individual module X, and then export them all from X at index module, be a appropriate way?

Cecil0o0:
### Description

After `just build` done finished, there were six `CIRCULAR_DEPENDENCY` warnings on the terminal.

image

I solved two of them in this PR.

And Then This PR contains a way for checking the circular dependency manually, I'm not sure whether it's a appropriate way to do the check?

related #1231


packages/rolldown/package.json
```diff
@@ -65,7 +65,17 @@
     "# Scrips for checking #": "_",
     "test": "vitest run --reporter verbose --hideSkippedTests",
     "test:update": "vitest run -u",
-    "type-check": "tsc"
+    "type-check": "tsc",
+    "circular-dep-check": "madge --circular src/index.ts src/cli/index.ts src/parallel-plugin.ts src/parallel-plugin-worker.ts"
```
IWANABETHATGUY: I am not sure if we should introduce another tool to do this, @hyf0

hyf0: Not needed. We have rolldown already as the check tools.

hyf0 (approved):

hyf0:
Thanks

PR #1237, opened 2024-05-26
refactor(rust): clarify the priority of side effect from different source

IWANABETHATGUY:
### Description
1. remove redundant side effect calc https://github.com/rolldown/rolldown/pull/1237/files#diff-c6103d95ff0ac45751c32c94a8f82ebe0ce206be44de7793137c3051885b0304L135-L144
2. lazy check side effect

### DIFF
--- a/crates/rolldown/src/module_loader/normal_module_task.rs
+++ b/crates/rolldown/src/module_loader/normal_module_task.rs
@@ -132,11 +132,11 @@ impl NormalModuleTask {
     let resource_id = ResourceId::new(Arc::clone(&self.resolved_path.path));
     let stable_resource_id = resource_id.stabilize(&self.ctx.input_options.cwd);
 
-    // The side effects pripority is:
+    // The side effects priority is:
     // 1. Hook side effects
     // 2. Package.json side effects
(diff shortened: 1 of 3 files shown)


crates/rolldown/src/module_loader/normal_module_task.rs
```diff
@@ -132,35 +132,31 @@ impl NormalModuleTask {
...
+          let analyzed_side_effects = stmt_infos.iter().any(|stmt_info| stmt_info.side_effect);
+          DeterminedSideEffects::Analyzed(analyzed_side_effects)
+        })
+    };
 
     let side_effects = match hook_side_effects {
       Some(side_effects) => match side_effects {
-        HookSideEffects::True => default_side_effects,
-        HookSideEffects::False => DeterminedSideEffects::PackageJson(false),
+        HookSideEffects::True => DeterminedSideEffects::UserDefined(true),
```
IWANABETHATGUY, line 155: Since side effect from `hook` has higher priority, we should apply the value directly.

hyf0 (approved):

PR #1211, opened 2024-05-22
feat: bootstrap Rolldown with Rolldown

hyf0:
### Description

* The `rolldown` package published to npm is built with Rolldown
* TypeScript files are transpiled by esbuild via a plugin

PR #1239, opened 2024-05-26
chore(repo): eliminate unused code, dep, package script.

Cecil0o0:
### Description

this PR contains three items:

1. After #1211, dependency `unbuild` hasn't been imported by any module.
2. A `_assert` private function which declared in a ts module hasn't been referenced by others.
3. There is no `zod` declaration at `external` field in `rolldown.config.mjs`, so `zod` would be as `devDependencies`, more appropriate than `dependencies`. It seems not to used by user project even if user install this dependency.


packages/rolldown/build.config.ts
hyf0, line 1: We gonna keep this file for a while to make `unbuild` as a fallback.

packages/rolldown/src/rolldown-build.ts
```diff
@@ -43,7 +42,3 @@ export class RolldownBuild {
     await this.#stopWorkers?.()
   }
 }
-
-function _assert() {
```
hyf0, line 47: This is used to ensure `RolldownBuild` has properies `generate` and `write` by `tsc`.

Cecil0o0: Thanks for the reply, I didn't understand it before.

PR #1251, opened 2024-05-27
feat: support throw some build_errors in  scan stage and link stage

zhusjfaker:
### Description

in the past code implementation

in the scan stage 

everyone will see the logic at https://github.com/rolldown/rolldown/blob/main/crates/rolldown/src/module_loader/normal_module_task.rs#L349

most of build error will be into as anyhow::Error and throw them in the "try_spawn_new_task"
https://github.com/rolldown/rolldown/blob/main/crates/rolldown/src/module_loader/module_loader.rs#L122

in the code stack 
“fetch_all_modules”
https://github.com/rolldown/rolldown/blob/main/crates/rolldown/src/stages/scan_stage.rs#L71

but  in the esbuild test case 

https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_packagejson_test.go#L65

they are need rolldown-reoslver:ResovleError into be a BuildError and  throw them at the function bundleup in the same time .

so i change the code logic please master review it 

Thank you for your hard work and have a nice day.

### DIFF
--- a/crates/rolldown/src/bundler.rs
+++ b/crates/rolldown/src/bundler.rs
@@ -95,34 +95,41 @@ impl Bundler {
     ret
   }
 
-  async fn try_build(&mut self) -> Result<LinkStageOutput> {
-    let build_info = self.scan().await?;
-
-    let link_stage = LinkStage::new(build_info, &self.options);
-    Ok(link_stage.link())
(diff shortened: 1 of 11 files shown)


zhusjfaker:
Although, of course, the code has already been passed.

.test_package_json_syntax_error_comment

.test_package_json_syntax_error_trailing_comma

I just haven't migrated the code for those cases yet.

IWANABETHATGUY:
too stale

PR #1284, opened 2024-05-30
feat(cjs): add `"use strict"` if possible

hyf0:
### Description

Implement what https://github.com/rolldown/rolldown/pull/1279 said and fixes #1277.

PR #1286, opened 2024-05-30
test(use_strict): rolldown should allow to parse non-strict code

hyf0:

### DIFF
--- a/crates/rolldown/tests/fixtures/misc/use_strict/allow_parse_non_strict_code_in_cjs_format/_config.json
+++ b/crates/rolldown/tests/fixtures/misc/use_strict/allow_parse_non_strict_code_in_cjs_format/_config.json
@@ -1,0 +1,6 @@
+{
+  "config": {
+    "format": "cjs"
+  },
+  "expectExecuted": false
+}
--- a/crates/rolldown/tests/fixtures/misc/use_strict/allow_parse_non_strict_code_in_cjs_format/artifacts.snap
+++ b/crates/rolldown/tests/fixtures/misc/use_strict/allow_parse_non_strict_code_in_cjs_format/artifacts.snap
@@ -1,0 +1,22 @@
+---
+source: crates/rolldown/tests/common/case.rs
+expression: content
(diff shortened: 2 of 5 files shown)


hyf0:
* **#1286** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1286?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1285** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1285?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1284** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1284?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1283** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1283?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

hyf0:
Rolldown parse code in with `assert!(sourceType.strict())` by default and plan to

> With `format: 'cjs'`, rolldown will first parse your code in strict mode, and will re-parse it in non-strict mode if it throws an error in the first parsing. [link](https://github.com/rolldown/rolldown/blob/92e145e64064f44079952bfde159eeb9bd42170d/web/docs/guide/in-depth/use-strict.md?plain=1#L40)

I notice that oxc won't throw errors for[ invalid code](https://github.com/rolldown/rolldown/blob/c50421c53d5816c7cb71a0db3bddcf6a6218e22e/crates/rolldown/tests/fixtures/misc/use_strict/allow_parse_non_strict_code_in_cjs_format/cjs.js) in strict mode parsing. The validation is done by the `Semantic#with_check_syntax_error`.

So there's no need to do re-parsing, we just need to ignore the errors related to strict mode that emitted by `Semantic` for non-strict module. Or we could choose to disable `Semantic#with_check_syntax_error` for non-strict modules.

No actions, just FYI. @Boshen cc

hyf0:
To make commonjs code benefit from "use strict",  we could even mark non-strict commonjs module that doesn't have errors which are related to `use strict` as strict. Just not sure if the check is 100% reliable.

PR #1287, opened 2024-05-30
fix: add return source logic when file not downloaded and add loader …

zhusjfaker:
…suite

### Description

in the old logic 

when the script exec " return fs.readFileSync(sourcePath).toString()"  and  golang file not downloaded and happend exception 

the logic will try to download file but not return the golang code source 

i fix it and return the expect code 

please review it thank you master and have nice day

### DIFF
--- a/scripts/misc/gen-esbuild-test.mjs
+++ b/scripts/misc/gen-esbuild-test.mjs
@@ -128,6 +128,13 @@ const suites = /** @type {const} */ ({
       'lower_using_unsupported_using_and_async',
     ],
   },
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY (approved):

zhusjfaker:
sorry i dont have right to re_run workflow to check the change

zhusjfaker:
@IWANABETHATGUY  please help me re_run this workflow ,i have no right to reboot it

IWANABETHATGUY (approved):

PR #1289, opened 2024-05-30
feat: mark `PluginContext#parse` as unsupported

underfin:


packages/rolldown/src/plugin/plugin-context.ts
```diff
@@ -5,13 +5,15 @@ import type { NormalizedInputOptions } from '../options/normalized-input-options
...
 import { error, logPluginError } from '../log/logs'
+import { unsupported } from '../utils'
 
 export class PluginContext {
   debug: LoggingFunction
   info: LoggingFunction
   warn: LoggingFunction
   error: (error: RollupError | string) => never
   resolve: BindingPluginContext['resolve']
+  parse: (input: string, options?: any) => any
```
hyf0, line 29: If we don't want to support `this.parse`, it should not exist on `PluginContext` type.

Probably do something like

```rust
// @ts-expect-error
PluginContext.prototype.parse = ...
```

hyf0 (approved):

Issue #1266, opened 2024-05-29
`treeshake.moduleSideEffects: false` not yet supported


hyf0:

hyf0:
https://rollupjs.org/configuration-options/#treeshake-modulesideeffects

Guess it's enough for now that we only accept boolean for `treeshake.moduleSideEffects` option.

PR #1293, opened 2024-05-31
feat(plugin): support return `sideEffect: 'no-treeshake'`

IWANABETHATGUY:
### Description
1. Part of ~https://github.com/rolldown/rolldown/issues/1266~

### DIFF
--- a/crates/rolldown/src/stages/link_stage/tree_shaking.rs
+++ b/crates/rolldown/src/stages/link_stage/tree_shaking.rs
@@ -32,7 +32,7 @@ fn include_module(ctx: &mut Context, module: &NormalModule) {
   if ctx.tree_shaking {
     let forced_no_treeshake = matches!(module.side_effects, DeterminedSideEffects::NoTreeshake);
     module.stmt_infos.iter_enumerated().for_each(|(stmt_info_id, stmt_info)| {
-      // Skip the first statement, which is the namespace object. It should included only if it is used no matter
+      // Skip the first statement, which is the namespace object. It should be included only if it is used no matter
       // tree shaking is enabled or not.
       if stmt_info_id.index() == 0 {
         return;
@@ -43,7 +43,7 @@ fn include_module(ctx: &mut Context, module: &NormalModule) {
     });
   } else {
     module.stmt_infos.iter_enumerated().for_each(|(stmt_info_id, _stmt_info)| {
-      // Skip the first statement, which is the namespace object. It should included only if it is used no matter
+      // Skip the first statement, which is the namespace object. It should be included only if it is used no matter
       // tree shaking is enabled or not.
       if stmt_info_id.index() == 0 {
         return;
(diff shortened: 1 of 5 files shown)


hyf0:
#1266 is about https://rollupjs.org/configuration-options/#treeshake-modulesideeffects. I think you misunderstand it.

packages/rolldown/src/plugin/bindingify-build-hooks.ts
```diff
@@ -149,28 +150,18 @@ export function bindingifyTransform(
...
-    }
-
-    if (ret.moduleSideEffects !== null) {
-      // @ts-ignore TODO The typing should import from binding
-      result.sideEffects = bindingifySideEffects(ret.moduleSideEffects)
-    }
-
-    return result
+      sideEffects: bindingifySideEffects(ret.moduleSideEffects),
+    } as unknown as BindingHookLoadOutput
```
hyf0: What's about this `as unknown`?  Code should be valid without it.

IWANABETHATGUY:
> #1266 is about https://rollupjs.org/configuration-options/#treeshake-modulesideeffects. I think you misunderstand it.

I don't close the original issue, it is good to merge.

hyf0:
> > #1266 is about https://rollupjs.org/configuration-options/#treeshake-modulesideeffects. I think you misunderstand it.
> 
> I don't close the original pull request, it is good to merge.

Wait a minue. I'm fixing the `as unknown`.

crates/rolldown/src/stages/link_stage/tree_shaking.rs
```diff
@@ -23,16 +23,23 @@ fn include_module(ctx: &mut Context, module: &NormalModule) {
...
+    // runtime module has no side effects and it's statements should be included
+    // by other modules's references.
+    return;
+  }
+
+  if ctx.tree_shaking {
+    let forced_no_treeshake = matches!(module.side_effects, DeterminedSideEffects::NoTreeshake);
     module.stmt_infos.iter_enumerated().for_each(|(stmt_info_id, stmt_info)| {
-      if stmt_info.side_effect {
+      if stmt_info.side_effect && !forced_no_treeshake {
```
IWANABETHATGUY: Any good of this? you need to calculate every time, we only calculate once previously.

hyf0: I want make it more clear to see the different between enabling treeshaking or not. But yes bad for performance. Let's do a better way.

hyf0 (approved):
@IWANABETHATGUY Done. Leave to you if there are still things need to be changed.

hyf0:
@IWANABETHATGUY Sorry to overwrite your commit. I will do MR next time.

hyf0:
I need to change some code in `normal_module_task.rs`. To prevent conflicts, let me merge the PR first. If there are still things need to be changed, let's do them in another PR.

IWANABETHATGUY:
> @IWANABETHATGUY Sorry to overwrite your commit. I will do MR next time.

It is alright,  I pushed the patch before I saw your comments, :rofl:

PR #1302, opened 2024-05-31
feat: support text loader

### LINKED ISSUE #1300: Support text loader
## Related

- https://esbuild.github.io/content-types/#text

Cecil0o0:
### Description

relative issue #1300 

esbuild's snapshot:

https://github.com/evanw/esbuild/blob/67cbf87a4909d87a902ca8c3b69ab5330defab0a/internal/bundler_tests/snapshots/snapshots_loader.txt#L942-L957


crates/rolldown/src/utils/parse_to_ast.rs
```diff
@@ -50,6 +52,10 @@ pub fn parse_to_ast(
     ParseType::Jsx => todo!(),
     ParseType::Ts => todo!(),
     ParseType::Tsx => todo!(),
+    ParseType::Text => OxcCompiler::parse(
```
hyf0: You should not touch the `ParserType`. See how rolldown deal with `json`.

hyf0 (changes requested):

crates/rolldown/src/utils/parse_to_ast.rs
Cecil0o0: has reverted and changed~

hyf0 (dismissed):
Thanks

hyf0:
CI failed.

Cecil0o0:
already fixed

crates/rolldown/src/utils/parse_to_ast.rs
```diff
@@ -42,6 +42,7 @@ pub fn parse_to_ast(
   let (source, parsed_type) = match loader {
     Loader::Js => (source, ParseType::Js),
     Loader::Json => (json_to_esm(&source)?.into(), ParseType::Js),
+    Loader::Text => (("export default \"".to_string() + &source + "\";").into(), ParseType::Js),
```
hyf0: I believe this will met problems while the content contains `"`. Could you solve this one and add related tests in `fixtures/loaders/txt/escape_quote`.

hyf0 (changes requested):

crates/rolldown/src/utils/parse_to_ast.rs
hyf0: 1. add a function `text_to_esm` in `rolldown/utils/text_to_esm.rs`
2. Convert the source to Value::String using `serde_json`
3. serilized the value use serde_json and you should be able to escape the quote correctly.

Not sure if this is in best performant. @IWANABETHATGUY might have some thoughts.

Cecil0o0: I have pushed a new commit for adding a fixture and declaring a `text_to_esm` function. I looked at esbuild's implementation, esbuild seems to use `base64` encoding for text loader to realize `escape_quote`.

https://github.com/evanw/esbuild/blob/67cbf87a4909d87a902ca8c3b69ab5330defab0a/internal/bundler/bundler.go#L269-L280

crates/rolldown/src/utils/text_to_esm.rs
```diff
@@ -0,0 +1,5 @@
+use anyhow::Ok;
+
+pub fn text_to_esm(txt: &str) -> anyhow::Result<String> {
```
IWANABETHATGUY, line 1: the `Result` is redundant.

crates/rolldown/src/utils/parse_to_ast.rs
hyf0: It seems that esbuilds uses codegen of itself to escape the quote. Using oxc to do this is a little bit complex. I probably stay with serde_json.

crates/rolldown/tests/fixtures/loaders/txt/escape_quote/x.txt
```diff
@@ -0,0 +1 @@
+"
```
hyf0:
```suggestion
'"`
```

hyf0 (changes requested):

Cecil0o0:
test case passed but not correct, try to fix it.

crates/rolldown/src/utils/parse_to_ast.rs
Cecil0o0: serde_json is quite useful~learned a new thing.

hyf0 (approved):
Thanks

PR #1305, opened 2024-05-31
chore(tree-shaking):  tweak determine module side effects

IWANABETHATGUY:

### DIFF
--- a/crates/rolldown/src/module_loader/normal_module_task.rs
+++ b/crates/rolldown/src/module_loader/normal_module_task.rs
@@ -158,7 +158,7 @@ impl NormalModuleTask {
 
     let side_effects = match hook_side_effects {
       Some(side_effects) => match side_effects {
-        HookSideEffects::True => DeterminedSideEffects::UserDefined(true),
+        HookSideEffects::True => lazy_check_side_effects(),
         HookSideEffects::False => DeterminedSideEffects::UserDefined(false),
         HookSideEffects::NoTreeshake => DeterminedSideEffects::NoTreeshake,
       },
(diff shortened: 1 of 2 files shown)


crates/rolldown/src/stages/link_stage/tree_shaking.rs
```diff
@@ -40,18 +40,14 @@ fn include_module(ctx: &mut Context, module: &NormalModule) {
...
-      forcefully_include_all_statements(ctx, module);
-    } else {
-      module.stmt_infos.iter_enumerated().for_each(|(stmt_info_id, stmt_info)| {
-        // No need to handle the first statement specially, which is the namespace object, because it doesn't have side effects and will only be included if it is used.
-        if stmt_info.side_effect {
-          include_statement(ctx, module, stmt_info_id);
-        }
-      });
-    }
+  let forced_no_treeshake = matches!(module.side_effects, DeterminedSideEffects::NoTreeshake);
```
IWANABETHATGUY, line 43: The semantics should be the same, and conditional checking looks simpler.

hyf0 (approved):

crates/rolldown/src/module_loader/normal_module_task.rs
```diff
@@ -158,7 +158,7 @@ impl NormalModuleTask {
 
     let side_effects = match hook_side_effects {
       Some(side_effects) => match side_effects {
-        HookSideEffects::True => DeterminedSideEffects::UserDefined(true),
+        HookSideEffects::True => lazy_check_side_effects(),
```
IWANABETHATGUY, line 161: > If false is returned for moduleSideEffects and no other module imports anything from this module, then this module will not be included even if the module would have side effects.

> If true is returned, Rollup will use its default algorithm to include all statements in the module that have side effects (such as modifying a global or exported variable).

hyf0: Will the plain import `import './foo.js'` be reserved by rollup/esbuild if `foo.js` is `HookSideEffects::True` and it actually doesn't have side effect.

PR #1325, opened 2024-06-03
ci: add missing type definitions

sxzz:
### Description

d.ts files are missing in the latest version.

See https://publint.dev/rolldown@0.10.3-snapshot-3681314-20240603002645


hyf0 (approved):
Thanks

PR #1326, opened 2024-06-03
chore: remove unnecessary clippy allow directives

lzt1008:
### Description

This PR addresses the removal of unnecessary clippy::allow directives.

Changes are just remove the `#[allow(clippy::...)]` and unnecessary cast.

### DIFF
--- a/crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
+++ b/crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
@@ -17,7 +17,7 @@ use crate::utils::call_expression_ext::CallExpressionExt;
 use super::ScopeHoistingFinalizer;
 
 impl<'me, 'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'me, 'ast> {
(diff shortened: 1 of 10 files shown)


hyf0:
Some changes changed the semantic of the original code. So this PR would be not accepted. Please split the change that removing unnecessary clippy directives to another PR.

lzt1008:
> Some changes changed the semantic of the original code. So this PR would be not accepted. Please split the change that removing unnecessary clippy directives to another PR.

semantic changes are removed, now this PR only contains removal of unnecessary clippy allow and a unnecessary cast.

Boshen:
How did you discover these problems? We may be able to enforce a rule for the future.

hyf0 (approved):
Thanks

lzt1008:
> How did you discover these problems? We may be able to enforce a rule for the future.

I just discovered these unnecessary clippy::allow directives by accident and manually find it all. Currently, there isn't an automated rule in place for identifying these unnecessary Clippy allows(see rust-lang/rust-clippy/issues/3122). Hopefully, we'll see this as a clippy rule implemented soon.

PR #1336, opened 2024-06-04
chore(package/rolldown): stop inline dependencies

hyf0:
### Description

Currently we use tsc to generate types and tsc won't inline types of dependencies. We need to add these dependencies to `dependencies `, so the types of users could be correct.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "license": "MIT",
   "devDependencies": {
     "rolldown": "workspace:*",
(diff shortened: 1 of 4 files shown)


hyf0:
* **#1336** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1336?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #1342, opened 2024-06-04
docs: describle `Module Type`

hyf0:
### Description

Part of #1341.

### DIFF
--- a/web/docs/guide/in-depth/module-type.md
+++ b/web/docs/guide/in-depth/module-type.md
@@ -1,8 +1,8 @@
 # Module Type
 
-As a web bundler, rolldown not only support bundling JavaScript files, but also support bundling other types of files, such as CSS(in the future), JSON, and so on.
+As a web bundler, rolldown not only accept JavaScript files as input items, but also support other types of input items, such as CSS(in the future), JSON, dataurl, and so on.
 
-Therefore, if a file could be recognized by rolldown without any extra plugins, it means that the type of this file is treated as a as first-class citizen by rolldown.
+Therefore, if an input item could be recognized by rolldown without any extra plugins, it means that the type of this input item is treated as a first-class citizen by rolldown.
 
 We call these fist-class types as `Module Type`s.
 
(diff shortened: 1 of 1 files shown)


hyf0:
* **#1342** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1342?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

web/docs/guide/in-depth/module-type.md
```diff
@@ -0,0 +1,36 @@
...
+
+Rolldown uses the extension of the file to determine what the `Module Type` is, but sometimes it's not enough. For example, a file contains JSON data, but the extension is `.data`. In this case, rolldown can't recognize it as a JSON file, because the extension is not `.json`.
+
+In this case, users need to tell rolldown what the `Module Type` is for `.data` extension by configuring `moduleTypes` field in `rolldown.config.mjs`.
+
+```js
+// rolldown.config.mjs
+export default {
+  moduleTypes: {
+    '.data': 'json',
```
magic-akari, line 21: There is only the usage of extension. 
In some other use cases, there is also searchQuery, such as:

```js
import foo from "./foo.data?url";
```

I think webpack has done a great job.
https://webpack.js.org/configuration/module/#ruleresource

hyf0: Yeah. Webpack provides a very high customized filter system, it's indeed powerful but sometimes could be complex for newbies. But that's not the key reason we implemented `moduleTypes` this way.

`moduleTypes` is a port of esbuild's loader feature, but the `loader` concept is kind of easily-misunderstanding with webpack's loader concept and the `Plugin#transform` hook. So we are try to giving an easier understandable concept for esbuild's loader feature and the API pattern is derived from esbuild. The API pattern is simple and works out for most of cases.

For high customized filter, since `ModuleType` could be speicified in `load` and `transform` hook. So we might have third-party/official plugin to implemented a filter system like webpack.


```js
import moduleTypeFilter from '...'

export default {
   plugins: [moduleTypeFilter([
     {
       query: /url/,
       ext: '.data',
       loader: 'url',
     },
     {
       ext: '.data',
       loader: 'json',
     }
   ])]
}
```

Issue #1301, opened 2024-05-31
Support binary loader


hyf0:
## Related

- https://esbuild.github.io/content-types/#binary
- https://github.com/rolldown/rolldown/blob/e88c85d7cf6b60ee04a9a901c8a7c033c4d7ad5d/crates/rolldown/src/utils/parse_to_ast.rs#L26-L56
- https://github.com/rolldown/rolldown/blob/e88c85d7cf6b60ee04a9a901c8a7c033c4d7ad5d/crates/rolldown/src/utils/normalize_options.rs#L15-L40

hyf0:
WARNING⚠️

Be aware this is a difficult task. We would very much welcome contributions, but please be aware that Rolldown's maintainers have limited time to provide support or feedback. Please assess your understanding of the problem, and ability to tackle it on your own without a great deal of outside help before diving in.


We are of course very happy to discuss and answer questions. But if you're fairly new to Rust, or unfamiliar with the subject of this issue, this one is probably not for you. please take a look at other [good first issues](https://github.com/rolldown/rolldown/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) and take the one you are familiar with if you're keen to get involved in Rolldown.

PR #1352, opened 2024-06-05
feat: support base64 loader

### LINKED ISSUE #1297: Support `base64` loader
## Related

- https://esbuild.github.io/content-types/#base64

ikkz:
### Description

resolves #1297 

This pr supports importing other files in base64 format, including text files and binary files

Maybe binary loader(#1301 ) can be implemented in a similar way? If so, I can also implement it in this pr


crates/rolldown/src/utils/normalize_options.rs
```diff
@@ -18,6 +18,7 @@ pub fn normalize_options(mut raw_options: crate::BundlerOptions) -> NormalizeOpt
       ("mjs".to_string(), ModuleType::Js),
       ("cjs".to_string(), ModuleType::Js),
       ("txt".to_string(), ModuleType::Text),
+      ("base64".to_string(), ModuleType::Base64),
```
hyf0: No need to add default extension for base64 loader.

crates/rolldown_loader_utils/src/text_to_esm.rs
hyf0, line 1: Unexpected changes. Could you explain what this change for

crates/rolldown_utils/src/base64.rs
```diff
@@ -3,3 +3,7 @@ use base64::{engine::general_purpose, Engine as _};
 pub fn to_url_safe_base64(input: impl AsRef<[u8]>) -> String {
   general_purpose::URL_SAFE_NO_PAD.encode(input)
 }
+
+pub fn to_standard_base64(input: impl AsRef<[u8]>) -> String {
+  general_purpose::STANDARD.encode(input)
```
hyf0, line 8: Could you verify that we are using the same standard as esbuild?

hyf0 (changes requested):

hyf0:
Great works! All looks fine except some parts I left comments for.

---

> Maybe binary loader(https://github.com/rolldown/rolldown/issues/1301 ) can be implemented in a similar way? If so, I can also implement it in this pr

Yeah. It's welcome and you could give it a try. If it works out, we'd love to merge the PR.

crates/rolldown_loader_utils/src/text_to_esm.rs
ikkz: I used this function in a previous draft, so I modified it for better readability, but the final version I submitted uses a separate `base64_to_esm`. If you feel that this optimization is unnecessary or irrelevant to this PR, I can remove it in the next revision.

hyf0: It's the most fastest way concat strings. Yeah. Just keep it as it is. https://github.com/hoodie/concatenation_benchmarks-rs

ikkz:
CI on Windows fails, maybe git converts`lf` to `crlf` when checking out, I will try to disable this conversion of related files in .gitattributes in the next commit

crates/rolldown_loader_utils/src/text_to_esm.rs
ikkz: ok, and I will modify `base64_to_esm` to this way.

crates/rolldown_utils/src/base64.rs
ikkz: Yes, the [base64 encoding method](https://github.com/evanw/esbuild/blob/67cbf87a4909d87a902ca8c3b69ab5330defab0a/internal/bundler/bundler.go#L284) of base64 loader in esbuild is the standard method in [RFC 4648](https://www.rfc-editor.org/rfc/rfc4648.html), and we are using the same standard according to the [doc](https://docs.rs/base64/latest/base64/#encoding-alphabet) of `base64` crate

ikkz:
@hyf0 This PR is ready to be merged if there are no other issues.
Since a lot of files have been modified in this PR, I will try to implement the binary loader in another PR 😀

hyf0 (approved):

PR #1364, opened 2024-06-07
feat!: support `jsx`, `ts` and `tsx` via oxc transform by default

Dunqing:
### Description

I'm not familiar with rolldown, where should I place the `transform`?


crates/rolldown_loader_utils/src/transformer.rs
```diff
@@ -0,0 +1,30 @@
+use std::mem;
+use std::{path::Path, sync::Arc};
+
+use oxc::span::SourceType;
+use oxc::transformer::{TransformOptions, Transformer};
+use rolldown_oxc_utils::{OxcAst, OxcCompiler};
+
+pub fn transform(path: &Path, source: Arc<str>, source_type: SourceType) -> anyhow::Result<OxcAst> {
```
hyf0: You could place this function in `creates/rolldown/utils`

crates/rolldown_loader_utils/src/transformer.rs
```diff
@@ -0,0 +1,30 @@
...
+    if source_type.is_jsx() {
+      transformer_options.react.jsx_plugin = true;
+    }
+
+    Transformer::new(
+      fields.allocator,
+      path,
+      source_type,
+      fields.source,
+      unsafe { mem::transmute_copy(&fields.trivias) },
```
Dunqing: I'm trying to delete the unsafe, but it seems to require changes to OxcAst

Boshen:
We need a feature toggle using an environment variable, turned off by default.

Dunqing:
> We need a feature toggle using an environment variable, turned off by default.

Let's discuss this topic next monday, and we also need to discuss how to configure `TransformerOptions` in `Rolldown`.

Dunqing:
Ready for review, But still wait `oxc` for release https://github.com/oxc-project/oxc/pull/3580

hyf0:
> Ready for review, But still wait `oxc` for release [oxc-project/oxc#3580](https://github.com/oxc-project/oxc/pull/3580)

It's ok for `OxcAst` to change `trivias: Trivias` to `trivias: Rc<Trivias>`, but this would cause `OxcAst` not send and sync anymore.

`Arc<Trivias>` is suitable for rolldown but introduce extra overhead on cases don't require parallelism. Would it be better to introduce a feature gate `parallel` and use `Lrc<T>` around the oxc codebase?  rustc use `Lrc` instead of using `Arc` and `Rc` directly. https://github.com/rust-lang/rust/blob/20ba13c38e5b8451b7b09e4009efc9010bfa3f2c/compiler/rustc_data_structures/src/sync.rs#L21 @Boshen cc

Cargo.toml
```diff
@@ -117,7 +117,7 @@ once_cell          = "1.19.0"
 oxc_resolver       = { version = "1.8.1", features = ["package_json_raw_json_api"] }
 rayon              = "1.10.0"
 regex              = "1.10.5"
-rustc-hash         = "1.1.0"
+rustc-hash         = { git = "https://github.com/rust-lang/rustc-hash" }                              # v1.2.0
```
Boshen, line 120: v1.2.0 was yanked.

crates/rolldown/src/utils/normalize_options.rs
```diff
@@ -17,6 +17,9 @@ pub fn normalize_options(mut raw_options: crate::BundlerOptions) -> NormalizeOpt
       ("js".to_string(), ModuleType::Js),
       ("mjs".to_string(), ModuleType::Js),
       ("cjs".to_string(), ModuleType::Js),
+      ("jsx".to_string(), ModuleType::Jsx),
+      ("ts".to_string(), ModuleType::Ts),
+      ("tsx".to_string(), ModuleType::Tsx),
       ("txt".to_string(), ModuleType::Text),
```
Boshen, line 25: "mts" and "cts" are missing 🤔

Boshen:
```
  --> crates/rolldown_plugin/src/transform_plugin_context.rs:39:5
   |
37 |     fn create_sourcemap(&self) -> SourceMap {
   |                                   --------- expected `SourceMap` because of return type
38 |       let magic_string = MagicString::new(self.original_code);
39 | /     magic_string.source_map(SourceMapOptions {
40 | |       hires: true,
41 | |       include_content: true,
42 | |       source: self.id.into(),
43 | |     })
   | |______^ expected `SourceMap`, found `oxc_sourcemap::sourcemap::SourceMap`
   |
   = note: `oxc_sourcemap::sourcemap::SourceMap` and `SourceMap` have similar names, but are actually distinct types
note: `oxc_sourcemap::sourcemap::SourceMap` is defined in crate `oxc_sourcemap`
  --> /Users/boshen/.cargo/registry/src/index.crates.io-6f17d22bba15001f/oxc_sourcemap-0.13.5/src/sourcemap.rs:11:1
```

I think we should move that crate into rolldown ...

hyf0:
> ```
> error[E0308]: mismatched types
>   --> crates/rolldown_plugin/src/transform_plugin_context.rs:39:5
>    |
> 37 |     fn create_sourcemap(&self) -> SourceMap {
>    |                                   --------- expected `SourceMap` because of return type
> 38 |       let magic_string = MagicString::new(self.original_code);
> 39 | /     magic_string.source_map(SourceMapOptions {
> 40 | |       hires: true,
> 41 | |       include_content: true,
> 42 | |       source: self.id.into(),
> 43 | |     })
>    | |______^ expected `SourceMap`, found `oxc_sourcemap::sourcemap::SourceMap`
>    |
>    = note: `oxc_sourcemap::sourcemap::SourceMap` and `SourceMap` have similar names, but are actually distinct types
> note: `oxc_sourcemap::sourcemap::SourceMap` is defined in crate `oxc_sourcemap`
>   --> /Users/boshen/.cargo/registry/src/index.crates.io-6f17d22bba15001f/oxc_sourcemap-0.13.5/src/sourcemap.rs:11:1
> ```
> 
> I think we should move that crate into rolldown ...

I'm not sure the if best way to solve this is moving crate. We sooner or later will met the similar case with not having the auth to move crates around.

We need figure out how we met this gotcha and how to prevent it. What's the best practice to solve this kind of problem? I have seen putting efforts on wrong direction without reviewing just cause more problems in the future.

@underfin wants to avoid converting cost between `rolldown_sourcemap` and `string_wizard`, so these two crates both use oxc`SourceMap` directly.

The error is caused by treating oxc's `SourceMap` as the same struct between two different versions.

- `rolldown_sourcemap` uses version A oxc`SourceMap`
- `string_wizard` uses version B oxc`SourceMap`

---

The solution I propose:

If we consider `string_wizard` as a standalone crate

- `string_wizard` should expose a own version of `SourceMap`. It should not expose any data structures related oxc.
- `string_wizard`'s `SourceMap` could expose it's fields with owned values. So we could use owned values to construct oxc's sourcemap without many overhead.
- The logic that convert `string_wizard`'s `SourceMap` to oxc's `SourceMap` should be place into rolldown's codebase.

---

`string_wizard`'s features are very scoped and stable, moving it to rolldown's repo is not reasonable just because the above reason.

Boshen:
Yeah either decouple string_wizard, expose things via traits.

Or, relax the version constraint to `0.*`.

hyf0 (approved):
Let's merge it first. I think it's ok we don't have a feature gate for this. Let's start to accept feedback of bugs.

hyf0:
@IWANABETHATGUY you may start enable related ts tests of esbuild.

I will soon make the rolldown's self-compiling using oxc ts transfomer and drop the esbuild plugin.

PR #1365, opened 2024-06-07
feat: support binary loader

### LINKED ISSUE #1301: Support binary loader

ikkz:
### Description
This pr resolves #1301

The implementation of `__toBinaryNode` and `__toBinary` was copied from esbuild.

To import the `rolldown:runtime` in the code generated by the loader, I added it to the `visited` hashmap, but I'm not sure if this has any other effects


.github/workflows/codecov.yml
```diff
@@ -24,6 +24,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Install node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
```
hyf0, line 30: What's this for?

ikkz: In `_test.mjs` I used `import.meta.dirname` which requires node 20

hyf0 (approved):
Looks good.

PR #1367, opened 2024-06-08
feat: support `dataurl` module type.

### LINKED ISSUE #1299: Support dataurl loader
## Related

- https://esbuild.github.io/content-types/#data-url

7086cmd:
resolves #1299.

As for the `Data Url`, I modified the `read_source` and directly converted it into Data Url through `MIME`, etc.

As for texts, especially `.svg`, I use `urlencoding` crate to avoid chars leading to errors.

Thanks to the inspiration of #1406, this pr can use `infer` crate to infer the MIME type through file content.


crates/rolldown_loader_utils/src/base64_to_esm.rs
```diff
@@ -1,3 +1,3 @@
-pub fn base64_to_esm(source: &str) -> String {
-  ["export default '", source, "';"].concat()
+pub fn base64_to_esm(source: &str) -> anyhow::Result<String> {
+  Ok(["export default '", source, "';"].concat())
```
IWANABETHATGUY: You don't need to add `Result` here.

crates/rolldown_loader_utils/src/binary_to_esm.rs
```diff
@@ -1,20 +1,26 @@
...
-    "} from '",
-    runtime_resource_id,
-    "'; export default ",
-    to_binary,
-    "('",
-    base64,
-    "')",
-  ]
-  .concat()
+  Ok(
```
IWANABETHATGUY: You don't need to add Result here.

7086cmd: Ok. It may be my obsession on uniforming :-), but due to the performance, and understandability, etc. I will delete it. Thanks for figuring it out!

7086cmd: But it seems that the `Result` for the `base64` function is necessary, otherwise may lead to `clippy::match_same_arms`.
Both `DataUrl` and `Base64` requires the base64 because I treat them same, while the `DataUrl` has the "preprocess" step reading the source.

7086cmd:
I refactored the code and removed useless `Result` now with tests passed.

crates/rolldown_utils/src/mime.rs
```diff
@@ -0,0 +1,25 @@
+pub fn get_data_url_mime_by_extension(extension: &str) -> Option<&'static str> {
```
hyf0: Would you mind checking out if there is ready a solid crate to do this thing?

Cargo.toml
```diff
@@ -129,6 +129,7 @@ tokio              = { version = "1.38.0", default-features = false }
 tracing            = "0.1.40"
 tracing-chrome     = "0.7.2"
 tracing-subscriber = "0.3.18"
+urlencoding        = "2.1.3"
```
hyf0, line 136: What's this for?

Chen-anon:
Hello, this is Ethan's friend's account. Ethan Goh's account enables 2FA and cannot login on campus (as a high school student), while Ethan is taking final examination of the semester for 3 days, and completed just now.

> Here is the reply from Ethan
> Thanks for reviewing. The `urlencoding` crate is added for the `SVG` format, taking `<script>` for example, it will transform it into `%3Cscript%3E`, which is used in some `<`-included SVG files. Without this crate, the compilation of official icon of rolldown [https://rolldown.rs/lightning-down.svg](https://rolldown.rs/lightning-down.svg) will raise errors.
> I have found a `MIME` crate, but I am unsure about formats can be bundled by `data url`. I will add the crate for `utils` tonight.

hyf0:
Cool. Just take your time.

7086cmd:
@hyf0 Thanks. I have replaced my codes to use the `mime` and `mime_guess` crate.

7086cmd:
[Sorry, I closed this pr in accident.]

7086cmd:
Is there anyone reviews (?

hyf0:
Please don't be hurry. We will review it when we got time.

7086cmd:
OK, thanks. since there will be a scholastic exam of 4 subjects on July 6 & 7, I may not sync with the PR on time :)

7086cmd:
Sorry, due to my confusing action, I mistakenly removed commits from my contribution repository. Now it is OK.

7086cmd:
Through `infer` crate, now this pr can infer invalid extensions.

Unlike `magic`, `infer` crate implements the function natively without the requirements of `libmagic`.

crates/rolldown_utils/src/mime.rs
7086cmd: I have used `mime` now. The crate `mime_guess` is used for inferring the mime through extension, and `infer` crate is used for interring the mime through file content (magic data), which is natively implemented by Rust and doesn't require `libmagic`.

Cargo.toml
7086cmd: The urlencoding crate is added for the SVG format, taking <script> for example, it will transform it into %3Cscript%3E, which is used in some <-included SVG files. Without this crate, the compilation of official icon of rolldown https://rolldown.rs/lightning-down.svg will raise errors.

hyf0:
Cool. I'll take the PR from here and made some changes.

ikkz:
> Thanks to the inspiration of #1406, this pr can use infer crate to infer the MIME type through file content.

I'm very happy that my pr was helpful to you, and I also learned about the excellent crate 'infer' from your code, which I hadn't discovered before.

However, I don't think [this function](https://github.com/rolldown/rolldown/pull/1406/files#diff-f3ba2c9c2543ad91cbb0222f1d389f90f78ff098280a33fc9fe1d990fbb15687R18) can be omitted. It is the implementation of dataurl compatibility for all browsers.

And after this PR is merged, maybe my PR #1406 can be modified to complete the tests in `esbuild/loader` for the dataurl loader? @hyf0

hyf0 (approved):
Thanks a lot.

hyf0:
> However, I don't think [this function](https://github.com/rolldown/rolldown/pull/1406/files#diff-f3ba2c9c2543ad91cbb0222f1d389f90f78ff098280a33fc9fe1d990fbb15687R18) can be omitted. It is the implementation of dataurl compatibility for all browsers.

@ikkz Yeah. Please open a new PR for this function.

PR #1368, opened 2024-06-09
feat: wasm plugin

underfin:
### Description

The pr intend to implement vite wasm plugin at rust side, maybe it could be as a rolldown internal `wasm` laoder, I'm not sure for it, if it is a good idea, we could migrate it at future.


hyf0 (approved):
Nice. However I would consider this as workaround. The best practice way to support wasm would be support syntax like

```js
WebAssembly.instantiateStreaming(fetch( new URL('./demo.wasm', import.meta.url)), importObject).then(
  (results) => {
    // Do something with the results!
  },
);

```

underfin:
Yeah. It will be supported as asset. It is the feature of vite asset plugin .

PR #1369, opened 2024-06-09
feat(ModuleType): support `empty`

### LINKED ISSUE #1298: Support empty loader
## Related

- https://esbuild.github.io/content-types/#empty-file

AliceLanniste:
### Description

the related issue #1298


crates/rolldown/src/utils/normalize_options.rs
```diff
@@ -18,6 +18,7 @@ pub fn normalize_options(mut raw_options: crate::BundlerOptions) -> NormalizeOpt
       ("mjs".to_string(), ModuleType::Js),
       ("cjs".to_string(), ModuleType::Js),
       ("txt".to_string(), ModuleType::Text),
+      ("empty".to_string(), ModuleType::Empty),
```
hyf0: This is not needed.

crates/rolldown/src/utils/parse_to_ast.rs
```diff
@@ -41,6 +41,7 @@ pub fn parse_to_ast(
     ModuleType::Binary => {
       (binary_to_esm(&source, options.platform, ROLLDOWN_RUNTIME_RESOURCE_ID).into(), ParseType::Js)
     }
+    ModuleType::Empty => ("".to_string().into(), ParseType::Js),
```
hyf0: Probably should transform `export {}` to mark it as an esm.

hyf0 (changes requested):

crates/rolldown/src/utils/parse_to_ast.rs
```diff
@@ -41,6 +41,7 @@ pub fn parse_to_ast(
     ModuleType::Binary => {
       (binary_to_esm(&source, options.platform, ROLLDOWN_RUNTIME_RESOURCE_ID).into(), ParseType::Js)
     }
+    ModuleType::Empty => ("export default {}".to_string().into(), ParseType::Js),
```
hyf0: `export {}` not `export default {}`

hyf0:
I need to take a look how esbuild implement empty loader. It seems that it not only pretend the file to be empty but have some other special behaviors.

crates/rolldown/src/utils/parse_to_ast.rs
AliceLanniste: when I set ` export { }`,it will get output  
image
so I changed `export default { }`

ryanxiang7:
> It seems that it not only pretend the file to be empty but have some other special behaviors.

I found that `esbuild` will convert the variables exported from an `empty` file to `void 0`. Maybe some conversions need to be done at the `transform` stage. :)


for example:

build config:
```js
// esbuild.mjs

import * as esbuild from 'esbuild'

await esbuild.build({
  entryPoints: ['./src/index.js'],
  bundle: true,
  format: 'esm',
  loader: { '.empty.js': 'empty' },
  outfile: './dist/out.js',
})
```
entry file:
```js
import { str } from './index.empty.js'

import * as s from './default.empty.js'

console.log('[🤔] from index.empty.js -> ', str)

console.log('[🤔] from default.empty.js -> ', s, s.age, s.name)

const arr = [1, 2, 3]

console.log('[🤔] index arr -> ', arr)
```

index.empty.js
```js
export const str = 'empty'
```

default.empty.js
```js
const age = 18
const name = 'esbuild'

export default {
  age,
  name,
}
```

output:
```js

// ... other code

// src/default.empty.js
var require_default_empty = __commonJS({
  "src/default.empty.js"() {
  }
});

// src/index.js
var s = __toESM(require_default_empty());
console.log("[\u{1F914}] from index.empty.js -> ", void 0); // `str` -> void 0
console.log("[\u{1F914}] from default.empty.js -> ", s, void 0, void 0); // `s.name`/`s.age` -> void 0, but `s` not.
var arr = [1, 2, 3];
console.log("[\u{1F914}] index arr -> ", arr);
```

AliceLanniste:
> I found that `esbuild` will convert the variables exported from an `empty` file to `void 0`. Maybe some conversions need to be done at the `transform` stage. :)
> 
> for example:
> 
> build config:
> 
> ```js
> // esbuild.mjs
> 
> import * as esbuild from 'esbuild'
> 
> await esbuild.build({
>   entryPoints: ['./src/index.js'],
>   bundle: true,
>   format: 'esm',
>   loader: { '.empty.js': 'empty' },
>   outfile: './dist/out.js',
> })
> ```
> 
> entry file:
> 
> ```js
> import { str } from './index.empty.js'
> 
> import * as s from './default.empty.js'
> 
> console.log('[🤔] from index.empty.js -> ', str)
> 
> console.log('[🤔] from default.empty.js -> ', s, s.age, s.name)
> 
> const arr = [1, 2, 3]
> 
> console.log('[🤔] index arr -> ', arr)
> ```
> 
> index.empty.js
> 
> ```js
> export const str = 'empty'
> ```
> 
> default.empty.js
> 
> ```js
> const age = 18
> const name = 'esbuild'
> 
> export default {
>   age,
>   name,
> }
> ```
> 
> output:
> 
> ```js
> // ... other code
> 
> // src/default.empty.js
> var require_default_empty = __commonJS({
>   "src/default.empty.js"() {
>   }
> });
> 
> // src/index.js
> var s = __toESM(require_default_empty());
> console.log("[\u{1F914}] from index.empty.js -> ", void 0); // `str` -> void 0
> console.log("[\u{1F914}] from default.empty.js -> ", s, void 0, void 0); // `s.name`/`s.age` -> void 0, but `s` not.
> var arr = [1, 2, 3];
> console.log("[\u{1F914}] index arr -> ", arr);
> ```

yes ,you are right .`moduleType::Empty` should be `void 0`,I found  ` "shimMissingExports": true` in _config.json, variables from empty file will convert to void 0. so I add code that sets  shimMissingExports to true when the ModuleType::Empty.

crates/rolldown/src/utils/normalize_options.rs
```diff
@@ -63,7 +63,9 @@ pub fn normalize_options(mut raw_options: crate::BundlerOptions) -> NormalizeOpt
     sourcemap: raw_options.sourcemap.unwrap_or(SourceMapType::Hidden),
     sourcemap_ignore_list: raw_options.sourcemap_ignore_list,
     sourcemap_path_transform: raw_options.sourcemap_path_transform,
-    shim_missing_exports: raw_options.shim_missing_exports.unwrap_or(false),
+    shim_missing_exports: raw_options.shim_missing_exports.unwrap_or_else(|| {
```
hyf0: This is not ideal and will cause enabling `shim_missing_exports` implicitly on all modues.

hyf0: https://github.com/rolldown/rolldown/blob/aef45e09fa93e7df8f18422fb8d5c47fae0c77f0/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs#L446

Should check it here.

Probably also need to add a field in `NormalModule` to store `ModuleType`.

---

This is not a easy contribution task anymore due to requiring touching logic of link stage.

AliceLanniste:
@hyf0 hi，my latest commit ,snap result is like this
```js


//#region b.empty
var require_b = __commonJSMin((exports, module) => {
});

//#endregion
//#region c.empty
var require_c = __commonJSMin((exports, module) => {
});

//#endregion
//#region main.js
var import_b = __toESM(require_b());
var import_c = __toESM(require_c());
console.log(import_b, import_c.default);

//#endregion
```
I can't handle the condition that a named import of a file without any exports. I read `esbuild/Link.go`,

>
https://github.com/evanw/esbuild/blob/67cbf87a4909d87a902ca8c3b69ab5330defab0a/internal/linker/linker.go#L3075
```go
// Is this a named import of a file without any exports?
	otherSourceIndex := record.SourceIndex.GetIndex()
	otherRepr := c.graph.Files[otherSourceIndex].InputFile.Repr.(*graph.JSRepr)
	if !namedImport.AliasIsStar && !otherRepr.AST.HasLazyExport &&
		// CommonJS exports
		otherRepr.AST.ExportKeyword.Len == 0 && namedImport.Alias != "default" &&
		// ESM exports
		!otherRepr.AST.UsesExportsRef && !otherRepr.AST.UsesModuleRef {
		// Just warn about it and replace the import with "undefined"
		return importTracker{sourceIndex: otherSourceIndex, importRef: ast.InvalidRef}, importCommonJSWithoutExports, nil
	}

```
>
I think should be rewrite the part of code in `bind_imports_and_exports.rs`,  I found `normalModule` type has no AST prop.

hyf0:
Thanks for the feedback. I kind of feel this task is not suitable for a newer anymore. You may leave it now, I'll finish the rest in this PR. Didn't thought that empty loader is actually harder than I think, sorry about that.

hyf0 (approved):
Thank you!

PR #1374, opened 2024-06-10
feat: advanced barrel exports optimization.

IWANABETHATGUY:
### Description
1. Finished nested namespace ref barrel exports optimization
2. Change eager inclusion in `tree_shaking.rs` into **lazy** inclusion.


crates/rolldown/src/ast_scanner/impl_visit.rs
```diff
@@ -11,6 +16,35 @@ use crate::utils::call_expression_ext::CallExpressionExt;
 
 use super::{side_effect_detector::SideEffectDetector, AstScanner};
 
+impl<'me> AstScanner<'me> {
```
hyf0: Could you move this method to `mod.ts`.

crates/rolldown/src/ast_scanner/impl_visit.rs
```diff
@@ -47,30 +81,55 @@ impl<'me, 'ast> Visit<'ast> for AstScanner<'me> {
     self.try_diagnostic_forbid_const_assign(symbol_id);
   }
 
-  fn visit_identifier_reference(&mut self, ident: &IdentifierReference) {
-    let symbol_id = self.resolve_symbol_from_reference(ident);
-    match symbol_id {
-      Some(symbol_id) if self.is_top_level(symbol_id) => {
-        self.add_referenced_symbol(symbol_id);
+  fn visit_member_expression(&mut self, expr: &MemberExpression<'ast>) {
```
hyf0, line 51: Do we need to call `walk::walk_member_expression` here?

IWANABETHATGUY: No, if you call `walk::walk_member_expresion`, then `identifier` visitor will be called, and another referenced symbol will be added to this module. Visitor of  `member_expr_reference` and `identifier_reference` should be mutually exclusive

IWANABETHATGUY: I could add `enter_node` and `leave_node` if we need.

crates/rolldown/src/stages/link_stage/mod.rs
```diff
@@ -282,20 +283,25 @@ impl<'a> LinkStage<'a> {
                         // Reference to `require_bar_cjs`
                         stmt_info
                           .referenced_symbols
-                          .push(importee_linking_info.wrapper_ref.unwrap());
-                        stmt_info.referenced_symbols.push(self.runtime.resolve_symbol("__toESM"));
+                          .push(importee_linking_info.wrapper_ref.unwrap().into());
```
IWANABETHATGUY, line 286: All these `into` invocation, convert `symbol_ref` into a  `SymbolOrMemberExprRef::Symbol`

crates/rolldown/tests/esbuild/import_star/other_file_export_self_as_namespace_unused_es6/artifacts.snap
```diff
@@ -17,13 +17,7 @@ input_file: crates/rolldown/tests/esbuild/import_star/other_file_export_self_as_
 
 ```js
```
IWANABETHATGUY, line 19: esbuild: 
```js
================================================================================
TestOtherFileExportSelfAsNamespaceUnusedES6
---------- /out.js ----------
// foo.js
var foo = 123;
export {
  foo
};

```

crates/rolldown/tests/esbuild/import_star/import_star_export_star_as_unused/artifacts.snap
```diff
@@ -9,15 +9,6 @@ input_file: crates/rolldown/tests/esbuild/import_star/import_star_export_star_as
 
 ```js
 
-
```
IWANABETHATGUY, line 12:
```js
================================================================================
TestImportStarExportStarAsUnused
---------- /out.js ----------
// entry.js
var foo = 234;
console.log(foo);

```

crates/rolldown/src/stages/link_stage/tree_shaking.rs
```diff
@@ -56,20 +85,80 @@ fn include_module(ctx: &mut Context, module: &NormalModule) {
...
+  canonical_export_list.into_iter().for_each(|(key, symbol_ref)| {
+    ctx.used_exports_info_vec[module.id].used_exports.insert(key);
+    include_symbol(ctx, symbol_ref, &[]);
+  });
+}
+
+#[track_caller]
+fn include_symbol(ctx: &mut Context, symbol_ref: SymbolRef, chains: &[CompactStr]) {
+  let mut cursor = 0;
+
```
IWANABETHATGUY, line 115: Resolve top level member expr chain reference.

crates/rolldown/tests/esbuild/dce/import_re_export_of_namespace_import/artifacts.snap
```diff
@@ -25,7 +25,6 @@ var require_bar = __commonJSMin((exports, module) => {
 //#region node_modules/pkg/index.js
 var pkg_index_ns = {};
 __export(pkg_index_ns, {
-	bar:() => import_bar.default,
```
IWANABETHATGUY, line 28: Only `foo` is used.

crates/rolldown/tests/esbuild/dce/tree_shaking_import_identifier/artifacts.snap
```diff
@@ -22,13 +22,10 @@ class Base {
 //#region a.js
 var a_ns = {};
 __export(a_ns, {
-	Keep:() => Keep,
-	REMOVE:() => REMOVE
```
IWANABETHATGUY, line 26: Only keep is used.
here is the result of webpack: 
```js
/******/ (() => { // webpackBootstrap
/******/ 	"use strict";
var __webpack_exports__ = {};

;// CONCATENATED MODULE: ./a/b.mjs
class Base {}

;// CONCATENATED MODULE: ./a/a.mjs

class Keep extends Base {}
class REMOVE extends Base {}


;// CONCATENATED MODULE: ./a/index.js

new Keep();

/******/ })()
;

```

IWANABETHATGUY: the namespace object ref will be eliminated after we merge https://github.com/rolldown/rolldown/pull/1381

crates/rolldown/Cargo.toml
```diff
@@ -13,6 +13,7 @@ workspace = true
 
 [dependencies]
 anyhow                = { workspace = true }
+bitflags              = "2.5.0"
```
hyf0: Please use workspace so we could unify the version across crates.

crates/rolldown/src/ast_scanner/impl_visit.rs
hyf0: Ok. I prefer not to use `enter_node` and `leave_node` mechanism in any situation. It's hard to control and maintain it's flow.

hyf0: I just look it again. Could you rewrite it like

```rust
fn visit_member_expression{
  if MemberExpression::StaticMemberExpression ... {
    // ...
    // Bailout for not creating another referenced symbol
    return
  }

  // nothing special, just keep walking
  walk::visit_member_expression(self)
}
```

The current logic is complex to see when we should keep walking when we should bailout. The other visitor's function is writing like this way. It would be good if we could keep this style.

- https://github.com/rolldown/rolldown/blob/aef45e09fa93e7df8f18422fb8d5c47fae0c77f0/crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs#L356-L360
- https://github.com/rolldown/rolldown/blob/aef45e09fa93e7df8f18422fb8d5c47fae0c77f0/crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs#L421-L464
- https://github.com/rolldown/rolldown/blob/aef45e09fa93e7df8f18422fb8d5c47fae0c77f0/crates/rolldown/src/ast_scanner/impl_visit.rs#L74-L79

crates/rolldown_common/src/types/symbol_ref.rs
```diff
@@ -14,3 +15,55 @@ impl From<(NormalModuleId, SymbolId)> for SymbolRef {
...
+  /// ```js
+  /// import {test} frmo './a.js'
+  /// test.a.b
+  /// ```
+  /// `test` stored with `SymbolId`
+  /// `a.b` is the rest chain
+  pub chains: Vec<CompactStr>,
+}
+
+impl MemberExprRef {}
```
hyf0: Is this meaningful?

crates/rolldown_common/src/types/symbol_ref.rs
```diff
@@ -14,3 +15,55 @@ impl From<(NormalModuleId, SymbolId)> for SymbolRef {
...
+pub enum SymbolOrMemberExprRef {
+  Symbol(SymbolRef),
+  MemberExpr(MemberExprRef),
+}
+
+impl SymbolOrMemberExprRef {
+  /// get the first part of the expression,
+  /// e.g. `test.a.b` will return `test`
+  /// for identifier, it will return itself
+  pub fn first_part(&self) -> &SymbolRef {
```
hyf0:
```suggestion
  pub fn symbol_ref(&self) -> &SymbolRef {
```

I think it's ok you just call it `symbol` or `symbol_ref`. Because when we are referring `symbol` in `test.a.b`, the symbol would still be `test`. `.a` and `.b` might point to some symbols, but for the expression `test.a.b` itself, the symbol would only be `test`.

crates/rolldown_common/src/types/symbol_ref.rs
```diff
@@ -14,3 +15,55 @@ impl From<(NormalModuleId, SymbolId)> for SymbolRef {
...
+    match self {
+      SymbolOrMemberExprRef::Symbol(s) => s,
+      SymbolOrMemberExprRef::MemberExpr(expr) => &expr.symbol,
+    }
+  }
+}
+
+/// Crossing module ref between symbols
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct MemberExprRef {
```
hyf0, line 39: Discussion only: Do you think if it would be better we just call it `SymbolRef`?

```rust
pub struct SymbolRef {
  owner: ...,
  symbol: ...,
  chains: ..., or static_members:...
}

```

Is there any case that we will only benefit from using enum?

hyf0:
Nice work. I have comment suggested changes and other parts are all fine. We are good to merge once we solve above comments.

https://github.com/rolldown/rolldown/pull/1374#discussion_r1641907286 is only an discussion and not required to be solved in this PR.

crates/rolldown_common/src/types/symbol_ref.rs
IWANABETHATGUY: There is already a struct called `SymbolRef`, do you mean change the original `SymbolRef` sturct? 
the downsides are: 
1. the memory usage will from **8** bytes to **24**
2. the `SymbolRef` will not `Copy` any more

IWANABETHATGUY: Since we used  `SymbolRef` a lot, so the performance difference should be considerable.

hyf0: >  do you mean change the original SymbolRef sturct?

Yeah.

> the memory usage will from 8 bytes to 24
> the SymbolRef will not Copy any more

I'm only think it in aspect of if it make code cleaner. Feel like "Not copy" will make the dx worse. Ok. Let's keep it how it is now and see how it envolves.

hyf0 (approved):

PR #1377, opened 2024-06-10
ci: actions/setup-node should use .node-version

Boshen:
CI need to use the node version from .node-version, as that's what contributors should use.

### DIFF
--- a/.github/workflows/benchmark-node.yml
+++ b/.github/workflows/benchmark-node.yml
@@ -69,7 +69,7 @@ jobs:
       - name: Install node
         uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version-file: .node-version
           cache: pnpm
 
       - name: Install dependencies
--- a/.github/workflows/benchmark-rust.yml
+++ b/.github/workflows/benchmark-rust.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Install node
         uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version-file: .node-version
           cache: pnpm
 
       - name: Install dependencies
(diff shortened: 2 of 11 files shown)


hyf0:
I'll explain the reason later.

Boshen:
> I'll explain the reason later.

Wait ... we also specified .node-version, we should align it .node-version instead then.

edit: changed to `.node-version` instead.

Boshen:
Why aren't CI failing 🤔

Boshen:
Should be safe to merge now.

I think you wanted the engines field for publishing packages, maybe it should be moved to the publishing packages instead. And we rely on .node-version for development.

hyf0:
Two concepts:
- node versions that rolldown supports
- local `.node-version` that we used to development rolldown. The `.node-version` must be the supported version of rolldown and also enable to be used to running the development flow(some old version that rolldown supports might not satisfy this requirement.

For CI,
- In general purpose and rust tests actions(lint-repo, lint-node, lint-rust, rust-test), we need to use `.node-version` due to it ensures the node-version is enable to run the development flow.
- In the node-test and wasm-test actions, currently we don't have the development workflow that require `.node-version` to run(this might happens in the future, but let's ignore it so far). The node-test action need to tests all supported node versions, so node-version would an input showed in matrix field. It shouldn't use `.node-version`, though the `.node-version` is a version that rolldown supported.

---

I do agree with you about `package.json#engine` should not be updated by renovate. `.node-version` is ok due to also that newest node's version is the version that rolldown needs to support.

---

This PR does solve problems, so let's merge it first. I will create an issue to clarify the above thing.

hyf0 (approved):

PR #1379, opened 2024-06-11
docs: add instruction on debugging oxc resolver

Boshen:


Dunqing (approved):
Looks good, I spent quite a bit of time investigating #1364 as to why it doesn't deal with `*.ts` files

PR #1381, opened 2024-06-12
feat: move dependency

IWANABETHATGUY:
### Description
1. require https://github.com/rolldown/rolldown/pull/1374 to be merged.


hyf0 (approved):
This is a great improvement and all changes seem to be fine, so let's merge it first. I left some comments as marks. Let's have a meeting to discuss about them and make sure we have the same understanding about them.

PR #1382, opened 2024-06-12
refactor: bindingfy sourcemap

underfin:
### Description

It avoid convert `JSON.stringfy` map at js side, it is unnecessary overhead.

### DIFF
--- a/packages/rolldown/src/types/sourcemap.ts
+++ b/packages/rolldown/src/types/sourcemap.ts
@@ -1,7 +1,7 @@
 import { BindingSourcemap } from '../binding'
 
 export interface SourceMapInputObject {
-  file?: string
+  file?: string | null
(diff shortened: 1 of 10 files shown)


packages/rolldown/src/types/sourcemap.ts
```diff
@@ -9,3 +11,22 @@ export interface SourceMapInputObject {
 }
 
 export type SourceMapInput = SourceMapInputObject | string | null
+
+export function bidingSourcemap(
```
IWANABETHATGUY: Do you mean `bindingSourcemap`?

underfin: Updated to `bindingifySourcemap`

hyf0 (approved):

PR #1383, opened 2024-06-12
refactor: use the unified `oxc` crate to make sure version is correct

Boshen:
### Description

Use the unified crate `oxc` instead of using the splitted crates so we don't need to change versions across multiple crate definitions in Cargo.toml.

You may export some aliases if you find this inconvenient.


IWANABETHATGUY (approved):

hyf0:
Guess this will cause rolldown's related crates have to wait for oxc to be compiled?

Boshen:
> Guess this will cause rolldown's related crates have to wait for oxc to be compiled?

`oxc` compiles really fast, they don't block anything, and they don't show up in the graph with `--timings`



image

PR #1387, opened 2024-06-13
fix: rolldown typing

underfin:
### Description

The original typing has some errors.
1. the dist types not `binding.d.ts` and `rollup.d.ts`, the `tsc` couldn't move them to type dist, it need to copy manually.
2. the alias path is not replaced at dist typing, tsc provides `typescript-transform-paths` to resolve them, but it has noise at build, so here remove alias path usages.

Other, we could build binding files to `/rolldown/bidings/` instead of `/rolldown/src/`, it avoids moving binary files manually.


hyf0:
This PR solve another problem related unbundled-dts. We have met some many of them and use many workaround now.

This is not ideal. Let's try to solve it in the best suitable way. We should either use rollup-plugin-dts or https://github.com/Rich-Harris/dts-buddy to generate bundled dts files.

underfin:
Yeah, but i need it to make vite build success， let's merge it for temporarily usage.

hyf0 (approved):
Ok let's merge it. I have created an issue for tracking.

Issue #1270, opened 2024-05-29
Support `output.file` option


hyf0:
Related:

- https://github.com/rolldown/rolldown/issues/624

IWANABETHATGUY:
@hyf0 Could we close now?

PR #1393, opened 2024-06-14
feat: support output.file option

AliceLanniste:
### Description

related pr #1270

### DIFF
--- a/crates/rolldown/src/bundler.rs
+++ b/crates/rolldown/src/bundler.rs
@@ -40,31 +40,39 @@ impl Bundler {
 impl Bundler {
   #[tracing::instrument(level = "debug", skip_all)]
   pub async fn write(&mut self) -> Result<BundleOutput> {
-    let dir = self.options.cwd.as_path().join(&self.options.dir).to_string_lossy().to_string();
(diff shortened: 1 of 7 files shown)


crates/rolldown/src/bundler.rs
```diff
@@ -40,31 +40,39 @@ impl Bundler {
...
-      let dest = dir.as_path().join(chunk.filename());
-      if let Some(p) = dest.parent() {
-        if !self.fs.exists(p) {
-          self.fs.create_dir_all(p).unwrap();
-        }
-      };
-      self.fs.write(dest.as_path(), chunk.content_as_bytes()).map_err(|err| {
-        anyhow::anyhow!("Failed to write file in {:?}", dir.as_path().join(chunk.filename()))
-          .context(err)
+    if let Some(dest) = &self.options.file {
```
hyf0, line 46: Feel like there will be problems in this implementation for working with asset modules and future copy and file loader #1295.

---

I just give it a rough thinking and it's not easy job to make this feature right. I'm so far too busy to give detailed instructions on how to support this feature.

PR #1406, opened 2024-06-17
feat: support dataurl loader

### LINKED ISSUE #1299: Support dataurl loader

ikkz:
### Description
resolves #1299

This PR implements the dataurl loader based on the logic of esbuild and sets up all the related tests in `tests/esbuild`:
1. guess MIME
   a. Based on file extension
   b. If it's not possible to determine from the extension, then guess based on file content (this part requires converting the DetectContentType from the Go standard library [here](https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/net/http/sniff.go;l=21) to a Rust version, which I left a TODO in the code and I will finish it in another PR)
2. dataurl encode: Choose the shorter of the two following encodings
   a. Base64
   b. Percent encode if the file content is valid utf8


7086cmd:
Hello! In my PR #1367, you said that the `encode_as_percent_escaped_dataurl` function shouldn't be omitted to improve compatibility. However, through the crate `urlencoding`, these chars (e.g. `\n`) can also be translated into percent-included strings. Here is a snap:

aa1210e6287060986e5d8a0bf0adf8e4

7086cmd:
> Hello! In my PR #1367, you said that the `encode_as_percent_escaped_dataurl` function shouldn't be omitted to improve compatibility. However, through the crate `urlencoding`, these chars (e.g. `\n`) can also be translated into percent-included strings. Here is a snap:
> 
> aa1210e6287060986e5d8a0bf0adf8e4

So, maybe the case of `\n`, etc. is solved :)

ikkz:
> Hello! In my PR #1367, you said that the `encode_as_percent_escaped_dataurl` function shouldn't be omitted to improve compatibility. However, through the crate `urlencoding`, these chars (e.g. `\n`) can also be translated into percent-included strings. Here is a snap:

Esbuild generates the shortest possible dataurl and is compatible with all browsers, although the results are not standard, so it is the expected behavior , see this [playground](https://esbuild.github.io/try/#YgAwLjIzLjAAeyBsb2FkZXI6IHsgIi5kYXRhIjogImRhdGF1cmwiIH0gfQBlAGVudHJ5LmRhdGEAPHN0cmluZyBzbmlwcGV0Pgo) and related issues https://github.com/evanw/esbuild/issues/1843 https://github.com/evanw/esbuild/issues/3418

And I'm also not sure if we use `mime_guess`, which contains [thousands of mappings](https://github.com/abonander/mime_guess/blob/master/src/mime_types.rs) from extension to mime type, I think most of them may not be necessary for the front-end bundler scenario and may increase the binary size

Do you have any suggestions for these issues? @hyf0

7086cmd:
While `mime_guess` does include some rare extensions’ MIME types, I don’t think we should simply ignore them. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs):

> Data URLs, URLs prefixed with the data: scheme, allow content creators to embed small files inline in documents. They were formerly known as "data URIs" until that name was retired by the WHATWG.

Given this explanation, it seems important to consider these rare types to maintain the loader’s completeness. Additionally, using a crate can make these cases clearer and more manageable.

ikkz:
@7086cmd As the definition suggests, in frontend business development, I typically use data URLs to embed small, commonly-used resources that can be directly consumed by the browser into the application's build artifacts. For example, for a button's icon, the result from a data URL loader can be directly passed to the `src` attribute of an `img` tag.

For less common resources, which usually require some data preprocessing before use, using a base64 loader might be a better choice. One reason is that the output format of a data URL loader is `data:[<mediatype>][;base64],<data>`, and developers might need to extract the `<data>` for processing, which requires additional string manipulation. Another reason is that the user clearly knows the type of the imported resource, and the MIME type returned by the data URL loader may not always be what the developer expects.

Therefore, in my view, the types recognized by [mime_type_by_extension](https://github.com/rolldown/rolldown/pull/1406/files#diff-4b612e077c82ae0e05e50eb0d419e02c05a04b83c6ac5440c0d0c9d0c38af942R3-R30) and [infer](https://github.com/bojand/infer?tab=readme-ov-file#supported-types) already cover the scenarios for using data URLs. Thus, I feel there is no need to expend time and space to support all MIME types.

7086cmd:
I agree with you. However, in my opinion, directly writing numerous matches that aren’t part of the core functionality isn’t ideal for a project. Instead, I’ll create a new crate specifically for handling MIME functions.

However, for the `percent`-decoded function related to esbuild's behaivor, I still believe that using the `urlencoding` crate is a better approach.

7086cmd:
@ikkz I have developed a new crate called [`mime_more`](https://github.com/7086cmd/mime_more.git) that features an `extension-light` capability, including extensive matches for various types, such as video and audio, which may be useful. We can explore integrating it into `rolldown` if you're interested.

I have extended the `mime_more` crate with `dataurl` and implemented the current logic into the `Dataurl` struct within `mime_more`.

However, I noticed that I haven't conducted any benchmarks for `mime_more` yet, so its performance speed is still uncertain.

ikkz:
@7086cmd Sorry for the late reply.
I have looked at the current code and your crate. It seems there is a word mistake.
Maybe you wanted to write "[textual](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs#:~:text=token%20if%20non%2D-,textual,-%2C%20and%20the%20data)" instead of "texture"?

PR #1419, opened 2024-06-19
feat: support banner hook

### LINKED ISSUE #1417: Support banner  footer Hooks
The hooks supported need to some step.
- add rust plugin function and called it at bundle, you could follow `renderChunk` at [here](https://github.com/rolldown/rolldown/blob/main/crates/rolldown_plugin/src/plugin.rs#L84)
- add rust side binding, you can follow `renderChunk` at [rust side](https://github.com/rolldown/rolldown/blob/main/crates/rolldown_binding/src/options/plugin/js_plugin.rs#L165) and [js side](https://github.com/rolldown/rolldown/blob/main/packages/rolldown/src/plugin/bindingify-output-hooks.ts#L30)
-  implement `ChunkInfo` and using it at hooks(it could be a single pr)
- add test for it at [here](https://github.com/rolldown/rolldown/tree/main/packages/rolldown/tests/fixtures/plugin)

The rollup document at [here](https://rollupjs.org/plugin-development/#banner)

komyg:
### Description

Fix: https://github.com/rolldown/rolldown/issues/1417 issue.


underfin:
@komyg Let us merge it at now, i will refactor code for render chunk code concat, so it merged to avoid unnecessary rebase.
Thank you a lot :)

underfin (approved):

PR #1422, opened 2024-06-20
fix: option hook can be async and can return overwritten input options

thepassle:
### Description

The `options` hook in Rollup can be used to overwrite the input options, by returning a new inputOptions object from the hook. It also can be async.

https://rollupjs.org/plugin-development/#options


Boshen:
Some guidance is required from the team for adding a few tests 😅

IWANABETHATGUY:
@underfin Could you give some guidance?

underfin:
Thanks for your point to my implement error, here should be `InputOptions` instead of `NormalizedInputOptions`. I updated all code and test for it.

underfin (approved):

PR #1457, opened 2024-06-27
feat(rolldown): enable DCE and `--define` prepass

Boshen:
### Description

This PR adds the infra for DCE and `--define`. Future PRs will refine these features to make them feature complete.


Boshen:
2 AST passes = 5 - 8% perf regression, seems good to me.

PR #1459, opened 2024-06-27
chore(renamer):  remove .rev() when rename top level symbol

IWANABETHATGUY:
### Description
1. align with esbuild, https://esbuild.github.io/try/#YgAwLjIwLjIALS1idW5kbGUgLS1taW5pZnk9ZmFsc2UgLS1mb3JtYXQ9ZXNtAGUAZW50cnkuanMAaW1wb3J0IHtjfSBmcm9tICcuL2ZpbGUuanMnCmNvbnN0IGEgPSAxMDAwMDsKY29uc29sZS5sb2coYSk7AABmaWxlLmpzAGV4cG9ydCBjb25zdCBjID0gJ3Rlcyc7CmZ1bmN0aW9uIGEoKSB7Cn0KYSgpOw
2. Keep the same order between `renamer` and `codegen`, before we rename from bottom to top, and code gen from 
top to bottom, this will have some issue.


hyf0:
> align with esbuild

It's intentional to not align with esbuild to make the entry modules looks like what it was. Anyway this is a small thing.

> Keep the same order between renamer and codegen, before we rename from bottom to top, and code gen from
top to bottom, this will have some issue.

Could you expand this part?

IWANABETHATGUY:
I found that this is not needed.

PR #1461, opened 2024-06-27
perf(rust/renamer): new deconflict strategy

### LINKED ISSUE #389: Need a new strategy for dealing with conflicted identifiers
https://github.com/rolldown/rolldown/blob/b8555ac034ebe97bb0d7f5bbb46a8a1bd7cf1d7d/crates/rolldown/src/utils/renamer.rs#L36-L56

The current algorithm counts from 0 and may end up with a huge number of allocations and hashes for `candidate_name`.

The challenge is to come up with the most efficient strategy for finding the new name.

One possible solution is to follow esbuild - the assigned new name is a new name on the scope

The original discussion is https://github.com/rolldown-rs/rolldown/issues/365.

IWANABETHATGUY:
### Description
1. Closed https://github.com/rolldown/rolldown/issues/389


crates/rolldown/src/utils/renamer.rs
```diff
@@ -24,32 +42,39 @@ impl<'name> Renamer<'name> {
...
-        while self.used_canonical_names.contains(&candidate_name) {
-          count += 1;
-          candidate_name = Cow::Owned(format!("{original_name}${count}").into());
+        loop {
+          match self.used_canonical_names.entry(candidate_name.clone()) {
+            Entry::Occupied(mut occ) => {
+              let next_conflict_index = *occ.get() + 1;
+              *occ.get_mut() = next_conflict_index;
+              candidate_name =
+                Cow::Owned(format!("{candidate_name}${next_conflict_index}",).into());
```
hyf0:
```suggestion
                Cow::Owned(format!("{original_name}${next_conflict_index}",).into());
```

crates/rolldown/src/utils/renamer.rs
```diff
@@ -24,32 +42,39 @@ impl<'name> Renamer<'name> {
...
     match self.canonical_names.entry(canonical_ref) {
-      std::collections::hash_map::Entry::Vacant(vacant) => {
-        let mut count = 0;
+      Entry::Vacant(vacant) => {
         let mut candidate_name = original_name.clone();
-        while self.used_canonical_names.contains(&candidate_name) {
-          count += 1;
-          candidate_name = Cow::Owned(format!("{original_name}${count}").into());
+        loop {
+          match self.used_canonical_names.entry(candidate_name.clone()) {
```
hyf0, line 62:
```suggestion
          match self.used_canonical_names.entry(original_name.clone()) {
```

Should you use `original_name` name here?

crates/rolldown/tests/fixtures/deconflict/issue_364/artifacts.snap
```diff
@@ -11,13 +11,13 @@ input_file: crates/rolldown/tests/fixtures/deconflict/issue_364
 import { default as assert } from "node:assert";
 
 //#region shared.js
-const a$2 = 'shared.js';
+const a$1$1 = 'shared.js';
```
hyf0: should be `a$n` rather than `a$1$1`

crates/rolldown/src/utils/renamer.rs
IWANABETHATGUY: We should use candidate name here, you could ref the algorithm https://github.com/rolldown/rolldown/pull/1461/files#diff-c87caec33470b59eac43cd6ac05bef999e6a4fa010e716fd3e546971148f01d2R15,

IWANABETHATGUY: We need to find a fixed point

crates/rolldown/src/utils/renamer.rs
IWANABETHATGUY: same above

crates/rolldown/src/utils/renamer.rs
hyf0: If there's is another `a`, in the current `algorithm`, will it be renamed to `a$1$1$1`? If so, I would like to see if we could fix this, it's kind of ugly in the output.

If we don't care the how output looks, we could follow how parcel did which is renaming the vars to a meaningless hashed value. This way it should be more faster.

crates/rolldown/tests/fixtures/deconflict/issue_364/artifacts.snap
IWANABETHATGUY: Alright, I get you, let me test if performance got regressed

crates/rolldown/src/utils/renamer.rs
IWANABETHATGUY: Fixed

IWANABETHATGUY: We can't use `original_name` here, it will lead a dead loop, you could see the case in https://github.com/rolldown/rolldown/pull/1461/files#diff-c87caec33470b59eac43cd6ac05bef999e6a4fa010e716fd3e546971148f01d2R15

crates/rolldown/src/utils/renamer.rs
IWANABETHATGUY: fixed

hyf0 (approved):
Well done!

PR #1464, opened 2024-06-28
refactor: remove extra path conversion in Bundler::write

Boshen:
I noticed that we are converting the path back and forth for no obvious reason.

### DIFF
--- a/crates/rolldown/src/bundler.rs
+++ b/crates/rolldown/src/bundler.rs
@@ -15,7 +15,6 @@ use rolldown_common::SharedFileEmitter;
 use rolldown_error::BuildError;
 use rolldown_fs::{FileSystem, OsFileSystem};
 use rolldown_plugin::{BoxPlugin, HookBuildEndArgs, HookRenderErrorArgs, SharedPluginDriver};
-use sugar_path::SugarPath;
 use tracing_chrome::FlushGuard;
 
 pub struct Bundler {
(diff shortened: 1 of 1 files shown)


crates/rolldown/src/bundler.rs
```diff
@@ -40,31 +39,27 @@ impl Bundler {
 impl Bundler {
   #[tracing::instrument(level = "debug", skip_all)]
   pub async fn write(&mut self) -> Result<BundleOutput> {
-    let dir = self.options.cwd.as_path().join(&self.options.dir).to_string_lossy().to_string();
+    let dir = self.options.cwd.join(&self.options.dir);
```
Boshen, line 42: `.to_string_lossy().to_string()` creates an extra allocation, and returns a path with [U+FFFD REPLACEMENT CHARACTER](https://doc.rust-lang.org/std/char/constant.REPLACEMENT_CHARACTER.html) so I don't know why it's needed.

crates/rolldown/src/bundler.rs
```diff
@@ -40,31 +39,27 @@ impl Bundler {
...
 
-    self.fs.create_dir_all(dir.as_path()).map_err(|err| {
-      anyhow::anyhow!(
-        "Could not create directory for output chunks: {:?} \ncwd: {}",
-        dir.as_path(),
-        self.options.cwd.display()
-      )
-      .context(err)
+    self.fs.create_dir_all(&dir).map_err(|err| {
+      anyhow::anyhow!("Could not create directory for output chunks: {:?}", dir).context(err)
```
Boshen, line 49: I removed `cwd` because `dir` already contains `cwd`.

crates/rolldown/src/bundler.rs
```diff
@@ -40,31 +39,27 @@ impl Bundler {
 impl Bundler {
   #[tracing::instrument(level = "debug", skip_all)]
   pub async fn write(&mut self) -> Result<BundleOutput> {
-    let dir = self.options.cwd.as_path().join(&self.options.dir).to_string_lossy().to_string();
+    let dir = self.options.cwd.join(&self.options.dir);
 
-    let mut output = self.bundle_up(true).await?;
+    let mut output = self.bundle_up(/* is_write */ true).await?;
```
Boshen, line 44: For boolean arguments, better to an inline comment for readability.

hyf0 (approved):

PR #1465, opened 2024-06-28
refactor(rolldown): clean up code around scan stage

Boshen:
Some readability improvements.

### DIFF
--- a/crates/rolldown/src/module_loader/module_loader.rs
+++ b/crates/rolldown/src/module_loader/module_loader.rs
@@ -7,7 +7,6 @@ use rolldown_error::BuildError;
 use rolldown_fs::OsFileSystem;
 use rolldown_oxc_utils::OxcAst;
 use rolldown_plugin::SharedPluginDriver;
-use rolldown_utils::rustc_hash::FxHashSetExt;
 use rustc_hash::{FxHashMap, FxHashSet};
 use std::sync::Arc;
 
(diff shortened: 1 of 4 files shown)


crates/rolldown/src/module_loader/module_loader.rs
```diff
@@ -174,32 +173,25 @@ impl ModuleLoader {
...
+        id: self.try_spawn_new_task(info, /* is_user_defined_entry */ true).expect_normal(),
         kind: EntryPointKind::UserDefined,
       })
-      .inspect(|e| {
-        user_defined_entry_ids.insert(e.id);
-      })
       .collect::<Vec<_>>();
 
+    // Store the already consider as entry module
+    let user_defined_entry_ids: FxHashSet<NormalModuleId> =
```
Boshen: Instead of inserting it in place, do it in one go to better utilize cpu cache, otherwise `user_defined_entry_ids` may go out of the cache in the loop above.

`user_defined_entries` is generally small, so there won't be much of an impact.

IWANABETHATGUY (dismissed):

crates/rolldown/src/module_loader/module_loader.rs
hyf0: `user_defined_entries` could be huge as there use case like making all files as input. https://rollupjs.org/configuration-options/#input:~:text=If%20you%20want%20to%20convert%20a%20set%20of%20files%20to%20another%20format%20while%20maintaining%20the%20file%20structure%20and%20export%20signatures

Boshen: Let me double check

Boshen: Reverted.

hyf0 (approved):

Issue #186, opened 2023-11-07
How to support `define` options.


underfin:
Because the vite use esbuild `define` option at optimizer, the rollup has `[@rollup/plugin-replace](https://github.com/rollup/plugins/tree/master/packages/replace#rollupplugin-replace)` to do it,  it using string match to replace, it is performance friendly but other ecosystem using last to do it, eg `swc/webpack`. 

I think we can use `@rollup/plugin-replace` to do it at milestone 1. It means we need to support some plugin API for the plugin, it is easy to implement.
But we need re-think it how to support it at rolldown for better performance and correct replace result.

kazupon:
I am interested in this issue and making the plugin work.

we consider performance, I think the plugin should be implemented on the Rust side.
However, we would ideally like to keep it compatible with the existing rollup plugin.
Do you have in your team about a plugin plan, how to implemetation idea?

shulaoda:
It seems that this feature has already been supported in PR #1882?

hyf0:
Yeah. But we haven't exposed the options from nodejs. Will closed this issue once it's done.

PR #1470, opened 2024-06-28
feat: simple binary expr folder

IWANABETHATGUY:
### Description
1. A deadly simple binary expr folder impl.
2. Eliminating code with `RemoveDeadCode` Visitor.

### DIFF
--- a/crates/rolldown/src/utils/fold_const_value.rs
+++ b/crates/rolldown/src/utils/fold_const_value.rs
@@ -1,15 +1,16 @@
 use oxc::allocator::Allocator;
-use oxc::ast::{ast::*, visit::walk_mut, AstBuilder, VisitMut};
+use oxc::ast::ast::{Expression, Program};
+use oxc::ast::{AstBuilder, VisitMut};
 use oxc::span::SPAN;
 use oxc::syntax::operator::BinaryOperator;
 
-/// This is the very basic version of Inline Binary, aiming to support `process.env.NODE_ENV === "production"` -> `true` or `false`.
+/// This is a very basic version of binary expression folder, aiming to support converting `process.env.NODE_ENV === "production"` -> `true` or `false`.
 /// Used before dead branch elimination.
-pub struct InlineBinaryValueLight<'a> {
+pub struct BasicInlineBinaryValue<'a> {
   ast: AstBuilder<'a>,
 }
 
-impl<'a> InlineBinaryValueLight<'a> {
+impl<'a> BasicInlineBinaryValue<'a> {
   pub fn new(allocator: &'a Allocator) -> Self {
     Self { ast: AstBuilder::new(allocator) }
   }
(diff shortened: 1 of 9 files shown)


crates/rolldown/src/utils/fold_const_value.rs
```diff
@@ -0,0 +1,51 @@
+use oxc::allocator::Allocator;
+use oxc::ast::{ast::*, visit::walk_mut, AstBuilder, VisitMut};
+use oxc::span::SPAN;
+use oxc::syntax::operator::BinaryOperator;
+
+/// This is the very basic version of Inline Binary, aiming to support `process.env.NODE_ENV === "production"` -> `true` or `false`.
+/// Used before dead branch elimination.
+pub struct InlineBinaryValueLight<'a> {
```
hyf0: Oxc is working on this: https://github.com/oxc-project/oxc/blob/main/crates/oxc_minifier/src/ast_passes/replace_global_defines.rs

I think is best we improve oxc rather than working on an temparay solution.

IWANABETHATGUY: I think it is related to https://github.com/rolldown/rolldown/issues/186, rather than `BinaryExpr` inline

hyf0: I got it wrong. Then I think you should improve https://github.com/oxc-project/oxc/blob/main/crates/oxc_minifier/src/ast_passes/remove_dead_code.rs. @Boshen cc

Feel like duplicate works of oxc. If we are using oxc's `Compressor`, this feature should be provided.

hyf0: From my understanding, boshen don't want rolldown use `Compressor` to do such things, that's why https://github.com/oxc-project/oxc/issues/3706 get closed and `RemoveDeadCode` is introduced.

But `RemoveDeadCode` is immature so far, so to do such things, I guess we need port some features of `Compressor` to `RemoveDeadCode`.

IWANABETHATGUY: The full implementation for constant folding is not trivial (And we don't need all the features of  it) . Also It may hurt performance in the scan phase, or oxc needs to implement a lightweight version of const folding, which is not necessary. The current implementation is enough for most cases, and why I mark this as a temporary solution is that I want to find a solution to merge all the preprocessor visitors in one go, which should be a parser hook or something else.

IWANABETHATGUY: > From my understanding, boshen don't want rolldown use `Compressor` to do such things, that's why [oxc-project/oxc#3706](https://github.com/oxc-project/oxc/issues/3706) get closed and `RemoveDeadCode` is introduced.
> 
> But `RemoveDeadCode` is immature so far, so to do such things, I guess we need port some features of `Compressor` to `RemoveDeadCode`.

https://github.com/oxc-project/oxc/blob/main/crates/oxc_minifier/src/compressor/fold.rs, this is what we want to improve.

IWANABETHATGUY: But it looks overkill for supporting `process.env.NODE_ENV === "production"`  and dead branch eliminating

hyf0: > why I mark this as a temporary solution is that I want to find a solution to merge all the preprocessor visitors in one go

Let me put it this way. We weren't urgent to support this feature. So let's have a discussion with @Boshen and try find a globally optimal solution.

Personally, I have questions:

- If we should maintain this logic in rolldown's repo, or should we do it in oxc's repo?
- Could/Should we reuse oxc's logic?

---

> The current implementation is enough for most cases

I don't against merging this PR, since it's almost finished. But I hope we won't stop there, since this is only temporary solution as you said.

hyf0: > But it looks overkill for supporting `process.env.NODE_ENV === "production"` and dead branch eliminating

Not sure. For example I'm thinking about how we support `require('./' + 'foo' + ".js")`. This could be done by the complete version of folding expression.

And

```js
if (sss || 'foo') {
  ...
}
```

should become

```js
{
  ...
}
```

We probably gonna need the complete version of folding expression.

Boshen:
I intend to port esbuild's `--define`. I think we should clearly what a compiler should do, and what a bundler should do so we can have clearer scopes.

IWANABETHATGUY:
> I intend to port esbuild's `--define`. I think we should clearly what a compiler should do, and what a bundler should do so we can have clearer scopes.

agree.

hyf0 (approved):

PR #1472, opened 2024-06-28
refactor(rolldown_resolver): remove raw json API to reduce memory usage

Boshen:
Less memory foot print = faster code

### DIFF
--- a/crates/rolldown_common/src/types/side_effects.rs
+++ b/crates/rolldown_common/src/types/side_effects.rs
@@ -18,30 +18,28 @@ impl DeterminedSideEffects {
   }
 }
 
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 /// A field in `package.json`
 pub enum SideEffects {
   Bool(bool),
   String(String),
   Array(Vec<String>),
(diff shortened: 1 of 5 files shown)


crates/rolldown_common/src/types/side_effects.rs
```diff
@@ -18,29 +18,31 @@ impl DeterminedSideEffects {
...
 
-#[derive(Clone, Debug)]
+#[derive(Debug, Clone, Default)]
 /// A field in `package.json`
 pub enum SideEffects {
   Bool(bool),
   String(String),
   Array(Vec<String>),
+  #[default]
+  None,
```
hyf0: Seems misunderstandable. `SideEffects::None` will be misunderstand as  not have side effect. `SideEffects::Undefined` seems more sutiable.

However, I don't really get the benefit of changing `Option<SideEffects>` to `SideEffects::None`. The previous seems decrible the data accurately.

Boshen: I don't have a preference, reverted.

hyf0 (approved):

PR #1474, opened 2024-06-28
perf: use jemalloc on systems that performs better, and mimalloc on windows

Boshen:
I forgot how I ended up with this setting for `oxlint`, but it was from trial-and-error :sweat_and_smile:

Due to the use of bumpalo arena allocator, my guess is that jemalloc performs better under less allocation pressure.

In the triangle benchmark, this reduced build time from 42ms to 39ms.

### DIFF
--- a/crates/rolldown_binding/Cargo.toml
+++ b/crates/rolldown_binding/Cargo.toml
@@ -35,11 +35,11 @@ rustc-hash           = { workspace = true }
 serde                = { workspace = true }
 tracing              = { workspace = true }
 
-[target.'cfg(all(not(target_os = "linux"), not(target_family = "wasm")))'.dependencies]
-mimalloc = { workspace = true }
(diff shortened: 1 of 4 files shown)


crates/rolldown_binding/Cargo.toml
```diff
@@ -35,11 +35,11 @@ rustc-hash           = { workspace = true }
 serde                = { workspace = true }
 tracing              = { workspace = true }
 
-[target.'cfg(all(not(target_os = "linux"), not(target_family = "wasm")))'.dependencies]
-mimalloc = { workspace = true }
+[target.'cfg(not(target_env = "msvc"))'.dependencies]
```
Boshen, line 38: Oh ... I need to keep `not(target_family = "wasm"` for both.

Boshen:
We need bigger benchmarks! I'll revisit this once we can bundle HUGE projects.

PR #1478, opened 2024-06-28
perf(rust): reduce operations `legitimize_identifier_name`

hyf0:
### Description

Yeah. Probably no improvements in practice.

### DIFF
--- a/crates/rolldown_utils/src/ecma_script.rs
+++ b/crates/rolldown_utils/src/ecma_script.rs
@@ -7,39 +7,38 @@ pub fn is_validate_identifier_name(name: &str) -> bool {
 
 pub fn legitimize_identifier_name(name: &str) -> Cow<str> {
   let mut legitimized = String::new();
-  let mut have_seen_invalid_char = false;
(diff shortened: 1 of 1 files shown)


crates/rolldown_utils/src/ecma_script.rs
```diff
@@ -7,39 +7,38 @@ pub fn is_validate_identifier_name(name: &str) -> bool {
...
+    let (first_valid_part, rest_part) = name.split_at(first_invalid_char_index);
+    legitimized.push_str(first_valid_part);
+    for char in rest_part.chars() {
+      if identifier::is_identifier_part(char) {
         legitimized.push(char);
+      } else {
+        legitimized.push('_');
       }
-    } else {
-      if !have_seen_invalid_char {
```
hyf0, line 29: In the old impl, `have_seen_invalid_char` is always readed in every iteration while the new impl just read `first_invalid_char_index` once or twice.

IWANABETHATGUY (approved):

Issue #1471, opened 2024-06-28
Using builtin `assert.expect` instead of `console.log` in snapshot testing


IWANABETHATGUY:
use `assert.expect` as possible as we could

- [ ] │   ├── bundler_ts
- [ ]  │   ├── dce 
- [ ]  │   ├── default
- [ ]  │   ├── glob
- [ ] │   ├── import_star
- [ ] │   ├── import_star_ts
- [ ] │   ├── loader
- [ ] │   ├── lower
- [ ] │   ├── packagejson
- [ ] │   └── splitting
├── fixtures
- [ ] │   ├── cjs_compat
- [ ] │   ├── code_splitting
- [ ] │   ├── deconflict
- [ ] │   ├── errors
- [ ] │   ├── function
- [ ] │   ├── issues
- [ ] │   ├── loaders
- [ ] │   ├── misc
- [ ] │   ├── rollup
- [ ] │   ├── tree_shaking
- [ ] │   └── warnings

Boshen:
Can you link the code so you can add a "good first issue" label?

hyf0:
Might wanna consider #1378. We will run our tests using node18 and node20.

IWANABETHATGUY:
> Might wanna consider #1378. We will run our tests using node18 and node20.

Not sure what exactly node18 version do you mean, but there is no compatible issue when I try `node v18.19.0`

![image](https://github.com/rolldown/rolldown/assets/17974631/5a8fd0c9-2eaa-45d2-ae9d-e00f3c7b53d6)

hyf0:
> > Might wanna consider #1378. We will run our tests using node18 and node20.
> 
> Not sure what exactly node18 version do you mean, but there is no compatible issue when I try `node v18.19.0`
> 
[...]

PR #1488, opened 2024-06-29
chore: replace `console.log` as `assert` in snapshot

IWANABETHATGUY:
### Description
1. Part of #1471


hyf0:
Good work! I wanted to do this before. Glad you did it proactively. Should we merge it first to avoid conflicts? You could create another PR for rest parts.

hyf0 (approved):

PR #1496, opened 2024-06-30
perf(rolldown): replace `HASH_PLACEHOLDER_RE` with simple string parsing

### LINKED ISSUE #1480: Regex `HASH_PLACEHOLDER_RE` is bloated
https://github.com/rolldown/rolldown/blob/9ddfe14db2012935e856a5e1ec9d43e907935a74/crates/rolldown/src/utils/extract_hash_pattern.rs#L4

image

Every call allocates too much memory.

Boshen:
closes #1480


hyf0 (approved):
Cool!

PR #1506, opened 2024-07-01
perf(rolldown): build oxc semantic data only once

Boshen:
It was accidental that oxc semantic data is built twice during the whole process. This PR passes the built semantic into the transformer so it is not built twice.

### DIFF
--- a/crates/rolldown/src/utils/make_ast_symbol_and_scope.rs
+++ b/crates/rolldown/src/utils/make_ast_symbol_and_scope.rs
@@ -1,15 +1,18 @@
+use oxc::semantic::{ScopeTree, SymbolTable};
 use rolldown_common::AstScopes;
-use rolldown_oxc_utils::OxcAst;
 
 use crate::types::ast_symbols::AstSymbols;
 
-pub fn make_ast_scopes_and_symbols(ast: &OxcAst) -> (AstScopes, AstSymbols) {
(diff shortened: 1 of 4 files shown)


Boshen:
We probably need to do a big refactoring around the build pipeline, the parse -> transform -> scan pipeline is getting confusing.

crates/rolldown/src/utils/make_ast_symbol_and_scope.rs
```diff
@@ -1,15 +1,18 @@
+use oxc::semantic::{ScopeTree, SymbolTable};
 use rolldown_common::AstScopes;
-use rolldown_oxc_utils::OxcAst;
 
 use crate::types::ast_symbols::AstSymbols;
 
-pub fn make_ast_scopes_and_symbols(ast: &OxcAst) -> (AstScopes, AstSymbols) {
-  let (mut symbol_table, scope) = ast.make_symbol_table_and_scope_tree();
```
Boshen, line 7: symbols and scopes was built here, it's now moved to `pre_process_ast`.

hyf0:
Might need to check out if the `transformer` will ensure the `Semantic` is synced with the transformed AST.

hyf0:
> We probably need to do a big refactoring around the build pipeline, the parse -> transform -> scan pipeline is getting confusing.

Could you expand it more? Which part is considered as confusing?

hyf0 (approved):
Cool. Almost 10% improvement on rome-ts.

PR #1510, opened 2024-07-02
feat: glob import plugin

underfin:
### Description
The pr intend to implement Vite GlobImportPlugin, doc is https://vitejs.dev/guide/features.html#glob-import. 

Before we discussed using `transform` hook return `ast` to do it, it can't implement because we always need to re-parse and re-codegen for it. So here add `transform_ast` hook to do this, if you any ideas, I'm feel free using `transform` hook to do it.

### DIFF
--- a/packages/rolldown/tests/fixtures/builtin-plugin/glob-import/basic/_config.ts
+++ b/packages/rolldown/tests/fixtures/builtin-plugin/glob-import/basic/_config.ts
@@ -1,0 +1,11 @@
+import { BuiltinGlobImportPlugin } from '../../../../../'
+import { defineTest } from '@tests'
+
+export default defineTest({
+  config: {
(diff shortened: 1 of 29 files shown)


packages/rolldown/tests/fixtures/builtin-plugin/glob-import/basic/_config.ts
```diff
@@ -0,0 +1,11 @@
+import { BuiltinGlobImportPlugin } from '../../../../../'
```
hyf0:
```suggestion
import { BuiltinGlobImportPlugin } from 'rolldown'
```

hyf0 (dismissed):

hyf0 (dismissed):

hyf0 (approved):

PR #1530, opened 2024-07-05
refactor: remove `raw_text_to_esm` function.

7086cmd:
My previous pr #1367 includes `raw_text_to_esm`, but it is unnecessary due to the existing function `text_to_esm`, and it passes the test, and is better for normalizing chars.


hyf0 (approved):

hyf0:
Thanks

PR #1541, opened 2024-07-07
chore: adding cargo run-fixture

IWANABETHATGUY:
### Description
1. Move `common` dir in `crates/rolldown/tests` to `rolldown_testing`, cause now we need to dependent on it in `bin/run-fixture`, it is not proper to move `common` to `crates/rolldown`, so move it to `rolldown_testing`
2. Usage: `cargo run-fixture crates/rolldown/tests/fixtures/tree_shaking/advanced_barrel_exports/_config.json`
3. Both relative path and absolute path are supported.


crates/rolldown_testing/src/workspace.rs
```diff
@@ -1,38 +1,9 @@
...
-      .unwrap() // crates
-      .parent()
-      .unwrap() // ${WORKSPACE_ROOT}
-      .to_path_buf();
-
-    // Check if the root directory contains the `pnpm-workspace.yaml` file
-    assert!(root_dir.join("pnpm-workspace.yaml").exists(), "Incorrect root directory detected. Expected to find `pnpm-workspace.yaml` in the root directory. But got {root_dir:?}");
-  };
-  root_dir
+  PathBuf::from(env!("WORKSPACE_DIR"))
```
IWANABETHATGUY, line 6: According to https://github.com/rust-lang/cargo/issues/3946, this method is more concise.

IWANABETHATGUY:
@Boshen Can you have a try, if this tool meets you need?

hyf0:
Looks all fine. Just `cargo shear` check failed. [Details](https://github.com/rolldown/rolldown/actions/runs/9832990969/job/27142576276?pr=1541)

hyf0 (approved):

PR #1545, opened 2024-07-08
fix(sourcemap): avoid remapping if renderChunk hook not return map

underfin:

### DIFF
--- a/crates/rolldown/src/utils/render_chunks.rs
+++ b/crates/rolldown/src/utils/render_chunks.rs
@@ -15,18 +15,16 @@ pub async fn render_chunks<'a>(
       .await
       .map(|(code, render_chunk_sourcemap_chain)| ChunkRenderReturn {
         code,
-        map: if let Some(sourcemap) = chunk.map.as_ref() {
+        map: chunk.map.and_then(|sourcemap| {
           if render_chunk_sourcemap_chain.is_empty() {
-            chunk.map
+            Some(sourcemap)
           } else {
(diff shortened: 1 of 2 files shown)


IWANABETHATGUY (dismissed):

crates/rolldown/src/utils/render_chunks.rs
```diff
@@ -15,15 +15,17 @@ pub async fn render_chunks<'a>(
...
+            let mut sourcemap_chain = Vec::with_capacity(render_chunk_sourcemap_chain.len() + 1);
             sourcemap_chain.push(sourcemap);
+            sourcemap_chain.extend(render_chunk_sourcemap_chain.iter());
+            collapse_sourcemaps(sourcemap_chain)
           }
-          sourcemap_chain.extend(render_chunk_sourcemap_chain.iter());
-          collapse_sourcemaps(sourcemap_chain)
+        } else {
+          None
         },
```
IWANABETHATGUY:
```rs
        map: chunk.map.as_ref().map(|sourcemap| {
          if render_chunk_sourcemap_chain.is_empty() {
            chunk.map
          } else {
            let mut sourcemap_chain = Vec::with_capacity(render_chunk_sourcemap_chain.len() + 1);
            sourcemap_chain.push(sourcemap);
            sourcemap_chain.extend(render_chunk_sourcemap_chain.iter());
            collapse_sourcemaps(sourcemap_chain)
          }
        }),
```

IWANABETHATGUY (approved):

PR #1546, opened 2024-07-08
chore(web):  remove playground

IWANABETHATGUY:
### Description
1. Remove `playground` since we have stackblitz support


Boshen:
The playground can be useful for us to test from main, and also add other debugging capabilities.

IWANABETHATGUY:
> The playground can be useful for us to test from main, and also add other debugging capabilities.

@hyf0 is planning to support the playground via **web container** which could support js plugin, so the current playground is deprecated.

hyf0 (approved):

PR #1550, opened 2024-07-08
refactor(rust/rolldown): rename xx`Id` to xx`Idx` for defined types by `oxc_index`

hyf0:
### Description

The `Id` concept will play an important role in the future and it's not worth to use it just for representing the index of some vectors.


crates/rolldown_common/src/types/module_idx.rs
```diff
@@ -0,0 +1,44 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum ModuleIdx {
+  Normal(EcmaModuleIdx),
+  External(ExternalModuleIdx),
+}
+
+impl ModuleIdx {
+  pub fn expect_ecma(self) -> EcmaModuleIdx {
+    match self {
+      Self::Normal(id) => id,
```
IWANABETHATGUY: Should we call it `Ecma` as well?

hyf0: Good catch.

IWANABETHATGUY (approved):

IWANABETHATGUY:
Looks good, but maybe we should find a better name for the variant of `ModuleIdx`, since `Ecma` may overlap `External` in some scenarios, since some external module is also `ecma` module .

PR #1449, opened 2024-06-27
chore: 🤖 remove todo message

IWANABETHATGUY:

PR #1558, opened 2024-07-09
fix: eliminate unused external module import stmt

IWANABETHATGUY:


hyf0:
Would be good to have a test to ensure the behavior of `treeshake: fasle` too.

IWANABETHATGUY:
> Would be good to have a test to ensure the behavior of `treeshake: fasle` too.

Yeah, I am working on it.

packages/rolldown/tests/fixtures/tree-shake/unused-external-import-stmt/_config.ts
hyf0, line 1: It seems that this test could be written in pure rust side?

crates/rolldown/src/utils/chunk/render_chunk_imports.rs
```diff
@@ -101,11 +101,15 @@ pub fn render_chunk_imports(
 
   imports_from_external_modules.iter().for_each(|(importee_id, named_imports)| {
     let importee = &graph.module_table.external_modules[*importee_id];
+    let external_module_side_effects = &importee.side_effects;
     let mut is_importee_imported = false;
     let mut import_items = named_imports
       .iter()
       .filter_map(|item| {
         let canonical_ref = graph.symbols.par_canonical_ref_for(item.imported_as);
+        if !graph.used_symbol_refs.contains(&canonical_ref) {
```
hyf0, line 110: `use_symbols_refs` should be removed. Will this block https://github.com/rolldown/rolldown/pull/1449 ?

IWANABETHATGUY: I try your method with `stmt_info.is_included`,  and it did not work in all cases. 
e.g.
```js
// index.js
import {a, b} from 'test'

console.log(a)
```
You need to store symbol usage somewhere to eliminate binding `b`, if we remove `used_symbol_refs` how you gonna do that?

hyf0: Good question. I added to my note.

IWANABETHATGUY: Let's merge it first, I am ok to remove all the `used_symbol_refs` if we find a better way to do this.

packages/rolldown/tests/fixtures/tree-shake/unused-external-import-stmt/_config.ts
IWANABETHATGUY: I prefer to put tests under the js side when the case has advanced configurations like `Function`, `Regex` which are hard to support in `json` configuration.

PR #1559, opened 2024-07-09
fix:  `treeshake.moduleSideEffects` treeshake typing

IWANABETHATGUY:


packages/rolldown/src/treeshake/index.ts
```diff
@@ -10,6 +9,7 @@ export const TreeshakingOptionsSchema =
   })
 
 export interface TreeshakingOptions {
```
IWANABETHATGUY, line 11: `TreeshakingOptions` is exposed to userland, so it should be `boolean` here.

PR #1564, opened 2024-07-09
fix: multi declarator tree shake

IWANABETHATGUY:
### Description
for case: 
```
// index.js
import {a,} from './share'
console.log(a)
// share.js
export const a = 1000, b = 1000;
```
#### Before
```
// share.js
const a = 100, b = 1000;
// main1.js
console.log(a);
```

#### After
```
// share.js
const a = 100;
// main1.js
console.log(a);
```


crates/rolldown/tests/fixtures/tree_shaking/multi-declarator/artifacts.snap
```diff
@@ -0,0 +1,51 @@
...
+# Assets
+
+## main.mjs
+
+```js
+import { default as assert } from "node:assert";
+
+//#region foo.js
+const foo = 'foo';
+const [c, d] = [1, 2];
```
IWANABETHATGUY, line 15: Both `c` and `d` is kept, align with **esbuild**, 
Maybe we could support more advanced tree shaking in the future.

packages/rolldown/tests/fixtures/sourcemap/inner/_config.ts
```diff
@@ -25,7 +25,7 @@ export default defineTest({
     const map = JSON.parse(output.output[1].source)
     expect(map.file).toMatchInlineSnapshot(`"main.js"`)
     expect(map.mappings).toMatchInlineSnapshot(
-      `";;AAAO,MAAM,MAAM;;;;ACEnB,QAAQ,IAAI,IAAI"`,
+      `";;AAAA,MAAa,MAAM;;;;ACEnB,QAAQ,IAAI,IAAI"`,
```
IWANABETHATGUY, line 28: I have verified the sourcemap is correct.
image

crates/rolldown/tests/fixtures/tree_shaking/multi-declarator/artifacts.snap
IWANABETHATGUY: https://esbuild.github.io/try/#YgAwLjIwLjIALS1idW5kbGUgLS1taW5pZnk9ZmFsc2UgLS1mb3JtYXQ9ZXNtAGUAZW50cnkuanMACmltcG9ydCB7IGZvbywgYyAgfSBmcm9tICcuL2ZvbycKCgphc3NlcnQuZXF1YWwoZm9vLCAnZm9vJykKYXNzZXJ0LmVxdWFsKGMsIDEpCgAAZm9vLmpzAGV4cG9ydCBjb25zdCBmb28gPSAnZm9vJywgYmFyID0gJ2JhcicKZXhwb3J0IGNvbnN0IFtjLCBkXSA9IFsxLCAyXQo

hyf0 (approved):

PR #1565, opened 2024-07-09
refactor: add `Module` and `ModuleIdx`

hyf0:
### Description

Preparation for supporting css.


hyf0:
* **#1565** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1565?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

hyf0:
Refactoring rolldown to a genernal bundler causes too many changes on the core side. So I'm gonna merge changes as early as it can to reduce conflicts.

crates/rolldown_common/src/module/mod.rs
```diff
@@ -1 +1,94 @@
 pub mod external_module;
+
+use crate::{EcmaModule, ExternalModule, ModuleIdx};
+
+#[derive(Debug)]
+pub enum Module {
```
hyf0, line 6: I didn't use trait here, because we don't have the needs to support custom modules from outside.

Another reason it that exhaustive matching force us to think the relation ship between each module type.

PR #1566, opened 2024-07-10
fix: mixed multi declarator tree shake

IWANABETHATGUY:
### Description
1. Considering this case: 
```js
// index.js
import {c} from './a.js'
console.log(c)
// a.js

export const a = 100, b = 200, [c, d] = [1, 2]
```
var declarator should also be separated.


crates/rolldown/tests/fixtures/tree_shaking/multi-declarator/foo.js
```diff
@@ -1,2 +1,2 @@
-export const foo = 'foo', bar = 'bar'
+export const { baz } = { baz: 'baz' }, foo = 'foo', bar = 'bar';
```
IWANABETHATGUY, line 1: `{baz}` is shaked, as expected.

hyf0: You probably need to consider the potential side-effect when deconstructing from object or array.

---

I don't meat on this case, but case like

```js
const [a, b] = ...

console.log(a)
```

Removing `b` needs to be very careful. I guess we don't need to deal with this case so far.

IWANABETHATGUY: Yeah, looks like esbuild only marks **object destructing** as side effects. https://esbuild.github.io/try/#YgAwLjIwLjIALS1idW5kbGUgLS1taW5pZnk9ZmFsc2UgLS1mb3JtYXQ9ZXNtAGUAZW50cnkuanMAaW1wb3J0IHsgZm9vLCBjfSBmcm9tICcuL2ZvbycKCgphc3NlcnQuZXF1YWwoZm9vLCAnZm9vJykKAABmb28uanMAZXhwb3J0IGNvbnN0IHtiYXp9ID0ge2JhejogJyd9LCBmb28gPSAnZm9vJywgYmFyID0gJ2JhcicKZXhwb3J0IGNvbnN0IFtjLCBkXSA9IFsxLCAyXQ

hyf0: Nah. It's tricky: 

https://esbuild.github.io/try/#YgAwLjIwLjIALS1idW5kbGUgLS1taW5pZnk9ZmFsc2UgLS1mb3JtYXQ9ZXNtAGUAZW50cnkuanMAaW1wb3J0IHsgZm9vLCBjfSBmcm9tICcuL2ZvbycKCgphc3NlcnQuZXF1YWwoZm9vLCAnZm9vJykKAABmb28uanMAY29uc3QgZGVtbyA9IFsxLCAyXQpleHBvcnQgY29uc3Qge2Jhen0gPSB7YmF6OiAnJ30sIGZvbyA9ICdmb28nLCBiYXIgPSAnYmFyJwpleHBvcnQgY29uc3QgW2MsIGRdID0gZGVtbw

IWANABETHATGUY: Yeah, I need to read spec carefully

crates/rolldown/src/ast_scanner/side_effect_detector/mod.rs
```diff
@@ -230,8 +230,13 @@ impl<'a> SideEffectDetector<'a> {
           }
         }
       }
+      let is_destructuring = matches!(
```
IWANABETHATGUY, line 233: destructuring side effects detection is [so complicated](https://github.com/evanw/js-destructuring-fuzzer/tree/master), let's mark them all have side effects for now. fixing https://github.com/rolldown/rolldown/pull/1566/files#diff-f40e0b0989ba3acc13279fe2e59b5fbc4e766fbcc1dca9b25daca07540275ae2R19

crates/rolldown/src/utils/tweak_ast_for_scanning.rs
```diff
@@ -62,7 +62,7 @@ fn split_top_level_variable_declaration<'a>(
         .iter()
         // TODO: support nested destructuring tree shake, `export const {a, b} = obj; export const
         // [a, b] = arr;`
-        .all(|declarator| matches!(declarator.id.kind, BindingPatternKind::BindingIdentifier(_)))
+        .any(|declarator| matches!(declarator.id.kind, BindingPatternKind::BindingIdentifier(_)))
```
IWANABETHATGUY, line 65: If there is no `BindingIdentifier` Declarator found, we don't need to split them at all, since they will not be tree shaken anyway. at least for now.

hyf0 (approved):

PR #1568, opened 2024-07-10
feat(experimental): support strict execution order

hyf0:
### Description

This feature try to solve the execution order problem caused by rollup and esbuild's code splitting logic. The nature of these problem is that the execution order of shared modules are hoisted unexpectedly.

- https://github.com/evanw/esbuild/issues/399
- https://github.com/evanw/esbuild/issues/2598
- https://github.com/vitejs/vite/issues/5142

This is done by adding helper function to simulate the original the semantic esm input while keep the static linking between modules.

### DIFF
--- a/crates/rolldown/src/stages/link_stage/wrapping.rs
+++ b/crates/rolldown/src/stages/link_stage/wrapping.rs
@@ -92,10 +92,7 @@ impl LinkStage<'_> {
       let linking_info = &self.metas[module_id];
 
       let need_to_wrap = self.input_options.experimental.is_strict_execution_order_enabled()
-        || match linking_info.wrap_kind {
-          WrapKind::Cjs | WrapKind::Esm => true,
-          WrapKind::None => false,
-        };
+        || matches!(linking_info.wrap_kind, WrapKind::Cjs | WrapKind::Esm);
 
       if need_to_wrap {
         wrap_module_recursively(
--- a/crates/rolldown/src/stages/link_stage/wrapping.rs
+++ b/crates/rolldown/src/stages/link_stage/wrapping.rs
@@ -91,18 +91,21 @@ impl LinkStage<'_> {
       let module_id = module.idx;
       let linking_info = &self.metas[module_id];
 
-      match linking_info.wrap_kind {
-        WrapKind::Cjs | WrapKind::Esm => {
-          wrap_module_recursively(
-            &mut Context {
-              visited_modules: &mut visited_modules_for_wrapping,
-              linking_infos: &mut self.metas,
-              modules: &self.module_table.modules,
-            },
-            module_id,
-          );
-        }
-        WrapKind::None => {}
+      let need_to_wrap = self.input_options.experimental.is_strict_execution_order_enabled()
+        || match linking_info.wrap_kind {
+          WrapKind::Cjs | WrapKind::Esm => true,
+          WrapKind::None => false,
+        };
+
+      if need_to_wrap {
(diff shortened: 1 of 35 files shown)


hyf0:
* **#1568** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1568?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1567** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1567?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

Boshen:
May I suggest linking some prior arts or articles on what this is?

crates/rolldown/tests/fixtures/function/experimental/strict_execution_order/esbuild_issue_399/strict/init-dep-2.js
```diff
@@ -0,0 +1,3 @@
+global.foo = {
+  log: () => console.log("foo.log() (from entry 2) called"),
```
IWANABETHATGUY, line 2: Can you mock the function and do some assertion? since `console.log` is useless in unit testing, https://sevic.dev/notes/spies-mocking-node-test-runner/, I just found node has builtin support.

hyf0: I was going to, but I don't want change the original input that copied from issues. These tests are just examples to show output. We'll add more serious tests once the feature is going be landed stablely.

IWANABETHATGUY: yeah, just wanna this feature to be stable, you can go ahead and improve in the following pr.

crates/rolldown/src/stages/link_stage/wrapping.rs
```diff
@@ -91,18 +91,21 @@ impl LinkStage<'_> {
...
-              visited_modules: &mut visited_modules_for_wrapping,
-              linking_infos: &mut self.metas,
-              modules: &self.module_table.modules,
-            },
-            module_id,
-          );
-        }
-        WrapKind::None => {}
+      let need_to_wrap = self.input_options.experimental.is_strict_execution_order_enabled()
+        || match linking_info.wrap_kind {
```
IWANABETHATGUY: looks like a simple `matches!` would be more concise.

IWANABETHATGUY (dismissed):
Looks good, maybe we also need to add some benchmark to compare the runtime performance between enabling `strict_execution_order` and disabling `strict_execution_order`

hyf0:
> Looks good, maybe we also need to add some benchmark to compare the runtime performance between enabling `strict_execution_order` and disabling `strict_execution_order`

Yeah. Let's wait for we are sure that this feature is going to be landed in production.

Boshen:
Do we have runtime tests to ensure the strictness? I remember webpack has them by asserting a whole bunch of things.

crates/rolldown_common/src/inner_bundler_options/types/experimental_options.rs
Boshen, line 1: This can go into `main` even if this PR doesn't get merged.

Boshen:
For future reference, @hyf0 verified the reproduction https://github.com/vitejs/vite/issues/5142#issuecomment-953890608 is fixed with the feature in this PR.

Issue #1569, opened 2024-07-10
iife format


underfin:
- only support for single entry(need to clone ast for per entry)
- iife wrapper is added after render_chunks hooks?
- indent chunk content for iife format

- [x] #1623
- [x] iife support `output.name`
- [ ] iife support `output. globals`
- [ ] https://github.com/rolldown/rolldown/issues/1628

hyf0:
> iife wrapper is added after render_chunks hooks?

I think it should be added before `render_chunks` hooks.

7086cmd:
The `output.exports` in IIFE format may be solved in #1691.

celarye:
> * only support for single entry(need to clone ast for per entry)

Is support for multiple entry points for the IIFE format planned? I am not sure how this works in Rollup but I know ESBuild supports it.

The following code panics:

```rs
let files_main: Vec<InputItem> = files
    .iter()
    .map(|f| InputItem {
        name: Some(format!("{}/index", &f)),
        import: format!("{}/src/{}/main.ts", cwd.to_string_lossy(), &f),
    })
    .collect();

let mut bundler = Bundler::new(BundlerOptions {
    input: Some(files_main),
    dir: Some(format!("{}/bundles", cwd.to_string_lossy())),
    name: Some(String::from("source")),
    format: Some(rolldown::OutputFormat::Iife),
    polyfill_require: Some(true),
    minify: Some(RawMinifyOptions::Bool(true)),
    ..Default::default()
});

let _result = bundler.write().await.unwrap();
```

```
thread 'main' panicked at /Users/celarye/.cargo/git/checkouts/rolldown-8b2c76fc50da01ed/d97c9cc/crates/rolldown/src/ecmascript/format/iife.rs:80:6:
iife format only have entry chunk
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

At the moment I need to create a new Bundler for every entry point, which is quite inefficient.

PR #1570, opened 2024-07-10
feat(output): support emit output of iife format

underfin:
### Description

The pr intend to spport iife format, the result of rollup and esbuild has some different.

Here is a [esbuild example](https://esbuild.github.io/try/#YgAwLjIzLjAALS1mb3JtYXQ9aWlmZSAtLWJ1bmRsZSAtLW91dGRpcj0iZGlzdCIAZQBlbnRyeS5qcwBpbXBvcnQoIi4vY2pzIikKaW1wb3J0KCIuL2VzbSIpLnRoZW4oKG0pID0+IG0udmFsdWUpAABjanMuanMAY29uc29sZS5sb2coKQBlAGVzbS5qcwBleHBvcnQgY29uc3QgdmFsdWUgPSAx) and [rollup example](https://rollupjs.org/repl/?version=4.18.1&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQSUyMiUyMiUyQyUyMm1vZHVsZXMlMjIlM0ElNUIlN0IlMjJjb2RlJTIyJTNBJTIyaW1wb3J0KCcuJTJGbW9kdWxlXzEnKSUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTJDJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlN0QlMkMlN0IlMjJjb2RlJTIyJTNBJTIyJTIyJTJDJTIyaXNFbnRyeSUyMiUzQWZhbHNlJTJDJTIybmFtZSUyMiUzQSUyMm1vZHVsZV8xLmpzJTIyJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMm91dHB1dCUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmlpZmUlMjIlN0QlN0QlN0Q=).

Difference: The esbuild will wrapper esm if an inline dynamic module because it needs to execute lazily. The pr also port behavior form esbuild.

For iife format other issues, let us track at [here](https://github.com/rolldown/rolldown/issues/1569)

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/code_splitting.rs
@@ -2,7 +2,13 @@ use std::cmp::Ordering;
 
 use itertools::Itertools;
 use oxc::index::IndexVec;
+<<<<<<< HEAD
 use rolldown_common::{Chunk, ChunkIdx, ChunkKind, ImportKind, Module, ModuleIdx};
+=======
+use rolldown_common::{
+  Chunk, ChunkIdx, ChunkKind, EcmaModuleIdx, EntryPointKind, ImportKind, ModuleIdx, OutputFormat,
+};
+>>>>>>> 076b949a (feat: support format iife for single entry)
 use rolldown_utils::{rustc_hash::FxHashMapExt, BitSet};
 use rustc_hash::FxHashMap;
 
@@ -35,8 +41,16 @@ impl<'a> GenerateStage<'a> {
       if let Module::Ecma(importee) = &self.link_output.module_table.modules[rec.resolved_module] {
         // Module imported dynamically will be considered as an entry,
         // so we don't need to include it in this chunk
+<<<<<<< HEAD
         if !matches!(rec.kind, ImportKind::DynamicImport) {
           self.determine_reachable_modules_for_entry(importee.idx, entry_index, module_to_bits);
+=======
+        if !matches!(rec.kind, ImportKind::DynamicImport)
+      // IIFE format should inline dynamic imports
+          || matches!(self.options.format, OutputFormat::Iife)
+        {
+          self.determine_reachable_modules_for_entry(importee_id, entry_index, module_to_bits);
+>>>>>>> 076b949a (feat: support format iife for single entry)
         }
       }
     });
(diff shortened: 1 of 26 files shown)


crates/rolldown/src/stages/link_stage/mod.rs
```diff
@@ -223,7 +223,37 @@ impl<'a> LinkStage<'a> {
               }
             }
           },
-          ImportKind::DynamicImport => {}
+          ImportKind::DynamicImport => {
+            if matches!(self.input_options.format, OutputFormat::Iife) {
```
underfin, line 227: Here port form https://github.com/evanw/esbuild/blob/main/internal/linker/linker.go#L1432

crates/rolldown/src/stages/link_stage/mod.rs
```diff
@@ -253,7 +283,7 @@ impl<'a> LinkStage<'a> {
             match &self.module_table.modules[rec.resolved_module] {
               Module::External(importee) => {
                 // Make sure symbols from external modules are included and de_conflicted
-                stmt_info.side_effect = importee.side_effects.has_side_effects();
+                stmt_info.side_effect = true;
```
hyf0: @IWANABETHATGUY cc

crates/rolldown/tests/fixtures/format/iife/_config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "config": {
+     "format": "iife"
+  },
+  "expectExecuted": true
```
hyf0: default to be true. No need to set it.

crates/rolldown/src/utils/chunk/render_chunk.rs
```diff
@@ -83,6 +83,11 @@ pub async fn render_chunk(
         }
       }
 
+      // TODO indent chunk content for iife format
```
hyf0, line 86: Does it mean we have to introduce `string_wizard`? If so, we need to rethink the importance of `string_wizard`.

crates/rolldown/src/stages/generate_stage/code_splitting.rs
```diff
@@ -80,6 +90,13 @@ impl<'a> GenerateStage<'a> {
       FxHashMap::with_capacity(self.link_output.entries.len());
     // Create chunk for each static and dynamic entry
     for (entry_index, entry_point) in self.link_output.entries.iter().enumerate() {
+      // IIFE format should inline dynamic imports
```
hyf0: Since dynamic imported modules are inlined, why there still are dynamic entries? Should we just not create dynamic entry in `module_loader`?

crates/rolldown/src/stages/link_stage/mod.rs
underfin: Here is rebase mistaken. Resolved

crates/rolldown/src/utils/chunk/render_chunk.rs
underfin: Yeah. I remember the sourcmap merge behavior is different with `magic-string`,we need to `magic-string-bundle`.

crates/rolldown/src/stages/generate_stage/code_splitting.rs
underfin: Good idea. Done.

crates/rolldown/src/stages/link_stage/mod.rs
IWANABETHATGUY: :ok:

hyf0 (approved):
Nice

PR #1574, opened 2024-07-10
chore: refactor make ast

IWANABETHATGUY:

### DIFF
--- a/crates/rolldown/src/module_loader/runtime_ecma_module_task.rs
+++ b/crates/rolldown/src/module_loader/runtime_ecma_module_task.rs
@@ -23,10 +23,18 @@ pub struct RuntimeEcmaModuleTask {
 
 pub struct RuntimeEcmaModuleTaskResult {
   pub runtime: RuntimeModuleBrief,
-  pub ast_symbol: AstSymbols,
+  pub ast_symbols: AstSymbols,
   pub ast: EcmaAst,
   // pub warnings: Vec<BuildError>,
   pub module: EcmaModule,
+}
+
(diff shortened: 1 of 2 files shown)


crates/rolldown/src/module_loader/runtime_ecma_module_task.rs
```diff
@@ -100,10 +109,7 @@ impl RuntimeEcmaModuleTask {
     Ok(())
   }
 
-  fn make_ast(
```
IWANABETHATGUY, line 102: Using a more specific name, since we may have `make_css_ast` in the future.

crates/rolldown/src/module_loader/runtime_ecma_module_task.rs
```diff
@@ -114,21 +120,21 @@ impl RuntimeEcmaModuleTask {
...
       ModuleDefFormat::EsmMjs,
       source,
       &facade_path,
       &ast.trivias,
     );
-    let namespace_symbol = scanner.namespace_object_ref;
+    let namespace_object_ref = scanner.namespace_object_ref;
     let scan_result = scanner.scan(ast.program());
 
-    Ok((ast, ast_scope, scan_result, symbol_for_module, namespace_symbol))
```
IWANABETHATGUY, line 131: Using a tuple with the length of 5 is a bad idea, Adding a temp struct to store the result.

IWANABETHATGUY:
Self-assigned since it is just a simple refactor, and left 12 hours with no one reply

PR #1576, opened 2024-07-10
perf: avoid unnecessary allocation

IWANABETHATGUY:
### Description
1. Remove unnecessary memory allocation of `include_str!().to_string`

### DIFF
--- a/crates/rolldown/src/module_loader/runtime_ecma_module_task.rs
+++ b/crates/rolldown/src/module_loader/runtime_ecma_module_task.rs
@@ -36,8 +36,7 @@ impl RuntimeEcmaModuleTask {
 
   #[tracing::instrument(name = "RuntimeNormalModuleTaskResult::run", level = "debug", skip_all)]
   pub fn run(self) -> anyhow::Result<()> {
-    let source: Arc<str> =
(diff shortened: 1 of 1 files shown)


hyf0 (approved):

hyf0:
I'm considering using https://docs.rs/arcstr/latest/arcstr/ to replace all usages of `Arc<str>` and `String`. Any suggestion?

IWANABETHATGUY:
> I'm considering using https://docs.rs/arcstr/latest/arcstr/ to replace all usages of `Arc<str>` and `String`. Any suggestion?

I am not familiar with that crate, but it would worth trying if we use the extra feature of `arcstr`.

Boshen:
It’s possible to create a const ArcStr from a literal via the [arcstr::literal!](https://docs.rs/arcstr/latest/arcstr/macro.literal.html) macro. This is probably the killer feature, to be honest.

These “static” ArcStrs are zero cost, take no heap allocation, and don’t even need to perform atomic reads/writes when being cloned or dropped (nor at any other time).

---

Seems like a good choice for this scenario.

IWANABETHATGUY:
known issue of  WASI CI, directly merged

PR #1583, opened 2024-07-11
fix: chunk modules should include empty module

underfin:

### DIFF
--- a/packages/rolldown/tests/fixtures/plugin/write-bundle/_config.ts
+++ b/packages/rolldown/tests/fixtures/plugin/write-bundle/_config.ts
@@ -4,12 +4,14 @@ import type { RolldownOutputChunk } from '../../../../src'
 import { defineTest } from '@tests'
 
 const entry = path.join(__dirname, './main.js')
+const foo = path.join(__dirname, './foo.js')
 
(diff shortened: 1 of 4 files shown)


packages/rolldown/tests/fixtures/plugin/write-bundle/_config.ts
```diff
@@ -25,8 +27,9 @@ export default defineTest({
           expect(chunk.facadeModuleId).toBe(entry)
           expect(chunk.exports.length).toBe(0)
           expect(chunk.imports).length(0)
-          expect(chunk.moduleIds).toStrictEqual([entry])
-          expect(Object.keys(chunk.modules).length).toBe(1)
+          // The `foo.js` should be include `modules/moduleIds` even it is empty.
```
hyf0, line 30: What's the different between empty modules and modules that got tree-shaked? Do tree-shaked modules need to included too?

underfin: The chunk render result is same, only the API need to including empty module, the tree-shaked modules  don't need to included .

hyf0 (approved):

PR #1591, opened 2024-07-11
test: enable more esbuild dce tests

IWANABETHATGUY:
more dce tests case


crates/rolldown/tests/esbuild/dce/.const_value_inlining_bundle/artifacts.snap.new
hyf0, line 1: I think it's ok to keep `.snap` in ignored fixtures, but `.new` files should be removed.

IWANABETHATGUY: still draft, not finished yet

IWANABETHATGUY: Done

Boshen:
Is this using oxc?

hyf0 (approved):

IWANABETHATGUY:
> Is this using oxc?

Some of the esbuild dec tests rely on `oxc` minifier, some are not.
These enabled test cases does not rely on `oxc` minifier

PR #1593, opened 2024-07-12
fix:  align using declaration side effects detection with esbuild.

IWANABETHATGUY:


hyf0 (approved):

crates/rolldown/src/ast_scanner/side_effect_detector/mod.rs
```diff
@@ -248,7 +248,9 @@ impl<'a> SideEffectDetector<'a> {
       Declaration::ClassDeclaration(cls_decl) => self.detect_side_effect_of_class(cls_decl),
       // Currently, using a fallback value to make the bundle correct,
       // finishing the implementation after we carefully read the spec
-      Declaration::UsingDeclaration(_) => true,
+      Declaration::UsingDeclaration(decl) => {
+        decl.is_await || self.detect_side_effect_of_using_declarators(&decl.declarations)
```
IWANABETHATGUY, line 252: According to https://github.com/rolldown/rolldown/blob/07c00a0be85ae53fb0b0a3a61e74afe549705746/crates/rolldown/tests/esbuild/dce/dce_of_using_declarations/entry.js#L1-L1

crates/rolldown/tests/esbuild/dce/dce_of_using_declarations/artifacts.snap
```diff
@@ -0,0 +1,29 @@
...
+using dispose_keep = {[Symbol.dispose]() {
+	console.log('side effect');
+}};
+await using await_asyncDispose_keep = {[Symbol.asyncDispose]() {
+	console.log('side effect');
+}};
+using undef_keep = undefined;
+await using await_undef_keep = undefined;
+const Symbol_dispose_remove = Symbol.dispose;
+const Symbol_asyncDispose_remove = Symbol.asyncDispose;
```
IWANABETHATGUY, line 25: esbuild and rollup consider these two lines are side effects free, but webpack did not. 
Since `Symbol` could be overridden, we better consider these two lines have side effects.

PR #1595, opened 2024-07-12
refactor: compat rollup plugin typing

underfin:
### Description

The pr refactor `FunctionPluginHooks`, because original typing has `name`, it couldn't compat rollup typing for vite. So here port rollup typing to make vite happy.


packages/rolldown/src/plugin/index.ts
```diff
@@ -249,9 +178,54 @@ export type SequentialPluginHooks =
...
+}
+
+export type AddonHookFunction = (
+  this: PluginContext,
+  chunk: RenderedChunk,
+) => string | Promise<string>
+
+export type AddonHook = string | AddonHookFunction
+
+export interface OutputPlugin
```
underfin, line 218: The `OutputPlugin` typing is not used at rolldown, we could remove it at future if user not used.

hyf0 (approved):

packages/rolldown/src/plugin/index.ts
```diff
@@ -249,9 +178,54 @@ export type SequentialPluginHooks =
...
+  // | 'resolveFileUrl'
+  // | 'resolveImportMeta'
+  | 'writeBundle'
 
 export type ParallelPluginHooks = Exclude<
   keyof FunctionPluginHooks | AddonHooks,
   FirstPluginHooks | SequentialPluginHooks
 >
+
+export type PluginHooks = {
```
hyf0, line 200: I don't think we need these complex type functionality before we stabllize the final plugin form, with considering adding filter in it.

- It doesn't simplify anything but adding more reading complexity
- We need to add more mechanism into the plugin system. I kind of hope we we do this kind of abstraction after stablizing the plugin system/form.

PR #1596, opened 2024-07-12
perf(sourcemap): lookup original token in parallel

underfin:


IWANABETHATGUY (approved):

underfin:
The coldspeed test result is also bad, The benchmark comment show improved for it, it is same with my local.

PR #1601, opened 2024-07-13
ci: record compilation time

### LINKED ISSUE #1539: [Infra]: Tracking crate build time.

Avoid introducing `rolldown` crate build time regression unintentionally.

no

IWANABETHATGUY:
### Description
1. Closed https://github.com/rolldown/rolldown/issues/1539

### DIFF
--- a/.github/workflows/metric.yml
+++ b/.github/workflows/metric.yml
@@ -1,0 +1,72 @@
+name: Metric and Commit Results
+
+on:
+  workflow_dispatch:     # Allows manual triggering of the workflow
+  push:
+      branches:
+        - main
+      paths-ignore:
+        - '**/*.md'
+
+jobs:
+  compile-time:
+    name: Compile Time
+    runs-on: ubuntu-latest
+    # if: ${{ github.ref_name == 'main' }}
+    outputs:
+      time: ${{ steps.run.outputs.COMPILE_TIME }}
+    steps:
+      - uses: taiki-e/checkout-action@v1
+      - name: Checkout repo
+        uses: actions/checkout@v4
+      - uses: Boshen/setup-rust@main
+      - run: |
+          cargo fetch
+          cargo clean # build from scatch using local system cache to avoid download from crates.io
+      - id: run
+        run: |
+          /usr/bin/time -f "%e" -o output cargo build --release
+          cat output
+          COMPILE_TIME=$(cat output)
+          echo $COMPILE_TIME
+          echo "COMPILE_TIME=$COMPILE_TIME" >> $GITHUB_OUTPUT
+
+  metric:
+    name: Metric 
+    needs: [compile-time]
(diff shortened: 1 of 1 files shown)


.github/workflows/metric.yml
```diff
@@ -0,0 +1,71 @@
...
+  compile-time:
+    name: Compile Time
+    runs-on: ubuntu-latest
+    # if: ${{ github.ref_name == 'main' }}
+    outputs:
+      time: ${{ steps.run.outputs.COMPILE_TIME }}
+    steps:
+      - uses: taiki-e/checkout-action@v1
+      - name: Checkout repo
+        uses: actions/checkout@v4
```
Boshen, line 20: checkout twice?

.github/workflows/metric.yml
```diff
@@ -0,0 +1,71 @@
...
+      - uses: Boshen/setup-rust@main
```
Boshen, line 21: Need to setup warm cache here to avoid downloading crates.

.github/workflows/metric.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20' # Specify the Node.js version you need
+          cache: 'pnpm'
+
+      - name: Install dependencies and push metric
+        run: |
+          pnpm install -r
```
Boshen, line 58: It seems like the scripts don't need any dependencies installed? Otherwise you're downloading all the bundlers, which is useless.

.github/workflows/metric.yml
IWANABETHATGUY: https://github.com/oxc-project/monitor-oxc/blob/a0f34edc0c745d0a68ee4991e15046d398468548/.github/workflows/ci.yml#L102-L104, just copy your code, :smile: , I will fixed that

.github/workflows/metric.yml
IWANABETHATGUY: https://github.com/rolldown/rolldown/pull/1601/files#diff-34dda2f0b86ed8628bd9b41950351992acbe115340027caa42b64233895e17dcR24, did here.

.github/workflows/metric.yml
IWANABETHATGUY: need install devDependency to execute node script, `node ./scripts/compile-time.mjs`

.github/workflows/metric.yml
Boshen: go check the logs if you aren't sure about the setup.

IWANABETHATGUY: https://github.com/rolldown/testcase/actions/runs/9919913382/job/27406237816, I test in `testcase` repo before.

hyf0 (approved):

PR #1622, opened 2024-07-15
refactor: improve logic of  `render_preliminary_assets` and add more comments

hyf0:

### DIFF
--- a/crates/rolldown/src/ecmascript/ecma_generator.rs
+++ b/crates/rolldown/src/ecmascript/ecma_generator.rs
@@ -53,17 +53,23 @@ impl Generator for EcmaGenerator {
           .into_iter()
           .peekable();
 
-        let maybe_runtime_module = rendered_iter.peek();
-
-        match maybe_runtime_module {
-          Some((id, _, _))
-            if *id == ctx.link_output.runtime.id()
-              && matches!(ctx.options.format, OutputFormat::Cjs) =>
-          {
-            let maybe_runtime_module = rendered_iter.next();
-            if let Some((_, _module_resource_id, Some(emitted_sources))) = maybe_runtime_module {
-              for source in emitted_sources {
-                concat_source.add_source(source);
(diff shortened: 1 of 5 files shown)


hyf0:
* **#1624** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1624?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1622** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1622?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

hyf0:
@underfin The panic is caused here.

https://github.com/rolldown/rolldown/blob/24078fb0d2e060d5c51f6c5d82b569139d886320/crates/rolldown/src/utils/chunk/render_chunk_imports.rs#L107-L161

Need to think of way to handle external modules in iife format.

underfin:
@hyf0  It panic because here has imports, if here remove panic, the user will meet a problem at executing it, because it will have unresolved symbols from imports. I would like panic way, the explicit error would be better than implicit error.

hyf0:
> @hyf0 It panic because here has imports, if here remove panic, the user will meet a problem at executing it, because it will have unresolved symbols from imports. I would like panic way, the explicit error would be better than implicit error.

> the explicit error would be better than implicit error.

Panic with error message "App format doesn't need to generate imports" is not better at all.

The ideal way is first fixing the panic problem and then fix https://github.com/rolldown/rolldown/issues/1623.

underfin:
Yeah. I get it. The vue package panic already fixed at https://github.com/oxc-project/oxc/pull/4268. The external module symbol not removed by dec. We could using better way to resolve it.

hyf0:
I'm gonna merge this PR because it is not negative and not only a solution on panic of iife format but also a improvement on `render_preliminary_assets` function.

PR #1627, opened 2024-07-15
feat: support output.name for iife

underfin:
### Description

 The pr intend to support `output.name` to make vue happy, it also support return value using exports.

Here concat chunk code has some complex, i prefer using rollup like style to process it, see [here](https://github.com/rollup/rollup/blob/master/src/finalisers/index.ts#L36). We will have one corresponding method for every format。


crates/rolldown/src/ecmascript/ecma_generator.rs
```diff
@@ -91,7 +92,15 @@ impl Generator for EcmaGenerator {
 
         // TODO indent chunk content for iife format
         if matches!(ctx.options.format, OutputFormat::Iife) {
-          concat_source.add_source(Box::new(RawSource::new("(function() {\n".to_string())));
+          let has_exports = !get_export_items(ctx.chunk, ctx.link_output).is_empty();
+          if let Some(name) = &ctx.options.name {
+            concat_source.add_source(Box::new(RawSource::new(format!(
+              "var {name} = (function({}) {{\n",
+              if has_exports { "exports" } else { "" }
```
hyf0, line 99: Probably should think about if we follows this behavior? Or we don't introduce special behavior for single default export.

- https://rollupjs.org/repl/?version=4.18.1&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQSUyMiUyMiUyQyUyMm1vZHVsZXMlMjIlM0ElNUIlN0IlMjJjb2RlJTIyJTNBJTIyZXhwb3J0JTIwZGVmYXVsdCUyMDQyJTIyJTJDJTIyaXNFbnRyeSUyMiUzQXRydWUlMkMlMjJuYW1lJTIyJTNBJTIybWFpbi5qcyUyMiU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJvdXRwdXQlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJpaWZlJTIyJTJDJTIybmFtZSUyMiUzQSUyMkElMjIlN0QlN0QlN0Q=
- https://rollupjs.org/repl/?version=4.18.1&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQSUyMiUyMiUyQyUyMm1vZHVsZXMlMjIlM0ElNUIlN0IlMjJjb2RlJTIyJTNBJTIyZXhwb3J0JTIwY29uc3QlMjBhJTIwJTNEJTIwMSU1Q25leHBvcnQlMjBkZWZhdWx0JTIwNDIlMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSUyQyUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMm91dHB1dCUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmlpZmUlMjIlMkMlMjJuYW1lJTIyJTNBJTIyQSUyMiU3RCU3RCU3RA==

underfin: Yeah. It should be consider. Let us one by one to port it.

underfin: Here has some complex behavior we need to aligin it. like https://github.com/rollup/rollup/blob/master/src/finalisers/shared/getExportBlock.ts#L224. We also only add one at cjs format at now.

crates/rolldown_common/src/inner_bundler_options/mod.rs
```diff
@@ -36,6 +36,7 @@ pub struct BundlerOptions {
   pub platform: Option<Platform>,
   pub shim_missing_exports: Option<bool>,
   // --- options for output
+  pub name: Option<String>,
```
hyf0, line 39:
```suggestion
  pub iife_name: Option<String>,
```

`name` is kind of confusing here. We can't change it in package `rolldown` due to compatibility with rollup. But let's make it more meaningful in the core side.

crates/rolldown/src/ecmascript/ecma_generator.rs
hyf0: > Here has some complex behavior we need to aligin it. like https://github.com/rollup/rollup/blob/master/src/finalisers/shared/getExportBlock.ts#L224. We also only add one at cjs format at now.

This goes to https://github.com/rolldown/rolldown/issues/1628. Should have an discussion first to see if we truely need to support it.

crates/rolldown_common/src/inner_bundler_options/mod.rs
underfin: It also will using at `umd/amd`, so here using `name` maybe is good for me. I prefer js options same as rust options, avoid caused some issue.

hyf0: > I prefer js options same as rust options

Reasonable comes first if there are no compatibility issues.

But I didn't thought of `umd/amd`. Let's keep using `name` before we come up with better name.

hyf0 (approved):

PR #1637, opened 2024-07-16
chore: add remapping benchmark

underfin:


hyf0 (approved):

IWANABETHATGUY:
The benchmark case is too simple, considering some cases cost more time, `ns` level benchmark will lead many noises in future `CI`

underfin:
Yeah. It has some problem, sourcemap is empty, I'm fix it.

PR #1634, opened 2024-07-15
feat: throw friendly warning message for ambiguous export

shulaoda:
Closes #1621.
Closes #397.

PR #1641, opened 2024-07-16
refactor:  rename `BuildError` to `BuildDiagnostic`

IWANABETHATGUY:
### Description
1. follow https://github.com/rolldown/rolldown/pull/1634, since we have https://github.com/rolldown/rolldown/pull/1634/files#diff-f68450ab0a8cfe082912fd82beb770be4085872ef174fd398796eb4eff2e8aceR237.


hyf0:
Nice. I'm also thinking about a `DiagnosticStore` to replace raw `errors` `warnings`.

IWANABETHATGUY:
I don't think a simple renaming would lead to performance regression, so merged it first.

> Nice. I'm also thinking about a `DiagnosticSto`

Let's merged it for now, and maybe improve in the future.

PR #1651, opened 2024-07-17
refactor: add abstraction for rendering chunk imports targeting different format

underfin:
### Description

The pr add a `collect_render_chunk_imports` to extract imports info and using it to generate imports for different format.


hyf0 (approved):
Naming could be improved. However, it's positive.

PR #1656, opened 2024-07-18
chore: bump oxc to v0.21.0

Dunqing:
### Description

The changelog is [here](https://github.com/oxc-project/oxc/pull/4338). Most of the snapshots changed due to `codegen` printing double quotes by default. 

Also, in this version, we have aligned the ts ast scope with TypeScript. Many ignored ts tests may have been resolved.


crates/rolldown/tests/fixtures/format/iife/basic/artifacts.snap
```diff
@@ -12,14 +12,14 @@ input_file: crates/rolldown/tests/fixtures/format/iife
 
 
 //#region main.js
-Promise.resolve().then(function() {
+(Promise.resolve()).then(function() {
```
IWANABETHATGUY, line 15: This newly added paren is unexpected, we could merge it first, could you open a new issue in `oxc` ? 
https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=AoJw9gtglgzgpgOhHGYA2A3OAKAlAgFwAs4A7bAMwFdSBjAqMc3AAgG8AoASGQKpFIsA-kIJgAogGUAstmQBHKlGRCKYMHlwBuDgF9tHIA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=Electron-1.8%252CNode-10.13&version=7.24.10&externalPlugins=&assumptions=%7B%7D

crates/rolldown/tests/fixtures/deconflict/wrapped_esm_default_function/artifacts.snap
```diff
@@ -31,9 +31,9 @@ var bar_default = {foo: foo$1};
 //#endregion
 //#region main.js
 const a = 2;
-const { foo } = bar_default;
-assert.strictEqual(typeof foo, 'function');
-(init_foo(), __toCommonJS(foo_ns));
+const { foo: foo } = bar_default;
```
IWANABETHATGUY, line 34: This change is also not expected, https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=MYewdgzgLgBA3jAZiEMC-MC8MBGBDAJwCgig&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=Electron-1.8%252CNode-10.13&version=7.24.10&externalPlugins=&assumptions=%7B%7D

crates/rolldown/tests/fixtures/deconflict/decl_nested_assign_pattern/artifacts.snap
```diff
@@ -11,13 +11,13 @@ input_file: crates/rolldown/tests/fixtures/deconflict/decl_nested_assign_pattern
...
 //#region names.js
-const { foo: { bar: { baz: baz$1 = '' } = {} } = {} } = {foo: {bar: {baz: 'baz2'}}};
+const { foo: { bar: { baz: baz$1 = "" } = {} } = {} } = {foo: {bar: {baz: "baz2"}}};
 
 //#endregion
 //#region main.js
-const { foo: { bar: { baz = '' } = {} } = {} } = {foo: {bar: {baz: 'baz'}}};
-assert.strictEqual(baz, 'baz');
-assert.strictEqual(baz$1, 'baz2');
+const { foo: { bar: { baz: baz = "" } = {} } = {} } = {foo: {bar: {baz: "baz"}}};
```
IWANABETHATGUY, line 18: bad case, https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=KYDwDg9gTgLgBAYwgOwM7wN5wGYQgLjiwCMBDKQk0gLzgF44ByRuAX3qPfYYy461wEicMhWFlqhRhJZ9WAKCA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=Electron-1.8%252CNode-10.13&version=7.24.10&externalPlugins=&assumptions=%7B%7D

crates/rolldown/tests/fixtures/cjs_compat/reexport_commonjs/artifacts.snap
```diff
@@ -36,7 +36,7 @@ var import_commonjs = __toESM(require_commonjs());
 assert.equal(import_commonjs$1.bar, 1);
 assert.equal(value, 1);
 assert.equal(foo_ns.foo, undefined);
-assert.equal(Object.keys(foo_ns).length, 2);
+assert.equal((Object.keys(foo_ns)).length, 2);
```
IWANABETHATGUY, line 39: bad case, https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=IYZxFMCcBcDpwI4FdgBsAUB5ARgK3AMZwDW4AniOgGYD2NA-gHYgCUsq4jA5tABYA0AAgBMLANxA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=Electron-1.8%252CNode-10.13&version=7.24.10&externalPlugins=&assumptions=%7B%7D

crates/rolldown/tests/esbuild/splitting/dynamic_and_not_dynamic_es6_into_es6/artifacts.snap
```diff
@@ -29,7 +29,7 @@ import { bar } from "./foo~1.mjs";
 import { default as assert } from "node:assert";
 
 //#region main.js
-import('./foo.mjs').then(({ bar: b }) => {
+(import("./foo.mjs")).then(({ bar: b }) => {
```
IWANABETHATGUY, line 32: same bad case, like `Promise.then`

crates/rolldown/tests/esbuild/loader/auto_detect_mime_type_from_extension/artifacts.snap
```diff
@@ -1,5 +1,5 @@
 ---
-source: crates/rolldown/tests/common/case.rs
+source: crates/rolldown_testing/src/case/case.rs
```
underfin, line 2: @IWANABETHATGUY I also meet it if snapshot content changes. Could you see see? It should not changed,

crates/rolldown/tests/esbuild/lower/lower_private_method2020/artifacts.snap
```diff
@@ -14,27 +14,27 @@ class Foo {
 	#field;
 	#method() {}
 	baseline() {
-		a().foo;
-		b().foo(x);
-		c()?.foo(x);
-		d().foo?.(x);
-		e()?.foo?.(x);
+		(a()).foo;
```
IWANABETHATGUY, line 17: bad case.

crates/rolldown/tests/esbuild/dce/disable_tree_shaking/artifacts.snap
IWANABETHATGUY, line 1: This is unexpected.

crates/rolldown/tests/fixtures/format/iife/basic/artifacts.snap
Dunqing: Caused by https://github.com/oxc-project/oxc/pull/4245#issue-2407249521

I will send an issue for this later

crates/rolldown/tests/fixtures/deconflict/wrapped_esm_default_function/artifacts.snap
Dunqing: Caused by https://github.com/oxc-project/oxc/pull/4319. I will create an issue to track this

crates/rolldown/tests/esbuild/loader/auto_detect_mime_type_from_extension/artifacts.snap
IWANABETHATGUY: It is expected, the new path is correct, this is because we moved the `case` to another crate, but `cargo insta` only update the snapshot when the content is changed.

IWANABETHATGUY:
Overall it looks good except some bad cases in `call expr chain` and `ObjectExpr` with shorthand, could you track all the bad cases upstream?

IWANABETHATGUY (dismissed):

Cargo.toml
```diff
@@ -139,9 +139,13 @@ urlencoding        = "2.1.3"
 vfs                = "0.12.0"
 xxhash-rust        = "0.8.10"
 # oxc crates share the same version
-arcstr             = "1.2.0"
-lightningcss       = { version = "1.0.0-alpha.57" }
-oxc                = { version = "0.20.0", features = ["sourcemap_concurrent", "transformer", "minifier"] }
+arcstr = "1.2.0"
+lightningcss = { version = "1.0.0-alpha.57" }
+oxc = { git = "https://github.com/oxc-project/oxc", branch = "release/crates-1721271704", features = [
```
IWANABETHATGUY: Can you change to the version, using the git URL would invalidate the cache.

crates/rolldown/tests/esbuild/dce/disable_tree_shaking/artifacts.snap
Dunqing: I wonder why my local can resolve `react/jsx-runtime`

IWANABETHATGUY: rebase the main branch, it should be solved.

IWANABETHATGUY: and, cargo test && cargo insta accept

Dunqing:
The node tests are difficult to find out why it fails, and I'm wondering why we don't print out the `Vitest` error.

hyf0:
> The node tests are difficult to find out why it fails, and I'm wondering why we don't print out the `Vitest` error.

Looks like tests on vitest are all passed. It's just one failed in rollup tests: https://github.com/rolldown/rolldown/actions/runs/9985933904/job/27597619752?pr=1656.

Dunqing:
> Looks like tests on vitest are all passed. It's just one failed in rollup tests: https://github.com/rolldown/rolldown/actions/runs/9985933904/job/27597619752?pr=1656.

Please see https://github.com/rolldown/rolldown/actions/runs/9985756953/job/27597155926

hyf0:
> > Looks like tests on vitest are all passed. It's just one failed in rollup tests: https://github.com/rolldown/rolldown/actions/runs/9985933904/job/27597619752?pr=1656.
> 
> Please see https://github.com/rolldown/rolldown/actions/runs/9985756953/job/27597155926

> Error: Failed in ./fixtures/output/hashFileNames/_config.ts

Run `just test-node rolldown --update` for this one.

> Failed in ./fixtures/tree-shake/module-side-effects-regex/_config.ts

This one needs to be investigate. I'll take look now.

Cargo.toml
Dunqing: Once the test passes or it is determined that the test failure has nothing to do with `oxc`, I will merge the release RP and replace it.

Dunqing:
> > Error: Failed in ./fixtures/output/hashFileNames/_config.ts
> 
> Run `just test-node rolldown --update` for this one.
> 
> > Failed in ./fixtures/tree-shake/module-side-effects-regex/_config.ts
> 
> This one needs to be investigate. I'll take look now.

I already fixed it. https://github.com/rolldown/rolldown/pull/1656/commits/51701d36277dc91058c454202ee56f2968ef1270

hyf0:
I found the cause of the failed rollup test. Not sure how to describle it.

Dunqing:
> I found the cause of the failed rollup test. Not sure how to describle it.

Could you tell me if it is caused by `oxc`?

hyf0:
```js
let o = {
	f() {
		assert.ok(this !== o);
	}
};
(1, o.f)();
(1, o.f)``;

(true && o.f)();
(true && o.f)``;

(true ? o.f : false)();
(true ? o.f : false)``;
```
is bundled to
```js
let o = {f() {
	assert.ok(this !== o);
}};
(1, o.f)();
(1, o.f)``;
o.f();
o.f``;
o.f();
o.f``;
```
```
o.f``;
```
is not the same as

```
(true && o.f)``;
```

---

We used 

https://github.com/rolldown/rolldown/blob/401dd61e7baa3cf5ddec36a386e301ad5ff5859f/crates/rolldown/src/utils/pre_process_ecma_ast.rs#L52

to simplify expressions. I guess this will cause
```
(true && o.f)``;
```
become
```
(o.f)``;
```
which is still ~correct and has the same semantic as~  edited: this is also wrong, see the following output of esbuild
```
(1, o.f)``;
```
.

But this PR oxc finally printed to
```
o.f``;
```
which is wrong. Not sure if we should fix it in the `RemoveDeadCode` or this is a bug of `oxc_codegen`.

---

See the output of esbuld

```
let o={f(){assert.ok(this!==o)}};(0,o.f)(),(0,o.f)``,(0,o.f)(),(0,o.f)``,(0,o.f)(),(0,o.f)``;
```
https://esbuild.github.io/try/#dAAwLjIzLjAALS1taW5pZnkAbGV0IG8gPSB7CglmKCkgewoJCWFzc2VydC5vayh0aGlzICE9PSBvKTsKCX0KfTsKKDEsIG8uZikoKTsKKDEsIG8uZilgYDsKCih0cnVlICYmIG8uZikoKTsKKHRydWUgJiYgby5mKWBgOwoKKHRydWUgPyBvLmYgOiBmYWxzZSkoKTsKKHRydWUgPyBvLmYgOiBmYWxzZSlgYDs

hyf0:
This is the main branch output

```js
let o = {f() {
	assert.ok(this !== o);
}};
(1, o.f)();
(1, o.f)``;
(true && o.f)();
(true && o.f)``;
(true ? o.f : false)();
(true ? o.f : false)``;
```

So it seems we passed this rollup test because we didn't support simplify these expressions before. So they keep the semantics.

hyf0:
This is a classic one. Esbuild has the issue before. https://github.com/evanw/esbuild/issues/2610

Dunqing:
From the description, it looks like the bug comes from `RemoveDeadCode`. 

> This is a classic one. Esbuild has the issue before. https://github.com/evanw/esbuild/issues/2610

Haha, The issue author is also you.

Let's fix it in the next patch

hyf0 (approved):
Nice!

Issue #1174, opened 2024-05-20
Rolldown should throw errors for implicit external modules


zhusjfaker:
Is a module not found in rolldown an external module? 

How to define an external module

from case 
path: /crates/rolldown/tests/esbuild/packagejson/test_package_json_exports_alternatives
test_package_json_exports_alternatives

expect esbuild output in terminal
```sh

✘ [ERROR] Could not resolve "pkg/apples/red.js"

    src/test_package_json_exports_alternatives/src/entry.js:1:21:
      1 │ import redApple from 'pkg/apples/red.js'
        ╵                      ~~~~~~~~~~~~~~~~~~~

  The module "./good-apples/red.js" was not found on the file system:

    src/test_package_json_exports_alternatives/node_modules/pkg/package.json:3:18:
      3 │     "./apples/": ["./good-apples/", "./bad-apples/"],
        ╵                   ~~~~~~~~~~~~~~~~

  You can mark the path "pkg/apples/red.js" as external to exclude it from the bundle, which will
  remove this error and leave the unresolved path in the bundle.

✘ [ERROR] Could not resolve "pkg/books/red"

    src/test_package_json_exports_alternatives/src/entry.js:3:20:
      3 │ import redBook from 'pkg/books/red'
        ╵                     ~~~~~~~~~~~~~~~

  The module "./good-books/red-book.js" was not found on the file system:

    src/test_package_json_exports_alternatives/node_modules/pkg/package.json:4:18:
      4 │     "./books/*": ["./good-books/*-book.js", "./bad-books/*-book.js"]
        ╵                   ~~~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "pkg/books/red" as external to exclude it from the bundle, which will remove
  this error and leave the unresolved path in the bundle.

```

but now rolldown will bundle process result is successful 

dist/entry.js

```js

// These modules are treated as external. not throw any error or panic 
import { default as redApple } from "pkg/apples/red.js";
import { default as redBook } from "pkg/books/red";

// node_modules/pkg/good-apples/green.js
var green_default = '\u{1f34f}';

// node_modules/pkg/good-books/green-book.js
var green_book_default = '\u{1f4d7}';

// src/entry.js
console.log({
	redApple,
	greenApple:green_default,
	redBook,
	greenBook:green_book_default
});

```

Cause analysis

crates/rolldown_plugin/src/utils/resolve_id_with_plugins.rs
```rust
// line 72
[...]

PR #1659, opened 2024-07-18
fix:  external warning in dce/disable-tree-shaking

IWANABETHATGUY:


crates/rolldown/tests/esbuild/dce/disable_tree_shaking/artifacts.snap
```diff
@@ -3,14 +3,6 @@ source: crates/rolldown_testing/src/case/case.rs
 expression: content
 input_file: crates/rolldown/tests/esbuild/dce/disable_tree_shaking
 ---
-# warnings
```
hyf0, line 6: It seems people tend to ignore warnings. Thinking about should we turn this into errors:

https://github.com/rolldown/rolldown/issues/1174

PR #1667, opened 2024-07-19
chore: add minify benchmark

underfin:
### Description

The pr add the minify benchmark and minify-sourcemap benchmark. Here remove `scan-xx-sourcemap` benchmark, because it's sourcmap is unmeaning, it will not generate sourcemap.

Also add it to see for https://github.com/rolldown/rolldown/pull/1664. i think render chunk remapping case is small..


hyf0 (approved):
Nice.

PR #1668, opened 2024-07-19
fix: correct the behavior in texture or empty extension-customized file.

7086cmd:
#### Overview
This modification enhances the `Data URL` functionality to ensure better MIME type conformity.

#### Changes

1. **MIME Type Adjustment**:
   - The MIME type detection for texture files has been improved. Previously, files like `text.data` with content `a\x00b\x80c\xFFd` would be recognized as `application/octet-stream` by the `get_mime` function. According to [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#textplain), unspecified texture file subtypes should be classified as `text/plain`, not `application/octet-stream`.

2. **Additional Testing Scenarios**:
   - New test cases have been added to cover various scenarios, including `abnormal_extension`, `binary`, `custom_content`, `empty`, and `text`.
     - ~~**`abnormal_extension`**: A 1x1 image was generated using the following Python code and manually renamed with a `.custom` extension. The file is small and does not significantly impact storage.~~

       <details><summary>Generating Code</summary>

       ```python
       import cv2
       import numpy as np

       image = np.zeros((1, 1, 3), np.uint8)
       image[:] = (128, 128, 128)
       cv2.imwrite('image.png', image)
       ```
       </details>
     
     - **`binary`, `empty`, and `text`**: These test cases are adapted from existing `base64` or `binary` tests. Notably, the `text` test case uses URL-encoded content, which differs from the output of esbuild. According to [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs#datatexthtml3cscript3ealert2827hi27293b3c2fscript3e), tags containing `<` should be transformed, which esbuild does not currently do. While URL encoding increases the file size, a better solution needs to be found.
     
     - ~~**`custom_content`**: This test case involves both customized extension and content, ensuring it falls under `application/octet-stream` as a binary file.~~

       <details><summary>Generating Code</summary>

       ```python
       import os

       def generate_random_file(file_path, size):
           with open(file_path, 'wb') as f:
               f.write(os.urandom(size))

       file_path = 'random_data_file.bin'
       file_size = 8

       generate_random_file(file_path, file_size)
       ```
       </details>

      Removed because GitHub doesn't support the file whose type can't be recognised.

> Note: The code snippets provided are for validation purposes only.


crates/rolldown/tests/fixtures/module_types/dataurl/abnormal_extension/_config.json
hyf0, line 1: Could you also verify the behavior of esbuild about guessing mime for these cases? Just check out if we are the same or better.

7086cmd: Screenshot 2024-07-20 at 01 00 56
The material is the `text` test dataset.

7086cmd:
```
(base) ethangoh@Ethans-MacBook-Pro text % esbuild main.js --bundle --loader:.data=dataurl --platform=node --format=esm
// main.js

// text.data
var text_default = "data:text/plain;charset=utf-8,\"'`+=/@%23$$%^&*(%0A\u4F60\u597D\uFF0C\u4E16\u754C%0A\u041F\u0440\u0438\u0432\u0435\u0442 \u043C\u0438\u0440%0A\u3053\u3093\u306B\u3061\u306F\u4E16\u754C%0A<script>";

// main.js
assert(typeof text_default === "string" && text_default.startsWith("data:text/plain;charset=utf-8,"));
```

The `text` data.

7086cmd:
```
(base) ethangoh@Ethans-MacBook-Pro empty % esbuild main.js --bundle --loader:.empty=dataurl --platform=node --format=esm
// main.js

// empty.empty
var empty_default = "data:text/plain;charset=utf-8,";

// main.js
assert(empty_default === "data:text/plain;charset=utf-8,");
```

The `empty` data.

7086cmd: The function invoked by `esbuild` that returns `text/plain` can be found [here](https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/net/http/sniff.go;l=194). Previously, our function's performance was suboptimal compared to others. However, it now aligns with the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#text).

7086cmd:
Strange enough. I passed the test locally, and it worked. In the repo, the binary file is different from the local one.

7086cmd:
The function invoked by esbuild that returns text/plain can be found [here](https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/net/http/sniff.go;l=194). Previously, our function's performance was suboptimal compared to others. However, it now aligns with the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#text).

hyf0 (approved):
Thank you!

PR #1675, opened 2024-07-21
fix: reexport star of a external module should be considered have side effect.

IWANABETHATGUY:


packages/rolldown/tests/fixtures/tree-shake/external-export-star1/main.js
```diff
@@ -0,0 +1 @@
+export * from 'fs'
```
IWANABETHATGUY: Before we would remove this statement since it is considered side effect free if the user specifies `treeshake.moduleSideEffects: false`

packages/rolldown/tests/fixtures/tree-shake/external-export-star1/_config.ts
```diff
@@ -0,0 +1,20 @@
...
+      moduleSideEffects: false,
+    },
+    external: ['fs'],
+  },
+  afterTest: (output) => {
+    output.output
+      .filter(({ type }) => type === 'chunk')
+      .forEach((chunk) => {
+        let code = (chunk as RolldownOutputChunk).code
+        expect(code.includes(`import * as import_fs from "fs"`))
```
hyf0: What's the blocker for writing this test in rust?

IWANABETHATGUY: https://github.com/rolldown/rolldown/pull/1558#discussion_r1670548422, just personal preference. Also, update I would not like to add too many snapshots because it would be some nightmare when you have some changes on codegen.

hyf0: This looks like manually snapshot. Beyond personal preferences, rust side test is easily to be debugging and has faster development feedback considering no need to compile the binding, so we should always prefer rust side tests if possible.

Tests of nodejs is only for testing API or config that can't represent in pure rust.

IWANABETHATGUY: https://github.com/rolldown/rolldown/blob/a3ada216a0d8a3404bdfde5feeb18c7c07a18f0c/crates/rolldown_common/src/inner_bundler_options/types/treeshake.rs#L3-L7, the deserialize is not easy to write for `treeshake`(especially when we fully support the configuration it may have function), at that moment, we will have half of the tree shake option related test in rust side, and another half in js side, so I still prefer writing test case in js

hyf0: Yeah. I understand your preferences.

> Tests of nodejs is only for testing API or config that can't represent in pure rust.

However, this is the intension. If there aren't blockers, please write pure rust test first.

underfin: I also need it. see https://github.com/rolldown/rolldown/pull/1681

hyf0 (approved):

crates/rolldown_common/src/inner_bundler_options/mod.rs
```diff
@@ -105,11 +106,24 @@ fn deserialize_treeshake<'de, D>(deserializer: D) -> Result<TreeshakeOptions, D:
 where
   D: Deserializer<'de>,
 {
-  let deserialized = Option::<bool>::deserialize(deserializer)?;
-  match deserialized {
-    Some(false) => Ok(TreeshakeOptions::False),
-    Some(true) | None => Ok(TreeshakeOptions::Option(types::treeshake::InnerOptions {
-      module_side_effects: types::treeshake::ModuleSideEffects::Boolean(true),
-    })),
+  let value = Option::<Value>::deserialize(deserializer)?;
```
hyf0, line 109: Not for this PR. But We could consider doing this https://github.com/rolldown/rolldown/pull/1681#discussion_r1686188850 in the future.

crates/rolldown_common/src/inner_bundler_options/mod.rs
```diff
@@ -76,7 +76,7 @@ pub struct BundlerOptions {
   #[cfg_attr(
     feature = "deserialize_bundler_options",
     serde(deserialize_with = "deserialize_treeshake", default),
-    schemars(with = "Option<bool>")
+    schemars(skip)
```
hyf0: If things still go wrong, just change the defnation to https://github.com/rolldown/rolldown/pull/1681#discussion_r1686188850 and everything should be ok to be auto `derived`.

crates/rolldown_common/src/inner_bundler_options/mod.rs
hyf0: We have used the generated shchema to validate options:

https://github.com/rolldown/rolldown/blob/2d2974e7b185b63cd7593093114e815e77b0d2a3/crates/rolldown_testing/src/test_config.rs#L39

crates/rolldown_common/src/inner_bundler_options/mod.rs
hyf0: We have used the shchema to validate options

https://github.com/rolldown/rolldown/blob/2d2974e7b185b63cd7593093114e815e77b0d2a3/crates/rolldown_testing/src/test_config.rs#L39

PR #1681, opened 2024-07-22
chore: deserialize TreeshakeOptions

underfin:
### Description

Before we can't add ` tree_shake: { modulesideEffects: false }` config, the pr intend to support it. The default sturcture is ugly, but it could be work at now.

### DIFF
--- a/crates/rolldown/tests/fixtures/function/external/commonjs_reexport_external/_config.json
+++ b/crates/rolldown/tests/fixtures/function/external/commonjs_reexport_external/_config.json
@@ -4,6 +4,6 @@
     "external": [
       "external"
     ],
-    "treeshake": false
+    "treeshake": "False"
   }
 }
--- a/crates/rolldown/tests/fixtures/tree_shaking/pure_annotation/_config.json
+++ b/crates/rolldown/tests/fixtures/tree_shaking/pure_annotation/_config.json
@@ -2,7 +2,7 @@
   "config": {
     "treeshake": {
       "Option": {
-        "module_side_effects": true
+        "moduleSideEffects": true
       }
     }
   }
--- a/crates/rolldown/tests/fixtures/tree_shaking/pure_annotation/_config.json
+++ b/crates/rolldown/tests/fixtures/tree_shaking/pure_annotation/_config.json
@@ -1,5 +1,9 @@
 {
   "config": {
-    "treeshake": true
+    "treeshake": {
+      "Option": {
+        "module_side_effects": true
+      }
+    }
   }
 }
--- a/crates/rolldown_common/src/inner_bundler_options/types/treeshake.rs
+++ b/crates/rolldown_common/src/inner_bundler_options/types/treeshake.rs
@@ -40,7 +40,11 @@ impl TreeshakeOptions {
 }
 
 #[derive(Debug)]
-#[cfg_attr(feature = "deserialize_bundler_options", derive(Deserialize, JsonSchema))]
+#[cfg_attr(
+  feature = "deserialize_bundler_options",
+  derive(Deserialize, JsonSchema),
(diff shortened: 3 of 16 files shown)


crates/rolldown/tests/fixtures/tree_shaking/pure_annotation/_config.json
```diff
@@ -1,5 +1,9 @@
 {
   "config": {
-    "treeshake": true
+    "treeshake": {
+      "Option": {
+        "module_side_effects": true
+      }
+    }
```
underfin, line 7: Usage for 

```
 "treeshake": {
       "module_side_effects": true
    }
```

crates/rolldown/tests/fixtures/function/external/commonjs_reexport_external/_config.json
```diff
@@ -4,6 +4,6 @@
     "external": [
       "external"
     ],
-    "treeshake": false
+    "treeshake": "False"
```
underfin, line 7: Usage for 

```
 "treeshake": false
```

crates/rolldown_common/src/inner_bundler_options/types/treeshake.rs
```diff
@@ -1,6 +1,11 @@
 use crate::types::js_regex::HybridRegex;
+#[cfg(feature = "deserialize_bundler_options")]
+use schemars::JsonSchema;
+#[cfg(feature = "deserialize_bundler_options")]
+use serde::{Deserialize, Deserializer};
 
 #[derive(Debug)]
+#[cfg_attr(feature = "deserialize_bundler_options", derive(Deserialize, JsonSchema))]
 pub enum TreeshakeOptions {
   False,
```
hyf0, line 10: Not for this PR. But We don't need to align the `False` with the rollup's API. Adding filed like `disabled` in `InnerOptions` and using `InnerOptions` directly as the `TreeshakeOptions` won't be a problem.

We should considering if we expose rolldown as a crate, are the exposed options friendly to users.

crates/rolldown/tests/fixtures/tree_shaking/pure_annotation/_config.json
```diff
@@ -1,5 +1,9 @@
 {
   "config": {
-    "treeshake": true
+    "treeshake": {
+      "Option": {
+        "module_side_effects": true
```
hyf0: Please enable the `camelCase` for consistency.

crates/rolldown_common/src/inner_bundler_options/types/treeshake.rs
hyf0: I also think
```rust
enum TreeshakeOptions {
  Bool(bool),
  Option(InnerOptions),
}
```
is fine.

Even though `true` might be meaningless. But considering adde code for dealing with `False`, meaningless `true` is not a big deal.

underfin: Yeah. I agree with you, let's keep the this at the pr, if we need to refactor the API open a new pr for it. cc @IWANABETHATGUY

crates/rolldown/tests/fixtures/tree_shaking/pure_annotation/_config.json
underfin: Done

IWANABETHATGUY:
I just implement this already, and I think my implementation has better scability.

hyf0 (dismissed):

IWANABETHATGUY:
https://github.com/rolldown/rolldown/pull/1675/files#diff-fef0515e8da9bf5d47508ef403741c46e31b0e52647ad29efd4c037df8e42ecb

underfin:
Ok. Let's merge one. I need it.

PR #1683, opened 2024-07-22
fix: should include the dependencies if the module only has re-export…

underfin:
… statements

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/code_splitting.rs
@@ -23,7 +23,7 @@ impl<'a> GenerateStage<'a> {
     let meta = &self.link_output.metas[module_id];
 
     // Here if the module only has re-export statements, the `is_included` is false at `module_side_effects: false`.
-    // But we also need to include it's dependencies.
+    // But we also need to determine it's dependencies.
     if module.is_included {
       if module_to_bits[module_id].has_bit(entry_index) {
         return;
--- a/crates/rolldown/src/stages/generate_stage/code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/code_splitting.rs
@@ -22,14 +22,14 @@ impl<'a> GenerateStage<'a> {
     };
     let meta = &self.link_output.metas[module_id];
 
-    if !module.is_included {
-      return;
+    // Here if the module only has re-export statements, the `is_included` is false at `module_side_effects: false`.
+    // But we also need to include it's dependencies.
+    if module.is_included {
+      if module_to_bits[module_id].has_bit(entry_index) {
+        return;
+      }
+      module_to_bits[module_id].set_bit(entry_index);
     }
-
-    if module_to_bits[module_id].has_bit(entry_index) {
-      return;
-    }
-    module_to_bits[module_id].set_bit(entry_index);
 
     module.import_records.iter().for_each(|rec| {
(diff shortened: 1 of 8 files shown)


crates/rolldown/src/stages/generate_stage/code_splitting.rs
```diff
@@ -22,14 +22,14 @@ impl<'a> GenerateStage<'a> {
     };
     let meta = &self.link_output.metas[module_id];
 
-    if !module.is_included {
-      return;
-    }
-
-    if module_to_bits[module_id].has_bit(entry_index) {
-      return;
+    // Here if the module only has re-export statements, the `is_included` is false at `module_side_effects: false`.
```
hyf0, line 25: Is this the cause of unexpectedly generated common chunks on bundling vue?

crates/rolldown/src/stages/generate_stage/code_splitting.rs
```diff
@@ -22,14 +22,14 @@ impl<'a> GenerateStage<'a> {
...
+    // But we also need to include it's dependencies.
+    if module.is_included {
+      if module_to_bits[module_id].has_bit(entry_index) {
+        return;
+      }
+      module_to_bits[module_id].set_bit(entry_index);
```
underfin, line 31: Because the `report-star.js` is_included is false, so here need to determine it's dependencies. If we could be change is_included to be true, we don't need this, i'm not sure for it. @IWANABETHATGUY

crates/rolldown/src/stages/generate_stage/code_splitting.rs
underfin: No. It is a another panic bug.

crates/rolldown/src/stages/generate_stage/code_splitting.rs
hyf0: https://github.com/rolldown/rolldown/blob/9719b2a61a0c43ef6c4f0aa08cef12935ab2b15d/crates/rolldown/src/stages/link_stage/tree_shaking.rs#L84-L100

Related code.

hyf0: Where does it panic before this PR? I feel we shouldn't included this module but we should also not panic.

underfin: See crates/rolldown/src/stages/generate_stage/code_splitting.rs:136:7

crates/rolldown/tests/fixtures/tree_shaking/export-star/artifacts.snap.new
```diff
@@ -0,0 +1,21 @@
+---
```
IWANABETHATGUY: I don't see any difference compared to branch `fix/export-star-external`

underfin: The `export-star` is not external..

crates/rolldown/src/stages/generate_stage/code_splitting.rs
hyf0: Get it. I'll consider this as a workaround fix. This is more of a bug of linking.

crates/rolldown/src/stages/generate_stage/code_splitting.rs
hyf0: It's probably related/

hyf0 (approved):

PR #1690, opened 2024-07-23
chore: remove unused wasi files

underfin:


IWANABETHATGUY:
Maybe you should remove these either, https://github.com/rolldown/rolldown/blob/babbd8996246672adfb7dad623e34f57827433b8/.prettierignore#L14-L17

IWANABETHATGUY (approved):

PR #1680, opened 2024-07-22
feat: support `output.exports` option.

7086cmd:
### Description

resolves #1628

This pull request addresses the `output.exports` option related to [Rollup](https://rollupjs.org/configuration-options/#output-exports).

The PR does not yet address deconfliction; I will implement it if needed. (Rollup uses `getExportMode`, which is `determine_export_mode` in this PR, during deconfliction).

### Details

1. Added options in both the Rust and TypeScript modules.
2. Introduced the function `determine_export_mode` to handle the `auto` value. It is adapted from Rollup but uses `match` instead of an `if` chain.
3. Managed `export_mode` in `crates/rolldown/src/utils/chunk/render_chunk_exports.rs`.

### Issues Waiting for Fixes in this PR

1. Remove the duplicated enums `ExportMode` and `OutputExports`, even though `OutputExports` includes `Auto` while `ExportMode` does not.
2. Improve support for the `iife` format. The `iife` format is still unstable in this project, and it's uncertain if it can handle the `output.exports` issue for `iife`. However, a prototype has been provided.
3. Used `anyhow::bail` for error handling. The error messages are not clear enough and need improvement. Contributors are welcome to address this, and `TODO` comments have been added to the relevant sections.

This is the first time I have contributed to the core code. I am willing to fix problems in this PR promptly! :-)

PR #1691, opened 2024-07-23
feat: support `output.exports` for `IIFE` format align with rollup.

7086cmd:
In PR #1680, support for `output.exports` was only available for the `Cjs` format. This PR extends the support to the `IIFE` format.

### Modifications

1. **Wrapper Function Update:**
   - In `crates/rolldown/src/ecmascript/format/iife.rs`, the declaration part of the wrapper function has been rewritten. The `exports` argument is now enabled only if `output.exports` is `named`, and the non-name mode issue has been fixed. Similar changes have been made to the ending part of the wrapper function.

2. **Return Value Alignment:**
   - In `crates/rolldown/src/utils/chunk/render_chunk_exports.rs`, the value is now returned directly instead of assigning it to `exports`, aligning with Rollup’s behavior when `output.exports` is `default` and the `format` is `iife`.

3. **Removal of `__esModule` Property:**
   - The statement `Object.defineProperty(exports, '__esModule', { value: true });` has been removed when `output.exports` is `default` and the `format` is either `cjs` or `iife`, as it previously caused errors in some cases.

4. **Testing:**
   - Added tests for the `iife` format to ensure correct handling of `output.exports`.

These changes ensure better alignment with Rollup and improve the robustness of the `IIFE` format support in the project.


7086cmd:
In the `FIXME` comment in `iife.rs`, I suggest that the `determine...` function could be called from an outer function. However, I am still figuring out the best location for this function and the most elegant way to pass the necessary arguments.

underfin (approved):

underfin:
Awesome, thank you again.

PR #1692, opened 2024-07-23
feat: add invalid export option to generate diagnostic

underfin:
### Description

The pr implement two things.

1. Add diagnostic to `GenerateContext`, it used for collect diagnostic at generate stage.
2. Collect InvalidExportOption errors.

### DIFF
--- a/crates/rolldown/src/ecmascript/ecma_generator.rs
+++ b/crates/rolldown/src/ecmascript/ecma_generator.rs
@@ -7,6 +7,7 @@ use anyhow::Result;
 use rolldown_common::{
   AssetMeta, EcmaAssetMeta, ModuleId, ModuleIdx, OutputFormat, PreliminaryAsset, RenderedModule,
 };
+use rolldown_error::DiagnosableResult;
 use rolldown_plugin::HookBannerArgs;
 use rolldown_sourcemap::Source;
 use rolldown_utils::rayon::{IntoParallelRefIterator, ParallelIterator};
@@ -21,7 +22,9 @@ pub struct EcmaGenerator;
 
 impl Generator for EcmaGenerator {
   #[allow(clippy::too_many_lines)]
-  async fn render_preliminary_assets<'a>(ctx: &mut GenerateContext<'a>) -> Result<GenerateOutput> {
+  async fn render_preliminary_assets<'a>(
+    ctx: &mut GenerateContext<'a>,
+  ) -> Result<DiagnosableResult<GenerateOutput>> {
     let mut rendered_modules = FxHashMap::default();
 
     let rendered_module_sources = ctx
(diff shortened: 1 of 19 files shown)


crates/rolldown/src/ecmascript/ecma_generator.rs
```diff
@@ -21,9 +21,7 @@ pub struct EcmaGenerator;
 
 impl Generator for EcmaGenerator {
   #[allow(clippy::too_many_lines)]
-  async fn render_preliminary_assets<'a>(
-    ctx: &GenerateContext<'a>,
-  ) -> Result<Vec<PreliminaryAsset>> {
+  async fn render_preliminary_assets<'a>(ctx: &mut GenerateContext<'a>) -> Result<GenerateOutput> {
```
hyf0:
```suggestion
  async fn render_preliminary_assets<'a>(ctx: &mut GenerateContext<'a>) -> Result<DiagnosableResult<GenerateOutput>> {
```

`.errors:` is not the recommended pattern anymore. It could not break the control flow.

If there are errors, return early. Don't pretend everything is ok.

crates/rolldown/src/stages/generate_stage/render_chunk_to_assets.rs
```diff
@@ -23,9 +24,11 @@ impl<'a> GenerateStage<'a> {
   pub async fn render_chunk_to_assets(
     &mut self,
     chunk_graph: &mut ChunkGraph,
-  ) -> anyhow::Result<Vec<Output>> {
+  ) -> anyhow::Result<BundleOutput> {
+    let mut errors = std::mem::take(&mut self.link_output.errors);
```
hyf0, line 28: https://github.com/rolldown/rolldown/pull/1692/files#r1687734703

crates/rolldown/src/ecmascript/ecma_generator.rs
underfin: Good idea. Let me update it.

crates/rolldown/src/ecmascript/ecma_generator.rs
```diff
@@ -106,18 +104,22 @@ impl Generator for EcmaGenerator {
...
+        augment_chunk_hash: None,
+        file_dir: file_dir.to_path_buf(),
+        preliminary_filename: ctx
+          .chunk
+          .preliminary_filename
+          .clone()
+          .expect("should have preliminary filename"),
+      }],
+      errors: std::mem::take(&mut ctx.errors),
+      warnings: std::mem::take(&mut ctx.warnings),
```
hyf0, line 133: Since we already stored the warnings in the context. Why do we need to take it out and return it in other places?

underfin: Here is not same warnings. The render_chunk_to_assets warnings come from link output. Here render_preliminary_assets is run in parallel, we can't using mutable reference of `LinkStageOutput#warnings`.

hyf0 (approved):

PR #1693, opened 2024-07-23
feat(experimental): support `enableComposingJsPlugins` option

hyf0:
### Description

I don't add tests since it's quite experimental. But we should think of how to add tests to test such behavior.


hyf0:
* **#1693** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1693?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

hyf0:
Could we have a performance test on rolldown-vite and your benchmark case? @IWANABETHATGUY

underfin (approved):

IWANABETHATGUY:
> Could we have a performance test on rolldown-vite and your benchmark case? @IWANABETHATGUY

I would have a try

packages/rolldown/src/utils/compose-js-plugins.ts
```diff
@@ -0,0 +1,276 @@
...
+    name: `Composed(${names.join(', ')})`,
+  }
+
+  ;(Object.keys(batchedHooks) as (keyof typeof batchedHooks)[]).forEach(
+    (hookName) => {
+      switch (hookName) {
+        case 'buildStart': {
+          if (batchedHooks.buildStart) {
+            const batchedHandlers = batchedHooks.buildStart
+            composed.buildStart = async function (options) {
```
IWANABETHATGUY, line 118: What if the user expects the `buildStart` hook to execute linearly? This may generate different behavior

hyf0: This hook is intend to be async parallel: 
image
.

packages/rolldown/src/utils/compose-js-plugins.ts
```diff
@@ -0,0 +1,276 @@
+import { Plugin, RolldownPlugin } from '../plugin'
+import _ from 'lodash-es'
```
yyx990803, line 2: This doesn't seem to be used at all?
Also, even if lodash is needed, we should use named import to avoid bundling all its methods.

packages/rolldown/src/utils/compose-js-plugins.ts
```diff
@@ -0,0 +1,276 @@
...
+        case 'generateBundle':
+        case 'moduleParsed':
+        case 'onLog':
+        case 'options':
+        case 'outputOptions':
+        case 'renderError':
+        case 'renderStart':
+        case 'resolveDynamicImport':
+        case 'writeBundle': {
+          throw new Error(
```
yyx990803, line 95: - Do we plan to support more of these in the future, or are some of them not possible to compose?
- Is it possible to compose as much as we can and only skip the incompatible plugins?

packages/rolldown/src/utils/compose-js-plugins.ts
IWANABETHATGUY: got it.

packages/rolldown/src/utils/compose-js-plugins.ts
```diff
@@ -0,0 +1,276 @@
...
+          }
+          break
+        }
+        case 'load': {
+          if (batchedHooks.load) {
+            const batchedHandlers = batchedHooks.load
+            composed.load = async function (id) {
+              for (const handler of batchedHandlers) {
+                const [handlerFn, _handlerOptions] = normalizeHook(handler)
+                const result = await handlerFn.call(this, id)
```
yyx990803, line 135: We should check the return value of `handlerFn` and only await when it is actually a Promise. The extra `await` tick is actually costly in hot code paths. This applies to all serial hooks.

Vue SSR got a nice perf boost by only `await` when necessary: https://github.com/vuejs/core/commit/05779a70bd0b567ae458a07636d229bd07c44c4e

packages/rolldown/src/utils/compose-js-plugins.ts
hyf0: Yes. It's not used🤡. Rolldown supports treeshaking on `_.x`. I guess it would be fine.

packages/rolldown/src/utils/compose-js-plugins.ts
hyf0: > Do we plan to support more of these in the future, or are some of them not possible to compose?

Yeah. I don't see any hook which is not composable yet. There are going to be supported in the future.

> Is it possible to compose as much as we can and only skip the incompatible plugins?

Yeah. We are doing it now. https://github.com/rolldown/rolldown/blob/8aad7d0fc33c4ca991352b86362b0c0fa383944c/packages/rolldown/src/utils/compose-js-plugins.ts#L250-L275

packages/rolldown/src/utils/compose-js-plugins.ts
hyf0: That's an very nice idea. I'll look into it.

packages/rolldown/src/utils/compose-js-plugins.ts
IWANABETHATGUY: Yeah, rolldown would delete the whole module if the import is not used at all since `lodash-es` is side effects free.

IWANABETHATGUY: > Yes. It's not used🤡. Rolldown supports treeshaking on `_.x`. I guess it would be fine.

But for your case, it will keep the whole module, since you use default import rather than `import * as _ from 'lodash-es'`, I advice removing the statement or using `import {add } from 'lodash-es'` or `import * as _ from 'lodash-es'`

IWANABETHATGUY: I tested with `webpack` and `rolldown`

PR #1694, opened 2024-07-23
feat: support `output.globals` for iife format

### LINKED ISSUE #1569: iife format

### LINKED ISSUE #1623: Gracefully handle external modules in iife format

7086cmd:
In this PR, I addressed issues related to `output.exports`, particularly for the `IIFE` format, and extended the solution to handle external modules.

resolves #1623.
resolves #1569.

### Modifications

1. **Support for `output.globals` Key**: 
   - Implemented support for the `{ [id: string]: string }` type in the `globals` key.
   - Note: Although Rollup supports both `{ [id: string]: string }` and `((id: string) => string)`, this PR only includes support for the former. Future PRs may add support for the function type.

2. **Arguments Handling for `IIFE` Format**: 
   - Adapted the handler function from `CJS`.
   - Utilized `filter_map` to identify and manage used external modules.
   - Passed these modules to the `render_iife_arguments` function, leveraging the `globals` key.
   - Open to suggestions for more elegant solutions.

### Tests

- Tests for `iife/external_modules` passed. (The actual behavior aligns with rollup)
- Added new tests for `iife/external_modules_with_globals` and JavaScript side tests.


7086cmd:
Some tests are not passing yet, so I will modify the PR's status to `open` until tests passed.

crates/rolldown/tests/fixtures/format/iife/external_modules_with_globals/artifacts.snap
```diff
@@ -0,0 +1,20 @@
...
+input_file: crates/rolldown/tests/fixtures/format/iife/external_modules_with_globals
+---
+# Assets
+
+## main.mjs
+
+```js
+(function(node_path) {
+
+const { default: nodePath } = node:path;
```
underfin: Here has a invalid idenifer 'node:path'

7086cmd: I don't know, but Rollup does so. External modules needs the arguments that call the function, which may defined in `output.globals`.

Example code after bundling with Rollup:

```js
(function (node_path) {
	'use strict';



	return node_path.join;

})(node_path);
```

7086cmd: The source code is to export the `join` from `node:path`. When I specify the `globals: { 'node:path': 'path' }`, it will generate code in following:

```js
(function (node_path) {
	'use strict';



	return node_path.join;

})(path);
```

crates/rolldown/src/ecmascript/format/iife.rs
```diff
@@ -65,14 +76,75 @@ pub fn render_iife(
...
+      output_args.push(normalize_name(global));
+    } else {
+      // TODO add warning for missing global
+      output_args.push(normalize_name(external));
+    }
+  });
+  (input_args.join(", "), output_args.join(", "))
+}
+
+fn normalize_name(global: &str) -> String {
```
underfin: The rolldown_utils has a `legitimize_identifier_name` could be used.

crates/rolldown/tests/fixtures/format/iife/external_modules_with_globals/artifacts.snap
7086cmd: Sorry for the misunderstanding. I will fix it soon.

packages/rolldown/tests/fixtures/output/format/iife/_config.ts
```diff
@@ -12,6 +12,7 @@ export default defineTest({
     expect(output.output[0].code).toMatchInlineSnapshot(`
       "var myModule = (function() {
 
+
```
underfin, line 15: Here could be better add a `globals` options to test binding workflow.

7086cmd: I added tests for not only `globals` but also `exports` in iife.

crates/rolldown/src/ecmascript/format/iife.rs
7086cmd: I have used it in these areas and removed `normalize...`.

crates/rolldown/tests/fixtures/format/iife/external_modules_with_globals/artifacts.snap
7086cmd: Fixed.

crates/rolldown/src/ecmascript/format/iife.rs
```diff
@@ -65,14 +78,79 @@ pub fn render_iife(
...
+            s.push_str(&format!(
+              "const {{ {} }} = {};\n",
+              specifiers.join(", "),
+              legitimize_identifier_name(require_path_str.as_str())
+            ));
+            Some(require_path_str)
+          }
+        }
+        RenderImportDeclarationSpecifier::ImportStarSpecifier(alias) => {
+          s.push_str(&format!("const {alias} = {require_path_str};\n"));
```
underfin: Here also need to using  legitimize_identifier_name

7086cmd: Solved. I performed `legitimize_identifier_name` in assigning to `require_path_str`.

```rust
      let require_path_str = legitimize_identifier_name(&stmt.path).to_string();
```

7086cmd: Solved. Sorry for leaving so many cock-up.

crates/rolldown_binding/src/utils/normalize_binding_options.rs
```diff
@@ -96,6 +96,15 @@ pub fn normalize_binding_options(
     module_types = Some(tmp);
   }
 
+  let mut globals = None;
+  if let Some(raw) = output_options.globals {
+    let mut tmp = HashMap::with_capacity(raw.len());
+    for (k, v) in raw {
```
underfin: If change `BundlerOptions#globals` to `: Option<FxHashMap<String, String>>`, Here could be using `  let globals = output_options.globals.map(|raw| FxHashMap::from_iter(raw.into_iter()));
`.

crates/rolldown/src/ecmascript/format/iife.rs
```diff
@@ -65,14 +78,79 @@ pub fn render_iife(
   }
 
   // iife wrapper end
-  concat_source.add_source(Box::new(RawSource::new(format!(
-    "}})({});",
-    if has_exports && named_exports { "{}" } else { "" }
-  ))));
+  if invoke {
+    concat_source.add_source(Box::new(RawSource::new(format!("}})({output_args});"))));
```
underfin, line 83: Why here add `invoke` condiction, look like it's unnecessary, the iif wrapper always need to invoke.

7086cmd: In `iife`, it is unnessary, but in `amd`, `umd`, it won't be invoked with parameters. I think this function can be reused in generating `amd`, or `umd` function.
I think merging the function to `wrap.rs` is better, but it may be best to refactor it in a separate PR, imo.

crates/rolldown_binding/src/utils/normalize_binding_options.rs
7086cmd: I don't know why it reports error both in this method or `FxHashMap::from(...)`. 

```
   --> crates/rolldown_binding/src/utils/normalize_binding_options.rs:142:5
    |
142 |     globals,
    |     ^^^^^^^ expected `Option<HashMap<String, String>>`, found `Option<HashMap<String, String, ...>>`
    |
    = note: expected enum `std::option::Option<std::collections::HashMap<_, _, RandomState>>`
               found enum `std::option::Option<std::collections::HashMap<_, _, FxBuildHasher>>`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `rolldown_binding` (lib) due to 1 previous error
error: Recipe `check-rust` failed on line 50 with exit code 101
error: Recipe `check` failed on line 56 with exit code 101
(
```

underfin: Look like not change `BundlerOptions#globals` to `Option<FxHashMap<String, String>>`

crates/rolldown/src/ecmascript/format/iife.rs
underfin: Ok. I see. Maybe you could add some comment at now.

crates/rolldown_binding/src/utils/normalize_binding_options.rs
7086cmd: Solved

crates/rolldown/src/ecmascript/format/iife.rs
7086cmd: Comment added now.

underfin:
The life  will add an extra variable to the chunk, it is need to deconflict, i will fix it at next.

underfin (approved):

7086cmd:
> The life will add an extra variable to the chunk, it is need to deconflict, i will fix it at next.

Yes, especially when the module export value as `exports`, I previously forgot to mention it.

By the way, when do we plan to support `amd` and `umd`? I'd like to explore it since I am familiar with the `IIFE` now :)

underfin:
We are support iife becuase we need it to support vue bundling, the amd/umd is not in our roadmap at now. I think we could improve iife and cjs some behavior align with rollup, Here has more behavior we not supported at now, see ` getExportBlock, getNamespaceMarkers ` at https://github.com/rollup/rollup/blob/master/src/finalisers/iife.ts#L123. If you are interested for it, maybe you could do it. Thank you for contributing.

PR #1696, opened 2024-07-23
refactor:  compose js plugin

IWANABETHATGUY:
### Description
1. Simplify related code


packages/rolldown/src/utils/compose-js-plugins.ts
```diff
@@ -133,9 +133,7 @@ function createComposedPlugin(plugins: Plugin[]): Plugin {
               for (const handler of batchedHandlers) {
                 const [handlerFn, _handlerOptions] = normalizeHook(handler)
                 const result = await handlerFn.call(this, id)
-                if (isNullish(result)) {
-                  continue
-                } else {
+                if (!isNullish(result)) {
```
IWANABETHATGUY, line 136: 1. `continue` is the default behavior, so only one branch is enough.

hyf0 (approved):

PR #1698, opened 2024-07-23
perf: deprecate `mime_guess` in favor of a lighter guessing function.

7086cmd:
I wrapped the `mime`, and `infer` crate to [`mime_more`](https://github.com/7086cmd/mime_more) (benchmark result also in README). It also provides lighter mime guessing functions, with the benchmark from using `mime_guess` in about 289 ns to 37 ns and less rare extensions.

Also, the behavior of the Data URL is aligned with esbuild (except the percent decoding).


hyf0:
I will suggest to inline this package if code is proven to be better. However, in this case, using outside crate is not friendly for future maintaining, if we want to change something.

7086cmd:
Ok. How about using the crate's `extension_light` which can reduce not only the program size but also the bundling time, and ignore other features?

Other implementations are the same as the original one in this repo.

hyf0:
> Ok. How about using the crate's `extension_light` which can reduce not only the program size but also the bundling time, and ignore other features?
> 
> Other implementations are the same as the original one in this repo.

If it's better I will suggest just copy those code to rolldown's repo. Using outside crate is not friendly for future maintaining, if we want to change something.

7086cmd:
OK, I see.

crates/rolldown_utils/src/light_guess.rs
```diff
@@ -0,0 +1,129 @@
+// Port from https://github.com/7086cmd/mime_more/blob/main/src/light_guess.rs
+use mime::Mime;
+use phf::{phf_map, Map};
+use std::str::FromStr;
+
+pub static MIME_TYPES: Map<&'static str, &'static str> = phf_map! {
```
hyf0, line 6: `mime_guess` used `phf` too. Guess we don't need to maintain this? https://github.com/abonander/mime_guess/blob/master/src/impl_phf.rs

7086cmd: Yes, I think. phf provides a better performance (O(1)), but about 60 items is ambiguous to tell the performance of phf and match apart. But when we need other MIME types, using phf can be better.

7086cmd: phf is not authored by me :) I just learned mime_guess.

crates/rolldown/tests/esbuild/loader/auto_detect_mime_type_from_extension/artifacts.snap
```diff
@@ -15,7 +15,7 @@ var test_ns, test_default;
 var init_test = __esmMin(() => {
 	test_ns = {};
 	__export(test_ns, {default: () => test_default});
-	test_default = "data:image/svg+xml;base64,YVx4MDBiXHg4MGNceEZGZA";
```
IWANABETHATGUY: FYI, 
```js
TestAutoDetectMimeTypeFromExtension
---------- /out.js ----------
// test.svg
var require_test = __commonJS({
  "test.svg"(exports, module) {
    module.exports = "data:image/svg+xml;base64,YQBigGP/ZA==";
  }
});
```
esbuild generated code above.

crates/rolldown/tests/esbuild/loader/require_custom_extension_data_url/artifacts.snap
```diff
@@ -15,7 +15,7 @@ var test_ns, test_default;
 var init_test = __esmMin(() => {
 	test_ns = {};
 	__export(test_ns, {default: () => test_default});
-	test_default = "data:text/plain;charset=utf-8,a%5Cx00b%5Cx80c%5CxFFd";
```
IWANABETHATGUY:
```js
TestRequireCustomExtensionDataURL
---------- /out.js ----------
// test.custom
var require_test = __commonJS({
  "test.custom"(exports, module) {
    module.exports = "data:application/octet-stream;base64,YQBigGP/ZA==";
  }
});

// entry.js
console.log(require_test());
```

crates/rolldown_utils/src/light_guess.rs
hyf0: I mean `mime_guess` is already used `phf`. we don't need to maintain our own version of `MIME_TYPES `.

https://github.com/abonander/mime_guess/blob/master/src/impl_phf.rs

IWANABETHATGUY:
All snapshot you could found under https://github.com/evanw/esbuild/tree/main/internal/bundler_tests/snapshots, just convert our **snake_case** into **PascalCase** and prefix with `Test`

7086cmd:
Okay, thanks. I will fix them when I'm free.

crates/rolldown_utils/src/light_guess.rs
```diff
@@ -0,0 +1,129 @@
...
+/// - https://github.com/rolldown/rolldown/pull/1406/files#diff-4b612e077c82ae0e05e50eb0d419e02c05a04b83c6ac5440c0d0c9d0c38af942
+/// - https://github.com/evanw/esbuild/blob/fc37c2fa9de2ad77476a6d4a8f1516196b90187e/internal/helpers/mime.go#L5
+/// Thanks to @ikkz and @evanw for the inspiration.
+pub fn mime_type_by_extension(ext: &str) -> Option<&'static str> {
+  if let Some(mime) = MIME_TYPES.get(ext) {
+    Some(*mime)
+  } else {
+    None
+  }
+}
```
IWANABETHATGUY:
```suggestion
pub fn mime_type_by_extension(ext: &str) -> Option<&'static str> {
  MIME_TYPES.get(ext).map(|mime| *mime)
}

```

crates/rolldown_utils/src/light_guess.rs
```diff
@@ -0,0 +1,129 @@
...
+
+pub fn try_from_ext(ext: &str) -> anyhow::Result<Mime> {
+  if let Some(mime) = mime_type_by_extension(ext) {
+    Ok(Mime::from_str(mime)?)
+  } else {
+    anyhow::bail!("No mime type found for extension: {}", ext);
+  }
+}
```
IWANABETHATGUY:
```suggestion
pub fn try_from_ext(ext: &str) -> anyhow::Result<Mime> {
  mime_type_by_extension(ext)
    .ok_or_else(|| anyhow::Error::msg(format!("No mime type found for extension: {}", ext)))
    .and_then(|mime| Ok(Mime::from_str(mime)?))
}

```

crates/rolldown_utils/src/mime.rs
```diff
@@ -1,12 +1,20 @@
...
 fn is_texture(data: &[u8]) -> bool {
   std::str::from_utf8(data).is_ok()
 }
 
+pub fn is_texture_mime(mime_type: &Mime) -> bool {
+  mime_type.type_() == mime::TEXT
+    || mime_type.subtype() == mime::JSON
+    || mime_type.subtype() == mime::SVG
+    || mime_type.subtype() == mime::XML
+}
```
IWANABETHATGUY:
```suggestion
pub fn is_texture_mime(mime_type: &Mime) -> bool {
  mime_type.type_() == mime::TEXT
    || matches!(mime_type.subtype(), mime::JSON | mime::SVG || mime::XML)
}

```

7086cmd:
It should be a problem in our test case, and sadly I don't have a way to handle it (indistinguishable file is not supported in GitHub, see it in #1668). In [esbuild](https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_loader_test.go#L195), the `\x` should be the transferred expression, for example, when writing

```go
hexString := "\x48\x65\x6C\x6C\x6F"
```

The `hexString` is just `hello` in output.

IWANABETHATGUY:
> It should be a problem in our test case, and sadly I don't have a way to handle it (indistinguishable file is not supported in GitHub, see it in #1668). In [esbuild](https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_loader_test.go#L195), the `\x` should be the transferred expression, for example, when writing
> 
> ```go
> hexString := "\x48\x65\x6C\x6C\x6F"
> ```
> 
> The `hexString` is just `hello` in output.

Looks like our output is already wrong before https://github.com/rolldown/rolldown/pull/1668/files#diff-d0d7250843734347c76a1838a8e0b2d2f19752da56098bfea90a4e9fc7ac67c0L18, 
**esbuild**
```js
---------- /out.js ----------
// test.custom
var require_test = __commonJS({
  "test.custom"(exports, module) {
    module.exports = "data:application/octet-stream;base64,YQBigGP/ZA==";
  }
});

// entry.js
console.log(require_test());
```

7086cmd:
> > It should be a problem in our test case, and sadly I don't have a way to handle it (indistinguishable file is not supported in GitHub, see it in #1668). In [esbuild](https://github.com/evanw/esbuild/blob/main/internal/bundler_tests/bundler_loader_test.go#L195), the `\x` should be the transferred expression, for example, when writing
> > ```go
> > hexString := "\x48\x65\x6C\x6C\x6F"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The `hexString` is just `hello` in output.
> 
> Looks like our output is already wrong before https://github.com/rolldown/rolldown/pull/1668/files#diff-d0d7250843734347c76a1838a8e0b2d2f19752da56098bfea90a4e9fc7ac67c0L18, **esbuild**
> 
> ```js
> ---------- /out.js ----------
> // test.custom
> var require_test = __commonJS({
>   "test.custom"(exports, module) {
>     module.exports = "data:application/octet-stream;base64,YQBigGP/ZA==";
>   }
> });
> 
> // entry.js
> console.log(require_test());
> ```

Even if before #1668, the behavior with the test may still be "incorrect" because we may regard every `\` as a char instead of the synthesis char with the `x00`.

IWANABETHATGUY (approved):

Issue #1706, opened 2024-07-24
banner should be placed before the generated `use strict` for cjs


hyf0:
[...]

PR #1705, opened 2024-07-24
feat: add use strict for iife

underfin:

PR #1707, opened 2024-07-24
fix: move `banner` ahead of the file in `cjs` format.

7086cmd:
related: #1706, [#1705](https://github.com/rolldown/rolldown/pull/1705#discussion_r1689236056).

Tests are also added.


7086cmd:
`Banner` is not supported to be detected inline still. Waiting for another one.

crates/rolldown/src/ecmascript/format/iife.rs
```diff
@@ -56,7 +57,8 @@ pub fn render_iife(
     input_args
   ))));
 
-  if determine_use_strict(ctx) {
+  // Rollup also enables `use strict` in empty IIFE contents.
+  if determine_use_strict(ctx) || matches!(ctx.options.exports, OutputExports::None) {
```
underfin: The rollup has a `output.strict` to control it, it's value default is true, because rollup only support esm. The rolldown is not need it, because rolldown support esm and cjs, it should be determine. 
We could add the option to do it if the user use it at future, i think it look like is unnecessary at now.

7086cmd: Ok, I'll fix it now.

underfin (approved):

PR #1714, opened 2024-07-24
fix: unused commonjs export

IWANABETHATGUY:


crates/rolldown/src/stages/link_stage/mod.rs
```diff
@@ -316,7 +316,7 @@ impl<'a> LinkStage<'a> {
                     match importee_linking_info.wrap_kind {
                       WrapKind::None => {}
                       WrapKind::Cjs => {
-                        stmt_info.side_effect = true;
```
IWANABETHATGUY, line 336: Before,   whatever the importee has side effects or not. we would include the stmt.

IWANABETHATGUY: So the module would be included even if the module is unused and side effects free.

crates/rolldown/src/stages/link_stage/tree_shaking.rs
```diff
@@ -116,8 +116,15 @@ fn include_symbol(ctx: &mut Context, symbol_ref: SymbolRef) {
   let canonical_ref_symbol = ctx.symbols.get(canonical_ref);
   let mut canonical_ref_owner = ctx.modules[canonical_ref.owner].as_ecma().unwrap();
   if let Some(namespace_alias) = &canonical_ref_symbol.namespace_alias {
```
IWANABETHATGUY, line 115: this means the symbol is `import default from 'a.cjs'`

crates/rolldown/src/stages/link_stage/tree_shaking.rs
```diff
@@ -116,8 +116,15 @@ fn include_symbol(ctx: &mut Context, symbol_ref: SymbolRef) {
...
+    if let Some(named_import) = canonical_ref_owner.named_imports.get(&canonical_ref) {
+      let import = &canonical_ref_owner.import_records[named_import.record_id];
+      include_statement(ctx, canonical_ref_owner, import.stmt_idx + 1);
```
IWANABETHATGUY: `0` index is preserved, so `+ 1` to get the real `stmt_info`

IWANABETHATGUY: include the statement, https://github.com/rolldown/rolldown/pull/1714/files#diff-035ce9006e573c9e7d8ea10510bcf1b73f4fa35cd515dd88df96c1d36cd34581L345-L346, we need to explicitly include now, because the side effects is not always `true` now.

crates/rolldown/src/stages/link_stage/tree_shaking.rs
```diff
@@ -116,8 +116,15 @@ fn include_symbol(ctx: &mut Context, symbol_ref: SymbolRef) {
...
+    };
     canonical_ref = namespace_alias.namespace_ref;
     canonical_ref_owner = ctx.modules[canonical_ref.owner].as_ecma().unwrap();
+  } else if let Some(named_import) = canonical_ref_owner.named_imports.get(&symbol_ref) {
```
IWANABETHATGUY: for case `import * as xxx from 'test.cjs'`

crates/rolldown/tests/esbuild/dce/import_re_export_of_namespace_import/artifacts.snap
```diff
@@ -16,16 +16,9 @@ var require_foo = __commonJSMin((exports, module) => {
...
-//#endregion
-//#region node_modules/pkg/bar.js
-var require_bar = __commonJSMin((exports, module) => {
-	module.exports = "abc";
-});
-
 //#endregion
 //#region node_modules/pkg/index.js
 var import_foo = __toESM(require_foo());
-var import_bar = __toESM(require_bar());
```
IWANABETHATGUY, line 22:
```js
TestImportReExportOfNamespaceImport
---------- /out.js ----------
// Users/user/project/node_modules/pkg/foo.js
var require_foo = __commonJS({
  "Users/user/project/node_modules/pkg/foo.js"(exports, module) {
    module.exports = 123;
  }
});

// Users/user/project/node_modules/pkg/index.js

// Users/user/project/entry.js
console.log(import_foo.default);
```

crates/rolldown/tests/esbuild/dce/package_json_side_effects_false_keep_bare_import_and_require_common_js/artifacts.snap
```diff
@@ -18,7 +18,6 @@ var require_demo_pkg_index = __commonJSMin((exports, module) => {
 
 //#endregion
 //#region src/entry.js
-var import_demo_pkg_index = __toESM(require_demo_pkg_index());
 require_demo_pkg_index();
 console.log("unused import");
```
IWANABETHATGUY, line 23:
```js
TestPackageJsonSideEffectsFalseKeepBareImportAndRequireCommonJS
---------- /out.js ----------
// Users/user/project/node_modules/demo-pkg/index.js
var require_demo_pkg = __commonJS({
  "Users/user/project/node_modules/demo-pkg/index.js"(exports) {
    exports.foo = 123;
    console.log("hello");
  }
});

// Users/user/project/src/entry.js
require_demo_pkg();
console.log("unused import");
```

crates/rolldown/tests/fixtures/cjs_compat/mix-cjs-esm/artifacts.snap
```diff
@@ -25,7 +25,6 @@ require_foo();
 //#endregion
 //#region esm-import-cjs-export.js
 var require_esm_import_cjs_export = __commonJSMin((exports, module) => {
-	var import_foo = __toESM(require_foo());
```
IWANABETHATGUY, line 28: https://esbuild.github.io/try/#YgAwLjIwLjIALS1idW5kbGUgLS1taW5pZnk9ZmFsc2UgLS1mb3JtYXQ9ZXNtAGUAZW50cnkuanMAaW1wb3J0ICcuL2ZvbycKbW9kdWxlLmV4cG9ydHMgPSAxCgAAZm9vLmpzAA

crates/rolldown/tests/fixtures/cjs_compat/reexport_commonjs/artifacts.snap
```diff
@@ -39,6 +39,6 @@ assert.equal(foo_ns.foo, undefined);
 assert.equal((Object.keys(foo_ns)).length, 2);
```
IWANABETHATGUY: This is unexpected, I would have a look, but I think it doesn't block the merging. since this is a hotfix

IWANABETHATGUY: tracked https://github.com/rolldown/rolldown/issues/1715

crates/rolldown/src/stages/link_stage/tree_shaking.rs
```diff
@@ -178,11 +185,19 @@ fn include_member_expr_ref(ctx: &mut Context, symbol_ref: SymbolRef, props: &[Co
 
   let (export_name, namespace_property_name) =
```
IWANABETHATGUY: same with member expr with chain

crates/rolldown/src/stages/link_stage/mod.rs
```diff
@@ -354,11 +357,17 @@ impl<'a> LinkStage<'a> {
                           stmt_info
                             .referenced_symbols
                             .push(importee_linking_info.wrapper_ref.unwrap().into());
-                          // dbg!(&importee_linking_info.wrapper_ref);
                           stmt_info
                             .referenced_symbols
                             .push(self.runtime.resolve_symbol("__toESM").into());
                           stmt_info.declared_symbols.push(rec.namespace_ref);
+                          rec_namespace_ref_to_stmt_idx.lock().expect("should locked").insert(
```
IWANABETHATGUY, line 364: A reverse map(declared symbol -> stmt that declared it) , for now it only stores symbols imported from a cjs module, because for now, esm import stmt still is marked has side effects.

PR #1716, opened 2024-07-24
perf: ⚡️ use base64-simd to accelerate calc base64

IWANABETHATGUY:


hyf0:
Is there any downside about using `simd`?

hyf0 (approved):

PR #1704, opened 2024-07-24
refactor: use `render_chunk_imports` and `render_wrapper`.

7086cmd:
> According to the conversation with @underfin, we may evaluate it when we start to support AMD, UMD, etc. formats.

Thanks to @underfin.

### Summary

In previous pull requests, the render import chunks were copied from the CommonJS (CJS) module format to the Immediately Invoked Function Expression (IIFE) format. This led to duplicated code, which is hard to maintain. 

### Changes Made

1. **Unified Function Wrapping**: The functions previously separated between CJS and IIFE have now been unified into a single function to reduce code duplication and improve maintainability. 
2. **Separation of Concerns**: Different logical operations, such as the import star of a sphere, have been moved to separate functions. This enhances the readability and understandability of the code.
3. **Enhanced Support for Module Formats in the Future**: Introduced a new function named `render_wrapper` to better support not only IIFE but also UMD (Universal Module Definition) and AMD (Asynchronous Module Definition) in the future.
4. **`use strict` in IIFE**: Rollup marks `use strict` in IIFE, so I added a argument to `render_wrapper` to check if the format (IIFE) needs `use strict`.

Rollup's separate format handler is kept, and duplicated logics are merged.

These changes streamline the codebase, making it easier to extend support to additional module formats and maintain consistency across different formats.

PR #1717, opened 2024-07-24
feat(wip): support `umd` format.

7086cmd:
### Description

> Special thanks to @underfin for guiding the implementation of the `umd` support.

> This PR needs #1704 to be merged.

This PR introduces the integration of the `umd` format in Vite bundling. Notably, while `umd` supports `amd`, full support for `amd` has not been implemented yet and is not currently planned. An entry point for `amd` has been created, which will throw errors if attempted to use.

### Modifications

1. Added support for the `umd` format on both the node package and Rust side.
2. Implemented `render_umd` by reusing the `render_wrapper` function refactored in #1704. This function is instrumental in generating the wrapper function for SystemJS, AMD, UMD, and IIFE. Although support for SystemJS and AMD is not planned at this moment, it might be considered once stable.
3. Refactored `determine_export_mode` to ensure it is called only once.
4. **Note:** This PR is still a work in progress as tests have not yet been added. However, merging it now is not critical since `umd` format support is not a necessary part of this feature at the moment.


hyf0:
`umd` is also not planned. Personally, I don't against merging this, but we don't have enough man power to maintain it. Even if it's merged, if it blocks some important refactor, related code and tests will be removed and ignored.

However, doing such thing is really a maintaining burden. I don't like to tell we aren't going to merge this. So please open a issue before rushing into a PR. We are also in the discord server, instant online chat is also welcome.

7086cmd:
OK, I see. There is no hurry to merge it now, I will keep marking it draft and modifying its content until we decide to support `umd` format. Shall I open the issue right now?

hyf0:
Not needed.

7086cmd:
There are so many things that've changed these days. I'll rewrite it when we need UMD.

PR #1724, opened 2024-07-25
fix:  should not url encode when plain text

IWANABETHATGUY:


crates/rolldown/tests/esbuild/loader/require_custom_extension_data_url/artifacts.snap
```diff
@@ -15,7 +15,7 @@ var test_ns, test_default;
 var init_test = __esmMin(() => {
 	test_ns = {};
 	__export(test_ns, {default: () => test_default});
-	test_default = "data:text/plain;charset=utf-8,a%5Cx00b%5Cx80c%5CxFFd";
+	test_default = "data:text/plain;charset=utf-8,a\\x00b\\x80c\\xFFd";
```
IWANABETHATGUY, line 18:
```js
var __commonJS = (cb, mod) => function __require() {
  return mod || (0, cb[__getOwnPropNames(cb)[0]])((mod = { exports: {} }).exports, mod), mod.exports;
};

// a/foo.custom
var require_foo = __commonJS({
  "a/foo.custom"(exports, module) {
    module.exports = "data:text/plain;charset=utf-8,a\\x00b\\x80c\\xFFd%0A";
  }
});

// a/entry.js
var foo = require_foo();
console.log(foo);
//# sourceMappingURL=index.js.map
```

hyf0 (approved):

7086cmd:
There may be some misunderstanding :) The `test.custom` file should be a binary file according to `esbuild` if we want to align our behavior to `esbuild`. Whether to perform the `url encode` or use the percent decode is still discussed with @ikkz still.

crates/rolldown/tests/esbuild/loader/require_custom_extension_data_url/artifacts.snap
7086cmd: The percent decoding may be required according to [here](https://github.com/rolldown/rolldown/pull/1406#issuecomment-2212384575), either using the `urlencoding` or using the `percent decode` function.

7086cmd: percent decode is required according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs#datahello2c20world21).

IWANABETHATGUY: for this example, it did not specify `text/plain` and charset, https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs#datahello2c20world21

IWANABETHATGUY:
> There may be some misunderstanding :) The `test.custom` file should be a binary file according to `esbuild` if we want to align our behavior to `esbuild`. Whether to perform the `url encode` or use the percent decode is still discussed with @ikkz still.

So do you mean esbuild behavior is wrong? If so, could you file a issue in esbuild repo? 
There are two concerns: 
1. What if we misunderstand the example in `MDN`? 
2. We don't want to make too much difference with `esbuild`, that is easy to distinguish how much progress we made to be 100%(or near 100%) compatible with `esbuild`.

 we could file an issue with in esbuild repo, and look how `evaw` think about the behavior in `text/plain` encoding, if it is indeed a bug, then we could fix it.

IWANABETHATGUY:
And this is the output of webpack, it did not URL encode the `text/plain` either, 
```js
/***/ "./a/foo.custom":
/***/ ((__unused_webpack_module, __webpack_exports__, __webpack_require__) => {

"use strict";
__webpack_require__.r(__webpack_exports__);
/* harmony export */ __webpack_require__.d(__webpack_exports__, {
/* harmony export */   "default": () => (__WEBPACK_DEFAULT_EXPORT__)
/* harmony export */ });
/* harmony default export */ const __WEBPACK_DEFAULT_EXPORT__ = ("data:text/plain;charset=utf-8,a\\x00b\\x80c\\xFFd\n");
// ..snip
```

7086cmd:
> If the data contains [characters defined in RFC 3986 as reserved characters](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2), or contains space characters, newline characters, or other non-printing characters, those characters must be [percent-encoded](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding).

In [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs#datatexthtml3cscript3ealert2827hi27293b3c2fscript3e), the `<script>` was be encoded into `%3cscript%3e`. Encoding `<` to `%3c` (encoding ASCII chars) is not compulsory, but it might be recommended. (in [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs#common_problems) it also displays a Data URL without `<`-percent-decoded).

In esbuild, it converts `你好` to `data:text/plain;charset=utf-8,\u4F60\u597D%0A`.

PR #1725, opened 2024-07-25
fix: base64 handling

IWANABETHATGUY:
### Description
1. Rewrite some file with binary format.
2. should use standard base64 encoding instead.


crates/rolldown/src/utils/parse_to_ecma_ast.rs
```diff
@@ -67,7 +67,7 @@ pub fn parse_to_ecma_ast(
         }
         // TODO: should we support non-utf8 text?
       } else {
-        let encoded = rolldown_utils::base64::to_url_safe_base64(&data);
+        let encoded = rolldown_utils::base64::to_standard_base64(&data);
```
IWANABETHATGUY, line 70: https://github.com/evanw/esbuild/blob/360d47230813e67d0312ad754cad2b6ee09b151b/internal/bundler/bundler.go#L320-L332

crates/rolldown/tests/esbuild/loader/auto_detect_mime_type_from_extension/artifacts.snap
```diff
@@ -15,7 +15,7 @@ var test_ns, test_default;
 var init_test = __esmMin(() => {
 	test_ns = {};
 	__export(test_ns, {default: () => test_default});
-	test_default = "data:image/svg+xml;base64,YVx4MDBiXHg4MGNceEZGZA";
```
IWANABETHATGUY, line 18:
```js
---------- /out.js ----------
// test.svg
var require_test = __commonJS({
  "test.svg"(exports, module) {
    module.exports = "data:image/svg+xml;base64,YQBigGP/ZA==";
  }
});

// entry.js
console.log(require_test());
```

crates/rolldown/tests/esbuild/loader/require_custom_extension_data_url/artifacts.snap
```diff
@@ -15,7 +15,7 @@ var test_ns, test_default;
 var init_test = __esmMin(() => {
 	test_ns = {};
 	__export(test_ns, {default: () => test_default});
-	test_default = "data:text/plain;charset=utf-8,a\\x00b\\x80c\\xFFd";
```
IWANABETHATGUY, line 18:
```js
---------- /out.js ----------
// test.custom
var require_test = __commonJS({
  "test.custom"(exports, module) {
    module.exports = "data:application/octet-stream;base64,YQBigGP/ZA==";
  }
});

// entry.js
console.log(require_test());
```

7086cmd: Now it is OK.

7086cmd (approved):
Ok. I think we may handle the non-ASCII chars next.

PR #1730, opened 2024-07-25
fix: should not panic if there are accessing on `star as`  member

hyf0:


justfile
```diff
@@ -2,7 +2,8 @@ set windows-shell := ["powershell"]
 set shell := ["bash", "-cu"]
 
 _default:
-    just --list -u
+    cargo r crates/rolldown/tests/fixtures/tree_shaking/export_star2/_config.json
```
hyf0: What's this @IWANABETHATGUY cc

PR #1746, opened 2024-07-26
fix: reserve `reserved-key` in json module type.

7086cmd:
Rolldown seems to panic when parsing the following JSON before this PR:

```json
{ "null": null }
```

I used functions in `oxc::syntax::keyword` to solve it.

Notice that `is_reserved_keyword_or_global_object` function is suffering wrong calling (related pr: oxc-project/oxc#4484), I'll separately call them in this PR.


hyf0 (approved):

hyf0:
Thanks!

crates/rolldown/tests/fixtures/function/module_types/json/object_with_reserved_key/_config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "config": {
+    "external": ["node:assert"]
```
IWANABETHATGUY, line 3: `node:assert` is external by default.

7086cmd: I removed them in #1886.

PR #1713, opened 2024-07-24
feat: support `output.footer` in plugins.

7086cmd:
resolves #1702.
Thanks to @komyg for your contribution! This PR involves some test errors. Check it out [here](https://github.com/rolldown/rolldown/pull/1702#issuecomment-2248059655).

PR #1755, opened 2024-07-27
refactor: use macros to reduce duplicated codes in injection fields.

7086cmd:
#### Introduction

This PR introduces several improvements and new features, focusing on code optimization and support for additional plugin functionalities.

#### Contents

1. **Code Optimization using Macros:**
    - Introduced macros to reduce duplicated code in `ecma_generator.rs`, specifically targeting the format renderers.
    - The `Plugin` trait is excluded from macro optimization due to an issue with async traits not being compatible with the macros.

2. **Plugin Enhancements:**
    - Added support for `intro` and `outro` in plugins, using #1713 as a reference example.

By utilizing macros, we've significantly reduced the amount of duplicated code, making the codebase cleaner and easier to maintain.

### DIFF
--- a/crates/rolldown/src/ecmascript/ecma_generator.rs
+++ b/crates/rolldown/src/ecmascript/ecma_generator.rs
@@ -8,13 +8,33 @@ use rolldown_common::{
   AssetMeta, EcmaAssetMeta, ModuleId, ModuleIdx, OutputFormat, PreliminaryAsset, RenderedModule,
 };
 use rolldown_error::DiagnosableResult;
-use rolldown_plugin::{HookBannerArgs, HookFooterArgs};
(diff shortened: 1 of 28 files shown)


hyf0:
Using macros is rejected. It's bad for maintaining.

Boshen:
We try to avoid macros as much as we can. A macro should be used for DSLs or super repetitive work.

The usage here can be replicated by simple functions (I think).

hyf0:
Yeah. Duplication isn't bad always. Introducing abstraction too early might be bad for future refactoring.

7086cmd:
I see. But the logic of these four is simply the same. I am wondering if there is a better way except for macros instead of copying them...?

PR #1763, opened 2024-07-28
feat: support `intro` and `outro` in plugins.

7086cmd:
#### Contents

1. **Code Optimization:**
   - Refined types like `HookBannerArgs` to `HookInjectionArgs`, etc.
   - **macros in #1755 are removed now**.

2. **Plugin Enhancements:**
    - Added support for `intro` and `outro` in plugins, using #1713 as a reference example.


crates/rolldown/tests/fixtures/function/outro/iife/artifacts.snap
```diff
@@ -16,7 +16,7 @@ input_file: crates/rolldown/tests/fixtures/misc/outro/iife
 var main_default = "banner";
 
 //#endregion
-// outro
 return main_default;
+// outro
```
underfin, line 20: Look like it is incorrect.

underfin (approved):

underfin:
Thank you.

crates/rolldown/tests/fixtures/function/outro/iife/artifacts.snap
underfin: The pervious is incorrect.

PR #1767, opened 2024-07-28
feat: support `output.indent` option and indent the wrapper content in `iife`.

7086cmd:
Fix https://github.com/rolldown/rolldown/blob/main/crates/rolldown/src/ecmascript/format/iife.rs#L73, and support [`output.indent`](https://rollupjs.org/configuration-options/#output-indent) for user to disable indents (if needed).


packages/rolldown/tests/fixtures/output/indent/no_indent/exports/_config.ts
hyf0, line 1: Not needed due to the rust tests already ensure the correctness,

7086cmd: Because when I was adding the Rust side, the `false` is an invalid key in `indent` in `_config.json`, but I don't know why, even if I specified the `type` as `[ 'string', 'boolean', 'null' ]`. So in this `no_indent` test, I am testing the transform from `false` to an empty string in the JavaScript side and just enabling the `String` type in the Rust side.

7086cmd: So I think this test may be necessary.

hyf0: >  Can also be false (no indent)
> https://rollupjs.org/configuration-options/#output-indent

Does `no indent` mean empty string?

7086cmd: [REPL](https://rollupjs.org/repl/?version=4.19.1&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGZ1bmN0aW9uJTIwYSgpJTIwJTdCJTVDbiUyMCUyMGNvbnNvbGUubG9nKDEyMzQ1KSU1Q24lN0QlMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSUyQyUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTdEJTJDJTdCJTIyY29kZSUyMiUzQSUyMiUyRiUyRiUyMG1hdGhzLmpzJTVDbiU1Q24lMkYlMkYlMjBUaGlzJTIwZnVuY3Rpb24lMjBpc24ndCUyMHVzZWQlMjBhbnl3aGVyZSUyQyUyMHNvJTVDbiUyRiUyRiUyMFJvbGx1cCUyMGV4Y2x1ZGVzJTIwaXQlMjBmcm9tJTIwdGhlJTIwYnVuZGxlLi4uJTVDbmV4cG9ydCUyMGNvbnN0JTIwc3F1YXJlJTIwJTNEJTIweCUyMCUzRCUzRSUyMHglMjAqJTIweCUzQiU1Q24lNUNuJTJGJTJGJTIwVGhpcyUyMGZ1bmN0aW9uJTIwZ2V0cyUyMGluY2x1ZGVkJTVDbiUyRiUyRiUyMHJld3JpdGUlMjB0aGlzJTIwYXMlMjAlNjBzcXVhcmUoeCklMjAqJTIweCU2MCU1Q24lMkYlMkYlMjBhbmQlMjBzZWUlMjB3aGF0JTIwaGFwcGVucyElNUNuZXhwb3J0JTIwY29uc3QlMjBjdWJlJTIwJTNEJTIweCUyMCUzRCUzRSUyMHglMjAqJTIweCUyMColMjB4JTNCJTVDbiU1Q24lMkYlMkYlMjBUaGlzJTIwJTVDJTIyc2lkZSUyMGVmZmVjdCU1QyUyMiUyMGNyZWF0ZXMlMjBhJTIwZ2xvYmFsJTVDbiUyRiUyRiUyMHZhcmlhYmxlJTIwYW5kJTIwd2lsbCUyMG5vdCUyMGJlJTIwcmVtb3ZlZC4lNUNud2luZG93LmVmZmVjdDElMjAlM0QlMjAnY3JlYXRlZCclM0IlNUNuJTVDbmNvbnN0JTIwaW5jbHVkZUVmZmVjdCUyMCUzRCUyMGZhbHNlJTNCJTVDbmlmJTIwKGluY2x1ZGVFZmZlY3QpJTIwJTdCJTVDbiU1Q3QlMkYlMkYlMjBPbiUyMHRoZSUyMG90aGVyJTIwaGFuZCUyQyUyMHRoaXMlMjBpcyUyMG5ldmVyJTVDbiU1Q3QlMkYlMkYlMjBleGVjdXRlZCUyMGFuZCUyMHRodXMlMjByZW1vdmVkLiU1Q24lNUN0d2luZG93LmVmZmVjdDElMjAlM0QlMjAnbm90JTIwY3JlYXRlZCclM0IlNUNuJTdEJTIyJTJDJTIyaXNFbnRyeSUyMiUzQWZhbHNlJTJDJTIybmFtZSUyMiUzQSUyMm1hdGhzLmpzJTIyJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMm91dHB1dCUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmlpZmUlMjIlMkMlMjJpbmRlbnQlMjIlM0FmYWxzZSUyQyUyMm5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiU3RCUyQyUyMnRyZWVzaGFrZSUyMiUzQXRydWUlN0QlN0Q=) here. I think we can regard `no indent` as an empty string. Since I used `Cow` to apply in `indent`, and I just used `Cow::Borrowed` when the string is empty, so I think there won't be a visible performance regression compared with no calling `indent` function.

The indent field is only for the wrapper function, for example, iife. Contents' indents in the code won't be removed.

IWANABETHATGUY:
Can you enable source map for one testcase?

7086cmd:
Seems that there are some problems in the source map. I'll mark this as a draft first.

crates/rolldown/tests/fixtures/function/format/iife/custom_indent/artifacts.snap
```diff
@@ -0,0 +1,38 @@
...
+## main.mjs
+
+```js
+(function(exports) {
+
+   "use strict";
+
+   //#region main.js
+function test() {
+	console.log("Hello, world!");
```
underfin, line 17: Here also need to implement Indet  for `SourcemapSource`, it also need to adjust token position, it's is complex at now. I think we could leave it at now before we have a good way.

7086cmd: Yes, it is required also. Sorry for forgetting it :(.

7086cmd: And actually `rollup` does not modify any of the inner indent.

PR #1773, opened 2024-07-28
fix: should not generate dynamic lookup shim if the namespace ref is not used

IWANABETHATGUY:
### Description

Closes https://github.com/rolldown/rolldown/pull/1777.


hyf0:
Could you also add test in https://github.com/rolldown/rolldown/pull/1777.

crates/rolldown/tests/fixtures/semantic/export_star_from_external_as_shared_entries/artifacts.snap
hyf0, line 1: You missing the `_test.mjs` file in #1777.

hyf0:
These `_reExport` code should only generated if the namespace_object_ref is used by some places.

crates/rolldown/tests/fixtures/semantic/export_star_from_external_as_shared_entries/artifacts.snap
hyf0: This test could be removed in this PR, since it will fixed in https://github.com/rolldown/rolldown/pull/1780.

PR #1777, opened 2024-07-28
Add test

hyf0:


IWANABETHATGUY:
I think it is not related to https://github.com/rolldown/rolldown/pull/1773?, since main branch also failed, #1773 just keep `export * from 'fs'` as is in entry module.

IWANABETHATGUY:
https://github.com/rolldown/rolldown/pull/1780, see result in main branch

IWANABETHATGUY:
Closed since it is already added in main branch.

PR #1780, opened 2024-07-29
fix: remove redundant reexport from external modules

IWANABETHATGUY:


IWANABETHATGUY:
I think it is not related to https://github.com/rolldown/rolldown/pull/1773?, since main branch also failed, #1773 just keep `export * from 'fs'` as is in entry module.

hyf0:
@IWANABETHATGUY It might be related. There are conditions that using `is_user_defined_entry` might fall into this problem too. Let's have a chat afternoon.

crates/rolldown/tests/fixtures/semantic/export_star_from_external_as_shared_entries_cjs/artifacts.snap
```diff
@@ -0,0 +1,36 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+expression: snapshot
+input_file: crates/rolldown/tests/fixtures/semantic/export_star_from_external_as_shared_entries_cjs
+---
+# Assets
+
+## entry.cjs
```
hyf0: format cjs behavior is still wrong.

crates/rolldown/tests/esbuild/import_star/re_export_star_external_es6/artifacts.snap
```diff
@@ -10,6 +10,8 @@ input_file: crates/rolldown/tests/esbuild/import_star/re_export_star_external_es
 ```js
 import "foo";
 
+export * from "foo"
+
 
 //#region entry.js
 var entry_ns = {};
```
IWANABETHATGUY: Was this more like a workaround or a correct fix? the `foo` is exported twice?

hyf0: Correct fix. But the current code of rendering imports is kind of over abstraction. Don't want to deal with it in this timing.

PR #1782, opened 2024-07-29
fix: generate valid file name for virtual module chunk

underfin:

### DIFF
--- a/crates/rolldown_utils/src/path_ext.rs
+++ b/crates/rolldown_utils/src/path_ext.rs
@@ -44,6 +44,11 @@ impl PathExt for std::path::Path {
       _ => file_name,
     };
 
+    // Avoid file name contained an unexpected NUL byte.
+    if let Some(stripped) = file_name.strip_prefix('\0') {
+      return stripped.to_owned().into();
+    }
+
     file_name
   }
 }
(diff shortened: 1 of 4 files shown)


crates/rolldown_utils/src/path_ext.rs
```diff
@@ -44,6 +44,11 @@ impl PathExt for std::path::Path {
       _ => file_name,
     };
 
+    // Avoid file name contained an unexpected NUL byte.
```
hyf0: Should we do it in the begainning? I saw people return `\0[abslolute path]` in `resolve_id` hook.

underfin: If normalized it at module loader resolve pharse, it maybe caused `debug_id` is incorrect and `RenderChunk#modules` is incorrect

hyf0: I mean should we put this logic ahead of this function. https://github.com/rolldown/rolldown/pull/1782#discussion_r1694878474

crates/rolldown_utils/src/path_ext.rs
```diff
@@ -59,4 +64,7 @@ fn test_representative_file_name() {
 
   let path = cwd.join("vue").join("mod.ts");
```
hyf0, line 65: Consider test case 

```rust
let path = format!("\0{}", cwd.join("vue").join("mod.ts").to_string().unwrap().to_string()).
```

underfin: Fixed it.

hyf0 (approved):

PR #1786, opened 2024-07-30
fix: dedup export all external in entry module

IWANABETHATGUY:


crates/rolldown/tests/fixtures/semantic/export_star_from_external_as_shared_entries_cjs/artifacts.snap
```diff
@@ -43,6 +43,7 @@ require("node:fs");
 //#region main.js
 var main_ns = {};
 __reExport(main_ns, require("node:fs"));
+__reExport(main_ns, require("node:fs"));
```
IWANABETHATGUY, line 46: Not related to this pr, this duplicate `__reExport` is not generate in module level.

hyf0 (approved):

PR #1808, opened 2024-07-31
chore: export factory for builtin plugin

IWANABETHATGUY:


packages/rolldown/tests/fixtures/builtin-plugin/glob-import/basic/_config.ts
```diff
@@ -1,11 +1,11 @@
-import { BuiltinGlobImportPlugin } from 'rolldown'
+import { builtinGlobImportPlugin } from 'rolldown'
 import { defineTest } from '@tests'
 import * as path from 'path'
 
 export default defineTest({
   config: {
     plugins: [
-      new BuiltinGlobImportPlugin({
+      builtinGlobImportPlugin({
```
hyf0: I don't think we need to emphasize the `builtin` prefix here. These plugins are __imported__ from `rolldown`, which already has the builtin semantic.

Could you remove it? Other changes are fine.

hyf0 (approved):

PR #1810, opened 2024-07-31
feat: add support for banner as str

komyg:
### Description

According to the [Rollup docs](https://rollupjs.org/configuration-options/#output-banner-output-footer), the `banner` can either be a function or a `string`. This PR adds support for having the `banner` as a `string`.

### DIFF
--- a/packages/rolldown/tests/fixtures/plugin/banner-with-str/_config.ts
+++ b/packages/rolldown/tests/fixtures/plugin/banner-with-str/_config.ts
@@ -1,0 +1,20 @@
+import { expect, vi } from 'vitest'
+import path from 'node:path'
+import { defineTest } from '@tests'
+
+const entry = path.join(__dirname, './main.js')
+
+export default defineTest({
+  config: {
+    input: entry,
+    plugins: [
+      {
+        name: 'test-plugin',
+        banner: '/* Lorem ipsum */',
+      },
+    ],
+  },
+  afterTest: (output) => {
+    expect(output.output[0].code).toContain('/* Lorem ipsum */')
+  },
+})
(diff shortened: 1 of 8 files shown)


packages/rolldown/tests/fixtures/plugin/banner-with-str/_config.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const entry = path.join(__dirname, './main.js')
+
+export default defineTest({
+  config: {
+    input: entry,
+    plugins: [
+      {
+        name: 'test-plugin',
+        banner: '/* Lorem ipsum */',
```
underfin: Could you add a test for this?

```
{

     banner: {
           handle: '// banner'
     }
}
```

underfin:
[Here](https://github.com/rolldown/rolldown/blob/main/packages/rolldown/src/plugin/index.ts#L216) need to be `export type AddonHook = string | AddonHookFunction`

packages/rolldown/tests/fixtures/plugin/banner-with-str/_config.ts
underfin: If you update  `/* Lorem ipsum */` to `/* banner */` will make output clear to see.

underfin (approved):

underfin:
Thank you.

underfin:
@hyf0 Look like `!merge` has some problem, please take a look.

IWANABETHATGUY:
> * The [Graphite merge queue](https://app.graphite.dev/merges?org=rolldown&repo=rolldown) couldn't merge this PR because **it failed for an unknown reason** (Stack merges are not currently supported for forked repositories. Please create a branch in the target repository in order to merge).

> 

The [Graphite merge queue](https://app.graphite.dev/merges?org=rolldown&repo=rolldown) couldn't merge this PR because it failed for an unknown reason (Stack merges are not currently supported for forked repositories. Please create a branch in the target repository in order to merge). 
**graphite** has been explained to you

PR #1812, opened 2024-07-31
feat: support `output.esModule` option

7086cmd:
This PR introduces support for namespace blocks in IIFE or CJS format outputs, treating them as ES Modules instead of relying solely on `output.exports` being `named`. @underfin (I'm not in a hurry to work on this as I will likely spend a lot of time on my competition project next month. :)

### Modifications

1. Removed redundant usage of `determine_export_mode`.
2. Added `render_namespace_block`, inspired by Rollup, but implemented based on behavior observed in the REPL.
3. Added corresponding tests to validate the changes.

### Behaviors

1. **Namespace block is disabled if `output.exports` isn't `named`.**
2. **Namespace block is disabled if `output.format` is `esm` or `app`.**  
   - `esm` should always be an ES module.
   - `app` would never be a module.
3. **Entry Point with Default Export:**
   - When `output.esModule` is set to `'always'` (or `true` for Rollup compatibility), or `'if-default-prop'` (similar to Rollup), it will render `Object.defineProperty(exports, '__esModule', { value: true })`.
   - Left an entry for `generatedCode.symbol` for future updates.

### Rationale for Not Using `boolean | 'if-default-prop'`

During the conversion from JavaScript to Rust, the program encountered a panic when the source value was a boolean. Overriding `true` or `false` with a string was considered inelegant, so this approach was not used.


hyf0:
[I'm refactoring cjs exports generating](https://github.com/rolldown/rolldown/pull/1805) logic to ensure the correct semantic of live binding. Related changes are going to be on hold.

7086cmd:
> [I'm refactoring cjs exports generating](https://github.com/rolldown/rolldown/pull/1805) logic to ensure the correct semantic of live binding. Related changes are going to be on hold.

Ok. It seems that this PR doesn't have an impact on the actual export behavior :) And it's no hurry to be merged.

packages/rolldown/tests/fixtures/output/esModule/always/_config.ts
hyf0, line 1: The naming of folder `esModule` should be `es-module` .

crates/rolldown/src/utils/chunk/determine_export_mode.rs
```diff
@@ -0,0 +1,44 @@
...
+pub fn determine_export_mode(
+  export_mode: &OutputExports,
+  module: &EcmaModule,
+  exports: &[(Rstr, SymbolRef)],
+) -> DiagnosableResult<OutputExports> {
+  match export_mode {
+    OutputExports::Named => Ok(OutputExports::Named),
+    OutputExports::Default => {
+      if exports.len() != 1 || exports[0].0.as_str() != "default" {
+        return Err(vec![BuildDiagnostic::invalid_export_option(
```
hyf0, line 15: New `BuildDiagnostic` should have a test in `tests/rolldown/errors or warnings`.

crates/rolldown_common/src/inner_bundler_options/types/es_module.rs
```diff
@@ -0,0 +1,28 @@
...
+#[cfg(feature = "deserialize_bundler_options")]
+use serde::Deserialize;
+
+#[derive(Debug, Default)]
+#[cfg_attr(
+  feature = "deserialize_bundler_options",
+  derive(Deserialize, JsonSchema),
+  serde(rename_all = "kebab-case", deny_unknown_fields)
+)]
+pub enum EsModuleType {
```
hyf0:
```suggestion
pub enum EsModuleFlag {
```

hyf0:
https://github.com/rolldown/rolldown/pull/1805 has been merged. I think it's ok for this PR move on. Here are some reviews:
- Naming of folders in `crates` should be snake case. kebab-case for `packages`.

crates/rolldown_common/src/inner_bundler_options/types/es_module.rs
hyf0, line 1: filename should be consistent with the type.

crates/rolldown_common/src/inner_bundler_options/types/es_module.rs
```diff
@@ -0,0 +1,28 @@
...
+  Always,
+  Never,
+  #[default]
+  IfDefaultProp,
```
hyf0: Would be nice to add some document here.

```rust
/// What does this affects
pub enum EsModuleType {
   /// What does this do. 
   Always,
   /// What does this do.
  Never,
   /// What does this do. And why it's default and special.
  #[default]
  IfDefaultProp,
}
```

You could reference the rollup's documents.

crates/rolldown/tests/rolldown/function/es_module/always/_config.json
hyf0, line 1: Could you add runtime tests for these behaviors?

You could reference the following tests:

- https://github.com/rolldown/rolldown/blob/main/crates/rolldown/tests/rolldown/topics/live_bindings/named_exports_in_common_chunks/_test.mjs
- https://github.com/rolldown/rolldown/blob/main/crates/rolldown/tests/rolldown/topics/live_bindings/named_exports_in_common_chunks_cjs/_test.cjs

crates/rolldown/src/utils/chunk/determine_export_mode.rs
7086cmd: I moved the `determine_export_mode` from the `render_chunk_exports` to a splitter folder, as it won't use this function in the `render_chunk_exports`.

hyf0 (approved):
LGTM. @underfin cc

packages/rolldown/tests/fixtures/output/es-module/common-chunks/_config.ts
```diff
@@ -0,0 +1,26 @@
...
+      format: 'cjs',
+      esModule: true,
+    },
+  },
+  afterTest: (output) => {
+    expect(
+      output.output
+        .filter(({ type }) => type === 'chunk')
+        .every((chunk) =>
+          (chunk as RolldownOutputChunk).code.includes(
```
underfin, line 20: The common chunks shouldn't has the `__esModule` flag, see [rollup example](https://rollupjs.org/repl/?version=4.20.0&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmltcG9ydCgnLiUyRmZpbGUnKS50aGVuKChpbXBvcnRlZCklMjAlM0QlM0UlMjAlN0IlNUNuJTIwJTIwYXNzZXJ0LnN0cmljdEVxdWFsKGltcG9ydGVkLnNoYXJlZCUyQyUyMCdzaGFyZWQnKSU1Q24lN0QpJTVDbiU1Q25leHBvcnQlMjBjb25zdCUyMG1haW4lMjAlM0QlMjAnbWFpbiclMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSUyQyUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTdEJTJDJTdCJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGNvbnN0JTIwc2hhcmVkJTIwJTNEJTIwJ3NoYXJlZCclNUNuJTIyJTJDJTIyaXNFbnRyeSUyMiUzQWZhbHNlJTJDJTIybmFtZSUyMiUzQSUyMmZpbGUuanMlMjIlN0QlNUQlMkMlMjJvcHRpb25zJTIyJTNBJTdCJTIyb3V0cHV0JTIyJTNBJTdCJTIyZXNNb2R1bGUlMjIlM0F0cnVlJTJDJTIyZXhwb3J0cyUyMiUzQSUyMm5hbWVkJTIyJTJDJTIyZm9ybWF0JTIyJTNBJTIyY2pzJTIyJTdEJTJDJTIydHJlZXNoYWtlJTIyJTNBdHJ1ZSU3RCU3RA==)

7086cmd: Got it.

crates/rolldown/src/utils/chunk/namespace_marker.rs
```diff
@@ -0,0 +1,41 @@
...
+  es_module_flag: &EsModuleFlag,
+  has_default_export: bool,
+  namespace_to_string_tag: bool,
+) -> String {
+  let es_module = determine_es_module(es_module_flag, has_default_export);
+  let result = render_marker(es_module, namespace_to_string_tag);
+  if result.is_empty() {
+    String::new()
+  } else {
+    format!("\n{result}")
```
underfin: Here `\n` is unnessary, the concat source will add it.

crates/rolldown/src/utils/chunk/namespace_marker.rs
```diff
@@ -0,0 +1,41 @@
...
+/// to have a namespace marker in the output.
+pub fn determine_es_module(es_module_flag: &EsModuleFlag, has_default_export: bool) -> bool {
+  match es_module_flag {
+    EsModuleFlag::Always => true,
+    EsModuleFlag::IfDefaultProp if has_default_export => true,
+    _ => false,
+  }
+}
+
+fn render_marker(es_module: bool, to_string_tag: bool) -> String {
```
underfin: Could you using Option<&str> as return typing? I think it is better at here. The `render_namespace_markers` also could using it.

packages/rolldown/tests/fixtures/output/es-module/common-chunks/_config.ts
7086cmd: It should be a bug in handling the dynamic import already appearing in the main branch, which infers the common chunk as an entry point and removed all of exports.

underfin (approved):

underfin:
cc @hyf0 Check assignees' approval Look like is not expected.

hyf0:
There aren't event like `approved`. It have to manually re-run or trigger by the assign operation. I'll see how to solve it.

PR #1816, opened 2024-07-31
chore(rust): move `SymbolOrMemberExprRef` and `MemberExprRef` to seperate files

hyf0:

PR #1814, opened 2024-07-31
chore(rust/scanner): prefer returning `SymbolRef` instead of `SymbolId`

hyf0:

PR #1813, opened 2024-07-31
chore(rust/scanner): improve naming

hyf0:

PR #1818, opened 2024-07-31
refactor(rust/treeshaking): get resolved symbol of `MemberExprRef` in one go

hyf0:
### Description

- This PR does does the same thing as the old code but storing the result in a map and use it directly rather do a resolving process with cache.
- This PR also split the member expression resolving logic into a extra step, it might slow down the performance but make the treeshaking process more clean and simple.
- https://github.com/rolldown/rolldown/pull/1805 rely on this PR, which makes the finalizing logic simpler.

### DIFF
--- a/crates/rolldown/src/ast_scanner/impl_visit.rs
+++ b/crates/rolldown/src/ast_scanner/impl_visit.rs
@@ -5,6 +5,7 @@ use oxc::{
     Visit,
   },
   codegen::{self, CodeGenerator, Gen},
+  span::GetSpan,
 };
 use rolldown_common::ImportKind;
 
@@ -58,14 +59,17 @@ impl<'me, 'ast> Visit<'ast> for AstScanner<'me> {
           }
         };
         match object_symbol_in_top_level {
-          // import statements are hoisted to the top of the module, so in this time being, all imports scanned.
-          Some(sym_ref) if self.result.named_imports.contains_key(&sym_ref) => {
+          // - Import statements are hoisted to the top of the module, so in this time being, all imports are scanned.
+          // - Having empty span will also results to bailout since we rely on span to identify ast nodes.
+          Some(sym_ref)
+            if self.result.named_imports.contains_key(&sym_ref) && !expr.span().is_unspanned() =>
+          {
             let props = props_in_reverse_order
               .into_iter()
               .rev()
               .map(|ident| ident.name.as_str().into())
               .collect::<Vec<_>>();
-            self.add_member_expr_reference(sym_ref, props);
+            self.add_member_expr_reference(sym_ref, props, expr.span());
             // Don't walk again, otherwise we will add the `object_symbol_in_top_level` again in `visit_identifier_reference`
             return;
           }
--- a/crates/rolldown_common/src/types/member_expr_ref.rs
+++ b/crates/rolldown_common/src/types/member_expr_ref.rs
@@ -1,4 +1,5 @@
-use oxc::span::CompactStr;
+use oxc::span::{CompactStr, Span};
+use rustc_hash::FxHashMap;
 
 use crate::SymbolRef;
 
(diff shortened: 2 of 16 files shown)


hyf0:
* **#1818** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1818?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1816** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1816?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1815** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1815?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1814** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1814?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1813** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1813?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/rolldown_common/src/types/member_expr_ref.rs
```diff
@@ -9,4 +10,32 @@ use crate::SymbolRef;
...
+  pub fn new(object_ref: SymbolRef, props: Vec<CompactStr>, span: Span) -> Self {
+    Self { object_ref, props, span }
+  }
+
+  // #[allow(clippy::manual_map)]: Current code is more readable.
+  #[allow(clippy::manual_map)]
+  pub fn resolved_symbol_ref(
+    &self,
+    resolved_map: &FxHashMap<Span, Option<(SymbolRef, Vec<CompactStr>)>>,
+  ) -> Option<SymbolRef> {
```
hyf0, line 27: return `None` need to be handled specially in code splitting and treeshaking. The old code fallback to `object_ref`, which is actually negetive.

crates/rolldown/src/ast_scanner/impl_visit.rs
```diff
@@ -58,14 +59,17 @@ impl<'me, 'ast> Visit<'ast> for AstScanner<'me> {
           }
         };
         match object_symbol_in_top_level {
-          // import statements are hoisted to the top of the module, so in this time being, all imports scanned.
-          Some(sym_ref) if self.result.named_imports.contains_key(&sym_ref) => {
+          // - Import statements are hoisted to the top of the module, so in this time being, all imports are scanned.
+          // - Having empty span will also results to bailout since we rely on span to identify ast nodes.
+          Some(sym_ref)
+            if self.result.named_imports.contains_key(&sym_ref) && !expr.span().is_unspanned() =>
```
IWANABETHATGUY, line 65: What if other plugin inserted a top-level `MemberExpr` in **transform** phase?

IWANABETHATGUY:
Overall, LGTM, except that using span as a key might be a footgun in the future.

crates/rolldown/src/ast_scanner/impl_visit.rs
hyf0: > Import statements are hoisted to the top of the module, so in this time being, all imports are scanned.

No matter what transform, this rule still applies.

> except that using span as a key might be a footgun in the future.

As for invalid span, we just do bailout. We'll switch `AstNoeId` once oxc stablize it.

IWANABETHATGUY: `AstNodeId` sounds more robust

IWANABETHATGUY: I saw a litlte issue when I try to inserted a ZERO SPAN member expr, e.g.
```js
// index.js
import * as test from './test.js'
console.log(test.a)

test.b() // `test.b()` inserted by transformer

// test.js
export const a = 1000;
export const b = 10000;

```
It would ref the whole namespace? is that expected? 
**output** 
```js

//#region rolldown:runtime
var __export = (target, all) => {
		get: all[name],
		enumerable: true
	});
};

//#endregion
//#region hyper-cube.js
var hyper_cube_ns = {};
__export(hyper_cube_ns, {
	a: () => a,
	b: () => b
});
const a = 100;
const b = 1000;
test.b();

//#endregion
//#region entry.js
a();
hyper_cube_ns.b();

//#endregion
//# sourceMappingURL=entry.js.map
```

IWANABETHATGUY: I saw a litlte issue when I try to inserted a ZERO SPAN member expr, e.g. 
```js
// MULTIPLE ENTRY MODULES
import * as test from './hyper-cube.js';

test.a()
test.b() // inserted by transformer.
// hyper-cube.js
export const a = 100;
export const b = 1000;

```
It would ref the whole namespace.
**output** 
```js

//#region rolldown:runtime
var __export = (target, all) => {
		get: all[name],
		enumerable: true
	});
};

//#endregion
//#region hyper-cube.js
var hyper_cube_ns = {};
__export(hyper_cube_ns, {
	a: () => a,
	b: () => b
});
const a = 100;
const b = 1000;

//#endregion
//#region entry.js
a();
hyper_cube_ns.b();

//#endregion
//# sourceMappingURL=entry.js.map
```


**expected** 
```js

//#region hyper-cube.js
const a = 100;
const b = 1000;

//#endregion
//#region entry.js
a();
b();

//#endregion
//# sourceMappingURL=entry.js.map
```

IWANABETHATGUY: the result is still right, just wait for `AstNodeId` stablized.

hyf0: Yeah. It's expected. We have code to prevent scanning such member expr by `!expr.span().is_unspanned()`.

PR #1823, opened 2024-07-31
feat: upgrade oxc

Boshen:
The next oxc version improves sourcemap performance, and fixes various codegen annoyances.


crates/rolldown/tests/esbuild/dce/dce_of_symbol_instances/artifacts.snap
```diff
@@ -1,6 +1,6 @@
 ---
-source: crates/rolldown_testing/src/case/case.rs
-expression: content
+source: crates/rolldown_testing/src/integration_test.rs
+expression: snapshot
```
Boshen: We can get rid of this line from the snapshot by a setting.

hyf0: Yeah. It's quite annoying. I'll take a look.

hyf0:
Run `just test-rust` local and the snapshot conflicts should be solved.

hyf0:
> Run `just test-rust` local and the snapshot conflicts should be solved.

My bad. You probably need to rebase the main branch to apply those newest settings. You could run `just test-rust` in solving conflicts and not need to solve manully.

crates/rolldown_binding/src/types/binding_output_chunk.rs
```diff
@@ -86,12 +86,7 @@ impl BindingOutputChunk {
 
   #[napi(getter)]
   pub fn map(&self) -> napi::Result<Option<String>> {
-    self
-      .inner
-      .map
-      .as_ref()
-      .map(|v| v.to_json_string().map_err(|e| napi::Error::from_reason(format!("{e:?}"))))
-      .transpose()
+    Ok(self.inner.map.as_ref().map(SourceMap::to_json_string))
```
Boshen, line 89: This no longer returns `Result`, we can remove `Result` in a follow up PR.

hyf0:
`just fmt` for fix format issues.

hyf0 (approved):
Nice

PR #1830, opened 2024-08-01
fix: remove unnecessary esm keyword span usage

IWANABETHATGUY:

### DIFF
--- a/crates/rolldown/src/ast_scanner/mod.rs
+++ b/crates/rolldown/src/ast_scanner/mod.rs
@@ -44,7 +44,7 @@ pub struct ScanResult {
 
 bitflags::bitflags! {
     /// Represents a set of flags.
-    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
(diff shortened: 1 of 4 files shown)


crates/rolldown/src/ast_scanner/mod.rs
```diff
@@ -52,8 +62,7 @@ pub struct AstScanner<'me> {
   symbols: &'me mut AstSymbols,
   current_stmt_info: StmtInfo,
   result: ScanResult,
-  esm_export_keyword: Option<Span>,
```
hyf0, line 55: We need these report to some errors/warnings in the future. See https://hyrious.me/esbuild-repl/?version=0.23.0&b=e%00entry.js%00export+const+a+%3D+1%0Amodule.exports+%3D+%7B%7D&b=%00foo.cjs%00module.exports.foo+%3D+1&o=--bundle+--format%3Desm+--external%3Aext+--outdir%3Ddist

IWANABETHATGUY: got it.

PR #1832, opened 2024-08-01
feat: commonjs-variable-in-esm diagnostic

### LINKED ISSUE #1831: [Feature Request]: Emit warnings for commonjs-variable-in-esm

https://hyrious.me/esbuild-repl/?version=0.23.0&b=e%00entry.js%00export+const+a+%3D+1%0Amodule.exports+%3D+%7B%7D&b=%00foo.cjs%00module.exports.foo+%3D+1&o=--bundle+--format%3Desm+--external%3Aext+--outdir%3Ddist

Related code:
- https://github.com/rolldown/rolldown/blob/726f6333ef6f765720bc7a827cb2b315f065c34a/crates/rolldown/src/ast_scanner/mod.rs#L42
- https://github.com/rolldown/rolldown/blob/726f6333ef6f765720bc7a827cb2b315f065c34a/crates/rolldown/src/ast_scanner/mod.rs#L549

IWANABETHATGUY:
### Description
Ref: https://github.com/evanw/esbuild/blob/main/CHANGELOG-2022.md#01411

Closed https://github.com/rolldown/rolldown/issues/1831

### DIFF
--- a/crates/rolldown/src/ast_scanner/mod.rs
+++ b/crates/rolldown/src/ast_scanner/mod.rs
@@ -123,22 +123,28 @@ impl<'me> AstScanner<'me> {
     if self.esm_export_keyword_start.is_some() {
       exports_kind = ExportsKind::Esm;
       if let Some(start) = self.cjs_module_ident_start {
-        self.result.warnings.push(BuildDiagnostic::commonjs_variable_in_esm(
-          self.file_path.to_string(),
-          self.source.clone(),
-          // SAFETY: we checked at the beginning
-          self.esm_export_keyword_start.expect("should have start offset"),
-          CjsExportStartOffset::Module(start),
-        ))
+        self.result.warnings.push(
+          BuildDiagnostic::commonjs_variable_in_esm(
+            self.file_path.to_string(),
+            self.source.clone(),
+            // SAFETY: we checked at the beginning
+            self.esm_export_keyword_start.expect("should have start offset"),
+            CjsExportStartOffset::Module(start),
+          )
+          .with_severity_warning(),
+        )
       }
       if let Some(start) = self.cjs_exports_ident_start {
-        self.result.warnings.push(BuildDiagnostic::commonjs_variable_in_esm(
-          self.file_path.to_string(),
-          self.source.clone(),
-          // SAFETY: we checked at the beginning
-          self.esm_export_keyword_start.expect("should have start offset"),
-          CjsExportStartOffset::Exports(start),
-        ))
+        self.result.warnings.push(
+          BuildDiagnostic::commonjs_variable_in_esm(
+            self.file_path.to_string(),
+            self.source.clone(),
+            // SAFETY: we checked at the beginning
+            self.esm_export_keyword_start.expect("should have start offset"),
+            CjsExportStartOffset::Exports(start),
+          )
+          .with_severity_warning(),
+        )
       }
     } else if self.cjs_exports_ident_start.is_some() || self.cjs_module_ident_start.is_some() {
       exports_kind = ExportsKind::CommonJs;
--- a/crates/rolldown/src/ast_scanner/mod.rs
+++ b/crates/rolldown/src/ast_scanner/mod.rs
@@ -19,7 +19,7 @@ use rolldown_common::{
   ModuleId, ModuleIdx, NamedImport, RawImportRecord, Specifier, StmtInfo, StmtInfos, SymbolRef,
 };
 use rolldown_ecmascript::{BindingIdentifierExt, BindingPatternExt};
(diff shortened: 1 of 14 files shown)


crates/rolldown/src/ast_scanner/mod.rs
```diff
@@ -52,12 +52,12 @@ pub struct AstScanner<'me> {
   symbols: &'me mut AstSymbols,
   current_stmt_info: StmtInfo,
   result: ScanResult,
-  esm_export_keyword: Option<Span>,
-  esm_import_keyword: Option<Span>,
+  esm_export_keyword_start: Option<u32>,
```
IWANABETHATGUY: Since all length of `keyword` and special `identifier` is fixed, just save their start offset is enough

crates/rolldown/src/ast_scanner/mod.rs
```diff
@@ -120,9 +120,33 @@ impl<'me> AstScanner<'me> {
     self.visit_program(program);
     let mut exports_kind = ExportsKind::None;
 
-    if self.esm_export_keyword.is_some() {
+    if self.esm_export_keyword_start.is_some() {
       exports_kind = ExportsKind::Esm;
-    } else if self.used_exports_ref || self.used_module_ref {
+      if let Some(start) = self.cjs_module_ident_start {
```
IWANABETHATGUY: Esbuild would emit for each `module` ident, and `exports` ident. here is an example.
https://hyrious.me/esbuild-repl/?version=0.23.0&b=e%00entry.js%00exports.test+%3D+1000%3B%0Aexports.a+%3D+1000%3B%0Aexports.b+%3D+1000%3B%0Amodule.exports+%3D+%7B%7D%0A%0Aexport+const+a+%3D+1%0A%0A%0A&b=%00foo.cjs%00module.exports.foo+%3D+1&o=--bundle+--format%3Desm+--external%3Aext+--outdir%3Ddist

crates/rolldown_error/src/events/commonjs_variable_in_esm.rs
```diff
@@ -0,0 +1,66 @@
...
+  Exports(u32),
+}
+impl CjsExportStartOffset {
+  pub fn start(&self) -> u32 {
+    match self {
+      CjsExportStartOffset::Module(start) | CjsExportStartOffset::Exports(start) => *start,
+    }
+  }
+
+  pub fn end(&self) -> u32 {
```
hyf0, line 21: Seems brings more complexity than it should just for saving a `u32` size.

crates/rolldown/src/ast_scanner/mod.rs
hyf0: Could you storing span to keep consistency align with others?

I'm not sure this actually save anything, since `CjsExportStartOffset` also takes u64 bytes.

IWANABETHATGUY: What do you mean by u64 bytes?

IWANABETHATGUY: 64 bytes?

IWANABETHATGUY: or size of `u64`?

IWANABETHATGUY: FYI, store span rather than `u32` with enum will allocates 4 more bytes, they don't cost the same memory.

hyf0: or size of u64

crates/rolldown_error/src/events/commonjs_variable_in_esm.rs
IWANABETHATGUY: fine, if you think this is complex.

hyf0: Yeah. I'm thinking if we accept such design, there's no reason about not applying this design principle for all possible places. So if we gonna refactor all our code with this design principle, the complexity it brings and bytes it saves seems not worth.

IWANABETHATGUY:
Fixed

crates/rolldown/src/ast_scanner/mod.rs
```diff
@@ -52,12 +52,12 @@ pub struct AstScanner<'me> {
   symbols: &'me mut AstSymbols,
   current_stmt_info: StmtInfo,
   result: ScanResult,
-  esm_export_keyword: Option<Span>,
-  esm_import_keyword: Option<Span>,
+  esm_export_keyword_start: Option<Span>,
```
hyf0:
```suggestion
  esm_export_keyword: Option<Span>,
```

hyf0 (approved):

PR #1833, opened 2024-08-01
fix: using sanitize_file_name to generate valid file name

underfin:
### Description

The rollup using `sanitizeFileName` to generate the valid file name, see [here](https://github.com/rollup/rollup/blob/master/src/utils/sanitizeFileName.ts). 

The related vite pr: https://github.com/vitejs/vite/pull/9737.

### DIFF
--- a/crates/rolldown_utils/src/sanitize_file_name.rs
+++ b/crates/rolldown_utils/src/sanitize_file_name.rs
@@ -1,0 +1,21 @@
+use std::borrow::Cow;
+
+// TODO support `output.sanitizeFileName`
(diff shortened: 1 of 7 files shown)


hyf0 (approved):

crates/rolldown_utils/src/sanitize_file_name.rs
```diff
@@ -0,0 +1,21 @@
+use std::borrow::Cow;
+
+// TODO support `output.sanitizeFileName`
+// Follow from https://github.com/rollup/rollup/blob/master/src/utils/sanitizeFileName.ts
```
IWANABETHATGUY, line 4: I quickly look the implementation, it seems this is not a fully equivalent port?

PR #1837, opened 2024-08-01
fix: resolve conflict `exports` key in `CJS` and `IIFE`.

### LINKED ISSUE #1623: Gracefully handle external modules in iife format

### LINKED ISSUE #1836: [Bug]: Duplicated `exports` when IIFE exports a value with `exports` and enables `named` mode.

https://github.com/7086cmd/rolldown-reproduction-conflict-named-iife-with-key.git

Clone it to local and build `just build native release` before running this. It is a bug in `named` exports, which may not be included in the release version/

### What is expected?

```js
(function(exports) {

"use strict";

//#region src/main.js
const exports$1 = 1;

//#endregion
Object.defineProperty(exports, '__esModule', { value: true });
Object.defineProperty(exports, 'exports', {
  enumerable: true,
  get: function () {
    return exports$1;
  }
});
return exports;
})({});
```
And Rollup's REPL [here](https://rollupjs.org/repl/?version=4.19.2&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGNvbnN0JTIwZXhwb3J0cyUyMCUzRCUyMDElM0IlMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSUyQyUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMm91dHB1dCUyMiUzQSU3QiUyMmV4cG9ydHMlMjIlM0ElMjJuYW1lZCUyMiUyQyUyMmZvcm1hdCUyMiUzQSUyMmlpZmUlMjIlMkMlMjJuYW1lJTIyJTNBJTIybXlCdW5kbGUlMjIlN0QlMkMlMjJ0cmVlc2hha2UlMjIlM0F0cnVlJTdEJTdE).

The `Object.defineProperty(...` is not related to this issue.

```js
(function(exports) {

"use strict";

//#region src/main.js
const exports = 1;

//#endregion
Object.defineProperty(exports, '__esModule', { value: true });
Object.defineProperty(exports, 'exports', {
  enumerable: true,
  get: function () {
    return exports;
  }
});
return exports;
})({});
```

It causes the conflict.

### System Info

```Shell
  System:
    OS: macOS 15.1
    CPU: (12) arm64 Apple M2 Pro
    Memory: 147.27 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.4.1 - /opt/homebrew/bin/node
    npm: 10.8.1 - /opt/homebrew/bin/npm
    pnpm: 9.6.0 - /opt/homebrew/bin/pnpm
    bun: 1.1.12 - ~/.bun/bin/bun
  Browsers:
    Chrome: 127.0.6533.88
    Chrome Canary: 129.0.6630.0
    Safari: 18.0
  npmPackages:
    rolldown: file:../../rolldown/packages/rolldown => 0.12.2
```

### Any additional comments?

It may related to the deconflict in [Rollup](https://github.com/rollup/rollup/blob/master/src/utils/deconflictChunk.ts#L25-L43). I may fix this when I'm free, but still open this issue for noticing this.

7086cmd:
resolves #1836.
resolves #1623.

I modified the `renamer` and checked if it wasn't `esm`. Since named exports are not allowed using `default` or `none`, we should always consider it as a reserved key.

> https://github.com/rolldown/rolldown/pull/1694#issuecomment-2245219042 need to do
> 
> From #1623.


crates/rolldown/src/utils/renamer.rs
```diff
@@ -35,12 +34,15 @@ pub struct Renamer<'name> {
 }
 
 impl<'name> Renamer<'name> {
-  pub fn new(symbols: &'name Symbols, _modules_len: usize) -> Self {
+  pub fn new(symbols: &'name Symbols, _modules_len: usize, format: &OutputFormat) -> Self {
+    let manual_reserved =
+      if matches!(format, OutputFormat::Esm) { vec![] } else { vec!["exports"] };
```
underfin: The cjs shouldn't add it.

7086cmd: [REPL](https://rollupjs.org/repl/?version=4.19.2&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGNvbnN0JTIwZXhwb3J0cyUyMCUzRCUyMDElM0IlMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSUyQyUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMm91dHB1dCUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmNqcyUyMiU3RCUyQyUyMnRyZWVzaGFrZSUyMiUzQXRydWUlN0QlN0Q=) here. It seems that we should also consider this situation 🤔 , but I will add some tests then.

underfin:
Please wait for me debugger rollup.

crates/rolldown/src/utils/renamer.rs
underfin: Here should be port from https://github.com/rollup/rollup/blob/master/src/Chunk.ts#L1377-L1394. cc @hyf0

7086cmd: I have modified it.

IWANABETHATGUY (approved):

hyf0 (approved):

PR #1838, opened 2024-08-01
feat: add query after each glob file name

IWANABETHATGUY:


packages/rolldown/tests/fixtures/builtin-plugin/glob-import/query-object/dir/b.js.snap
IWANABETHATGUY, line 1: I don't find a way to inspect the query in filename, other than snapshot.

IWANABETHATGUY: Need to inspect if the query was generated correctly.

hyf0 (approved):

PR #1840, opened 2024-08-02
feat: add module_preload_polyfill plugin

### LINKED ISSUE #1827: modulePreloadPolyfillPlugin
The modulePreloadPolyfillPlugin provides a virtual module for the preload polyfill, the source code see [here](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/plugins/modulePreloadPolyfill.ts).

The rolldown always has a example plugin to do similar things, you could follow [it](https://github.com/rolldown/rolldown/blob/main/crates/rolldown_plugin_wasm/src/lib.rs#L19-L37). You also need to add plugin binding and export it, add test it at [rolldown js side test](https://github.com/rolldown/rolldown/tree/main/packages/rolldown/tests/fixtures/builtin-plugin).

shulaoda:
### Description
closed #1827 

This PR implements [Vite's module-preload-polyfill plugin](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/plugins/modulePreloadPolyfill.ts) as a built-in Rolldown plugin.

### Notes

I don't know how to implement the following content `skip & isModernFlag` and whether it is necessary.
```js
// `isModernFlag` is only available during build since it is resolved by `vite:build-import-analysis`
const skip = config.command !== 'build' || config.build.ssr

...

    load(id) {
      if (id === resolvedModulePreloadPolyfillId) {
        if (skip) {
          return ''
        }
        if (!polyfillString) {
          polyfillString = `${isModernFlag}&&(${polyfill.toString()}());`
        }
        return { code: polyfillString, moduleSideEffects: true }
      }
    },
```


hyf0:
LGTM. Could you fix the `CI / Repo Validation`.

crates/rolldown_plugin_module_preload_polyfill/src/module_preload_polyfill.js
```diff
@@ -0,0 +1,66 @@
+// Port from https://github.com/vitejs/vite/blob/main/packages/vite/src/node/plugins/modulePreloadPolyfill.ts#L34
```
IWANABETHATGUY, line 1: I think you could remove this line, since it would appear in runtime code, you could claim it in your plugin code.

shulaoda: agree

shulaoda: It doesn't seem to appear in the runtime code, and it seems to be declared so in other plugins as well?

https://github.com/rolldown/rolldown/blob/f5ffbf4238e0ddc97180b4c3e257396b109b42cf/packages/rolldown/tests/fixtures/builtin-plugin/module_preload_polyfill/main.snap#L1-L3

https://github.com/rolldown/rolldown/blob/f5ffbf4238e0ddc97180b4c3e257396b109b42cf/crates/rolldown_plugin_wasm/src/wasm_runtime.js#L1

https://github.com/rolldown/rolldown/blob/f5ffbf4238e0ddc97180b4c3e257396b109b42cf/crates/rolldown_plugin_dynamic_import_vars/src/dynamic_import_helper.js#L1

shulaoda: Should I remove them together?

IWANABETHATGUY: I know why, because currently, oxc will not preserve all comments. ok, you could ignore that, we could remove it when necessary.

underfin:
> skip & isModernFlag

Look like the `skip` is unnecessary at here, we could avoid injecting the import at BuildHtmlPlugin,
The `isModernFlag` is enable at format is es, it also could avoid injecting.

underfin (approved):

underfin:
Thank you.

PR #1845, opened 2024-08-02
feat: module preload config skip

IWANABETHATGUY:


IWANABETHATGUY:
After discussing with @underfin we decide to keep align with the upstream vite plugin `1:1` as much as possible , this could reduce a lot of work when preparing migration.

underfin (approved):

PR #1846, opened 2024-08-02
chore:  add test-node-only command

IWANABETHATGUY:
### Description
1. add `test-node-only` command in justfile,  this allows us to run the test rather than test after the build, which could save a few seconds.

### DIFF
--- a/justfile
+++ b/justfile
@@ -65,21 +65,21 @@ test-rust:
     cargo test --no-fail-fast
 
 # Supported presets: all, rolldown, rollup
-test-node preset="all" *args="":
+test-node preset="all" *args="": _build-native-debug
+    just _test-node-{{preset}} {{args}}
+
+test-node-only preset="all" *args="":
     just _test-node-{{preset}} {{args}}
 
 _test-node-all:
-  just build native debug
   pnpm run --recursive --parallel --filter=!rollup-tests test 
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -65,21 +65,21 @@ test-rust:
     cargo test --no-fail-fast
 
 # Supported presets: all, rolldown, rollup
-test-node preset="all" *args="":
+test-node preset="all" *args="": _build-native-debug
+    just _test-node-{{preset}} {{args}}
+
+test-node-only preset="all" *args="":
```
hyf0, line 71: `only` is confusing. But I can't think of a good name either. You might able rename to these `_test-node-all` to `_run-node-test-all` and use `just _run-node-test-all`. I'm open to both.

hyf0 (dismissed):

justfile
IWANABETHATGUY: but the command with the prefix `_` will lose auto-completion.

hyf0 (approved):
Feel free to merge. I'm testing CI just now.

justfile
hyf0: Get it. I kind of feel being limited while writing these commands. These `_test-node-all` is kind of hacky. A real cli would be more intuative on applying args. Such as `just test-node --no-build`. However, so far let's just bare  `just`.

PR #1849, opened 2024-08-02
fix: circular ref in rolldown package bundle

IWANABETHATGUY:
### Description
```bash
[CIRCULAR_DEPENDENCY] Warning: Circular dependency: src/utils/index.ts -> src/utils/create-bundler.ts -> src/options/bindingify-input-options.ts -> src/plugin/bindingify-plugin.ts -> src/pl
ugin/plugin-context-data.ts -> src/utils/index.ts.
```


hyf0:
https://github.com/rolldown/rolldown/blob/6081511cc9db238f744bbd236ab6d50befc38e99/packages/rolldown/src/utils/index.ts#L1-L9

Main evils are here. Should just delete `index.ts` and make everything work.

IWANABETHATGUY:
> https://github.com/rolldown/rolldown/blob/6081511cc9db238f744bbd236ab6d50befc38e99/packages/rolldown/src/utils/index.ts#L1-L9
> 
> Main evils are here. Should just delete `index.ts` and make everything work.

ok

hyf0 (approved):

PR #1858, opened 2024-08-04
test: composing js plugin

### LINKED ISSUE #1709: [Test] test infra about `composingJsPlugin`

Make sure `rolldown` output has the same behavior after enabling **experimental?.enableComposingJsPlugins**

N/A

IWANABETHATGUY:
### Description
Closed https://github.com/rolldown/rolldown/issues/1709

### DIFF
--- a/packages/rolldown/tests/fixture.test.ts
+++ b/packages/rolldown/tests/fixture.test.ts
@@ -27,7 +27,32 @@ function main() {
         throw new Error(`Failed in ${testConfigPath}`, { cause: err })
       }
     })
+
   }
+
+  for (const [testConfigPath, testConfig] of Object.entries(testConfigPaths)) {
+    const dirPath = nodePath.dirname(testConfigPath)
+    const testName = dirPath.replace('./fixtures/', '')
+
+    test.skipIf(testConfig.skipComposingJsPlugin)(`${testName}-composing-js-plugin`, async () => {
+      testConfig.config = testConfig.config ?? {}
+      testConfig.config.experimental = testConfig.config.experimental ?? {}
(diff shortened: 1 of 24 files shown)


packages/rolldown/tests/fixture.test.ts
```diff
@@ -27,12 +27,36 @@ function main() {
...
+      }
+    })
+
+  }
+
 }
 
 async function compileFixture(fixturePath: string, config: TestConfig) {
   let outputOptions: OutputOptions = config.config?.output ?? {}
-  delete config.config?.output
```
IWANABETHATGUY, line 35: I don't know what this is used for, but since we ran the test twice, deleting would lead to some tests failing, so I removed it,

packages/rolldown/tests/src/types.ts
```diff
@@ -2,6 +2,7 @@ import type { RolldownOptions, RolldownOutput } from 'rolldown'
 
 export interface TestConfig {
   skip?: boolean
+  skipComposingJsPlugin?: boolean
```
IWANABETHATGUY, line 5: Add a way to temp workaround composingJsPlugin test

packages/rolldown/vitest.config.mts
```diff
@@ -6,6 +6,7 @@ export default defineConfig({
     testTimeout: 20000,
     // Disabled, Because the error printed by rust cannot be seen
     disableConsoleIntercept: true,
+    mockReset: true,
```
hyf0, line 10: Maybe a document link to describle what this do.

hyf0 (approved):

packages/rolldown/vitest.config.mts
IWANABETHATGUY: Done

PR #1866, opened 2024-08-05
feat: provide more friendly diagnostics if could not read path from a resolveId hook

### LINKED ISSUE #1851: [Feature Request]: Provide more friendly diagnostics if could not read path from a `resolveId` hook

```js
// rolldown.config.js

export default defineConfig({
	input: "./index.js",
	plugins: [
		{
			resolveId(id) {
				if (id === "test.js") {
					return id;
				}
			},
		},
	],
});
// index.js
import test from 'test.js'

```
**output**
```bash
 ERROR  Rolldown internal error: No such file or directory (os error 2)                                                                                                          

 



 ERROR  Rolldown internal error: No such file or directory (os error 2)                                                                                                          
```

n/a

shulaoda:
### Description

Closes #1851

I don't know if it should be simply regarded as a `rolldown internal error`.

At first, I tried to implement the incorrect format of `esbuild`, but later I found that `rolldown` is not enough to support me in doing so.
wecom-temp-17469-893230717fe496ece769b4a6e33e9aeb

Therefore, I attempted to implement the incorrect format of `rollup`.
wecom-temp-92390-a41e8d41a7c53c93daa4c7e305564416

Due to the current lack of support for plugins in `rolldown testing`, I am unable to conduct testing on `rust`. In addition, testing in `fixtures` can lead to `build errors`, so currently I can only ignore it.
image


IWANABETHATGUY:
I think the pr still needs some refactoring and basic infra to be merged, it should not cost too much time.

shulaoda:
> I think the pr still needs some refactoring and basic infra to be merged, it should not cost too much time.

What should I do? What should be the expected effect?

IWANABETHATGUY:
> > I think the pr still needs some refactoring and basic infra to be merged, it should not cost too much time.
> 
> What should I do? What should be the expected effect?

It is not related to your pull request, resolve ID should not be `String`, It will cause none trivial overhead. you could continue when I refactor the related type.

IWANABETHATGUY:
Also we lack the test infra to diagnostics in js side, if there is no test guaranteed, implementation is somehow meaningless(except it is a hotfix).

IWANABETHATGUY:
after #1871, you could acquire `module_request` and module_request_span, can you use module_request instead of resolved id?I think now, we have nothing to block aligning diagnostic with esbuild.

shulaoda:
> after #1871, you could acquire `module_request` and module_request_span, can you use module_request instead of resolved id?I think now, we have nothing to block aligning diagnostic with esbuild.

Let me start to try it. 😀

shulaoda:
Due to my inability to intercept `Rust`'s error output on the `js side`, it is a bit difficult to perform diagnostic test on the `js side`. Currently, I have only implemented `onerror` to obtain `Build failed`.
image

IWANABETHATGUY:
```diff
diff --git a/crates/rolldown/tests/rolldown/mod.rs b/crates/rolldown/tests/rolldown/mod.rs
index 9ef19e33..936bf634 100644
--- a/crates/rolldown/tests/rolldown/mod.rs
+++ b/crates/rolldown/tests/rolldown/mod.rs
@@ -1,2 +1,3 @@
 mod errors;
 mod issues;
+mod warnings;
diff --git a/crates/rolldown/tests/rolldown/warnings/mod.rs b/crates/rolldown/tests/rolldown/warnings/mod.rs
new file mode 100644
index 00000000..5da035d2
--- /dev/null
+++ b/crates/rolldown/tests/rolldown/warnings/mod.rs
@@ -0,0 +1,2 @@
+#[path = "./unloadable_import/mod.rs"]
+mod unloadable_import;
diff --git a/crates/rolldown/tests/rolldown/warnings/unloadable_import/artifacts.snap b/crates/rolldown/tests/rolldown/warnings/unloadable_import/artifacts.snap
new file mode 100644
index 00000000..303153af
--- /dev/null
+++ b/crates/rolldown/tests/rolldown/warnings/unloadable_import/artifacts.snap
@@ -0,0 +1,17 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+---
+# Errors
+
+## UNRESOLVED_IMPORT
+
+```text
+[UNRESOLVED_IMPORT] Error: Could not resolve test.js
+   ╭─[entry.js:1:8]
+   │
+ 1 │ import 'test.js'
+   │        ────┬────  
+   │            ╰────── No such file or directory (os error 2)
+───╯
+
+```
diff --git a/crates/rolldown/tests/rolldown/warnings/unloadable_import/entry.js b/crates/rolldown/tests/rolldown/warnings/unloadable_import/entry.js
new file mode 100644
index 00000000..0d98f1ed
--- /dev/null
+++ b/crates/rolldown/tests/rolldown/warnings/unloadable_import/entry.js
@@ -0,0 +1 @@
+import 'test.js'
diff --git a/crates/rolldown/tests/rolldown/warnings/unloadable_import/mod.rs b/crates/rolldown/tests/rolldown/warnings/unloadable_import/mod.rs
new file mode 100644
index 00000000..62029f1d
--- /dev/null
+++ b/crates/rolldown/tests/rolldown/warnings/unloadable_import/mod.rs
@@ -0,0 +1,53 @@
+use std::{borrow::Cow, sync::Arc};
+
+use rolldown::{BundlerOptions, InputItem};
+use rolldown_plugin::{
+  HookResolveIdArgs, HookResolveIdOutput, HookResolveIdReturn, Plugin, SharedPluginContext,
+};
+use rolldown_testing::{abs_file_dir, integration_test::IntegrationTest, test_config::TestMeta};
+
+#[derive(Debug)]
+struct UnLoadableImport;
+
+impl Plugin for UnLoadableImport {
+  fn name(&self) -> Cow<'static, str> {
+    "unloadable-import".into()
+  }
+
+  async fn resolve_id(
+    &self,
+    _ctx: &SharedPluginContext,
+    args: &HookResolveIdArgs<'_>,
+  ) -> HookResolveIdReturn {
+    if args.specifier == "test.js" {
+      return Ok(Some(HookResolveIdOutput {
+        id: args.specifier.to_string(),
+        ..Default::default()
+      }));
+    }
+    Ok(None)
+  }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn should_emit_diagnostic_when_could_not_load_dependency() {
+  let cwd = abs_file_dir!();
+
+  IntegrationTest::new(TestMeta {
+    expect_executed: false,
+    expect_error: true,
+    ..Default::default()
+  })
+  .run_with_plugins(
+    BundlerOptions {
+      input: Some(vec![InputItem {
+        name: Some("entry".to_string()),
+        import: "./entry.js".to_string(),
+      }]),
+      cwd: Some(cwd),
+      ..Default::default()
+    },
+    vec![Arc::new(UnLoadableImport)],
+  )
+  .await;
+}
```

IWANABETHATGUY:
Can you patch this rust test, and revert all modifications on the js side? The new JS tests seems meaningless.

crates/rolldown/tests/rolldown/errors/unresolved_import/unknown_module_type_entry/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+        external: Some(false),
+        ..Default::default()
+      }));
+    }
+    Ok(None)
+  }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn should_rewrite_dynamic_imports_that_import_external_modules() {
```
IWANABETHATGUY: Can you correct the naming?

crates/rolldown/tests/rolldown/errors/unresolved_import/unknown_module_type_entry/artifacts.snap
```diff
@@ -0,0 +1,11 @@
+---
```
IWANABETHATGUY, line 1: The test could not passed in CI machine, can you use a universal path, or just revert it.

crates/rolldown/tests/rolldown/errors/unresolved_import/unknown_module_type/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+        external: Some(false),
+        ..Default::default()
+      }));
+    }
+    Ok(None)
+  }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn should_rewrite_dynamic_imports_that_import_external_modules() {
```
IWANABETHATGUY: correct the naming

packages/rolldown/tests/fixture.test.ts
IWANABETHATGUY, line 1: Please revert this

crates/rolldown/tests/rolldown/errors/unresolved_import/unknown_module_type_entry/artifacts.snap
shulaoda: It doesn't seem like this problem has occurred?

IWANABETHATGUY: https://github.com/rolldown/rolldown/actions/runs/10265126020/job/28400667544?pr=1866#:~:text=created%20previously%20unseen%20snapshot%20/home/runner/work/rolldown/rolldown/crates/rolldown/tests/rolldown/errors/unresolved_import/unknown_module_type_entry/artifacts.snap, this should be a bug related to our test infra, cc @hyf0

IWANABETHATGUY: image, snapshot got updated without confirm

hyf0: Got it. Let me fix it.

shulaoda: I understand, is it due to the problem of absolute path?
image

IWANABETHATGUY: > I understand, is it due to the problem of absolute path? image

Yeah, should use relative path instead.

hyf0: Done at https://github.com/rolldown/rolldown/pull/1878.

crates/rolldown/src/utils/load_source.rs
```diff
@@ -61,7 +61,7 @@ pub async fn load_source(
           }
         }
       } else {
-        Err(anyhow::format_err!("Fail to guess module type for {:?}. So rolldown could load this asset correctly. Please use the load hook to load the resource", resolved_id.id))
+        Err(anyhow::format_err!("Fail to guess module type for {:?}. So rolldown could load this asset correctly. Please use the load hook to load the resource", resolved_id.debug_id(options.cwd.clone())))
```
IWANABETHATGUY: `resolved_id.debug_id(options.cwd.as_path())`

IWANABETHATGUY: > So rolldown could load

So rolldown could not load

IWANABETHATGUY:
Overall, LGTM, but I prefer the `unloadable_dependency` rather than `unresolve_import`, since the error happened in `load_source` phase, it is resolved by `resolve` hook already. cc @hyf0 what do you think?

hyf0:
> Overall, LGTM, but I prefer the `unloadable_dependency` rather than `unresolve_import`, since the error happened in `load_source` phase, it is resolved by `resolve` hook already. cc @hyf0 what do you think?

Yeah. A new error would be more proper.

crates/rolldown_error/src/events/unresolved_import.rs
```diff
@@ -1,20 +1,54 @@
 use crate::types::diagnostic_options::DiagnosticOptions;
+use arcstr::ArcStr;
+use oxc::span::Span;
 
 use super::BuildEvent;
-use std::path::PathBuf;
+
+#[derive(Debug)]
```
IWANABETHATGUY: The name is also not proper, the `id` and `source` came from `importer`, but the span came from `module_request` which is related to importee, if you cannot find a better naming, just flatten them.

shulaoda: They are interrelated, an aggregation of `Option`, and flattening can become very complex. Let's come up with a good name. 🤔☝️

shulaoda: How about this one below?
```rust
pub struct UnloadableDependencyContext {
  pub source: ArcStr,
  pub importer_id: ArcStr,
  pub importee_span: Span,
}

```

hyf0 (approved):

IWANABETHATGUY:
Thanks

PR #1871, opened 2024-08-05
chore:  add span of `module_request` in  raw import record

IWANABETHATGUY:
### Description
1. This pr add the ability to get the span of `module_request`.
2. Aiming to keep the type size unchanged, I did some refactors, If I add `Span` directly the type size will increase from `40` to `48` bytes.

Related to https://github.com/rolldown/rolldown/pull/1866

### DIFF
--- a/crates/rolldown_common/src/types/import_record.rs
+++ b/crates/rolldown_common/src/types/import_record.rs
@@ -52,23 +52,35 @@ pub struct RawImportRecord {
   pub kind: ImportKind,
   /// See [ImportRecord] for more details.
   pub namespace_ref: SymbolRef,
-  /// See [ImportRecord] for more details.
-  pub contains_import_star: bool,
-  /// See [ImportRecord] for more details.
-  pub contains_import_default: bool,
-  /// See [ImportRecord] for more details.
-  pub is_plain_import: bool,
+  pub module_request_start: u32,
+  pub import_record_meta: ImportRecordMeta,
+}
+
+bitflags::bitflags! {
+  #[derive(Debug)]
+  pub struct ImportRecordMeta: u8 {
+    /// See [ImportRecord] for more details.
+    const CONTAINS_IMPORT_STAR = 1;
+    /// See [ImportRecord] for more details.
+    const CONTAINS_IMPORT_DEFAULT = 1 << 1;
+    /// See [ImportRecord] for more details.
+    const IS_PLAIN_IMPORT = 1 << 2;
+  }
 }
 
 impl RawImportRecord {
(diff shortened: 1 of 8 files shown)


crates/rolldown_common/src/types/import_record.rs
```diff
@@ -52,23 +52,35 @@ pub struct RawImportRecord {
...
   /// See [ImportRecord] for more details.
   pub namespace_ref: SymbolRef,
-  /// See [ImportRecord] for more details.
-  pub contains_import_star: bool,
-  /// See [ImportRecord] for more details.
-  pub contains_import_default: bool,
-  /// See [ImportRecord] for more details.
-  pub is_plain_import: bool,
+  pub module_request_start: u32,
+  pub import_record_meta: ImportRecordMeta,
```
hyf0:
```suggestion
  pub meta: ImportRecordMeta,
```

crates/rolldown_common/src/types/import_record.rs
```diff
@@ -78,9 +90,13 @@ impl RawImportRecord {
       resolved_module,
       kind: self.kind,
       namespace_ref: self.namespace_ref,
-      contains_import_star: self.contains_import_star,
-      contains_import_default: self.contains_import_default,
-      is_plain_import: self.is_plain_import,
+      contains_import_star: self
```
hyf0: I guess you could also pass the `meta` directly to `ImportRecord` too.

crates/rolldown_common/src/types/import_record.rs
hyf0: If you think this might cause confusing with considering the `meta` word used in link stage. You could call it `flags` too.

crates/rolldown_common/src/types/import_record.rs
```diff
@@ -94,10 +104,5 @@ pub struct ImportRecord {
   /// We will turn `import { foo } from './cjs.js'; console.log(foo);` to `var import_foo = require_cjs(); console.log(importcjs.foo)`;
   /// `namespace_ref` represent the potential `import_foo` in above example. It's useless if we imported n esm module.
   pub namespace_ref: SymbolRef,
-  /// If it is `import * as ns from '...'` or `export * as ns from '...'`
-  pub contains_import_star: bool,
-  /// If it is `import def from '...'`, `import { default as def }`, `export { default as def }` or `export { default } from '...'`
-  pub contains_import_default: bool,
-  /// If it is `import {} from '...'` or `import '...'`
-  pub is_plain_import: bool,
+  pub meta: ImportRecordMeta,
```
IWANABETHATGUY, line 115: `Span` in `ImportRecord` is unused, so I don't pass it into `ImportRecord`

hyf0 (approved):

PR #1874, opened 2024-08-06
feat(node): support passing `custom` in `PluginContext#resolve`

underfin:
### Description

The pr intend to support `PluginContext#resolve#option#cutsom`, it will be transfer to `ResolveId#ResolveOption#custom`  to do some plugin resolve logic.

It is used at vite resolver, and the struct is complex, see [here](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/plugins/resolve.ts#L181).  We can't directly convert it to rust struct. So here store a index to reference the js side custom options. It is difficult to expand, but i think it's less usage at rust side plugins, we could improve it if we have demand.

Other things, here fixed a logic about `PluginContextData`, it should be singleton, But it make complex to shared data at  parallel worker plugin, We could leave it at now util we find a good idea.

### DIFF
--- a/packages/rolldown/src/plugin/plugin-context-data.ts
+++ b/packages/rolldown/src/plugin/plugin-context-data.ts
@@ -61,4 +61,8 @@ export class PluginContextData {
   getResolveCustom(index: number) {
     return this.resolveCustomMap.get(index)
   }
+
+  removeResolveCustom(index: number) {
+    this.resolveCustomMap.delete(index)
+  }
 }
--- a/packages/rolldown/src/plugin/plugin-context-data.ts
+++ b/packages/rolldown/src/plugin/plugin-context-data.ts
@@ -1,11 +1,12 @@
 import { BindingPluginContext } from '../binding'
-import { ModuleInfo, ModuleOptions } from '..'
+import { CustomPluginOptions, ModuleInfo, ModuleOptions } from '..'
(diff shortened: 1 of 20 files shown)


packages/rolldown/src/plugin/plugin-context-data.ts
```diff
@@ -1,11 +1,12 @@
 import { BindingPluginContext } from '../binding'
-import { ModuleInfo, ModuleOptions } from '..'
+import { CustomPluginOptions, ModuleInfo, ModuleOptions } from '..'
 import { transformModuleInfo } from '../utils/transform-module-info'
 
 export class PluginContextData {
   modules = new Map<string, ModuleInfo>()
   moduleIds: Array<string> | null = null
   moduleOptionMap = new Map<string, ModuleOptions>()
+  resolveCustomMap = new Map<number, CustomPluginOptions>()
```
hyf0, line 9: This seems causing memory leak.

hyf0:
Let me take a quick look up see if we could support truely custom in rust side. If can't, let's merge this PR for moving fast.

packages/rolldown/src/plugin/plugin-context-data.ts
underfin: Fixed it.

hyf0 (approved):

PR #1875, opened 2024-08-06
chore: showing usages of `Span` of module declarations

hyf0:


Boshen:
@overlookmotel will try and replace this with a stable id.

overlookmotel:
Please see #1899.

PR #1885, opened 2024-08-07
perf(sourcemap): pre-allocate concat sourcemap builder

underfin:


hyf0 (approved):
Interesting. Any performance gain locally?

underfin:
bundle/bundle@threejs10x-sourcemap
                        time:   [142.46 ms 143.12 ms 143.90 ms]
                        change: [−8.4670% −5.0458% −2.3010%] (p = 0.00 < 0.05)
                        Performance has improved.

overlookmotel:
@underfin Any guess why this produces 5% speed-up locally, but no sign of a difference on CI benchmarks?

underfin:
> @underfin Any guess why this produces 5% speed-up locally, but no sign of a difference on CI benchmarks?

I hasn't idea for it...

PR #1886, opened 2024-08-07
chore: remove extra external `node:assert` in test configs.

7086cmd:
`node:assert` is external by default.

_Originally posted by @IWANABETHATGUY in https://github.com/rolldown/rolldown/pull/1746#discussion_r1693361950_


hyf0 (approved):
Thanks

PR #1891, opened 2024-08-07
feat(node/rolldown): support `define` functionality

### LINKED ISSUE #186: How to support `define` options.

hyf0:
### Description

Closes #186.

So far the target expr could only be identifier or member expression, but the replacement could be any expr.

### DIFF
--- a/crates/rolldown_binding/src/options/binding_input_options/mod.rs
+++ b/crates/rolldown_binding/src/options/binding_input_options/mod.rs
@@ -74,6 +74,7 @@ pub struct BindingInputOptions {
   pub treeshake: Option<treeshake::BindingTreeshake>,
 
   pub module_types: Option<HashMap<String, String>>,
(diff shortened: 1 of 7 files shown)


hyf0:
* **#1891** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1891?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

IWANABETHATGUY (approved):

PR #1894, opened 2024-08-07
chore: refactor composing js plugin bailout logic

IWANABETHATGUY:

### DIFF
--- a/packages/rolldown/src/utils/compose-js-plugins.ts
+++ b/packages/rolldown/src/utils/compose-js-plugins.ts
@@ -4,7 +4,6 @@ import { normalizeHook } from './normalize-hook'
 import { isNullish } from './misc'
 import { BuiltinPlugin } from '../plugin/builtin-plugin'
 
-
 // FIXME: Conflict with the `skip` option in `PluginContext#resolve`. Since we can't detect it in advance,
 // we have to bailout all plugins with `resolveId` hook.
 const supportedHooks = new Set([
@@ -12,7 +11,7 @@ const supportedHooks = new Set([
   'load',
   'transform',
   'buildEnd',
-  'renderChunk'
+  'renderChunk',
 ])
 
 function createComposedPlugin(plugins: Plugin[]): Plugin {
@@ -55,14 +54,6 @@ function createComposedPlugin(plugins: Plugin[]): Plugin {
           }
           break
         }
-        // case 'resolveId': {
-        //   const handlers = batchedHooks.resolveId ?? []
-        //   batchedHooks.resolveId = handlers
-        //   if (plugin.resolveId) {
-        //     handlers.push(plugin.resolveId)
-        //   }
-        //   break
-        // }
         case 'buildEnd': {
           const handlers = batchedHooks.buildEnd ?? []
           batchedHooks.buildEnd = handlers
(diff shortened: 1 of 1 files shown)


packages/rolldown/src/utils/compose-js-plugins.ts
```diff
@@ -59,14 +54,6 @@ function createComposedPlugin(plugins: Plugin[]): Plugin {
           }
           break
         }
-        case 'resolveId': {
```
hyf0: We could keep this as long as we will skip plugin that has resolveId hook previously.

packages/rolldown/src/utils/compose-js-plugins.ts
```diff
@@ -269,13 +257,7 @@ function isComposablePlugin(plugin: RolldownPlugin): plugin is Plugin {
     return false
   }
 
-  if ('resolveId' in plugin) {
-    // FIXME: Conflict with the `skip` option in `PluginContext#resolve`. Since we can't detect it in advance,
-    // we have to bailout all plugins with `resolveId` hook.
-    return false
-  }
-
-  if (Object.keys(plugin).some((k) => unsupportedHooks.has(k))) {
```
hyf0, line 278: Seems “some” would execute less times?

packages/rolldown/src/utils/compose-js-plugins.ts
```diff
@@ -4,19 +4,14 @@ import { normalizeHook } from './normalize-hook'
...
-  'options',
-  'outputOptions',
-  'renderError',
-  'renderStart',
-  'resolveDynamicImport',
-  'writeBundle',
-  'resolveId',
+// FIXME: Conflict with the `skip` option in `PluginContext#resolve`. Since we can't detect it in advance,
+// we have to bailout all plugins with `resolveId` hook.
+const supportedHooks = new Set([
```
IWANABETHATGUY: Using `supportedHooks` rather than `unsupportedHooks` could avoid accentently unsync the logic between filter and `createComposedPlugin` impl, e.g. 

before we did not specify `footer` in `unsupportedHooks`, this could cause a panic when we try to compose two continuous plugins that use the `footer` hook.

packages/rolldown/src/utils/compose-js-plugins.ts
hyf0: Also need to consider the “name” property.

packages/rolldown/src/utils/compose-js-plugins.ts
IWANABETHATGUY: Looks like a bug, can I comment out?

packages/rolldown/src/utils/compose-js-plugins.ts
IWANABETHATGUY: added `name` and `api`

IWANABETHATGUY: Or, I can't

packages/rolldown/src/utils/compose-js-plugins.ts
IWANABETHATGUY: Ok, looks like I can't due to exaustive check

hyf0 (approved):

PR #1895, opened 2024-08-07
refactor: soundness check on composing js plugins

IWANABETHATGUY:
### Description
1. Now we could ensure each filter out unsupported hook will not throw error at runtime. 
2. limitation, we can't detect if user implementing compose method to a look that is added to `unsupportedHooks` set


hyf0:
Wait for a while. See if I could bring a better type gymnastic😂.

packages/rolldown/src/utils/compose-js-plugins.ts
```diff
@@ -271,7 +255,7 @@ function isComposablePlugin(plugin: RolldownPlugin): plugin is Plugin {
     return false
   }
 
-  if (Object.keys(plugin).some((k) => unsupportedHooks.has(k))) {
+  if (R.keys(plugin).some((hookName) => isUnsupportedHooks(hookName))) {
```
hyf0, line 258: `isComposablePlugin` now shared the same logic on detecting unsupported plugin with `createComposedPlugin`

packages/rolldown/package.json
```diff
@@ -114,9 +114,9 @@
     "fs-extra": "^11.2.0",
     "glob": "^10.4.1",
     "locate-character": "^3.0.0",
-    "lodash-es": "^4.17.21",
     "mri": "^1.2.0",
     "npm-run-all2": "^6.2.0",
+    "remeda": "^2.10.0",
```
hyf0, line 119: `remeda` is typed-friendly utils library like lodash. It provides strong-typed version of `object.keys`, `map`, `filter` etc..

hyf0 (approved):

PR #1896, opened 2024-08-07
chore(rust): remove extra spaces in CJS.

7086cmd:
There are some extra blanks after the field:

```js
"Object.keys($NAME).forEach(function (k) {
    enumerable: true,
    get: function () { return $NAME[k]; }
  });
});
```
For example:
QQ_1723044960308


hyf0 (approved):

hyf0:
Thanks

PR #1898, opened 2024-08-07
feat: align OutputOptions#esModule with rollup

### LINKED ISSUE #1888: [Bug]: Align `OutputOptions#esModule` with rollup

Not needed.

### What is expected?

https://github.com/rollup/rollup/blob/df12edfea6e9c1a71bda1a01bed1ab787b7514d5/src/rollup/types.d.ts#L716

https://github.com/rolldown/rolldown/blob/d19beb4e463e5ed502cc56ffcb196a5333a4c843/packages/rolldown/src/options/output-options.ts#L44-L49

### System Info

```Shell
Not needed.
```

### Any additional comments?

_No response_

shulaoda:
### Description
closes #1888 

Need some help, it doesn't seem to work.
https://github.com/rolldown/rolldown/blob/e0751e4db539ce8e25ca8f31bb027475a0dd9e98/crates/rolldown_binding/src/options/binding_output_options/mod.rs#L37-L38
`Error: Failed to load native binding`


shulaoda:
@7086cmd cc

shulaoda:
@hyf0 cc

7086cmd:
You don't need to modify the Rust side here.

packages/rolldown/src/options/bindingify-output-options.ts
```diff
@@ -75,20 +75,13 @@ function bindingifySourcemap(
...
-    case 'never':
-    case false:
-      return 'never'
-
-    case 'if-default-prop':
-    case undefined:
-      return 'if-default-prop'
+  if (typeof esModule === 'boolean') {
+    return esModule
+  }
```
7086cmd: You can remove the following lines:

```js
    case ‘never’:
```

and

```js
    case ‘always’:
```

packages/rolldown/src/options/normalized-output-options.ts
```diff
@@ -22,7 +22,7 @@ export interface NormalizedOutputOptions extends OutputOptions {
   footer: AddonFunction
   intro: AddonFunction
   outro: AddonFunction
-  esModule: 'always' | 'never' | 'if-default-prop' | boolean
+  esModule: boolean | 'if-default-prop'
```
7086cmd, line 25: You can modify it to `’always’ | ‘never’ | ‘if-default-prop’`.

crates/rolldown_binding/src/options/binding_output_options/mod.rs
```diff
@@ -32,8 +33,9 @@ pub struct BindingOutputOptions {
   // compact: boolean;
   pub dir: Option<String>,
   // pub entry_file_names: String, // | ((chunkInfo: PreRenderedChunk) => string)
-  #[napi(ts_type = "'always' | 'never' | 'if-default-prop'")]
-  pub es_module: Option<String>,
+  #[serde(deserialize_with = "deserialize_es_module")]
+  #[napi(ts_type = "boolean | 'if-default-prop'")]
+  pub es_module: Option<Either<bool, String>>,
```
7086cmd: Please maintain the current state.

7086cmd (changes requested):
> Yeah. `rolldown` crate should keep using the current definition, which I think it is more proper. But js and binding crate should just pass values.

_Originally posted by @hyf0 in https://github.com/rolldown/rolldown/issues/1888#issuecomment-2272974628_

packages/rolldown/src/options/bindingify-output-options.ts
7086cmd: In the configuration file, we can simply remove the boolean value. During the normalization step on the JavaScript side, we only need to transform the boolean to the `’always’ | ‘never’`. On the Rust side, let’s maintain the current value to prevent panic.

7086cmd:
> Implementing this idea may present challenges. My proposed approach involves only passing the `boolean` value from the configuration to the JavaScript side, where it will be transformed into the appropriate `never` and `always` from `true` or `false` value during the normalization step. While using a union type in Rust may be acceptable, I personally prefer to avoid it.

_Originally posted by @7086cmd in https://github.com/rolldown/rolldown/issues/1888#issuecomment-2274711587_

packages/rolldown/src/options/bindingify-output-options.ts
shulaoda: Yes, I understand what you mean. 
I thought the modification on the `binding crate` side was feasible, as mentioned by  @hyf0, but it seems to have encountered the problem you encountered. I don't know if it's a problem with my code writing.

7086cmd: > ### Rationale for Not Using `boolean | 'if-default-prop'`
> 
> During the conversion from JavaScript to Rust, the program encountered a panic when the source value was a boolean. Overriding `true` or `false` with a string was considered inelegant, so this approach was not used.

In the initial prototype of the previous PR, I also attempted to use it. Regrettably, I experienced a similar panic.

7086cmd (approved):
It is acceptable, and I am uncertain whether there should be an inefficient and confusing approach to passing `boolean | string` from JavaScript to Rust.

underfin (approved):

underfin:
Thank you.

PR #1899, opened 2024-08-07
Identify spanless module decls using memory address

overlookmotel:
Demo only, not for merging. I have never peered inside Rolldown codebase before, so have no idea what I'm doing!

I'm not 100% sure of the problem we're trying to solve, but my rough understanding is: you are using `Span`s of AST nodes as unique identifiers, but this breaks down where the nodes are newly generated and so have a dummy span `SPAN`.

This is built on top of #1875, demonstrating how you can use memory address of AST nodes as a unique identifier, in place of `Span`.

If you don't like the weird pointer stuff, there is one other method I can think of to tackle this problem:

When you insert new `import` / `export` statements which need to be tracked, set a fake span on them with `Span::new(u32::MAX, id)`. This can never clash with a real span, and will be unique as long as `id` is incremented on each insertion. Then later on, when finished processing the AST, set the spans of these nodes back to empty `SPAN`.


Boshen:
> Span::new(u32::MAX, id)

This is a nice temporary hack ;-)

hyf0:
> When you insert new import / export statements which need to be tracked, set a fake span on them with Span::new(u32::MAX, id). This can never clash with a real span, and will be unique as long as id is incremented on each insertion. Then later on, when finished processing the AST, set the spans of these nodes back to empty SPAN.

Yeah. I thought this before, we could have a new ast pass the will fill up empty span and ensure the uniqueness. I can't think the downside of this solution, but it seems tricky.

However, I think it's ok we use this solution.

overlookmotel:
I think we can come up with a better solution later on which allows you to "stash" references to AST nodes and then "unstash" them later on to mutate them (using some unsafe code). But I'm still working out the safety model of Oxc's `Traverse` trait which would enable this.

For now either the memory address or fake spans method I think should work as a hacky stopgap.

Shall I close this PR now?

hyf0:
Yeah. You could close this one. Using `span` would be more proper even in a hack way. It's just import declarations that rely on this. Ensuring uniqueness of span would also benefits other parts. https://github.com/rolldown/rolldown/pull/1927

overlookmotel:
OK, closing.

A couple of notes:

Just for your info, I have been thinking that at some point it could be good to make `Span` a type which cannot be modified once parser has created it. This would allow us to make it a safety invariant that slicing source text with a span `source_text[span.start..span.end]` will always be on UTF-8 char boundaries, and we can use the faster [std::str::get_unchecked](https://doc.rust-lang.org/stable/std/primitive.str.html#method.get_unchecked).

If we did that, it might preclude your use of `Span` here (or maybe not - I guess `str[n..n]` is always safe if `n <= source_text.len()`).

Secondly, the `MemoryAddress` solution shown in this PR looks pretty hacky, but actually it's not. It is a fundamental guarantee of Oxc's arena allocator that the contents of a `Box<T>` never moves in memory, so the `MemoryAddress` technique is completely legitimate. We may add an API for it to Oxc quite soon to support some other use cases we have.

PR #1901, opened 2024-08-08
chore: bump oxc to 0.24.0

Boshen:
This update adds symbols and scopes to the minifier for better constant folding in the near future.

We'll try and fix this multiple semantic pass problem soon.


Boshen:
hmm ... this is not good, let me tweak some stuff.

Boshen:
Discussed with @hyf0 offline that we'll accept this performance regression in favor of correctness.

hyf0:
Performance is down mainly due to oxc brings up a feature-complete DCE. So we do more work and take more time.

Related:

- https://github.com/oxc-project/backlog/issues/85

hyf0 (approved):

PR #1902, opened 2024-08-08
fix:  share replace global define config

IWANABETHATGUY:


crates/rolldown/src/module_loader/module_loader.rs
```diff
@@ -82,8 +84,29 @@ impl ModuleLoader {
...
-    let common_data =
-      Arc::new(TaskContext { options: Arc::clone(&options), tx, resolver, fs, plugin_driver });
+    let common_data = Arc::new(TaskContext {
+      options: Arc::clone(&options),
+      tx,
+      resolver,
+      fs,
+      plugin_driver,
+      meta: TaskContextMeta {
+        replace_global_define_config: if !options.define.is_empty() {
```
hyf0: I guess you could just put `ReplaceGlobalDefinesConfig` in the `options.define`.

IWANABETHATGUY: The only thing I am hesitated to do that is if we want the raw `NormalizedBundlerOptions` in the future. For now the `NormalizedBundlerOptions` is almost `1:1` for `rolldown.config.js`. Is it worth to preserve `1:1` semantic of the original configuration?

hyf0: No need worried about that. The purpose of `NormalizedBundlerOptions` is make we happy, no need to dealing with something raw. However, the current implementation of the PR is also fine.

hyf0 (approved):

crates/rolldown/src/module_loader/module_loader.rs
IWANABETHATGUY: Ok, then I will replace it until we find it is necessary to keep the config as is.

IWANABETHATGUY: If we  replace `define` with `ReplaceGlobalDefinesConfig`(but we need handle error and propagate errors), which will lead change all upstream function return types to `Result<T>`. let's merge it first.

PR #1904, opened 2024-08-08
refactor: avoid unsafe code at BindingOutputs

underfin:
### Description

The `BindingOutputs` previous using `std::mem:: transmute` to owner `Vec<Output>` mutate reference, but meet segment fault at `delete bundle[key]`.

The pr intend to avoid using unsafe code to do it, here using `Arc<Mutex<RefCell<Vec<Output>>>>` instead of it, it means we need to owner `Vec<Output>` and return `Vec<Output>`  at `generateBundle` and `writeBundle` hook. 
Here not change the `generateBundle#bundle` to  `Arc<Mutex<RefCell<Vec<Output>>>>`, because it will make the rust side API is difficult to mutate. I only create it and take it to  `Vec<Output>` at binding side. 

Here meet one problem is the napi does not drop the struct to free the reference number of`Arc<Mutex<RefCell<Vec<Output>>>>`, so here i manually to drop it.

The pervious `FinalBindingOutputs` also could be using it at now, so i updated it.

### DIFF
--- a/crates/rolldown_binding/src/options/plugin/js_plugin.rs
+++ b/crates/rolldown_binding/src/options/plugin/js_plugin.rs
@@ -3,7 +3,11 @@ use crate::types::{
   js_callback::MaybeAsyncJsCallbackExt,
 };
 use rolldown_plugin::{Plugin, __inner::SharedPluginable, typedmap::TypedMapKey};
-use std::{borrow::Cow, ops::Deref, sync::Arc};
+use std::{
+  borrow::Cow,
+  ops::Deref,
+  sync::{Arc, Mutex},
+};
 
 use super::{
   binding_transform_context::BindingTransformPluginContext,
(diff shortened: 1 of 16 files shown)


hyf0:
Yeah. I'm doing the same thing here: https://github.com/rolldown/rolldown/pull/1905.

crates/rolldown_binding/src/types/binding_outputs.rs
```diff
@@ -32,60 +43,35 @@ impl BindingOutputs {
...
-        assets.push(BindingOutputAsset::new(unsafe { std::mem::transmute(asset.as_mut()) }));
-      }
-      rolldown_common::Output::Chunk(_) => {}
-    });
-    assets
+  /// TODO
+  /// The napi look like is not drop the strut after the function call, so we need to call it manually.
+  #[allow(clippy::should_implement_trait)]
+  #[napi]
+  pub fn drop(&mut self) {
```
hyf0, line 74: This is not needed in https://github.com/rolldown/rolldown/pull/1905, since the binding code only hold a bunch of `weak` references.

crates/rolldown_binding/src/types/binding_outputs.rs
```diff
@@ -1,29 +1,40 @@
...
+
 use napi_derive::napi;
 
 use super::{binding_output_asset::BindingOutputAsset, binding_output_chunk::BindingOutputChunk};
 
-/// The `BindingOutputs` owner `Vec<Output>` the mutable reference, it avoid `Clone` at call `writeBundle/generateBundle` hook, and make it mutable.
 #[napi]
 pub struct BindingOutputs {
-  inner: &'static mut Vec<rolldown_common::Output>,
+  inner: Arc<Mutex<RefCell<Vec<rolldown_common::Output>>>>,
```
hyf0, line 13: Is the `RefCell` necessary? There is already a `Mutex`.

crates/rolldown_binding/src/types/binding_outputs.rs
```diff
@@ -1,29 +1,40 @@
...
-    });
+    self
+      .inner
+      .lock()
+      .expect("should hasn't lock at BindingOutputs")
+      .borrow_mut()
+      .iter_mut()
+      .for_each(|o| match o {
+        rolldown_common::Output::Chunk(chunk) => {
+          chunks.push(BindingOutputChunk::new(unsafe { std::mem::transmute(chunk.as_mut()) }));
```
hyf0, line 34: I'm also thinking if we could remove this part of unsafe.

crates/rolldown_binding/src/options/plugin/js_plugin.rs
```diff
@@ -291,32 +295,36 @@ impl Plugin for JsPlugin {
...
+    mut bundle: Vec<rolldown_common::Output>,
+  ) -> rolldown_plugin::HookWriteBundleReturn {
     if let Some(cb) = &self.write_bundle {
-      cb.await_call((
-        Arc::clone(ctx).into(),
-        BindingOutputs::new(unsafe { std::mem::transmute(bundle) }),
-      ))
-      .await?;
+      let output = Arc::new(Mutex::new(std::cell::RefCell::new(bundle)));
+      cb.await_call((Arc::clone(ctx).into(), BindingOutputs::new(Arc::clone(&output)))).await?;
```
hyf0, line 321: This spreads `Arc` arounds, we probably should avoid that if possible to prevent the panic latter calling `Arc::into_inner`.

crates/rolldown_binding/src/types/binding_outputs.rs
underfin: Yeah. The `weak` is better than here.

crates/rolldown_binding/src/types/binding_outputs.rs
underfin: It is easy to muate `RefCell<T>` than `Arc<T>`.

crates/rolldown_binding/src/options/plugin/js_plugin.rs
underfin: The `Weak` avoid it.

crates/rolldown_binding/src/types/binding_outputs.rs
IWANABETHATGUY: Can you give an example? I don't see benefits.

underfin:
close it favor #1905

crates/rolldown_binding/src/types/binding_outputs.rs
underfin: Yeah. I get it.

PR #1907, opened 2024-08-08
feat: add module type to transform hook

IWANABETHATGUY:


crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
```diff
@@ -65,11 +65,11 @@ pub struct BindingPluginOptions {
 
   #[serde(skip_deserializing)]
   #[napi(
-    ts_type = "(ctx:  BindingTransformPluginContext, id: string, code: string) => MaybePromise<VoidNullable<BindingHookTransformOutput>>"
+    ts_type = "(ctx:  BindingTransformPluginContext, id: string, code: string, module_type: string) => MaybePromise<VoidNullable<BindingHookTransformOutput>>"
```
hyf0:
```suggestion
    ts_type = "(ctx:  BindingTransformPluginContext, id: string, code: string, meta: TransformHookExtraArgs) => MaybePromise<VoidNullable<BindingHookTransformOutput>>"
```

would be better. We might need to add more fields in it.

packages/rolldown/tests/fixtures/plugin/transform/_config.ts
```diff
@@ -8,10 +8,11 @@ export default defineTest({
     plugins: [
       {
         name: 'test-plugin',
-        transform: function (code, id) {
+        transform: function (code, id, module_type) {
```
hyf0:
```suggestion
        transform: function (code, id, meta) {
```

packages/rolldown/tests/fixtures/plugin/transform/_config.ts
```diff
@@ -8,10 +8,11 @@ export default defineTest({
...
           transformFn()
           if (id.endsWith('foo.js')) {
             expect(code).toStrictEqual('')
+            expect(module_type).toEqual('js')
```
hyf0:
```suggestion
            expect(meta.module_type).toEqual('js')
```

Looks more suitable.

crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
hyf0: We could call it module meta in js side, since it's more intutiive. In rust side, you could call it anything with a reasonable name.

packages/rolldown/src/plugin/index.ts
```diff
@@ -24,6 +24,8 @@ import { ParallelPlugin } from './parallel-plugin'
 
 export type ModuleSideEffects = boolean | 'no-treeshake' | null
 
+export type ModuleType = 'js' | 'jsx' | 'ts' | 'tsx' | 'json' | 'text' | 'base64' | 'dataurl' | 'binary' | 'empty' | (string & {})
```
hyf0: Might want to link a issue here.

crates/rolldown_binding/src/options/plugin/binding_plugin_options.rs
IWANABETHATGUY: why not `TransformHookExtra` rather than `TransformHookExtraOptions` ? It is metadata not options.

hyf0: Yeah. We could use `TransformHookExtra`.  We could use names different from the js side as long as it's reasonable in rust side.

IWANABETHATGUY: TransformHookExtraArgs sounds more reasonable.

hyf0 (approved):

PR #1908, opened 2024-08-08
chore(docs): fix the command instruction

7086cmd:
The `pnpm docs` command is used to open the official documentation website from `npm`. Therefore, we need to modify the command to `run docs`.

Additionally, shall we modify the `package.json`’s `name` field to `rolldown`? When running `pnpm docs` in the root directory, it will open `https://www.npmjs.com/package/monorepo`, as the `name` field is currently `rolldown`.


hyf0 (approved):
thanks

PR #1914, opened 2024-08-09
feat: transform plugin

IWANABETHATGUY:
### Description
1. port vite transform plugin
2. Only added basic test case, the rest testcases about `exclude` and `include` will be added after a new option to disable default transform behavior.


Cargo.toml
```diff
@@ -123,7 +124,7 @@ lightningcss        = { version = "1.0.0-alpha.57" }
 memchr              = "2.7.2"
 mimalloc            = "0.1.42"
 mime                = "0.3.17"
-napi                = { version = "3.0.0-alpha", features = ["async", "deferred_trace"] }
+napi                = { version = "3.0.0-alpha", features = ["async", "deferred_trace", "anyhow"] }
```
IWANABETHATGUY, line 127: allow `napi::Error::from(anyhow::Error)`

packages/rolldown/src/plugin/builtin-plugin.ts
```diff
@@ -68,6 +77,10 @@ export function wasmPlugin() {
   return new WasmPlugin()
 }
 
+export function ecmaTransformPlugin(config?: TransformPluginConfig) {
```
hyf0:
```suggestion
export function transformPlugin(config?: TransformPluginConfig) {
```

I think you could just call it `transform` and don't need to worried about if it's conflicted if we add css, other language support.

hyf0: Just like the minifier and other config. 

`Input#entryFilenames` `Output#minifier` is for JS.
`Input#cssEntryFilenames` `Output#cssMinifier` for css.

We could use non-prefix word to represent config related to JS, since it's most common part of bundler.

hyf0: Better add some jsdoc on this function and saying that in this plugin `jsx`, `ts`, `tsx` would be transformed and `ModuleType` would be changed too.

hyf0 (approved):

PR #1915, opened 2024-08-09
feat: support flags for hybrid regex

### LINKED ISSUE #1911: [Bug]: Should support flag in HybridRegex

n/a

### What is expected?

support flag in JsRegex

did not support yet

### System Info

```Shell
n/a
```

### Any additional comments?

_No response_

shulaoda:
### Description
closes #1911

### DIFF
--- a/crates/rolldown_common/src/types/js_regex.rs
+++ b/crates/rolldown_common/src/types/js_regex.rs
@@ -16,15 +16,8 @@ impl HybridRegex {
   }
 
   pub fn with_flags(pattern: &str, flags: &str) -> anyhow::Result<Self> {
-    let mut result = String::new();
-    for c in flags.bytes() {
-      match c {
-        b'i' | b'm' | b's' | b'u' | b'R' | b'U' | b'x' => result.push(c as char),
-        _ => {}
-      }
-    }
+    let regex_pattern = if flags.is_empty() { pattern } else { &format!("(?{flags}){pattern}") };
 
-    let regex_pattern = if result.is_empty() { pattern } else { &format!("(?{result}){pattern}") };
     match regex::Regex::new(regex_pattern).map(HybridRegex::Optimize) {
       Ok(reg) => Ok(reg),
       Err(_) => regress::Regex::with_flags(pattern, flags)
--- a/crates/rolldown_common/src/types/js_regex.rs
+++ b/crates/rolldown_common/src/types/js_regex.rs
@@ -1,3 +1,5 @@
+use regress::Flags;
+
 /// According to the doc of `regress`, https://docs.rs/regress/0.10.0/regress/#comparison-to-regex-crate
 /// **regress supports features that regex does not, in particular backreferences and zero-width lookaround assertions.**
 /// these features are not commonly used, so in most cases the slow path will not be reached.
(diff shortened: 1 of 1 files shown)


crates/rolldown_common/src/types/js_regex.rs
```diff
@@ -7,6 +9,64 @@ pub enum HybridRegex {
...
+        b's' => flags |= 1 << 2, // 0000_0100
+        b'u' => flags |= 1 << 3, // 0000_1000
+        b'R' => flags |= 1 << 4, // 0001_0000
+        b'U' => flags |= 1 << 5, // 0010_0000
+        b'x' => flags |= 1 << 6, // 0100_0000
+        _ => {}
+      }
+    }
+
+    let mut result = String::new();
```
7086cmd: I was thinking, would it be better if you moved this to the `rolldown_utils` module?

shulaoda: Those functionalities is only part of the internal implementation of `HybridRegex` and is not exposed externally. I think it might be unnecessary. 🤔

7086cmd: I think you might move the `to_regex_flags` to the utils module. It might not be directly related to the structure 🤔 .

IWANABETHATGUY:
The impl is too complex, just pass flag `str` to `regress::Regex::with_flags` is enough, the error is expected, if user pass some flag `regress` don't support

IWANABETHATGUY:
What you only need is handle flags for `regex` https://docs.rs/regex/latest/regex/#grouping-and-flags

shulaoda:
> What you only need is handle flags for `regex` https://docs.rs/regex/latest/regex/#grouping-and-flags

Should I verify the repeatability of `flags` for `regex` or let `regex` handle it?

IWANABETHATGUY:
> > What you only need is handle flags for `regex` https://docs.rs/regex/latest/regex/#grouping-and-flags
> 
> Should I verify the repeatability of `flags` for `regex` or let `regex` handle it?

No, just let lib handle it, the order does not matter.

crates/rolldown_common/src/types/js_regex.rs
shulaoda: It seems like I've made things a bit complicated. 😂

shulaoda:
In addition, I changed the confusing name `source` to `pattern`.

crates/rolldown_common/src/types/js_regex.rs
```diff
@@ -8,10 +8,28 @@ pub enum HybridRegex {
...
+  pub fn with_flags(pattern: &str, flags: &str) -> anyhow::Result<Self> {
+    let mut result = String::new();
+    for c in flags.bytes() {
+      match c {
+        b'i' | b'm' | b's' | b'u' | b'R' | b'U' | b'x' => result.push(c as char),
+        _ => {}
+      }
+    }
+
+    let regex_pattern = if result.is_empty() { pattern } else { &format!("(?{result}){pattern}") };
```
IWANABETHATGUY:
```suggestion
    let regex_pattern = if flags.is_empty() { pattern } else { &format!("(?{flags}){pattern}") };

```

IWANABETHATGUY: The extra allocation is not necessary, we should fallback to `regress` and even raise an error if the `flags` is not supported.

IWANABETHATGUY:
Also could you add a basic test to ensure regression?
```
mod test {
  #[test]
  fn with_flags() {
    let reg = super::HybridRegex::with_flags("a", "i").unwrap();
    assert!(reg.matches("A"));

    let reg = super::HybridRegex::new("a").unwrap();
    assert!(!reg.matches("A"));
  }
}

```

IWANABETHATGUY (approved):
Thanks

PR #1916, opened 2024-08-09
feat: allow rewrite `moduleType` in transform Hook

IWANABETHATGUY:


crates/rolldown/src/utils/transform_source.rs
```diff
@@ -10,14 +10,15 @@ pub async fn transform_source(
   source: String,
   sourcemap_chain: &mut Vec<SourceMap>,
   side_effects: &mut Option<HookSideEffects>,
-  module_type: &ModuleType,
+  module_type: &mut ModuleType,
 ) -> Result<String> {
   plugin_driver
     .transform(
-      &HookTransformArgs { id: &resolved_id.id, code: &source, module_type },
+      &HookTransformArgs { id: &resolved_id.id, code: &source, module_type: &ModuleType::Empty },
```
IWANABETHATGUY, line 17: a dummy placeholder, since `module_type` needs to be `mut`.

hyf0 (approved):

PR #1926, opened 2024-08-10
chore: bump oxc to 0.24.1

Boshen:
This releases adds 

## inject plugin

```rust
    let config =
        InjectGlobalVariablesConfig::new(vec![InjectImport::named_specifier("jquery", None, "$")]);
    InjectGlobalVariables::new(&allocator, config).build(&mut symbols, &mut scopes, program);
```

API: https://github.com/oxc-project/oxc/blob/970c36942cffb0d0a57f39e4170add9bce9f74ae/crates/oxc_minifier/src/plugins/inject_global_variables.rs#L10-L56

I'll draft another PR with inject plugin usage for you to integrate.

PR #1927, opened 2024-08-10
feat(rust): ensure span uniqueness of oxc AST

hyf0:
### Description

Rolldown uses `Span` as the way to indentify AST ndoes. Now the `inject` feature would generate improt statements with empty Span. They might be duplicated, we need ensure it's uniqueness.

Contexts:
- Closes https://github.com/rolldown/rolldown/pull/1875
- https://github.com/rolldown/rolldown/pull/1926
- https://github.com/rolldown/rolldown/pull/1899#issuecomment-2274936578

### DIFF
--- a/crates/rolldown/src/utils/ecma_visitors/mod.rs
+++ b/crates/rolldown/src/utils/ecma_visitors/mod.rs
@@ -1,0 +1,48 @@
+use oxc::{
+  ast::{visit::walk_mut, VisitMut},
+  span::{GetSpanMut, Span},
+};
+use rustc_hash::FxHashSet;
+
+/// Make sure there aren't any duplicate spans in the AST.
+pub struct EnsureSpanUniqueness {
+  // visited_spans: FxHashMap</* start */ u32, /* ends */ FxHashSet<u32>>,
+  visited_spans: FxHashSet<Span>,
+  next_unique_span_start: u32,
+}
+
+impl<'a> VisitMut<'a> for EnsureSpanUniqueness {
+  fn visit_program(&mut self, it: &mut oxc::ast::ast::Program<'a>) {
+    self.next_unique_span_start = it.span.end + 1;
+    walk_mut::walk_program(self, it);
+  }
+
+  // TODO: it's better use `visit_span`, but it's not implemented yet by oxc. https://github.com/oxc-project/oxc/issues/4799
+  fn visit_module_declaration(&mut self, it: &mut oxc::ast::ast::ModuleDeclaration<'a>) {
+    self.ensure_uniqueness(it.span_mut());
+    walk_mut::walk_module_declaration(self, it);
+  }
+}
+
+impl EnsureSpanUniqueness {
+  pub fn new() -> Self {
+    Self { visited_spans: FxHashSet::default(), next_unique_span_start: 1 }
+  }
+
+  fn ensure_uniqueness(&mut self, span: &mut Span) {
+    if self.visited_spans.contains(span) {
+      *span = self.generate_unique_span();
+    }
+    self.visited_spans.insert(*span);
+  }
+
+  fn generate_unique_span(&mut self) -> Span {
+    let mut span_candidate = Span::new(self.next_unique_span_start, self.next_unique_span_start);
+    while self.visited_spans.contains(&span_candidate) {
+      self.next_unique_span_start += 1;
+      span_candidate = Span::new(self.next_unique_span_start, self.next_unique_span_start);
+    }
+    debug_assert!(span_candidate.is_empty());
+    span_candidate
+  }
+}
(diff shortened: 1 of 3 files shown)


hyf0:
* **#1927** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1927?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/rolldown/src/utils/ecma_visitors/mod.rs
```diff
@@ -0,0 +1,48 @@
+use oxc::{
+  ast::{visit::walk_mut, VisitMut},
+  span::{GetSpanMut, Span},
+};
+use rustc_hash::FxHashSet;
+
+/// Make sure there aren't any duplicate spans in the AST.
+pub struct EnsureSpanUniqueness {
```
IWANABETHATGUY, line 8: Maybe rename to `EnsureModuleDeclarationSpanUniqueness`?

hyf0: It's not just for `ModuleDeclaration`s. Other nodes type would be added after https://github.com/oxc-project/oxc/issues/4799.

IWANABETHATGUY: I see

overlookmotel:
A few thoughts:

1. If you're only checking spans of `ModuleDeclaration`s, you don't need a full AST traversal. They can only be at top level, so you could just iterate through `program.statements`.

2. If you're only looking for duplicate "fake" spans, as `start` and `end` of "fake" spans are always equal, `visited_spans: FxHashSet<Span>` can be replaced with `visited_span_starts: FxHashSet<u32>` and `ensure_uniqueness` can be:

```rs
fn ensure_uniqueness(&mut self, span: &mut Span) {
    // `span.start == span.end` will be false for all real spans,
    // and equality check is much cheaper than hashmap lookup
    if span.start == span.end && self.visited_span_starts.contains(span.start) {
        *span = self.generate_unique_span();
        self.visited_span_starts.insert(span.start);
    }
}
```

3. Please see my comment on https://github.com/oxc-project/oxc/issues/4799 as to why using `visit_span` is probably a bad idea.

hyf0:
> If you're only checking spans of ModuleDeclarations.  you don't need a full AST traversal.

There are other parts need this too, including `MemberExpression`, so we need  a almost full AST traversal. I didn't do it because they aren't required for supporting https://github.com/rolldown/rolldown/pull/1926.

>  If you're only looking for duplicate "fake" spans, as start and end of "fake" spans are always equal, visited_spans: FxHashSet<Span> can be replaced with visited_span_starts: FxHashSet<u32> and ensure_uniqueness can be:

Nope. We need to make sure all AST nodes don't have duplicate span.

---

When I say "all AST nodes" it means almost all AST nodes, since we care about `MemberExpression`. A full AST traversal seems unavoidable.

overlookmotel:
Ah ha. Thanks for explaining. What `MemberExpression`s are you looking for? I can see you'd need to also track `import()` and `require()`, but I can't imagine why `MemberExpression`s.

> Nope. We need to make sure all AST nodes don't have duplicate span.

Out of interest, how is it possible for 2 AST nodes to have same span if they are "real" spans i.e. from Oxc parser? Is this something that can happen when you combine two ASTs into one?

Sorry for "noob" questions. I have very little knowledge of Rolldown, but would like to gain some!

IWANABETHATGUY:
> Ah ha. Thanks for explaining. What `MemberExpression`s are you looking for? I can see you'd need to also track `import()` and `require()`, but I can't imagine why `MemberExpression`s.
> 
> > Nope. We need to make sure all AST nodes don't have duplicate span.
> 
> Out of interest, how is it possible for 2 AST nodes to have same span if they are "real" spans i.e. from Oxc parser? Is this something that can happen when you combine two ASTs into one?
> 
> Sorry for "noob" questions. I have very little knowledge of Rolldown, but would like to gain some!

After oxc transform (or any custom transform), the newly added `Ast` will have `SPAN` which are `(0, 0)`. They are both same

hyf0:
> Ah ha. Thanks for explaining. What `MemberExpression`s are you looking for? I can see you'd need to also track `import()` and `require()`, but I can't imagine why `MemberExpression`s.
> 
> > Nope. We need to make sure all AST nodes don't have duplicate span.
> 
> Out of interest, how is it possible for 2 AST nodes to have same span if they are "real" spans i.e. from Oxc parser? Is this something that can happen when you combine two ASTs into one?
> 
> Sorry for "noob" questions. I have very little knowledge of Rolldown, but would like to gain some!

https://github.com/rolldown/rolldown/pull/1818#discussion_r1699442259. We will resolve `MemberExpression` stored in the first visit to a `symbol` and rewrite the `MemberExpression` to that `Symbol`. We rely on `Span` to identify the `MemberExpression` if they are the same in two visit pass.

PR #1929, opened 2024-08-10
feat: support plugin order

underfin:
### Description

Support  plugin order.

TODO:

The `ComposingJsPlugin` also need to support it.


hyf0:
I'm doing review now. All changes are fine, just let me tweak something and we could merge it.

hyf0 (approved):
Good works!

- I'll make `ComposingPlugin` bailout for plugins that use `order` in the next PR. There are ways to make `ComposingPlugin` support this `order`, but I haven't think through it.
- We should stored called hooks meta and could be used latter for module filter features.

PR #1930, opened 2024-08-10
fix: 🐛 auto guess module type when load hook no specify

IWANABETHATGUY:


packages/rolldown/tests/fixtures/plugin/load/guess-module-type/_config.ts
```diff
@@ -0,0 +1,21 @@
+import { defineTest } from '@tests'
```
IWANABETHATGUY, line 1: Before the test will fail, since load a module and did not return module type will fallback to `Js`. Oxc will raise a parse error.

PR #1931, opened 2024-08-10
fix: Add namespace for each builtin plugin

IWANABETHATGUY:
### Description
1. use kebab case for naming plugin, just follow https://github.com/rolldown/vite/blob/3bf86e3f715c952a032b476b60c8c869e9c50f3f/packages/vite/src/node/plugins/dynamicImportVars.ts#L174-L174

### DIFF
--- a/crates/rolldown_plugin_transform/src/lib.rs
+++ b/crates/rolldown_plugin_transform/src/lib.rs
@@ -40,7 +40,7 @@ pub struct TransformPlugin {
 /// only handle ecma like syntax, `jsx`,`tsx`,`ts`
 impl Plugin for TransformPlugin {
   fn name(&self) -> Cow<'static, str> {
-    Cow::Borrowed("transform")
+    Cow::Borrowed("builtin:transform")
   }
 
   async fn transform(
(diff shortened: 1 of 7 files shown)


crates/rolldown_plugin_transform/src/lib.rs
```diff
@@ -40,7 +40,7 @@ pub struct TransformPlugin {
 /// only handle ecma like syntax, `jsx`,`tsx`,`ts`
 impl Plugin for TransformPlugin {
   fn name(&self) -> Cow<'static, str> {
-    Cow::Borrowed("module_preload_polyfill")
+    Cow::Borrowed("transform")
```
hyf0: We should have a convention for this. `builtin:transform` would be more proper I guess.

@underfin cc

hyf0:
```suggestion
    Cow::Borrowed("builtin:transform")
```

IWANABETHATGUY: All plugins don't have a namespace for now, prefix a namespace is better.

IWANABETHATGUY:
Let's prefix each built-in plugin name with `builtin`.

hyf0 (approved):

PR #1933, opened 2024-08-10
feat(rust/rolldown): support `inject` imports

Boshen:
This PR adds the missing pieces for inject plugin.

### DIFF
--- a/crates/rolldown/src/utils/pre_process_ecma_ast.rs
+++ b/crates/rolldown/src/utils/pre_process_ecma_ast.rs
@@ -2,12 +2,14 @@ use std::path::Path;
 
 use oxc::ast::VisitMut;
 use oxc::minifier::{
-  CompressOptions, Compressor, ReplaceGlobalDefines, ReplaceGlobalDefinesConfig,
+  CompressOptions, Compressor, InjectGlobalVariables, InjectGlobalVariablesConfig, InjectImport,
+  ReplaceGlobalDefines, ReplaceGlobalDefinesConfig,
 };
-use oxc::semantic::{ScopeTree, SymbolTable};
+use oxc::semantic::{ScopeTree, SemanticBuilder, SymbolTable};
 use oxc::span::SourceType;
 use oxc::transformer::{TransformOptions, Transformer};
-use rolldown_ecmascript::EcmaAst;
+
+use rolldown_ecmascript::{EcmaAst, WithMutFields};
 
 use crate::types::oxc_parse_type::OxcParseType;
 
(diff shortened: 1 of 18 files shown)


crates/rolldown/src/utils/pre_process_ecma_ast.rs
```diff
@@ -23,6 +25,19 @@ pub fn pre_process_ecma_ast(
...
 ) -> anyhow::Result<(EcmaAst, SymbolTable, ScopeTree)> {
+  // Build initial semantic data and check for semantic errors.
+  let semantic_ret = ast.program.with_mut(|WithMutFields { program, source, .. }| {
+    SemanticBuilder::new(source, source_type).build(program)
+  });
+
+  // TODO:
+  // if !semantic_ret.errors.is_empty() {
+  // return Err(anyhow::anyhow!("Semantic Error: {:#?}", semantic_ret.errors));
+  // }
```
Boshen, line 38: This broke a few test cases so I left this as a TODO.

crates/rolldown/src/utils/pre_process_ecma_ast.rs
```diff
@@ -43,21 +58,37 @@ pub fn pre_process_ecma_ast(
...
+    // Use built-in inject plugin.
+    // <https://www.npmjs.com/package/@rollup/plugin-inject>
+    // TODO: move config to Rolldown options
+    // NOTE: `InjectGlobalVariablesConfig` is `Arc`ed.
+    // let config = InjectGlobalVariablesConfig::new(vec![InjectImport::named_specifier(
+    // "es6-promise",
+    // [> "default" <] None,
+    // "Promise",
+    // )]);
+    // InjectGlobalVariables::new(allocator, config).build(&mut symbols, &mut scopes, program);
```
Boshen: Please uncomment this block to get things going.

hyf0:
Any idea about why performance get down so much even we don't use inject? 

---edited

Besides adding skipping logic, even we do a full-ast traverse with empty options, it shouldn't slow that much. [This PR](https://github.com/rolldown/rolldown/pull/1927) adds a almost full-ast traverse pass, but have little affection.

crates/rolldown/tests/rolldown/function/inject/main.js
```diff
@@ -0,0 +1,8 @@
+import assert from 'node:assert'
+
+assert.strictEqual(Promise, 'promise-shim')
+assert.strictEqual(P, 'promise-shim')
+assert.strictEqual($, 'jquery')
+assert.strictEqual(fs.default, 'node-fs')
+// FIXME: oxc injects invalid statements `import { default as 'Object.assign' from 'object-assign-shim'` }`, so it fails.
```
hyf0, line 7: @Boshen cc

There is a bug here.

I take a rough look

oxc transform to the following code

```js
import { default as 'Object.assign'} from 'object-assign-shim'`
```

This is invalid js code.

while it should be 

```js
import { default as Object_assign} from 'object-assign-shim'`
```
or
```js
import Object_assign from 'object-assign-shim'`
```

hyf0 (approved):

Boshen:
> Any idea about why performance get down so much even we don't use inject?
> 
> ---edited
> 
> Besides adding skipping logic, even we do a full-ast traverse with empty options, it shouldn't slow that much. [This PR](https://github.com/rolldown/rolldown/pull/1927) adds a almost full-ast traverse pass, but have little affection.

This is because semantic is now run for each file in this file, where previously its only run for ts or jsx files.

crates/rolldown/tests/rolldown/function/inject/main.js
Boshen: https://github.com/oxc-project/oxc/issues/4818

PR #1934, opened 2024-08-11
fix(rolldown): fix inject test where `alias` can be a member expression

Boshen:
### Description

In inject plugin https://www.npmjs.com/package/@rollup/plugin-inject, `imported` is optional and is defaulted to `default` https://github.com/rollup/plugins/blob/8550c4b1925b246adbd3af48ed0e5f74f822c951/packages/inject/src/index.js#L125

Only `alias` can be a member expression.


hyf0 (approved):
Thanks.

PR #1938, opened 2024-08-11
feat: allow passing array of array in `config.plugins`

IWANABETHATGUY:
### Description
1. This could allow the user or developer to compose two or more different types plugin into one, this is easy for distributing.
2. See test as an example.
3. Progressive porting js plugin.
4. Prerequisite of porting `vite:build-import-analysis`

### DIFF
--- a/packages/rolldown/src/utils/normalize-input-options.ts
+++ b/packages/rolldown/src/utils/normalize-input-options.ts
@@ -12,6 +12,7 @@ export async function normalizeInputOptions(
 ): Promise<NormalizedInputOptions> {
   const { input, ...rest } = config
   let plugins = await normalizePluginOption(config.plugins)
+  plugins = plugins.flat()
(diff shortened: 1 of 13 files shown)


packages/rolldown/src/utils/normalize-input-options.ts
```diff
@@ -11,7 +11,7 @@ export async function normalizeInputOptions(
   config: InputOptions,
 ): Promise<NormalizedInputOptions> {
   const { input, ...rest } = config
-  let plugins = await normalizePluginOption(config.plugins)
+  let plugins = (await normalizePluginOption(config.plugins)).flat()
```
hyf0: I thought we have this before.

https://github.com/rolldown/rolldown/blob/38a2c3e6a9d782efdea45a71d15cae589647ea37/packages/rolldown/src/utils/normalize-plugin-option.ts

hyf0 (approved):

PR #1949, opened 2024-08-11
fix(treeshaking): `require` should consider as having side effects

### LINKED ISSUE #1917: [Bug]: cjs dependency ends up empty when `treeshake.moduleSideEffects: false` is set

To reproduce:

* clone `vuejs/core` 's latest `rolldown-main` branch
* Run `pnpm build compiler-sfc -f cjs`
* Check `packages/compiler-sfc/dist/compiler-sfc.cjs.js` - line 3604

```js
//#region node_modules/.pnpm/util-deprecate@1.0.2/node_modules/util-deprecate/node.js
var require_node = __commonJSMin((exports, module) => {}); // line 3604
// ...
var deprecate = require_node(); // line 3647
```

* Somehow the dep results in an empty module, but the export is called on line 3649, so that breaks everything

The offending dependency is `postcss-selector-parser` -> `util-deprecate`, however I tried bundling this dep alone and cannot reproduce this issue. It only happens when bundling Vue.

### What is expected?

Module not empty

Module is empty

### System Info

```Shell
rolldown 0.12.2-snapshot-29d29ba-20240809002935
```

### Any additional comments?

_No response_

hyf0:
### Description

`require` should consider having side-effect no matter what in current codebase.

---

```
require('no-side-effect'); // Shouldn't consider having side-effect but we haven't handled such cases.
```

### DIFF
--- a/crates/rolldown/src/stages/link_stage/mod.rs
+++ b/crates/rolldown/src/stages/link_stage/mod.rs
@@ -271,7 +271,7 @@ impl<'a> LinkStage<'a> {
             match &self.module_table.modules[rec.resolved_module] {
               Module::External(importee) => {
                 // Make sure symbols from external modules are included and de_conflicted
-                let is_reexport_all = match rec.kind {
+                match rec.kind {
                   ImportKind::Import => {
                     if matches!(self.options.format, OutputFormat::Cjs)
                       && !rec.meta.contains(ImportRecordMeta::IS_PLAIN_IMPORT)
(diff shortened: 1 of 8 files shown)


hyf0:
* **#1950** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1950?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1949** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/1949?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #1957, opened 2024-08-12
feat: add wasm fallback plugin

### LINKED ISSUE #1919: wasmFallbackPlugin
https://github.com/rolldown/vite/blob/ededbc26533fc3ff7069e00ab8990ead9fd3cf64/packages/vite/src/node/plugins/wasm.ts#L79-L96

ikkz:
resolves #1919

### DIFF
--- a/crates/rolldown_plugin_wasm_fallback/src/lib.rs
+++ b/crates/rolldown_plugin_wasm_fallback/src/lib.rs
@@ -1,0 +1,27 @@
+use std::borrow::Cow;
+
+use rolldown_plugin::{HookLoadArgs, HookLoadReturn, Plugin, SharedPluginContext};
+
+#[derive(Debug)]
+pub struct WasmFallbackPlugin {}
+
+impl Plugin for WasmFallbackPlugin {
+  fn name(&self) -> Cow<'static, str> {
+    Cow::Borrowed("builtin:wasm-fallback-plugin")
+  }
+
+  #[allow(clippy::case_sensitive_file_extension_comparisons)]
+  async fn load(&self, _ctx: &SharedPluginContext, args: &HookLoadArgs<'_>) -> HookLoadReturn {
+    if args.id.ends_with(".wasm") {
+      // TODO: Replace the link here after rolldown's document is ready
+      Err(anyhow::anyhow!(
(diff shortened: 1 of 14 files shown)


crates/rolldown_plugin_wasm_fallback/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+
+impl Plugin for WasmFallbackPlugin {
+  fn name(&self) -> Cow<'static, str> {
+    Cow::Borrowed("builtin:wasm-fallback-plugin")
+  }
+
+  async fn load(&self, _ctx: &SharedPluginContext, args: &HookLoadArgs<'_>) -> HookLoadReturn {
+    if std::path::Path::new(args.id)
+      .extension()
+      .map_or(false, |ext| ext.eq_ignore_ascii_case("wasm"))
```
IWANABETHATGUY: Use `args.id.ends_with` instead, construct a `Path` is slow

packages/rolldown/tests/fixture.test.ts
```diff
@@ -19,11 +19,20 @@ function main() {
...
+            nodePath.join(import.meta.dirname, dirPath),
+            testConfig,
+          )
+        } catch (err) {
+          if (testConfig.catchError) {
+            await testConfig.catchError(err)
+          } else {
+            throw err
+          }
+        }
```
IWANABETHATGUY:
```suggestion
          const output = await compileFixture(
            nodePath.join(import.meta.dirname, dirPath),
            testConfig,
          ).catch((err) => {
            if (testConfig.catchError) {
              testConfig.catchError(err)
              return
            }
            throw err
          })

```

packages/rolldown/tests/fixture.test.ts
```diff
@@ -22,8 +22,14 @@ function main() {
...
           testConfig,
-        )
-        if (testConfig.afterTest) {
+        ).catch(async (err) => {
+          if (testConfig.catchError) {
+            await testConfig.catchError(err)
+            return
+          }
+          throw err
+        })
```
ikkz, line 31: An `async` function is passed to catch because it allows to throw new errors in TestConfig#catchError

IWANABETHATGUY (approved):
Thanks

PR #1961, opened 2024-08-12
chore: move cargo shear to lint

ikkz:
This PR moves `cargo shear` into `lint-rust`, so that `just roll` can also execute this command.


hyf0 (approved):
Thanks

PR #1962, opened 2024-08-12
chore: improve the successful output of setup in `justfile`

shulaoda:
### Description
When studying `just`, I found that `echo` can cause duplicate output, so we should change it to `@echo`. This issue is too insignificant, I hope this PR is useful.

image


hyf0 (approved):
Thanks

Issue #1954, opened 2024-08-12
replace `citty` with builtin `node:parseArgs` to parse command line args.


IWANABETHATGUY:
### What problem does this feature solve?

`citty` is not actively maintained, and since our lowerest node js requirement is `>= 18.20.3` we could use `node:parseArgs` to parse command line args.

related to https://github.com/rolldown/rolldown/issues/1852
### What does the proposed API look like?

same as before.

ikkz:
I want to try it

IWANABETHATGUY:
Thanks in advance, but It would great to finish https://github.com/rolldown/rolldown/issues/1919 this ASAP, since #1919 have higher priority, if you don't have enough time, then I could take #1919, and you continue this issue.

ikkz:
> Thanks in advance, but It would great to finish #1919 this ASAP, since #1919 have higher priority, if you don't have enough time, then I could take #1919, and you continue this issue.

I have written a draft for #1919 and plan to open a PR today.

IWANABETHATGUY:
> > Thanks in advance, but It would great to finish #1919 this ASAP, since #1919 have higher priority, if you don't have enough time, then I could take #1919, and you continue this issue.
> 
> I have written a draft for #1919 and plan to open a PR today.

that's great

7086cmd:
Hello @ikkz. To complete the task, you just need to remove the `citty` parameter and modify the rolldown on the track using parseArgs. I will take care of the -c and -h parameters more efficiently with other arguments :)

ikkz:
> Hello @ikkz. To complete the task, you just need to remove the `citty` parameter and modify the rolldown on the track using parseArgs. I will take care of the -c and -h parameters more efficiently with other arguments :)

@7086cmd Thank you for your guidance. I'll open a draft PR tomorrow

PR #1963, opened 2024-08-12
feat(node): support CLI options.

### LINKED ISSUE #1955: [rolldown/cli] Auto-Generate CLI Args from Options Schema

Looking to enhance the CLI by auto-generating command line arguments based on the options schema.
## Desc 
- Automatically generate CLI args from schema options (booleans, strings, arrays of strings).This would make the CLI more flexible, reducing the need for a config file and aligning args directly with schema types.

n/a

7086cmd:
resolves #1955.

This public relations campaign leverages the `parseArgs` function provided by the `node:util` module and the JSON schema generated by the `zod` library to automatically generate CLI options. This approach enhances the elegance and convenience of the process.

In my opinion, the runtime-generated schema is performance-degrading. While it is feasible to generate it during compilation via a plugin, it is currently too complex for this particular pull request, and also requires support `TypeScript` for config file.


packages/rolldown/src/cli/index.ts
```diff
@@ -1,54 +1,33 @@
...
+    const cliOptions = parseCliArguments()
+
+    if (cliOptions.config) {
+      await bundleWithConfig(cliOptions.config, cliOptions)
+      process.exit(0)
+    }
+
+    if ('input' in cliOptions.input) {
+      // If input is specified, we will bundle with the input options
+      await bundleWithCliOptions(cliOptions)
```
7086cmd: Here, because the input argument is passed by the positional, so I think this might be the best way to identify if user input the input file in command line.

The command line will override the configure file if there's conflicts, in rollup.

packages/rolldown/src/options/input-options.ts
```diff
@@ -98,6 +98,18 @@ const inputOptionsSchema = z.strictObject({
...
+  external: z.array(z.string()).optional(),
+  onLog: z.string().optional(), // command
+  onwarn: z.string().optional(),
+  inject: z.record(z.string()).optional(),
+  // Doesn't support treeshake options as it is ambiguous
+  treeshake: z.undefined().optional(),
+  // Doesn't support plugin yet.
+  plugins: z.undefined().optional(),
+})
+
```
7086cmd: In command lines, options don't support functions, so I override these function-accepted options and just allow accepted types.

packages/rolldown/src/options/output-options.ts
```diff
@@ -50,6 +50,22 @@ const outputOptionsSchema = z.strictObject({
...
+  banner: z.string().optional(),
+  footer: z.string().optional(),
+  intro: z.string().optional(),
+  outro: z.string().optional(),
+  // It is hard to handle the union type in json schema, so use this first.
+  esModule: z
+    .literal('if-default-prop')
+    .or(z.literal('always'))
+    .or(z.literal('never'))
+    .optional(),
```
7086cmd: Here, because the type of 'boolean | string' in command line is ambiguous, so I used this first, and then normalize it.

IWANABETHATGUY:
review this after #1954 is solved.

IWANABETHATGUY (approved):

IWANABETHATGUY:
Overall looks good, and could be polished in the future.

PR #1964, opened 2024-08-12
refactor(rust): improve esModule type in binding

shulaoda:
### Description
Related to #1888 

Due to some mistakes, I mistakenly believed that it was not possible to use `napi-rs`'s `Either` for value passing in Rolldown. You can refer to this issue https://github.com/napi-rs/napi-rs/issues/2215 for more details.

### DIFF
--- a/crates/rolldown_binding/src/utils/normalize_binding_options.rs
+++ b/crates/rolldown_binding/src/utils/normalize_binding_options.rs
@@ -121,8 +121,8 @@ pub fn normalize_binding_options(
     dir: output_options.dir,
     sourcemap: output_options.sourcemap.map(Into::into),
     es_module: output_options.es_module.map(|es_module| match es_module {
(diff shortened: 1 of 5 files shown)


crates/rolldown_binding/src/utils/normalize_binding_options.rs
```diff
@@ -120,7 +120,10 @@ pub fn normalize_binding_options(
     asset_filenames: output_options.asset_file_names,
     dir: output_options.dir,
     sourcemap: output_options.sourcemap.map(Into::into),
-    es_module: output_options.es_module.map(Into::into),
+    es_module: output_options.es_module.map(|es_module| match es_module {
+      Either::A(a) => a.into(),
```
IWANABETHATGUY: can you use a better naming instead `a` and `b`?

hyf0 (approved):

PR #1965, opened 2024-08-12
feat: avoid OutputAsset#name clone asset

shulaoda:
### Description
Related to #1586 

https://github.com/rolldown/rolldown/blob/73808a9dfce628a7083b1aad6a80b63ad0b013ed/crates/rolldown_common/src/file_emitter.rs#L86-L91

### DIFF
--- a/crates/rolldown_binding/src/types/binding_outputs.rs
+++ b/crates/rolldown_binding/src/types/binding_outputs.rs
@@ -40,7 +40,10 @@ impl BindingOutputs {
       let mut inner = inner.lock().expect("PoisonError raised");
       inner.iter_mut().for_each(|o| match o {
         rolldown_common::Output::Asset(asset) => {
-          assets.push(BindingOutputAsset::new(unsafe { std::mem::transmute(asset.as_mut()) }));
+          assets.push(BindingOutputAsset::new(unsafe {
+            &mut *(std::ptr::from_mut::<std::sync::Arc<rolldown_common::EmittedAsset>>(asset)
+              .cast::<rolldown_common::EmittedAsset>())
+          }));
         }
         rolldown_common::Output::Chunk(_) => {}
       });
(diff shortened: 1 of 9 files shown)


shulaoda:
@underfin cc

IWANABETHATGUY:
/Users/runner/work/_temp/3c2fef40-cddd-4593-8d5d-7a047a72e37e.sh: line 1:  4534 Segmentation fault: 11  pnpm run --recursive --parallel --filter=!rollup-tests test

crates/rolldown_binding/src/types/binding_outputs.rs
```diff
@@ -96,7 +99,10 @@ impl FinalBindingOutputs {
       let mut inner = inner.lock().expect("PoisonError raised");
       inner.iter_mut().for_each(|o| match o {
         rolldown_common::Output::Asset(asset) => {
-          assets.push(BindingOutputAsset::new(unsafe { std::mem::transmute(asset.as_mut()) }));
+          assets.push(BindingOutputAsset::new(unsafe {
+            &mut *(std::ptr::from_mut::<std::sync::Arc<rolldown_common::EmittedAsset>>(asset)
+              .cast::<rolldown_common::EmittedAsset>())
+          }));
```
IWANABETHATGUY: I don't thnk this is a safe cast.

shulaoda: Yes, I am currently investigating this problem. I'm not sure if we should really avoid cloning, it seems to affect the `files` in `FileEmitter`.

IWANABETHATGUY: According to the benchmark `asset` clone is not the bottleneck, clone it is fine.

shulaoda:
When `emitted_files` exist, it seems that the assets in `FileEmitter`'s `files` are no longer in use. Perhaps we can delete it directly.

shulaoda:
It seems necessary, I will close this PR.

PR #1971, opened 2024-08-13
refactor: migrate from citty to node's built-in `parseArgs`

### LINKED ISSUE #1852: [Bug]: Should not emit twice for the same diagnostic

see https://github.com/rolldown/rolldown/issues/1851 as an example.

### What is expected?

Should only emit once.

emit twice.

### System Info

```Shell
n/a
```

### Any additional comments?

_No response_

### LINKED ISSUE #1954: replace `citty` with builtin `node:parseArgs` to parse command line args.

ikkz:
resolves #1954 
resolves #1852


ikkz:
Usage output and the current error output of #1852

image

packages/rolldown/src/cli/index.ts
```diff
@@ -1,55 +1,34 @@
...
-  // help?: boolean
-  // h?: boolean
-}
+async function main() {
+  const { values } = parseArgs({
+    options: mapValues(CLI_OPTIONS, omit(['description'])),
+    // We need to support both `rolldown -c` and `rolldown -c rolldown.config.js`,
+    // the value of the option could be either a boolean or a string in this case,
+    // so `strict` needs to be set to `false`
+    strict: false,
```
7086cmd, line 16: Indeed, it should be `false`. By the way, I wonder if we should specify the `rolldown.config.js` file or move it to the constants file (as it may be considered hardcoding if I remove the “constants” file.).

Additionally, I do not believe the automatically ignored parameter is elegant. Once the API becomes stable and independent, I think we should look into a better way to separate the default configuration or the configuration file that’s been specified :thinking:.

packages/rolldown/src/cli/index.ts
```diff
@@ -1,55 +1,34 @@
...
+import { parseArgs } from 'node:util'
+import { mapValues, omit } from 'remeda'
 import { bundle } from './commands/bundle'
-import {
-  version,
-  description,
-} from '../../package.json' assert { type: 'json' }
+import { showHelp } from './commands/help'
 import { DEFAULT_CONFIG_FILENAME } from './constants'
+import { CLI_OPTIONS } from './options'
```
7086cmd, line 7: The value should not be static when the automatically generated arguments are implemented, so I suggest using the `camelCase` method instead of the UPPER SNAKE_CASE method.

7086cmd (approved):
Great, and I will take care of matters pertaining to arguments in #1963. cc @IWANABETHATGUY.

packages/rolldown/src/cli/index.ts
ikkz: In this PR, the value is still static; you can change it to camelCase in your subsequent PR.

7086cmd: Oh, I see.

packages/rolldown/src/cli/index.ts
ikkz: > Indeed, it should be `false`. By the way, I wonder if we should specify the `rolldown.config.js` file or move it to the constants file (as it may be considered hardcoding if I remove the “constants” file.).

It should be fine to keep it in constants.ts, as there might be some default values and so on for the CLI in the future.

> Additionally, I do not believe the automatically ignored parameter is elegant. Once the API becomes stable and independent, I think we should look into a better way to separate the default configuration or the configuration file that’s been specified 🤔.

Perhaps when the user specifies both the config option and other options, we could log a message to inform them that we will ignore the other options?

7086cmd: > It should be fine to keep it in constants.ts, as there might be some default values and so on for the CLI in the future.

Ok.

> Perhaps when the user specifies both the config option and other options, we could log a message to inform them that we will ignore the other options?

I'll do it in #1963.

packages/rolldown/src/cli/commands/help.ts
```diff
@@ -0,0 +1,39 @@
...
+import {
+  version,
+  description,
+} from '../../../package.json' assert { type: 'json' }
+import { CLI_OPTIONS } from '../options'
+import { bold, cyan, underline } from '../colors'
+
+const HELP_TEMPLATE = `rolldown version ${version}
+
+${description}
```
IWANABETHATGUY: > Fast JavaScript/TypeScript bundler in Rust with Rollup-compatible API. (rolldown v0.12.2)                                                                                        

Can you keep the layout as before?

image

packages/rolldown/src/cli/commands/help.ts
```diff
@@ -0,0 +1,39 @@
...
+          if (short) {
+            optionStr += `-${short}, `
+          }
+          optionStr += `--${option} `
+          if (hint) {
+            optionStr += `<${hint}> `
+          }
+          return cyan(optionStr.padEnd(30)) + description
+        })
+        .join('\n'),
```
IWANABETHATGUY, line 33: Can you add an indentation for each `option` ?

IWANABETHATGUY: image

Also we should fix the wrong indentation before

packages/rolldown/src/cli/commands/help.ts
ikkz: current output and before:
image

packages/rolldown/src/cli/commands/help.ts
ikkz: Done, two spaces are added bofore each option

IWANABETHATGUY (approved):

PR #1977, opened 2024-08-14
feat: align cli options with unix ux patterns.

7086cmd:
resolves #1924.

```
Fast JavaScript/TypeScript bundler in Rust with Rollup-compatible API. (rolldown v0.12.2)                                                                       

USAGE rolldown -c <config> or rolldown <input> <options>
                                                                                                                                                                
OPTIONS                                                                                                                                                         
                                                                                                                                                                
  --config -c, <filename>     Path to the config file (default: rollup.config.js).                                                                              
  --dir -d, <dir>             Output directory, defaults to dist.
  --external -e, <external>   Module ids to exclude from the bundle. (multiple)
  --format -f, <format>       Output format of the generated bundle (supports esm, cjs, and iife).
  --globals -g, <globals>     Map of global variable names to their module name in the bundle in IIFE / UMD format. (multiple)
  --help -h,                  Show help.
  --minify -m,                Minify the bundled file.
  --name -n, <name>           Name for UMD / IIFE format wrapper functions.
  --platform -p, <platform>   Platform for which the code should be generated (node, browser, neutral).
  --sourcemap -s, <sourcemap> Generate sourcemap (-s inline for inline, or pass the -s on the last argument if you want to generate .map file).
  --version -v,               Show version number.
  --asset-file-names <name>   .
  --banner <banner>           Code to insert the top of the bundled file (outside the wrapper function).
  --chunk-file-names <name>   .
  --cwd <cwd>                 Current working directory.
  --define <define>           Define global variables. (multiple)
  --entry-file-names <name>   .
  --es-module                 Always generate __esModule marks in non-ESM formats, defaults to if-default-prop (use --no-esModule to always disable).
  --exports <exports>         Specify a export mode (auto, named, default, none).
  --extend                    Extend global variable defined by name in IIFE or UMD formats.
  --footer <footer>           Code to insert the bottom of the bundled file (outside the wrapper function).
  --inject <inject>           Inject import statements on demand. (multiple)
  --intro <intro>             Code to insert the top of the bundled file (inside the wrapper function).
  --log-level <log-level>     Log level (silent, info, debug, warn).
  --module-types <types>      Module types for customized extensions. (multiple)
  --no-external-live-bindings Use external live bindings.
  --no-treeshake              Enable treeshaking.
  --outro <outro>             Code to insert the bottom of the bundled file (inside the wrapper function).
  --shim-missing-exports      .
  --sourcemap-ignore-list     .
                                                                                                                                                                
EXAMPLES                                                                                                                                                        
                                                                                                                                                                
  1. Bundle with a config file rolldown.config.mjs:                                                                                                             
    rolldown -c rolldown.config.mjs                                                                                                                             
                                                                                                                                                                
  2. Bundle the src/main.ts to dist with cjs format:                                                                                                            
    rolldown src/main.ts -d dist -f cjs                                                                                                                         
                                                                                                                                                                
  3. Bundle the src/main.ts and handle the .png assets to Data URL:                                                                                             
    rolldown src/main.ts -d dist --moduleTypes .png=dataurl                                                                                                     
                                                                                                                                                                
  4. Bundle the src/main.tsx and minify the output with sourcemap:                                                                                              
    rolldown src/main.tsx -d dist -m -s                                                                                                                         
                                                                                                                                                                
  5. Create self-executing IIFE using external jQuery as $ and _:                                                                                               
    rolldown src/main.ts -d dist -n bundle -f iife -e jQuery,window._ -g jQuery=$                                                                               
                                                                                                                                                                
NOTES                                                                                                                                                           
                                                                                                                                                                
  * Due to the API limitation, you need to pass -s for .map sourcemap file as the last argument.                                                                
  * If you are using the configuration, please pass the -c as the last argument if you ignore the default configuration file.                                   
  * CLI options will override the configuration file.                                                                                                           
  * For more information, please visit https://rolldown.rs/.                                                                                              
```


IWANABETHATGUY:
Can you add some tests?

7086cmd:
> Can you add some tests?

The sole objective of this PR is to enhance the help page and rectify the `-v` (formerly displays the help) option. I am uncertain about the appropriate channel for addressing other tests :joy:. And [this](https://github.com/rolldown/rolldown/pull/1977#issue-2464954454) is the help message in this PR.

7086cmd:
> Can you add some tests?

Completed.

7086cmd:
The issue appears to be related to the usage of the `nodejs 22.6.0`-featured API, specifically the `allowNegative` option. This option is crucial, and I am contemplating whether we should modify the `node-version`.

packages/rolldown/src/cli/arguments/index.ts
```diff
@@ -43,14 +48,23 @@ export function parseCliArguments() {
...
     allowPositionals: true,
-    allowNegative: true,
     // We can't use `strict` mode because we should handle the default config file name.
     strict: false,
   })
 
   tokens
     .filter((token) => token.kind === 'option')
     .forEach((option) => {
+      if (option.name.startsWith('no-')) {
```
hyf0, line 65: I'm not sure about this `no-xxx` patterns. We don't need to follow what rollup did in cli.

I'll suggest we made decisions based on implementing a cli with the best UX, not just simply rollup did it.

hyf0: - `rolldown --treeshake`
- `rolldown --treeshake=false` | `rolldown --no-treeshake`
- `rolldown --treeshake=false --external-live-bindings=false`  | `rolldown --no-treeshake --no-external-live-bindings`

hyf0: There's usage that using rolldown-cli programmatically. `no-xxx` pattern is not friendly for rendering cli args programmatically.

```js
const command = `rolldown --treeshake=${treeshake}`
```

```js
const command = `rolldown --${ !treeshake ? 'no-'  : ''}treeshake`
```

7086cmd: I have implemented the no- name prefix to handle the boolean value whose default values are true. The only problem is that a union type with string and boolean is hard to handle.

7086cmd: I think we can follow up Unix's UX Pattern since it’s easier to support and more popular to use.

For arguments,

1. Offer both short and long options where applicable. For example, `-o` and `--output`.
2. Use the presence of an option to toggle a feature. For disabling, use `--no-` prefixes for long options.
3. Allow flexibility in how users pass values. Support both space-separated (for short option) and equals-sign (for long option) syntax.
4. Clearly define mandatory and optional positional arguments, and ensure options can be used flexibly with them.
5. Support stacked short options for a concise CLI experience.
6. Always provide a clear and consistent way to negate options where applicable.
7. Always include `--help` and `--version` as standard options.
8. Support environment variables for configurations, with CLI options taking precedence.
9. Use defaults where possible but ensure clear communication when required arguments are omitted.
10. Support configuration files for flexibility but ensure CLI arguments can override any configuration.

*(Generated by ChatGPT)*

7086cmd: > I'm not sure about this no-xxx patterns. We don't need to follow what rollup did in cli.

Incidentally, the code responsible for handling arguments prefixed with “no-“ is referenced from the official website of Node.js: [node.js official website](https://nodejs.org/api/util.html#parseargs-tokens).

hyf0 (approved):
Thanks!

PR #1981, opened 2024-08-14
fix: ensure ordering of assign chunk name

underfin:
### Description

 The esbuild using `Chunk#bits` to sorted chunks, but the order of `Chunk#bits` is not stable, eg `BitSet(0) 00000001_00000000` > `BitSet(8) 00000000_00000001`. It couldn't ensure the order of dynamic chunks and common chunks.
 Consider the compare `Chunk#exec_order` should be faster than `Chunk#bits`, we use `Chunk#exec_order` to sort chunks.
 Note Here could be make sure the order of chunks.
 - entry chunks are always before other chunks
 - static chunks are always before dynamic chunks
 - other chunks has stable order at per entry chunk level


crates/rolldown/src/stages/generate_stage/code_splitting.rs
```diff
@@ -215,15 +211,26 @@ impl<'a> GenerateStage<'a> {
...
 
-    let sorted_chunk_idx_vec =
-      chunks.indices().sorted_by_key(|id| &chunks[*id].bits).collect::<Vec<_>>();
+    // The esbuild using `Chunk#bits` to sorted chunks, but the order of `Chunk#bits` is not stable, eg `BitSet(0) 00000001_00000000` > `BitSet(8) 00000000_00000001`. It couldn't ensure the order of dynamic chunks and common chunks.
+    // Consider the compare `Chunk#exec_order` should be faster than `Chunk#bits`, we use `Chunk#exec_order` to sort chunks.
+    // Note Here could be make sure the order of chunks.
+    // - entry chunks are always before other chunks
+    // - static chunks are always before dynamic chunks
+    // - other chunks has stable order at per entry chunk level
+    let sorted_chunk_idx_vec = chunks
```
hyf0, line 220: https://github.com/rolldown/rolldown/blob/cc41c949e4003d6f7ae6a8b0f205111298a5a7c5/crates/rolldown/src/stages/generate_stage/code_splitting.rs#L169-L216

chunk is already sorted here. Could we just use `chunks.indices().collect<Vec<_>>`?

crates/rolldown/tests/rolldown/misc/assign-chunk-name-order/1/file.js
hyf0, line 1: Folder name should have snake cases.

crates/rolldown/src/stages/generate_stage/code_splitting.rs
underfin: The share chunks ordering is not stable because we using unstale ordering at https://github.com/rolldown/rolldown/blob/main/crates/rolldown/src/stages/generate_stage/code_splitting.rs#L141

underfin: If we fix it look like it could be get expected stable ordering for assgin name

hyf0: > The share chunks ordering is not stable because we using unstale ordering at https://github.com/rolldown/rolldown/blob/main/crates/rolldown/src/stages/generate_stage/code_splitting.rs#L141

Shared chunks are sorted by their first module's execution order. It seems stable.

https://github.com/rolldown/rolldown/blob/cc41c949e4003d6f7ae6a8b0f205111298a5a7c5/crates/rolldown/src/stages/generate_stage/code_splitting.rs#L204-L209

https://github.com/rolldown/rolldown/blob/cc41c949e4003d6f7ae6a8b0f205111298a5a7c5/crates/rolldown/src/stages/generate_stage/code_splitting.rs#L163-L167

underfin: > https://github.com/rolldown/rolldown/blob/cc41c949e4003d6f7ae6a8b0f205111298a5a7c5/crates/rolldown/src/stages/generate_stage/code_splitting.rs#L169-L216
> 
> chunk is already sorted here. Could we just use `chunks.indices().collect<Vec<_>>`?

Oh. I missing the your meanings. Here the order is not correct, because the share chunk execute order could be at first. But the entry chunks should be at first for assign name.

hyf0: https://github.com/rolldown/rolldown/blob/cc41c949e4003d6f7ae6a8b0f205111298a5a7c5/crates/rolldown/src/stages/generate_stage/code_splitting.rs#L182-L191

Probably should this logic.

hyf0 (approved):

crates/rolldown/src/stages/generate_stage/code_splitting.rs
underfin: > If we fix it look like it could be get expected stable ordering for assgin name

Here I have a try to changed it, but looke like the `sorted_modules` hasn't dynamic import ids.

PR #1861, opened 2024-08-04
feat: support namespaced name and `output.extend` for IIFE format.

7086cmd:
This pull request supports the use of namespaced names (including the dot character) and the `output.extend` option for IIFE format aligning with Rollup. Tests are included, and further refactoring is planned (not included in this PR, as it is better to conduct refactoring when more formats are supported, in my opinion).

PR #1986, opened 2024-08-15
perf: improve generate_namespace_definition

shulaoda:
### Description
Related to #1861 
https://github.com/rolldown/rolldown/blob/745c1a57e66433df7182b5f9f4aa65aa00aa001b/crates/rolldown/src/ecmascript/format/utils/namespace.rs#L49-L51

### DIFF
--- a/crates/rolldown/src/ecmascript/format/utils/namespace.rs
+++ b/crates/rolldown/src/ecmascript/format/utils/namespace.rs
@@ -28,29 +28,20 @@ use rolldown_utils::ecma_script::is_validate_assignee_identifier_name;
 ///    ```
 fn generate_namespace_definition(name: &str) -> (String, String) {
   let parts: Vec<&str> = name.split('.').collect();
+  let mut initial_code = String::new();
+  let mut final_code = String::from("this");
 
-  let initialization_code = parts
-    .iter()
-    .enumerate()
-    .scan(String::new(), |state, (i, part)| {
-      // We use `scan` to generate the declaration sentence level-by-level.
-      let caller = generate_caller(part);
-      state.push_str(&caller);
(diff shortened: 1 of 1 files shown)


shulaoda:
@7086cmd cc

crates/rolldown/src/ecmascript/format/utils/namespace.rs
```diff
@@ -28,29 +28,20 @@ use rolldown_utils::ecma_script::is_validate_assignee_identifier_name;
...
-    .collect::<String>();
+  for (i, part) in parts.iter().enumerate() {
+    let caller = generate_caller(part);
+    final_code.push_str(&caller);
 
-  // TODO do not call the `generate_caller` function twice.
-  let final_code =
-    format!("this{}", parts.iter().map(|&part| generate_caller(part)).collect::<String>());
+    if i < parts.len() - 1 {
+      let callers = &final_code[4..];
```
7086cmd: Could you please clarify if this is related to the `this` keyword? It would be preferable to avoid hardcoding this, as we need to support different namespace roots.

shulaoda: Yes, because I pre-set `this` in `final_code`. Do I need to do anything else?

7086cmd: It is advisable to modify the code to check the length of the `this` variable instead of hardcoding (`&final_code[4..];`). We may support the UMD format, which uses `global` instead of `this`. Also [here](https://rollupjs.org/configuration-options/#context).

7086cmd (approved):
Nice.

IWANABETHATGUY (approved):

PR #1991, opened 2024-08-15
fix: typo of `transform-plugin-context.ts`

IWANABETHATGUY:


IWANABETHATGUY:
Just a typo fix, so skip reviewing.

PR #1995, opened 2024-08-16
test: test infra for cli

### LINKED ISSUE #1994: [Feature Request]: E2E Test the CLI gracefully.

Previously, the end-to-end test for the command line interface (CLI) was exclusively utilized for verifying the execution of commands, but not for assessing the behavior of the application. It appears that the `stdout` output does not contain the actual `info` message.

In the end-to-end test, ensure the command-line interface (CLI) behaves gracefully. Related #1993.

IWANABETHATGUY:
### Description
1. `consola` intercept all stdout, since the instance is created in another process,  the mock technique is not useful https://github.com/unjs/consola?tab=readme-ov-file#mocktypes
2. create a testLogger for snapshot testing.
3. Closed https://github.com/rolldown/rolldown/issues/1994


7086cmd (approved):
Excellent news! We will never only receive the `help` message when using the `rolldown -c` command :)

hyf0 (approved):

packages/rolldown/src/cli/utils.ts
```diff
@@ -6,16 +6,44 @@ import type { ConfigExport } from '../types/config-export'
 /**
  * Console logger
  */
-export const logger = createConsola({
-  formatOptions: {
-    date: false,
-  },
-})
+export const logger = process.env.TEST
```
ikkz, line 9: Perhaps a more unique env identifier such as `ROLLDOWN_TEST` would be better since `TEST` is too common in development and may conflict with users' env

IWANABETHATGUY: nice advice.

PR #2006, opened 2024-08-17
ci: exit with none zero code when snapshots updated (only in github actions CI env).

### LINKED ISSUE #2005: [Bug]: should exit none 0 when the snapshots updated.

1. try to update some code in core, that will affect codegen.
2. run `cargo test` 
3. echo the exit code

### What is expected?

should exit none 0 when the snapshots updated.

exit  0 when the snapshots updated.

### System Info

```Shell
n/a
```

### Any additional comments?

_No response_

IWANABETHATGUY:
### Description
Closed https://github.com/rolldown/rolldown/issues/2005

### DIFF
--- a/crates/rolldown/tests/integration_rolldown.rs
+++ b/crates/rolldown/tests/integration_rolldown.rs
@@ -5,6 +5,7 @@ use rolldown_testing::{
   fixture::Fixture,
   integration_test::IntegrationTest,
   test_config::{read_test_config, TestConfig, TestMeta},
+  workspace,
 };
 use sugar_path::SugarPath;
 use testing_macros::fixture;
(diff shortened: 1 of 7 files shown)


crates/rolldown/tests/integration_rolldown.rs
```diff
@@ -61,5 +62,15 @@ async fn filename_with_hash() {
...
+  let snapshot_path = workspace::crate_dir("rolldown")
+    .join("tests/snapshots/integration_rolldown__filename_with_hash.snap");
+  let before_snapshot = std::fs::read_to_string(&snapshot_path).unwrap();
+
   insta::assert_snapshot!(snapshot_outputs.join("\n"));
+
+  let in_github_action = std::env::var("GITHUB_ACTIONS").is_ok();
+  let after_snapshot = std::fs::read_to_string(&snapshot_path).unwrap();
+  if (before_snapshot != after_snapshot && in_github_action) {
+    std::process::exit(-1);
```
hyf0: Should  this be done by insta?

https://github.com/rolldown/rolldown/blob/dde9b33b409bfc909d089769d2817f178e00f30f/.github/workflows/reusable-cargo-test.yml#L36-L37

IWANABETHATGUY: you are right.

IWANABETHATGUY: https://github.com/rolldown/rolldown/actions/runs/10431525166/job/28891291395?pr=2006, but actually it did not work

IWANABETHATGUY: image

IWANABETHATGUY: I agree with you that setting the env is correct way to fix this issue.

hyf0 (approved):

IWANABETHATGUY:
The ci failed because Windows error message is different from linux and mac, 
image

hyf0:
Let's ignore this test temparaly.

PR #2007, opened 2024-08-17
feat: compose addon hooks

### LINKED ISSUE #1999: [Feature Request]: Support `banner`, `footer`, `intro` and `outro` for composejs plugin

Making `banner`, `footer`, `intro` and `outro` composable for compose js plugin
for now, these hook will bailout https://github.com/rolldown/rolldown/blob/1b49ed1102469e42283b38fa115cfb215bf25629/packages/rolldown/src/utils/compose-js-plugins.ts#L10-L28

n/a

ikkz:
resolves #1999

### DIFF
--- a/packages/rolldown/src/utils/compose-js-plugins.ts
+++ b/packages/rolldown/src/utils/compose-js-plugins.ts
@@ -9,8 +9,6 @@ import { AssertNever } from './type-assert'
 
 const unsupportedHookName = [
   'augmentChunkHash',
-  'banner',
-  'footer',
   'generateBundle',
   'moduleParsed',
   'onLog',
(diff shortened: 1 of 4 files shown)


IWANABETHATGUY:
Can you double-check the implementation? It did not compose at all.

ikkz:
> Can you double-check the implementation? It did not compose at all.

I added some log [here](https://github.com/rolldown/rolldown/blob/main/packages/rolldown/src/utils/normalize-input-options.ts#L16) 

image

and output of `just test-node rolldown -t "multi-addon-composing"`:

image

It seems that the 8 test plugins have been composed to 1 plugin.

Did I misunderstand this issue?

IWANABETHATGUY:
> plugins

It's my mistake, we need a technique to determine if the plugin is composed, I am working on that.

IWANABETHATGUY:
Can you add a test like https://github.com/rolldown/rolldown/pull/2008/files#diff-e709cc275fbba560e8cd300f426f59152876a8ce9c02bf8283a61fdc258cf319, so we could verify it is correctly composed

ikkz:
> Can you add a test like https://github.com/rolldown/rolldown/pull/2008/files#diff-e709cc275fbba560e8cd300f426f59152876a8ce9c02bf8283a61fdc258cf319, so we could verify it is correctly composed

Done

IWANABETHATGUY (approved):
thanks

PR #2008, opened 2024-08-17
chore:  enhance testing composejs plugin

IWANABETHATGUY:
### Description
1. Adding an example to test if the plugins is correctly composed.

### DIFF
--- a/packages/rolldown/src/rolldown-build.ts
+++ b/packages/rolldown/src/rolldown-build.ts
@@ -26,7 +26,6 @@ export class RolldownBuild {
       this.#bundler = bundler
       this.#stopWorkers = stopWorkers
       this.normalizedInputOptions = normalizedInputOptions
-      console.log(`this.normalizedInputOptions: `, this.normalizedInputOptions)
     }
     return this.#bundler
   }
--- a/packages/rolldown/src/rolldown-build.ts
+++ b/packages/rolldown/src/rolldown-build.ts
@@ -6,11 +6,13 @@ import { createBundler } from './utils/create-bundler'
 import type { RolldownOutput } from './types/rolldown-output'
 import type { HasProperty, TypeAssert } from './utils/type-assert'
 import type { InputOptions } from './options/input-options'
+import { NormalizedInputOptions } from './options/normalized-input-options'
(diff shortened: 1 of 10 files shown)


packages/rolldown/src/rolldown-build.ts
```diff
@@ -6,11 +6,13 @@ import { createBundler } from './utils/create-bundler'
 import type { RolldownOutput } from './types/rolldown-output'
 import type { HasProperty, TypeAssert } from './utils/type-assert'
 import type { InputOptions } from './options/input-options'
+import { NormalizedInputOptions } from './options/normalized-input-options'
 
 export class RolldownBuild {
   #inputOptions: InputOptions
   #bundler?: Bundler
   #stopWorkers?: () => Promise<void>
+  normalizedInputOptions?: NormalizedInputOptions
```
hyf0: Not sure about leaking this to outside just for testing.

hyf0:
What if we expose `composeJsPlugins` under `rolldown/expermental` to outside and test it directly?

packages/rolldown/src/rolldown-build.ts
IWANABETHATGUY: that's maybe not the best way to do it, but I don't find other methods to get normalized plugin length.

hyf0: I believe testing `composeJsPlugins` directly is more proper.

IWANABETHATGUY:
> What if we expose `composeJsPlugins` under `rolldown/expermental` to outside and test it directly?

Sounds like a good idea.

scripts/meta/constants.js
```diff
@@ -1,4 +1,5 @@
 import * as selfExports from './constants.js'
+// oxlint-disable
```
IWANABETHATGUY, line 2: avoid race condition

hyf0:
We don't need to write everything in fixture test. We still use fixture-based tests to testing if enabling composingPlugins give correct output. But the behavior of `composeJsPlugins` itself could be tested directly in another test files.

hyf0 (approved):

PR #2016, opened 2024-08-18
perf: ⚡️ remove extra collection when using rayon

IWANABETHATGUY:

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
+++ b/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
@@ -64,7 +64,8 @@ impl<'a> GenerateStage<'a> {
       &mut index_imports_from_other_chunks,
     );
 
-    let index_sorted_cross_chunk_imports = std::mem::take(&mut index_cross_chunk_imports.raw)
+    let index_sorted_cross_chunk_imports = index_cross_chunk_imports
+      .raw
       .into_par_iter()
       .map(|cross_chunk_imports| {
         let mut cross_chunk_imports = cross_chunk_imports.into_iter().collect::<Vec<_>>();
--- a/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
+++ b/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
@@ -64,10 +64,7 @@ impl<'a> GenerateStage<'a> {
       &mut index_imports_from_other_chunks,
     );
 
-    let index_sorted_cross_chunk_imports = index_cross_chunk_imports
-      .into_iter()
-      // FIXME: Extra traversing. This is a workaround due to `par_bridge` doesn't ensure order https://github.com/rayon-rs/rayon/issues/551#issuecomment-882069261
-      .collect::<Vec<_>>()
+    let index_sorted_cross_chunk_imports = std::mem::take(&mut index_cross_chunk_imports.raw)
       .into_par_iter()
       .map(|cross_chunk_imports| {
         let mut cross_chunk_imports = cross_chunk_imports.into_iter().collect::<Vec<_>>();
--- a/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
+++ b/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
@@ -64,6 +64,7 @@ impl<'a> GenerateStage<'a> {
       &mut index_imports_from_other_chunks,
     );
 
+    // TODO: support rayon trait in `oxc_index`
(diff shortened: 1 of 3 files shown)


crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
```diff
@@ -64,10 +64,7 @@ impl<'a> GenerateStage<'a> {
       &mut index_imports_from_other_chunks,
     );
 
-    let index_sorted_cross_chunk_imports = index_cross_chunk_imports
-      .into_iter()
-      // FIXME: Extra traversing. This is a workaround due to `par_bridge` doesn't ensure order https://github.com/rayon-rs/rayon/issues/551#issuecomment-882069261
```
hyf0, line 69: Should keep this. What this PR does is still to be considered as a workaround.

IWANABETHATGUY: why this is a workaround?

IWANABETHATGUY: According to https://github.com/rayon-rs/rayon/issues/669#issuecomment-507814077, the ordering now is preserved without another traverse

crates/rolldown/src/utils/chunk/finalize_chunks.rs
```diff
@@ -56,10 +59,9 @@ pub fn finalize_assets(
     index_vec![Xxh3::default(); preliminary_assets.len()];
 
   let index_final_hashes: IndexVec<AssetIdx, String> = index_asset_hashers
-    .into_iter_enumerated()
-    // FIXME: Extra traversing. This is a workaround due to `par_bridge` doesn't ensure order https://github.com/rayon-rs/rayon/issues/551#issuecomment-882069261
-    .collect::<Vec<_>>()
+    .raw
     .into_par_iter()
+    .enumerate()
     .map(|(asset_idx, mut hasher)| {
```
hyf0, line 65: `asset_idx` become `usize` instead of `AssetIdx`, this is unexpected. I'm not sure why `oxc_index` allows using `usize` as the valid index type. But the reason we use `IndexVec` is to have a domain-specific `Vec` to prevent misusing wrong index. Otherwise we could just use std `Vec`.

crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
hyf0: https://github.com/rolldown/rolldown/pull/2016/files#r1721029588

hyf0: It's not about rayon here. IMO, this should be solved by the `oxc_index` to have a native support of rayon.

IWANABETHATGUY: Then that should be TODO: wait oxc_index support `IntoParIterator` rather than `Fixme:`, since the bug is fixed, no overhead needed.

hyf0: Yeah.

crates/rolldown/src/utils/chunk/finalize_chunks.rs
hyf0: @overlookmotel Sorry for bothering. Could you take have a rough look if this is expected/reasonable behavior of `oxc_index`?

IWANABETHATGUY:
TODO: is added

crates/rolldown/src/utils/chunk/finalize_chunks.rs
```diff
@@ -55,11 +57,11 @@ pub fn finalize_assets(
   let index_asset_hashers: IndexVec<AssetIdx, Xxh3> =
     index_vec![Xxh3::default(); preliminary_assets.len()];
 
+  // TODO: support rayon trait in `oxc_index`
```
hyf0:
```suggestion
  // TODO: support `par_into_iter_enumerated`/rayon trait in `oxc_index`
```

hyf0 (approved):

crates/rolldown/src/utils/chunk/finalize_chunks.rs
overlookmotel: @hyf0 Sorry I'm so late to come back on this. It got lost in my Github notifications.

I've made a PR to prevent `IndexVec`s being indexed into with a `usize` (https://github.com/oxc-project/oxc/pull/5733). As you say, that's a bad idea! That oddity was inherited from `index_vec` crate.

Opened an issue for the par iter problem: https://github.com/oxc-project/oxc/issues/5734

PR #2019, opened 2024-08-19
chore: remove unused ts file

IWANABETHATGUY:


underfin (approved):

packages/rolldown/src/utils/plugin/index.ts
hyf0, line 1: Could you keep this one? I'm gonna use it in composingJsPlugins.

IWANABETHATGUY: ok

PR #2021, opened 2024-08-19
feat(node/composingJsPlugins): support `resolveId` hook

hyf0:
### Description

You could actually call `this.resolveId` in `buildStart`... hooks, so the current bailout logic is wrong. To get a correct semantic, we probably need to bailout all hooks that could call `this.resolveId`. This basically make `composingJsPlugins` unuseable.

However, this PR solve the issue partially, which means, with this PR, we could ensure the correct semantic of calling `this.resolve` in `resolveId` hook.

Fixes for other hooks would be raised later. Basically we need intercept all hooks's `PluginContext` for plugins that implemented `resolveId` hook.

### DIFF
--- a/packages/rolldown/src/constants/plugin-context.ts
+++ b/packages/rolldown/src/constants/plugin-context.ts
@@ -1,0 +1,9 @@
+/**
+ * If Composed plugins call `this.resolve` with `skipSelf: true`, the composed plugins will be skipped as a whole.
+ * To prevent that, we use this symbol to store the actual caller of `this.resolve` with `skipSelf: true`. And we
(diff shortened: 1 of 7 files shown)


hyf0:
* **#2022** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2022?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2021** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2021?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

hyf0:
just FYI @IWANABETHATGUY cc

PR #2025, opened 2024-08-20
chore: ⚡️ faster lint repo

IWANABETHATGUY:
prettier, cspell can not max utilize CPU resource, they are often single threaded, so using `npm-run-all` to run them in parallel
### Description
# Before
________________________________________________________
Executed in    9.79 secs    fish           external
   usr time   14.91 secs  459.00 micros   14.91 secs
   sys time    2.41 secs  137.00 micros    2.41 secs

# After
________________________________________________________
Executed in    5.16 secs    fish           external
   usr time   13.93 secs  495.00 micros   13.93 secs
   sys time    2.16 secs  138.00 micros    2.16 secs


IWANABETHATGUY:
it did not affect core

PR #2026, opened 2024-08-20
chore: 🤖 try to use oxc builder alloc based api

IWANABETHATGUY:
### Description
1. try to use `oxc_ast` builder API in more case.
2. align with `oxc_ast` naming convention, function return `Boxed` ast prefixed with `alloc_`


IWANABETHATGUY:
just a refactor and did not affect the core, skip reviewing.

PR #2027, opened 2024-08-20
WIPfix: 🐛 comment gen regression

### LINKED ISSUE #2013: [Bug]: A bad case in comment generation.
[https://github.com/rolldown/rolldown/pull/2011/files#diff-f9f6cb612053f087f69bd1fe270e5ac6deddbaf63b53107eca4038a4a688b9b9R109
This diff is unexpected, since ](https://github.com/rolldown/rolldown/pull/2011/files#diff-e21d568d8f246859eb84351e0780db56bc283d2dd79f8b58805c1e8898f4f10fR21)
the comment should not be generated.

IWANABETHATGUY:
### Description
1. Closed #2013

### DIFF
--- a/crates/rolldown/tests/esbuild/dce/no_side_effects_comment/artifacts.snap
+++ b/crates/rolldown/tests/esbuild/dce/no_side_effects_comment/artifacts.snap
@@ -106,7 +106,7 @@ export { f as default };
 ```js
 
 //#region stmt-export-default-before-async-fn-anon.js
-/* #__NO_SIDE_EFFECTS__ */ async function stmt_export_default_before_async_fn_anon_default() {}
+async function stmt_export_default_before_async_fn_anon_default() {}
 
 //#endregion
 export { stmt_export_default_before_async_fn_anon_default as default };
@@ -116,7 +116,7 @@ export { stmt_export_default_before_async_fn_anon_default as default };
 ```js
 
 //#region stmt-export-default-before-async-fn-name.js
-/* #__NO_SIDE_EFFECTS__ */ async function f() {}
+async function f() {}
 
 //#endregion
 export { f as default };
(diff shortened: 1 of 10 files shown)


crates/rolldown/tests/esbuild/dce/no_side_effects_comment/artifacts.snap
IWANABETHATGUY, line 1: To fixing this issue, `oxc_codegen` should allow passing comment move_map. e.g.

```js
/* @__NO_SIDE_EFFECTS__ */ export function a() {}
//                         ^^^^^^ will be removed after transform
//                         when oxc codegen cursor move to `function a()`
//                         it try to take first leading comment, `/* @__NO_SIDE_EFFECTS__ */`
//                         the content between `/* @__NO_SIDE_EFFECTS__ */` and `function a()` is ` export `
//                         so the attachment would fail, since the content should be all whitespace.

```

IWANABETHATGUY: This is another hack, we should came up a better idea after the AST Id stabilize

crates/rolldown/tests/rolldown/tree_shaking/pure_annotation/artifacts.snap
```diff
@@ -18,7 +18,7 @@ const c = test();
 //#endregion
 //#region d.js
 function test2() {}
-const e = /* #__PURE__ */ test2();
```
IWANABETHATGUY, line 21: This is expected, 
the original code is 

```js
function test2() {}

export const d = /* #__PURE__ */                 test2(), e =  test2();
              
```

`PURE` should attached to the first `test2()`.

PR #2029, opened 2024-08-20
fix:  JSON.parse() should not have side effect

IWANABETHATGUY:
### Description
1. Before the test case will fail, since `JSON.parse` is considered to have side effects.


packages/rolldown/tests/fixtures/builtin-plugin/json/side-effects/main.js
```diff
@@ -0,0 +1,3 @@
+import json from './package.json'
```
IWANABETHATGUY, line 1: Make `json` unused.

underfin (approved):

PR #2034, opened 2024-08-21
fix: Allowing gcc builds on windows

Kreijstal:
### Description

gcc runs on windows, gcc does not accept /FORCE

image


hyf0 (approved):
Thanks

PR #2037, opened 2024-08-22
chore: add task name prefix for tasks in `lint-repo`.

7086cmd:
Utilize the `--label-name` options to display the task name, such as `lint-pretter`, as the prefix in the `npm-run-all` command. This approach enhances the user experience by providing a more intuitive and visually appealing display of lint errors.

This does not impact the core, so self-assigned.


hyf0:
I believe there's no need to use `self-assigned` to approve this PR.

hyf0 (approved):

PR #2039, opened 2024-08-22
feat: support `-p <plugins>` in cli.

7086cmd:
Align with Rollup and suport `builtin plugins`, but not supported for “parallel plugins” at this time.

I am seeking an approach for testing built-in plugins.


package.json
```diff
@@ -29,6 +29,7 @@
   "license": "MIT",
   "devDependencies": {
     "@ls-lint/ls-lint": "^2.2.3",
+    "@rollup/plugin-dsv": "^3.0.4",
```
7086cmd, line 33: This is utilized for testing the plugin.

packages/rolldown/src/cli/plugins/resolver.ts
```diff
@@ -0,0 +1,151 @@
...
+      break
+    case 'load-fallback':
+      builtinPlugins.loadFallbackPlugin()
+      break
+    case 'json':
+      builtinPlugins.jsonPlugin()
+      break
+    default:
+      // alias plugin doesn't support non-config options
+      return
```
7086cmd, line 124: I am curious to know if there is an alternative method to find it if it is a built-in plugin.

7086cmd:
stale. Let's find a better way to implement it for built-in plugins.

PR #2040, opened 2024-08-22
perf: use `simdutf8` to validate UTF-8 when reading files

overlookmotel:
### Description

I'm not familiar with Rolldown's codebase, so I'm not sure whether this function is the main one used to read source text from files - so this may be in wrong place/not enough places.

`std::fs::read_to_string` internally calls `std::str::from_utf8` to check that the file is valid UTF-8. `str::from_utf8` is not so efficient as it doesn't use SIMD. `simduft8` should be faster.

I assume that this difference won't show up on benchmarks as they likely won't include IO. But I'd hope it'd have an effect in the real world. How would we go about testing that assumption? Are there any benchmarks which do exercise this function?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2279,6 +2279,7 @@ version = "0.1.0"
 dependencies = [
  "dunce",
  "oxc_resolver",
+ "simdutf8",
  "vfs",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -141,6 +141,7 @@ schemars            = "0.8.21"
 self_cell           = "1.0.4"
 serde               = { version = "1.0.203", features = ["derive"] }
 serde_json          = "1.0.117"
+simdutf8            = { version = "0.1.4", features = ["aarch64_neon"] }
 smallvec            = "1.13.2"
 sugar_path          = { version = "1.2.0", features = ["cached_current_dir"] }
 testing_macros      = "0.2.13"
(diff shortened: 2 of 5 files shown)


Boshen:
I assume this would break our wasm and wasi builds without and feature target gates.

overlookmotel:
`simduft8` supports WASM, although it's not clear to me if it's enabled by default: https://github.com/rusticstuff/simdutf8#wasm32

But it shouldn't break WASM anyway.

Boshen:
We can always test the effect of such functionalities in oxc first, oxlint in this case.

overlookmotel:
Do oxlint's benchmarks include file IO?

Boshen:
> Do oxlint's benchmarks include file IO?

I need to profile locally.

Boshen:
In monitor-oxc, validating 84190 files (before and after):

image

Performance is greatly improved, but maybe M3 is too powerful to make a dent on the overall performance improvement.

IWANABETHATGUY (approved):

IWANABETHATGUY:
@Boshen since you assign to yourself, this pull request needs your approval to merge.

Boshen (approved):

overlookmotel:
I wasn't sure if this change also needs making elsewhere in Rolldown, if other parts of the code also read from file system. @IWANABETHATGUY can you advise?

PR #2042, opened 2024-08-23
feat(rolldown): bump oxc to v0.25.0

Boshen:


crates/rolldown/tests/rolldown/function/minify/basic/artifacts.snap
```diff
@@ -6,26 +6,26 @@ source: crates/rolldown_testing/src/integration_test.rs
 ## main.mjs
 
 ```js
-import {default as assert} from "node:assert";const a$1="shared.js",a="index.js";assert.equal(a,"index.js");assert.equal(a$1,"shared.js");
+import {default as assert} from "node:assert";const b="shared.js",c="index.js";assert.equal(c,"index.js");assert.equal(b,"shared.js");
```
Boshen, line 9: This is properly mangled now.

crates/rolldown/tests/esbuild/dce/no_side_effects_comment/artifacts.snap
```diff
@@ -106,7 +106,7 @@ export { f as default };
 ```js
 
 //#region stmt-export-default-before-async-fn-anon.js
-/* #__NO_SIDE_EFFECTS__ */ async function stmt_export_default_before_async_fn_anon_default() {}
+async function stmt_export_default_before_async_fn_anon_default() {}
```
Boshen, line 109: @IWANABETHATGUY Can you double check these `__NO_SIDE_EFFECTS__` change?

crates/rolldown/tests/rolldown/function/minify/basic/artifacts.snap
IWANABETHATGUY: nice

crates/rolldown/tests/esbuild/dce/no_side_effects_comment/artifacts.snap
IWANABETHATGUY: they are expected, https://github.com/rolldown/rolldown/pull/2027 I clarify the reason.

IWANABETHATGUY (approved):

PR #2047, opened 2024-08-23
feat: hook filter

IWANABETHATGUY:


crates/rolldown_plugin/src/plugin.rs
```diff
@@ -252,4 +254,16 @@ pub trait Plugin: Any + Debug + Send + Sync + 'static {
   fn transform_ast_meta(&self) -> Option<PluginHookMeta> {
     None
   }
+
+  fn get_transform_filter(&self) -> anyhow::Result<Option<TransformHookFilter>> {
```
hyf0: `get_transform_filter` => `transform_filter`.

https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

IWANABETHATGUY:
merged since it needs to test in rolldown-vite, all tests are passed, TODO: needs support filter with `compose-js-plugin`

PR #2051, opened 2024-08-24
feat: build import analysis plugin

IWANABETHATGUY:


crates/rolldown/src/utils/ecma_visitors/mod.rs
```diff
@@ -22,6 +22,11 @@ impl<'a> VisitMut<'a> for EnsureSpanUniqueness {
     self.ensure_uniqueness(it.span_mut());
     walk_mut::walk_module_declaration(self, it);
   }
+
+  fn visit_import_expression(&mut self, it: &mut oxc::ast::ast::ImportExpression<'a>) {
+    self.ensure_uniqueness(it.span_mut());
+    walk_mut::walk_import_expression(self, it);
+  }
```
IWANABETHATGUY, line 29: This pr will insert `import('dep')` it may lead conflict if you have multiple `await ()` that needs to transform.

crates/rolldown_plugin/src/plugin_driver/build_hooks.rs
```diff
@@ -194,8 +194,10 @@ impl PluginDriver {
     for (_, plugin, ctx) in
       self.iter_plugin_with_context_by_order(&self.order_by_transform_ast_meta)
     {
-      args.ast =
-        plugin.call_transform_ast(ctx, HookTransformAstArgs { cwd: args.cwd, ast: args.ast })?;
+      args.ast = plugin.call_transform_ast(
+        ctx,
+        HookTransformAstArgs { cwd: args.cwd, ast: args.ast, id: args.id },
```
IWANABETHATGUY, line 199: used to filter out `module` in `node_modules`, https://github.com/rolldown/vite/blob/8eed5766df862049aae43a7f1611acb83363b89a/packages/vite/src/node/plugins/importAnalysisBuild.ts#L217-L219

hyf0 (approved):

PR #2065, opened 2024-08-26
perf(rolldown): generate unique span by decrementing u32::MAX

Boshen:
We assume

* newly inserted span is always empty (`(0, 0)`)
* it is unlikely that a span is at `u32::MAX`

### DIFF
--- a/crates/rolldown/src/utils/ecma_visitors/mod.rs
+++ b/crates/rolldown/src/utils/ecma_visitors/mod.rs
@@ -2,12 +2,9 @@ use oxc::{
   ast::{visit::walk_mut, VisitMut},
   span::{GetSpanMut, Span},
 };
-use rustc_hash::FxHashSet;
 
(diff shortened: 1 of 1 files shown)


Boshen:
No performance change? And where is codspeed?

IWANABETHATGUY:
image,  looks like it is already uploaded.
Don't know why not display as a comment

image

Boshen:
image

Boshen:
We have a very stupid thing:

https://github.com/oxc-project/oxc/blob/46b641b75d9da1ad1c3a0dd589766956880a583f/crates/oxc_span/src/span/mod.rs#L348-L352

```
impl Hash for Span {
    fn hash<H: Hasher>(&self, _state: &mut H) {
        // hash to nothing so all ast spans can be comparable with hash
    }
}
```

We need to make some changes in places where we need to hash the span in rolldown ...

Issue #2057, opened 2024-08-25
[Tracking] `ReplacePlugin`


hyf0:
`ReplacePlugin` https://github.com/rolldown/rolldown/pull/2056 is the builtin version of rolldown that align with `rollup-plugin-reaplce`.

## To be finished

- [x] Migrate all tests
  - Migrated: https://github.com/rolldown/rolldown/tree/1a5b2a3ebfa0e8e5bbb2e81eb33ed297a51788bf/crates/rolldown_plugin_replace/tests/form
  - To be migrated: https://github.com/rollup/plugins/tree/master/packages/replace/test/fixtures/form
- [x] Support `objectGuards` option
- [x] Support `preventAssignment` option
- [ ] Support sourcemap

References:

- https://github.com/rollup/plugins/blob/master/packages/replace/README.md

ikkz:
Can I try this?

hyf0:
Sure.

vhfmag:
Hi @hyf0! I'm taking a look into this issue and I'm seeing this error when trying to use `MagicString::source_map` to generate a source map:

```
mismatched types
`oxc_sourcemap::sourcemap::SourceMap` and `oxc_sourcemap::sourcemap::SourceMap` have similar names, but are actually distinct types
perhaps two different versions of crate `oxc_sourcemap` are being used?rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic message [0]?0#file:///Users/vmagalhaes/dev/rolldown/crates/rolldown_plugin_replace/src/plugin.rs)
```

This seems to happen because the latest `magic_string` depends on `oxc_sourcemap` 0.25.0, while rolldown itself depends on `oxc_sourcemap` 0.31.0. I just thought I'd check if generating a source map with `MagicString::source_map` seems like a good plan so that I can look into upgrading `oxc_sourcemap` in `magic_string`. Thanks!

PR #2074, opened 2024-08-26
test: add tests for replace plugin

ikkz:
related issue: #2057 

migrated from https://github.com/rollup/plugins/tree/master/packages/replace/test/fixtures/form

These tests have not been included in this pr:
- replacement-function: ~~will be added once I implement support for function as `Replacement`~~
- observe-plugin-options: we already have separated `values` and plugin options so it is unnecessary

### DIFF
--- a/crates/rolldown_plugin_replace/tests/form/assignment/mod.rs
+++ b/crates/rolldown_plugin_replace/tests/form/assignment/mod.rs
@@ -1,0 +1,28 @@
+use std::sync::Arc;
+
+use rolldown::BundlerOptions;
+
+use rolldown_plugin_replace::{ReplaceOptions, ReplacePlugin};
+use rolldown_testing::{abs_file_dir, integration_test::IntegrationTest, test_config::TestMeta};
+
+// doesn't replace lvalue in assignment
+// #[tokio::test(flavor = "multi_thread")]
+#[allow(dead_code)]
+async fn assignment() {
+  let cwd = abs_file_dir!();
+
+  IntegrationTest::new(TestMeta { expect_executed: false, ..Default::default() })
+    .run_with_plugins(
+      BundlerOptions {
+        input: Some(vec!["./input.js".to_string().into()]),
+        cwd: Some(cwd),
+        ..Default::default()
+      },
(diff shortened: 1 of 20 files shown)


hyf0:
`replacement-function` is not required. There's no plan to support it.

crates/rolldown_plugin_replace/src/plugin.rs
```diff
@@ -100,3 +100,24 @@ impl Plugin for ReplacePlugin {
...
 }
+
+#[cfg(test)]
+mod tests {
+  use super::{ReplaceOptions, ReplacePlugin};
+  use string_wizard::MagicString;
+
+  #[allow(unused_variables)]
+  #[test]
+  fn typescript_declare() {
```
hyf0: This test should be written in form just like others.

ikkz:
> `replacement-function` is not required. There's no plan to support it.

 I see

crates/rolldown_plugin_replace/tests/form/assignment/mod.rs
```diff
@@ -0,0 +1,28 @@
+use std::sync::Arc;
+
+use rolldown::BundlerOptions;
+
+use rolldown_plugin_replace::{ReplaceOptions, ReplacePlugin};
+use rolldown_testing::{abs_file_dir, integration_test::IntegrationTest, test_config::TestMeta};
+
+// doesn't replace lvalue in assignment
+// #[tokio::test(flavor = "multi_thread")]
+#[allow(dead_code)]
```
hyf0, line 10: You could use `// #[tokio::test(flavor = "multi_thread"), ignore="TODO: implement the feature"]`

hyf0: Anyway, this is small thing.

hyf0 (approved):
Thanks!

crates/rolldown_plugin_replace/tests/form/assignment/mod.rs
ikkz: > You could use `// #[tokio::test(flavor = "multi_thread"), ignore="TODO: implement the feature"]`

ok, it is more elegant and I will use it in subsequent PRs

PR #2081, opened 2024-08-28
feat: support `output.inlineDynamicImports` option.

### LINKED ISSUE #1841: output.inlineDynamicImports (format umd/iife or ssr worker build)
https://rollupjs.org/configuration-options/#output-inlinedynamicimports

7086cmd:
resolves #1841.

The logic for the `inline dynamic imports` option has been implemented, but it is currently hard-coded only for `iife` format. I have added an option called `inlineDynamicImports` that aligns with Rollup’s approach to manually handle this scenario.

Rollup’s default value for this option is [`false`](https://rollupjs.org/configuration-options/#output-inlinedynamicimports). However, I believe we can check if the format is `iife` to determine whether to enable this option, but it may be not so elegant. Any suggestions on how to proceed?


hyf0:
Does rollup `inlineDynamicImports` no matter if `inlineDynamicImports` enabled while targeting IFFE format?

crates/rolldown/src/stages/link_stage/mod.rs
```diff
@@ -62,42 +62,47 @@ pub struct LinkStage<'a> {
...
+      .map(|module| LinkingMetadata {
+        dependencies: module
+          .import_records()
+          .iter()
+          .filter_map(|rec| {
+            if options.inline_dynamic_imports || !matches!(rec.kind, ImportKind::DynamicImport) {
+              Some(rec.resolved_module)
+            } else {
+              if matches!(options.format, OutputFormat::Iife) {
+                // Throw an error.
```
underfin: The rollup check chunks count to give warnings, see https://github.com/rollup/rollup/blob/master/src/Bundle.ts#L68. I think you should check it at code splitting.

crates/rolldown_error/src/build_error/error_constructors.rs
```diff
@@ -176,6 +177,16 @@ impl BuildDiagnostic {
     Self::new_inner(ForbidConstAssign { filename, source, name, reference_span, re_assign_span })
   }
 
+  pub fn invalid_option(situation: &str, option: String) -> Self {
+    Self::new_inner(InvalidOption {
+      invalid_option_types: match situation {
+        "output.format" => InvalidOptionTypes::UnsupportedCodeSplittingFormat,
```
underfin: Here is better using an enum to flag `invalidOptionWithUnsupportedCodeSplittingFormat`.

underfin (dismissed):

crates/rolldown/src/stages/link_stage/mod.rs
underfin: > Does rollup inlineDynamicImports no matter if inlineDynamicImports enabled while targeting IFFE format?

cc @hyf0 The rollup checked at here.

7086cmd: In this scenario, Rollup throws an error (as seen [here](https://github.com/rollup/rollup/blob/master/src/Bundle.ts#L228)).

However, regarding [#1393](https://github.com/rolldown/rolldown/pull/1393#discussion_r1641910564), there's a concern:
> There might be issues with this implementation when working with asset modules and future copy and file loaders (see [#1295](https://github.com/rolldown/rolldown/issues/1295)).

Rollup's approach can be found [here](https://github.com/rollup/rollup/blob/186706b7229943c97e0f2742a2944c6a5d74b0ff/src/Bundle.ts#L228-L244).

I propose implementing this logic at this point and deferring the handling of asset modules to the `file` / `copy` loader implementation. This approach should also apply to the single file checker.

crates/rolldown_error/src/build_error/error_constructors.rs
7086cmd: Rollup also handles errors in a similar way, as seen [here](https://github.com/rollup/rollup/blob/186706b7229943c97e0f2742a2944c6a5d74b0ff/src/Bundle.ts#L237-L252). Given the similarity in logic, I've used the `InvalidOption` error with `UnsupportedCodeSplittingFormat`.

We can adopt a similar approach to Rollup: when supporting `output.file` and `output.sourcemapFile`, we can use the same error handling mechanism. This aligns with the multi-chunk output configuration checker.

Specifically, we can apply this error handling when dealing with `output.file` and `output.sourcemapFile`, ensuring this approach is consistent with the multi-chunk output configuration checker.

7086cmd:
> Does rollup `inlineDynamicImports` no matter if `inlineDynamicImports` enabled while targeting IFFE format?

I'm considering automatically setting the value to `true` for `iife` and `umd` formats (in the future). This logic should be implemented on the JavaScript side. However, I encountered unexpected testing errors during implementation. If automatically setting to `true` is beneficial, I'll attempt to normalize this behavior in JavaScript again.

underfin (approved):

crates/rolldown/tests/rolldown/errors/invalid_option/unsupported_code_splitting_format/artifacts.snap
```diff
@@ -0,0 +1,52 @@
...
+
+```# warnings
+
+## MISSING_NAME_OPTION_FOR_IIFE_EXPORT
+
+```text
+[MISSING_NAME_OPTION_FOR_IIFE_EXPORT] Warning: If you do not supply "output.name", you may not be able to access the exports of an IIFE bundle.
+
+```
+## MISSING_NAME_OPTION_FOR_IIFE_EXPORT
```
underfin: Here duplicated warnings need to fixed it. We could fix it at new pr.

PR #2089, opened 2024-08-29
fix: entry chunk should be at first

underfin:
### Description

The rollup entry chunks should be at first, we need to make sure the behavior.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/code_splitting.rs
@@ -210,7 +210,6 @@ impl<'a> GenerateStage<'a> {
       .for_each(|(i, chunk)| {
         chunk.exec_order = i.try_into().expect("Too many chunks, u32 overflowed.");
       });
-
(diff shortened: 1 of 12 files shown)


hyf0 (approved):

crates/rolldown/src/stages/generate_stage/code_splitting.rs
```diff
@@ -238,9 +237,21 @@ impl<'a> GenerateStage<'a> {
               a_should_be_first
             }
           }
-          _ => chunks[*a].exec_order.cmp(&chunks[*b].exec_order),
+          (
+            ChunkKind::EntryPoint { is_user_defined: a_is_user_defined, .. },
+            ChunkKind::EntryPoint { is_user_defined: b_is_user_defined, .. },
+          ) => {
+            if *a_is_user_defined && *b_is_user_defined {
+              // Using user specific order of entry
```
underfin, line 245: Here fixed user specific order of entry

hyf0: Should we make sure the entry asset order is the same as the user's `input` option?

hyf0:
Since there are no failing on tests, let's merged it first.

hyf0 (approved):

PR #2092, opened 2024-08-30
chore: print seconds when build time exceeds 1 second.

### LINKED ISSUE #2091: [Feature Request]: Print execution time in seconds for > 1s

When the total execution time exceeds 1 second, it should print 1.23 seconds instead of 1230ms.

I'm getting 

```
Finished in 10479.08 ms
Finished in 6696.75 ms
```

which is really hard to read.

```
Finished in 10.47 s
Finished in 6.69 s
```

7086cmd:
resolves #2091 
However, I am curious about how I can add tests.

### DIFF
--- a/packages/rolldown/src/cli/commands/bundle.ts
+++ b/packages/rolldown/src/cli/commands/bundle.ts
@@ -61,9 +61,12 @@ async function bundleInner(
 
   logger.log(``)
   const duration = endTime - startTime
-  logger.success(
-    `Finished in ${colors.bold(duration < 1000 ? `${duration.toFixed(2)} ms` : `${(duration / 1000).toFixed(2)} s`)}`,
-  )
+  // If the build time is more than 1s, we should display it in seconds.
+  const spent =
+    duration < 1000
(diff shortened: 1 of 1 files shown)


packages/rolldown/src/cli/commands/bundle.ts
```diff
@@ -60,8 +60,9 @@ async function bundleInner(
   printBundleOutputPretty(bundleOutput)
 
   logger.log(``)
+  const duration = endTime - startTime
   logger.success(
-    `Finished in ${colors.bold((endTime - startTime).toFixed(2))} ms`,
+    `Finished in ${colors.bold(duration < 1000 ? `${duration.toFixed(2)} ms` : `${(duration / 1000).toFixed(2)} s`)}`,
```
IWANABETHATGUY: prefer extract this conditional expr as a variable, since it is hard to read.

IWANABETHATGUY (approved):

IWANABETHATGUY:
> resolves #2091 However, I am curious about how I can add tests.

Adding a unit test for the `time string formatter` is enough if you want.

PR #2093, opened 2024-08-30
feat(plugin/replace): prevent assignment

ikkz:
related issue: #2057

This pr implements the same function with rollup's preventAssignment: https://github.com/rollup/plugins/blob/master/packages/replace/src/index.js#L77-L82

And part of the logic is put into the match loop, because of `fancy-regex`'s limitation that variable size of expression cannot be used in lookbehind (https://github.com/fancy-regex/fancy-regex/issues/74 and https://github.com/fancy-regex/fancy-regex/issues/109)


hyf0:
> And part of the logic is put into the match loop, because of fancy-regex's limitation that variable size of expression cannot be used in lookbehind (https://github.com/fancy-regex/fancy-regex/issues/74 and https://github.com/fancy-regex/fancy-regex/issues/109)

Looks like both regex and fancy_regex have issues on porting regex of js. I'm considering taking https://github.com/ridiculousfish/regress or https://github.com/hyf0/ecma_regex/issues more seriously.

hyf0 (approved):
Thank you!

PR #2102, opened 2024-09-01
refactor: replace qualified path with use for `oxc::ast::ast`

IWANABETHATGUY:
### Description
1. we don't need to worry about the conflicting since the naming only used for AstNode


hyf0 (approved):
Just sharing: I normally used `ast::xxx` pattern because typing `ast::` will raise autocompletion from IDE.

IWANABETHATGUY:
> Just sharing: I normally used `ast::xxx` pattern because typing `ast::` will raise autocompletion from IDE.

ok, I just feel the `oxc::ast::ast::` is too verbose for each usage,  as you said, we could replace the ast as `oxc::ast::ast`, a good trade-off.

PR #2103, opened 2024-09-01
chore(ci): add autofix

### LINKED ISSUE #2003: Use `autofix.ci` to fix code issues.

I observed the usage of the `autofix.ci` command in `slidev`, which enables automatic resolution of certain easily fixable issues, such as formatting problems. By utilizing the `just fix` command, open-source projects can freely access the `autofix.ci` command to promptly address these issues without the need to repeatedly employ the `just fmt` command in pull requests.

Only requires configuration of a CI pipeline.

shulaoda:
### Description

closes #2003

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -1,0 +1,27 @@
+name: autofix.ci # needed to securely identify the workflow
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  autofix:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: ./.github/actions/setup-node
+
+      - uses: ./.github/actions/setup-rust
+        with:
+          tools: just, taplo-cli, cargo-shear
+          cache-key: debug-build
+          components: clippy rustfmt
+
+      - run: just fix
+
+      - uses: autofix-ci/action@dd55f44df8f7cdb7a6bf74c78677eb8acd40cd0a
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -1,9 +1,6 @@
 name: autofix.ci # needed to securely identify the workflow
 
 on:
-  push:
-    branches:
-      - main
   pull_request:
     branches:
       - main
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -122,20 +122,6 @@ jobs:
       os: ${{ matrix.target }}
       changed: ${{ needs.changes.outputs.node-changes == 'true' }}
 
-  node-validation:
```
hyf0: I think we still need this in case there are something wrong with autofix.

.github/workflows/autofix.yml
```diff
@@ -0,0 +1,27 @@
+name: autofix.ci # needed to securely identify the workflow
+
+on:
+  push:
+    branches:
+      - main
```
hyf0: Does autofix add extra commit? If so, it should not apply to main branch.

shulaoda: Let me experiment

.github/workflows/ci.yml
shulaoda: Do we still need to keep these?
https://github.com/rolldown/rolldown/blob/4c9feeb058f0212228e430450004967c99af9177/.github/workflows/ci.yml#L154-L158

IWANABETHATGUY: Format checking could be removed since we have autofix.

IWANABETHATGUY (approved):
LGTM, cc @hyf0

.github/workflows/ci.yml
hyf0: Emm. I'm not sure about this. I consider autofix is just a robot that executing `just fix`  for you. But no matter who  are executing `just fix` we still need to validate the fix.

.github/workflows/ci.yml
```diff
@@ -151,12 +151,6 @@ jobs:
       - name: Lint Spell
         run: pnpm lint-spell
 
-      - name: Lint Prettier
```
hyf0: Could you keep this? We still need these to validate if autofix is executing correctly.

shulaoda: yes

hyf0 (approved):
This is cool.

PR #2106, opened 2024-09-01
chore: add husky pre-commit

### LINKED ISSUE #2104: [Feature Request]: Why not use `husky` for `pre-commit` interception?

IMO, we can actually format or lint during the `git commit` stage, rather than heavily relying on the `CI/CD` pipeline. This way, we can avoid some unnecessary errors before the code is even committed.

shulaoda:
### Description

closes #2104

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+just fix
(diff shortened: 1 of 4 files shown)


IWANABETHATGUY:
This is not right, we should not run `just fix` if people only change documentation and should not run `rust-fix` if people only modify js files.  please refer https://github.com/lint-staged/lint-staged

PR #2107, opened 2024-09-01
perf(rolldown): use `itoa` for integer to string conversion

Boshen:
image

```rust
let i: u32 = 0;
format!("{i}");
```

is really costly in Rust as it involves heap allocation.

> This crate provides a fast conversion of integer primitives to decimal strings.
> The implementation comes straight from libcore but avoids the performance penalty of going through core::fmt::Formatter.

I haven't checked the rest of the codebase for such patterns, feel free to change them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2190,6 +2190,7 @@ dependencies = [
  "indexmap",
  "insta",
  "itertools 0.13.0",
+ "itoa",
  "memchr",
  "oxc",
  "regex",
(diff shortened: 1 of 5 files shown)


Boshen:
https://github.com/rolldown/rolldown/blob/8913103ac91a5256972ab58dbc639b5f5f935c2e/crates/rolldown/src/utils/renamer.rs#L137

This heap allocation is still costly, we should:

* improve the algorithm for less collisions, perhaps some form of short uuid?
* construct the string with `ComactString`, as most variables are really short.

---

The name is constantly being hashed as well https://github.com/rolldown/rolldown/blob/8913103ac91a5256972ab58dbc639b5f5f935c2e/crates/rolldown/src/utils/renamer.rs#L134

hyf0 (approved):

PR #2113, opened 2024-09-01
perf(rolldown): only recreate semantic data when ast is changed in pre_process_ecma_ast

Boshen:
Semantic data do not need to be recreated if ast is unchanged.

I intend to improve this situation later on, but this is a good first step.

* https://github.com/oxc-project/oxc/issues/5390
* https://github.com/oxc-project/backlog/issues/85

### DIFF
--- a/crates/rolldown/src/utils/pre_process_ecma_ast.rs
+++ b/crates/rolldown/src/utils/pre_process_ecma_ast.rs
@@ -27,6 +27,9 @@ pub fn pre_process_ecma_ast(
   replace_global_define_config: Option<&ReplaceGlobalDefinesConfig>,
   bundle_options: &NormalizedBundlerOptions,
 ) -> anyhow::Result<(EcmaAst, SymbolTable, ScopeTree)> {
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY (dismissed):

IWANABETHATGUY:
what about these two passes https://github.com/rolldown/rolldown/pull/2113/files#diff-0c67b3fa7a5959bb0aaaa3af2917786974449438d8eb43846b2d887c0cc5554cR107-R111, maybe they will affect semantic data?

Boshen:
> what about these two passes https://github.com/rolldown/rolldown/pull/2113/files#diff-0c67b3fa7a5959bb0aaaa3af2917786974449438d8eb43846b2d887c0cc5554cR107-R111, maybe they will affect semantic data?

Notice how `ast_changed` is only applied to the compressor right now.

hyf0 (approved):
Looks like positive. If there are further optimizations, let's do it in another PR.

PR #2121, opened 2024-09-02
feat(plugin/replace): objectGuards

ikkz:
Related issue: #2057. This pr implements the same function with [rollup's](https://github.com/rollup/plugins/blob/master/packages/replace/README.md#objectguards)

Additionally, the rollup's `objectGuards` doesn't take effects actually, even in the test case's [snapshot](https://github.com/rollup/plugins/blob/master/packages/replace/test/snapshots/form.js.md?plain=1#L46). It caused by the wrong implementation of function [`expandTypeofReplacements`](https://github.com/rollup/plugins/blob/master/packages/replace/src/index.js#L41). In this pr, I rewrite this function and I plan to fix it in rollup


hyf0 (approved):
Thanks!

PR #2122, opened 2024-09-02
perf: codegen in multiple chunks

IWANABETHATGUY:
### Description
1. split computation heavy task from `tokio join all ` , to avoid interference between tokio worker and rayon thread.
2. the performance gain dependents on number of chunks.  For `oxc_monitor` the build time is reduced from 
**3.5** to **2.8**, confirmed by @Boshen .
## Before
image
## After
image


crates/rolldown/src/stages/generate_stage/render_chunk_to_assets.rs
```diff
@@ -163,18 +165,49 @@ impl<'a> GenerateStage<'a> {
...
-      EcmaGenerator::instantiate_chunk(&mut ctx).await
-    }))
+    let chunk_to_codegen_ret = chunk_graph
+      .chunks
+      .raw
+      .par_iter()
+      .map(|item| {
+        item
+          .modules
+          .par_iter()
```
Boshen, line 229: 2 `par_iters` is really costly, try flatten them first in a single thread and then `part_iter` it?

Spawning threads is really slow :-/

IWANABETHATGUY: rayon will reuse the global thread pool,  the real cost part is sent to the work queue(it is trivial compared to ast codegen). I don't flatten them because you need to group by `chunkId`  them whatever at the end.

hyf0:
> he performance gain dependents on number of chunks. For oxc_monitor the build time is reduced from
3.5 to 2.8, @Boshen can confirm that.

This is nice!

Cargo.toml
```diff
@@ -167,7 +167,7 @@ oxc_transform_napi = { version = "0.26.0" }
 
 [profile.release]
 codegen-units = 1
-debug         = false     # Set to `true` for debug information
+debug         = 1     # Set to `true` for debug information
```
hyf0: Do we need to keep this?

IWANABETHATGUY: No,  I am profiling with some screenshot, hang on for a few seconds.

IWANABETHATGUY:
I merged it since oxc_monitor needs it.

PR #2126, opened 2024-09-02
refactor: remove pointless `Cow` in `Renamer`

overlookmotel:
### Description

`Renamer::used_canonical_names` was an `FxHashMap<Cow<'name, Rstr>, u32>`. But its keys are always `Cow::Owned`,  so the `Cow` wrapper serves no purpose and is pure overhead. Replace it with `FxHashMap<Rstr, u32>`.

I happened to come across this while looking at the renaming algorithm.


hyf0 (approved):
Thanks

PR #2134, opened 2024-09-03
chore(repo/vscode): ignored expanding `#[napi]` macro

hikomoon:
### Description

because vscode refuses to autocomplete/autosuggest code when using the #[napi] macro
use .vscode/setting.json file for fix napi-rs IDE support problem maybe more convenient for every forks.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -44,5 +44,6 @@
     "temp": true,
     "tmp": true
   },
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY:
I don't use vscode, so cc @hyf0

hyf0 (approved):
Thanks. This is quite useful for developing `rolldown_binding`.

PR #2140, opened 2024-09-03
chore: add husky pre-commit

### LINKED ISSUE #2104: [Feature Request]: Why not use `husky` for `pre-commit` interception?

shulaoda:
### Description

Revert of #2106, closes #2104


shulaoda:
@IWANABETHATGUY cc

IWANABETHATGUY:
I will have a look when I got some time.

IWANABETHATGUY:
This is not working when I change some rs files, 

```
✖ cargo fix --allow-dirty --allow-staged:
error: unexpected argument '~/Documents/rolldown-rs/rolldown/crates/rolldown/src/bundler.rs' found

Usage: cargo fix [OPTIONS]
```

IWANABETHATGUY:
Can you just remove all `rs` related lint-stage config if it does not work for `.rs` files?

shulaoda:
> Can you just remove all `rs` related lint-stage config if it does not work for `.rs` files?

How about this?

IWANABETHATGUY:
can you change this `lint-staged` to below? 
```
  "lint-staged": {
    "*.toml": "pnpx @taplo/cli format",
    "*.{js,ts,json,md,yml,yaml,vue}": [
      "prettier --write",
      "pnpm lint-code -- --fix"
    ]
  }
```

package.json
```diff
@@ -49,5 +50,13 @@
     "singleQuote": true,
     "trailingComma": "all",
     "arrowParens": "always"
+  },
+  "lint-staged": {
+    "*.rs": "rustfmt --edition 2021",
```
IWANABETHATGUY: We don't need format rs, since we have https://github.com/shulaoda/rolldown/blob/f8aca97e0fe01011524d34e3c82bba668ef0ecd4/.vscode/settings.json#L2-L5

shulaoda: You are right.

IWANABETHATGUY (approved):

PR #2153, opened 2024-09-04
fix: `process.env.NODE_ENV` inside paren expr

IWANABETHATGUY:


crates/rolldown/src/utils/pre_process_ecma_ast.rs
```diff
@@ -92,7 +92,12 @@ pub fn pre_process_ecma_ast(
     }
 
     // Perform dead code elimination.
-    let options = CompressOptions::dead_code_elimination();
+    let options = CompressOptions {
+      fold_constants: true,
+      remove_dead_code: true,
+      remove_syntax: true,
```
IWANABETHATGUY, line 98:
```
/// Remove syntax from the AST.
///
/// * Parenthesized Expression
/// * `debugger`
/// * `console.log`
```

hyf0 (approved):

crates/rolldown_ecmascript/src/ecma_ast/gen.rs
```diff
@@ -0,0 +1,16 @@
+use oxc::codegen::{self, CodeGenerator, Gen};
+
+pub trait ToSourceString {
```
IWANABETHATGUY, line 3: Used for convert any stmt or program to source_string, useful for debugging.

PR #2154, opened 2024-09-04
fix: mark knownGlobals global ident and global member access as side effects free

IWANABETHATGUY:
### Description
ref: 
https://github.com/evanw/esbuild/blob/9c13ae1f06dfa909eb4a53882e3b7e4216a503fe/internal/config/globals.go#L30-L850

https://github.com/evanw/esbuild/blob/9c13ae1f06dfa909eb4a53882e3b7e4216a503fe/internal/config/globals.go#L940-L961


crates/rolldown/tests/esbuild/dce/dce_of_using_declarations/artifacts.snap
```diff
@@ -19,8 +19,6 @@ await using await_asyncDispose_keep = { [Symbol.asyncDispose]() {
 } };
 using undef_keep = undefined;
 await using await_undef_keep = undefined;
-const Symbol_dispose_remove = Symbol.dispose;
-const Symbol_asyncDispose_remove = Symbol.asyncDispose;
 console.log(null_keep, undef_keep);
```
IWANABETHATGUY, line 23:
```
// entry.js
using null_keep = null;
await using await_null_keep = null;
using throw_keep = {};
using dispose_keep = { [Symbol.dispose]() {
  console.log("side effect");
} };
await using await_asyncDispose_keep = { [Symbol.asyncDispose]() {
  console.log("side effect");
} };
using undef_keep = void 0;
await using await_undef_keep = void 0;
console.log(
  null_keep,
  undef_keep
);
```
Still not same as esbuild, but not relevant to this pr, these two stmt is removed either in esbuild snapshot

hyf0 (approved):

PR #2162, opened 2024-09-04
fix(rolldown): side effect detector mark `+1` `-1` as number literal

IWANABETHATGUY:

### DIFF
--- a/crates/rolldown/src/ast_scanner/side_effect_detector/utils.rs
+++ b/crates/rolldown/src/ast_scanner/side_effect_detector/utils.rs
@@ -227,7 +227,7 @@ pub fn is_primitive_literal(scope: &AstScopes, expr: &Expression) -> bool {
     | Expression::NumericLiteral(_)
     | Expression::StringLiteral(_)
     | Expression::BigIntLiteral(_) => true,
-    // oxc number literal did not included `+1` and `-1`
+    // Include `+1` / `-1`.
     Expression::UnaryExpression(e)
       if matches!(e.operator, |UnaryOperator::UnaryNegation| UnaryOperator::UnaryPlus) =>
     {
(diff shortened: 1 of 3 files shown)


crates/rolldown/src/ast_scanner/side_effect_detector/utils.rs
```diff
@@ -227,6 +227,12 @@ pub fn is_primitive_literal(scope: &AstScopes, expr: &Expression) -> bool {
     | Expression::NumericLiteral(_)
     | Expression::StringLiteral(_)
     | Expression::BigIntLiteral(_) => true,
+    // oxc number literal did not included `+1` and `-1`
```
Boshen: It's not oxc ... The spec grammar is always an UnaryExpression.

crates/rolldown/src/ast_scanner/side_effect_detector/utils.rs
```diff
@@ -227,6 +227,12 @@ pub fn is_primitive_literal(scope: &AstScopes, expr: &Expression) -> bool {
...
+    Expression::UnaryExpression(e)
+      if matches!(e.operator, |UnaryOperator::UnaryNegation| UnaryOperator::UnaryPlus) =>
```
Boshen: Don't you need to check for number litral on `e.expression`?

crates/rolldown/src/ast_scanner/side_effect_detector/utils.rs
Boshen, line 230: Include `+1` / `-1`.

Boshen:
```suggestion
    // Include `+1` / `-1`.
```

crates/rolldown/src/ast_scanner/side_effect_detector/utils.rs
IWANABETHATGUY: you are right, I forgot that.

PR #2168, opened 2024-09-05
chore(rolldown): oxc v0.27.0

Boshen:
Test before publishing a new version.

### DIFF
--- a/crates/rolldown/src/utils/pre_process_ecma_ast.rs
+++ b/crates/rolldown/src/utils/pre_process_ecma_ast.rs
@@ -92,13 +92,8 @@ pub fn pre_process_ecma_ast(
     }
 
     // Perform dead code elimination.
-    let options = CompressOptions {
-      fold_constants: true,
-      remove_dead_code: true,
-      remove_syntax: true,
-      ..CompressOptions::all_false()
-    };
-    let compressor = Compressor::new(allocator, options);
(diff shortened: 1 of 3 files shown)


IWANABETHATGUY (approved):

crates/rolldown/src/utils/pre_process_ecma_ast.rs
```diff
@@ -92,13 +92,8 @@ pub fn pre_process_ecma_ast(
...
 
     // Perform dead code elimination.
-    let options = CompressOptions {
-      fold_constants: true,
-      remove_dead_code: true,
-      remove_syntax: true,
-      ..CompressOptions::all_false()
-    };
-    let compressor = Compressor::new(allocator, options);
+    // NOTE: `CompressOptions::dead_code_elimination` will remove `ParenthesizedExpression`s from the AST.
```
hyf0, line 95: Would it be problematic in the future? @IWANABETHATGUY cc

IWANABETHATGUY: No, they are literally the same thing.
https://github.com/oxc-project/oxc/blob/1bed5ce2a5fc89b3aa89fdf5e92ad8abc089e4a9/crates/oxc_minifier/src/options.rs#L95-L102

PR #2169, opened 2024-09-05
chore(ci): add workflow dispatch to CI action

inottn:
### Description
Considering the potential inconsistencies between the local and CI, allowing manual triggering of actions will make it more convenient to perform tests in forked repository.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ on:
     branches:
       - main
       - 'renovate/**' # For renovate bot "automerge": true
+  workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}


hyf0:
I'm open to this. Let me try to understand its usage. Where does you need to trigger the CI manually?

inottn:
> I'm open to this. Let me try to understand its usage. Where does you need to trigger the CI manually?

I sometimes encounter a situation where everything works fine locally, but the CI fails after creating a PR. So, before creating a PR, I merge the corresponding branch into the main branch and push it to trigger the CI, ensuring it passes before creating the PR. This process is quite cumbersome for me, as I need to revert the main branch when syncing upstream code. Therefore, I hope to be able to trigger the CI directly on the forked branch to simplify this workflow.

hyf0 (approved):
Don't see anything negative, so approved.

PR #2171, opened 2024-09-05
chore(rolldown): remove `console.log`

sxzz:
### Description

There's another `console.log` 😅


hyf0:
@Boshen Does oxc have lint on prevent from writing `console.log`?

hyf0 (approved):
Thanks

IWANABETHATGUY:
> @Boshen Does oxc have lint on prevent from writing `console.log`?

Yeah, but if we enabled the `no-console` we will get 54 errors, 
```
Finished in 13ms on 170 files with 107 rules using 32 threads.
Found 0 warnings and 54 errors.

```

PR #2182, opened 2024-09-06
chore: use `oxc::index` instead of `oxc_index` to reduce dependency in root `Cargo.toml`

IWANABETHATGUY:


hyf0:
This is on purpose to avoid building `rolldown_plugin` and it's dependents got blocked by building whole `oxc`. I don't have the data, if it doesn't affect the actually build time, I'm ok with using `oxc::index`.

IWANABETHATGUY:
I will investigate it when I got some time.

PR #2191, opened 2024-09-07
refactor: use `builder.void_zero`

IWANABETHATGUY:


IWANABETHATGUY:
skip review since it is just a simple reuse `oxc` builder api

PR #2196, opened 2024-09-09
fix(plugin): fix `expand_typeof_replacements`

AliceLanniste:
### Description

Related issue:#2057  . fix the function of `expand_typeof_replacements`

### DIFF
--- a/crates/rolldown_plugin_replace/src/utils.rs
+++ b/crates/rolldown_plugin_replace/src/utils.rs
@@ -10,23 +10,37 @@ static OBJECT_RE: LazyLock<Regex> = LazyLock::new(|| {
 pub(crate) fn expand_typeof_replacements(
   values: &HashMap<String, String>,
 ) -> HashMap<String, String> {
-  values
-    .keys()
-    .filter_map(|key| {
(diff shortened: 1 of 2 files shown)


hyf0 (approved):
Cool. Thank you!

hyf0:
CI failed.

ikkz:
@AliceLanniste Hi, I would like to further discuss this function. 
In your version, `process.env.NODE_ENV` will be expanded to `typeof process`, `typeof env`, but I think in actual use, the more common usage is as follows:
```
if (typeof process!== 'undefined' && typeof process.env === "object" && process.env.NODE_ENV === 'production') {
  console.log('production');
}
```
This will cause `typeof process.env === "object"` not to be replaced and `console.log('production');` won't be executed, which can also be seen in the snapshot of this pr. 

Secondly, taking your replacement result `typeof process!==` as an example, after constructing the replacement regular expression, the default delimiter will be added on both sides of this replacement: `\btypeof process!==\b(?!\.)`. In the above usage, this will not match because the equal sign is not followed by a word boundary.

Before:
image

Delimiters added:
image

PR #2202, opened 2024-09-10
perf(plugin_replace): set more reasonable number for regex `backtrack_limit`

Boshen:
Set `backtrack_limit` for `delimiters` and `lookahead` because they contain backtracking pattern `!?` and `*`.
Cannot set the number too low or it will be ignored by `fancy_regex::Error::RuntimeError` in `try_replace`.
Setting `backtrack_limit` to a large number will cause huge performance regression.
See <https://github.com/fancy-regex/fancy-regex/blob/main/PERFORMANCE.md#fancy-regex>.

---

Before:

24% program execution time

image

After:

It's gone 

image

### DIFF
--- a/crates/rolldown_plugin_replace/src/plugin.rs
+++ b/crates/rolldown_plugin_replace/src/plugin.rs
@@ -60,9 +60,11 @@ impl ReplacePlugin {
     let pattern = format!("{delimiter_left}({joined_keys}){delimiter_right}{lookahead}");
     Self {
       matcher: RegexBuilder::new(&pattern)
-        // Give a `usize::MAX` will cause bundle time tripled in some cases, so we need to use sensible limit
-        // to have a balance between performance and correctness.
-        .backtrack_limit(1_000_000)
+        // Set `backtrack_limit` for `delimiters` and `lookahead` because they contain backtracking pattern `!?` and `*`.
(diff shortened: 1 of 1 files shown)


crates/rolldown_plugin_replace/src/plugin.rs
```diff
@@ -60,9 +60,9 @@ impl ReplacePlugin {
     let pattern = format!("{delimiter_left}({joined_keys}){delimiter_right}{lookahead}");
     Self {
       matcher: RegexBuilder::new(&pattern)
-        // Give a `usize::MAX` will cause bundle time tripled in some cases, so we need to use sensible limit
-        // to have a balance between performance and correctness.
-        .backtrack_limit(1_000_000)
+        // Set a higher `backtrack_limit` for `lookahead` because it contains a backtracking pattern `*`.
+        // There are no backtracking patterns otherwise, setting to `10` just in case.
+        .backtrack_limit(if lookahead.is_empty() { 10 } else { 1000 })
```
hyf0: Probably a decent `1000` would be enough, since `lookahead` is set by default.

hyf0 (approved):

Boshen:
Interesting, test failed ... let me take a look at the test cases :-(

PR #2204, opened 2024-09-10
chore: remove unnecessary config file `rusty-hook.toml`

shulaoda:
### Description

Initially, we used `rusty-hook` to integrate with `pre-commit` (PR #5). However, due to some oversight, we accidentally removed it (PR #536), while the `rusty-hook.toml` configuration file was still retained. Maybe we can restore its use.

### DIFF
--- a/rusty-hook.toml
+++ b/rusty-hook.toml
@@ -1,5 +1,0 @@
-[hooks]
-pre-commit = "just check-typo && just fmt && just lint"
-
-[logging]
-verbose = true


hyf0:
With `husky` already enabled, I guess this cause execute `lint-staged` twice while committing. This is unnecessary.

>  while the rusty-hook.toml configuration file was still retained. 

So instead of enabling `rusty-hooks`, you could remove it in this PR.

shulaoda:
> With `husky` already enabled, I guess this cause execute `lint-staged` twice while committing. This is unnecessary.
> 
> > while the rusty-hook.toml configuration file was still retained.
> 
> So instead of enabling `rusty-hooks`, you could remove it in this PR.

They are mutually exclusive, so I removed `husky` and kept only `rusty-hook`. The only thing we need to confirm is which one we should keep.

hyf0:
Just keep using `husky` would be enough.

hyf0 (approved):

PR #2219, opened 2024-09-12
chore: udpate binding files

underfin:


IWANABETHATGUY:
https://github.com/rolldown/rolldown/blob/main/packages/rolldown/src/binding.d.ts#L548-L562, it seems binding already update to date.

IWANABETHATGUY:
I don't mean to close you pr, I just rebase the main and it is auto closed by github

PR #2221, opened 2024-09-12
fix: 🐛 invalid module repr name

IWANABETHATGUY:


crates/rolldown/tests/rolldown/misc/invalid_ident_repr/artifacts.snap
```diff
@@ -0,0 +1,25 @@
...
+
+```js
+import { default as assert } from "node:assert";
+
+
+//#region 1aaa.js
+var _aaa_ns = {};
+__export(_aaa_ns, { default: () => _aaa_default });
+const a$1 = "shared.js";
+var _aaa_default = a$1;
```
IWANABETHATGUY, line 16: Before this will generate `1aaa_default` which will cause runtime error

hyf0 (approved):

PR #2229, opened 2024-09-13
refactor(rolldown): change `pre_process_ecma_ast` to a struct

Boshen:
So we can keep and change some states.

PR #2228, opened 2024-09-13
chore(rolldown): oxc 0.29.0

Boshen:

PR #2230, opened 2024-09-13
perf(rolldown): use `SemanticBuilder::with_stats` when running semantic analysis on same AST twice

### LINKED ISSUE #2227: Perf: Use `SemanticBuilder::with_stats` when running semantic analysis on same AST twice
https://github.com/oxc-project/oxc/pull/5757 adds new APIs to `Semantic` and `SemanticBuilder` which are useful when running semantic analysis on same AST twice.

```rs
// First semantic run
let semantic = SemanticBuilder::new(source_text).build(program).semantic;
let stats = semantic.stats();

// Later on, running semantic again, pass the stats into `SemanticBuilder`
let semantic = SemanticBuilder::new(source_text)
    .with_stats(stats)
    .build(program)
    .semantic;
```

This avoids an AST traversal on 2nd semantic run. We see up to 12% perf improvement on large source files.

I *think* Rolldown does a 2nd semantic run sometimes.

Alternatively, if you combine multiple ASTs into 1 AST, and then run semantic on the end result, this may also be helpful:

```rs
let semantic1 = SemanticBuilder::new(source_text1).build(program1).semantic;
let stats1 = semantic1.stats();

let semantic2 = SemanticBuilder::new(source_text2).build(program2).semantic;
let stats2 = semantic2.stats();

let combined_program = combine_somehow(program1, program2);

let combined_stats = oxc_semantic::Stats::new(
    stats1.nodes + stats2.nodes,
    stats1.scopes + stats2.scopes,
    stats1.symbols + stats2.symbols,
    stats1.references + stats2.references,
);
let combined_semantic = SemanticBuilder::new("")
    .with_stats(combined_stats)
    .build(combined_program)
    .semantic;
```

Is that a pattern you'd use? If so, we can add an API to add 2 `Stats` together.

Boshen:
closes #2227

### DIFF
--- a/crates/rolldown/src/utils/pre_process_ecma_ast.rs
+++ b/crates/rolldown/src/utils/pre_process_ecma_ast.rs
@@ -5,7 +5,7 @@ use oxc::minifier::{
   CompressOptions, Compressor, InjectGlobalVariables, ReplaceGlobalDefines,
   ReplaceGlobalDefinesConfig,
 };
-use oxc::semantic::{ScopeTree, SemanticBuilder, SymbolTable};
+use oxc::semantic::{ScopeTree, SemanticBuilder, Stats, SymbolTable};
 use oxc::span::SourceType;
 use oxc::transformer::{TransformOptions, Transformer};
 
@@ -21,6 +21,9 @@ use super::tweak_ast_for_scanning::tweak_ast_for_scanning;
 pub struct PreProcessEcmaAst {
   /// Only recreate semantic data if ast is changed.
   ast_changed: bool,
+
+  /// Semantic statistics.
+  stats: Stats,
 }
 
 impl PreProcessEcmaAst {
@@ -45,6 +48,7 @@ impl PreProcessEcmaAst {
     // return Err(anyhow::anyhow!("Semantic Error: {:#?}", semantic_ret.errors));
     // }
 
+    self.stats = semantic_ret.semantic.stats();
     let (mut symbols, mut scopes) = semantic_ret.semantic.into_symbol_table_and_scope_tree();
 
     // Transform TypeScript and jsx.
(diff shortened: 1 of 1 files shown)


Boshen:
* **#2230** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2230?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2229** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2229?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2228** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2228?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

 Join @Boshen and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

overlookmotel:
Ah I now see that the transformer has run in between the 1st semantic run and the 2nd. This makes the stats from 1st semantic run less useful, as the counts will be too low - often more scopes/symbols/references are created in transform (JSX transform with `development: true` in particular creates tons of references).

Counts being too high is largely fine, but being even 1 too low is bad news, as then those `Vec`s will have to grow and reallocate.

We also at present have no way of getting an accurate count of nodes post-transform (we will once transformer adds `NodeId`s to nodes).

A workaround for now would be:

1. Get the updated scopes/symbols/references counts from `ScopeTree` and `SymbolTable` after all transforms are done.
2. Take a guess of how number of nodes may have increased since original semantic run.

```rs
// Do this *after* all transforms
let stats_for_2nd_semantic = Stats::new(
    // Very generous over-estimate on `nodes`
    nodes: self.stats.nodes * 2,
    scopes: scopes.len() as u32,
    symbols: symbols.len() as u32,
    references: symbols.references.len() as u32,
);
```

NB: Experimentation on https://github.com/oxc-project/oxc/pull/5703 and https://github.com/oxc-project/oxc/pull/5762 suggests that over-estimating counts is better than under-estimating.

Once https://github.com/oxc-project/oxc/pull/5762 is merged, should also use `with_excess_capacity` on the first semantic run, if AST is going to be transformed.

hyf0 (approved):

PR #2238, opened 2024-09-15
fix(filename_template): `[hash:6]` in `entryFileNames` does not get replaced with hash

### LINKED ISSUE #2206: [Bug]: `[hash:6]` in `entryFileNames` does not get replaced with hash

1. Open https://stackblitz.com/edit/github-6jajke-nva34g?file=rolldown.config.mjs
3. see `dist/entry-[hash:6].js` is generated
4. run `npm run build:up` to run rollup
5. see `dist/entry-q8VKMz.js` is generated

### What is expected?

`dist/entry-hash.js` (`hash` part can be any 6-character long hash) is generated

`dist/entry-[hash:6].js` is generated (`[hash:6]` is not replaced with the real hash)

### System Info

```Shell
Stackblitz

  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    rolldown: https://pkg.pr.new/rolldown@d6233d8 => 0.13.2
```

### Any additional comments?

This is not something urgent.

shulaoda:
### Description

closes #2206


crates/rolldown_common/src/inner_bundler_options/types/filename_template.rs
```diff
@@ -26,14 +29,27 @@ pub struct FileNameRenderOptions<'me> {
   pub ext: Option<&'me str>,
 }
 
+static HASH_PLACEHOLDER: LazyLock<Regex> =
+  LazyLock::new(|| Regex::new(r"\[hash(:(\d*))?]").expect("Invalid hash regex"));
```
Boshen: Similar to aa347f5, regex is unnecessary here.

IWANABETHATGUY:
only one test in js side is enough, since what you need to ensure is file name has `hash` in dist

crates/rolldown_common/src/inner_bundler_options/types/filename_template.rs
```diff
@@ -33,7 +34,7 @@ impl FilenameTemplate {
       tmp = tmp.replace("[name]", name);
     }
     if let Some(hash) = options.hash {
-      tmp = tmp.replace("[hash]", hash);
+      tmp = tmp.replace(options.hash_pattern.unwrap_or("[hash]"), hash);
```
7086cmd: I recommend completing the task in the file_emitter.rs file. For reference, the relevant code can be found [here](https://github.com/rolldown/rolldown/blob/main/crates/rolldown_common/src/file_emitter.rs#L86).

packages/rolldown/tests/fixtures/output/fileNames/hash/_config.ts
```diff
@@ -0,0 +1,25 @@
+import type { RolldownOutputChunk } from 'rolldown'
+import { defineTest } from '@tests'
+import { expect } from 'vitest'
+
+export default defineTest({
+  config: {
+    output: {
+      entryFileNames: '[name]-[hash:6].js',
+      chunkFileNames: '[name]-[hash:7].js',
```
7086cmd, line 9: What if we need more than 8?

7086cmd (changes requested):

packages/rolldown/tests/fixtures/output/fileNames/hash/_config.ts
shulaoda: We seem to support hash lengths of `5-22`.
https://github.com/rolldown/rolldown/blob/e52ef2f22015b919e958bdcf332bfca7425144d2/crates/rolldown/src/utils/hash_placeholder.rs#L44

shulaoda:
I have a question: 
Is it possible for there to be a situation where `[hash]-[hash:6].js` results in duplicate hashes? If so, there might be some errors with our current implementation.

packages/rolldown/tests/fixtures/output/fileNames/hash/_config.ts
7086cmd: Is [this](https://github.com/rolldown/rolldown/blob/main/crates/rolldown_common/src/file_emitter.rs#L86) `8` related?

shulaoda:
Both `esbuild` and `rollup` handle duplicate `[hash]-[hash]`, perhaps we should also support this feature.

packages/rolldown/tests/fixtures/output/fileNames/hash/_config.ts
shulaoda: Got it, I will improve.

IWANABETHATGUY:

crates/rolldown/src/utils/hash_placeholder.rs
```diff
@@ -22,16 +22,21 @@ const CHARS: &[u8] = b"0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTU
...
-    out_string.push(CHARS[current_digit as usize] as char);
+    index += 1;
+
     if value == 0 {
       break;
     }
   }
-  out_string
+
+  std::str::from_utf8(&buffer[..index]).unwrap().to_string()
```
IWANABETHATGUY: String::from_utf8_lossy(&a).to_string()

crates/rolldown/src/utils/hash_placeholder.rs
```diff
@@ -42,20 +47,20 @@ pub struct HashPlaceholderGenerator {
...
-        padding.push('0');
-      }
-      seed_base64 = [padding, seed_base64].concat();
-    }
+    let seed_base64 = to_base64(self.seed);
+
+    // TODO(hyf0): improve this
+    assert!(seed_base64.len() <= allow_middle_len, "seed is too large");
+
+    let mut placeholder = String::with_capacity(len + 5);
```
IWANABETHATGUY: pre alloc should consider all sub string length.

IWANABETHATGUY (approved):

PR #2242, opened 2024-09-16
feat(plugin/transformer): support `targets` key for options.

7086cmd:
The `oxc_transformer` supports parsing the options from the `targets` argument.

Incidentally, would it be feasible to consolidate these two `Transformer` instances into a single entity? In my opinion, this could lead to significant performance optimization.


hyf0 (approved):
Thanks

Issue #1554, opened 2024-07-09
Rollup APIs used by Vite


underfin:
Rolldown need to support some api for vite build.

- [x] build watch
- [x] https://github.com/rolldown/rolldown/issues/3500
- [x] [input.exports 'auto'](https://github.com/rolldown/rolldown/issues/1628)
- [x] #1728
- [x] output.name (library)
- [ ] output.hoistTransitiveImports (import code import performance)
- [x] https://github.com/rolldown/rolldown/issues/206
- [x] #1841
- [x] [output.format umd (ssr or library build )](https://github.com/rolldown/rolldown/pull/2489)
- [x] #2355
- [x] PluginContext#resolveId skipSelf
- [x] https://github.com/rolldown/rolldown/pull/769 ( src/node/plugins/index.ts:157:18  getSortedPluginsByHook)

IWANABETHATGUY:
It seems we already support ` PluginContext#load` and 
` PluginContext#resolveId skipSelf`?

7086cmd:
It appears that the `hoistTransitiveImports` option is currently set to false by default in Rolldown’s behavior (related: rollup/rollup#3332). Is it necessary to support the option being set to `true`?

sapphi-red:
I think we can close this issue now as rolldown-vite is not blocked by rolldown.

PR #2244, opened 2024-09-17
feat: support `output.generatedCode.symbols` and `constBindings`.

7086cmd:
**Vite-related PR:** https://github.com/vitejs/vite/pull/5018. It appears that only `constBindings` and `symbols` are impacted.

**Related issue:** https://github.com/rolldown/rolldown/issues/1554, option `output.generatedCode 2015`, also: https://github.com/rolldown/rolldown/issues/206.

Apart from these two and `reservedNamesAsProps`, it is advisable to utilize the `target` property through a transformer, and comment here: https://github.com/rolldown/rolldown/issues/206#issuecomment-2335111971.


7086cmd:
stale.

PR #2272, opened 2024-09-21
feat: support `css` module type in dataurl.

7086cmd:
Due to the absence of bundle CSS support, the availability of ESbuild tests remains undefined.

Let’s remove the unused functionality in `base64.rs`.


hyf0 (approved):
Thanks!

PR #2290, opened 2024-09-24
chore(esbuild-diff):  refactor with ts

IWANABETHATGUY:
### Description
1. for better  code completion and maintainability


IWANABETHATGUY:
Skip the review since it is just internal tooling, and just a refactor without logic change.

PR #2312, opened 2024-09-24
fix(moduleType/json): should trim the utf-8 bom.

### LINKED ISSUE #2297: [Bug]: error happens when JSON has UTF-8 BOM

1. Open https://stackblitz.com/edit/github-6jajke-ykjhyo?file=src%2Fobj.json
1. npm run build runs automatically.
1. The error happens

### What is expected?

No error happens.

The following error happens:

> Rolldown internal error: expected value at line 1 column 1

### System Info

```Shell
Stackblitz

  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    rolldown: https://pkg.pr.new/rolldown@351d2c7 => 0.13.2
```

### Any additional comments?

Faced this while trying rolldown with Vite.

7086cmd:
resolves #2297.


IWANABETHATGUY:
Can you add testcase in js side? The case only care about if we could handle `utfbom` nobody care about the snapshot, since the `utf8 bom` is invisible.

crates/rolldown_loader_utils/src/json_to_esm.rs
```diff
@@ -3,9 +3,13 @@ use rolldown_utils::ecma_script::is_validate_assignee_identifier_name;
 use serde_json::Value;
 // TODO: handling https://github.com/tc39/proposal-json-superset
 
+fn trim_bom(raw: &str) -> &str {
+  raw.strip_prefix("\u{FEFF}").unwrap_or(raw)
```
Boshen: Isn't this just https://doc.rust-lang.org/std/primitive.str.html#method.trim_start_matches

crates/rolldown_loader_utils/src/json_to_esm.rs
```diff
@@ -3,9 +3,13 @@ use rolldown_utils::ecma_script::is_validate_assignee_identifier_name;
 use serde_json::Value;
 // TODO: handling https://github.com/tc39/proposal-json-superset
 
+fn trim_bom(raw: &str) -> &str {
```
IWANABETHATGUY: NIT, inline this

IWANABETHATGUY (approved):

PR #2315, opened 2024-09-25
perf(plugin/replace): replace regex with manual iterator

overlookmotel:
### Description

In replace plugin, replace the regex used to split key into parts with hand-written iteration. The iterator validates the parts (replacing the function of the regex) and splits it into strings in one go. It has a fast path for ASCII (common case).


overlookmotel:
Discussed on Discord with @hyf0. Decision was that this is too much complexity for too little gain, as the regex is relatively simple and this code only runs once during plugin initialization, so it's not on a hot path. I'll open a separate PR which keeps the regex but removes some of the temporary allocations which this PR removed.

PR #2321, opened 2024-09-25
perf(plugin/replace): `expand_typeof_replacements` avoid temporary vecs

overlookmotel:
### Description

Pure refactor. Avoid temporary `Vec`s in `expand_typeof_replacements`.

Also avoid a pointless conversion of `Vec` to `HashMap` when that `HashMap` is immediately converted back to an iterator in the caller anyway.

PR #2319, opened 2024-09-25
test(plugin/replace): add more tests for `expand_typeof_replacements`

overlookmotel:
### Description

Just add more tests.

PR #2322, opened 2024-09-25
fix(plugin/replace): fix `object_guards` option

overlookmotel:
### Description

[@rollup/plugin-replace](https://www.npmjs.com/package/@rollup/plugin-replace)'s docs give this example for `objectGuards` option:

```js
replace({
  values: {
    'process.env.NODE_ENV': '"production"'
  }
});
```
```js
// Input
if (typeof process !== 'undefined' && process.env.NODE_ENV === 'production') {
  console.log('production');
}
// Without `objectGuards`
if (typeof process !== 'undefined' && 'production' === 'production') {
  console.log('production');
}
// With `objectGuards`
if ('object' !== 'undefined' && 'production' === 'production') {
  console.log('production');
}
```

Rolldown's version was not following this behavior. The example above was transformed to:

```js
if (typeof process !== "undefined" && true) {
  console.log("production");
}
```

This PR fixes that. The implementation now is almost identical to [Rollup's implementation](https://github.com/rollup/plugins/blob/master/packages/replace/src/index.js).

### DIFF
--- a/crates/rolldown_plugin_replace/tests/form/process_check/artifacts.snap
+++ b/crates/rolldown_plugin_replace/tests/form/process_check/artifacts.snap
@@ -8,7 +8,7 @@ source: crates/rolldown_testing/src/integration_test.rs
 ```js
 
 //#region input.js
-if (typeof process !== "undefined" && typeof process.env === "object" && true) {
+{
 	console.log("production");
 }
 
--- a/crates/rolldown_plugin_replace/src/utils.rs
+++ b/crates/rolldown_plugin_replace/src/utils.rs
@@ -15,21 +15,10 @@ pub(crate) fn expand_typeof_replacements(
   for key in values.keys() {
     if OBJECT_RE.is_match(key) {
       // Skip last part
-      let mut parts = key.split('.');
-      let mut part = parts.next().unwrap();
-      for next in parts {
-        replacements.extend([
-          (format!("typeof {part} ==="), "\"object\" ===".to_string()),
-          (format!("typeof {part}==="), "\"object\"===".to_string()),
-          (format!("typeof {part} !=="), "\"object\" !==".to_string()),
-          (format!("typeof {part}!=="), "\"object\"!==".to_string()),
-          (format!("typeof {part} =="), "\"object\" ===".to_string()),
-          (format!("typeof {part}=="), "\"object\"===".to_string()),
-          (format!("typeof {part}!="), "\"object\"!==".to_string()),
-          (format!("typeof {part} !="), "\"object\" !==".to_string()),
-        ]);
-        part = next;
-      }
+      replacements.extend(key.match_indices('.').map(|(index, _)| {
+        let match_str = &key[..index];
+        (format!("typeof {match_str}"), "\"object\"".to_string())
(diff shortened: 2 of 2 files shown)


overlookmotel:
* **#2322** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2322?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2321** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2321?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2320** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2320?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2319** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2319?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2318** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2318?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

 Join @overlookmotel and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

overlookmotel:
Is there a reason why Rolldown was generating a much more complex regex for `typeof` than Rollup? I assume this must have been intentional, but I can't see why.

Second question: Do the benchmarks include the replace plugin? Or are the differences on benchmarks in this stack just noise?

crates/rolldown_plugin_replace/tests/form/process_check/artifacts.snap
```diff
@@ -8,7 +8,7 @@ source: crates/rolldown_testing/src/integration_test.rs
 ```js
 
 //#region input.js
-if (typeof process !== "undefined" && typeof process.env === "object" && true) {
+{
```
overlookmotel, line 11: It would be better to disable DCE for this test, so can see the actual output from replace plugin, but I couldn't figure our how to do that. In Rollup, `treeshake: false` does that, but `treeshake: TreeshakeOptions::Boolean(false)` seemed to have no effect.

hyf0 (approved):

crates/rolldown_plugin_replace/tests/form/process_check/artifacts.snap
hyf0: @IWANABETHATGUY cc

hyf0:
### Merge activity

* **Sep 25, 1:30 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/rolldown/rolldown/2322) rebased this pull request after merging its parent, because this pull request is set to merge when ready.

crates/rolldown_plugin_replace/tests/form/process_check/artifacts.snap
IWANABETHATGUY: Ok then we should align with rollup

overlookmotel: Thanks for doing that (#2335).

PR #2324, opened 2024-09-25
chore: update esbuild diff 

IWANABETHATGUY:
### Description
1. rest changes are different introduced by  #2314


crates/rolldown/tests/esbuild/importstar/namespace_import_re_export_star_unused_missing_es6/entry.js
```diff
@@ -1,5 +1,3 @@
 import assert from 'node:assert'
 import * as ns from './foo'
-assert.deepEqual(ns, {
-  x: 123
-})
+assert.deepEqual(ns.foo, void 0)
```
IWANABETHATGUY, line 3: align with esbuild https://github.com/evanw/esbuild/blob/9c13ae1f06dfa909eb4a53882e3b7e4216a503fe/internal/bundler_tests/bundler_importstar_test.go#L1294-L1296

PR #2325, opened 2024-09-25
ci: check esbuild tests diff in ci

IWANABETHATGUY:


hyf0:
I'm not sure about this. Should we track diff for every PR?

IWANABETHATGUY:
> I'm not sure about this. Should we track diff for every PR?

any other thoughts?

hyf0:
Does `ued` only appied for tests of esbuild or all tests? Let's have a quick chat later.

PR #2331, opened 2024-09-26
fix(esbuild-diff): remove temp var when exported variable

IWANABETHATGUY:


crates/rolldown/tests/esbuild/importstar/export_other_nested_common_js/artifacts.snap
```diff
@@ -18,11 +18,10 @@ var require_foo = __commonJSMin((exports) => {
 var import_foo = __toESM(require_foo());
 
 //#endregion
-var y = import_foo.x;
```
IWANABETHATGUY, line 21: align with esbuild

crates/rolldown/tests/esbuild/importstar/export_other_common_js/artifacts.snap
```diff
@@ -18,11 +18,10 @@ var require_foo = __commonJSMin((exports) => {
 var import_foo = __toESM(require_foo());
 
 //#endregion
-var bar = import_foo.bar;
```
IWANABETHATGUY, line 21: This var only used when reexport, inline it will generate less  code, and also align with esbuild.

IWANABETHATGUY: here is a example, if we use `bar` and then reexoprt it, 

```js
import {bar} from './foo'

console.log(`bar: `, bar)
export {bar}
```
here is the result: 

```js

//#region foo.js
var require_foo = __commonJSMin((exports) => {
	exports.foo = 123;
});

//#endregion
//#region entry.js
console.log(`bar: `, import_foo.bar); // you could see here, it still use  `import.bar`

// snip
```

```

hyf0 (approved):

PR #2335, opened 2024-09-26
fix:  dicable dce when disable tree shake

IWANABETHATGUY:
### Description
              It would be better to disable DCE for this test, so can see the actual output from replace plugin, but I couldn't figure our how to do that. In Rollup, `treeshake: false` does that, but `treeshake: TreeshakeOptions::Boolean(false)` seemed to have no effect.

_Originally posted by @overlookmotel in https://github.com/rolldown/rolldown/pull/2322#discussion_r1775296379_

### DIFF
--- a/crates/rolldown/tests/esbuild/default/var_relocating_bundle/_config.json
+++ b/crates/rolldown/tests/esbuild/default/var_relocating_bundle/_config.json
@@ -21,7 +21,6 @@
         "name": "function-nested_js",
         "import": "function-nested.js"
       }
-    ],
-    "treeshake": false
+    ]
   }
 }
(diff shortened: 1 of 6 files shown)


hyf0 (approved):

crates/rolldown/tests/esbuild/default/var_relocating_bundle/_config.json
```diff
@@ -21,7 +21,6 @@
         "name": "function-nested_js",
         "import": "function-nested.js"
       }
-    ],
-    "treeshake": false
```
IWANABETHATGUY, line 25: esbuild did not disable tree shake in this suite

IWANABETHATGUY: https://github.com/evanw/esbuild/blob/d34e79e2a998c21bb71d57b92b0017ca11756912/internal/bundler_tests/bundler_default_test.go#L6333-L6344

PR #2336, opened 2024-09-26
refactor: avoid clone asset source at emit asset to bundle

underfin:


hyf0 (dismissed):

hyf0 (approved):
Might affects rebuild, let's ignore it for now.

Issue #2330, opened 2024-09-26
[Bug]: investigate memory alloc failed in https://github.com/rolldown/rolldown/actions/runs/11046049843/job/30684736322


IWANABETHATGUY:
### Reproduction link or steps

n/a

### What is expected?

Don't panic

### What is actually happening?

panic

### System Info

```Shell
https://github.com/rolldown/rolldown/actions/runs/11046049843/job/30684736322
```

IWANABETHATGUY:
closed in favor of #2407 and vitest update

PR #2338, opened 2024-09-26
chore: node test in single thread

IWANABETHATGUY:
### Description
ref: 
https://github.com/vitest-dev/vitest/pull/5047
try to run every test in sequential
related to #2330


IWANABETHATGUY:
did not help, closed

PR #2339, opened 2024-09-26
chore(esbuild-diff): rewrite `esbuild` `__commonJS` to `__commonJSMin`

IWANABETHATGUY:


IWANABETHATGUY:
Since it not affect the core, just related to esbuild diff tool

IWANABETHATGUY:
skip flaky ci of alloc bytes

PR #2342, opened 2024-09-26
chore(esbuild-diff): enable missing imporstar tests

IWANABETHATGUY:
### Description
1. enable importstar tests case that ignore by generator.


IWANABETHATGUY:
skip review since only enabled missing test cases, and did not affect core

PR #2343, opened 2024-09-26
chore(esbuild-diff): bypass esbuild diff.md

IWANABETHATGUY:
### Description
1. add bypass mechanism to bypass some test case that semantic and runtime behavior same as esbuild, but it's impossible to ignore noise like
    - deconflict order, we starts with `root` and esbuild starts with `leaf` module
    -  esbuild cjs interop can't be recognized by cjs-module-lexer, our impl is compatible with it.


IWANABETHATGUY:
skip review since it is about esbuild diff tool, not affect core.

PR #2345, opened 2024-09-27
feat: add hmr module loader

underfin:
### Description

The `HmrModuleLoader` is used to fetch changed modules at hmr, it has some difference with `ModuleLoader`.

- If the module is already exited at the pervious module table, it is only replaced, the `ModuleIdx` is not changed.
- If the module is new one, here alloc new  `ModuleIdx`  to it
- If the module is removed, here is not deal it, because it need to changed `ModuleTable#Module` to `Option`, it is a larger change at now.
- If some errors happened,it shouldn't exit process.

### DIFF
--- a/crates/rolldown/src/bundler.rs
+++ b/crates/rolldown/src/bundler.rs
@@ -6,17 +6,21 @@ use super::stages::{
 };
 use crate::{
   bundler_builder::BundlerBuilder,
+  module_loader::hmr_module_loader::HmrModuleLoader,
   stages::{generate_stage::GenerateStage, scan_stage::ScanStage},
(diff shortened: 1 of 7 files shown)


hyf0:
Let's not be hurry to merge this to main. There are the reasons:
- It's all about future-vite, and it's far from now.
- No manpower/plan put on ensuring correctness of this code during development of other features, since it's not a high priority thing.
- There will be planned refactor on related code(eg module loader). We don't want something future block the current development.

 However, This PR could be a good reference to show how the HMR works in the current codebase. Let's keep this PR open  as a draft.

PR #2346, opened 2024-09-27
feat: conditional `commonJSMin` and `esmMin`

### LINKED ISSUE #86: How to apply __esmMin or __esm runtime symbol
Esbuil provide `__esmMin` which using production and `__esm` which using development, it is controlled by `MinifyIdentifiers` option. But it is need discuss how to control at rolldown. Here has two ways to control it.

- Add similar option to do it, maybe `MinifyIdentifiers`
- Add mode which value be `production` or `development`

IWANABETHATGUY:
### Description
Closed https://github.com/rolldown/rolldown/issues/86


crates/rolldown/tests/rolldown/misc/common_js_min/artifacts.snap
```diff
@@ -0,0 +1,10 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+---
+# Assets
+
+## main.mjs
+
+```js
+var __create=Object.create,__defProp=Object.defineProperty,__getOwnPropDesc=Object.getOwnPropertyDescriptor,__getOwnPropNames=Object.getOwnPropertyNames,__getProtoOf=Object.getPrototypeOf,__hasOwnProp=Object.prototype.hasOwnProperty,__commonJSMin=(l,m)=>()=>(m||l((m={exports:{}}).exports,m),m.exports),__copyProps=(l,m,n,o)=>{if(m&&typeof m==="object"||typeof m==="function")for(var p=__getOwnPropNames(m),q=0,r=p.length,s;q<r;q++){s=p[q];if(!__hasOwnProp.call(l,s)&&s!==n)__defProp(l,s,{get:((t)=>m[t]).bind(null,s),enumerable:!(o=__getOwnPropDesc(m,s))||o.enumerable})}return l},__toESM=(l,m,n)=>(n=l!=null?__create(__getProtoOf(l)):{},__copyProps(m||!l||!l.__esModule?__defProp(n,"default",{value:l,enumerable:!0}):n,l)),require_shared=__commonJSMin((exports,m)=>{m.exports=123}),import_shared=__toESM(require_shared());assert.equal(import_shared.foo,123);
```
IWANABETHATGUY, line 9: You could see the `commonJsMin` is used when minify is enable.

hyf0 (approved):

PR #2348, opened 2024-09-28
test(esbuild): enable `loader_data_url_unknown_mime`.

7086cmd:


crates/rolldown/tests/esbuild/loader/loader_data_url_unknown_mime/_config.json
```diff
@@ -7,5 +7,5 @@
       }
     ]
   },
-  "expectError": true
+  "expectExecuted": false
```
7086cmd, line 10: In contrast, esbuild does not generate an error, but rather executes an error.

7086cmd (commented):
I have already executed the command `just eud`.

hyf0 (approved):
Thanks

PR #2349, opened 2024-09-28
test(esbuild): enable more module type tests.

7086cmd:
I neglected to enable this feature in the previous instance. Additionally, the `just ued` command is also executed.


IWANABETHATGUY:
`just ued` will not gonna work, since it only generate snap in `esbuild/importstar` dir

IWANABETHATGUY (approved):

PR #2358, opened 2024-09-29
feat: warning diagnostic import-is-undefined

IWANABETHATGUY:
### Description
1. ref https://esbuild.github.io/try/#YgAwLjIwLjIALS1idW5kbGUgLS1taW5pZnk9ZmFsc2UgLS1mb3JtYXQ9ZXNtAGUAZW50cnkuanMAaW1wb3J0ICogYXMgbnMgZnJvbSAnLi9mb28uanMnCgpjb25zb2xlLmxvZyhucy5kZWZhdWx0KQoKAABmb28uanMALy8gZm9vLmpzCmV4cG9ydCBjb25zdCBhID0gMTAwOw


crates/rolldown/tests/esbuild/importstar/import_default_namespace_combo_issue446/artifacts.snap
```diff
@@ -38,6 +38,30 @@ source: crates/rolldown_testing/src/integration_test.rs
 ```text
 [UNRESOLVED_IMPORT] Warning: "external" is imported by "external-ns.js", but could not be resolved – treating it as an external dependency.
 
+```
+## IMPORT_IS_UNDEFINED
+
+```text
+[IMPORT_IS_UNDEFINED] Warning: Import `def` will always be undefined because there is no matching export in 'internal.js'
```
IWANABETHATGUY: Currently our diagnostic is sub-optimal compared to `esbuild` since we don't store the `span` for each `props` so we could not highlight the target prop accurately, for most scenarios member expr prop will not duplicate.

hyf0 (approved):

PR #2365, opened 2024-10-01
feat(rolldown): add meta for plugins

sxzz:
### Description

Add missing `meta` property and additionally add `rolldownVersion`.


hyf0 (approved):
Thanks

PR #2368, opened 2024-10-04
chore: enable esbuild default tests

IWANABETHATGUY:
### Description
All about the pr is just enabling tests of `esbuild default`

image
you probably need to review it locally.


hyf0 (approved):

PengBoUESTC:
I notice that some of `just ued` results throw errors, because these source scripts are compiled to  `mjs` but do not conform to `strict mode` standards. Is this WIP? @IWANABETHATGUY 

`
with_statement_tainting_no_bundle
err:  SyntaxError: 'with' in strict mode (5:2)
`

PR #2372, opened 2024-10-04
feat(rust): add `rolldown_std_utils` crate to store utils/extensions for the rust std library

hyf0:
### Description

`rolldown_utils` is kind of swell. We need a crate to store utils code that is `std` only. These code would be used widely, so it would be better don't put much third-party dependecies in `rolldown_std_utils` as `rolldown_utils` did.


hyf0:
* **#2372** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2372?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/rolldown_std_utils/src/pretty_type_name.rs
```diff
@@ -0,0 +1,30 @@
+use std::borrow::Cow;
+
+use regex::Regex;
+use std::sync::LazyLock;
+
+static MODULE_MATCHER_RE: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"(?:\w+::)").unwrap());
```
hyf0, line 6: I wonder if we could remove the `regex` dependency and still print the type name in a pretty way.

IWANABETHATGUY (approved):

PR #2393, opened 2024-10-09
fix: `advancedChunks.groups[*].test` should be a `RegExp`

### LINKED ISSUE #2391: [Bug]: `advancedChunks.groups[*].test` expects a string instead of a RegExp

https://github.com/rolldown/rolldown/blob/7bc51f099a916dbe31bc0582995c58cf0d0f8924/packages/rolldown/src/options/output-options.ts#L93
https://github.com/rolldown/rolldown/blob/7bc51f099a916dbe31bc0582995c58cf0d0f8924/crates/rolldown_binding/src/utils/normalize_binding_options.rs#L193-L195

### What is expected?

Allow passing a RegExp and disallow passing a string.
```ts
advancedChunks: {
  groups: [
    {
      name: 'chunk',
      test: /chunk.css/,
    },
  ],
},
```
This is more intuitive.
I think it can allow strings if it works like `id.includes(group.test)` instead of `new RegExp(group.test).test(id)`.

This is disallowed
```ts
advancedChunks: {
  groups: [
    {
      name: 'chunk',
      test: /chunk.css/,
    },
  ],
},
```
and this is allowed.
```ts
advancedChunks: {
  groups: [
    {
      name: 'chunk',
      test: 'chunk.css',
    },
  ],
},
```

### System Info

```Shell
System:
    OS: Windows 11 10.0.22631
    CPU: (24) x64 AMD Ryzen 9 5900X 12-Core Processor
    Memory: 17.20 GB / 31.92 GB
  Binaries:
    Node: 20.17.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.8.2 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.12.0 - C:\Program Files\nodejs\pnpm.CMD
    bun: 1.1.29 - ~\AppData\Local\Microsoft\WinGet\Links\bun.EXE
  Browsers:
    Edge: Chromium (127.0.2651.74)
    Internet Explorer: 11.0.22621.3527
  npmPackages:
    rolldown: https://pkg.pr.new/rolldown@9a0db67 => 0.13.2
```

### Any additional comments?

_No response_

shulaoda:
### Description

closes #2391 

This requires some modifications to struct `BindingStringOrRegex`, which will take some time

### DIFF
--- a/crates/rolldown_binding/src/options/binding_output_options/types/binding_advanced_chunks_options.rs
+++ b/crates/rolldown_binding/src/options/binding_output_options/types/binding_advanced_chunks_options.rs
@@ -1,8 +1,7 @@
 use derivative::Derivative;
-use napi::Either;
 use serde::Deserialize;
 
-use crate::options::plugin::types::binding_js_or_regex::JsRegExp;
+use crate::options::plugin::types::binding_js_or_regex::BindingStringOrRegex;
 
 #[napi_derive::napi(object, object_to_js = false)]
 #[derive(Deserialize, Derivative)]
@@ -20,9 +19,7 @@ pub struct BindingAdvancedChunksOptions {
 #[derivative(Debug)]
 pub struct BindingMatchGroup {
   pub name: String,
-  #[napi(ts_type = "RegExp | String")]
-  #[serde(skip_deserializing)]
-  pub test: Option<Either<JsRegExp, String>>,
+  pub test: Option<BindingStringOrRegex>,
   // pub share_count: Option<u32>,
   pub priority: Option<u32>,
   pub min_size: Option<f64>,
--- a/crates/rolldown_binding/src/options/binding_output_options/types/binding_advanced_chunks_options.rs
+++ b/crates/rolldown_binding/src/options/binding_output_options/types/binding_advanced_chunks_options.rs
@@ -1,5 +1,8 @@
 use derivative::Derivative;
+use napi::Either;
 use serde::Deserialize;
+
+use crate::options::plugin::types::binding_js_or_regex::JsRegExp;
 
 #[napi_derive::napi(object, object_to_js = false)]
 #[derive(Deserialize, Derivative)]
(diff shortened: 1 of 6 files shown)


shulaoda:
I have to set the type of test to `String | RegExp` because it is used in some test cases' `_config.json` in crates.

shulaoda:
Perhaps we can change the `BindingStringOrRegex` to `Either<String, JsRegExp>`, so that we don't need to perform additional processing on it on the js side. @IWANABETHATGUY cc

IWANABETHATGUY:
> What is expected?
> Allow passing a RegExp and disallow passing a string.

> advancedChunks: {
>   groups: [
 >  {
 >     name: 'chunk',
>      test: /chunk.css/,
>    },
>  ],
> },

It seems this pr did not follow the expected, I leave  @hyf0 and @sapphi-red to ensure if we need to support string

shulaoda:
> I have to set the type of test to `String | RegExp` because it is used in some test cases' `_config.json` in crates.

I'm not sure if we support the `_config.ts` configuration format in the `crates`. if we do, we can remove the usage of `advancedChunks.groups` from `_config.json`.

IWANABETHATGUY:
> Perhaps we can change the `BindingStringOrRegex` to `Either<String, JsRegExp>`, so that we don't need to perform additional processing on it on the js side. @IWANABETHATGUY cc

you could open another pr.

sapphi-red:
I think it's fine to allow strings if it works like `.contains` like we have in `filter` options of `transform` hook. I said "string should be disallowed" because I thought that would be easier to implement than supporting it.

burhanuday:
How about validating if the regex is valid using `refine` from zod?
Something like
```js
         test: z.string().refine((v) => {
              // Test if the string is a valid RegExp
              try {
                new RegExp(v)
                return true
              } catch (e) {
                return false
              }
            }),
```

shulaoda:
> How about validating if the regex is valid using `refine` from zod? Something like

I tried it, but it doesn't seem to work on the ts side.

hyf0:
> I think it's fine to allow strings if it works like `.contains` like we have in `filter` options of `transform` hook. I said "string should be disallowed" because I thought that would be easier to implement than supporting it.

Yeah. Supporting string would be more performant than regex since it only needs to do linear matching.

hyf0:
Looks like we need a proper  `BindingStringOrRegex` while writing binding. I'll take the PR from here.

hyf0:
I'm gonna merge this PR first and try to improve it in the future.

hyf0 (approved):

PR #2397, opened 2024-10-09
chore: enable dce esbuild tests

IWANABETHATGUY:


IWANABETHATGUY:
skip review since it is not related to core.

PR #2398, opened 2024-10-09
chore: allow `diff.md` add reason section

IWANABETHATGUY:
### Description
1. This allows us to aggregate failed reasons programmatically.


IWANABETHATGUY:
Skip review since it is not related to core.

PR #2402, opened 2024-10-10
chore: use .js extension in rust tests output

IWANABETHATGUY:
### Description
1. Don't added `.cjs` and `.mjs` in rust tests
2. Inserted a dummy `package.json` with `type` to ensure `import` and `export` could be used as before
3. related to esbuild test diff


IWANABETHATGUY:
Skip review since it is not related to core, and we have discussed before.

PR #2406, opened 2024-10-10
feat: support more than one instance of hash in file names

### LINKED ISSUE #2246: Filename supports duplicate hash like `[hash]-[hash:6].js`

burhanuday:
### Description

Closes #2246


shulaoda:
You need to consider the case like `[hash]-[hash:6]-[hash:8].js`.

Additionally, the `hash_placeholder` below looks like `!~\\{[0-9a-zA-Z_$]{1,17}\\}~`, so simply slicing the hash based on size will not work.
https://github.com/rolldown/rolldown/blob/c5b837963dda7d71444e8c70322d4e30d3e3536a/crates/rolldown/src/stages/generate_stage/mod.rs#L227-L231

burhanuday:
> Additionally, the hash_placeholder below looks like !~\\{[0-9a-zA-Z_$]{1,17}\\}~, so simply slicing the hash based on size will not work.

I did not understand this part. Why will slicing by the number of characters not work?

shulaoda:
> I did not understand this part. Why will slicing by the number of characters not work?

Firstly, when replacing the hash in the chunk name, `[hash]-[hash:5]-[hash:6].js` will be replaced with a `preliminary_name` like `!~{abcdefgh}~-!~{abcdefgh}~-!~{abcdefgh}~.js`.
https://github.com/rolldown/rolldown/blob/c5b837963dda7d71444e8c70322d4e30d3e3536a/crates/rolldown/src/stages/generate_stage/mod.rs#L227-L231

During the final output of the chunk, a series of conflict resolution steps will be performed (I am not fully familiar with this part of the code yet). This process will replace the `!~{abcdefgh}~` in the `preliminary_name` (the content might change due to conflicts) with `abcdefgh`, resulting in `abcdefgh-abcdefgh-abcdefgh.js`.
https://github.com/rolldown/rolldown/blob/c5b837963dda7d71444e8c70322d4e30d3e3536a/crates/rolldown/src/utils/chunk/finalize_chunks.rs#L107-L114

So, by truncating the hash to achieve a specific length, you might alter the structure of `!~{abcdefgh}~`.

Of course, you can also check if the test snapshot has changed correctly by executing `just test`.

PR #2409, opened 2024-10-12
chore: enable all esbuild ts tests

IWANABETHATGUY:


IWANABETHATGUY:
skip review, since this only related to tests

PR #2411, opened 2024-10-12
chore: add step to check for minimum node version during setup

burhanuday:
### Description

Using a NodeJS version less than what is in the .node-version file gives a cryptic error, see screenshot below

Screenshot 2024-10-12 at 10 57 21 AM

This PR adds a check if the user has the minimum required version installed

When minimum version is not installed, the script errors out early

Screenshot 2024-10-12 at 10 56 41 AM

When the installed version is above the minimum, the script works as expected

Screenshot 2024-10-12 at 10 57 50 AM

### DIFF
--- a/scripts/misc/setup-prerequisites/node.js
+++ b/scripts/misc/setup-prerequisites/node.js
@@ -1,6 +1,10 @@
-const MIN_MAJOR_VERSION = 20
-const MIN_MINOR_VERSION = 14
-const MIN_PATCH_VERSION = 0
+import { readFileSync } from 'node:fs'
+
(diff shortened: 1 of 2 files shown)


scripts/misc/setup-prerequisites/node.js
```diff
@@ -0,0 +1,17 @@
+const MIN_MAJOR_VERSION = 20
```
hyf0: Could we create this data by reading from `.node-version`?

hyf0 (approved):
Thanks

burhanuday:
Are there any blockers to merging this? @hyf0

hyf0:
Nope. Just forgot to click the auto merge button.

PR #2415, opened 2024-10-12
chore: esbuild testcase multi entrypoint mapping

IWANABETHATGUY:
### Description
image
1.


IWANABETHATGUY:
skip review since it is related to esbuild tests and don't affect core.

PR #2416, opened 2024-10-12
chore: esbuild tests import_star reason

IWANABETHATGUY:
### Description
1. Adding diff reason for each case undering `esbuild/import_star`


IWANABETHATGUY:
skip review, since it is related to tests, and not affect core.
image

PR #2425, opened 2024-10-14
chore: esbuild tests default reason

IWANABETHATGUY:


IWANABETHATGUY:
skip reviewing, since it is related to esbuild tests part, just add reason to `diff.md`

PR #2427, opened 2024-10-14
chore: add reason for esbuild tests `importstar_ts`

IWANABETHATGUY:


IWANABETHATGUY:
skip review since it is not related to core, only tests

PR #2429, opened 2024-10-14
chore: add esbuild tests compatibility stats

IWANABETHATGUY:


IWANABETHATGUY:
skip review since it is not related to core, only generates `stats.md`

PR #2433, opened 2024-10-15
refactor(esbuild-diff): use remark to parse markdown

IWANABETHATGUY:
### Description
1. `marked-it` only generates `tokens` which is not friendly for analyzing.


IWANABETHATGUY:
skip review since it related to `esbuild -diff` tool
image

PR #2441, opened 2024-10-15
fix: 🐛 crossplatform glob expr

IWANABETHATGUY:


IWANABETHATGUY:
image
This is a hotfix, so skip review, also it is not related to core.

PR #2454, opened 2024-10-17
feat:  add jsx related options in rust side

IWANABETHATGUY:
### Description
1. Adding `jsx` related option in rust test side
2. update some test configuration under esbuild.
3. passing some esbuild tests


crates/rolldown_common/src/inner_bundler_options/mod.rs
```diff
@@ -135,7 +135,11 @@ pub struct BundlerOptions {
   pub inline_dynamic_imports: Option<bool>,
   pub advanced_chunks: Option<AdvancedChunksOptions>,
   pub checks: Option<ChecksOptions>,
-  #[cfg_attr(feature = "deserialize_bundler_options", serde(default), schemars(skip))]
+  #[cfg_attr(
+    feature = "deserialize_bundler_options",
+    serde(deserialize_with = "deserialize_jsx", default),
+    schemars(with = "Option<HashMap<String, String>>")
```
hyf0, line 141: Could we have a struct instead of this? Writing a custom deserialize function seems not worth and not easy to maintain.

hyf0: See https://serde.rs/remote-derive.html

hyf0: I also wonder if it's a good idea to expose oxc options directly in rust. This will make rolldown's rust user need to both install oxc and rolldown to get the `JsxOptions` struct. And if the user's oxc version is different from the one that rolldown uses, there might be dependency resolution problem.

Writing a new `JsxOptions` in rolldown itself is indeed cumbersome but I think it's a long-term-friendly thing to do.

IWANABETHATGUY: yeah, sounds good, but it may be out of the scope of this pr, could be polished in the future.

IWANABETHATGUY: I think it is worth doing, but at least wait for oxc jsx lowering is stable.

hyf0 (approved):

PR #2455, opened 2024-10-17
fix: data url guess mime by extension

IWANABETHATGUY:


crates/rolldown_utils/src/light_guess.rs
```diff
@@ -3,88 +3,62 @@ use mime::Mime;
 use phf::{phf_map, Map};
 use std::str::FromStr;
 
-pub static MIME_TYPES: Map<&'static str, &'static str> = phf_map! {
+use crate::mime::MimeExt;
+pub static MIME_TYPES: Map<&'static str, (&'static str, bool)> = phf_map! {
     // Text
```
IWANABETHATGUY, line 19: Before, we intercepted too much extension with path compared to `esbuild`, this leads some disalignment.

IWANABETHATGUY: https://github.com/evanw/esbuild/blob/d34e79e2a998c21bb71d57b92b0017ca11756912/internal/helpers/mime.go#L4

crates/rolldown_utils/src/mime.rs
```diff
@@ -1,24 +1,47 @@
...
-use std::{path::Path, str::FromStr};
+use std::{fmt::Display, path::Path, str::FromStr};
 
 fn is_texture(data: &[u8]) -> bool {
   std::str::from_utf8(data).is_ok()
 }
 
-pub fn guess_mime(path: &Path, data: &[u8]) -> anyhow::Result<Mime> {
+#[derive(Debug)]
+pub struct MimeExt {
```
IWANABETHATGUY, line 10: Adding `MimeExt` because a lots of `type` with `charset=utf8` is not defined in `Mime`, only those are supported 
```
    TEXT_PLAIN_UTF_8, "text/plain; charset=utf-8", 4, None, 10;
    TEXT_HTML_UTF_8, "text/html; charset=utf-8", 4, None, 9;
    TEXT_CSS_UTF_8, "text/css; charset=utf-8", 4, None, 8;

    TEXT_CSV_UTF_8, "text/csv; charset=utf-8", 4, None, 8;

    TEXT_TAB_SEPARATED_VALUES_UTF_8, "text/tab-separated-values; charset=utf-8", 4, None, 25;
```

crates/rolldown/tests/esbuild/loader/loader_data_url_base64_invalid_utf8/artifacts.snap
```diff
@@ -8,7 +8,7 @@ source: crates/rolldown_testing/src/integration_test.rs
 ```js
 
 //#region binary.txt
-var binary_default = "data:text/plain;base64,/w==";
```
IWANABETHATGUY, line 11: This looks regressed,because before we use extension to determine `.txt`'s mime, this did not align with `esbuild`

IWANABETHATGUY: https://github.com/evanw/esbuild/blob/d34e79e2a998c21bb71d57b92b0017ca11756912/internal/bundler/bundler.go#L769-L777

crates/rolldown_utils/src/light_guess.rs
```diff
@@ -3,88 +3,59 @@ use mime::Mime;
 use phf::{phf_map, Map};
 use std::str::FromStr;
 
-pub static MIME_TYPES: Map<&'static str, &'static str> = phf_map! {
+use crate::mime::MimeExt;
+pub static MIME_TYPES: Map<&'static str, (&'static str, bool)> = phf_map! {
```
hyf0: How about use a struct instead of tuple here? It's kind of confusing about the `bool`.

IWANABETHATGUY: ok

hyf0 (approved):

PR #2461, opened 2024-10-18
feat: support PluginContext addWatchFile

underfin:


crates/rolldown/src/bundler.rs
```diff
@@ -188,12 +189,16 @@ impl Bundler {
...
+    output.watch_files = {
+      let mut files = link_stage_output
+        .module_table
+        .modules
+        .iter()
+        .filter_map(|m| m.as_normal().map(|m| m.id.as_str().into()))
+        .collect::<Vec<ArcStr>>();
+      files.extend(self.plugin_driver.watch_files.iter().map(|f| f.clone()));
+      files
+    };
```
underfin, line 201: The rollup also collected the files from `PluginDriver` to `RollupBuild`(it only has one `graph#watchFiles`). ref https://github.com/rollup/rollup/blob/master/src/utils/PluginContext.ts#L56

IWANABETHATGUY: This should be lazy.

hyf0 (approved):

PR #2462, opened 2024-10-18
WIP fix: extension alias

IWANABETHATGUY:


crates/rolldown/tests/esbuild/packagejson/package_json_bad_main/artifacts.snap
```diff
@@ -1,25 +1,24 @@
...
 source: crates/rolldown_testing/src/integration_test.rs
 ---
+# warnings
+
+## UNRESOLVED_IMPORT
+
+```text
+[UNRESOLVED_IMPORT] Warning: "demo-pkg" is imported by "src/entry.js", but could not be resolved – treating it as an external dependency.
+
+```
```
IWANABETHATGUY, line 11: This regression due to `oxc_resolver` can not handle package json bad main when configure `extensionAlias`

PR #2465, opened 2024-10-19
fix: toplevel class decl rewrite

IWANABETHATGUY:
### Description
1.Currently, only module that needs_wrap will rewrite `ClassDeclaration` to `VarDeclaration`, esbuild will try to rewrite all top level classDeclaration https://github.com/evanw/esbuild/blob/d34e79e2a998c21bb71d57b92b0017ca11756912/internal/js_parser/js_parser_lower_class.go#L2028-L2043
2. align with https://github.com/evanw/esbuild/blob/d34e79e2a998c21bb71d57b92b0017ca11756912/internal/js_parser/js_parser_lower_class.go#L2277-L2283


crates/rolldown/src/module_finalizers/scope_hoisting/mod.rs
```diff
@@ -184,29 +192,6 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
...
-                ..TakeIn::dummy(self.alloc)
-              }
-              .into_in(self.alloc),
-            ),
-            ..TakeIn::dummy(self.alloc)
-          }
-          .into_in(self.alloc),
-        ))
-      }
-      ast::Declaration::FunctionDeclaration(_) => {
```
IWANABETHATGUY, line 206: unreachable

crates/rolldown/src/module_finalizers/scope_hoisting/mod.rs
```diff
@@ -184,29 +192,6 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
           ))
         }
       }
-      ast::Declaration::ClassDeclaration(cls_decl) => {
```
IWANABETHATGUY, line 187: Move to visitor

crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
```diff
@@ -352,6 +351,26 @@ impl<'me, 'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'me, 'ast> {
       ident.name,
       self.ctx.module.repr_name
     );
+    // detect if the toplevel class decl name is referenced in inner scope
```
IWANABETHATGUY: This impl is wrong, the name can be renamed in deconflict phase.

hyf0 (dismissed):

IWANABETHATGUY:
I am going to disable `just ued` in windows ci, since I have no way to investigate why it failed, and it is not worth to put too much effort on this, enable on macos and linux is enough.

hyf0:
> I am going to disable `just ued` in windows ci, since I have no way to investigate why it failed, and it is not worth to put too much effort on this, enable on macos and linux is enough.

Let's have a discussion tomorrow about if we should support development in windows.

hyf0 (approved):

PR #2469, opened 2024-10-20
chore: esbuild tests ts category diff reason

IWANABETHATGUY:


IWANABETHATGUY:
skip review since it is not affected core,

PR #2474, opened 2024-10-21
refactor(iife): extract separable logic

underfin:
### Description

Extract separable logic for umd usage.

### DIFF
--- a/crates/rolldown/src/ecmascript/format/iife.rs
+++ b/crates/rolldown/src/ecmascript/format/iife.rs
@@ -23,10 +23,7 @@
 //! 10. Render the footer if it exists.
 
 use crate::ecmascript::format::utils::namespace::generate_identifier;
-use crate::utils::chunk::collect_render_chunk_imports::{
-  collect_render_chunk_imports, ExternalRenderImportStmt, RenderImportDeclarationSpecifier,
-  RenderImportStmt,
-};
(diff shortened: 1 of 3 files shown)


hyf0 (approved):

crates/rolldown/src/ecmascript/format/iife.rs
```diff
@@ -178,90 +167,28 @@ pub fn render_iife(
   Ok(concat_source)
 }
 
-/// Handling external imports needs to modify the arguments of the wrapper function.
-fn render_iife_chunk_imports(ctx: &GenerateContext<'_>) -> (String, Vec<ExternalRenderImportStmt>) {
```
underfin, line 182: I review the logic, the current implementation need to deconflict symbols from external symbols.

``` .ts
// input 
import { a } from 'external'
console.log(a)

// rolldown iife output
(function(exports, external) {
   const { a } = external; // the `a` should be deconflict.
   console.log(a)
})()

// rollup iife output
(function(exports, external) {
   console.log(external.a)
})()
```

You could see the above example. We need to select one implementation to do it.
- deconflict symbols from external symbols for `const { a } = external;`
- rewrite ast for `console.log(external.a)`

I'm ok for both, but look like the first one is suitable, because we don't need to add logic to finalizer for umd/cjs/iife.
If you have other ideas, please let me know.

underfin: cc @hyf0

PR #2483, opened 2024-10-22
fix: hashbang code generation

IWANABETHATGUY:
### Description
1. Before, we didn't handle `hashbang` and kept it as is, then let `oxc` generate it at the module level. However, if there are multiple modules with hashbang, then a syntax will be raised [e.g.](https://oxc-project.github.io/oxc/playground/?code=3YCAAID6gICAgICAgICRCMLHCqOzQ%2FePs33VF20nRjvZPGHjK6NVbP%2BOYkjjnp2ajneuSMN4lq4M9qtRZz9233dFn8ZEcQDCqYIwB%2BliBTc2KVJdUkzo73ZtlTcrXJvcyCMio33NX398BwSA). This pr fixes that issue.
ref https://github.com/evanw/esbuild/blob/d34e79e2a998c21bb71d57b92b0017ca11756912/internal/linker/linker.go#L5663-L5673


hyf0:
Could you also add a `hashbang` chapter to describe the behavior of rolldown while dealing with `hashbang` like https://github.com/rolldown/rolldown/blob/main/docs/guide/in-depth/use-strict.md.

crates/rolldown/tests/esbuild/default/hashbang_banner_use_strict_order/bypass.md
```diff
@@ -1,5 +1,5 @@
 # Reason
-1. hashban not align
+1. iife impl follows rollup
```
IWANABETHATGUY: https://rollupjs.org/repl/?version=4.24.0&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMiUyMyElMjBpbiUyMGZpbGUlNUNuJ3VzZSUyMHN0cmljdCclNUNuZm9vKCklNUNuJTIyJTJDJTIyaXNFbnRyeSUyMiUzQXRydWUlMkMlMjJuYW1lJTIyJTNBJTIybWFpbi5qcyUyMiU3RCUyQyU3QiUyMmNvZGUlMjIlM0ElMjIlNUNuJTIyJTJDJTIyaXNFbnRyeSUyMiUzQWZhbHNlJTJDJTIybmFtZSUyMiUzQSUyMmNvZGUuanMlMjIlN0QlNUQlMkMlMjJvcHRpb25zJTIyJTNBJTdCJTIyb3V0cHV0JTIyJTNBJTdCJTIyYmFubmVyJTIyJTNBJTIyJTIzISUyMGZyb20lMjBiYW5uZXIlMjIlMkMlMjJmb3JtYXQlMjIlM0ElMjJpaWZlJTIyJTdEJTdEJTdE

hyf0 (approved):

Issue #2438, opened 2024-10-15
[Bug]: node watch tests seems unstable


underfin:
### Reproduction link or steps

see https://github.com/rolldown/rolldown/actions/runs/11322896333/job/31484809957 at macos

### What is expected?

-

### What is actually happening?

-

### System Info

```Shell
-
```

underfin:
https://github.com/rolldown/rolldown/actions/runs/11456512537/job/31874824001 at windows

underfin:
https://github.com/rolldown/rolldown/actions/runs/11495843374/job/31996251667?pr=2494
https://github.com/rolldown/rolldown/actions/runs/11495843374/job/31996905774?pr=2494

PR #2484, opened 2024-10-22
feat: watch include exclude

underfin:
### Description

ref https://rollupjs.org/configuration-options/#watch-exclude

### DIFF
--- a/crates/rolldown/src/watcher/watcher.rs
+++ b/crates/rolldown/src/watcher/watcher.rs
@@ -56,7 +56,7 @@ impl Watcher {
     };
     let inner = RecommendedWatcher::new(
       move |res| {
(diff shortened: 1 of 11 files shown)


hyf0 (approved):

IWANABETHATGUY:
Seems the watcher still has some unstable issues https://github.com/rolldown/rolldown/actions/runs/11456512537/job/31874824001

underfin:
> Seems the watcher still has some unstable issues https://github.com/rolldown/rolldown/actions/runs/11456512537/job/31874824001

Yeah. Let us track it at here. https://github.com/rolldown/rolldown/issues/2438

IWANABETHATGUY:
It seems the Windows ci panic could be reproduced stably

underfin:
Look like the error is gone

PR #2485, opened 2024-10-22
chore: bump `oxc_resolver` to 2.0.0

IWANABETHATGUY:

### DIFF
--- a/crates/rolldown/tests/esbuild/default/package_alias_match_longest/artifacts.snap
+++ b/crates/rolldown/tests/esbuild/default/package_alias_match_longest/artifacts.snap
@@ -3,22 +3,40 @@ source: crates/rolldown_testing/src/integration_test.rs
 ---
 # warnings
 
-## UNRESOLVED_IMPORT
+## DIAGNOSABLE_RESOLVE_ERROR
 
 ```text
-[UNRESOLVED_IMPORT] Warning: "pkg/bar/baz" is imported by "entry.js", but could not be resolved – treating it as an external dependency.
+[DIAGNOSABLE_RESOLVE_ERROR] Warning: Could not resolve "pkg/bar/baz" in entry.js
+   ╭─[entry.js:5:8]
+   │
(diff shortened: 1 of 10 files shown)


crates/rolldown/tests/esbuild/default/package_alias_match_longest/artifacts.snap
```diff
@@ -3,22 +3,40 @@ source: crates/rolldown_testing/src/integration_test.rs
...
+   │            ╰────── Matched alias not found
+───╯
 
 ```
-## UNRESOLVED_IMPORT
+## DIAGNOSABLE_RESOLVE_ERROR
 
 ```text
-[UNRESOLVED_IMPORT] Warning: "pkg/foo/bar/baz" is imported by "entry.js", but could not be resolved – treating it as an external dependency.
+[DIAGNOSABLE_RESOLVE_ERROR] Warning: Could not resolve "pkg/foo/bar/baz" in entry.js
```
hyf0, line 33:
```suggestion
[UNRESOLVED_IMPORT] Warning: Could not resolve "pkg/foo/bar/baz" in entry.js
```

`UNRESOLVED_IMPORT` seems more meaningful than `DIAGNOSABLE_RESOLVE_ERROR`.

hyf0: And since we are doing `treating it as an external dependency.`, it's better to keep this message.

IWANABETHATGUY: I don't apply the suggestion, since it is auto generated

IWANABETHATGUY: Also, `unresolved_import` can't give user code frame, I will refactor the resolve diagnostic after this pr.

hyf0 (approved):

PR #2486, opened 2024-10-22
refactor: resolve diagnostic

IWANABETHATGUY:


crates/rolldown/src/ast_scanner/mod.rs
```diff
@@ -228,11 +228,13 @@ impl<'me> AstScanner<'me> {
     self.scopes.get_root_binding(name)
   }
 
+  /// `is_dummy` means if it the import record is created during ast transformation.
   fn add_import_record(
     &mut self,
     module_request: &str,
     kind: ImportKind,
     module_request_start: u32,
+    is_dummy: bool,
```
hyf0, line 237: Could you store the `Span` directly? It's more clear than pass both `start` and `is_dummy `. Don't worry about if the size get bigger.

hyf0: There is a concern about that if we keep reusing this error for missing `@import` in css files. If so, storing `span_start` only might not be enough.

IWANABETHATGUY: for 
https://github.com/rolldown/rolldown/blob/f970f3db50d0251eb340cec310b5e08a013af1dd/crates/rolldown/tests/rolldown/topics/css/basic/main.css#L1-L5

```css
@import "dep.css";
        ^ span_start
                ^ span_start + module_request.len()
```
I don't see any difference compare to calc `span_end` in ecma import record.

IWANABETHATGUY: Adding `span_end` only 4 byptes, but because 8 bytes alignment, the struct will increase 8 bytes. I don't think this is trivial.

hyf0: For `url("xxx")`, I kind of hope that we could emphasize  any range we want instead of only the quoted part.

hyf0 (approved):
This is a big PR, let's merge it first if there aren't serious blockers.

crates/rolldown/src/ast_scanner/mod.rs
IWANABETHATGUY: since all diagnostics emitter is context aware (it will also display **n** lines above,  **n** lines below), it is enough for most user. I think it is not worth increasing 8 bytes to only provide a little friendly diagnostics.

PR #2487, opened 2024-10-23
chore: base64 test source file not align

IWANABETHATGUY:


crates/rolldown/tests/esbuild/loader/require_custom_extension_base64/test.custom
IWANABETHATGUY, line 1: ref: https://github.com/evanw/esbuild/blob/d34e79e2a998c21bb71d57b92b0017ca11756912/internal/bundler_tests/bundler_loader_test.go#L175, 
Before we use string representation directly, this caused the difference of base64, now we use binary instead of string.

hyf0 (approved):

PR #2489, opened 2024-10-23
feat: format umd

underfin:


crates/rolldown/tests/rolldown/function/export_mode/umd/complex_name_named/_config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "config": {
+    "format": "umd",
+    "name": "test.module",
+    "exports": "named",
+    "banner": "import assert from \"node:assert\"",
+    "footer": "assert(globalThis.test.module.a === 1);"
```
hyf0: How about using `_test.mjs`? This seems to abuse the banner and footer.

```js
// __test.mjs
import assert from "node:assert"
import "./dist/main.js"
assert(globalThis.test.module.a === 1);
```

crates/rolldown/tests/rolldown/function/export_mode/umd/complex_name_named/artifacts.snap
```diff
@@ -0,0 +1,24 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+---
+# Assets
+
+## main.js
+
+```js
+import assert from "node:assert"
```
hyf0: Would be better if this doesn't show in the snapshot. https://github.com/rolldown/rolldown/pull/2489/files#r1812325035

crates/rolldown/tests/rolldown/function/export_mode/umd/complex_name_named/_config.json
underfin: Good idea,Fixed it.

hyf0 (approved):
Good job!

PR #2492, opened 2024-10-24
chore: convert rest none js related module type ast lazy export

IWANABETHATGUY:


hyf0 (approved):

crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
```diff
@@ -470,6 +470,7 @@ impl<'a> BindImportsAndExportsContext<'a> {
     debug_assert!(
       matches!(importee.exports_kind, ExportsKind::Esm | ExportsKind::CommonJs)
         || importee.has_lazy_export
+        || importee.module_type == ModuleType::Empty
```
IWANABETHATGUY, line 473: Can't determine module exports kind if the modules type is empty, since the source file is `""`

crates/rolldown/src/utils/parse_to_ecma_ast.rs
```diff
@@ -66,20 +66,22 @@ pub fn parse_to_ecma_ast(
...
+      has_lazy_export = true;
+      (text_to_string_literal(&dataurl)?, OxcParseType::Js)
     }
     ModuleType::Binary => {
       let source = source.try_into_bytes()?;
       let encoded = rolldown_utils::base64::to_standard_base64(source);
       (binary_to_esm(&encoded, options.platform, RUNTIME_MODULE_ID), OxcParseType::Js)
     }
-    ModuleType::Empty => ("export {}".to_string(), OxcParseType::Js),
+    ModuleType::Empty => ("".to_string(), OxcParseType::Js),
```
IWANABETHATGUY: https://github.com/evanw/esbuild/blob/d34e79e2a998c21bb71d57b92b0017ca11756912/internal/bundler/bundler.go#L215-L217

PR #2497, opened 2024-10-24
refactor: store `is_export_star` info in `ImportRecord` meta directly

IWANABETHATGUY:
### Description
1. since they just did the same thing and even with vec `star_exports` we still need to get the original `ImportRec` by index, and then do some stuff.
2. this refactor also save some memory


crates/rolldown/src/stages/link_stage/wrapping.rs
```diff
@@ -103,7 +103,7 @@ impl LinkStage<'_> {
         );
       }
 
-      if !module.star_exports.is_empty() {
+      if module.has_star_export() {
```
hyf0, line 106: Most of modules doesn't have `star_export`, this cause a `o(1)` check method become `o(n)` check which is called on every modules.

crates/rolldown/src/stages/link_stage/mod.rs
```diff
@@ -235,24 +235,27 @@ impl<'a> LinkStage<'a> {
...
-                .entry(rec.resolved_module)
-                .or_insert_with(|| {
-                  // Created `SymbolRef` is only join the de-conflict process to avoid conflict with other symbols.
-                  self.symbols.create_facade_root_symbol_ref(
-                    importer.idx,
-                    legitimize_identifier_name(&ext.name).into_owned().into(),
-                  )
-                });
+        importer
+          .import_records
```
hyf0, line 242: The same as https://github.com/rolldown/rolldown/pull/2497/files#r1814691768.

crates/rolldown/src/stages/link_stage/mod.rs
```diff
@@ -300,7 +303,7 @@ impl<'a> LinkStage<'a> {
               let importee_linking_info = &self.metas[importee.idx];
               match rec.kind {
                 ImportKind::Import => {
-                  let is_reexport_all = importer.star_exports.contains(rec_id);
+                  let is_reexport_all = rec.meta.contains(ImportRecordMeta::IS_EXPORT_START);
```
hyf0, line 309: You could record this meta to get o(1) check here. However, it's better to keep the `star_exports` field for other parts.

crates/rolldown/src/stages/link_stage/mod.rs
hyf0: This one is still slower though it have o(1) check `has_star_export`. Anyway, probably should investigate it when we have profiling data.

hyf0 (approved):

crates/rolldown/src/stages/link_stage/mod.rs
IWANABETHATGUY: Yeah, but it needs to meet `is_entry` so only a few module will be executed.

PR #2507, opened 2024-10-25
fix: avoid watcher deadlock

### LINKED ISSUE #2438: [Bug]: node watch tests seems unstable

underfin:
### Description

close https://github.com/rolldown/rolldown/issues/2438#issuecomment-2434691950. The timeout is caused by deadlock.

### DIFF
--- a/crates/rolldown/src/watcher/watcher.rs
+++ b/crates/rolldown/src/watcher/watcher.rs
@@ -141,6 +141,8 @@ impl Watcher {
         }
       }
     }
+    // The inner mutex should be dropped to avoid deadlock with bundler lock at `Watcher::close`
+    std::mem::drop(inner);
     self.emitter.emit(WatcherEvent::Event, BundleEventKind::BundleEnd.into()).await?;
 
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY (approved):
I think we should solve this issue by refactoring the code to avoid deadlock rather than explicitly dropping the lock, cause you may miss dropping the lock in another place.
But since the flaky Ci is so painful, merge it for now.

underfin:
> I think we should solve this issue by refactoring the code to avoid deadlock rather than explicitly dropping the lock, cause you may miss dropping the lock in another place. But since the flaky Ci is so painful, merge it for now.

Yeah. I agree with you. But here always has two lock exits for `notify` and `bundler`. Maybe we could have a good idea to resolve it at future, but i hasn't idea to resolve it at now.

PR #2508, opened 2024-10-26
chore(justfile): add `just run` command for running local build

Boshen:
I couldn't find a `run` command anywhere.

### DIFF
--- a/justfile
+++ b/justfile
@@ -135,6 +135,10 @@ build target="native" mode="debug":
 _build-native-debug:
     just build native debug
 
+# `just build && just run` to run the local build
+run:
+    ./packages/rolldown/bin/cli.js
+
 # BENCHING
 
 bench-rust:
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY:
Still can't figure out what is this used for, did you mean build something use `cli`  without needed to install the workspace dependency?

Boshen:
> Still can't figure out what is this used for, did you mean build something use `cli` without needed to install the workspace dependency?

How do you guys debug a local build? We should give some instructions on how to debug a local build. As a first contributor I have no clue on how to run the project after the project is built.

IWANABETHATGUY:
change code here:
https://github.com/rolldown/rolldown/blob/81c89dd662d5515d757a23a8c23889492844da9d/crates/rolldown/examples/basic/entry.js#L1-L4
and change configuration here: 
https://github.com/rolldown/rolldown/blob/81c89dd662d5515d757a23a8c23889492844da9d/crates/rolldown/examples/basic.rs#L9-L18

PR #2509, opened 2024-10-26
refactor: improve `BindingStringOrRegex`

shulaoda:
### Description

              > Perhaps we can change the `BindingStringOrRegex` to `Either<String, JsRegExp>`, so that we don't need to perform additional processing on it on the js side. @IWANABETHATGUY cc

you could open another pr.

_Originally posted by @IWANABETHATGUY in https://github.com/rolldown/rolldown/issues/2393#issuecomment-2403729777_


IWANABETHATGUY (dismissed):

crates/rolldown_binding/src/options/binding_input_options/binding_watch_option.rs
```diff
@@ -6,7 +6,7 @@ use crate::options::plugin::types::binding_js_or_regex::{
   bindingify_string_or_regex_array, BindingStringOrRegex,
 };
 
-#[napi_derive::napi(object)]
+#[napi_derive::napi(object, object_to_js = false)]
```
IWANABETHATGUY, line 9: Why need `object_to_js`  here?
https://napi.rs/changelog/napi_derive#disable-tonapivalue-for-object

crates/rolldown_binding/src/options/plugin/types/binding_js_or_regex.rs
```diff
@@ -1,29 +1,85 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct BindingStringOrRegex(pub Either<String, JsRegExp>);
+
+impl Default for BindingStringOrRegex {
+  fn default() -> Self {
+    Self(Either::A(String::default()))
+  }
+}
```
IWANABETHATGUY: I don't see any `Default` usage

crates/rolldown_binding/src/options/binding_input_options/binding_watch_option.rs
shulaoda: This is because the current `BindingStringOrRegex` only implements the `FromNapiValue` trait and does not implement `ToNapiValue`. At present, it seems that `object_to_js` is unnecessary. If we need it in the future, we can implement it again, but there may be some performance trade-offs.

crates/rolldown_binding/src/options/plugin/types/binding_js_or_regex.rs
shulaoda: This is because its parent object has `Default` trait, so we must implement the trait for it.

shulaoda: Let me remove it because it doesn't seem necessary.

IWANABETHATGUY (approved):

PR #2516, opened 2024-10-28
feat(sourcemap): Add support for injecting Debug IDs 

timfish:
Adds support for injecting debug IDs into source files and sourcemaps as per the [TC39 proposal](https://github.com/tc39/source-map/blob/main/proposals/debug-id.md).


IWANABETHATGUY:
Cool, but it seems the proposal only reached stage one, we would like to merge it once the proposal is stable.

hyf0:
Is this feature supported in practice already? Does the maintain toolchain support this feature? If so, it's ok to merge it.

timfish:
> Is this feature supported in practice already? 

- Sentry users have been relying on debug ids in production for a number of years. Sentry process hundreds of millions of files per month containing debug ids
- Bun added debug id injection to it's bundler when they added sourcemap support
- [A Github search](https://github.com/search?q=%2F%5C%2F%5C%2F%23%20debugId%3D%2F&type=code) suggests they are also used by Backtrace and Expo

> Does the maintain toolchain support this feature?

I recently added the required support to [oxc::sourcemap](https://github.com/oxc-project/oxc/pull/6221).

We've published [bundler plugins and polyfills](https://www.npmjs.com/search?q=%40debugids) but in the long run it's more efficient to implement directly in each bundler since most already have an internal hashing mechanism that can be used. There's no point hashing every source file twice!

This PR still requires tests but I just wanted to confirm this would likely be merged before spending the extra time!

yyx990803:
@timfish thanks for the PR and the context - I think this is good to have, assuming debug IDs will just be ignored by tools that don't support them.

hyf0 (approved):
Thanks

PR #2517, opened 2024-10-29
fix(cli):  `-s` flag not working & Update cli help info

liruifengv:
### Description

Change the `-s` flag default to `true`.

```
or pass the -s on the last argument if you want to generate .map file).
```
As the above description, just pass the `-s` flag will generate `.map` file, so it default should be `true`.

Now default is `false`, it not works!


IWANABETHATGUY:
LGTM, could you add a test case like this ? https://github.com/rolldown/rolldown/blob/5fc923df94ca54c1fff35c5bc26d179813a9305d/packages/rolldown/tests/cli/cli-e2e.test.ts#L74-L81

IWANABETHATGUY:
wait, my bad, I am not familiar with the cli part, but it seems this is not the correct fix. change the default value from `false` to `true` would also change the flag

liruifengv:
> wait, my bad, I am not familiar with the cli part, but it seems this is not the correct fix. change the default value from `false` to `true` would also change the flag

What do you mean?

liruifengv:
image
It seems like there's a problem here?

liruifengv:
> image It seems like there's a problem here?

There is a problem with the logic here.
https://github.com/rolldown/rolldown/blob/main/packages/rolldown/src/cli/commands/help.ts#L74

IWANABETHATGUY:
Yeah, this should not be `no-sourcemap`, I prefer adding a new field `reverse` https://github.com/rolldown/rolldown/blob/5fc923df94ca54c1fff35c5bc26d179813a9305d/packages/rolldown/src/cli/arguments/alias.ts#L7-L7 here and use the new field to control whether to add `no` prefix, for now the `default` field has tow usage : 
1. one for controlling if add a `no` prefix to option
2. one for getting default value.

liruifengv:
OK. I will take a look tomorrow.

liruifengv:
Hi @IWANABETHATGUY I have do something:
- add `reverse` new field
- handle add prefix logic in `/src/cli/arguments/index.ts`
- remove add prefix logic in `help.ts`
- Update snapshot for `help` e2e test
- Add a new e2e test for pass `-s` option

packages/rolldown/tests/cli/__snapshots__/cli-e2e.test.ts.snap
```diff
@@ -7,46 +7,46 @@ USAGE rolldown -c <config> or rolldown <input> <options>
 
 OPTIONS
 
-  --config -c, <filename>     Path to the config file (default: \`rolldown.config.js\`).
```
IWANABETHATGUY: These indentation should be preserved for better display.

liruifengv: Done~

packages/rolldown/tests/cli/__snapshots__/cli-e2e.test.ts.snap
```diff
@@ -43,8 +43,8 @@ OPTIONS
   --jsx.refresh               React refresh transformation.
   --log-level <log-level>     Log level (silent, info, debug, warn).
   --module-types <types>      Module types for customized extensions.
-  --no-external-live-bindings Use external live bindings.
-  --no-treeshake              Enable treeshaking.
+  --no-external-live-bindings Do not use external live bindings.
+  --no-treeshake              Do not enable treeshaking.
```
IWANABETHATGUY: prefer `Disable`

packages/rolldown/src/cli/arguments/index.ts
```diff
@@ -37,13 +37,13 @@ export const options = Object.fromEntries(
     if (config && config?.abbreviation) {
       result.short = config?.abbreviation
     }
+    // add `do not` prefix for need reverse options description
+    if (config && config.reverse) {
+      result.description = `do not ${result.description}`
```
IWANABETHATGUY: Use `Disable` instead

IWANABETHATGUY (approved):
The rest part LGTM

IWANABETHATGUY:
The pr could merge once the issues is solved.

packages/rolldown/src/cli/arguments/index.ts
liruifengv: The `use` and `enable` are hardcoded in the description string, Do you mean that I should replace the enable with disable instead of adding prefixes (Do not)?

https://github.com/rolldown/rolldown/blob/1197c99aa8cd9141ef611d6254024273e3ec0e10/packages/rolldown/src/options/output-options.ts#L78

liruifengv: Something like
```suggestion
-     result.description = `do not ${result.description}`
+     if (result.description.startsWith("enable") {
+       result.description.replace("enable", "disable")
+     } else {
+       result.description = `disable ${result.description}`
+     }
```

IWANABETHATGUY: replace description `use external live bindings` with `external live bindings`, then https://github.com/rolldown/rolldown/pull/2517/files#diff-45393b8f7b8fa098c265346c4a08841b717ae192e5e4f011b7329f9d78d63961R42 replace it with `Disable` should enough.

PR #2523, opened 2024-10-30
feat(json_plugin): add named_exports

jason89521:
### Description

Support `namedExports` for builtin json plugin.

### Impl

If `stringify` not set, transform the json to js object with the `named_exports` property.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -136,6 +136,7 @@ itertools           = "0.13.0"
 itoa                = "1.0.11"
 json-strip-comments = "1.0.4"
 jsonschema          = { version = "0.26.0", default-features = false }
+lazy_static         = "1.5.0"
 lightningcss        = { version = "1.0.0-alpha.57" }
 memchr              = "2.7.2"
 mimalloc            = "0.1.42"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -136,7 +136,6 @@ itertools           = "0.13.0"
 itoa                = "1.0.11"
 json-strip-comments = "1.0.4"
 jsonschema          = { version = "0.26.0", default-features = false }
-lazy_static         = "1.5.0"
 lightningcss        = { version = "1.0.0-alpha.57" }
 memchr              = "2.7.2"
 mimalloc            = "0.1.42"
--- a/crates/rolldown_plugin_json/src/lib.rs
+++ b/crates/rolldown_plugin_json/src/lib.rs
@@ -49,13 +49,13 @@ impl Plugin for JsonPlugin {
       }));
     }
 
-    let output = to_esm(value, self.named_exports);
+    let output = to_esm(&value, self.named_exports);
 
-    return Ok(Some(HookTransformOutput {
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -129,6 +129,7 @@ itertools           = "0.13.0"
 itoa                = "1.0.11"
 json-strip-comments = "1.0.4"
 jsonschema          = { version = "0.26.0", default-features = false }
+lazy_static         = "1.5.0"
```
hyf0: Please use the std's `OnceLock`.

jason89521: Thank you for the tip! I learned something new with Rust.

I also realized that I forgot to use the global variable, so I've already added it in the fix commit.

[24b86d2](https://github.com/rolldown/rolldown/pull/2523/commits/24b86d212870c84d085163d1c930411134133d3b)

crates/rolldown_plugin_json/src/lib.rs
```diff
@@ -92,9 +106,60 @@ fn is_special_query(ext: &str) -> bool {
...
+  static FORBIDDEN_IDENTIFIERS: OnceLock<HashSet<&'static str>> = OnceLock::new();
+  FORBIDDEN_IDENTIFIERS.get_or_init(|| {
+    let mut set: HashSet<&'static str> =
+      RESERVED_WORDS.split_whitespace().chain(BUILTINS.split_whitespace()).collect();
+    set.insert("");
+    set
+  })
+}
+
+fn make_legal_identifier(ident: &str) -> String {
```
IWANABETHATGUY: We have implementation here https://github.com/rolldown/rolldown/blob/d116b6be242be48fe23ccd3c0fc8e0c17bced3a5/crates/rolldown_utils/src/ecma_script.rs#L12-L12

IWANABETHATGUY: And you don't need to  care about the `reserved` keywords, it will be automatically deconflict in the finalized phase.

jason89521: image

I used `legitimize_identifier_name` to check whether an identifier is valid. However, for identifiers like `const` or `true`, the function returns the same identifier, which causes the parser to fail when parsing the code. Did I miss something?

IWANABETHATGUY: My bad, this is phase before parsing, so you need to convert it by hand.
you could use  https://github.com/rolldown/rolldown/blob/34cd0dc8c658946a12c614a1558433bce6766070/crates/rolldown_utils/src/ecma_script.rs#L8-L8 function, to check, this should be more performant.

jason89521: Used `is_validate_assignee_identifier_name` in [f1d4290](https://github.com/rolldown/rolldown/pull/2523/commits/f1d42908e7a24bd06536df5d41d40fb377767c0d)

hyf0 (approved):

hyf0:
Thanks

PR #2524, opened 2024-10-30
refactor(rolldown_binding): use a safer method to check `JsRegExp`

shulaoda:
### Description

Related to #2509 

I found that `instanceof` is less reliable compared to `Object.prototype.toString.call`, as it doesn’t work consistently across environments like web workers and iframes.

### DIFF
--- a/crates/rolldown_binding/src/options/plugin/types/binding_js_or_regex.rs
+++ b/crates/rolldown_binding/src/options/plugin/types/binding_js_or_regex.rs
@@ -2,7 +2,7 @@ use std::fmt::Debug;
 
 use napi::{
   bindgen_prelude::{FromNapiValue, Function, TypeName, ValidateNapiValue},
-  sys, Either, Env, Error, JsObject, JsUnknown, NapiValue, Status,
+  sys, Either, Env, Error, JsObject, NapiValue, Status,
 };
 use serde::{Deserialize, Deserializer};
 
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY:
Interesting, I am wondering if node binding could be used in `worker_thead` ? if the answer is no, we may don't need to care about it.

shulaoda:
> Interesting, I am wondering if node binding could be used in `worker_thead` ? if the answer is no, we may don't need to care about it.

Yes, but I’m not entirely sure if `rolldown` will have a use case for this scenario. I discovered this issue while improving the `JsRegExp` in https://github.com/web-infra-dev/rspack/pull/8254 , which has many related test cases for web workers.

IWANABETHATGUY:
> > Interesting, I am wondering if node binding could be used in `worker_thead` ? if the answer is no, we may don't need to care about it.
> 
> Yes, but I’m not entirely sure if `rolldown` will have a use case for this scenario. I discovered this issue while improving the `JsRegExp` in [web-infra-dev/rspack#8254](https://github.com/web-infra-dev/rspack/pull/8254) , which has many related test cases for web workers.

Do you mean this refactor make them pass more wob worker related tests? I assume this improvement only happen when load a node addon in a web worker right? But IIRC, there is no way to require module in web worker or iframe, this optimization seems useless.

IWANABETHATGUY:
feel free to fix me, I have not experiment  it yet.

shulaoda:
> I assume this improvement only happen when load a node addon in a web worker right?

I think so, let me draft it first and make further improvements when possible issues arise in the future.

IWANABETHATGUY:
> > I assume this improvement only happen when load a node addon in a web worker right?
> 
> I think so, let me draft it first and make further improvements when possible issues arise in the future.

ok

PR #2532, opened 2024-10-31
fix: conditional expr import and require

IWANABETHATGUY:
### Description
1. related commit https://github.com/evanw/esbuild/commit/2bff66e2b6c7ba1b1b759ea682f4f1380d67c633


hyf0 (approved):
Nice!

crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
```diff
@@ -365,7 +365,7 @@ impl<'me, 'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'me, 'ast> {
   #[allow(clippy::collapsible_else_if, clippy::too_many_lines)]
   fn visit_expression(&mut self, expr: &mut ast::Expression<'ast>) {
     if let Some(call_expr) = expr.as_call_expression_mut() {
-      if call_expr.is_global_require_call(self.scope) && !call_expr.span.is_empty() {
+      if call_expr.is_global_require_call(self.scope) && !call_expr.span.is_unspanned() {
```
IWANABETHATGUY, line 368: unspand means != (0, 0), however, `is_empty` means span.start == span.end`, `EnsureSpanUnique` visitor only guarantee the span not equal to `0, 0`, change to `is_unspanned` to make sure we could rewrite transformed `require`

crates/rolldown/src/utils/ecma_visitors/mod.rs
```diff
@@ -27,11 +27,18 @@ impl<'a> VisitMut<'a> for EnsureSpanUniqueness {
...
+      self.ensure_uniqueness(it.span_mut());
+    }
+    walk_mut::walk_call_expression(self, it);
+  }
 }
 
 impl EnsureSpanUniqueness {
   pub fn new() -> Self {
-    Self { visited_spans: FxHashSet::default(), next_unique_span_start: 1 }
+    Self { visited_spans: FxHashSet::from_iter([SPAN]), next_unique_span_start: 1 }
```
IWANABETHATGUY, line 41: Exclude `SPAN` (0, 0)

PR #2533, opened 2024-10-31
feat: print friendly msg for watch rebuild

underfin:
### Description

The pr print friendly msg for watch rebuild, here is example.

```
Waiting for changes...                                                                                                                   
Found index.js changed, rebuilding...                                                                                                    
✔ Rebuilt dist in 209.00 ms.                                                                                                            
```

### DIFF
--- a/crates/rolldown_common/src/types/watch.rs
+++ b/crates/rolldown_common/src/types/watch.rs
@@ -37,7 +37,7 @@ impl From<WatcherChange> for WatcherEventData {
 pub enum BundleEventKind {
   Start,
   BundleStart,
-  BundleEnd,
+  BundleEnd(BundleEndEventData),
   End,
 }
 
@@ -46,7 +46,7 @@ impl Display for BundleEventKind {
     match self {
       BundleEventKind::Start => write!(f, "START"),
       BundleEventKind::BundleStart => write!(f, "BUNDLE_START"),
-      BundleEventKind::BundleEnd => write!(f, "BUNDLE_END"),
+      BundleEventKind::BundleEnd(_) => write!(f, "BUNDLE_END"),
       BundleEventKind::End => write!(f, "END"),
     }
   }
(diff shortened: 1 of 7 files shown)


packages/rolldown/src/watcher.ts
```diff
@@ -42,7 +42,15 @@ export class Watcher {
         break
       case 'event':
         this.inner.on(BindingWatcherEvent.Event, async (data) => {
-          await listener(data)
+          if (data!.code === 'BUNDLE_END') {
+            await listener({
+              code: 'BUNDLE_END',
+              duration: Number(data!.duration),
```
IWANABETHATGUY, line 48: Why not just output `number` rust side?

crates/rolldown_common/src/types/watch.rs
```diff
@@ -56,10 +56,19 @@ impl From<BundleEventKind> for WatcherEventData {
   fn from(kind: BundleEventKind) -> Self {
     let mut map = HashMap::default();
     map.insert("code".to_string(), kind.to_string());
+    if let BundleEventKind::BundleEnd(data) = kind {
```
IWANABETHATGUY, line 59: The key could just use `&'static str`, you don't need to use `to_string`

packages/rolldown/src/watcher.ts
underfin: Because the data typing is ` WatcherEventData(Option<HashMap<String, String>>)`, binding `HashMap<String, String/Number>` is complex. 
I have tried to convert it to JSON string and parse it at js side, but it is also not good idea.

crates/rolldown_common/src/types/watch.rs
underfin: Updated.

IWANABETHATGUY (approved):

PR #2534, opened 2024-10-31
feat: create bundler for each bundle write or generate

### LINKED ISSUE #2495: [Bug]: Calling `bundle.write` multiple times with different output options does not work

1. Open https://stackblitz.com/edit/github-6jajke-ngvxxh?file=build.mjs
2. `node build.mjs` runs automatically
3. See `dist.es.js` is only output (`dist.cjs.js` is not output)

### What is expected?

Both `dist.es.js` and `dist.cjs.js` is output.

Only `dist.es.js` is output.

### System Info

```Shell
Stackblitz

  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    rolldown: https://pkg.pr.new/rolldown@1fa5d4b => 0.13.2
```

### Any additional comments?

_No response_

underfin:
### Description

close https://github.com/rolldown/rolldown/issues/2495

### DIFF
--- a/packages/rolldown/src/rolldown-build.ts
+++ b/packages/rolldown/src/rolldown-build.ts
@@ -1,33 +1,29 @@
 import { Bundler } from './binding'
 import type { OutputOptions } from './options/output-options'
 import { transformToRollupOutput } from './utils/transform-to-rollup-output'
-import { createBundler } from './utils/create-bundler'
+import { BundlerWithStopWorkers, createBundler } from './utils/create-bundler'
 
 import type { RolldownOutput } from './types/rolldown-output'
 import type { HasProperty, TypeAssert } from './utils/type-assert'
 import type { InputOptions } from './options/input-options'
-import { Watcher } from './watcher'
 
 export class RolldownBuild {
   #inputOptions: InputOptions
-  #bundler?: Bundler
-  #stopWorkers?: () => Promise<void>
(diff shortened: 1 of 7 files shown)


packages/rolldown/src/rolldown-build.ts
```diff
@@ -43,17 +39,15 @@ export class RolldownBuild {
...
-
-  async watch(outputOptions: OutputOptions = {}): Promise<Watcher> {
-    const bundler = await this.#getBundler(outputOptions)
-    const bindingWatcher = await bundler.watch()
-    const watcher = new Watcher(bindingWatcher)
-    watcher.watch()
-    return watcher
+    // If the bundler not create, create one to make close related things(eg. hooks) could be work.
+    if (this.#bundlers.length === 0) {
+      await this.#getBundler({})
```
IWANABETHATGUY, line 44: Maybe we could just return?

underfin: It will break the behavior

``` .tsx
const bundle = await rollup({
        input: './main.js',
        plugins: [
            {
              closeBundle: () => {
                console.log(1111) // should called
              },
            },
          ],
    });
    await bundle.close()
```

underfin:
close by https://github.com/rolldown/rolldown/pull/2535

PR #2537, opened 2024-10-31
fix:  remove plain import decl if importee is side effects free

IWANABETHATGUY:


crates/rolldown/tests/esbuild/default/top_level_await_forbidden_require_dead_branch/artifacts.snap
```diff
@@ -21,9 +22,7 @@ var init_b = __esm({ "b.js"() {} });
 //#endregion
 //#region a.js
 var a_exports = {};
-var init_a = __esm({ "a.js"() {
-	init_b();
```
IWANABETHATGUY, line 25: this is expected, since `b.js` is side effects free

hyf0 (approved):

PR #2545, opened 2024-11-01
fix: resolve id return false should mark module as external

underfin:

### DIFF
--- a/packages/rolldown/src/plugin/bindingify-build-hooks.ts
+++ b/packages/rolldown/src/plugin/bindingify-build-hooks.ts
@@ -170,8 +170,14 @@ export function bindingifyResolveDynamicImport(
         specifier,
         importer ?? undefined,
       )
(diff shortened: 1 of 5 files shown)


packages/rolldown/src/plugin/bindingify-build-hooks.ts
```diff
@@ -114,9 +114,15 @@ export function bindingifyResolveId(
         importer ?? undefined,
         newExtraOptions,
       )
-      if (ret == false || ret == null) {
+      if (ret == null) {
         return
       }
+      if (ret == false) {
```
IWANABETHATGUY: Did this needs to be strict equal?

underfin: Yeah. Good catch.

hyf0 (approved):

PR #2550, opened 2024-11-01
feat(cli): call watcher close at process exit

underfin:
### Description

Inspired rollup implementation, ref https://github.com/rollup/rollup/blob/master/cli/run/watch-cli.ts#L31

### DIFF
--- a/packages/rolldown/package.json
+++ b/packages/rolldown/package.json
@@ -101,7 +101,8 @@
     "dtsHeader": "type MaybePromise<T> = T | Promise<T>\ntype Nullable<T> = T | null | undefined\ntype VoidNullable<T = void> = T | null | undefined | void\nexport type BindingStringOrRegex = string | RegExp\n\n"
   },
   "dependencies": {
-    "zod": "^3.23.8"
+    "zod": "^3.23.8",
+    "signal-exit": "4.1.0"
   },
   "devDependencies": {
     "@napi-rs/cli": "^3.0.0-alpha.60",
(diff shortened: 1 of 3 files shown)


IWANABETHATGUY:
Can we add a test to ensure that if there are any errors, the watcher can be closed with a zero exit code?

packages/rolldown/package.json
```diff
@@ -101,7 +101,8 @@
     "dtsHeader": "type MaybePromise<T> = T | Promise<T>\ntype Nullable<T> = T | null | undefined\ntype VoidNullable<T = void> = T | null | undefined | void\nexport type BindingStringOrRegex = string | RegExp\n\n"
   },
   "dependencies": {
-    "zod": "^3.23.8"
+    "zod": "^3.23.8",
+    "signal-exit": "4.1.0"
```
hyf0: Better put it in `devDependencies`, since we will inline this package during bundling.

underfin: Done

underfin:
> Can we add a test to ensure that if there are any errors, the watcher can be closed with a zero exit code?

I tested it, because the `watcher.close` is a async,`onExit` maybe could't wait `watcher.close` finish, it is unstable. It is difficult to add a stable test for it, so let us skip it at now.

underfin:
I get some info after research, it is difficult to ensure async task is finished at process exit. But here has a helper info us to out. 

If you ^C a process, the kernel sends a SIGINT to the process group; that usually terminates both the parent and its children, grandchildren, and so on.

So we only need to care about is using watch cli bin at child process, Like test watch-cli using child process, it is unstable at process exit. So here i suggest to skip it at now.

IWANABETHATGUY (approved):

underfin:
tracking test at https://github.com/rolldown/rolldown/issues/2584

PR #2560, opened 2024-11-02
refactor: don't replace facade hash on binary data

hyf0:

PR #2559, opened 2024-11-02
refactor: tweak methods of `StrOrBytes`

hyf0:

PR #2557, opened 2024-11-02
refactor: allow the content of `InstantiatedChunk` to be binary data

hyf0:

PR #2539, opened 2024-10-31
feat: first-class asset support

hyf0:

PR #2561, opened 2024-11-02
feat(rust/test): show filename of binary files in snapshot

hyf0:

### DIFF
--- a/crates/rolldown/tests/rolldown/function/module_types/asset/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/function/module_types/asset/artifacts.snap
@@ -4,6 +4,8 @@ snapshot_kind: text
 ---
 # Assets
 
+## assets/foo-vpuLoj-k.txt
+
 ## main.js
 
 ```js
--- a/crates/rolldown_testing/src/integration_test.rs
+++ b/crates/rolldown_testing/src/integration_test.rs
@@ -210,22 +210,7 @@ impl IntegrationTest {
       let artifacts = assets
         .iter()
         .filter_map(|asset| {
-          let content = match asset {
-            Output::Chunk(inner) => &inner.code,
-            Output::Asset(inner) => match &inner.source {
-              rolldown_common::StrOrBytes::Str(inner) => inner,
-              // Snapshot buffer is meaningless
-              rolldown_common::StrOrBytes::Bytes(_) => return None,
-            },
-          };
-          let content = if self.test_meta.hidden_runtime_module {
-            RUNTIME_MODULE_OUTPUT_RE.replace_all(content, "")
-          } else {
-            Cow::Borrowed(content.as_str())
-          };
-
           let filename = asset.filename();
-
           let file_ext = filename.as_path().extension().and_then(OsStr::to_str).map_or(
             "unknown",
             |ext| match ext {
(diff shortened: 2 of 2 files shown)


hyf0:
* **#2562** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2562?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2561** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2561?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2560** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2560?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2559** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2559?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2558** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2558?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2557** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2557?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2539** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2539?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/rolldown/tests/rolldown/function/module_types/asset/artifacts.snap
```diff
@@ -4,6 +4,10 @@ snapshot_kind: text
 ---
 # Assets
 
+## assets/foo-vpuLoj-k.txt
+
+BINARY CONTENT DETECTED
```
IWANABETHATGUY: Can we try to display it if it is a valid utf8, the snapshot is misleading.

hyf0: That would lead us to do a pre-check on binary data. I don't against it. But it still doesn't solve the problem of how printing binary files like `.png`.

Or we just print nothing. Any solution suggested?

---

> That would lead us to do a pre-check on binary data.

I thought it again. If a output's content is binary(though it might be valid text), snapshot the content and compare it between commits is kind of meaningless for ensuring correctness.

hyf0:
### Merge activity

* **Nov 2, 2:26 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/rolldown/rolldown/2561) rebased this pull request after merging its parent, because this pull request is set to merge when ready.

crates/rolldown/tests/rolldown/function/module_types/asset/artifacts.snap
IWANABETHATGUY: FYI, this is esbuild snapshot, I think just print them as ` String::from_utf8_lossy` is good enough rather than `Binary Detected`

hyf0: I don't get it. What's point of printing `String::from_utf8_lossy` on png files?

--- edited

> rather than `Binary Detected`

This is not a good solution, but I don't consider printing `String::from_utf8_lossy` is better a option. If the asset is png file, printing `String::from_utf8_lossy` just create bunch of meaningless garbled characters. Showing them in the snapshot seems not have any benifit.

hyf0: I changed it to print nothing.

IWANABETHATGUY: For most of the tests the png is small, prints them just ok, I don't see too much garbled char in the esbuild snapshot, they just print the file asset with string https://github.com/evanw/esbuild/blob/d34e79e2a998c21bb71d57b92b0017ca11756912/internal/bundler_tests/snapshots/snapshots_default.txt#L4580-L4626, and with your solution, It is hard to work with esbuild test diff.

IWANABETHATGUY (approved):

crates/rolldown/tests/rolldown/function/module_types/asset/artifacts.snap
IWANABETHATGUY: let's merge it first, if there are too many diff with `file` loader compare to esbuild in the future, I will add the snapshot back.

PR #2566, opened 2024-11-03
refactor: simplify `replace_placeholder_with_hash`

IWANABETHATGUY:

### DIFF
--- a/crates/rolldown/src/utils/chunk/finalize_chunks.rs
+++ b/crates/rolldown/src/utils/chunk/finalize_chunks.rs
@@ -1,4 +1,4 @@
-use std::{hash::Hash, mem};
+use std::hash::Hash;
 
 use arcstr::ArcStr;
 use itertools::Itertools;
@@ -121,9 +121,11 @@ pub fn finalize_assets(
       // TODO: PERF: should check if this asset has dependencies/placeholders to be replaced
       match &mut asset.content {
         StrOrBytes::Str(content) => {
-          *content =
-            replace_placeholder_with_hash(mem::take(content), &final_hashes_by_placeholder)
-              .into_owned();
+          if let std::borrow::Cow::Owned(str) =
+            replace_placeholder_with_hash(content, &final_hashes_by_placeholder)
+          {
+            *content = str;
+          }
         }
         StrOrBytes::Bytes(_content) => {}
       }
--- a/crates/rolldown_utils/src/hash_placeholder.rs
+++ b/crates/rolldown_utils/src/hash_placeholder.rs
@@ -79,11 +79,10 @@ impl HashPlaceholderGenerator {
 /// ```
 #[expect(clippy::implicit_hasher)]
 pub fn replace_placeholder_with_hash<'a>(
-  source: impl Into<Cow<'a, str>>,
+  source: &'a str,
   final_hashes_by_placeholder: &FxHashMap<ArcStr, &'a str>,
 ) -> Cow<'a, str> {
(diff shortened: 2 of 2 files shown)


IWANABETHATGUY:
* **#2566** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2566?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @IWANABETHATGUY and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

crates/rolldown/src/utils/chunk/finalize_chunks.rs
```diff
@@ -122,8 +122,7 @@ pub fn finalize_assets(
       match &mut asset.content {
         StrOrBytes::Str(content) => {
           *content =
-            replace_placeholder_with_hash(mem::take(content), &final_hashes_by_placeholder)
-              .into_owned();
+            replace_placeholder_with_hash(content, &final_hashes_by_placeholder).into_owned();
```
hyf0: Does this cause a extra `clone` if the replacement doesn't happen?

IWANABETHATGUY: Can you elaborate where is the potential `clone` happen? the function use `&'a str` and  the `Regex::replace_all` return `Cow::borrow` if the there is no matched found in content. as doc said: 
> Replaces all non-overlapping matches in the haystack with the
replacement provided. This is the same as calling `replacen` with
`limit` set to `0`.

> If no match is found, then the haystack is returned unchanged. In that
case, this implementation will likely return a `Cow::Borrowed` value
such that no allocation is performed.

hyf0: The implicit `clone` is done by `into_onwed`. If no replacement happen, you'll get a `Cow::Borrowed(&str)`.

`into_owned` would convert the `Cow::Borrowed(&str)` to `String`, that's where the clone happening.

IWANABETHATGUY: I see

crates/rolldown/src/utils/chunk/finalize_chunks.rs
```diff
@@ -121,9 +121,11 @@ pub fn finalize_assets(
       // TODO: PERF: should check if this asset has dependencies/placeholders to be replaced
       match &mut asset.content {
         StrOrBytes::Str(content) => {
-          *content =
-            replace_placeholder_with_hash(mem::take(content), &final_hashes_by_placeholder)
-              .into_owned();
+          if let std::borrow::Cow::Owned(str) =
```
hyf0, line 124: Maybe it's better to hide this logic inside the `replace_placeholder_with_hash` function. It would be confusing for people who doesn't understand how `regex` works.

I think the original `replace_placeholder_with_hash` is good enough for this use case. Though it still create a dummy `String::default`.

IWANABETHATGUY: This is relatively subjective to say which one is better, previous impl still does not hide full implementation inside the function, you need to take `content` and then convert it from `Cow` to `String` after processed.

hyf0: Yeah. Thanks for the explanation. It's indeed quite hard to dealing with due to `&T`, `T` and the ownership of rust.

PR #2585, opened 2024-11-04
fix: should reference importee namespace_object_ref for export all decl

underfin:

### DIFF
--- a/crates/rolldown/tests/rolldown/function/external/double-namespace-reexport/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/function/external/double-namespace-reexport/artifacts.snap
@@ -1,0 +1,26 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+snapshot_kind: text
+---
+# Assets
+
+## main.js
+
+```js
+import "external";
+
+
+//#region second.js
+var second_exports = {};
+import * as import_external from "external";
+__reExport(second_exports, import_external);
+
+//#endregion
+//#region first.js
+var first_exports = {};
+__reExport(first_exports, second_exports);
+
+//#endregion
+var foo = first_exports.foo;
+export { foo };
+```
--- a/crates/rolldown/tests/rolldown/function/external/double-namespace-reexport/main.js
+++ b/crates/rolldown/tests/rolldown/function/external/double-namespace-reexport/main.js
@@ -1,0 +1,2 @@
+/* eslint-disable */
+export { foo } from './first.js';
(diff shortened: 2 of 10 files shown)


crates/rolldown/tests/rolldown/function/external/double-namespace-reexport/artifacts.snap
```diff
@@ -0,0 +1,26 @@
...
+
+## main.js
+
+```js
+import "external";
+
+
+//#region second.js
+var second_exports = {};
+import * as import_external from "external";
```
underfin, line 15: Here the import should be at top of file to instead of `import "external";` I will fixed it at next.

crates/rolldown/tests/rolldown/function/external/double-namespace-reexport/main.js
```diff
@@ -0,0 +1,2 @@
+/* eslint-disable */
```
underfin, line 1: cc @Boshen. The `.oxlintignore` already ignored `crates/**` path, but it will failed at `git commit -a -m`.I run `npm run lint-code` is ok. I'm not sure it is has some different...

crates/rolldown/tests/rolldown/function/external/double-namespace-reexport/artifacts.snap
hyf0: Not needed. It's expected behavior and the same as esbuild did it https://hyrious.me/esbuild-repl/?version=0.23.0&b=e%00entry.js%00import+%27.%2Ffoo%27%0A%0Aexport+%7B+foo+%7D+from+%27.%2Ffirst.js%27%3B&b=%00first.js%00export+*+from+%27external%27%3B&b=%00foo.js%00console.log%28%27foo%27%29&o=--bundle+--external%3Aexternal+--format%3Desm

crates/rolldown/tests/rolldown/function/external/double-namespace-reexport/main.js
hyf0: I met this too https://discord.com/channels/1138376133969203262/1199281741165248522/1287715251135053895

hyf0 (approved):

crates/rolldown/tests/rolldown/function/external/double-namespace-reexport/artifacts.snap
underfin: But the top of `import 'external'` is unnecessary. Other named import from external is placed at top, i think it is better if  make them consistent.

hyf0: > But the top of import 'external' is unnecessary.

You could try to solve this one. But `import * as import_external from "external";` is in the right place, don't require to fix it.

PR #2601, opened 2024-11-05
chore: support rollup test runner import files

underfin:

### DIFF
--- a/packages/rollup-tests/test/function/index.js
+++ b/packages/rollup-tests/test/function/index.js
@@ -12,7 +12,7 @@ const {
 	// verifyAstPlugin
 } = require('../utils.js');
 
-function requireWithContext(code, context, exports) {
+function requireWithContext(code, context, exports, isImport) {
 	const module = { exports };
 	const contextWithExports = { ...context, module, exports };
 	const contextKeys = Object.keys(contextWithExports);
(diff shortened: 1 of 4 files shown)


packages/rollup-tests/test/function/index.js
```diff
@@ -24,24 +24,36 @@ function requireWithContext(code, context, exports) {
...
 		}
 		const code = codeMap[outputId];
 		return code === undefined
-			? require(importee)
+			? isImport ? import(importee) : require(importee)
 			: (exportsMap[outputId] = requireWithContext(
-					code,
-					{ require: requireFromOutputVia(outputId), ...context },
-					(exportsMap[outputId] = {})
+				// Here replaced `import` to avoid unexpected error
```
IWANABETHATGUY, line 46: I recommended you split the rewrite logic in another file,this could make it more clear for the tests runner logic, especially when you got lot of rewrite rules.

IWANABETHATGUY (approved):

packages/rollup-tests/test/function/index.js
underfin: Here split the rewrite logic need to copy functions, it look like is unnecessary at now.

PR #2602, opened 2024-11-05
fix(plugin): ensure correct execution order

underfin:
### Description

- support js side hooks plugin order
- disable `buildStart` hook run in parallel, it break plugin order behavior. We could make the plugin run in parallel for supported hooks at future.
- refactor plugin order test follow rollup test

### DIFF
--- a/crates/rolldown_plugin/src/plugin_driver/build_hooks.rs
+++ b/crates/rolldown_plugin/src/plugin_driver/build_hooks.rs
@@ -12,41 +12,46 @@ use crate::{
 use anyhow::Result;
 use rolldown_common::{side_effects::HookSideEffects, ModuleInfo, ModuleType};
 use rolldown_sourcemap::SourceMap;
-use rolldown_utils::{futures::block_on_spawn_all, unique_arc::UniqueArc};
+use rolldown_utils::unique_arc::UniqueArc;
 
 use super::hook_filter::{filter_load, filter_resolve_id, filter_transform};
 
 impl PluginDriver {
   #[tracing::instrument(level = "trace", skip_all)]
   pub async fn build_start(&self) -> HookNoopReturn {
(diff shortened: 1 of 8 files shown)


crates/rolldown_plugin/src/plugin_driver/build_hooks.rs
```diff
@@ -12,41 +12,46 @@ use crate::{
...
 
-    for r in ret {
-      r?;
+    // for r in ret {
+    //   r?;
+    // }
+
+    for (_, plugin, ctx) in self.iter_plugin_with_context_by_order(&self.order_by_build_start_meta)
+    {
+      plugin.call_build_start(ctx).await?;
```
hyf0, line 54: https://rollupjs.org/plugin-development/#buildstart

`buildStart` is expected to be ran in parallel.

`order` also don't affect the `parallel` nature.

https://rollupjs.org/plugin-development/#buildstart:~:text=For%20parallel%20hooks%2C%20it%20changes%20the%20order%20in%20which%20the%20synchronous%20part%20of%20the%20hook%20is%20run.

> For parallel hooks, it changes the order in which the synchronous part of the hook is run.

I think the problem is that we treat all `handler` as `async` so their jobs so the actually sync one doesn't get executed orderly. Jobs are randomly picked by `tokio`.

hyf0: Other cause would be the rollup will make sure the sync code are executed, but we in rust would even delay the sync code of js side. This is tricky.

hyf0: Need a expert on this @Brooooooklyn cc

hyf0 (approved):
Let's merge it to ensure the correct behavior first.

Issue #2392, opened 2024-10-09
[Bug]: `this.getModuleInfo(id)` returns `null` in `load` hook


sapphi-red:
### Reproduction link or steps

1. https://stackblitz.com/edit/github-6jajke-xxwzrl?file=rolldown.config.mjs
2. `npm run build` runs automatically
3. see `/home/projects/github-6jajke-xxwzrl/src/entry.js false undefined` output in the terminal
4. run `npm run build:up` to run rollup
5. see `/home/projects/github-6jajke-xxwzrl/src/entry.js true true` output in the terminal

### What is expected?

`this.getModuleInfo(id)` returns the module info

### What is actually happening?

 `this.getModuleInfo(id)` returns `null`

### System Info

```Shell
Stackblitz

  System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.20.3 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    rolldown: https://pkg.pr.new/rolldown@2e3e2ed => 0.13.2
```

underfin:
It is diffcult at now, because the module created after `resolveid/load/transform` hook, so it is null at the hooks.

PR #2625, opened 2024-11-06
refactor: PluginContext#getModuleInfo

underfin:
### Description

The pr refactor the `PluginContext#getModuleInfo` implementation, using `PluginDriver#modules` to store module info table instead of using reference `ModuleTable`. It has some advantages.
- remove unsafe reference `ModuleTable`
- make https://github.com/rolldown/rolldown/issues/2392 has a chance could be fixed.
- make `PluginContext#load` logic easy to do, if the module already exit, here doesn't need to fetch it.

Here maybe has some performance regression because we need to `set_module_info` when some fields change, but it is acceptable. Because the operation almost happened at `ModuleTask`, only has overhead at refresh `importers/dynamic_importers` at main thread.

### DIFF
--- a/crates/rolldown/src/module_loader/module_loader.rs
+++ b/crates/rolldown/src/module_loader/module_loader.rs
@@ -323,12 +323,19 @@ impl ModuleLoader {
         if let Some(module) = module.as_normal_mut() {
           // Note: (Compat to rollup)
           // The `dynamic_importers/importers` should be added after `module_parsed` hook.
-          for importer in std::mem::take(&mut self.intermediate_normal_modules.importers[id]) {
+          let importers = std::mem::take(&mut self.intermediate_normal_modules.importers[id]);
+          for importer in &importers {
             if importer.kind.is_static() {
-              module.importers.push(importer.importer_path);
+              module.importers.push(importer.importer_path.clone());
             } else {
-              module.dynamic_importers.push(importer.importer_path);
+              module.dynamic_importers.push(importer.importer_path.clone());
             }
+          }
+          if !importers.is_empty() {
+            self
+              .shared_context
+              .plugin_driver
+              .set_module_info(&module.id, Arc::new(module.to_module_info()));
           }
         }
         module
--- a/crates/rolldown/src/module_loader/module_loader.rs
+++ b/crates/rolldown/src/module_loader/module_loader.rs
@@ -13,7 +13,7 @@ use oxc::transformer::ReplaceGlobalDefinesConfig;
 use rolldown_common::side_effects::{DeterminedSideEffects, HookSideEffects};
 use rolldown_common::{
   EntryPoint, EntryPointKind, ExternalModule, ImportKind, ImportRecordIdx, ImporterRecord, Module,
-  ModuleIdx, ModuleTable, ResolvedId, SymbolNameRefToken, SymbolRefDb,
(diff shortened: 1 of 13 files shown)


crates/rolldown/src/module_loader/module_loader.rs
```diff
@@ -323,13 +323,20 @@ impl ModuleLoader {
...
+          for importer in &importers {
             if importer.kind.is_static() {
-              module.importers.push(importer.importer_path);
+              module.importers.push(importer.importer_path.clone());
             } else {
-              module.dynamic_importers.push(importer.importer_path);
+              module.dynamic_importers.push(importer.importer_path.clone());
             }
           }
+          if !importers.is_empty() {
```
IWANABETHATGUY, line 334: This step could be parallel

IWANABETHATGUY: not a blocker, if the benchmark shows no regression, I am ok with it.

underfin: Yeah. I'm doing it.

IWANABETHATGUY (approved):

crates/rolldown/src/module_loader/module_loader.rs
underfin: The step run at parallel is not showing performance improve at benchmark, the start thread is more has overhead at here.

IWANABETHATGUY: That's alright, we could do it when it shows regression.

underfin:
The benchmark result should only see `scan` benchmark, here changes only happened at the scan stage.
After some benchmarking, looks like the little regression caused by setting `ModuleInfo`  at `ModuleTask`. We tracking it at #2628.

crates/rolldown/src/module_loader/module_task.rs
```diff
@@ -228,7 +228,9 @@ impl ModuleTask {
       asset_view,
     };
 
-    self.ctx.plugin_driver.module_parsed(Arc::new(module.to_module_info())).await?;
+    let module_info = Arc::new(module.to_module_info());
+    self.ctx.plugin_driver.set_module_info(&module.id, Arc::clone(&module_info));
```
IWANABETHATGUY, line 232: It seems you set it twice, https://github.com/rolldown/rolldown/pull/2625/files#diff-c9c7714142367c1f6a60995eb1e3ec2255fcab1aa62141a33de5268a814bdf57R335-R338

underfin: Yeah.Here beacause the `importers/dynamic_importers` be collected finished, we need to refresh some fields for modules. I have a try to spreate it to `PluginContext#moduleImporters`, and merge it at `PluginContext#getModuleId` called, but it is also not improving benchmark. The finally i found the performance regression caused by [here](https://github.com/rolldown/rolldown/pull/2625/files#diff-b77ae2f219a9298326818f35e202bf7403010390dae7ae1e6cd27dd9f0865131R231-R232).

IWANABETHATGUY:
Ok, let's improve after you finish 
> make https://github.com/rolldown/rolldown/issues/2392 has a chance could be fixed.
make PluginContext#load logic easy to do, if the module already exit, here doesn't need to fetch it.

PR #2626, opened 2024-11-06
refactor(rolldown_sourcemap): tweak `line_count` related code

hyf0:
### Description

- use the faster `line_count` function in all places
- only pre compute sourcemap data in parallel if necessary
- hide the pre-compute logic inside the function, so people don't bother to understand why we compute it and pass it instead of compute it when we needed.

### DIFF
--- a/crates/rolldown_sourcemap/src/source.rs
+++ b/crates/rolldown_sourcemap/src/source.rs
@@ -5,7 +5,9 @@ use crate::lines_count;
 pub trait Source {
   fn sourcemap(&self) -> Option<&SourceMap>;
   fn content(&self) -> &str;
-  fn lines_count(&self) -> u32;
+  fn lines_count(&self) -> u32 {
+    lines_count(self.content())
+  }
   #[allow(clippy::wrong_self_convention)]
   fn join(
     &self,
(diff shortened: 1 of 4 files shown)


hyf0:
* **#2626** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2626?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

underfin (approved):

underfin:
Look like the performance is not expected.

crates/rolldown_sourcemap/src/source.rs
```diff
@@ -83,7 +85,7 @@ impl Source for SourceMapSource {
   }
 
   fn lines_count(&self) -> u32 {
-    self.lines_count
+    self.pre_computed_lines_count.unwrap_or_else(|| lines_count(&self.content))
```
hyf0, line 88: @underfin this part was `self.pre_computed_lines_count.unwrap_or(lines_count(&self.content))` before. Should be ok now.

IWANABETHATGUY:
Maybe we should add another bench suite for `rolldown-source` and `string_wizard`,  the improvement could not be measured.

hyf0:
> Maybe we should add another bench suite for `rolldown-source` and `string_wizard`, the improvement could not be measured.

`string_wizard` has benchmark suite, I just didn't move them here. Current benchmark suite for `rolldown_sources` should be enough.

PR #2629, opened 2024-11-06
fix: emit error when `output.format` did not support `topLevelAwait`

IWANABETHATGUY:
### Description
1. related to https://github.com/evanw/esbuild/commit/5fe21253ee75fb4c5ea395a0877b2a5ab51c3575#diff-f312cd3ffe8d88cf424d3f0bd04b41e5b0ad36e549874b5688028bf9091a61f5L2238

### DIFF
--- a/crates/rolldown/src/ast_scanner/impl_visit.rs
+++ b/crates/rolldown/src/ast_scanner/impl_visit.rs
@@ -90,6 +90,40 @@ impl<'me, 'ast: 'me> Visit<'ast> for AstScanner<'me, 'ast> {
     walk::walk_member_expression(self, expr);
   }
 
+  fn visit_for_of_statement(&mut self, it: &ast::ForOfStatement<'ast>) {
+    if it.r#await {
+      if let Some(format) = self.options.as_ref().map(|option| &option.format) {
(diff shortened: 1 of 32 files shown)


crates/rolldown/src/ast_scanner/impl_visit.rs
```diff
@@ -90,6 +90,40 @@ impl<'me, 'ast: 'me> Visit<'ast> for AstScanner<'me, 'ast> {
     walk::walk_member_expression(self, expr);
   }
 
+  fn visit_for_of_statement(&mut self, it: &ast::ForOfStatement<'ast>) {
+    if it.r#await {
+      if let Some(format) = self.options.as_ref().map(|option| &option.format) {
+        if !format.keep_esm_import_export() {
```
hyf0: I kind of fail to understand what `keep_esm_import_export` means when reading esbuild's code.

hyf0 (approved):

crates/rolldown/src/ast_scanner/impl_visit.rs
IWANABETHATGUY: my bad, it should change to `keep_esm_import_export_syntax`,

PR #2631, opened 2024-11-06
fix: defer determine WrapKind of `EcmaView` of css module

IWANABETHATGUY:


crates/rolldown/src/utils/parse_to_ecma_ast.rs
```diff
@@ -52,8 +52,8 @@ pub fn parse_to_ecma_ast(
     ModuleType::Ts => (source.try_into_string()?, OxcParseType::Ts),
     ModuleType::Tsx => (source.try_into_string()?, OxcParseType::Tsx),
     ModuleType::Css => {
```
IWANABETHATGUY, line 56: The implementation is different since our CSS implementation is not the same as the esbuild https://github.com/evanw/esbuild/blob/d34e79e2a998c21bb71d57b92b0017ca11756912/internal/bundler/bundler.go#L278-L281

hyf0 (approved):

crates/rolldown/src/utils/parse_to_ecma_ast.rs
```diff
@@ -52,8 +52,8 @@ pub fn parse_to_ecma_ast(
...
-      let content = "export {}".to_string();
-      (content, OxcParseType::Js)
+      has_lazy_export = true;
+      ("({})".to_owned(), OxcParseType::Js)
```
IWANABETHATGUY: `{}` will be parsed as `blockStatement` so using this trick generate a `ObjectExpr`, oxc will unwrap it in codegen.

hyf0:
FYI: I was planned to solve the `don't generate ecma asset when css as entry` problem in a more proper way. This PR would still leaves a empty js files. However, applying the proper fix requires some refactor, so I probably won't do it recently.

IWANABETHATGUY:
> FYI: I was planned to solve the `don't generate ecma asset when css as entry` problem in a more proper way. This PR would still leaves a empty js files. However, applying the proper fix requires some refactor, so I probably won't do it recently.

I see.

hyf0:
More context: The current css is still not align with vite's behavior due to some limitation of internal code, which is the css-related thing I'm currently thinking of.

PR #2633, opened 2024-11-06
refactor: remove `Directive` to `StringLiteral` hack in `PreProcessor`

IWANABETHATGUY:


crates/rolldown/src/utils/tweak_ast_for_scanning.rs
```diff
@@ -29,19 +27,6 @@ impl<'a, 'ast> PreProcessor<'a, 'ast> {
 
 impl<'ast, 'a: 'ast> VisitMut<'ast> for PreProcessor<'a, 'ast> {
   fn visit_program(&mut self, program: &mut ast::Program<'ast>) {
-    if self.has_lazy_export {
```
IWANABETHATGUY, line 32: This hack is used because `oxc` will parse 
```js
"literal"
```
as directive, now we use `("literal")` instead, so not needed

hyf0 (approved):

Issue #2634, opened 2024-11-06
[Bug]: Watcher does not handle swapfiles


thesmartwon:
### Reproduction link or steps

```js
import { watch } from 'rolldown';
await watch({ input: 'main.js' });
```

```sh
> echo 'console.log(1)' > main.js
> node watch.js &
> top
node watcher-worker.js 100% CPU
> cat dist/main.js

//#region main.js
console.log(1);

//#endregion%

> echo 'console.log(2)' > main.js
> cat dist/main.js

//#region main.js
console.log(1);

//#endregion%

```

### What is expected?

Not 100% CPU and
```sh
> echo 'console.log(2)' > main.js
> cat dist/main.js

//#region main.js
console.log(2);

//#endregion%

```

thesmartwon:
I tried removing
```js
const watcherWorkerPath = require.resolve("rolldown/watcher-worker");
const child = spawn(process.argv[0], [watcherWorkerPath], {signal: this.controller.signal});
child.on("error", () => {});
```
...and `node watch.js` does not use 100% CPU, does not exit, but still does not rebundle after changes.

underfin:
@thesmartwon, could you provide your operating system? Look like the file update event has cross-platform issue.

thesmartwon:
Linux. I'll test on MacOS shortly.

IWANABETHATGUY:
Weird, I tested on Linux ubuntu 22.04 it worked well, I could reproduce the 100% CPU issue

IWANABETHATGUY:
Can you use `npx envinfo --system --npmPackages rolldown --binaries --browsers` to print your system info?

thesmartwon:
Works on MacOS with Deno 2.0.4 but still uses 100% CPU.

Doesn't on Linux with Deno 2.0.2. I'll try upgrading and I'll try node.

thesmartwon:
Linux with Deno 2.0.5 does not work. Linux with Node 23.1.0 does not work.

Rolldown 0.14.0, x64 Arch Linux 6.11

IWANABETHATGUY:
I use Deno 2.0.3 on Linux ubuntu 22.04, it worked, interesting.

thesmartwon:
Tried a fresh local debug build. Still doesn't work.

FWIW, `wait_for_change` is working on the Rust side. I'll continue to debug.

thesmartwon:
Edit 2: It's `vim` because it uses a swapfile to write!

[...]

PR #2635, opened 2024-11-06
fix: avoid watcher worker use more cpu

underfin:
### Description
related https://github.com/rolldown/rolldown/issues/2634


thesmartwon:
`exec`ing a file like this is unnecessary. You can just call `setInterval(() => {}, 1e9 /** Use less power */)` on the main thread if you don't want it to exit without Ctrl+C.

underfin:
closed by https://github.com/rolldown/rolldown/pull/2636

PR #2636, opened 2024-11-06
fix: use timer instead of watcher worker

thesmartwon:
### Description

Half of #2634

Alternative to #2635

Let me know @underfin if this breaks something else, but I think it should be functionally equivalent and avoid an extra thread...


underfin (approved):

underfin:
Thanks your good ideas.

PR #2648, opened 2024-11-07
chore: update esbuild test status

IWANABETHATGUY:

### DIFF
--- a/crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/artifacts.snap
+++ b/crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/artifacts.snap
@@ -6,6 +6,9 @@ snapshot_kind: text
 
 ## assets/common-h8VKzHVD.png
 
+```png
+x
+```
 ## common-3yRNVswI.css
 
 ```css
--- a/crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/diff.config.js
+++ b/crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/diff.config.js
@@ -1,0 +1,6 @@
+module.exports = {
+  resolver: {
+    "/out/common-LSAMBFUD.png": "assets/common-h8VKzHVD.png",
+    "/out/entries/entry.css": "entries_entry.css"
+  }
+}
--- a/crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/_config.json
+++ b/crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/_config.json
@@ -14,5 +14,6 @@
       ".css": "css",
       ".png": "asset"
     }
-  }
+  },
+  "snapshotBytes": true
 }
--- a/crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/diff.md
+++ b/crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/diff.md
@@ -1,24 +1,6 @@
 # Reason
-1. css reference .png
+1. generate wrong output when css as entry and has shared css
 # Diff
-## /out/common-LSAMBFUD.png
-### esbuild
-```js
-x
-```
-### rolldown
-```js
-
-```
-### diff
-```diff
-===================================================================
---- esbuild	/out/common-LSAMBFUD.png
-+++ rolldown	
-@@ -1,1 +0,0 @@
--x;
-
-```
 ## /out/entries/entry.css
 ### esbuild
 ```js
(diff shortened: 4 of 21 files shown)


IWANABETHATGUY:
* **#2648** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2648?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @IWANABETHATGUY and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/artifacts.snap
```diff
@@ -6,6 +6,9 @@ snapshot_kind: text
 
 ## assets/common-h8VKzHVD.png
 
+```png
+x
```
hyf0, line 10: Not sure if we want to show this in the snapshot file. The content seems meaningless.

IWANABETHATGUY: Only used for esbuild diff, by default it is hidden

hyf0: So this is for ensuring the binary data generated by rolldown is the same as esbuild generated?

crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/diff.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  resolver: {
+    "/out/common-LSAMBFUD.png": "assets/common-h8VKzHVD.png",
```
hyf0, line 3: I just notice that, if the esbuild's or rolldown's hash algo is changed, these mappings would be broken and all need to be changed manaully. Should we disable `[hash]` for asset too?

crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/artifacts.snap
IWANABETHATGUY: yes

crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/diff.config.js
IWANABETHATGUY: The hash algorithm should not be changed frequently(especially for esbuild), so I think we don't need to do this at this times Or if there are we will generate something like `common$1.png` if there are multiple assets has same basename

crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/artifacts.snap
hyf0: https://github.com/rolldown/rolldown/blob/02a65e2d19aa38072d3009509d9bd2cd5adcd30e/crates/rolldown_testing/src/integration_test.rs#L255

> ensuring the binary data generated by rolldown is the same as esbuild generated?

I think current `String::from_utf8_lossy(bytes)` can't achieve that goal, it would replace invalid bytes, which makes the data dirty. It's better to use the raw data to do a strict equal.

crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/diff.config.js
hyf0: > The hash algorithm should not be changed frequently

I agree with this. But once it changed(maybe two or three times), we need to fix all related tests manually. It would be frustrated for people who touches hash-related code.

crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/artifacts.snap
IWANABETHATGUY: Actually, this is only used for snapshot esbuild test assets, and I didn't find any invalid bytes in the esbuild test case. This is no hurry to do.

crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/diff.config.js
IWANABETHATGUY: I will try to support regex match in `esbuild-diff` so hash change, so people should not worry about asset hash algo modification

crates/rolldown/tests/esbuild/loader/loader_file_one_source_two_different_output_paths_css/artifacts.snap
hyf0: Maybe it's because esbuild didn't use a real picture. Well, since it's only in esbuild's test scope, no need to worry about it for now.

hyf0 (approved):

PR #2658, opened 2024-11-08
feat: implement vite resolve plugin

sapphi-red:


crates/rolldown_binding/src/options/plugin/binding_builtin_plugin.rs
```diff
@@ -142,6 +146,65 @@ pub struct BindingBuildImportAnalysisPluginConfig {
...
 
+#[napi_derive::napi(object)]
+#[derive(Debug, Deserialize, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct BindingViteResolvePluginConfig {
+  pub resolve_options: BindingViteResolvePluginResolveOptions,
+  pub environment_consumer: String,
+
+  pub runtime: String,
+  pub node_builtins: Vec<String>,
```
IWANABETHATGUY: I am not sure If we need to pass `node_builtins`, there is already one util function exist in the codebase, https://github.com/rolldown/rolldown/blob/dee08ac3d507e0ac2b3f83196295ad5809ede261/crates/rolldown_common/src/ecmascript/node_builtin_modules.rs#L78-L78

IWANABETHATGUY:
LGTM, since it is used for `vite` only, let's merge it when you think it is ready.

IWANABETHATGUY (approved):

crates/rolldown_plugin_vite_resolve/src/vite_resolve_plugin.rs
```diff
@@ -0,0 +1,402 @@
...
+
+        if external {
+          let resolved_ext = get_extension(&resolved.id);
+          if resolved_ext == "js" || resolved_ext == "mjs" || resolved_ext == "cjs" {
+            let id = args.specifier;
+            let mut resolved_id = id;
+            if is_deep_import(id) && resolved_ext != get_extension(id) {
+              if let Some(pkg_json) = oxc_resolved_result.unwrap().package_json() {
+                let has_exports_field =
+                  pkg_json.raw_json().as_object().unwrap().get("exports").is_some();
```
sapphi-red: @Boshen Does it make sense to have a feature to check the existence of `exports` field in `oxc_resolver`? This is the only reason I'm enabling `package_json_raw_json_api` feature. or is it fine to rely on `package_json_raw_json_api` feature?

This code is adding needed extensions for imports pointing external packages.
For example, you have `import foo from 'pkg/a'` in your code and `pkg` has `a.js` and doesn't have `exports` field. In the application code base, you can write `pkg/a` because Vite allows omitting extensions. But after the bundle that import needs to be `pkg/a.js` because Node.js does not allow omitting extensions in ESM. This code transforms `pkg/a` to `pkg/a.js` so that users can write `pkg/a` for imports pointing external packages.
Maybe it make sense to have this whole feature on the oxc-resolver side?

sapphi-red:
@IWANABETHATGUY It's ready for me. Would you take another look just in case because I pushed a bunch of codes?

IWANABETHATGUY (approved):

PR #2661, opened 2024-11-08
chore: enable rollup form test

underfin:


packages/rollup-tests/package.json
```diff
@@ -14,7 +14,8 @@
     "fs-extra": "^11.2.0",
     "mocha": "^10.4.0",
     "rolldown": "workspace:*",
-    "source-map-support": "^0.5.21"
+    "source-map-support": "^0.5.21",
+    "oxc-transform": "0.35.0"
```
hyf0, line 18: What's the usage of this?

hyf0 (approved):

packages/rollup-tests/package.json
underfin: Used to prettier code to diff rollup snapshot code

PR #2662, opened 2024-11-08
chore: update esbuild inject tests related stats

IWANABETHATGUY:


crates/rolldown/src/utils/pre_process_ecma_ast.rs
```diff
@@ -103,8 +103,13 @@ impl PreProcessEcmaAst {
         scopes = ret.scopes;
         self.ast_changed = true;
       }
-
       if !bundle_options.inject.is_empty() {
+        // if the define replace something, we need to recreate the semantic data.
```
hyf0, line 107: Just FYI @Boshen

hyf0 (approved):

crates/rolldown/src/utils/pre_process_ecma_ast.rs
IWANABETHATGUY: boshen knows this,  I discussed it with him before, and on the long run `oxc` will support update the `semantic` info on the fly, so we don't need to worry about the overhead.

PR #2680, opened 2024-11-10
support fileLoader

AliceLanniste:
### Description
support file loader

related #1295


IWANABETHATGUY:
We already have `asset` module loader which is almost same as file loader

AliceLanniste:
&gt; We already have `asset` module loader which is almost same as file loader

Thank you for telling me about this.

IWANABETHATGUY:
> > We already have `asset` module loader which is almost same as file loader
> 
> Thank you for telling me about this.

I am sorry about wasting your time, maybe you could ask about that before start working to avoid doing some needless work.

PR #2682, opened 2024-11-10
chore: remove redundant types and reorganize files

shulaoda:
### Description

Essentially, it is a series of improvements derived from the first issue.

- remove redundant type `StringOrRegExp`
- reorganize `type-assert` to `types/assert.ts`
- reorganize `normalizedStringOrRegex` to `utils/normalize-string-or-regex.ts`

I'm considering whether we could merge `bindingify-output-options` and `normalize-output-options`, as `rollup` only has a single `normalize-output-options`. When developing the output functionality, I often forget to destructure and return the new option in one of them. (also `input-options`)


hyf0 (approved):
Thank you!

PR #2687, opened 2024-11-11
feat: support PluginContext.load

### LINKED ISSUE #2355: PluginContext#load

underfin:
### Description

Support `PluginContext.load`, ref https://rollupjs.org/plugin-development/#this-load. close https://github.com/rolldown/rolldown/issues/2355.

The pr using msg channel to hand extra modules need to load. Here need to use one channel to avoid https://github.com/rolldown/rolldown/pull/2687#discussion_r1836055076.

> Here has a another way to do it. Using `load_via_plugins`, like `resolve_id_via_plugins`, but it need to a lot of change the relation crate dependencies and meet a cycle dependencies problem at moving `create_ecma_view`, it is difficult to resolve. try here at https://github.com/rolldown/rolldown/pull/2690.

The pr implements it as much as possible, but it also has a problem, consider some cases at here.
- `ModuleTables` has not the module `A`, here only fetch it.
- `ModuleTables` has module `A` and already fetched, here return already fetched result, using `PluginContext#modules`.
- `ModuleTables` has module `A` but it is not fetched finished, here may lost result because here can't ensure the fetch time before the module task is finished. The case maybe happend because it is run at parallel. We could re-fetch it at the case, but it call the module-related hook again, It could be put as low priority to resolve.


crates/rolldown/src/module_loader/module_loader.rs
```diff
@@ -306,6 +309,14 @@ impl ModuleLoader {
         }
       }
       self.remaining -= 1;
+
+      let mut count = load_module_rx_guard.len();
+      while count > 0 {
```
underfin: Here the channel can't starting recevier msg because the `await this.load` is block the `ModuleLoader#channel`. I will try to merge the two channel using one channel.

crates/rolldown/src/module_loader/runtime_module_task.rs
```diff
@@ -1,34 +1,29 @@
...
 
-use super::Msg;
 use crate::{
   ast_scanner::{AstScanner, ScanResult},
-  runtime::{RuntimeModuleBrief, RUNTIME_MODULE_ID},
   utils::tweak_ast_for_scanning::PreProcessor,
 };
 pub struct RuntimeModuleTask {
-  tx: tokio::sync::mpsc::Sender<Msg>,
+  tx: Arc<tokio::sync::mpsc::Sender<ModuleLoaderMsg>>,
```
hyf0: `Sender` is using `Arc` internally. Seems no need to wrap it with `Arc` again?

crates/rolldown/src/module_loader/module_loader.rs
```diff
@@ -76,11 +72,7 @@ impl ModuleLoader {
     fs: OsFileSystem,
     resolver: SharedResolver,
   ) -> anyhow::Result<Self> {
-    // 1024 should be enough for most cases
-    // over 1024 pending tasks are insane
-    let (tx, rx) = tokio::sync::mpsc::channel::<Msg>(1024);
-
-    let tx_to_runtime_module = tx.clone();
+    let tx_to_runtime_module = Arc::clone(&plugin_driver.tx);
```
hyf0: I don't have a solid reason, but this doesn't seems good. Do we have to place the channel in the `PluginDriver`?

crates/rolldown_plugin/src/plugin_context.rs
```diff
@@ -67,6 +101,27 @@ impl From<PluginContextImpl> for PluginContext {
...
+  ) -> anyhow::Result<()> {
+    self.context_load_modules.insert(specifier.into(), LoadCallback(Box::new(load_callback_fn)));
+    self
+      .tx
+      .send(ModuleLoaderMsg::FetchModule(ResolvedId {
+        id: specifier.into(),
+        ignored: false,
+        module_def_format: ModuleDefFormat::Unknown,
+        is_external: false,
+        package_json: None,
```
hyf0, line 119: Side note: So using `this.load` API would cause the `package.json#sideEffects` being ignored. Because it skips the resolve step, so we can't get a package json via oxc-resolver.

Rollup doesn't aware of `node_modules` so it doesn't have the problem I guess.

crates/rolldown_plugin/src/plugin_context.rs
```diff
@@ -67,6 +101,27 @@ impl From<PluginContextImpl> for PluginContext {
 }
 
 impl PluginContextImpl {
+  pub async fn load(
```
hyf0, line 102: It just came to me that `load` should not blocks the running task. What this PR does now is the right direction.

crates/rolldown_plugin/src/plugin_driver/mod.rs
```diff
@@ -43,6 +49,12 @@ impl PluginDriver {
   ) -> SharedPluginDriver {
     let watch_files = Arc::new(DashSet::default());
     let modules = Arc::new(DashMap::default());
+    let context_load_modules = Arc::new(DashMap::default());
+    // 1024 should be enough for most cases
+    // over 1024 pending tasks are insane
+    let (tx, rx) = tokio::sync::mpsc::channel(100);
+    let tx = Arc::new(tx);
+    let rx = Arc::new(Mutex::new(rx));
```
hyf0: Maybe create a one-shot channel in place when you need it so we don't need to use `ArcMutex` to wrap it?

crates/rolldown/src/module_loader/module_loader.rs
underfin: If there not store channel at the `PluginDriver`, it needs to set `tx` to `Arc<PluginDriver>`, mutate the `Arc<PluginDriver>` is a little messy. Let me try to fixed it.

crates/rolldown/src/module_loader/runtime_module_task.rs
underfin: Yeah. Here should remove `Arc`.

crates/rolldown_plugin/src/plugin_context.rs
underfin: Yeah. Here the rollup needs to manual set `moduleSideEffects` at here, the pr not including it.

crates/rolldown/src/module_loader/module_loader.rs
underfin: Fixed

crates/rolldown_plugin/src/plugin_driver/mod.rs
underfin: here already refactor finished.

crates/rolldown/src/module_loader/module_task.rs
```diff
@@ -231,12 +239,13 @@ impl ModuleTask {
 
     let module_info = Arc::new(module.to_module_info());
     self.ctx.plugin_driver.set_module_info(&module.id, Arc::clone(&module_info));
-    self.ctx.plugin_driver.module_parsed(module_info).await?;
+    self.ctx.plugin_driver.module_parsed(Arc::clone(&module_info)).await?;
+    self.ctx.plugin_driver.mark_context_load_modules_loaded(&module.id).await?;
```
hyf0, line 244: Because we are running in parallel, is it possible that a task of `foo.js` is running in line 244 and stop here, then a `load(foo.js)` message send to `module_loader` thread so this callback would never be resolved.

crates/rolldown/src/module_loader/module_task.rs
```diff
@@ -231,12 +239,13 @@ impl ModuleTask {
...
```
hyf0, line 243: Would be good if we don't need to await here. Is it possible just `tokio::spawn` it?

crates/rolldown/src/module_loader/module_task.rs
hyf0: I guess we should check left fallbacks in the end of the module loader.

hyf0:
Except above comments, others looks fine. Good job!

crates/rolldown/src/module_loader/module_task.rs
underfin: Here `in the end of the module loader` couldn't happened, the `await this.load` always block it, we need to find a good pointer to check. Other way as https://github.com/rolldown/rolldown/pull/2687#issue-2648390432 said, we could refetch it again. I will have a try to resolve it at next.

crates/rolldown/src/module_loader/module_task.rs
underfin: Not sure why don't need to await here. We need to callback the js side function, here need to using `awiat`. If the callback hasn't exit, it is also correct.

crates/rolldown/src/module_loader/module_task.rs
hyf0: Oh. The js hook which calls `await this.load` would blocks the rust hook, so it never reachs the end of the module loader.

crates/rolldown/src/module_loader/module_task.rs
hyf0: Let's assume if we don't await here, what bad thing might happens?

- The `self.ctx.plugin_driver.mark_context_load_modules_loaded(&module.id)` is being `spawn`, so the js callback would be executed in another thread. And it gets correct result, because ` self.ctx.plugin_driver.set_module_info(&module.id, Arc::clone(&module_info));` is already called.

hyf0 (approved):

crates/rolldown/src/module_loader/module_task.rs
underfin: Yeah. It could be done moving to other thread. But looks like it hasn't profit from it, it maybe seize the thread from module loader task.

Issue #2606, opened 2024-11-05
[Feature Request]: function type `treeshake.moduleSideEffects` and `"no-external"` is difficult to rewrite to RegExp


sapphi-red:
### What problem does this feature solve?

In rollup `treeshake.moduleSideEffects` accepts a function in addition to what rolldown accepts.
In Vite, we are passing the following function to `treeshake.moduleSideEffects`:
```js
 moduleSideEffects(id, external) {
      id = id.replaceAll('\\', '/')
      if (
        id.includes('node_modules/astring') ||
        id.includes('node_modules/acorn')
      ) {
        return false
      }
      return !external
    },
```
https://github.com/vitejs/vite/blob/ad9a987aa4ae3d9e246541d918bb2ec331bbccae/packages/vite/rollup.config.ts#L46-L57

To replace this with a RegExp, it needs to be
```js
moduleSideEffects: new RegExp(`^((?!${['acorn', 'astring', ...externals].join('|')}).)*$`),
```
This is difficult to write.
Also I'm using `externals` variable here, but this variable cannot be constructed correctly on users side, as a plugin might set something external and the user cannot know about that.

The same can be said for `moduleSideEffects: "no-external"`. It needs to be rewritten into
```js
moduleSideEffects: new RegExp(`^((?!${externals.join('|')}).)*$`),
```

refs #2367

### What does the proposed API look like?

Change `treeshake.moduleSideEffects` to accept an include RegExp and an exclude RegExp.
```ts
type ModuleSideEffectsOption = {
  include: RegExp
  exclude: RegExp
}
```
This would make it easier to write the RegExp, but still it's not possible to control externalized imports correctly. I'm not sure how to solve that one. Maybe adding an option that can control the behavior of externalized imports might be sufficient?
```ts
type ModuleSideEffectsOption = {
  external: boolean
  include: RegExp
  exclude: RegExp
}
```

IWANABETHATGUY:
Adding a function is not hard, but it is not recommended due to calling ts function *N* times (n is module count), will drastically decrease the performance. I will think of your proposal, also I think the static `sideEffects`(string + regex + `sideEffects` in package.json)  matcher should absolutely enough, because I don't see there are too many people complaining about webpack static `sideEffects` matcher design.

hyf0:
Interesting. Out of the main topic, we probably should consider changing places that accepts `Regex` to `Regex || Regex[]`.

PR #2695, opened 2024-11-11
feat: Add option moduleSideEffectsRule for `treeshake.moduleSideEffects`

IWANABETHATGUY:
### Description
1. part of https://github.com/rolldown/rolldown/issues/2606


packages/rolldown/src/treeshake/module-side-effects.ts
```diff
@@ -1,21 +1,27 @@
...
 >
 
-export const ModuleSideEffectsOptionSchema = z.boolean().or(z.string())
+export const ModuleSideEffectsRuleSchema = z
+  .object({
+    test: z.instanceof(RegExp).optional(),
+    external: z.boolean().optional(),
+    sideEffects: z.boolean(),
+  })
+  .refine((data) => {
```
hyf0, line 13: NIT: Would be good if this a compile time error.

hyf0 (approved):

packages/rolldown/src/treeshake/module-side-effects.ts
IWANABETHATGUY: I am working on that.

Issue #819, opened 2024-04-10
[Tracking] Rollup Plugin Compat Status


underfin:
### Fully Supported Hooks

- [x] `options`
- [x] `buildStart`
- [x] `buildEnd`
- [x] `renderStart`
- [x] `renderError`
- [x] `generateBundle`
- [x] `writeBundle`
- [x] `closeBundle`
- [x] `banner`
- [x] `footer`
- [x] `intro`
- [x] `outro`
- [x] `augmentChunkHash`
- [x] `closeWatcher`
- [x] `watchChange`
- [x] `renderChunk`

## Partually Supported Hooks

Hooks in this section can work as intended for most commonly seen plugins, but they may have some niche options / behavior that are not implemented yet.

Our priority at this stage is implement enough compatibility for most plugins to run, and determine whether to support the niche options at a later stage based on user demand. Some options are explicitly marked as "will not support" due to architectural incompatibility and limited use case.

- `outputOptions`
  - the `PluginContext` is not supported yet.
- `resolveId`
  - the 3rd argument `options.attributes` is not supported yet
  - the following fileds in the return value are not supported yet:
    - `attributes`
    - `resolvedBy`
    - `syntheticNamedExports`
- `load`
  - the following fileds in the return value are not supported yet:
    - `attributes`
    - `syntheticNamedExports`
    - `ast` <sup>will not support</sup>
- `transform`
  - the following fileds in the return value are not supported yet:
    - `attributes`
    - `syntheticNamedExports`
    - `ast` <sup>will not support</sup>
- `moduleParsed`
  - the argument `moduleInfo` is only partially compatible
- `resolveDynamicImport`
  - the 3rd argument `options.attributes` is not supported yet
- `onLog`
  - the `meta` property is not supported yet

### Not Yet Supported

- [ ] `shouldTransformCachedModule`
- [ ] `resolveFileUrl`
- [ ] `resolveImportMeta`
- [ ] `renderDynamicImport`

### Supported Properties

- [x] `debug`
- [x] `error`
- [x] `info`
- [x] `warn`
- [x] `getFileName`
- [x] `getModuleIds`
- [x] `getModuleInfo`
- [x] `addWatchFile`
- [x] `getCombinedSourcemap`
- [x] `parse`

### Partially Supported Properties

- `resolve`
  - `options.isEntry` and `options.attributes` are not supported yet
- `emitFile`
  - only support emitting asset and chunk for now
- `load`
[...]

PR #2696, opened 2024-11-11
feat(plugin/virtual): builtin virtual plugin

ikkz:
Basically a translation of https://github.com/rollup/plugins/blob/master/packages/virtual/src/index.ts

### DIFF
--- a/crates/rolldown_testing_config/src/lib.rs
+++ b/crates/rolldown_testing_config/src/lib.rs
@@ -42,6 +42,9 @@ pub struct TestMeta {
   /// If `true`, the bundle will be called with `write()` instead of `generate()`.
   #[serde(default = "true_by_default")]
   pub write_to_disk: bool,
+  /// If `true`, assets won't be snapshot
+  #[serde(default)]
+  pub snapshot_skip_assets: bool,
 }
 
 impl Default for TestMeta {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2584,6 +2584,7 @@ dependencies = [
  "rolldown_plugin_module_preload_polyfill",
  "rolldown_plugin_replace",
  "rolldown_plugin_transform",
+ "rolldown_plugin_virtual",
  "rolldown_plugin_wasm_fallback",
  "rolldown_plugin_wasm_helper",
  "rolldown_sourcemap",
@@ -2821,6 +2822,17 @@ dependencies = [
  "rolldown_plugin",
  "rolldown_utils",
  "sugar_path",
+]
+
+[[package]]
+name = "rolldown_plugin_virtual"
+version = "0.1.0"
+dependencies = [
+ "rolldown",
+ "rolldown_plugin",
+ "rolldown_testing",
+ "sugar_path",
+ "tokio",
 ]
 
 [[package]]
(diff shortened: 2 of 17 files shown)


crates/rolldown_testing_config/src/lib.rs
```diff
@@ -42,6 +42,9 @@ pub struct TestMeta {
   /// If `true`, the bundle will be called with `write()` instead of `generate()`.
   #[serde(default = "true_by_default")]
   pub write_to_disk: bool,
+  /// If `true`, assets won't be snapshot
+  #[serde(default)]
+  pub snapshot_skip_assets: bool,
```
ikkz, line 47: Added because without this option, the snapshot will be
```
---
snapshot_kind: text
---
# Assets

## entry.js

```js

//#region \0virtual:batman
var __virtual_batman_default = "na na na na na";

//#endregion
//#region \0virtual:/files/projects/rolldown/crates/rolldown_plugin_virtual/tests/src/robin.js  😨😨😨😨
var robin_default = "batman";

//#endregion
//#region src/entry.js
console.assert(__virtual_batman_default === "na na na na na");
console.assert(robin_default === "batman");

//#endregion
```
the path is not stable across machines. If there's a better way, please let me know.

hyf0:
Good job. Could you remove the binding part? We want to keep this plugin in the pure rust side for now. We want to control the current binary size.

yyx990803:
Thanks for the contribution! So we discussed this a little bit - and we don't really think we should add this as a built-in.

Every built-in plugin in Rolldown has a cost: it increases our maintenance surface and binary size. So when we decide what plugins should be built-in, there are two important considerations:

1. They should provide significant performance advantage over pure-JS plugins.
2. They should be very commonly used.
(And hidden 3 - it's needed in Vite)

In this case, the virtual plugin is [much less used compare to e.g. replace or node-resolve plugins](https://npmtrends.com/@rollup/plugin-node-resolve-vs-@rollup/plugin-replace-vs-@rollup/plugin-virtual). Occasional use of the Rollup virtual plugin also won't be a major performance bottleneck. So having it as a built-in doesn't seem like a good trade-off.

I hope you don't get discouraged for not having this PR merged! There are plenty of opportunities for future contributions, and we really appreciate the ones by you that are already merged. But when you plan to work on a relatively big feature in the future, it might be a good idea to communicate the intention via an issue first to avoid wasted work.

ikkz:
Thank you for your patient explanation; I fully understand the reasons now. I initially planned to open this PR because I used the virtual plugin during the migration process and saw a note [here](https://github.com/rolldown/rolldown/issues/819):  

> @rollup/plugin-virtual - we could provider a rust plugin to make it faster

However, I completely understand that the goals and plans of such a large project are subject to change at any time, so it's totally fine to close this PR. And I also look forward to continuing to contribute to Rolldown in the future.

IWANABETHATGUY:
> Thank you for your patient explanation; I fully understand the reasons now. I initially planned to open this PR because I used the virtual plugin during the migration process and saw a note [here](https://github.com/rolldown/rolldown/issues/819):
> 
> > @rollup/plugin-virtual - we could provider a rust plugin to make it faster
> 
> However, I completely understand that the goals and plans of such a large project are subject to change at any time, so it's totally fine to close this PR. And I also look forward to continuing to contribute to Rolldown in the future.

Thanks for point that out, I have updated the stats. We don't need to port it as builtin now because we have supported hook filter now, so that only a few virtual modules will call the js plugin, it should not be the bottleneck now.

niksy:
Does it make sense to make this external package that can be installed and used with Rolldown?

PR #2697, opened 2024-11-11
feat: support `RenderedModule.renderedLength`

### LINKED ISSUE #2686: [Bug]: `RenderedModule.renderedLength` support

1. Open https://stackblitz.com/edit/github-xv4yss-ggycbu?file=rolldown.config.mjs
2. `npm i` runs automatically
4. `chunk modules [ undefined, undefined ]` is output in the terminal
5. run `npm run build:up` to use rollup
6. `chunk modules [ 46, 19 ]` is output in the terminal

### What is expected?

`chunk modules [ number, number ]` is output (`RenderedModule.renderedLength` is supported)

`chunk modules [ undefined, undefined ]` is output (`RenderedModule.renderedLength` is not supported)

### System Info

```Shell
Stackblitz

  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    rolldown: https://pkg.pr.new/rolldown@f94549f => 0.14.0
```

### Any additional comments?

Found while trying to bundle Vite with Rolldown.
This property is necessary to make `rollup-plugin-license` to work ([relevant code](https://github.com/mjeanroy/rollup-plugin-license/blob/66333ea04aa0e204f43f976439fe4fae50dd8cdb/src/index.js#L58-L60)).

shulaoda:
### Description

closes #2686

There are still two remaining issues:

- Our `codegen` differs from `rollup`, and it appends `\n` by default at the end
- The chunk traversal order in our implementation does not match that of `rollup`

This is just a simple attempt, and feel free to close it at any time.

### DIFF
--- a/crates/rolldown/src/ecmascript/ecma_generator.rs
+++ b/crates/rolldown/src/ecmascript/ecma_generator.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use crate::{
   types::generator::{GenerateContext, GenerateOutput, Generator},
   utils::{chunk::generate_rendered_chunk, render_ecma_module::render_ecma_module},
@@ -21,7 +23,8 @@ use super::format::{
   app::render_app, cjs::render_cjs, esm::render_esm, iife::render_iife, umd::render_umd,
 };
 
-pub type RenderedModuleSources = Vec<(ModuleIdx, ModuleId, Option<Vec<Box<dyn Source + Send>>>)>;
+pub type RenderedModuleSources =
+  Vec<(ModuleIdx, ModuleId, Option<Arc<[Box<dyn Source + Send + Sync>]>>)>;
 
 pub struct EcmaGenerator;
 
--- a/crates/rolldown/src/ecmascript/ecma_generator.rs
+++ b/crates/rolldown/src/ecmascript/ecma_generator.rs
@@ -51,10 +51,10 @@ impl Generator for EcmaGenerator {
       })
       .collect::<Vec<_>>();
 
-    rendered_module_sources.iter().for_each(|(_, module_id, _)| {
+    rendered_module_sources.iter().for_each(|(_, module_id, sources)| {
       // FIXME: NAPI-RS used CStr under the hood, so it can't handle null byte in the string.
       if !module_id.starts_with('\0') {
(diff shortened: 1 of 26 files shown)


IWANABETHATGUY:
Can you add a test case?

crates/rolldown/src/ecmascript/ecma_generator.rs
```diff
@@ -48,10 +48,15 @@ impl Generator for EcmaGenerator {
...
+    rendered_module_sources.iter().for_each(|(_, module_id, sources)| {
       // FIXME: NAPI-RS used CStr under the hood, so it can't handle null byte in the string.
       if !module_id.starts_with('\0') {
-        rendered_modules.insert(module_id.clone(), RenderedModule { code: None });
+        let rendered_length = sources.as_ref().and_then(|sources| {
+          sources
+            .get(1)
+            .and_then(|source| u32::try_from(source.content().encode_utf16().count()).ok())
+        });
+        rendered_modules.insert(module_id.clone(), RenderedModule { code: None, rendered_length });
```
hyf0: If code is return to the JS side, we probably should be able to just call `this.code.length` instead of calculating it in rust side.

shulaoda: Do you mean we could discard the `renderedLength` option and simply return `code`, allowing users to call `this.code.length` on the js side instead?

hyf0:
My reviews:

1. Change `render_ecma_module` to return `Option<Arc<Vec<Box<dyn Source + Send>>>>`
2. Remove `RenderedModule#rendered_length` in rust side
3. Add `RenderedModule#inner_code` to store `Option<Arc<Vec<Box<dyn Source + Send>>>>`
4. Add `RenderedModule#code()` to return `Option<String>` which will turn `inner_code` to actually code/string.
5. `BindingRenderedModule#code` should be a getter, which calls `RenderedModule#code()` internally.
6. `BindingRenderedModule#renderedLength` is a pure js side getter that returns `this.code.length`.

hyf0:
Let me do some chages.

hyf0 (approved):
Thanks for the fix. I was thinking how to solve these test problems automatically. It's too cumbersome to manually update these tests.

PR #2709, opened 2024-11-12
fix:  cli doesn't show config loading error details

IWANABETHATGUY:
### Description
1. related to https://github.com/rolldown/rolldown/issues/2705, as 
https://github.com/rolldown/rolldown/pull/2707#issuecomment-2470659332 said, this is caused by `consola` not support display `Error.cause`

### DIFF
--- a/packages/rolldown/src/cli/index.ts
+++ b/packages/rolldown/src/cli/index.ts
@@ -28,6 +28,6 @@ async function main() {
 }
 
 main().catch((err) => {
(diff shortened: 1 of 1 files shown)


hyf0 (approved):

IWANABETHATGUY:
This pr didn't solve the issues, but at least display the original stack trace, we should replace `consola` in other Logger, but I don't have enough bandwidth to do that, this is just a hotfix.

PR #2717, opened 2024-11-13
fix(umd): check name option if chunk has exports

underfin:


crates/rolldown/tests/rolldown/function/export_mode/umd/auto/none/_config.json
```diff
@@ -1,7 +1,6 @@
 {
   "config": {
     "exports": "auto",
-    "format": "umd",
-    "name": "module"
+    "format": "umd"
   }
```
underfin, line 5: Before remove the `name` the test is failed.

hyf0 (approved):

PR #2720, opened 2024-11-13
feat: dynamic import treeshaking

IWANABETHATGUY:
### Description
1. part 1, only tree shake 
```js
import('mod').then(mod => {
  mod.xxx
})
```
bailout destruct pattern for now


hyf0:
How this feature works with `inlineDynamicImports`?

- Does enabling `inlineDynamicImports` break this?
- Do we need to disable this if `inlineDynamicImports: true`.

crates/rolldown_common/src/module_loader/task_result.rs
```diff
@@ -12,4 +14,5 @@ pub struct NormalModuleTaskResult {
   pub warnings: Vec<BuildDiagnostic>,
   pub module: Module,
   pub ecma_related: Option<(EcmaAst, SymbolRefDbForModule)>,
+  pub dynamic_import_rec_exports_usage: FxHashMap<ImportRecordIdx, DynamicImportExportsUsage>,
```
hyf0, line 17: Maybe put this in `ecma_related` fields?

IWANABETHATGUY:
> How this feature works with `inlineDynamicImports`?
> 
> * Does enabling `inlineDynamicImports` break this?
> * Do we need to disable this if `inlineDynamicImports: true`.

test added

hyf0 (approved):
Don't see anything wrong. Good job!

packages/rolldown/tests/fixtures/tree-shake/inline-dynamic-imports/_config.ts
```diff
@@ -0,0 +1,14 @@
+import { defineTest } from '@tests'
+import { expect } from 'vitest'
+
+export default defineTest({
+  config: {
+    output: {
+      inlineDynamicImports: false,
```
hyf0:
```suggestion
      inlineDynamicImports: true,
```

`inlineDynamicImports` is `false` by default. I assume you want to test the `true` condition?

crates/rolldown_common/src/module_loader/task_result.rs
IWANABETHATGUY: This is temp data structure, it will be normalized to `FxHashmap<ModuleId, DynamicUsage>` after fetching all modules, I don't see any usage related to specific modules, so I don't add this field to `ecma_view`. Also only those modules are used as a dynamic entry will have a none empty field, for most of the modules this field is just a waste of memory.

packages/rolldown/tests/fixtures/tree-shake/inline-dynamic-imports/_config.ts
IWANABETHATGUY: Nice catch

crates/rolldown_common/src/module_loader/task_result.rs
hyf0: > so I don't add this field to ecma_view.

`ecma_related` is not `ecma_view`.

> for most of the modules this field is just a waste of memory.

I'm thinking the same with you. Most of the module tasks don't need to create `dynamic_import_rec_exports_usage: FxhashMap` .

IWANABETHATGUY: > Maybe put this in `ecma_related` fields?

looks good, will refactor it in next pr.

PR #2738, opened 2024-11-14
fix: align resolve behavior of `new URL(..., import.met.url)` with webpack

hyf0:
### Description

So the behavior is for bare identifier, bundler will resolve it as relative path first instead of npm modules. It means that `foo` would be resolve to `./foo.js` first. It'll hide the npm module `foo`.

More contexts:

- To prevent the above behavior, parcel supports [`npm:xxx` pattern](https://parceljs.org/blog/v2/#new-npm%3A-scheme) to force resolve to node modules first. Eg: `new URL("npm:foo", import.meta.url)`. 

This behavior is not supported by rolldown. We'll wait users feedbacks and see if we need to support this feature.

### DIFF
--- a/crates/rolldown/src/ast_scanner/new_url.rs
+++ b/crates/rolldown/src/ast_scanner/new_url.rs
@@ -34,7 +34,7 @@ impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
     };
     let path = &first_arg_string_literal.value;
     let idx =
-      self.add_import_record(path, ImportKind::Import, expr.span, ImportRecordMeta::empty());
+      self.add_import_record(path, ImportKind::NewUrl, expr.span, ImportRecordMeta::empty());
     self.result.import_records[idx].asserted_module_type = Some(ModuleType::Asset);
     self.result.new_url_references.insert(expr.span, idx);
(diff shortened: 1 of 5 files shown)


hyf0:
* **#2738** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2738?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2738?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

hyf0:
Behavior of this PR/Webpack's behavior should be a superset of vite's I think. So it should be compatible with vite's current behavior by defualt. @sapphi-red cc

sapphi-red:
> Behavior of this PR/Webpack's behavior should be a superset of vite's I think.

I think Vite does the same. (Vite is a bit strict about extensions though).
https://github.com/vitejs/vite/blob/cf813cc5db06907a33cdf0793299ece3540178d2/packages/vite/src/node/plugins/assetImportMetaUrl.ts#L118-L123

PR #2739, opened 2024-11-15
feat: vite-resolve-plugin dev support

sapphi-red:


crates/rolldown_binding/src/options/plugin/binding_builtin_plugin.rs
```diff
@@ -44,7 +44,7 @@ impl std::fmt::Debug for BindingBuiltinPlugin {
 }
 
 #[allow(clippy::enum_variant_names)]
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Deserialize, Clone)]
```
IWANABETHATGUY: seems `Copy` is enough?

sapphi-red: This diff was needed while I was iterating but it seems this is not needed anymore.

IWANABETHATGUY (approved):

PR #2756, opened 2024-11-15
fix: keep external module live bindings at format cjs/umd/iife

underfin:


crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
```diff
@@ -606,7 +602,21 @@ impl<'a> BindImportsAndExportsContext<'a> {
...
-        ImportStatus::External => todo!(),
+        ImportStatus::External(symbol_ref) => {
+          if self.options.format.keep_esm_import_export_syntax() {
+            // Imports from external modules should not be converted to CommonJS
+            // if the output format preserves the original ES6 import statements
+            break MatchImportKind::Normal { symbol: tracker.imported_as };
+          }
+
+          match &tracker.imported {
+            Specifier::Star => MatchImportKind::Namespace { namespace_ref: symbol_ref },
```
underfin, line 611: Here esbuild using `namedImport.NamespaceRef`, ref https://github.com/evanw/esbuild/blob/main/internal/linker/linker.go#L2639. 

But it caused generate different external namespace, example `import_external/import_external2/import_external3`

``` .js
var import_external = require("external");
var import_external2 = require("external");
var import_external3 = __toESM(require("external"));
console.log(import_external.a, import_external2.b, import_external3.default);
```

So here i using `External#symbol_ref` to make it from one symbol. Acutally here i using `namedImport.NamespaceRef` at first, but it generated many import stmts from external, like esbuild.

crates/rolldown/src/module_loader/module_loader.rs
```diff
@@ -178,11 +179,20 @@ impl ModuleLoader {
...
+          let symbol_ref = self.symbol_ref_db.create_facade_root_symbol_ref(
+            idx,
+            legitimize_identifier_name(resolved_id.id.as_str()).into(),
+          );
           let ext = ExternalModule::new(
             idx,
             ArcStr::clone(&resolved_id.id),
             external_module_side_effects,
-            SymbolNameRefToken::new(idx, legitimize_identifier_name(&resolved_id.id).into()),
+            symbol_ref,
```
underfin, line 195: Here pervious using `SymbolNameRefToken`, it is different with `SymbolRef`, because here need to linker it to `NamespaceAlias#SymbolRef`, so here reback using `SymbolRef`.

here we could to change `NamespaceAlias#SymbolRef` to  enum `SymbolNameRefToken/SymbolRef` to be avoid here using `SymbolRef`. It is some complex so it is not including it at here.

crates/rolldown/tests/esbuild/default/export_wildcard_fs_node_common_js/artifacts.snap
```diff
@@ -8,18 +8,18 @@ snapshot_kind: text
...
+Object.keys(___external$1).forEach(function (k) {
   if (k !== 'default' && !Object.prototype.hasOwnProperty.call(exports, k)) Object.defineProperty(exports, k, {
     enumerable: true,
-    get: function () { return ___external[k]; }
+    get: function () { return ___external$1[k]; }
   });
 });
-var fs = require("fs");
-Object.keys(fs).forEach(function (k) {
+var fs$1 = require("fs");
```
IWANABETHATGUY: unexpected

crates/rolldown/src/utils/renamer.rs
```diff
@@ -108,6 +108,7 @@ impl<'name> Renamer<'name> {
     conflictless_name.to_string()
   }
 
+  #[allow(dead_code)]
   pub fn add_symbol_name_ref_token(&mut self, token: &SymbolNameRefToken) {
```
IWANABETHATGUY, line 112: Why do we need this?

crates/rolldown/tests/esbuild/default/export_wildcard_fs_node_common_js/artifacts.snap
underfin: Yeah. Here has some the perivous implement not delete, let me update it.

crates/rolldown/src/utils/renamer.rs
underfin: reason is here https://github.com/rolldown/rolldown/pull/2756#discussion_r1843979416

crates/rolldown/tests/esbuild/default/import_fs_node_common_js/artifacts.snap
```diff
@@ -9,10 +10,10 @@ source: crates/rolldown_testing/src/integration_test.rs
 "use strict";
 
 const fs = __toESM(require("fs"));
-const { default: defaultValue, readFileSync } = __toESM(require("fs"));
+const fs = __toESM(require("fs"));
```
IWANABETHATGUY, line 13: this should be deconflicted

crates/rolldown/tests/esbuild/default/export_wildcard_fs_node_common_js/artifacts.snap
underfin: The `LinkingMetadata#require_bindings_for_star_exports` is unless, here removed it.

crates/rolldown/tests/esbuild/default/import_fs_node_common_js/artifacts.snap
underfin: It cuased by render twice with import star stmt and import named stmt. The pr already two larger, i will fix it at next.

hyf0:
Some notes:

```
issues of cjs format and external module
	- Hard-coded support for code splitting in cjs format
		- Too keep live binding, speicifer need to be rewritten to property access
	- Hard-coded support for external module
		- Still needs to be rewritten to property access
			- This so far rely on create `SymbolRef` for external module, which shouldn't have symbols in fact.
	- imports for external module doesn't get merged under the current symbol system ported from esbuild
```

hyf0 (approved):

crates/rolldown/tests/esbuild/default/import_fs_node_common_js/artifacts.snap
hyf0: https://github.com/rolldown/rolldown/issues/2768

hyf0:
Try to make a series of small PRs if possible.

PR #2763, opened 2024-11-16
perf: improve the String concatenation performance

Brooooooklyn:
Considering
```rust
  let base64 = "base64".repeat(1000);

  let mime_ext = MimeExt { mime: "image/png".parse().unwrap(), is_utf8_encoded: true };

  // Mix of String and str
  group.bench_function("mixed_types_concat", |b| {
    let mime_ext_string = mime_ext.to_string();
    b.iter(|| black_box(concat_string!("data:", mime_ext_string, ";base64,", base64)));
  });
  group.bench_function("mixed_types_format", |b| {
    b.iter(|| black_box(format!("data:{mime_ext};base64,{base64}")));
  });
```

`concat_string` is 2x faster than `format!`

```
concat_string/mixed_types_concat
                        time:   [98.035 ns 101.56 ns 105.10 ns]
                        change: [+8.7720% +13.706% +18.797%] (p = 0.00 < 0.05)
                        Performance has regressed.
concat_string/mixed_types_format
                        time:   [217.54 ns 221.50 ns 225.72 ns]
                        change: [+2.2251% +4.5076% +6.7315%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

### DIFF
--- a/crates/rolldown/src/ecmascript/format/esm.rs
+++ b/crates/rolldown/src/ecmascript/format/esm.rs
@@ -2,6 +2,7 @@ use arcstr::ArcStr;
 use itertools::Itertools;
 use rolldown_common::{ChunkKind, ExportsKind, Module, WrapKind};
 use rolldown_sourcemap::SourceJoiner;
+use rolldown_utils::concat_string;
 
 use crate::{
   ecmascript::ecma_generator::RenderedModuleSources,
@@ -53,8 +54,7 @@ pub fn render_esm<'code>(
           })
           .dedup()
           .for_each(|ext_name| {
-            let import_stmt = format!("export * from \"{}\"\n", &ext_name);
-            source_joiner.append_source(import_stmt);
+            source_joiner.append_source(concat_string!("export * from \"", ext_name, "\"\n"));
           });
       }
     }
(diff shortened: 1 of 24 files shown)


Brooooooklyn:
* **#2763** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2763?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2763?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

crates/rolldown_utils/src/concat_string.rs
```diff
@@ -0,0 +1,77 @@
+// https://github.com/FaultyRAM/concat-string
```
hyf0, line 1: How about using this crate directly?

hyf0:
Do you think you could merge this PR to your branch? https://github.com/rolldown/rolldown/pull/2764. 

It's about suggesting people  should use  `concat-string` instead of `format!` if possible.

Note: Words may need to be modified, if we don't use `concat-string` crate.

crates/rolldown/src/ecmascript/format/esm.rs
```diff
@@ -138,7 +144,11 @@ fn render_esm_chunk_imports(ctx: &GenerateContext<'_>) -> String {
         }
       }
       RenderImportDeclarationSpecifier::ImportStarSpecifier(alias) => {
-        s.push_str(&format!("import * as {alias} from \"{path}\";\n",));
+        s.push_str("import * as ");
```
hyf0, line 147: Maybe a helper method like

```rust
trait StringExt {
  fn push_strs((&mut self, strs: &[&str]) {
    let size_hint = ...;
    self.reserve_addiational();
    // push_str
  }  
}
}
```

Macro would be better for this due to this will create a temporary array.

hyf0:
You could run `just test update` to update all snapshots both in rust and nodejs.

crates/rolldown/src/ecmascript/format/esm.rs
Brooooooklyn: Let's do it in a separate pr

hyf0 (approved):

PR #2769, opened 2024-11-17
fix(interop): improve compatibility of esm imports cjs

hyf0:
### Description

In nodejs

```
import foo from './foo.cjs'
console.log(foo)
```

`foo` == `module.exports` in `foo.cjs`

---

In rolldown,

```
import foo from './foo.cjs'
console.log(foo)
```

is converted to

```
console.log(foo.default)
```

Without this PR,

`foo.default` == `module.exports.default`

With this PR,

`foo.default` == `module.exports`


hyf0:
* **#2770** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2770?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2769** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2769?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2769?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

crates/rolldown/tests/rolldown/function/resolve/browser_filed_false/artifacts.snap
```diff
@@ -13,7 +14,7 @@ var require_util = __commonJS({ "node_modules/package/util.js"() {} });
 
 //#endregion
 //#region node_modules/package/index.js
-var import_util = __toESM(require_util());
+var import_util = __toESM(require_util(), 1);
```
hyf0, line 17: This result is wired but expected. This PR is not the root cause. It's because the way we deal with "false browser field" is different as esbuild.

Also the runtime behavior is not changed since `false`ed module is empty. I'm not gonna touch it on this PR.

crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
```diff
@@ -532,7 +538,13 @@ impl<'me, 'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'me, 'ast> {
                   )),
                 );
               }
-              WrapKind::None => {}
+              WrapKind::None => {
+                // The nature of `import()` is to load the module dynamically/lazily, so imported modules would
+                // must be wrapped, so we could make sure the module is executed lazily.
+                if cfg!(debug_assertions) {
```
IWANABETHATGUY, line 544: why only `unreachable` in `debug` mode?

IWANABETHATGUY (approved):

crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
hyf0: Don't want panic happen in release mode. And this also keep the code behave like the old code. I haven't sure if we should panic always in such patterns. I could change it once we come up with the conclusion.

PR #2777, opened 2024-11-18
fix: avoid unresolved promise if load same id at concurrent

### LINKED ISSUE #2751: [Bug]: gets stuck when calling `this.load` with same id in parallel

1. Open https://stackblitz.com/edit/github-xv4yss-qaagcu?file=rolldown.config.mjs,package.json
3. Rolldown command does not end

Note this happens not only on Stackblitz but also on Windows as well.

### What is expected?

The build finishes.

The build does not finish.

### System Info

```Shell
Stackblitz

  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    rolldown: https://pkg.pr.new/rolldown@cce5cde => 0.14.0
```

### Any additional comments?

I found this when I tried to enable some disabled tests in rolldown-vite.

underfin:
### Description

close https://github.com/rolldown/rolldown/issues/2751

### DIFF
--- a/packages/rolldown/src/plugin/plugin-context.ts
+++ b/packages/rolldown/src/plugin/plugin-context.ts
@@ -74,17 +74,32 @@ export class PluginContext extends MinimalPluginContext {
         moduleSideEffects: options.moduleSideEffects || null,
       }
       data.updateModuleOption(id, rawOptions)
-      let resolveFn
-      // TODO: If is not resolved, we need to set a time to avoid waiting.
-      const promise = new Promise((resolve, _) => {
-        resolveFn = resolve
-      })
-      await context.load(
-        id,
-        bindingifySideEffects(options.moduleSideEffects),
-        resolveFn!,
-      )
-      await promise
+
+      async function createLoadModulePromise() {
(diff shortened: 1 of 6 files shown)


packages/rolldown/src/plugin/plugin-context.ts
```diff
@@ -74,17 +74,32 @@ export class PluginContext extends MinimalPluginContext {
...
-      })
-      await context.load(
-        id,
-        bindingifySideEffects(options.moduleSideEffects),
-        resolveFn!,
-      )
-      await promise
+
+      async function createLoadModulePromise() {
+        if (data.loadModulePromiseMap.has(id)) {
```
IWANABETHATGUY: get and checking if it is null, could avoid hash twice.

underfin: Done.

IWANABETHATGUY:
I try the issue case locally,  it does not seems to work, still stuck when bundling

IWANABETHATGUY (approved):

PR #2780, opened 2024-11-18
refactor: render chunk imports for format esm

underfin:
### Description

Consider the external imports is different at different format, Here separate it to avoid over-abstraction, i will remove `collect_render_chunk_imports` usage.
# Esm 
- The import star decl need to generate 
- The import named decl need to generate
# Other format
only have generate one stmt, like `const external = require('external')`


hyf0 (approved):

underfin:
Draft it, look like the performance has regression.

underfin:
> Draft it, look like the performance has regression.

Re-run it the result look like is expected, here also haven't break performance things.

PR #2798, opened 2024-11-19
refactor: scope hoisting finalizer

IWANABETHATGUY:

### DIFF
--- a/crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
+++ b/crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
@@ -202,12 +202,11 @@ impl<'me, 'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'me, 'ast> {
     // 3. hoisted_names
     // 4. wrapped module declaration
     let declaration_of_module_namespace_object = if is_namespace_referenced {
-      let mut stmts = self.generate_declaration_of_module_namespace_object();
+      let stmts = self.generate_declaration_of_module_namespace_object();
       if needs_wrapper {
         stmts
       } else {
-        stmts.extend(program.body.take_in(self.alloc));
-        program.body.extend(stmts);
+        program.body.splice(0..0, stmts);
         vec![]
       }
     } else {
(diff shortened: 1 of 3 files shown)


IWANABETHATGUY:
* **#2799** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2799?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2798** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2798?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2798?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

crates/rolldown_common/src/types/symbol_ref.rs
```diff
@@ -29,34 +29,26 @@ impl SymbolRef {
 
   /// Not all symbols have flags info, we only care about part of them.
   /// If you want to ensure the flags info exists, use `flags_mut` instead.
-  pub fn flags<'db>(&self, db: &'db dyn GetLocalDb) -> Option<&'db SymbolRefFlags> {
+  pub fn flags<'db, T: GetLocalDb>(&self, db: &'db T) -> Option<&'db SymbolRefFlags> {
```
hyf0, line 32: NIT: Previous code is using syntax sugar and should perform the same as changed one. ~~Not sure why we want to desugar it~~. What's the reason of desugar it?

crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
```diff
@@ -202,12 +202,11 @@ impl<'me, 'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'me, 'ast> {
...
     // 4. wrapped module declaration
     let declaration_of_module_namespace_object = if is_namespace_referenced {
-      let mut stmts = self.generate_declaration_of_module_namespace_object();
+      let stmts = self.generate_declaration_of_module_namespace_object();
       if needs_wrapper {
         stmts
       } else {
-        stmts.extend(program.body.take_in(self.alloc));
-        program.body.extend(stmts);
+        program.body.splice(0..0, stmts);
```
hyf0, line 209: NIT: would

```rust
program.body.clear();
program.body.extends(stmts)
```

more readable and perform the same.

hyf0 (approved):

crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
IWANABETHATGUY: maybe not the same thing, `splice(0..0, stmts)` prepend all `stmts` into `program.body`, let me add some comments.

crates/rolldown_common/src/types/symbol_ref.rs
IWANABETHATGUY: https://doc.rust-lang.org/std/keyword.dyn.html, `dyn` is dynamically dispatched, generic param should be more performant.

hyf0: I thought that `&dyn trait` would be handled specially compared to `Box<dyn trait>`, which rustc would chose to use static dispatch for `&dyn trait` if it could be infered. But I can't find materilas to prove this. Guess I thought it wrong. Glad to know that from this conversation.

IWANABETHATGUY: Yeah, I don't find any source to say that either, but we could saw the asm https://godbolt.org/z/dWK6nc8hr, you could see for `test` function it has an indirect `call` instruction

PR #2565, opened 2024-11-02
feat(rust): support `file` option

hyf0:

PR #2800, opened 2024-11-19
feat: use `BuildResult` for `validate_options_for_multi_chunk_output`

shulaoda:
### Description

Related to #2565

https://github.com/rolldown/rolldown/blob/d796b1efa9bd973b60e012ff5bccdb05ee157593/crates/rolldown/src/utils/chunk/validate_options_for_multi_chunk_output.rs#L3-L6


underfin (approved):

hyf0:
Thanks

PR #2803, opened 2024-11-20
perf:  ast scanner `is_top_level`

IWANABETHATGUY:


crates/rolldown/src/ast_scanner/mod.rs
```diff
@@ -174,6 +174,7 @@ impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
       .scope_stack
       .iter()
       .filter_map(|item| *item)
+      .rev()
```
IWANABETHATGUY, line 177: scope_stack is from `root` -> `the most nested scope`, if the top of the stack is not top level, we don't need to iterate anymore.

hyf0 (approved):

PR #2808, opened 2024-11-20
feat: enable `inlineDynamicImports` automatically if format is `iffe` or `umd`

### LINKED ISSUE #2792: [Feature Request]: Enable `inlineDynamicImports` automatically if format is `iffe` or `umd`

Improve UX.

This issue includes two main tasks:
- Set `inlineDynamicImports: true` if format is `iffe` or `umd`  https://github.com/rolldown/rolldown/blob/8ef1b106a4796c54a04d018e317a29c41cb75009/crates/rolldown/src/utils/normalize_options.rs#L133
- If `inlineDynamicImports` is set to `false` by user, a warning should be emitted and we still set `inlineDynamicImports` to be `true`.

shulaoda:
### Description

closes #2792 

Currently, there are some issues with triggering the warning.


crates/rolldown_binding/src/bundler.rs
```diff
@@ -66,12 +67,18 @@ impl Bundler {
...
+    let bundler = Self {
       cwd: ret.bundler_options.cwd.clone().unwrap_or_else(|| std::env::current_dir().unwrap()),
       inner: Arc::new(Mutex::new(NativeBundler::with_plugins(ret.bundler_options, ret.plugins))),
       log_level,
       on_log,
-    })
+    };
+
+    if !ret.warnings.is_empty() {
+      block_on(bundler.handle_warnings(ret.warnings));
```
hyf0: You don't need to handle warnings. Just make it exist in the final `BundleOutput#warnings`. It gets printed automatically by handlers in `rolldown_binding`/`integrstion_test.rs`.

shulaoda: We cannot store `warnings` within `Bundler::new`. Should we consider adding a `warnings` field to the struct `Bundler` or `NativeBundler`?

hyf0: Yeah. Adding a `Bundler#warnings` would be fine.

shulaoda: I eventually added it to `NativeBundler` because its methods are mutable and common for both `rolldown_binding` and `integration_test.rs`.

crates/rolldown/src/bundler.rs
```diff
@@ -120,6 +123,8 @@ impl Bundler {
       .build_end(error_for_build_end_hook.map(|error| HookBuildEndArgs { error }).as_ref())
       .await?;
 
+    scan_stage_output.warnings.append(&mut self.warnings);
```
hyf0: You should append these warnings in the end of `write`/`generate`. It's more predictable.

crates/rolldown/src/utils/normalize_options.rs
```diff
@@ -83,6 +89,24 @@ pub fn normalize_options(mut raw_options: crate::BundlerOptions) -> NormalizeOpt
     experimental.strict_execution_order = Some(true);
   }
 
+  let inline_dynamic_imports = match raw_options.format {
```
hyf0: Normalize raw format first and match normalized format. No need to dealing with raw format here.

hyf0 (changes requested):

hyf0 (approved):

hyf0:
Thanks

crates/rolldown/src/bundler.rs
shulaoda: I'm not sure whether this should be added here, because if we are simply using the `scan` exposed by `napi`, this warning won't actually be exposed.

https://github.com/rolldown/rolldown/blob/a64f4825e7359dc5474a8e2fc78ce4ef144561d3/crates/rolldown_binding/src/bundler.rs#L115-L119

hyf0: 👌Get it

PR #2809, opened 2024-11-20
docs: update bluesky link

gautemo:
### Description

Bluesky link on https://rolldown.rs/ points towards old handle. This PR solves this by updating the URL to https://bsky.app/profile/rolldown.rs.


hyf0 (approved):
Thanks!

PR #2810, opened 2024-11-20
chore: make watcher test stable

underfin:
### Description

Here using retry sometimes to make the test stable. The `perf)watching same file multiply times` is removed, because it is meaningless at here, i will add benchmark to instead of it at next.


hyf0 (approved):

hyf0:
Directly merged to try avoid unstable tests.

PR #2818, opened 2024-11-20
feat: support function for `output.globals`

### LINKED ISSUE #1925: [Feature Request]: Align `output.globals` with Rollup.

In Rollup, the `output.globals` option supports not only the HashMap (`{ [id: string]: string }` in TypeScript, `HashMap` in Rust) but also requires the function (`(id: string) => string`). While implementing this may be challenging, you can refer to the `external` option in the `input` option.

Perhaps we should consider adding a `contrib: hard` tag if this functionality is required.

For instance, if only external modules like `node:path` are present, transform all of these external modules into `path` within an IIFE.
```ts
export default defineConfig({
    external: (id: string) => "path",
})
```

shulaoda:
### Description

closes #1925

Align with [rollup#output-globals](https://rollupjs.org/configuration-options/#output-globals)

https://github.com/rolldown/rolldown/blob/ab438c355063765640db18cb03b8d08fa2eee39d/packages/rolldown/src/utils/normalize-output-options.ts#L46-L47

### DIFF
--- a/crates/rolldown/src/ecmascript/format/iife.rs
+++ b/crates/rolldown/src/ecmascript/format/iife.rs
@@ -34,7 +34,6 @@ use crate::{
   },
 };
 use arcstr::ArcStr;
-use futures::executor::block_on;
 use rolldown_common::{ChunkKind, ExternalModule, OutputExports, WrapKind};
 use rolldown_error::{BuildDiagnostic, BuildResult};
 use rolldown_sourcemap::SourceJoiner;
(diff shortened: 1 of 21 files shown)


crates/rolldown/src/ecmascript/format/iife.rs
```diff
@@ -201,17 +202,20 @@ fn render_iife_factory_arguments(
...
-      let target = legitimize_identifier_name(&external.name).to_string();
-      warnings.push(
-        BuildDiagnostic::missing_global_name(external.name.clone(), ArcStr::from(&target))
-          .with_severity_warning(),
-      );
-      factory_arguments.push(target);
-    }
-  });
+  for external in externals {
+    let global = block_on(globals.call(external.name.as_str()));
```
underfin: Here using `block_on` is not expected, it could be using `async/await` to do it. But it looks like caused by `SourceJoiner<'source>` is not `Send`. Let us address it at next.

hyf0: Pre-generates these and pass them as args into the `render_iife` function.

shulaoda: > Pre-generates these and pass them as args into the `render_iife` function.

I found that doing this would complicate the logic of `render_iife` and `render_umd`, so I simply implemented the `Send` trait for `SourceJoiner`.

underfin (approved):

crates/rolldown/src/ecmascript/format/iife.rs
underfin: Good jod, thank you.

PR #2823, opened 2024-11-21
refactor: watcher emitter

### LINKED ISSUE #2806: [Bug]: dead lock happend at `WatcherEmitter#emit` and `WatcherEmitter#on`

 Is there a way to provider a channel by napi, the rust side emit event, the js side lister the event. I meet the deadlock problem at `WatcherEmitter#emit` and `WatcherEmitter#on` , the code at https://github.com/rolldown/rolldown/blob/main/crates/rolldown/src/watcher/emitter.rs.

The production case: 

``` .js
import { watch } from 'rolldown'

const watcher = await watch({
  input: ['./index.js'],
})

watcher.on('event', (event) => {
    if (event.code === 'BUNDLE_END') {
      watcher.on('event', () => console.log(111)) // dead lock, try to get  lock of `WatcherEmitter#on`, but the rust side lock of `WatcherEmitter#emit` not released.
    }
});
```

### What is expected?

Should not deadlock.

Happened deadlock.

### System Info

```Shell
-
```

### Any additional comments?

_No response_

underfin:
### Description

close https://github.com/rolldown/rolldown/issues/2806

Using `channel` to do it. when receiving msg only call binding function once.


hyf0 (approved):

underfin:
Draft it here not sure the `notify` is lost event at test `PluginContext.addWatchFile` at macOS.

Issue #2835, opened 2024-11-21
Remove all options normalization in js side and everything should work as usual


hyf0:
This is preparation for https://github.com/rolldown/rolldown/issues/1041.

https://github.com/rolldown/rolldown/blob/0b8dbaa8548bd64f59a1281bcd5f8ca6326d3935/packages/rolldown/src/utils/normalize-output-options.ts#L30-L52

- All these `??` should be removed.

https://github.com/rolldown/rolldown/blob/0b8dbaa8548bd64f59a1281bcd5f8ca6326d3935/packages/rolldown/src/options/normalized-output-options.ts#L8-L19

- Inconsistent `NormalizedOutputOptions` compared to rollup is acceptable for now.

See https://github.com/rolldown/rolldown/pull/2834 for example.

shulaoda:
I'm free to do this.

hyf0:
@shulaoda Try make smaller PRs for each option. No need to solve them all in one PR.

sxzz:
@hyf0 

This change could be considered a major update that breaks https://github.com/unplugin/unplugin-isolated-decl

1. This plugin didn't treat Rolldown specially but regarded it as Rollup.
2. It relies on the default value of `NormalizedOutputOptions.entryFileNames`, which is `[name].js` in Rollup, but now it's undefined.
3. `NormalizedInputOptions.input` is `string[] | Record<string, string>` in Rollup and can be a `string` in Rolldown.

If this is by design, I hope the Rolldown team will reconsider.

hyf0:
> @hyf0 
> 
> This change could be considered a major update that breaks https://github.com/unplugin/unplugin-isolated-decl
> 
> 1. This plugin didn't treat Rolldown specially but regarded it as Rollup.
> 2. It relies on the default value of `NormalizedOutputOptions.entryFileNames`, which is `[name].js` in Rollup, but now it's undefined.
> 3. `NormalizedInputOptions.input` is `string[] | Record<string, string>` in Rollup and can be a `string` in Rolldown.
> 
> If this is by design, I hope the Rolldown team will reconsider.

These options would be supported after https://github.com/rolldown/rolldown/issues/1041. We still pass these values to js side.

PR #2836, opened 2024-11-21
feat: don't normalize options on js side - part1

shulaoda:
### Description

Related to #2835 

@hyf0 I've noticed that modifying each option seems to be quite simple. Is there anything else I haven't considered? Do I still need to break them down into individual smaller PRs?


hyf0:
> I've noticed that modifying each option seems to be quite simple.

Yes, you could raise one PR to solve them.

shulaoda:
I might need to split this into two PRs. The main goal is to remove `normalizeOptions` on js side, right? I will remove these two functions in the next PR.

hyf0 (approved):
Cool

PR #2846, opened 2024-11-21
chore(rolldown): oxc v0.37.0

Boshen:


hyf0:
You need to run `just ued` to update `diff.md` files. It's used to compared rolldown's output with esbuild's.

hyf0 (approved):

Issue #1041, opened 2024-05-06
Implement `NormalizedInputOptions` and `NormalizedOutputOptions` to rust side


hyf0:
## Why

The main problem is that the concept `NormalizedInputOptions` is try to give a final form of options while it still in Js side not passing to the rust.

Take `external` option as an example, `NormalizedInputOptions` require `external` option to be a `function`, which means though users passed static `['external']` data and it would still needs be wrapped in a js function, that's bad for performance. If `external` option is setted, every resolve process would need to call Js function in sequence.

## How

I looked into the ``NormalizedInputOptions` and delete we highly not going to support. We got

```js
export interface NormalizedInputOptions {
  // acorn: Record<string, unknown>
  // acornInjectPlugins: (() => unknown)[]
  // cache: false | undefined | RollupCache
  context: string
  experimentalCacheExpiry: number
  experimentalLogSideEffects: boolean
  external: IsExternal
  /** @deprecated Use the "inlineDynamicImports" output option instead. */
  inlineDynamicImports: boolean | undefined
  input: string[] | { [entryAlias: string]: string }
  logLevel: LogLevelOption
  makeAbsoluteExternalsRelative: boolean | 'ifRelativeSource'
  /** @deprecated Use the "manualChunks" output option instead. */
  manualChunks: ManualChunksOption | undefined
  // maxParallelFileOps: number
  /** @deprecated Use the "maxParallelFileOps" option instead. */
  // maxParallelFileReads: number
  moduleContext: (id: string) => string
  onLog: LogHandler
  onwarn: (warning: RollupLog) => void
  perf: boolean
  plugins: Plugin[]
  preserveEntrySignatures: PreserveEntrySignaturesOption
  /** @deprecated Use the "preserveModules" output option instead. */
  preserveModules: boolean | undefined
  preserveSymlinks: boolean
  shimMissingExports: boolean
  strictDeprecations: boolean
  treeshake: false | NormalizedTreeshakingOptions
}
```

The problem here is that for properties whose value is function is complicated/impossible to pass them from rust to js, so we only have

```js
export interface NormalizedInputOptions {
  context: string
  experimentalCacheExpiry: number
  experimentalLogSideEffects: boolean
  /** @deprecated Use the "inlineDynamicImports" output option instead. */
  inlineDynamicImports: boolean | undefined
  input: string[] | { [entryAlias: string]: string }
  logLevel: LogLevelOption
[...]

PR #2848, opened 2024-11-21
feat: don't normalize options on js side - part2

shulaoda:
### Description

Related to #2835 

I found that removing `normalizeInputOptions` is a bit tricky because it is tightly coupled with the `Plugin` on the js side. 

I reviewed the context of issue #1041, but I'm still unsure if I'm on the right track. Therefore, I'm submitting this PR to ensure that what I'm doing is correct. I would appreciate it if you could point out the next steps I should take.

> I also think about it before, we could using original `external` option to improve it. If the user not changed normalized `external` option, we just using original `external` option to binding option, it could be `array|function`.
> 
> 
> 
> I think about this too. But it requires us to change the definition of `NormalizedInputOptions`. If this is acceptable, we could just use `InputOptions` as the `NormalizedInputOptions`. I kind of feel that this makes the normalization step in js side meaningless, so I didn't propose this way.
 _Originally posted by @hyf0 in [#1041](https://github.com/rolldown/rolldown/issues/1041#issuecomment-2098313025)_


hyf0 (approved):
Good works

PR #2858, opened 2024-11-22
feat: don't normalize options on js side - part3

shulaoda:
### Description

Related to #2835 

The `onLog` has a deep coupling with `PluginContext`, which might make it more complex, so I'm not sure if it's worth doing. 

The simpler approach is to directly add a static `onLog` method to `PluginContext`.


packages/rolldown/src/utils/normalize-tree-shake.ts
```diff
@@ -1,32 +0,0 @@
...
-      moduleSideEffects: true,
-      annotations: true,
-    }
-  }
-  let normalizedConfig: NormalizedTreeshakingOptions = {
-    moduleSideEffects: true,
-  }
-  if (config.moduleSideEffects === undefined) {
-    normalizedConfig.moduleSideEffects = true
-  } else if (config.moduleSideEffects === 'no-external') {
```
IWANABETHATGUY, line 21: I don't think this is correct.

hyf0 (approved):

PR #2870, opened 2024-11-22
feat: don't normalize options on js side - part4

### LINKED ISSUE #2835: Remove all options normalization in js side and everything should work as usual

shulaoda:
### Description

closes #2835

I'm uncertain whether we should remove the current `NormalizedInputOptions` and `NormalizedOutputOptions` and replace them with `InputOptions` and `OutputOptions`. If this change is necessary, it would be the final part.


hyf0 (approved):
> I'm uncertain whether we should remove the current NormalizedInputOptions and NormalizedOutputOptions and replace them with InputOptions and OutputOptions

No needed.

`NormalizedInputOptions` `NormalizedOutputOptions` will be refactored in https://github.com/rolldown/rolldown/issues/1041. You could leave it to me. Great works!

PR #2893, opened 2024-11-24
refactor: use `BuildResult` for `PreProcessEcmaAst`

shulaoda:
### Description

I'm not sure if this will break the original design intent, so feel free to close it at any time.

@IWANABETHATGUY cc


IWANABETHATGUY:
That looks better, thanks!

IWANABETHATGUY (approved):

PR #2913, opened 2024-11-25
fix: set external module info

underfin:

### DIFF
--- a/crates/rolldown_common/src/types/module_info.rs
+++ b/crates/rolldown_common/src/types/module_info.rs
@@ -12,3 +12,17 @@ pub struct ModuleInfo {
   pub imported_ids: Vec<ModuleId>,
   pub dynamically_imported_ids: Vec<ModuleId>,
 }
+
+impl From<ModuleId> for ModuleInfo {
+  fn from(id: ModuleId) -> Self {
+    Self {
+      code: None,
+      id,
+      is_entry: false,
+      importers: Vec::new(),
+      dynamic_importers: Vec::new(),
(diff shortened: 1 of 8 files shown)


crates/rolldown_common/src/types/module_info.rs
```diff
@@ -12,3 +12,17 @@ pub struct ModuleInfo {
   pub imported_ids: Vec<ModuleId>,
   pub dynamically_imported_ids: Vec<ModuleId>,
 }
+
+impl From<ModuleId> for ModuleInfo {
```
hyf0: I'll suggest using `ModuleInfo::new`.

hyf0 (approved):

crates/rolldown_common/src/types/module_info.rs
underfin: Here `ModuleInfo::new` needs to care about more fields, the external module only has `id`.

hyf0: It's for clear reading

```diff
- self.shared_context.plugin_driver.set_module_info(&id.clone(), Arc::new(id.into()));
+ self.shared_context.plugin_driver.set_module_info(&id.clone(), Arc::new(moduleInfo::new(id)));
```

underfin: > Here `ModuleInfo::new` needs to care about more fields, the external module only has `id`.

The `MouleInfo::new` does not follow the rust struct style, let using struct litreal at here.

PR #2924, opened 2024-11-26
fix: ensure support nest and async plugins

underfin:

### DIFF
--- a/packages/rolldown/src/plugin/index.ts
+++ b/packages/rolldown/src/plugin/index.ts
@@ -300,13 +300,10 @@ export interface Plugin<A = any> extends OutputPlugin, Partial<PluginHooks> {
 }
 
 export type RolldownPlugin<A = any> = Plugin<A> | BuiltinPlugin | ParallelPlugin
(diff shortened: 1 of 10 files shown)


packages/rolldown/src/plugin/index.ts
```diff
@@ -300,13 +300,10 @@ export interface Plugin<A = any> extends OutputPlugin, Partial<PluginHooks> {
 }
 
 export type RolldownPlugin<A = any> = Plugin<A> | BuiltinPlugin | ParallelPlugin
-export type RolldownPluginOption<A = any> = NullValue<RolldownPlugin<A>> | false
+export type RolldownPluginOption<A = any> = MaybePromise<
+  NullValue<RolldownPlugin<A>> | false | RolldownPluginOption[]
+>
```
underfin, line 305: The typing migrate from rollup https://github.com/rollup/rollup/blob/master/src/rollup/types.d.ts#L619

hyf0 (approved):

PR #2925, opened 2024-11-26
chore: oxc v0.38.0

Boshen:


Boshen:
> Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@babel/runtime' imported from /Users/boshen/github/rolldown/crates/rolldown/tests/esbuild/lower/lower_nullish_coalescing_assignment_issue1493/dist/entry.js

How and where should we bundle this helper? We use babel runtime for the time being but will eventually switch to our own helper https://github.com/oxc-project/oxc/issues/7422

hyf0:
@Boshen It's tricky here. When you use bable in webpack/rollup, it's your job to install `@babel/runtime`. But oxc in rolldown is a bit like a builtin thing, probably should add it in the `deps` I guess.

packages/rolldown/src/binding.d.ts
```diff
@@ -794,6 +794,21 @@ export interface TransformOptions {
...
+   * Example:
+   *
+   * * 'es2015'
+   * * ['es2020', 'chrome58', 'edge16', 'firefox57', 'node12', 'safari11']
+   *
+   * @default `esnext` (No transformation)
+   *
+   * @see [esbuild#target](https://esbuild.github.io/api/#target)
+   */
+  target?: string | Array<string>
```
Boshen, line 811: Note: `target` is exposed on the napi side as well.

package.json
```diff
@@ -42,7 +42,8 @@
     "oxlint": "0.12.0",
     "prettier": "^3.3.1",
     "rolldown": "workspace:*",
-    "typescript": "^5.6.3"
+    "typescript": "^5.6.3",
+    "@babel/runtime": "7.26.0"
```
Boshen, line 46: Note: This is required for tests to run for the entire monorepo.

crates/rolldown/src/ast_scanner/mod.rs
```diff
@@ -7,7 +7,7 @@ pub mod side_effect_detector;
 use arcstr::ArcStr;
 use oxc::ast::ast::MemberExpression;
 use oxc::ast::{ast, AstKind};
-use oxc::index::IndexVec;
+use oxc_index::IndexVec;
```
Boshen: Note: for all these oxc index changes, I removed `oxc_index` from `oxc::index`.

hyf0:
Need `just lint-rust` to catch lint errors.

crates/rolldown/tests/esbuild/default/define_optional_chain/artifacts.snap
hyf0, line 1: Why this is deleted? If the output is considered as expected compared to the esbuild's output. You could rename it to `bypass.md` and run `just ued` to update the esbuild's test coverage status.

underfin (approved):

crates/rolldown/tests/esbuild/default/define_optional_chain/artifacts.snap
underfin: Look like the `bypass.md` is not unnecessary, it already passed at https://github.com/rolldown/rolldown/pull/2925/files#diff-c141b120dd177e198e208800d017c4a8c75c700b7d36925bbbc3eb4d2c26fcbfR189

Boshen: I don't know, it removed itself after I ran `just test-update` + `just ued`

IWANABETHATGUY: the diff.md will be auto deleted if our output is same as esbuild

PR #2928, opened 2024-11-26
fix: sort NODEJS_PREFIXED_BUILTINS

sapphi-red:
### Description

This should be sorted because we are calling `binary_search` here.
https://github.com/rolldown/rolldown/blob/78749354b8bd65833a2313402f65dc06638d37af/crates/rolldown_common/src/ecmascript/node_builtin_modules.rs#L17

### DIFF
--- a/crates/rolldown_common/src/ecmascript/node_builtin_modules.rs
+++ b/crates/rolldown_common/src/ecmascript/node_builtin_modules.rs
@@ -4,10 +4,9 @@ use oxc_resolver::NODEJS_BUILTINS;
 const NODEJS_PREFIXED_BUILTINS: &[&str] = &[
   // https://nodejs.org/api/modules.html#built-in-modules-with-mandatory-node-prefix
   "node:sea",
+  "node:sqlite",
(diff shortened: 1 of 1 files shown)


crates/rolldown_common/src/ecmascript/node_builtin_modules.rs
```diff
@@ -4,10 +4,9 @@ use oxc_resolver::NODEJS_BUILTINS;
 const NODEJS_PREFIXED_BUILTINS: &[&str] = &[
   // https://nodejs.org/api/modules.html#built-in-modules-with-mandatory-node-prefix
   "node:sea",
+  "node:sqlite",
   "node:test",
   "node:test/reporters",
-  // https://nodejs.org/api/sqlite.html
```
sapphi-red, line 9: removed this comment as `node:sqlite` is now included in the list of https://nodejs.org/api/modules.html#built-in-modules-with-mandatory-node-prefix

hyf0 (approved):

PR #2929, opened 2024-11-26
feat: give specific default plugin name

underfin:
### Description

Normalize plugin name to instead of `unknown`, follow rollup https://github.com/rollup/rollup/blob/master/src/rollup/rollup.ts#L167.

### DIFF
--- a/packages/rolldown/src/plugin/bindingify-plugin.ts
+++ b/packages/rolldown/src/plugin/bindingify-plugin.ts
@@ -126,6 +126,7 @@ export function bindingifyPlugin(
     bindingifyCloseWatcher(args)
 
   const result: BindingPluginOptions = {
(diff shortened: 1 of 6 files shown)


packages/rolldown/src/plugin/bindingify-plugin.ts
```diff
@@ -126,7 +126,7 @@ export function bindingifyPlugin(
     bindingifyCloseWatcher(args)
 
   const result: BindingPluginOptions = {
-    name: plugin.name ?? 'unknown',
+    name: plugin.name!,
```
hyf0, line 130: Could you add a comment saying that nameless plugin would be surely assign a name by rolldown?

hyf0 (approved):

packages/rolldown/src/plugin/bindingify-plugin.ts
underfin: Done

PR #2944, opened 2024-11-26
chore: remove unnecessary `@ts-ignore`

shulaoda:
### Description

I will work on resolving the ts type issues related to `filter` next PR.

https://github.com/rolldown/rolldown/blob/0535a45140334005d359aa8f11bf9c1acd7c3e52/packages/rolldown/src/plugin/bindingify-build-hooks.ts#L163-L165


hyf0 (approved):

hyf0:
Thanks

Issue #1378, opened 2024-06-10
Clarify node versions that rolldown supports and ensure correctness about it


hyf0:
Two concepts:
- node versions that rolldown supports
- local `.node-version` that we used to development rolldown. The `.node-version` must be the supported version of rolldown and also enable to be used to running the development flow(some old version that rolldown supports might not satisfy this requirement.

For CI,
- In general purpose and rust tests actions(lint-repo, lint-node, lint-rust, rust-test), we need to use `.node-version` due to it ensures the node-version is enable to run the development flow.
- In the node-test and wasm-test actions, currently we don't have the development workflow that require `.node-version` to run(this might happens in the future, but let's ignore it so far). The node-test action need to tests all supported node versions, so node-version would an input showed in matrix field. It shouldn't use `.node-version`, though the `.node-version` is a version that rolldown supported.

---

I do agree with you about `package.json#engine` should not be updated by renovate. `.node-version` is ok due to also that newest node's version is the version that rolldown needs to support.

---

This PR does solve problems, so let's merge it first. I will create an issue to clarify the above thing.

_Originally posted by @hyf0 in https://github.com/rolldown/rolldown/issues/1377#issuecomment-2158880347_

hyf0:
I propose that supportings all node newest lts versions and newest node version(odd-numbered releases won't become LTS version), which are

- 22
- 20
- 18

https://endoflife.date/nodejs

hyf0:
> I propose that supportings all node newest lts versions and newest node version(odd-numbered releases won't become LTS version), which are
> 
> * 22
> * 20
> * 18
> 
> https://endoflife.date/nodejs

We will hit problems while using 18 and 20. We are using features like `import.meta.dirname` which are not supported in old node versions. Since we are using the same node version to run test infra code and compiled code. Dealing with `import.meta.dirname` is complex.

hyf0:
Related problems:
- https://github.com/rolldown/rolldown/issues/1376

hyf0:
Meeting conclusion:

Rolldown will make sure all tests of rolldown are passed in active LTS versions of node. Odd-numbered releases of node are not included in this promise.

We will use latest node's version to develop rolldown, which ensures rolldown is usable even in odd-numbered releases in a way. However, this is not a official guarantee.

hyf0:
Ensure this means more CI time, not sure if we want to handle it so far.

PR #2945, opened 2024-11-26
fix: update package.json engines node minimal version

joinmouse:
### Description

scripts/misc/setup-prerequisites/node.js check node minimal version is read .node-version(20.14.0)

so package.json engines node should 20.14.0


hyf0:
See https://github.com/rolldown/rolldown/issues/1378#issue-2344422031.

There are different things.

PR #2970, opened 2024-11-28
refactor(plugin/vite-resolve): use oxc-resolver to resolve optional peer dep value from package.json

sapphi-red:
### Description

Removed the custom package.json resolving logic and use oxc-resolver for that (adds a cache too).


crates/rolldown_plugin_vite_resolve/src/package_json_cache.rs
```diff
@@ -1,24 +1,99 @@
...
+};
 
 use dashmap::DashMap;
 use rolldown_common::PackageJson;
+use serde::{de::IgnoredAny, Deserialize};
 
 #[derive(Debug, Default)]
 pub struct PackageJsonCache {
-  package_json_cache: DashMap<PathBuf, Arc<PackageJson>>,
+  side_effects_cache: DashMap<PathBuf, Arc<PackageJson>>,
```
Boshen: Can you use fx hash for all these hashmaps? e.g. https://github.com/oxc-project/oxc-resolver/blob/fe70f41f0dd1a7d0cfb854839c0b1110cbad7a35/src/cache.rs#L30

crates/rolldown_plugin_vite_resolve/src/package_json_cache.rs
```diff
@@ -1,24 +1,101 @@
...
+      self.optional_peer_dep_cache.insert(oxc_pkg_json.realpath.clone(), Arc::clone(&pkg_json));
+      pkg_json
+    }
+  }
+}
+
+#[derive(Debug, Default)]
+pub struct PackageJsonWithOptionalPeerDependencies {
+  pub name: String,
+  pub optional_peer_dependencies: BTreeSet<String>,
```
Boshen: One more here.

crates/rolldown_plugin_vite_resolve/src/package_json_cache.rs
```diff
@@ -1,24 +1,101 @@
...
+        .collect(),
+    })
+  }
+}
+
+#[derive(Deserialize)]
+struct PackageJsonWithPeerDependenciesRaw {
+  pub name: String,
+  #[serde(rename = "peerDependencies")]
+  pub peer_dependencies: Option<BTreeMap<String, IgnoredAny>>,
```
Boshen: Another one.

crates/rolldown_plugin_vite_resolve/src/package_json_cache.rs
```diff
@@ -1,24 +1,101 @@
...
+  #[serde(rename = "peerDependenciesMeta")]
+  pub peer_dependencies_meta: Option<BTreeMap<String, PackageJsonPeerDependenciesMetaRaw>>,
```
Boshen: Another one.

Boshen (commented):
Does this plugin have access to an `OxcResolver` instance?

crates/rolldown_plugin_vite_resolve/src/package_json_cache.rs
sapphi-red: Made #2971 for other places that wasn't using fx hash.

crates/rolldown_plugin_vite_resolve/src/package_json_cache.rs
sapphi-red: Ideally, I wanted these two to be a Map backed by a Vec as I don't need the key access. I don't know if `HashMap` has smaller cost than `BTreeMap` or not. If that's the case, I'll replace these with a HashMap.

crates/rolldown_plugin_vite_resolve/src/package_json_cache.rs
sapphi-red: Do you think it's better to use a HashSet here? The size of this set should be small and will only looked up in minor cases.
I'm not sure whether which of `BTreeSet` or `HashSet` or a sorted `Vec`+`binary_search` is better.

sapphi-red:
> Does this plugin have access to an `OxcResolver` instance?

I think no, if you mean the `OxcResolver` instance in rolldown's internal resolver.

crates/rolldown_plugin_vite_resolve/src/package_json_cache.rs
sapphi-red: Ah, actually I need the key access for `peer_dependencies_meta`.

Boshen: `HashMap` should perf better in this case.

crates/rolldown_plugin_vite_resolve/src/package_json_cache.rs
Boshen: `HashMap` is always cheaper.

crates/rolldown_plugin_vite_resolve/src/package_json_cache.rs
IWANABETHATGUY: Unless you needs visit the key in order, hashmap should outperform in most scenarios.

hyf0 (approved):
No blocking issues. Let's merge it first.

PR #2972, opened 2024-11-28
refactor: using reference at generate diagnostic

underfin:
### Description

Refactor it because the `BuildEnd#error` need to use it, prepare for https://github.com/rolldown/rolldown/pull/2973.


hyf0 (approved):
LGTM.

NIT: Is it hard to make to support `clone` and work with `into_xxx`?

underfin:
> LGTM.
> 
> NIT: Is it hard to make to support `clone` and work with `into_xxx`?

I'm not try to do it, because it has more overhead. So i using reference to avoid clone for all errors.

PR #2973, opened 2024-11-28
fix: the buildEnd error should handle js errors

underfin:


crates/rolldown/src/bundler.rs
```diff
@@ -110,21 +109,17 @@ impl Bundler {
     {
       Ok(v) => v,
       Err(errs) => {
-        // TODO: So far we even call build end hooks on unhandleable errors . But should we call build end hook even for unhandleable errors?
-        error_for_build_end_hook = Some(errs.first().unpack_ref().to_string());
+        let errors = UniqueArc::new(errs.into_vec());
```
hyf0: What's the issue of using `Arc` directly.

underfin: Using `Arc` and  then `into_inner`, i have a thinking Inertia to using `UniqueArc`. I updated to `Arc`.

hyf0 (approved):

PR #2968, opened 2024-11-27
feat(bundle-size): generate shorthand prop if possible for namespace decl

hyfdev:

PR #2967, opened 2024-11-27
refactor(runtime): extend `__export` to allow value assign

hyfdev:
### Description

This PR is a pre-step for https://github.com/rolldown/rolldown/pull/2968.

PR #2974, opened 2024-11-28
refactor: move namespace decl to bottom if possible

hyfdev:
### Description

Pre-step for https://github.com/rolldown/rolldown/pull/2968.


hyf0:
* **#2974** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2974?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/2974?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

crates/rolldown/src/module_finalizers/scope_hoisting/mod.rs
```diff
@@ -27,6 +27,7 @@ pub struct ScopeHoistingFinalizer<'me, 'ast> {
   pub alloc: &'ast Allocator,
   pub snippet: AstSnippet<'ast>,
   pub comments: oxc::allocator::Vec<'ast, Comment>,
+  pub contains_reference_to_namespace_ref_of_this_module: bool,
```
hyf0, line 30: The biggest complexity comes from how we make `contains_reference_to_namespace_ref_of_this_module` right.

crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
```diff
@@ -172,6 +165,10 @@ impl<'me, 'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'me, 'ast> {
         }
         WrapKind::None => {}
       }
+    }
+
+    if self.contains_reference_to_namespace_ref_of_this_module || needs_wrapper {
```
hyf0, line 172: This PR actually works for `needs_wrapper: true` too, but it require more dedicated code, I remove them for simplicity.

crates/rolldown/src/module_finalizers/scope_hoisting/mod.rs
hyf0: Such as we need to consider if rewrited `member_expr` point to the `namespace_ref_of_this_module`.

crates/rolldown/src/module_finalizers/scope_hoisting/rename.rs
```diff
@@ -84,6 +84,11 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
     is_callee: bool,
   ) -> Option<Expression<'ast>> {
     if let Some(new_expr) = self.generate_finalized_expr_for_reference(ident_ref, is_callee) {
```
IWANABETHATGUY: symbol_ref could be reused from https://github.com/rolldown/rolldown/blob/8ee2182a117626d7d0f122b507fe1119b01c8a19/crates/rolldown/src/module_finalizers/scope_hoisting/rename.rs?plain=1#L16-L43, since the return value is already `some`

IWANABETHATGUY:
Please bypass all regressed test, because the only different is codegen order.

crates/rolldown/src/module_finalizers/scope_hoisting/rename.rs
hyf0: Yeah. This is more like POC, I'll optimize these code in this PR after solving the major problem.

underfin:
> Do you think we could move namespace decl down to then bottom so this won't get the error of reference value before declaring

I think the pr intend to resolve it? Look like it's unnecessary, because the unsafe case already fallback to `getter`, eg `export const value = 1`.

IWANABETHATGUY (approved):

hyf0:
I'm not gonna merge this PR standalone. This PR would be merged after solving https://github.com/rolldown/rolldown/pull/2967.

hi-ogawa:
It looks like webpack has exports definition at the top for cyclic import support. Here is an example I was investigating related Vite SSR issue https://github.com/hi-ogawa/reproductions/tree/main/cyclic-import

I'd assume this PR is an optimization only for scope hoisting bundle, so this doesn't look relevant, but I just wanted to mention we cannot probably do the same for app format.

hyf0:
> It looks like webpack has exports definition at the top for cyclic import support. Here is an example I was investigating related Vite SSR issue https://github.com/hi-ogawa/reproductions/tree/main/cyclic-import
> 
> I'd assume this PR is an optimization only for scope hoisting bundle, so this doesn't look relevant, but I just wanted to mention we cannot probably do the same for app format.

Yeah. It's a situation that needs to fallback to generating top.

PR #2975, opened 2024-11-28
refactor(rolldown/utils): optimise bindingifyInputOptions

JayFate:
### Description

optimise bindingifyInputOptions.


underfin:
Thank you. But the original code is ok to usage, here close it to avoid a similar code style pr.

PR #2982, opened 2024-11-29
fix: nested member expression mutation from imports

underfin:


crates/rolldown/tests/rollup/mutations-in-imports/artifacts.snap
```diff
@@ -0,0 +1,34 @@
...
+```js
+
+//#region foo.js
+const foo = { a: 0 };
+
+//#endregion
+//#region main.js
+foo.a = 1;
+(void 0).a = 1;
+
```
underfin, line 32: The esbuild repl result https://hyrious.me/esbuild-repl/?version=0.24.0&b=e%00entry.js%00import+*+as+file+from+%27.%2Ffile%27%0A%0Afile.y.a+%3D1%0Afile.b.a+%3D1&b=%00file.js%00export+const+b+%3D+%7B+a%3A+2+%7D&o=--bundle+--outdir%3Ddist+--external%3Aexternal+--format%3Desm

crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
```diff
@@ -263,6 +264,24 @@ impl<'me, 'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'me, 'ast> {
     walk_mut::walk_expression(self, expr);
   }
 
+  // foo.js `export const bar = { a: 0 }`
+  // main.js `import * as foo_exports from './foo.js';\n foo_exports.bar.a = 1;`
+  // The `foo_exports.bar.a` ast is `StaticMemberExpression(StaticMemberExpression)`, The outer StaticMemberExpression span is `foo_exports.bar.a`, the `visit_expression(Exprssion::MemberExpression)` is called with `foo_exports.bar`, the span is inner StaticMemberExpression.
```
underfin, line 282: It is reason why the `visit_expression(Exprssion::MemberExpression)` not handle the case.

underfin: If we migrated to  oxc `Traverse` API, the issue should be gone.

IWANABETHATGUY (approved):

PR #2988, opened 2024-11-29
perf: use fxhash for `HashMap`s

sapphi-red:
### Description

This PR needs https://github.com/napi-rs/napi-rs/pull/2374.


hyf0 (approved):
To have a correct semantic, these all should be `FxIndexMap` I guess.

However, let's worried about it in another PR

crates/rolldown_binding/src/utils/normalize_binding_options.rs
```diff
@@ -123,7 +124,7 @@ pub fn normalize_binding_options(
 
   let mut module_types = None;
   if let Some(raw) = input_options.module_types {
-    let mut tmp = HashMap::with_capacity(raw.len());
+    let mut tmp: FxHashMap<_, _> = FxHashMapExt::with_capacity(raw.len());
```
sapphi-red, line 127: Is it possible to make this work without the explicit type (`: FxHashMap<_, _>`) by changing the types of `FxHashMapExt::with_capacity`?

hyf0: I guess so. Instead of returning `Self` in `with_capacity`, we could return `FxHashMap` directly since the trait is only gonna be used on `FxHashMap`.

```rust
pub trait FxHashMapExt<K, V> {
  fn with_capacity(capacity: usize) -> FxHashMap<K, V>;
}
```

Should works

sapphi-red: I now tried that but it seems it doesn't work.
The error was

> cannot call associated function on trait without specifying the corresponding `impl` type cannot call associated function of trait

hyf0: `FxhashMap::with_capacity` just works.  No need to use `FxHashMapExt:: with_capacity`.

No shadowing problem, because `FxhashMap` doesn't have `with_capacity` method in the first place. That's why I add `FxHashMapExt`.

PR #2989, opened 2024-11-30
chore: remove unnecessary forced `inputOptions#onLog`

shulaoda:
### Description

Related to #2870

Since we have already implemented `NormalizedInputOptionsImpl`, the following hack behavior is unnecessary.

https://github.com/rolldown/rolldown/blob/59f76dc54c9d39e3ad7099c03ffdaa477f4e116e/packages/rolldown/src/options/bindingify-input-options.ts#L30-L35


hyf0 (approved):
Thanks!

PR #2990, opened 2024-11-30
feat: support `preserve` for `outputOptions#comments`

shulaoda:
### Description

This PR was initially created to address the inconsistency between `preserve-legal` and `Comments::Preserve`.

https://github.com/rolldown/rolldown/blob/f1c8656cd231ddff20ded658665a987a78e41fd6/crates/rolldown_binding/src/utils/normalize_binding_options.rs#L230-L234

Later, I realized that the `preserve` option would be exposed, so I added it to the default options type for `comments`.

https://github.com/rolldown/rolldown/blob/9cd90be46446794693d3f55d6b851bd01bed2847/crates/rolldown_binding/src/types/binding_normalized_options.rs#L234-L237

### DIFF
--- a/packages/rolldown/src/options/output-options.ts
+++ b/packages/rolldown/src/options/output-options.ts
@@ -191,9 +191,10 @@ export interface OutputOptions {
    * Control comments in the output.
    *
    * - `none`: no comments
+   * - `preserve`: preserve all comments except legal comments
    * - `preserve-legal`: preserve comments that contain `@license`, `@preserve` or starts with `//!` `/*!`
    */
-  comments?: 'none' | 'preserve-legal';
+  comments?: 'none' | 'preserve' | 'preserve-legal';
   plugins?: RolldownOutputPluginOption;
   polyfillRequire?: boolean;
   target?: ESTarget;
--- a/packages/rolldown/tests/fixtures/function/comments/preserve/output.snap
+++ b/packages/rolldown/tests/fixtures/function/comments/preserve/output.snap
@@ -1,0 +1,18 @@
+
+//#region ../none/main.js
+let x;
+console.log("in a");
+console.log("in b");
+//! Legal comments1
+/*! Legal comments2 */
+foo;
+bar;
+function test() {}
+var main_default = () => {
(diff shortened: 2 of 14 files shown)


packages/rolldown/src/options/output-options.ts
```diff
@@ -70,9 +70,10 @@ export interface OutputOptions {
    * Control comments in the output.
    *
    * - `none`: no comments
+   * - `preserve`: preserve all comments except legal comments
```
shulaoda, line 73: a

shulaoda, line 194: I'm not sure if this description is accurate, as I wrote it based on the execution logic.

https://github.com/rolldown/rolldown/blob/f1c8656cd231ddff20ded658665a987a78e41fd6/crates/rolldown_ecmascript/src/ecma_compiler.rs#L57-L60

packages/rolldown/tests/fixtures/function/comments/preserve/output.snap
```diff
@@ -0,0 +1,17 @@
+
```
shulaoda, line 1: Finally, the shotsnap for these three test cases seems unexpected. There may be an issue with `oxc_codegen` regarding comments, and I'll open an issue to track it later.

hyf0:
`preserve` doesn't get well supported by oxc so far, so we didn't expose it. I'm gonna make this PR draft until oxc finish this feature. It probably takes a while, due to oxc team has other high priority things to do.

packages/rolldown/tests/fixtures/function/comments/preserve/output.snap
shulaoda: Related to #4006

shulaoda:
There are some bug issues present: #4006, #4118

IWANABETHATGUY:
> There are some bug issues present: #4006, #4118

#4006 is fixed already

Boshen:
We can still merge this, but change the documentation to:

`preserve`: all statement comments and annotation comments.

PR #3009, opened 2024-12-02
fix: avoid buildEnd panic

underfin:
### Description
Fix the panic of `Arc into_inner should success after call buildEnd hook`. The previous test is not assert test result, so make the test passed. ref https://github.com/rolldown/rolldown/actions/runs/12115381503/job/33774198333


crates/rolldown/src/bundler.rs
```diff
@@ -108,20 +109,13 @@ impl Bundler {
...
       Err(errs) => {
-        let errors = Arc::new(errs.into_vec());
+        let errors = UniqueArc::new(errs.into_vec());
         self
           .plugin_driver
-          .build_end(Some(&HookBuildEndArgs {
-            errors: Arc::clone(&errors),
-            cwd: &self.options.cwd,
-          }))
+          .build_end(Some(&HookBuildEndArgs { errors: errors.weak_ref(), cwd: &self.options.cwd }))
```
underfin, line 115: Other thing is the rollup only have one error because the js `Error` mechanism，the rolldown `error` is mutilply, we should merge it to one error.

hyf0 (approved):

PR #3016, opened 2024-12-02
ci: enable `Node Test` on node 18, 20 and 22

### LINKED ISSUE #1378: Clarify node versions that rolldown supports and ensure correctness about it

hyf0:
### Description

Closes https://github.com/rolldown/rolldown/issues/1378.

### DIFF
--- a/.github/workflows/reusable-node-test.yml
+++ b/.github/workflows/reusable-node-test.yml
@@ -65,3 +65,17 @@ jobs:
 
       - name: Build Examples For Node20
         run: pnpm --filter '@example/**' build
+
+      - name: Setup Node22 For Testing
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - name: Node Test For Node22
+        run: pnpm run --recursive --parallel --filter=!rollup-tests test
+
+      - name: Rollup Test Node22
+        run: pnpm run --filter rollup-tests test
+
+      - name: Build Examples For Node22
+        run: pnpm --filter '@example/**' build
--- a/.github/workflows/reusable-node-test.yml
+++ b/.github/workflows/reusable-node-test.yml
@@ -23,7 +23,7 @@ jobs:
         with:
           submodules: true # Pull submodules for additional files
 
-      - name: Setup Node
+      - name: Setup Node For Development
         uses: ./.github/actions/setup-node
 
       - name: Setup Rust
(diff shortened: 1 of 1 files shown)


hyf0:
* **#3016** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3016?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3016?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

.github/workflows/reusable-node-test.yml
```diff
@@ -38,11 +38,30 @@ jobs:
       - name: Type Check
         run: pnpm type-check
 
-      - name: Node Test
+      - name: Setup Node18 For Testing
```
IWANABETHATGUY, line 41: Maybe we could use matrix?

hyf0: Using matrix would case building binding multiple times, which is the slowest step in the job. Though there are running in parallel, but it would cost more machine to run CI in the same time.  Seems a waste and would delay running action of other PR.

hyf0:
Running tests only takes 10~ seconds in node18, so I enable node22 too.

IWANABETHATGUY (approved):

PR #3033, opened 2024-12-04
chore: improve cli bundle command

shulaoda:
### Description

We now support `output.file`, so I have made improvements to it and aligned it with `rollup`.

https://github.com/rolldown/rolldown/blob/176666fa328cf77657addaa04787061e9ac836b1/packages/rolldown/src/cli/commands/bundle.ts#L33


hyf0 (approved):
Good job!

PR #3036, opened 2024-12-04
test: panics when error happens

sapphi-red:
### Description

https://github.com/rolldown/rolldown/commit/2ff61462c1bbdc00955d99e81df2ced8976db582 made the build to panic when an error happens.
It panics because of this line.
https://github.com/rolldown/rolldown/commit/2ff61462c1bbdc00955d99e81df2ced8976db582#diff-6d7d83333f1eb087a0cf62417600d1dee89f55d50777f83108c9ec41d9f84e7cR122
The reason why the strong count is not 0 is because `BindingHookError` which is passed to the JS side holds the reference.
https://github.com/rolldown/rolldown/blob/dbbc45a5126c92fd7239617733f539d2ae631f9d/crates/rolldown_binding/src/options/plugin/types/binding_hook_error.rs#L12
This is to make things like below work.
```js
let e
const plugin = {
  buildEnd(err) {e = err}
}
// run build
console.log(e)
```

I'm not sure how this should be fixed. The ways I can think of is:

1. clone the value here (https://github.com/rolldown/rolldown/blob/dbbc45a5126c92fd7239617733f539d2ae631f9d/crates/rolldown/src/bundler.rs#L115)
2. make `scan` to return `Arc<BatchedBuildDiagnostic>` https://github.com/rolldown/rolldown/blob/dbbc45a5126c92fd7239617733f539d2ae631f9d/crates/rolldown/src/bundler.rs#L97
3. change `BindingHookError::errors` to be a `sync::Weak` so that the reference is dropped after `buildError` call https://github.com/rolldown/rolldown/blob/dbbc45a5126c92fd7239617733f539d2ae631f9d/crates/rolldown_binding/src/options/plugin/types/binding_hook_error.rs#L12

### DIFF
--- a/packages/rolldown/tests/plugin/plugin.test.ts
+++ b/packages/rolldown/tests/plugin/plugin.test.ts
@@ -2,48 +2,49 @@ import { expect, test, vi, describe } from 'vitest'
 import { rolldown, Plugin } from 'rolldown'
 
 async function buildWithPlugin(plugin: Plugin) {
-  try {
-    const build = await rolldown({
-      input: './main.js',
-      cwd: import.meta.dirname,
-      plugins: [plugin],
-    })
-    await build.write({})
-  } catch {
-    // Here `renderError` test will crash it, here avoid bubble it.
-    // console.log(error)
-  }
+  const build = await rolldown({
+    input: './main.js',
+    cwd: import.meta.dirname,
+    plugins: [plugin],
+  })
+  await build.write({})
 }
 
 test('Plugin renderError hook', async () => {
   const renderErrorFn = vi.fn()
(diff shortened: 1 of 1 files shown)


sapphi-red:
@underfin May I hand over this one to you? I think this requires more codebase knowledge than I have now.

underfin:
Yeah. I'm fix it at https://github.com/rolldown/rolldown/pull/3009. It is caused by `JsError` can't `ToNapiValue::into_value` twcie, I fix it at https://github.com/napi-rs/napi-rs/pull/2388.

sapphi-red:
Ah, I see 👍

PR #3032, opened 2024-12-04
refactor(watcher): add WatcherTask

underfin:
### Description

related to https://github.com/rolldown/rolldown/issues/2842. Here add `WatcherTask` to make `Watcher` support multiply `Bundler`.

PR #3037, opened 2024-12-04
feat(watcher): support multiply watch options

underfin:
### Description

The `Bundler::watch` api is not suitable for watcher if it is multiply watch options, so here add a new `BindingWatcher`. The binding the multiply `Bundler` to create `BindingWatcher` is impossible, so here pass binding multiply options to create it.

The rollup is create `chokidar` for each `Bundler`, it has more overhead at rolldown, because we need to create thread to listen event. So the `notify` option is used for all `Bundler`, i will check it at next.


underfin:
* **#3037** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3037?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3037?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3032** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3032?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

packages/rolldown/src/binding.d.ts
```diff
@@ -167,18 +167,6 @@ export interface ArrowFunctionsOptions {
   spec?: boolean
 }
 
-export interface ArrowFunctionsOptions {
```
underfin, line 170: It's is duplicated.

crates/rolldown_binding/src/bundler.rs
```diff
@@ -19,6 +18,13 @@ use napi_derive::napi;
 use rolldown::Bundler as NativeBundler;
 use rolldown_error::{BuildDiagnostic, BuildResult, DiagnosticOptions};
 
+#[napi(object, object_to_js = false)]
+pub struct BindingBundlerOption {
```
hyf0:
```suggestion
pub struct BindingBundlerOptions {
```

Plural would be better.

packages/rolldown/src/utils/create-bundler-option.ts
```diff
@@ -0,0 +1,95 @@
...
+      outputOptions,
+      onLog,
+      logLevel,
+    )
+
+    // Convert `OutputOptions` to `BindingInputOptions`
+    const bindingOutputOptions = bindingifyOutputOptions(outputOptions)
+
+    return {
+      bundlerOption: {
```
hyf0:
```suggestion
      bundlerOptions: {
```

hyf0 (approved):

PR #3042, opened 2024-12-04
feat(rolldown_plugin_transform): support both target & browserslist

sxzz:
### Description

Support both `target` & `browserslist`, and it's required by downstream https://github.com/sxzz/tsdown

### DIFF
--- a/crates/rolldown_plugin_transform/src/lib.rs
+++ b/crates/rolldown_plugin_transform/src/lib.rs
@@ -21,6 +21,7 @@ pub struct TransformPlugin {
   pub jsx_inject: Option<String>,
   pub react_refresh: bool,
 
+  // TODO: support specific transform options. Firstly we can use `target` & `browserslist` but we'd better allowing user to pass more options.
   pub target: Option<String>,
   pub browserslist: Option<String>,
 }
--- a/crates/rolldown_plugin_transform/src/lib.rs
+++ b/crates/rolldown_plugin_transform/src/lib.rs
@@ -61,16 +61,15 @@ impl Plugin for TransformPlugin {
     };
 
     let env = if self.target.is_some() && self.browserslist.is_some() {
-      return Err(anyhow::anyhow!(
-        "Cannot specify both `target` and `browserslist` at the same time"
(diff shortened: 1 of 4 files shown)


crates/rolldown_plugin_transform/src/lib.rs
```diff
@@ -58,16 +59,21 @@ impl Plugin for TransformPlugin {
         return Err(anyhow::format_err!("Error occurred when parsing {}\n: {:?}", args.id, errs));
       }
     };
+
+    let env = if self.target.is_some() && self.browserslist.is_some() {
+      return Err(anyhow::anyhow!(
+        "Cannot specify both `target` and `browserslist` at the same time"
+      ));
+    } else if let Some(target) = &self.target {
+      EnvOptions::from_target(target).expect("Failed to create transform options")
```
Boshen: These should return the error messages from these functions, instead of panicking with `expect`.

crates/rolldown_plugin_transform/src/lib.rs
```diff
@@ -58,16 +59,20 @@ impl Plugin for TransformPlugin {
         return Err(anyhow::format_err!("Error occurred when parsing {}\n: {:?}", args.id, errs));
       }
     };
+
+    let env = if self.target.is_some() && self.browserslist.is_some() {
+      Err("Cannot specify both `target` and `browserslist` at the same time".to_string())
```
hyf0, line 64: Is this required by oxc-transfomer? For better UX, I guess we could support accepting  both `target` and `browserslist`. @Boshen cc

hyf0 (approved):
Let me do a merge first. But I think the API could be improved. `targets` and `broweserlist` should not be conflict-able.

crates/rolldown_plugin_transform/src/lib.rs
sxzz: At this point, I'm guessing the Oxc API isn't ready yet.

hyf0:
CI failed: you need run `just build` to update `binding.d.ts`.

PR #3052, opened 2024-12-05
feat: expose the option `checks` to `inputOptions`

shulaoda:
### Description

https://github.com/rolldown/rolldown/blob/225f6a41679fff165eff297ba21578a72b4d88eb/packages/rolldown/tests/fixtures/onwarn/_config.ts#L8


packages/rolldown/src/options/input-options.ts
```diff
@@ -137,6 +141,7 @@ export interface InputOptions {
   watch?: WatchOptions | false
   dropLabels?: string[]
   keepNames?: boolean
+  checks?: ChecksOptions
```
hyf0, line 148: Should also add it  in schema.

packages/rolldown/src/options/input-options.ts
```diff
@@ -56,6 +56,10 @@ export interface WatchOptions {
   chokidar?: any
 }
 
+export interface ChecksOptions {
+  circularDependency?: boolean
```
hyf0, line 64: Add jsdoc

```suggestion
  Wether to emit warnings when detecting circular dependencies.
  @default false
  circularDependency?: boolean
```

packages/rolldown/src/options/input-options.ts
shulaoda: Related to #2193 
It seems to have already been updated.

packages/rolldown/src/binding.d.ts
```diff
@@ -606,6 +628,25 @@ export interface IsolatedDeclarationsOptions {
   sourcemap?: boolean
 }
 
+export interface IsolatedDeclarationsOptions {
```
shulaoda, line 631: Related to #3058 

I tried what you did on windows, but it didn't work. @hi-ogawa cc

hyf0 (approved):
Good job.

PR #3064, opened 2024-12-06
feat(node): support `SourceMap::toUrl`/`SourceMap::toString`

sapphi-red:

### DIFF
--- a/packages/rolldown/src/utils/transform-to-rollup-output.ts
+++ b/packages/rolldown/src/utils/transform-to-rollup-output.ts
@@ -1,4 +1,5 @@
 import type {
+  SourceMap,
   RolldownOutput,
   RolldownOutputAsset,
   RolldownOutputChunk,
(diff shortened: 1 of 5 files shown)


packages/rolldown/src/utils/transform-to-rollup-output.ts
```diff
@@ -20,6 +21,21 @@ import {
...
 import { normalizeErrors } from './error'
 import { transformChunkModules } from './transform-rendered-chunk'
+import { Buffer } from 'node:buffer'
+
+function transformToRollupSourceMap(map: string): SourceMap {
+  const parsed: Omit<SourceMap, 'toString' | 'toUrl'> = JSON.parse(map)
+  const obj: SourceMap = {
+    ...parsed,
+    toString() {
+      return JSON.stringify(obj)
```
underfin, line 31: Here could be directly using `map`

packages/rolldown/src/utils/transform-to-rollup-output.ts
```diff
@@ -20,6 +21,21 @@ import {
...
+    },
+    toUrl() {
+      return `data:application/json;charset=utf-8;base64,${Buffer.from(obj.toString(), 'utf-8').toString('base64')}`
```
underfin, line 34: Maybe  here also could be using `map` to instead of `obj.toString()`.

packages/rolldown/src/utils/transform-to-rollup-output.ts
sapphi-red: Doing that will break this test.
https://github.com/rolldown/rolldown/pull/3064/files#diff-effa1bb20ad69d9982696be2e4798f5ef323ae4129a0461a45ae63e69c331e64R20-R27
I'm not sure if people would do this though.
If we are not going to support that case, I think we should mark all properties in `SourceMap` as `readonly`.

underfin: Get it.

underfin (approved):

PR #3068, opened 2024-12-06
test: handle external dynamic import in `plugin/resolve-id/basic`

shulaoda:
### Description

It can now be properly enabled.

https://github.com/rolldown/rolldown/blob/60d7b16675a18df42e3cec0a5b74d0838e65694c/packages/rolldown/tests/fixtures/plugin/resolve-id/basic/_config.ts#L39


hyf0 (approved):
Thanks

PR #3085, opened 2024-12-09
feat: give warning for multiply notify option

underfin:
### Description

related https://github.com/rolldown/rolldown/pull/3037.

The rollup is create chokidar for each Bundler, it has more overhead at rolldown, because we need to create thread to listen event. So the notify option is used for all Bundler, here give warning for multiply notify options.


hyf0 (approved):
Is the warning means that only one of watch options would works?

underfin:
> Is the warning means that only one of watch options would works?

Yeah, the reason is  https://github.com/rolldown/rolldown/pull/3085#issue-2726174694

PR #3086, opened 2024-12-09
fix: revert require rewrite

### LINKED ISSUE #2684: [Feature Request]: keep `require` function name as-is

[`artichokie`](https://github.com/sapphi-red/artichokie), a library to run code in a worker, uses `Function::toString` on the main thread side and eval that string on the worker side.

When I use it like below:
```js
const worker = new Worker(() => {
    const path = require('node:path')
    return async (relative) => path.resolve(relative)
})

const result = await worker.run('./foo')
console.log(result) // /absolute/path/to/foo
```
and bundle that code by Rolldown. The output (sometimes?) becomes:
```js
const worker = new Worker(() => {
    const path = __require('node:path')
    return async (relative) => path.resolve(relative)
})

const result = await worker.run('./foo')
console.log(result) // /absolute/path/to/foo
```
(the `require` is renamed to `__require`.)

This code no longer runs because `__require` is not defined on the worker side (Because artichokie stringifies the function and put that on the worker side).

I know this is quite a hacky usecase, so I expect this issue to be closed as won't fix, but thought it would be nice to be supported.

Slightly related: if rolldown renames functions, it might need some feature like [`keepNames` in esbuild](https://esbuild.github.io/api/#keep-names).

No API changes. Just make rolldown to keep `require` as-is.

IWANABETHATGUY:
### Description
1. Closed https://github.com/rolldown/rolldown/issues/2684


crates/rolldown/src/runtime/runtime-tail-node.js
```diff
@@ -1,4 +1,4 @@
 // This is created by the rolldown. It's used to polyfill the global `require` function when you bundling code
 // with target `format: 'esm'` and `platform: 'node'`. It's used to make sure the global `require` could 
 // works in Node.js esm environment.
-export var __require = /* @__PURE__ */ createRequire(import.meta.url);
\ No newline at end of file
+export var require = /* @__PURE__ */ createRequire(import.meta.url);
```
IWANABETHATGUY, line 4: used in next pr.

crates/rolldown/tests/rolldown/issues/2685/_config.json
```diff
@@ -2,5 +2,6 @@
   "config": {
     "platform": "node"
   },
-  "hiddenRuntimeModule": false
+  "hiddenRuntimeModule": false,
+  "expectExecuted": false
```
IWANABETHATGUY, line 6: Will fixed in next pr.

crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
```diff
@@ -519,11 +519,6 @@ impl<'ast> ScopeHoistingFinalizer<'_, 'ast> {
       //  we just keep these `require` calls as it is
       if let Some(rec_id) = self.ctx.module.imports.get(&call_expr.span).copied() {
         let rec = &self.ctx.module.import_records[rec_id];
-        // use `__require` instead of `require`
-        if rec.meta.contains(ImportRecordMeta::CALL_RUNTIME_REQUIRE) {
```
hyf0, line 523: Consider the need for polyfill require.

crates/rolldown/src/runtime/runtime-tail-node.js
hyf0, line 4: ~~Keep using `__require` would be better.~~ `require` seems to be easily conflict with others. If happed, it still breaks the worker case. `require` is reserved in `cjs` format, so this would certainly be renamed to `require$1`.

~~Maybe keep using `__require` and allow to turn off `polyfill` for require.~~

@sapphi-red cc

crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
IWANABETHATGUY: then still need to rewrite `require`

PR #3103, opened 2024-12-10
feat(plugin/json): support `stringify: 'auto'`

sapphi-red:
### Description

Add support for `stringify: 'auto'`.


hyf0 (approved):

crates/rolldown_binding/src/options/plugin/binding_builtin_plugin.rs
```diff
@@ -88,14 +88,35 @@ pub struct BindingModulePreloadPolyfillPluginConfig {
 }
 
 #[napi_derive::napi(object)]
-#[derive(Debug, Deserialize, Default)]
-#[serde(rename_all = "camelCase")]
+#[derive(Debug, Default)]
```
sapphi-red, line 91: When is `Deserialize` used? Is it fine to simply remove it?

crates/rolldown_plugin_json/src/lib.rs
```diff
@@ -25,27 +34,62 @@ impl Plugin for JsonPlugin {
...
+            // TODO: escape key
+            default_object_code += &format!("  \"{key}\": {},\n", &serialize_value(value)?);
+          }
+        }
+        default_object_code += "}";
+
+        code += &format!("export default {default_object_code};\n");
+        return Ok(Some(HookTransformOutput {
+          code: Some(code),
+          map: Some(SourceMap::default()),
```
sapphi-red, line 60: The JS plugin has `map: { mappings: '' }`. Is it correct to use `SourceMap::default()` for that?

crates/rolldown_binding/src/options/plugin/binding_builtin_plugin.rs
hyf0: Yeah. It could be removed. It's basically for `#[serde(rename_all = "camelCase")]`. Seems napi-rs supports this by default.

crates/rolldown_plugin_json/src/lib.rs
hyf0: @underfin cc

underfin: Yeah. It could be using it.

Issue #3099, opened 2024-12-09
[infra] Migrate to `valibot` from `zod`


hyf0:
Contexts:
- One reason of using `zod` is that we want to automatically infer types from schema. But now, to write jsdoc, we have to define types first and then write schema to fit with that type.
- `valibot` is way smaller than `zod`.

~~The biggest the problem of this migrating is that rolldown's cli rely on zod's defined schema to generate cli commands descriptions.~~

`valibot` supports `description`: https://valibot.dev/api/description/. I guess we could reuse valibot's schema when migrating.

So we should first make cli not rely on zod's schema and then start the migrating.

hyf0:
https://valibot.dev/guides/migrate-from-zod/ supports most of the API similar to zod's, so the migrating won't be  a big work.

Another problem is that we are using #3096, which requires to write types explicitly.

https://github.com/rolldown/rolldown/blob/f2f09d7ac8be79a53493127fa4e073440334ada8/packages/rolldown/src/options/output-options-schema.ts#L167

If `validbot` doesn't have helper type like `ZodType` to infer schema from type, it will be a blocker for migrating.

shulaoda:
> Another problem is that we are using [#3096](https://github.com/rolldown/rolldown/pull/3096), which requires to write types explicitly.

Will this cause any conflicts? Should we wait for it to be merged before starting work? Before beginning, I will study and try using `valibot`.

shulaoda:
Related to #1955

hyf0:
@shulaoda 

> Should we wait for it to be merged before starting work?

Yes

hyf0:
Notes:

- We don't have to use `valibot`. It's just a good choice for now.

shulaoda:
> Notes:
> 
> * We don't have to use `valibot`. It's just a good choice for now.

In most cases, the input and output schemas are only used in the CLI, and apart from that, `zod` is only used in `utils/zod-ext.ts`.

hyf0:
> > Notes:
> > 
> > * We don't have to use `valibot`. It's just a good choice for now.
> 
> In most cases, the input and output schemas are only used in the CLI, and apart from that, `zod` is only used in `utils/zod-ext.ts`.

We will use it to validate config in `rolldown` API in the future.

shulaoda:
> * We don't have to use `valibot`. It's just a good choice for now.

[...]

PR #3106, opened 2024-12-10
chore(infra): migrate the `inputSchema` from `zod` to `valibot`

shulaoda:
### Description

Related to #3099


shulaoda:
> WIP: wait for [fabian-hiller/valibot#968](https://github.com/fabian-hiller/valibot/issues/968)

~~Since we haven't exposed the schema yet, we can temporarily ignore the issue by using as any, or by adding an input-options-schema.d.ts file to specify its type. In the future, if we use oxc, this issue might be resolved?~~ @hyf0 cc

After solving a problem, we can use `v.GenericSchema`.

packages/rolldown/src/schemas/input-options.ts
```diff
@@ -0,0 +1,273 @@
...
+    'onwarn',
+    'onLog',
+    'resolve',
+    'experimental',
+    'profilerNames',
+    'watch',
+  ],
+)
+
+export function validateInputCliOptions(options: any): boolean {
```
shulaoda: This is only temporary method.

I'm thinking that the cli input can't actually distinguish between `inputOptions` and `outputOptions`. Should we consider combining `inputOptionsSchema` and `outputOptionsSchema` into a single file?

shulaoda: Alternatively, we could use another approach by defining a custom `validateOptions(options)` method. First, we use the `inputOptionsSchema` to determine the corresponding input keys, then filter the `options` into `inputOptions` and `otherOptions`. We validate `inputOptions` and similarly extract `outputOptions` from `otherOptions` for validation. Finally, any remaining keys would be categorized as `ignoreOptions`, and we could issue a warning indicating they are being ignored.

shulaoda: @hyf0 cc

hyf0: > inputOptionsSchema and outputOptionsSchema into a single file?

A giant `validator.ts ` is fine.

packages/rolldown/package.json
```diff
@@ -98,6 +98,8 @@
     "dtsHeader": "type MaybePromise<T> = T | Promise<T>\ntype Nullable<T> = T | null | undefined\ntype VoidNullable<T = void> = T | null | undefined | void\nexport type BindingStringOrRegex = string | RegExp\n\n"
   },
   "dependencies": {
+    "@valibot/to-json-schema": "1.0.0-beta.3",
```
hyf0, line 101: What's this for?

shulaoda: Prepare for this:

https://github.com/rolldown/rolldown/blob/149566a616f4fb29e0ea41a9fd8fa5d1ba8ce9c4/packages/rolldown/src/cli/arguments/schema.ts#L34-L36

hyf0: Why we need json schema? Does valibot itself already able to validate options?

shulaoda: The current CLI logic relies on a JSON schema, which is used to process some of the options before ultimately parsing them with `parseArgs`.

https://github.com/rolldown/rolldown/blob/149566a616f4fb29e0ea41a9fd8fa5d1ba8ce9c4/packages/rolldown/src/cli/arguments/index.ts#L18-L63

hyf0 (approved):
Thanks

packages/rolldown/package.json
hyf0: Get it.

PR #3110, opened 2024-12-10
refactor: remove all derived `Deserialize` in `rolldown_binding` crate

### LINKED ISSUE #3109: Refactor: remove all derived `Deserialize` in `rolldown_binding` crate
Yeah. It could be removed. It's basically for `#[serde(rename_all = "camelCase")]`. Seems napi-rs supports this by default.

_Originally posted by @hyf0 in https://github.com/rolldown/rolldown/pull/3103#discussion_r1877541921_

- `struct`s in `rolldown_binding` doesn't need to have a derived `Deserialize` because we won't create them by deserializing anything.
- `#[serde(rename_all = "camelCase")]` is supported by napi-rs itself
- This should reduce binary sizes.
- If we need to rename the property name, we could just simply write
```rust
struct Foo {
  whatever_fooBar: ...,
}
```
`rolldown_binding` won't be consumed by other rust crate.

shulaoda:
### Description

closes #3109

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2468,7 +2468,6 @@ dependencies = [
  "rolldown_tracing",
  "rolldown_utils",
  "rustc-hash",
- "serde",
  "tracing",
 ]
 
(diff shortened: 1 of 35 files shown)


hyf0 (approved):
Thanks!

PR #3111, opened 2024-12-10
fix: wrong output when external import name is string literal

### LINKED ISSUE #3094: Import string name from a `external` module generate invalid code.
As you can see the output is already wrong before, I will fix it in another pr.

_Originally posted by @IWANABETHATGUY in https://github.com/rolldown/rolldown/pull/3082#discussion_r1875836880_

IWANABETHATGUY:
### Description
Closed: https://github.com/rolldown/rolldown/issues/3094


crates/rolldown_common/src/types/named_import.rs
```diff
@@ -50,14 +57,80 @@ impl Display for Specifier {
...
+  fn from(rstr: Rstr) -> Self {
+    if is_validate_identifier_name(rstr.as_str()) {
+      Self::Identifier(rstr)
+    } else {
+      Self::String(rstr)
+    }
+  }
+}
+
+impl<'a, 'ast: 'a> From<&'a ModuleExportName<'ast>> for ImportOrExportName {
```
IWANABETHATGUY, line 131: Not sure we need to reuse `ModuleExportName`, this may cause conflicted in some scenario.

hyf0:
I've sent my reviews on discord.

PR #3135, opened 2024-12-12
perf(rolldown): remove `AppendOnlyVec`

Boshen:
It's still costly to sync the atomics in `AppendOnlyVec`: https://docs.rs/append-only-vec/latest/src/append_only_vec/lib.rs.html#41-45

A simple map + collect should suffice.

image


Boshen:
I don't know what I am doing :-/

PR #3140, opened 2024-12-12
feat(string_wizard): add `serde` feature

Boshen:
When `string_wizard` is used inside `oxc`, and with `oxc_index` having the `serde` feature enabled, feature unification breaks apart. And there is no way to force enable a dependencie's dependency's feature. This is the last resort :-(

```
error[E0433]: failed to resolve: use of undeclared crate or module `serde`
 --> crates/string_wizard/src/chunk.rs:5:1
  |
5 | / oxc_index::define_index_type! {
6 | |     pub struct ChunkIdx = u32;
7 | | }
  | |_^ use of undeclared crate or module `serde`
  |
  = note: this error originates in the macro `$crate::__internal_maybe_index_impl_serde` which comes from the expansion of the macro `oxc_index::define_index_type` (in Nightly builds, run with -Z macro-backtrace for more info)
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1731,6 +1731,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5eca5d9726cd0a6e433debe003b7bc88b2ecad0bb6109f0cef7c55e692139a34"
 dependencies = [
  "rayon",
+ "serde",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


hyf0 (approved):

Boshen:
Thank you, a new version after merge is needed.

hyf0:
> Thank you, a new version after merge is needed.

0.0.25

PR #3145, opened 2024-12-13
feat: expose more fine-grained minify option

sapphi-red:
### Description

Changes the `minify` option to accept more fine-grained options.

```ts
export interface OutputOptions {
  minify?: boolean | 'dead-code-elimination-only' | MinifyOptions
  // omitted other properties
}

export interface MinifyOptions {
  mangle: boolean
  compress: boolean
  removeWhitespace: boolean
}
```

I wanted to enable deadCodeElimination by default (when when `renderChunks` or other hooks changed the output) as that would be a more sensible default. But I didn't do that because if I run it, the `region` comments gets stripped.


sapphi-red:
Let's wait for oxc-minifier to expose the new options.

sapphi-red:
Probably the perf regression is caused by this part, it runs some compression pass twice.
https://github.com/rolldown/rolldown/pull/3145/files#diff-536ccef0bb17e2204824effeb837078ef269727f8fa154dbeaa56cd4e24bb26fR120-R135

PR #3147, opened 2024-12-13
refactor: transform source

underfin:
### Description

Here has some changes at here.

- avoid unnecessarily `to_string`
- avoid creating unnecessarily `HookTransformArgs`
- perf: avod create the sourcemap if the code hasn't changed, the vite has some usages.

### DIFF
--- a/crates/rolldown_plugin/src/plugin_driver/build_hooks.rs
+++ b/crates/rolldown_plugin/src/plugin_driver/build_hooks.rs
@@ -170,20 +170,20 @@ impl PluginDriver {
 
   pub async fn transform(
     &self,
-    args: &HookTransformArgs<'_>,
+    id: &str,
+    original_code: String,
     sourcemap_chain: &mut Vec<SourceMap>,
     side_effects: &mut Option<HookSideEffects>,
-    original_code: &str,
     module_type: &mut ModuleType,
   ) -> Result<String> {
-    let mut code = args.code.to_string();
+    let mut code = original_code;
     let mut original_sourcemap_chain = std::mem::take(sourcemap_chain);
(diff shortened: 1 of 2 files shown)


crates/rolldown_plugin/src/plugin_driver/build_hooks.rs
```diff
@@ -238,17 +237,23 @@ impl PluginDriver {
       if map.get_source_content(0).map_or(true, str::is_empty) {
         map.set_source_contents(vec![original_code]);
       }
-      map
+      Some(map)
+    } else if let Some(code) = code {
+      if original_code == code {
```
hyf0, line 242: Seems a high cost operation if code doesn't change.

Does rollup have this too?

hyf0: I'll suggest if user explicitly return the code, we consider it changed no matter if really changes or not.

hyf0 (approved):

crates/rolldown_plugin/src/plugin_driver/build_hooks.rs
underfin: > Seems a high cost operation if code doesn't change.
> 
> Does rollup have this too?

The rollup sourcemap implemention is different with rolldown.



> I'll suggest if user explicitly return the code, we consider it changed no matter if really changes or not.

Here look like has a high overhead, but it only happened at only return code and not return map, here we need to create sourcemap for it and also need to remapping, It is worth to using a pre-check to avoid it. 
Yeah. It has overhead for the code actually changed and the code length is not changed. But the fetching module and render module is running at parallel, make the slow task as much as possible faster finish will get better performance.

PR #3151, opened 2024-12-13
refactor: split extensions into `ast_ext` for better DX

shulaoda:
### Description

The `ext/mod.rs` file in `rolldown_ecmascript_utils` contained many extensions related to `oxc::ast::ast` nodes, but accumulating them in a single file wasn't ideal for DX. Therefore, I refactored them into separate files under the `ast_ext` folder.


hyf0 (approved):
Thanks

PR #3152, opened 2024-12-13
feat: define `itoa!` macro to simplify `itoa::Buffer::new().format(..)`

shulaoda:
### Description

Related to #2107 and https://github.com/web-infra-dev/rspack/pull/7762

I defined the `itoa!` macro to improve the DX when introducing `itoa` for `rspack`. I found it to be a good idea, so I decided to bring it into `rolldown`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.2"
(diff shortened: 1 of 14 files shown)


hyf0:
We want to avoid macro if possible. In this case typing a few more words won't be a big problem.

shulaoda:
> We want to avoid macro if possible. In this case typing a few more words won't be a big problem.

Got it.

PR #3154, opened 2024-12-14
refactor: improve the DX of `ModuleTask`

shulaoda:
### Description

1. Simplify the error return type
2. Declare variable before use


hyf0 (approved):
Thanks

PR #3156, opened 2024-12-15
refactor: use safer logic to get `ModuleTable#modules`

shulaoda:
### Description

The `importers` and `modules` in `intermediate_normal_modules` are mapped one-to-one by `idx`. If a module fails to be fetched during the `fetch_all_modules` process, `flatten()` will break the corresponding mapping. 

https://github.com/rolldown/rolldown/blob/ddbe0a9e8252f61380a4a5b57d82e32a00d0791b/crates/rolldown/src/module_loader/module_loader.rs#L407-L414
https://github.com/rolldown/rolldown/blob/ddbe0a9e8252f61380a4a5b57d82e32a00d0791b/crates/rolldown/src/module_loader/module_loader.rs#L418

Although most cases are already handled by the code below:
https://github.com/rolldown/rolldown/blob/ddbe0a9e8252f61380a4a5b57d82e32a00d0791b/crates/rolldown/src/module_loader/module_loader.rs#L387-L389

What I’ve done is simply make the code safer. If my understanding is incorrect, feel free to close this.

### DIFF
--- a/crates/rolldown/src/module_loader/module_loader.rs
+++ b/crates/rolldown/src/module_loader/module_loader.rs
@@ -410,29 +410,30 @@ impl ModuleLoader {
       .modules
       .into_iter()
       .enumerate()
-      .filter_map(|(id, module)| {
-        module.map(|mut module| {
-          if let Some(module) = module.as_normal_mut() {
-            let id = ModuleIdx::from(id);
-            // Note: (Compat to rollup)
-            // The `dynamic_importers/importers` should be added after `module_parsed` hook.
-            let importers = std::mem::take(&mut self.intermediate_normal_modules.importers[id]);
-            for importer in &importers {
-              if importer.kind.is_static() {
-                module.importers.push(importer.importer_path.clone());
-              } else {
-                module.dynamic_importers.push(importer.importer_path.clone());
-              }
-            }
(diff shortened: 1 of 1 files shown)


crates/rolldown/src/module_loader/module_loader.rs
```diff
@@ -402,35 +402,37 @@ impl ModuleLoader {
...
-        if let Some(module) = module.as_normal_mut() {
-          // Note: (Compat to rollup)
-          // The `dynamic_importers/importers` should be added after `module_parsed` hook.
-          let importers = std::mem::take(&mut self.intermediate_normal_modules.importers[id]);
-          for importer in &importers {
-            if importer.kind.is_static() {
-              module.importers.push(importer.importer_path.clone());
-            } else {
-              module.dynamic_importers.push(importer.importer_path.clone());
+      .filter_map(|(id, module)| {
```
hyf0: I don't think `filter_map` is proper too. Should just use `unwrap` for such case. Having `None` in the `intermediate_normal_modules` is invalid and shouldn't be ignored.

shulaoda: Got it.

hyf0 (approved):
Thanks

PR #3157, opened 2024-12-15
refactor: use `visit_span` instead for `EnsureSpanUniqueness`

shulaoda:
### Description

Related to https://github.com/oxc-project/oxc/issues/4799 and https://github.com/oxc-project/oxc/pull/7816

Since oxc already supports `visit_span`, I tried to refactor it.

I’m curious if empty spans other than `SPAN` exist, such as `Span { start: 5, end: 5 }`. If we can confirm they don't, the logic could be further simplified.

PS: 
- Compared to the original logic, we might perform more `if it.start == it.end` checks, but the impact of this is negligible.
- I didn’t use `it.is_empty()` for `it.start == it.end` because `is_empty` is not marked as `#[inline]`, and I’m concerned it might affect performance.
- I’ve looked into this issue in the past, which is why I was eager to propose this refactor after `oxc` support it. If there’s a better approach or consideration, feel free to close this.


shulaoda:
I don't know what I'm doing, and sticking with the original logic might be the better choice. :-/

PR #3162, opened 2024-12-16
refactor: improve the error handling of `RuntimeModuleTask#run`

shulaoda:
### Description

Currently, if an error occurs in `RuntimeModuleTask#run`, it will never be handled (except in the case of `wasm`). 

https://github.com/rolldown/rolldown/blob/e8033ef406271a01080fc8a07d426eb9387e2a05/crates/rolldown/src/module_loader/runtime_module_task.rs#L61-L67
https://github.com/rolldown/rolldown/blob/e8033ef406271a01080fc8a07d426eb9387e2a05/crates/rolldown/src/module_loader/module_loader.rs#L112-L122

While we know that `RuntimeModuleTask#run` is unlikely to fail under normal circumstances, if an error does occur, the entire program will never be able to terminate properly because `remaining > 0` will be always. 

Therefore, I aligned it with `ModuleTask#run` to ensure that errors are handled correctly.

### DIFF
--- a/crates/rolldown/src/module_loader/module_loader.rs
+++ b/crates/rolldown/src/module_loader/module_loader.rs
@@ -94,7 +94,8 @@ impl ModuleLoader {
         })?
       },
     };
-    let common_data = Arc::new(TaskContext {
+
+    let shared_context = Arc::new(TaskContext {
       options: Arc::clone(&options),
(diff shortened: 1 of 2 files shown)


hyf0 (approved):
Thanks

PR #3165, opened 2024-12-17
refactor: simplify return type of `resolve_user_defined_entries`

shulaoda:
### Description

We have implemented our own `DiagnosableResolveError` which is from `oxc_resolver::ResolveError`, but its applicability was too limited, so I decided to retain the original error handling logic, which is based on errors converted from `anyhow::Error`.

https://github.com/rolldown/rolldown/blob/a519f4c704baf83ad0f4dce383cdcc88783c15ef/crates/rolldown_error/src/events/resolve_error.rs#L7-L13

### DIFF
--- a/crates/rolldown/src/stages/scan_stage.rs
+++ b/crates/rolldown/src/stages/scan_stage.rs
@@ -6,7 +6,7 @@ use rolldown_common::{
   dynamic_import_usage::DynamicImportExportsUsage, EntryPoint, ImportKind, ModuleIdx, ModuleTable,
   ResolvedId, RuntimeModuleBrief, SymbolRefDb,
 };
-use rolldown_error::{BuildDiagnostic, BuildResult};
+use rolldown_error::{BuildDiagnostic, BuildResult, ResultExt};
 use rolldown_fs::OsFileSystem;
 use rolldown_plugin::SharedPluginDriver;
 use rolldown_resolver::ResolveError;
@@ -139,7 +139,7 @@ impl ScanStage {
           ResolveError::PackagePathNotExported(..) => {
             errors.push(BuildDiagnostic::unresolved_entry(args.specifier, Some(e)));
           }
-          _ => return Err(e)?,
+          _ => return Err(e).map_err_to_unhandleable()?,
         },
       }
     }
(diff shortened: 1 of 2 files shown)


crates/rolldown/src/stages/scan_stage.rs
```diff
@@ -140,17 +139,15 @@ impl ScanStage {
...
-            return Err(e.into());
-          }
+          _ => return Err(e)?,
         },
       }
     }
 
     if !errors.is_empty() {
-      return Ok(Err(errors.into()));
+      Err(errors)?;
```
underfin, line 148: Here look like missing `return`.

shulaoda: Actually, this is syntactic sugar. `return Err(errors.into());` and `Err(errors)?;` are equivalent, because the `?` operator automatically performs both the `return` and `into` conversion for an `Error`.

shulaoda: This might be confused, I will add `return` to them. Since it was also used in the previous PRs, I will open a separate PR to resolve it.

underfin: I get it, i'm not notice it. I think it's ok.

crates/rolldown_error/src/build_error/mod.rs
```diff
@@ -120,6 +120,12 @@ impl From<anyhow::Error> for BuildDiagnostic {
   }
 }
 
+impl From<oxc_resolver::ResolveError> for BatchedBuildDiagnostic {
+  fn from(err: oxc_resolver::ResolveError) -> Self {
+    Self::new(vec![BuildDiagnostic::unhandleable_error(err.into())])
```
underfin: cc @hyf0 The `ResolveError` look like needs to give a specific error.

underfin (approved):

crates/rolldown_error/src/build_error/mod.rs
shulaoda: We have `DiagnosableResolveError` for it,  but its applicability was too limited.

hyf0: Yeah. It could be improved.

PR #3178, opened 2024-12-18
chore(test): separate watch test

underfin:


hyf0 (approved):

hyf0:
Seems this PR breaks `just test-node rolldown --update` @underfin cc

PR #3181, opened 2024-12-18
fix: meta.dependency should consider reexport chain owner that has side effects.

IWANABETHATGUY:


hyf0 (approved):

crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
```diff
@@ -140,7 +150,16 @@ impl LinkStage<'_> {
...
       meta.resolved_exports = resolved_exports;
     });
-
+    let side_effects_modules = self
+      .module_table
+      .modules
+      .iter_enumerated()
+      .filter_map(|(idx, item)| item.side_effects().has_side_effects().then(|| idx))
+      .map(|item| item)
+      .inspect(|item| {
```
hyf0: Shouldn't be called in release.

PR #3182, opened 2024-12-18
feat: emit a warning when `minify: true`

### LINKED ISSUE #3173: [Feature Request]: Emit a warning when `minify: true`

Since the builtin minifier is still under development, we need to emit a warning for `minify: true` to warn that it's not ready for production usage.

.

shulaoda:
### Description

closes #3173 

This is a simple feature issue that I noticed nobody was working on, so I implemented it.


hyf0 (approved):

hyf0:
Thanks

PR #3194, opened 2024-12-20
refactor: improve the error return type for `GenerateStage`

shulaoda:
### Description

- Remove unnecessary usages of `anyhow::Result`
- Use `BuildResult` instead of `anyhow::Result` for the methods of `GenerateStage`


hyf0 (approved):

hyf0:
Thanks

PR #3195, opened 2024-12-20
fix(iife/umd): named exports if module wrapped

underfin:

### DIFF
--- a/crates/rolldown/tests/rolldown/function/extend/entry-wrapped-cjs-named/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/function/extend/entry-wrapped-cjs-named/artifacts.snap
@@ -1,0 +1,99 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+snapshot_kind: text
+---
+# Assets
+
+## main.js
+
+```js
+
+
+//#region main.js
+var require_main = __commonJS({ "main.js"() {
+	require_main();
+} });
+
+//#endregion
+export default require_main();
+
+```
+---
+
+Variant: (format: Cjs)
+
+# Assets
+
+## main.js
+
+```js
+
+
+//#region main.js
+var require_main = __commonJS({ "main.js"() {
+	require_main();
+} });
+
+//#endregion
+module.exports = require_main();
+
+```
+---
+
+Variant: (format: Iife)
+
+# Assets
+
+## main.js
+
+```js
+(function(exports) {
+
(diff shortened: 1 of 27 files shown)


crates/rolldown/tests/rolldown/function/extend/entry-wrapped-cjs-named/artifacts.snap
```diff
@@ -0,0 +1,99 @@
...
+//#region main.js
+var require_main = __commonJS({ "main.js"() {
+	require_main();
+} });
+
+//#endregion
+Object.defineProperty(exports, 'default', {
+  enumerable: true,
+  get: function () {
+    return require_main();
```
underfin, line 63: Here should be add `default` to `exports`.

crates/rolldown_testing_config/src/lib.rs
```diff
@@ -1,13 +1,16 @@
...
-use rolldown_common::{BundlerOptions, OutputFormat};
+use rolldown_common::{BundlerOptions, OutputExports, OutputFormat};
 use schemars::JsonSchema;
 use serde::Deserialize;
 
 #[derive(Deserialize, JsonSchema, Default)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 pub struct ConfigVariant {
   pub format: Option<OutputFormat>,
+  pub extend: Option<bool>,
```
hyf0, line 11: These new fields should be added in

https://github.com/rolldown/rolldown/blob/d5e30898d9a34413afa524212aee476c1f9e7ca5/crates/rolldown_testing_config/src/lib.rs#L38

, so we could visualize them easily in snapshot.

hyf0 (approved):

crates/rolldown/tests/rolldown/function/extend/entry-wrapped-cjs-named/artifacts.snap
```diff
@@ -0,0 +1,99 @@
...
+Variant: (exports: Named, extend: true, format: Iife, name: "module")
+
+# Assets
+
+## main.js
+
+```js
+(function(exports) {
+
+Object.defineProperty(exports, '__esModule', { value: true });
```
hyf0: Seems wrong. why does not cjs format emit this?

underfin: Fixed it.

PR #3196, opened 2024-12-20
chore(rolldown): oxc v0.43.0

Boshen:


crates/rolldown/src/utils/renamer.rs
```diff
@@ -144,11 +144,12 @@ impl<'name> Renamer<'name> {
       stack: &mut Vec<Cow<FxHashMap<Rstr, u32>>>,
       canonical_names: &mut FxHashMap<SymbolRef, Rstr>,
     ) {
-      let bindings = module.scope.get_bindings(scope_id);
+      let mut bindings = module.scope.get_bindings(scope_id).iter().collect::<Vec<_>>();
+      bindings.sort_unstable_by_key(|(_, symbol_id)| *symbol_id);
```
Boshen, line 148: `get_bindings` no longer returns an indexmap (order keys by insertion order), so we need to sort here.

Boshen (commented):
Most changes are for replacing `&CompactString` with `&str`.

hyf0 (approved):

PR #3212, opened 2024-12-24
fix(interop): should detect importer is esm module

underfin:
### Description

The interop should detect importer module is esm if imported is cjs module, the https://github.com/rolldown/rolldown/pull/2769 only fixed with `import` case, not updated `reexports` and `dynamic import` case, the pr fixed other cases.


crates/rolldown/tests/rolldown/function/format/app/import/artifacts.snap
```diff
@@ -19,7 +20,7 @@ module.exports = 1;
 __toCommonJS(exports);
 var external = require("external");
 var cjs_exports = __toESM(require("cjs.js"));
-var commonjs_exports = __toESM(require("commonjs.mjs"), 1);
+var commonjs_exports = __toESM(require("commonjs.mjs"));
```
underfin, line 23: The perivous is error, because it read `commonjs.mjs` as importer.

crates/rolldown/tests/rolldown/cjs_compat/reexports_from_cjs/artifacts.snap
hyf0, line 1: I run this test in the main branch and it gets the same output as this PR. https://github.com/rolldown/rolldown/pull/3215

If you want to fix the issue of `app` format output, could you limit the changes only happen on isolating finalizer and not touch scope hoisting finalizer?

---

See https://github.com/rolldown/rolldown/pull/3212#discussion_r1896628781

crates/rolldown/src/module_finalizers/scope_hoisting/mod.rs
```diff
@@ -95,12 +95,6 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
         // `import_foo`
         let binding_name_for_wrapper_call_ret = self.canonical_name_for(rec.namespace_ref);
 
-        // If the module is an ESM module that follows the Node.js ESM spec, such as
-        // - extension is `.mjs`
-        // - `package.json` has `"type": "module"`
-        // , we need to consider to stimulate the Node.js ESM behavior for maximum compatibility.
-        let should_consider_node_esm_spec = self.ctx.module.ecma_view.def_format.is_esm();
```
hyf0, line 102: Could you keep this part of code if it doesn't have bugs? It makes understanding that why we should add a `1` in the `toESM` call easier. I run the PR's test in the main branch and it gets the same output as this PR. https://github.com/rolldown/rolldown/pull/3215

Other changes are fine.

crates/rolldown/tests/rolldown/cjs_compat/reexports_from_cjs/artifacts.snap
underfin: I missing some files, please check it again. The main branch render result is this. 

``` .js
var import_commonjs = __toESM(require_commonjs(), 1);
__reExport(reexports_exports, __toESM(require_commonjs())); // here is incorrect
```

crates/rolldown/src/module_finalizers/scope_hoisting/mod.rs
underfin: I moved the comment to `crates/rolldown_common/src/module/normal_module.rs`. That has one `Interop` mode abstract。

underfin: Updated.

hyf0 (approved):

PR #3249, opened 2024-12-27
feat: remove unnecessary compat `default` property access when there is no `module.exports ` assignment

IWANABETHATGUY:
### Description
1. The unresolved issue, transitive `module.exports` like react pattern. 
```js
// react.production.js
exports.a = xx;
... snip

// react.js
module.exports = require('react.production.js')

// index.js
import React from 'react'

React.useState(); // generated code still has `default` property access, since `react.js`  used `module.exports` to export something.
```
Will handle this case in next pr.


crates/rolldown/tests/rolldown/cjs_compat/unnecessary_compat_default_property_access/commonjs.js
```diff
@@ -0,0 +1 @@
+exports.foo = 1000;
```
IWANABETHATGUY, line 1: It is safe to optimize, even there is `exports[computed] = ` assignment

crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
```diff
@@ -619,10 +620,25 @@ impl BindImportsAndExportsContext<'_> {
...
           }
-          Specifier::Literal(alias) => MatchImportKind::NormalAndNamespace {
-            namespace_ref: importer_record.namespace_ref,
-            alias: alias.clone(),
-          },
+          Specifier::Literal(alias) => {
+            if alias.as_str() == "default" {
+              // `ExternalModule` has no `ExportsKind`, here the module must be `NormalModule`
+              let module = self.index_modules[tracker.importee].as_normal().unwrap();
+              if module.ast_usage.contains(EcmaModuleAstUsage::ModuleRef) {
```
underfin, line 627: It's not safe, because the importee commonjs module maybe contains `__esModule` flag.

IWANABETHATGUY: Can you provide a example?

IWANABETHATGUY: I got it.

IWANABETHATGUY:
I found a more general way.

Issue #3243, opened 2024-12-26
[Bug]: rolldown:runtime: es5 umd not working due to non-es5 code being injected


kmalakoff:
### Reproduction link or steps

Use a tsconfig file with es5 as a target and format umd. 

### What is expected?

It should only inject es5 compatible code

### What is actually happening?

It's injecting es6 runtime. See the arrow functions:

```
var __copyProps = (to, from, except, desc) => {
	if (from && typeof from === "object" || typeof from === "function") for (var keys = __getOwnPropNames(from), i = 0, n = keys.length, key; i < n; i++) {
		key = keys[i];
		if (!__hasOwnProp.call(to, key) && key !== except) __defProp(to, key, {
			get: ((k) => from[k]).bind(null, key),
			enumerable: !(desc = __getOwnPropDesc(from, key)) || desc.enumerable
		});
	}
	return to;
};
var __toESM = (mod, isNodeMode, target) => (target = mod != null ? __create(__getProtoOf(mod)) : {}, __copyProps(isNodeMode || !mod || !mod.__esModule ? __defProp(target, "default", {
	value: mod,
	enumerable: true
}) : target, mod));

```

### System Info

```Shell
System:
    OS: macOS 15.2
    CPU: (14) arm64 Apple M4 Pro
    Memory: 4.21 GB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.12.0 - ~/.nvm/versions/node/v22.12.0/bin/node
    npm: 11.0.0 - /usr/local/bin/npm
    pnpm: 9.4.0 - ~/Library/pnpm/pnpm
    Watchman: 2024.12.02.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 131.0.6778.205
    Safari: 18.2
```

### Any additional comments?

I'm excited to use rolldown! Please expedite your es5 target testing

underfin:
Here has two things need to do.
- ~~The rolldown `target` option is not enable for all files~~
- The `target` is not binding to js side.
- respect the fields at `tsconfig.json`

hyf0:
- The team has few thoughts on es5 support. But the current conclusion is there will be no official support on transforming output to es5.
  - There might will be a third-party plugin to transform the the bundled output to es5.
- Rolldown doesn't aware of `tsconfig.json` and we gonna keep this behavior.
[...]

PR #3251, opened 2024-12-27
chore: add target lowering test case.

underfin:
### Description
related https://github.com/rolldown/rolldown/issues/3243


IWANABETHATGUY:
This fix seems not working, because the real issue is `target` is not expose in js binding. The `transform` test is added before, you could expose the `target` in binding and add a test in js side.

IWANABETHATGUY (approved):
I rename the pr title

PR #3257, opened 2024-12-28
feat: esm_flag analyzer

IWANABETHATGUY:

PR #3258, opened 2024-12-28
TEST feat: eliminate_unnecessary_default

IWANABETHATGUY:
### Description

To ensure that the behavior remains consistent with the previous interop behavior after removing the `default` property access, we need to identify all cases where `import_default_alias.default == import_default_alias`. This can be inferred from the runtime.

```js
var __toESM = (mod, isNodeMode, target) => {
    return (
        __copyProps(
            isNodeMode || !mod || !mod.__esModule
                ? __defProp(target, "default", {
                        value: mod,
                        enumerable: true,
                    })
                : target,
            mod,
        )
    );
};
```

It is not difficult to see that the conditions are:
1. `isNodeMode` is `true || 1`
2. The module itself is an empty module
3. `mod.__esModule` is false

Condition one can be obtained from [this link](https://github.com/rolldown/rolldown/blob/eb47e2a4b5e07de61e6eb8101f4f07ecbf7c9c30/crates/rolldown_common/src/module/normal_module.rs?plain=1#L191-L201).
Condition two can be inferred through analysis; an empty module will default to being treated as `esm`, so the only case is `module.exports = undefined`.
Condition three can also be inferred through analysis, almost done in https://github.com/rolldown/rolldown/pull/3257.

However, it is important to note that when accessing the property `default` after `import_default_alias`, an unsafe situation may arise. For example:
```js
// index.js
import foo from './foo.js';
console.log(foo.default.bar === 123);
// foo.js
module.exports = { default: { bar: 123 } };
```

Here, condition 1 is not satisfied, condition 2 is not satisfied, but condition 3 is satisfied. Therefore, after `__toEsm`, the `mod` object will be transformed into:
```js
{
  default: { default: { bar: 123 } }
}
```
In this case, `foo.default.bar` is no longer equal to `foo.default.default.bar`. Specifically,
`foo.default.bar === undefined`, 
`foo.default.default.bar === 123`.

Currently, there are two solutions:
- One approach is to allow the `VisitMut` trait to access the parent node so that when rewriting `import_default_alias`, it knows whether the `default` property is accessed.
    - **Advantages**: This method is relatively intuitive and has low overhead.
    - **Disadvantages**: Currently, the `Traverse` trait requires mutating semantic data. We cannot mutate semantic data in the `module_finalizer` phase (which violates borrow rules). We need to implement something similar to a `TraverseSimple` trait in oxc that does not change the semantic data version but allows access to parent nodes during AST mutation.
- The other approach involves recording each time a `MemberExpression` is accessed, checking the span during the rewrite of `IdentifierReference` to see if it meets the conditions. 
    - **Disadvantages**: This method has higher overhead (compared to the first solution) and requires additional data structures for recording.

---

### DIFF
--- a/crates/rolldown/src/ast_scanner/esmodule_flag_analyzer.rs
+++ b/crates/rolldown/src/ast_scanner/esmodule_flag_analyzer.rs
@@ -1,0 +1,159 @@
+use oxc::ast::{
+  ast::{self, Expression, IdentifierReference, PropertyKey},
+  AstKind,
+};
+use rolldown_ecmascript_utils::ExpressionExt;
+
+use super::AstScanner;
+
+impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
+  pub fn check_es_module_flag(
+    &self,
+    ident: &IdentifierReference<'ast>,
+    ty: EsModuleFlagCheckType,
+  ) -> Option<bool> {
+    let cursor = self.visit_path.len() - 1;
+    let parent = self.visit_path.get(cursor)?;
+    match parent {
+      AstKind::MemberExpression(member_expr) => match ty {
+        // two scenarios:
+        // 1. module.exports.__esModule = true;
+        // 2. Object.defineProperty(module.exports, "__esModule", { value: true });
+        EsModuleFlagCheckType::ModuleExportsAssignment => {
+          let property_name = member_expr.static_property_name()?;
+          if property_name != "exports" {
+            return Some(false);
+          }
+          let parent_parent_kind = self.visit_path.get(cursor - 1)?;
+          match parent_parent_kind {
+            AstKind::MemberExpression(parent_parent) => {
+              let property_name = parent_parent.static_property_name()?;
+              if property_name != "__esModule" {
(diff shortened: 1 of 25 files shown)


IWANABETHATGUY:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3258?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3258** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3258?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3258?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3257** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3257?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #3259, opened 2024-12-29
test: resolve processes not exiting in test cases

shulaoda:
### Description

When using `gracefulCancel: true`, the termination signal can only be obtained via `getCancelSignal`, but our watch implementation uses `onExit`, which causes the process to never terminate.

```bash
[plugin shim]
Error: EBUSY: resource busy or locked, copyfile 'D:\shulaoda\rolldown\packages\rolldown\src\rolldown-binding.win32-x64-msvc.node' -> 'D:\shulaoda\rolldown\packages\rolldown\dist\shared\rolldown-binding.win32-x64-msvc.node'
```

image

### DIFF
--- a/packages/rolldown/tests/cli/cli-e2e.test.ts
+++ b/packages/rolldown/tests/cli/cli-e2e.test.ts
@@ -204,10 +204,10 @@ describe('watch cli', () => {
     const controller = new AbortController()
     execa({
       cwd,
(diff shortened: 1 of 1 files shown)


hyf0:
@underfin cc

underfin (approved):

underfin:
Thank you.

Issue #3255, opened 2024-12-28
[Refactor]: Move hook filter related logic to `rolldown_binding`


hyf0:
Hook filter is bring up to improve the performance of calling js callback in rust. Currently we implement it as first-class concept in the `rolldown_plugin`, but it's not necessary.

Hook filter is only for js callbacks, so related logic could be moved to `rolldown_binding`. In this way, we could reduce the complexity of `rolldown_plugin` itself and keep related concepts lean.

hyf0:
Taken by @shulaoda

PR #3260, opened 2024-12-29
refactor(rust): tweak code of `rolldown_binding`

shulaoda:
### Description

Before starting on #3255, improve the DX of `rolldown_binding`.


hyf0 (approved):
Thanks

PR #3262, opened 2024-12-29
chore(justfile): fix test-update

shulaoda:
### Description

              Seems this PR breaks `just test-node rolldown --update` @underfin cc

_Originally posted by @hyf0 in https://github.com/rolldown/rolldown/issues/3178#issuecomment-2562646748_
            

I don't know why it needs to be split. 
Perhaps we can control whether to run the watch-specific test cases separately using the `-t` command.

### DIFF
--- a/packages/rollup-tests/package.json
+++ b/packages/rollup-tests/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "scripts": {
     "test": "mocha --file ./src/intercept/check.js test/test.js",
-    "test-u": "mocha --jobs 1 --file ./src/intercept/update-test-status.js test/test.js",
+    "test--update": "mocha --jobs 1 --file ./src/intercept/update-test-status.js test/test.js",
     "type-check": "tsc --noEmit"
   },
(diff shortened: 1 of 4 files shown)


packages/rollup-tests/package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "scripts": {
     "test": "mocha --file ./src/intercept/check.js test/test.js",
-    "test--update": "mocha --jobs 1 --file ./src/intercept/update-test-status.js test/test.js",
+    "test-u": "mocha --jobs 1 --file ./src/intercept/update-test-status.js test/test.js",
```
underfin: It will break `just test-node rollup --update`. We need to fix 

```
test-update:
    just test-rust
    just test-node all -u -> --update
```

shulaoda: Indeed, I intended to add a `test-u` command to support both `-u` and `--update`, but it seems I forgot to push it.

underfin (approved):

PR #3265, opened 2024-12-30
refactor: move hook filter related logic to `rolldown_binding`

### LINKED ISSUE #3255: [Refactor]: Move hook filter related logic to `rolldown_binding`

### LINKED ISSUE #3256: [Infra] Remove some unnecessary tests in nodejs
We intend the tests of nodejs are only for testing the API. The correctness of features are ensured by tests of rust's side.

Now there's are some tests could removed.

For example:

For feature of `shimMissingExports`, in nodejs,we only need to ensure

- ~~The default value(normalized value if unset) of `shimMissingExports` is `false`~~ This would be done using a snapshot test.
- If user pass `true`, in the `buildStart` hook, `NormalizeInputOptions#shimMissingExports` should be `true`.

There's no need test the output in ndoejs with enabling `shimMissingExports`.

---

However, this only applies for some cases. For options that accepts js callback, we have to test them in nodejs.

## Guide

With the above example, if you think some options is not needed to be tests in nodejs, comment it in this issue. If it's true, let's remove them.

shulaoda:
### Description

closes #3255


crates/rolldown_binding/src/options/plugin/js_plugin.rs
```diff
@@ -135,16 +164,30 @@ impl Plugin for JsPlugin {
...
-      Ok(None)
+    let Some(cb) = &self.load else { return Ok(None) };
+
+    if let Some(load_filter) = &self.load_filter {
+      let stabilized_path = Path::new(args.id).relative(ctx.cwd());
+      let normalized_id = stabilized_path.to_string_lossy();
+
+      let exclude =
+        load_filter.exclude.clone().map(bindingify_string_or_regex_array).transpose()?;
+      let include =
```
shulaoda, line 175: Improved in #3266

crates/rolldown_binding/src/options/plugin/js_plugin.rs
```diff
@@ -472,34 +523,65 @@ impl Plugin for JsPlugin {
...
+) -> anyhow::Result<bool> {
+  let Some(transform_hook_filter_options) = transform_filter else {
+    return Ok(true);
+  };
+
+  let mut fallback_ret =
+    if let Some(ref module_type_filter) = transform_hook_filter_options.module_type {
+      if module_type_filter
+        .iter()
+        .map(ModuleType::from_str_with_fallback)
```
shulaoda, line 549: Improved in #3266

hyf0 (approved):
Well done!

PR #3280, opened 2025-01-01
chore(rolldown_binding): remove unnecessary `type_aliases.rs`

shulaoda:
### Description

Managing `type_aliases.rs` in the project root is less ideal than moving it to `types/mod.rs`, making the structure clearer and more organized.

I'm not sure if it is a good idea and it's just my personal suggestion. Feel free to close it if it's not suitable.

### DIFF
--- a/crates/rolldown_binding/src/lib.rs
+++ b/crates/rolldown_binding/src/lib.rs
@@ -23,6 +23,5 @@ pub mod parallel_js_plugin_registry;
 pub mod types;
 pub mod utils;
 pub use oxc_transform_napi;
-mod type_aliases;
 pub mod watcher;
 pub mod worker_manager;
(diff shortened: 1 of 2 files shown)


crates/rolldown/src/type_alias.rs
shulaoda, line 1: Related to #2407

They are no longer in use, so I think we should remove them.

hyf0:
It is intended to put `type_aliases.rs`  in the root level. I expect there will only one `type_aliases.rs` in each crate, while a crate might have multiple `types` folder. And `types` is for storing data structures. A `type_aliases.rs` in `types` folder means there's a struct or trait called `TypeAliases`.

shulaoda:
> It is intended to put `type_aliases.rs` in the root level. I expect there will only one `type_aliases.rs` in each crate, while a crate might have multiple `types` folder. And `types` is for storing data structures. A `type_aliases.rs` in `types` folder means there's a struct or trait called `TypeAliases`.

Got it.

hyf0 (approved):

Issue #3281, opened 2025-01-01
[Bug]: Watch flag ignores directory flag


trymeouteh:
### Reproduction link or steps

- Install Rolldown
- Create a JS project
- Have the packaging file named `bundle.js`
- Run one of the following commands

```
rolldown bundle.js -d dist -o my-package.js -f umd -n myPackage -w
rolldown bundle.js -d dist -o my-package.js -f umd -n myPackage -w -s
rolldown bundle.js -d dist -o my-package.js -f umd -n myPackage -w -s inline
rolldown bundle.js -d dist -o my-package.js -f umd -n myPackage -m -w
rolldown bundle.js -d dist -o my-package.js -f umd -n myPackage -m -w -s inline
```

### What is expected?

- The command will create the bundle in the `/dist` folder

### What is actually happening?

- The command will create the bundle in the root folder.

### System Info

```Shell
System:
    OS: Linux 6.8 Linux Mint 22 (Wilma)
    CPU: (12) x64 AMD Ryzen 5 PRO 6650U with Radeon Graphics
    Memory: 20.84 GB / 30.12 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.12.0 - /run/user/1000/fnm_multishells/30985_1735769087210/bin/node
    npm: 10.9.0 - /run/user/1000/fnm_multishells/30985_1735769087210/bin/npm
  Browsers:
    Brave Browser: 131.1.73.104
```

### Any additional comments?

Using rolldown v1.0.0-beta.1

trymeouteh:
It will appear that without the watch flag, that it will ignore the directory flag? Is the directory flag been deprecated?

underfin:
Only one of the `output.dir`(`-d`) and `output.file`(`-o`) is valid, the other is ignored. Using `rolldown bundle.js —o dist/my-package.js —f umd —n myPackage —w` could resolve this.

trymeouteh:
That make sense. I welcome this change since it works the same as rollup.

PR #3282, opened 2025-01-02
feat: give error when using both the file and the dir option

underfin:
### Description
related https://github.com/rolldown/rolldown/issues/3281


packages/rolldown/src/utils/bindingify-output-options.ts
```diff
@@ -28,6 +29,11 @@ export function bindingifyOutputOptions(
     globals,
     file,
   } = outputOptions
+
+  if (typeof dir === 'string' && typeof file === 'string') {
```
hyf0, line 33: Could you move this validation to rust side?

underfin: It's difficult at now, here need to refactor the `NormalizedBundlerOptions#dir` to `Option<String>` at first. Would you interesting refactor it?

hyf0: Yes.

hyf0 (approved):

Issue #3240, opened 2024-12-26
[Bug]: Paths in `writeBundle` are not consistent with Rollup


filipsobol:
### Reproduction link or steps

https://github.com/filipsobol/rolldown-inconsistent-paths-repro

1. Install dependencies.
2. Run `npm run build`.
3. Check logs in console.

### What is expected?

Rolldown should output the same paths in `writeBundle` hook as Rollup.

### What is actually happening?

Rolldown outputs paths that are different from Rollup.

```js
{
  bundler: 'Rollup',
  dir: undefined,
  file: 'dist/index.js',
  bundles: [ 'index.js' ]
}
```

```js
{
  bundler: 'Rolldown',
  dir: 'dist',
  file: 'dist/index.js',
  bundles: [ 'dist/index.js' ]
}
```

This makes the following code return different results (double `dist` in the path in Rolldown):

```js
writeBundle( { dir, file }, bundle ) {
  const outputDir = resolve( process.cwd(), dir ?? dirname( file ) );
  const assets = Object.keys( bundle ).map( name => join( outputDir, name ) );

  console.log( { assets } );
}
```

### System Info

```Shell
-
```

PR #3284, opened 2025-01-02
refactor: validate `file` and `dir` option in rust side

### LINKED ISSUE #3283: validate `file` and `dir ` option in rust side
It's difficult at now, here need to refactor the `NormalizedBundlerOptions#dir` to `Option<String>` at first. Would you interesting refactor it?

_Originally posted by @underfin in https://github.com/rolldown/rolldown/pull/3282#discussion_r1900571864_

shulaoda:
### Description

closes #3283


crates/rolldown/src/utils/normalize_options.rs
```diff
@@ -12,8 +12,37 @@ pub struct NormalizeOptionsReturn {
   pub warnings: Vec<BuildDiagnostic>,
 }
 
+fn verify_raw_options(raw_options: &crate::BundlerOptions) -> Vec<BuildDiagnostic> {
+  let mut warnings: Vec<BuildDiagnostic> = Vec::new();
+
+  if raw_options.dir.is_some() && raw_options.file.is_some() {
+    warnings.push(
```
shulaoda, line 19: I prefer to set it as a warning since we have already indicated that `dir` is an invalid option, so it can be safely ignored. This shouldn't block the entire bundler process. @underfin cc
https://github.com/rolldown/rolldown/blob/82b4e3062abaa7e4acf3ee1eea0553c7fd8ef13a/crates/rolldown/src/bundler.rs#L102-L106

shulaoda: I will defer to your decisions, this is merely my personal suggestion.

crates/rolldown/src/utils/normalize_options.rs
```diff
@@ -12,8 +12,37 @@ pub struct NormalizeOptionsReturn {
...
+      BuildDiagnostic::invalid_option(InvalidOptionType::InvalidOutputDirOption)
```
underfin, line 20: It should be an error for it.

packages/rolldown/tests/error/error.test.ts
```diff
@@ -1,139 +0,0 @@
-import { expect, test, vi, describe } from 'vitest'
```
underfin: The test need to preserve.

shulaoda: Should it be migrated to the rust side?

crates/rolldown/src/utils/normalize_options.rs
underfin: If supports both `dir` and `file`, the logic of the generated filename needs to be compatible with it, it will increase unnecessary complex.

underfin: Thank you for pointing me to the right way to validate it, i consider it need to support `NormalizedBundlerOptions::dir: Option<String>`.

packages/rolldown/tests/error/error.test.ts
underfin: No. Here has some tests for js plugin hook error. You could move the `should error with output.dir + output.file` test to rust side.

crates/rolldown/src/utils/normalize_options.rs
shulaoda: I understand your thoughts now, and I will try to find a way to better throw the option error.

crates/rolldown/src/utils/normalize_options.rs
shulaoda: Shouldn't we change it to an error?

crates/rolldown/src/utils/normalize_options.rs
underfin: For `NormalizedBundlerOptions::dir: Option<String>`, here has a other reason is https://github.com/rolldown/rolldown/issues/3240. The `normalize_options` always gives a default value for `dir` caused the js side `NormalizedOption#dir` is error when used `file` option. cc @hyf0

crates/rolldown/src/utils/normalize_options.rs
underfin: It shoulde a error. I resolve it because it is discussed at https://github.com/rolldown/rolldown/pull/3284#discussion_r1900787713

crates/rolldown/src/utils/normalize_options.rs
shulaoda: Could we keep this as a warning? An error might lead to extensive changes, which wouldn’t be the better way to validate it. After supporting `dir: Option<String>`, we can issue a warning indicating that both `dir` and `file` can't coexist. Then, we can default to using `file` and set `dir` to `None`.

shulaoda: In cases of multiple chunks, an error is thrown if `file` is not empty. Therefore, defaulting `dir` to `dist` seems reasonable.

https://github.com/rolldown/rolldown/blob/82b4e3062abaa7e4acf3ee1eea0553c7fd8ef13a/crates/rolldown/src/stages/generate_stage/mod.rs#L69-L71

underfin: I think it's ok, what do you think cc @hyf0

shulaoda: If the issue is solely related to the `dir` property in the js side `NormalizedOption`, we can handle it in its getter by checking if `file` is not empty. If `file` is not empty, the `getter` can return `undefined`.

shulaoda: > In cases of multiple chunks, an error is thrown if `file` is not empty. Therefore, defaulting dir to dist seems reasonable.

If the issue is solely related to the `dir` property in the js side `NormalizedOption`, we can handle it in its getter by checking if `file` is not empty. If `file` is not empty, the `getter` can return `undefined`.

shulaoda: > In cases of multiple chunks, an error is thrown if `file` is not empty. Therefore, defaulting dir to dist seems reasonable.

If the issue is solely related to the `dir` property in the js side `NormalizedOption`, we can handle it in its getter by checking if `file` is not empty. If `file` is not empty, the getter can return `undefined`.

shulaoda: Actually, if neither `dir` nor `file` is set, `rollup` will not execute the `writeBundle` hook. This gives us yet another reason to support `dir: Option<String>`.

hyf0: I wonder if `verify_raw_options` is enough so that we don't need to refactor code to `dir: String` anymore.

Though there's default value for `dir` if users only pass `file`, `dir` would be ignored if `file` is set.

Changing `dir` to `Option<String>` doesn't really affect the core logic, it only help that if we used `dir` while `file` exist, there will be a runtime panic(unwrap).

So, whether to change `dir` or not are both fine to me.

> Could we keep this as a warning?

I actually consider `warning` would be better. Just said that `dir` would be ignored, if `file` is set.

crates/rolldown/tests/rolldown/warnings/invalid_option/invalid_output_dir_option/artifacts.snap
```diff
@@ -0,0 +1,23 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+snapshot_kind: text
+---
+# warnings
+
+## INVALID_OPTION
+
+```text
+[INVALID_OPTION] Warning: Invalid value for option "output.dir" - you must set either "output.file" for a single-file build or "output.dir" when generating multiple chunks.
```
hyf0, line 10: If this gonna be a warning, I think we should add and underline the part "`output.dir` would be ignored."

hyf0:
```suggestion
[INVALID_OPTION] Warning: Invalid value for option "output.dir" - "output.dir" is ignored. you must set either "output.file" for a single-file build or "output.dir" when generating multiple chunks.
```

hyf0 (approved):

PR #3291, opened 2025-01-03
chore(infra): migrate the output schemas from `zod` to `valibot`

shulaoda:
### Description

Related to #3099

### DIFF
--- a/packages/rolldown/src/utils/validator.ts
+++ b/packages/rolldown/src/utils/validator.ts
@@ -313,11 +313,14 @@ const addonFunctionSchema = v.pipe(
   ),
 )
 
-const chunkFileNamesFunctionSchema = v.pipe(
-  v.function(),
-  v.args(v.tuple([v.custom<PreRenderedChunk>(() => true)])),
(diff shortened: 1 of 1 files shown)


packages/rolldown/src/utils/validator.ts
```diff
@@ -268,6 +276,251 @@ const inputCliOptionsSchema = v.omit(
   ],
 )
 
+/// --- OutputSchema ---
+
+enum ESTarget {
```
shulaoda, line 280: This is only temporary. I'm not sure what the differences are between `enum` and `union` with `literal` in `valibot` after `toJsonSchema`, so I will make the necessary adjustments later.

hyf0: Could you add a `TODO` mark?

shulaoda: Don't worry, I will soon adapt it to cli and handle it.

hyf0 (approved):

hyf0 (approved):

PR #3295, opened 2025-01-05
refactor(cli): migrate the cli parsing logic from `zod` to `valibot`

shulaoda:
### Description

Related to #3099 

For easier review, all `zod` related content will be removed after this PR is merged.


packages/rolldown/src/utils/validator.ts
```diff
@@ -439,7 +441,6 @@ const outputOptionsSchema = v.strictObject({
     v.description('Control comments in the output'),
   ),
   target: v.pipe(
-    // TODO: Verify if it needs to be changed to a union literal
```
shulaoda, line 442: There is no difference between the two in terms of parsing logic.

hyf0 (approved):

PR #3300, opened 2025-01-06
refactor(infra): change schema name from camelCase to PascalCase

### LINKED ISSUE #3099: [infra] Migrate to `valibot` from `zod`

shulaoda:
### Description

closes #3099 

The related task has been completed, and I have aligned the schema naming convention with `valibot`.

### DIFF
--- a/packages/rolldown/src/utils/validator.ts
+++ b/packages/rolldown/src/utils/validator.ts
@@ -11,7 +11,7 @@ import type {
   SourcemapPathTransformOption,
 } from '../types/misc'
 
-const stringOrRegExpSchema = v.union([v.string(), v.instance(RegExp)])
+const StringOrRegExpSchema = v.union([v.string(), v.instance(RegExp)])
 
 const LogLevelSchema = v.union([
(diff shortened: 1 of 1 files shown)


hyf0 (approved):

hyf0:
Is there any way we could not rely the json-schema in rolldown?

shulaoda:
> Is there any way we could not rely the json-schema in rolldown?

It is primarily used for customizing `parseArgs` options in the CLI, and we may remove it if we switch to another command-line parsing tool. Let me investigate if there is a better alternative.

PR #3305, opened 2025-01-07
fix: run just setup limited node version

unStone:
### Description

When I was building the project from scratch need NODE version >= 20.14.0;
but docs show me is 20.11.0
image

When I read the source code, I found running `just setup` also running `node ./scripts/misc/setup-prerequisites/node.js` and min nodeVersion is used `.node-version`(20.14.0)


underfin (approved):

underfin:
Thank you.

PR #3312, opened 2025-01-08
feat: eliminate unnecessary default property access

IWANABETHATGUY:
### Description

To ensure that the behavior remains consistent with the previous interop behavior after removing the `default` property access, we need to identify all cases where `import_default_alias.default == import_default_alias`. This can be inferred from the runtime.

```js
var __toESM = (mod, isNodeMode, target) => {
    return (
        __copyProps(
            isNodeMode || !mod || !mod.__esModule
                ? __defProp(target, "default", {
                        value: mod,
                        enumerable: true,
                    })
                : target,
            mod,
        )
    );
};
```

It is not difficult to see that the conditions are:
1. `isNodeMode` is `true || 1`
2. The module itself is an empty module
3. `mod.__esModule` is false

Condition one can be obtained from [this link](https://github.com/rolldown/rolldown/blob/eb47e2a4b5e07de61e6eb8101f4f07ecbf7c9c30/crates/rolldown_common/src/module/normal_module.rs?plain=1#L191-L201).
Condition two can be inferred through analysis; an empty module will default to being treated as `esm`, so the only case is `module.exports = undefined`.
Condition three can also be inferred through analysis, almost done in https://github.com/rolldown/rolldown/pull/3257.

However, it is important to note that when accessing the property `default` after `import_default_alias`, an unsafe situation may arise. For example:
```js
// index.js
import foo from './foo.js';
console.log(foo.default.bar === 123);
// foo.js
module.exports = { default: { bar: 123 } };
```

Here, condition 1 is not satisfied, condition 2 is not satisfied, but condition 3 is satisfied. Therefore, after `__toEsm`, the `mod` object will be transformed into:
```js
{
  default: { default: { bar: 123 } }
}
```
In this case, `foo.default.bar` is no longer equal to `foo.default.default.bar`. Specifically,
`foo.default.bar === undefined`, 
`foo.default.default.bar === 123`.

---

### DIFF
--- a/crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
+++ b/crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
@@ -35,6 +35,8 @@ impl<'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'_, 'ast> {
         // bailout if the importee is a external module
         // see rollup/test/function/samples/side-effects-only-default-exports/ as an
         // example
+        // TODO: maybe we could relex the restriction if `platform: node` and the external module
+        // is a node builtin module
         self.ctx.modules[importee_idx].as_normal()?;
         self.ctx.symbol_db.get(*symbol_ref).namespace_alias.as_ref().and_then(|alias| {
           if alias.property_name.as_str() == "default" {
--- a/crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
+++ b/crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
@@ -10,6 +10,7 @@ use oxc::{
 };
 use rolldown_common::{ExportsKind, Module, StmtInfoIdx, SymbolRef, ThisExprReplaceKind, WrapKind};
 use rolldown_ecmascript_utils::{ExpressionExt, TakeIn};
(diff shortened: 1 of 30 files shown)


IWANABETHATGUY:
Rest part I will left in another pr.

crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
```diff
@@ -302,6 +329,9 @@ impl<'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'_, 'ast> {
         }
       }
     } else {
+      if let Some(ref_id) = self.try_get_valid_namespace_alias_ref_id_from_member_expr(expr) {
```
hyf0, line 332: If this method could recognize the pattern here, why we don't just do the rewriting here?

It doesn't need to maintain `interested_namespace_alias_ref_id` and change the definition of `finalized_expr_for_symbol_ref`.

hyf0 (approved):

crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
IWANABETHATGUY: Then we have two position to rewrite `IdentifierReference`， It may reduce the maintainability

PR #3318, opened 2025-01-09
perf(link_stage): improve `LinkStage#sort_modules`

shulaoda:
### Description

- Remove unnecessary `executed_ids.insert(id);`
- Remove meaningless `executed_ids.shrink_to(self.module_table.modules.len())`
- Remove unnecessary `circular_dependencies.into_iter().collect::<Vec<_>>()` and `.copied()`

### DIFF
--- a/crates/rolldown/src/stages/link_stage/sort_modules.rs
+++ b/crates/rolldown/src/stages/link_stage/sort_modules.rs
@@ -2,6 +2,7 @@ use std::iter;
 
 use rolldown_common::{Module, ModuleIdx};
 use rolldown_error::BuildDiagnostic;
+use rolldown_utils::rustc_hash::FxHashSetExt;
 use rustc_hash::{FxHashMap, FxHashSet};
 
 use super::LinkStage;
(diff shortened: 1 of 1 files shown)


crates/rolldown/src/stages/link_stage/sort_modules.rs
```diff
@@ -88,7 +90,6 @@ impl LinkStage<'_> {
           }
         }
         Status::WaitForExit(id) => {
-          executed_ids.insert(id);
```
shulaoda, line 91: The initial state is always `ToBeExecuted`, and it is considered executed when it transitions to the `WaitForExit` state.

https://github.com/rolldown/rolldown/blob/e6a880a1c677039c7b91847bbe132bfaeb79e0a7/crates/rolldown/src/stages/link_stage/sort_modules.rs#L73-L74

crates/rolldown/src/stages/link_stage/sort_modules.rs
```diff
@@ -39,13 +40,14 @@ impl LinkStage<'_> {
       .chain(iter::once(Status::ToBeExecuted(self.runtime.id())))
       .collect::<Vec<_>>();
 
+    let mut next_exec_order = 0;
+
+    let mut executed_ids = FxHashSet::with_capacity(self.module_table.modules.len());
     let mut stack_indexes_of_executing_id = FxHashMap::default();
-    let mut executed_ids = FxHashSet::default();
-    executed_ids.shrink_to(self.module_table.modules.len());
```
shulaoda, line 44: This code is meaningless because `shrink_to` adjusts and releases excess capacity, but `executed_ids` has just been initialized.

crates/rolldown/src/stages/link_stage/sort_modules.rs
```diff
@@ -108,13 +109,12 @@ impl LinkStage<'_> {
     }
 
     if !circular_dependencies.is_empty() {
-      let cycles = circular_dependencies.into_iter().collect::<Vec<_>>();
-      for cycle in cycles {
+      for cycle in circular_dependencies {
```
shulaoda, line 112: Instead of collecting a new `Vec`, use it directly.

hyf0 (approved):
Thanks

PR #3319, opened 2025-01-09
feat(link_stage): check if importer is a js module

shulaoda:
### Description

https://github.com/rolldown/rolldown/blob/42ee4c997588b2ee205eceae36b48a18c95aed96/crates/rolldown/src/stages/link_stage/mod.rs#L147

### DIFF
--- a/crates/rolldown/src/stages/link_stage/mod.rs
+++ b/crates/rolldown/src/stages/link_stage/mod.rs
@@ -143,84 +143,87 @@ impl<'a> LinkStage<'a> {
   #[tracing::instrument(level = "debug", skip_all)]
   fn determine_module_exports_kind(&mut self) {
     let entry_ids_set = self.entries.iter().map(|e| e.id).collect::<FxHashSet<_>>();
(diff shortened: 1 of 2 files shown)


crates/rolldown/src/stages/link_stage/mod.rs
```diff
@@ -143,84 +143,87 @@ impl<'a> LinkStage<'a> {
...
-      importer.import_records.iter().for_each(|rec| {
-        let importee_id = rec.resolved_module;
-        let Module::Normal(importee) = &self.module_table.modules[importee_id] else {
-          return;
-        };
+    self
+      .module_table
+      .modules
+      .iter()
+      .filter_map(|module| module.as_normal().filter(|importer| importer.is_js_module()))
```
shulaoda, line 150: The key change is here and the rest is just due to formatting style.

hyf0:
You may ignore the

> TODO(hyf0): should check if importer is a js module

I haven't thought rough related logic, so I intend keep it this way for a while.

PR #3321, opened 2025-01-09
perf(link_stage): improve `LinkStage#wrap_modules`

shulaoda:
### Description

When `need_to_wrap` is `true`, `wrap_module_recursively` is executed, causing `module.import_records.iter()` to be processed repeatedly. We can optimize this to avoid redundant iterations.

https://github.com/rolldown/rolldown/blob/42ee4c997588b2ee205eceae36b48a18c95aed96/crates/rolldown/src/stages/link_stage/wrapping.rs#L89-L101

### DIFF
--- a/crates/rolldown/src/stages/link_stage/wrapping.rs
+++ b/crates/rolldown/src/stages/link_stage/wrapping.rs
@@ -15,15 +15,15 @@ struct Context<'a> {
 }
 
 fn wrap_module_recursively(ctx: &mut Context, target: ModuleIdx) {
-  let is_visited = &mut ctx.visited_modules[target];
-  if *is_visited {
-    return;
-  }
-  *is_visited = true;
-
   let Module::Normal(module) = &ctx.modules[target] else {
     return;
   };
+
+  // Only consider `NormalModule`
+  if ctx.visited_modules[target] {
+    return;
+  }
+  ctx.visited_modules[target] = true;
 
   if matches!(ctx.linking_infos[target].wrap_kind, WrapKind::None) {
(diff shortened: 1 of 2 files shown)


crates/rolldown/tests/esbuild/dce/tree_shaking_in_esm_wrapper/artifacts.snap
hyf0, line 1: unnecessary change

crates/rolldown/src/stages/link_stage/wrapping.rs
```diff
@@ -78,27 +78,11 @@ impl LinkStage<'_> {
...
 
     for module in self.module_table.modules.iter().filter_map(Module::as_normal) {
       let module_id = module.idx;
-      let linking_info = &self.metas[module_id];
-
-      let need_to_wrap = self.options.experimental.is_strict_execution_order_enabled()
-        || matches!(linking_info.wrap_kind, WrapKind::Cjs | WrapKind::Esm);
-
-      if need_to_wrap {
-        wrap_module_recursively(
```
hyf0, line 93: Interesting. This call got removed and everything still works as expected.

shulaoda: In fact, its side effects have been moved here.  `|| need_to_wrap`

https://github.com/rolldown/rolldown/blob/0d1d06b91feac3270c18a66a4e3d37047b2eca25/crates/rolldown/src/stages/link_stage/wrapping.rs#L118

hyf0 (approved):

PR #3323, opened 2025-01-10
feat(rolldown): support multiple hash placeholder

### LINKED ISSUE #2246: Filename supports duplicate hash like `[hash]-[hash:6].js`

alexzhang1030:
### Description

closes #2246

### DIFF
--- a/crates/rolldown_common/src/chunk/mod.rs
+++ b/crates/rolldown_common/src/chunk/mod.rs
@@ -13,7 +13,7 @@ use arcstr::ArcStr;
 use rolldown_rstr::Rstr;
 use rolldown_std_utils::PathExt;
 use rolldown_utils::{
-  extract_hash_pattern::extract_hash_pattern, hash_placeholder::HashPlaceholderGenerator,
+  extract_hash_pattern::extract_hash_patterns, hash_placeholder::HashPlaceholderGenerator,
   indexmap::FxIndexMap, BitSet,
 };
 use rustc_hash::FxHashMap;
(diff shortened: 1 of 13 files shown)


shulaoda:
Could you add a test in [packages/rolldown/tests/fixtures/output/fileNames/hash](https://github.com/rolldown/rolldown/pull/2238/files#diff-694c21ae584943abdf888d27ff97a5884679536862c25501b2f0b373ba496b2e)?

alexzhang1030:
I don't know why the hash has been updated since this PR, can you help me figure it out? /cc @shulaoda

crates/rolldown_common/src/chunk/mod.rs
```diff
@@ -147,7 +147,7 @@ impl Chunk {
 
     let rendered = filename_template.render(&FileNameRenderOptions {
       name: Some(&name),
-      hash: hash_placeholder.as_deref(),
+      hashes: hash_placeholder.as_deref(),
       ..Default::default()
     });
```
shulaoda, line 153: 1. Set `entry_filenames` as `[name]-[hash]-[hash:6].js`

2. Debug `println!("{rendered:?}, {hash_placeholder:?}");` at [here](https://github.com/rolldown/rolldown/pull/3323/files#diff-ff987458a29cbb840e078962ecd1861b881139604fe178056d8b30f9c61dc823R153)

```bash
"entry-!~{000}~-!~{001}~.js", Some(["!~{000}~", "!~{001}~", "!~{2}~"])
```

This indicates that the hashes in the name do not correspond one-to-one.

PR #3327, opened 2025-01-10
feat: support PluginContext emitChunk

underfin:
### Description

The `PluginContext#emitChunk` is used to add an extra entry point at rollup. Its special feature is that you can specify the filename of the entry chunk. It generates a specified entry name at the module-federation remote. Here is an example. We need to generate a chunk with `filename: 'remoteEntry.js'`.

``` .js
{
    filename: 'remoteEntry.js',
    exposes: { './App': './src/App', }
}
```

It is a children method of `emitFile` with `type: 'chunk'` at rollup, ref https://rollupjs.org/plugin-development/#this-emitfile. But here is the difficult to make the method to sync. So here add an async `emitChunk` method at the pr, i will try to make it to sync at the future.

The rollup generate specified chunk name implementation is add the `fileName` to `Moudle#chunkFileNames` and use it generate chunk at https://github.com/rollup/rollup/blob/master/src/Chunk.ts#L420


underfin:
[@makkarpov/nuxt-service-worker](https://github.com/makkarpov/nuxt-service-worker/blob/ec2f7a4f545deed4f9039d78278e4e0daccd7377/src/build.ts#L89)
[rollup-plugin-workers](https://github.com/johanholmerin/rollup-plugin-workers/blob/19864bbd9725f39fab60dbd091720d45b7a22b41/index.js#L102)
astro
- https://github.com/withastro/astro/blob/1c36331782b63044aed167a092e4b913d2f1a656/packages/astro/src/vite-plugin-scripts/index.ts#L56
- https://github.com/withastro/astro/blob/1c36331782b63044aed167a092e4b913d2f1a656/packages/astro/src/core/server-islands/vite-plugin-server-islands.ts#L69
The astro generate entry chunk with dynamic serve components, so we need to support it.

hyf0 (approved):

PR #3333, opened 2025-01-10
refactor(rust): simplify `determine_use_strict`

shulaoda:
### Description

We can simplify it.

### DIFF
--- a/crates/rolldown/src/utils/chunk/determine_use_strict.rs
+++ b/crates/rolldown/src/utils/chunk/determine_use_strict.rs
@@ -1,17 +1,8 @@
-use rolldown_common::ExportsKind;
-
 use crate::types::generator::GenerateContext;
 
 pub fn determine_use_strict(ctx: &GenerateContext) -> bool {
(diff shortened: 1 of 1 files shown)


crates/rolldown/src/utils/chunk/determine_use_strict.rs
```diff
@@ -1,17 +1,8 @@
...
-
-  if modules.peek().is_none() {
-    // No modules, no need to add "use strict"
-    return false;
-  }
-
-  modules.all(|ecma_module| {
-    let is_esm = matches!(&ecma_module.exports_kind, ExportsKind::Esm);
-    is_esm || ctx.link_output.ast_table[ecma_module.ecma_ast_idx()].0.contains_use_strict
+  ctx.renderable_ecma_modules().all(|ecma_module| {
```
shulaoda, line 4: Sorry, I didn't realize that `all(..)` returns `false` when modules is empty. This behavior feels quite counterintuitive.

PR #3216, opened 2024-12-24
feat: merge esm namespace in cjs module

hyfdev:
### Description

- This feature is conflicted with treeshaking on unused cjs modules. So this PR would breaks the treeshaking on unused cjs modules.

PR #3343, opened 2025-01-12
feat: merge generated `__toESM` calls of cjs modules

hyf0:
### Description

Continue of https://github.com/rolldown/rolldown/pull/3216.

### DIFF
--- a/crates/rolldown/tests/esbuild/dce/package_json_side_effects_false_keep_bare_import_and_require_common_js/artifacts.snap
+++ b/crates/rolldown/tests/esbuild/dce/package_json_side_effects_false_keep_bare_import_and_require_common_js/artifacts.snap
@@ -14,6 +14,7 @@ var require_demo_pkg = __commonJS({ "node_modules/demo-pkg/index.js"(exports) {
 	exports.foo = 123;
 	console.log("hello");
 } });
+var import_demo_pkg = __toESM(require_demo_pkg());
 
 //#endregion
 //#region src/entry.js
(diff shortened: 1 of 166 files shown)


crates/rolldown/tests/esbuild/default/es6_from_common_js/artifacts.snap
```diff
@@ -23,11 +25,12 @@ var require_bar = __commonJS({ "bar.js"(exports) {
 		return "bar";
 	};
 } });
+var import_bar = __toESM(require_bar());
 
 //#endregion
 //#region entry.js
-var import_foo = __toESM(require_foo());
-var import_bar = __toESM(require_bar());
+require_foo();
```
yyx990803: These calls are no longer necessary and should be removed

yyx990803 (dismissed):

crates/rolldown/tests/esbuild/default/es6_from_common_js/artifacts.snap
hyf0: done

IWANABETHATGUY (approved):

crates/rolldown/tests/esbuild/dce/package_json_side_effects_false_keep_bare_import_and_require_common_js/artifacts.snap
```diff
@@ -14,6 +14,7 @@ var require_demo_pkg = __commonJS({ "node_modules/demo-pkg/index.js"(exports) {
 	exports.foo = 123;
 	console.log("hello");
 } });
+var import_demo_pkg = __toESM(require_demo_pkg());
```
IWANABETHATGUY, line 17: Here is the only case seems not expect. Maybe you could fix it in next pr.

hyf0: https://github.com/rolldown/rolldown/issues/3359

PR #3347, opened 2025-01-13
chore(build): colorize starting segment of every each line.

Cecil0o0:
### Description

image

choose green color for semantic color of `success/done`, and then colorize starting segment to identify a positive status for a line of instruction such as `Cleaning` or `Copying`.


IWANABETHATGUY (approved):
Thanks

PR #3351, opened 2025-01-14
refactor: make PluginContext#emitChunk to be sync

underfin:
### Description

Make the `PluginContext#emitChunk` to be sync, make it compatible with rollup.

### DIFF
--- a/crates/rolldown_binding/src/options/plugin/binding_plugin_context.rs
+++ b/crates/rolldown_binding/src/options/plugin/binding_plugin_context.rs
@@ -79,8 +79,10 @@ impl BindingPluginContext {
   }
 
   #[napi]
-  pub async fn emit_chunk(&self, file: BindingEmittedChunk) -> anyhow::Result<String> {
-    self.inner.emit_chunk(file.into()).await.map(|id| id.to_string())
+  pub fn emit_chunk(&self, file: BindingEmittedChunk) -> anyhow::Result<String> {
+    futures::executor::block_on(async {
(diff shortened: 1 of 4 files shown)


hyf0:
Would you mind to check if it works in stackblitz?

underfin:
> Would you mind to check if it works in stackblitz?

It is failed, it caused the module tx is not setting, i track it at https://github.com/rolldown/rolldown/issues/3354

crates/rolldown_binding/src/options/plugin/binding_plugin_context.rs
```diff
@@ -79,8 +79,8 @@ impl BindingPluginContext {
   }
 
   #[napi]
-  pub async fn emit_chunk(&self, file: BindingEmittedChunk) -> anyhow::Result<String> {
-    self.inner.emit_chunk(file.into()).await.map(|id| id.to_string())
+  pub fn emit_chunk(&self, file: BindingEmittedChunk) -> String {
+    self.inner.emit_chunk(file.into()).to_string()
```
hyf0: How about we use `block_on` on this method and keep the original one async?

hyf0 (approved):

crates/rolldown_binding/src/options/plugin/binding_plugin_context.rs
underfin: Good idea. It is a more clear way and could be reduce the overhead for rust plugins.

PR #3352, opened 2025-01-14
feat(node): expose moduleFederationPlugin

underfin:


packages/rolldown/src/builtin-plugin/constructors.ts
```diff
@@ -77,3 +78,9 @@ export function viteResolvePlugin(
   })
   return makeBuiltinPluginCallable(builtinPlugin)
 }
+
+export function moduleFederationPlugin(
+  config: BindingModuleFederationPluginOption,
+): BuiltinPlugin {
+  return new BuiltinPlugin('builtin:load-fallback', config)
```
hyf0: Is `builtin:load-fallback` right?

underfin: Fixed it.

hyf0 (approved):

PR #3360, opened 2025-01-15
feat: export `RolldownPluginOption` type

alpheusmtx:
### Description

Export `RolldownPluginOption` type in case it will be used.

```ts

export default defineConfig({
    // plugins?: RolldownPluginOption;
    plugins: [],
});
```


IWANABETHATGUY (commented):
Thanks

IWANABETHATGUY (approved):

PR #3378, opened 2025-01-18
chore(rolldown): oxc 0.47.1

Boshen:


crates/rolldown/tests/esbuild/ts/ts_enum_cross_module_inlining_definitions/enums.ts
```diff
@@ -1,6 +1,7 @@
+var foo;
 export enum a {
 	implicit_number,
 	explicit_number = 123,
 	explicit_string = 'xyz',
 	non_constant = foo,
```
Boshen: This is ts syntax error 

```
export enum a {
	implicit_number,
	explicit_number = 123,
	explicit_string = 'xyz',
	non_constant = foo,
}
```

```
Cannot find name 'foo'.
```

See https://github.com/oxc-project/oxc/pull/8543

hyf0: esbuild source: https://github.com/evanw/esbuild/blob/df815ac27b84f8b34374c9182a93c94718f8a630/internal/bundler_tests/bundler_ts_test.go#L2246-L2266

esbuild output: https://github.com/evanw/esbuild/blob/df815ac27b84f8b34374c9182a93c94718f8a630/internal/bundler_tests/snapshots/snapshots_ts.txt#L334-L348


Without adding `var foo;`, is the error a runtime error or compile-error emitted by oxc?

If it's a runtime error, you could set the `_config.json#expectExecuted: 'false'`and not add `var foo;`. We want to keep sync with the original esbuild's test input.

 If it's a compile-error, maybe we should investigate why esbuild doesn't treat this as a compile-error. In the meantime, we could set `_config.json#expectError: true` and still not add `var foo` to make the test passed.

Boshen: tsc errors with `Cannot find name 'foo'.`

Runtime errors with `Uncaught ReferenceError: foo is not defined`.

hyf0: If so, I think you could  set the `_config.json#expectExecuted: 'false'` and not add `var foo;`. It should also make the test passed.

hyf0 (approved):

PR #3379, opened 2025-01-18
feat(string_wizard): add `.hasChanged()`

Boshen:
API: https://github.com/rich-harris/magic-string?tab=readme-ov-file#shaschanged impl: https://github.com/Rich-Harris/magic-string/pull/202 part of https://github.com/nuxt/nuxt/pull/30066#issuecomment-2532929079

### DIFF
--- a/crates/string_wizard/Cargo.toml
+++ b/crates/string_wizard/Cargo.toml
@@ -17,7 +17,7 @@ serde         = { workspace = true, optional = true }
 insta = { workspace = true }
 
 [features]
-default   = ["sourcemap"]
+default   = ["sourcemap", "serde"]
 serde     = ["dep:serde", "oxc_index/serde"]
 sourcemap = ["dep:oxc_sourcemap"]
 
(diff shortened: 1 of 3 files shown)


crates/string_wizard/Cargo.toml
```diff
@@ -17,7 +17,7 @@ serde         = { workspace = true, optional = true }
 insta = { workspace = true }
 
 [features]
-default   = ["sourcemap"]
+default   = ["sourcemap", "serde"]
```
Boshen, line 20: `cargo check -p string_wizard` fails to compile if this is not the default.

crates/string_wizard/src/magic_string/mod.rs
```diff
@@ -83,6 +83,11 @@ impl<'text> MagicString<'text> {
     self.len() == 0
   }
 
+  /// Indicates if the string has been changed.
+  pub fn has_changed(&self) -> bool {
+    self.source != self.to_string()
```
hyf0: There's a fast path `self.source.len() != self.len() || self.source != self.to_string()`.

hyf0 (approved):

Boshen:
@hyf0 can you publish a new version 🙇

hyf0:
@Boshen Done 0.0.26.

PR #3384, opened 2025-01-20
fix: sort function panic due to user-provided comparison function does not correctly implement a total order 

IWANABETHATGUY:
### Description
1. fix https://github.com/vitejs/vite-ecosystem-ci/blob/rolldown-vite/README-temp.md#analog, user-provided comparison function does not correctly implement a total order


IWANABETHATGUY:
I didn't find the situation when the total order is not satisfied, then I tried to replace `sort_unstable` with `sort` and it worked without panicing, try to use gpt's method but it did not work.

hyf0 (approved):

hyf0:
Let's wait for #3387.

PR #3405, opened 2025-01-23
fix(splitting): indirect external symbol

underfin:
### Description

fix nuxt + rolldown crashes at crates/rolldown_common/src/types/symbol_ref_db.rs, ref https://github.com/vitejs/vite-ecosystem-ci/blob/rolldown-vite/README-temp.md#nuxt.

The issue happened because we merge all external symbol to one, but not consider the exported external symbol is could split, the referenced external symbol is used other chunks exported symbols, the symbol is not deconflicted at importer chunk.

The pr intend to avoid connecting cross module symbols from external modules, it will ensure correctness for calculate cross  module symbols. Here is a [esbuild repl](https://hyrious.me/esbuild-repl/?version=0.24.0&b=e%00entry.js%00import+%7B+read+%7D+from+%27.%2Findirect%27%0A%0Aconsole.log%28read%29%0A%0Aimport%28%27.%2Findirect%27%29&b=%00indirect.js%00export+%7B+read+%7D+from+%22node%3Aassert%22%3B&o=%7B%0A++bundle%3A+true%2C%0A++outdir%3A+%27dist%27%2C%0A++external%3A+%5B%27node%3Aassert%27%5D%2C%0A++format%3A+%27esm%27%2C%0A++keepNames%3A+true%2C%0A++splitting%3A+true%0A%7D). It meaning we only merge import symbols from external modules, if it is exported, it can't be merged.

<br class="Apple-interchange-newline">

### DIFF
--- a/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
+++ b/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
@@ -554,7 +554,14 @@ impl BindImportsAndExportsContext<'_> {
 
       let rec = &module.import_records[named_import.record_id];
       let is_external = matches!(self.index_modules[rec.resolved_module], Module::External(_));
-      if is_esm && is_external {
+
+      if is_esm
+        && is_external
+        && self.metas[module_id]
+          .resolved_exports
+          .iter()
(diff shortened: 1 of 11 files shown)


IWANABETHATGUY:
I will review it later, busy with other work.

IWANABETHATGUY (approved):

PR #3476, opened 2025-02-02
refactor(rust): inline unnecessary `extract_rest_member_expr_chain`

shulaoda:
### Description

- Inline `extract_rest_member_expr_chain` to avoid redundant calls.

### DIFF
--- a/crates/rolldown/src/ast_scanner/side_effect_detector/utils.rs
+++ b/crates/rolldown/src/ast_scanner/side_effect_detector/utils.rs
@@ -281,34 +281,28 @@ pub fn extract_member_expr_chain<'a>(
   if max_len == 0 {
     return None;
   }
+
   let mut chain = vec![];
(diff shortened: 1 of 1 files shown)


crates/rolldown/src/ast_scanner/side_effect_detector/utils.rs
```diff
@@ -72,26 +72,6 @@ pub(crate) enum PrimitiveType {
   Unknown,
 }
 
-fn merged_known_primitive_types(
```
IWANABETHATGUY: This method is copied from esbuild, and maybe used in the future, so I don't think it is necessarily to inline them manually

shulaoda: Got it.

IWANABETHATGUY (approved):

PR #3489, opened 2025-02-04
refactor(rust): simplify `scan_ast` and `make_ecma_ast`

shulaoda:
### Description

Concentrate the required variables from `AstScanner` into the scan result.


IWANABETHATGUY:
LGTM, what I am thinking about next is, we could flat `result` into `AstScanner`, and then extract `AstScannerResult` after transformation .

IWANABETHATGUY (approved):

PR #3503, opened 2025-02-06
feat: warn when a file is emitted with the same name

### LINKED ISSUE #3444: [Feature Request]: warn when a file is emitted with the same name

https://stackblitz.com/edit/github-nci4o6nk-euzlsxjf?file=rolldown.config.mjs
No error happens with rolldown. An error is output with rollup.

A warning is output like rollup does.

underfin:
### Description
close https://github.com/rolldown/rolldown/issues/3444


examples/basic-vue/index.js
IWANABETHATGUY, line 1: should revert this.

crates/rolldown_common/src/file_emitter.rs
```diff
@@ -207,14 +210,26 @@ impl FileEmitter {
...
     self.files.iter_mut().for_each(|mut file| {
       let (key, value) = file.pair_mut();
       if self.emitted_files.contains(key) {
         return;
       }
       self.emitted_files.insert(key.clone());
 
+      // Follow rollup using lowercase filename to check conflicts
+      let lowercase_filename = value.filename.as_str().to_lowercase().into();
+      if self.emitted_filenames.contains(&lowercase_filename) {
```
IWANABETHATGUY, line 227: Use `entry` so that you don't need to hash twice

underfin: Here the `emitted_filenames` is `FxDashSet`, it hasn't `entry` method. Here using `FxDashSet` because the `FileEmitter` wrapped with `Arc`.

IWANABETHATGUY (approved):

PR #3505, opened 2025-02-06
feat: export more types

sapphi-red:
### Description

Exported types that are used by Vite.


packages/rolldown/src/log/logging.ts
```diff
@@ -6,6 +6,11 @@ export type LogLevelWithError = LogLevel | 'error'
 export type RollupLog = any
```
sapphi-red, line 6: Some types in this file are duplicated. The same types exist in `src/types/misc`. (I didn't fix that though)

IWANABETHATGUY (approved):

PR #3509, opened 2025-02-06
fix(node): tsc build only emit declaration

### LINKED ISSUE #3506: [Bug]: `dist/types` in the packed package includes JS files alongside the dts files

see https://www.npmjs.com/package/rolldown?activeTab=code and https://www.npmjs.com/package/rolldown/v/1.0.0-beta.3-commit.306867e?activeTab=code

### What is expected?

JS files are not included in `types` directory as I guess it's not used for anything.
The JS files also exist in `cjs`, `esm`, `shared` directory and I guess those are the used ones.

JS files are not included in `types` directory, making the package size bigger unnecessarily.

### System Info

```Shell
N/A
```

### Any additional comments?

For both beta releases and canary releases.

underfin:
### Description
close https://github.com/rolldown/rolldown/issues/3506, ref https://www.typescriptlang.org/tsconfig/#emitDeclarationOnly.

### DIFF
--- a/packages/rolldown/tsconfig.json
+++ b/packages/rolldown/tsconfig.json
@@ -8,6 +8,7 @@
     "outDir": "./dist/types",
     "resolveJsonModule": true,
     "noEmit": false,
-    "tsBuildInfoFile": ".tsbuildinfo"
+    "tsBuildInfoFile": ".tsbuildinfo",
+    "emitDeclarationOnly": true
   }
 }


IWANABETHATGUY:
after this pr, there are only two `.d.ts` in under `dist/types`: 

image

hyf0 (approved):

PR #3555, opened 2025-02-11
feat(rolldown): support multiple hash placeholder

### LINKED ISSUE #2246: Filename supports duplicate hash like `[hash]-[hash:6].js`

shulaoda:
### Description

closes #2246, #3323 

The issue lies in the implementation of `extract_hash_patterns`, where using `hash_pattern.pattern.len()` as the starting position for `remaining` is incorrect. @alexzhang1030 cc

Next, I will attempt to optimize the `FilenameTemplate#render`.


crates/rolldown_common/src/inner_bundler_options/types/filename_template.rs
```diff
@@ -32,10 +32,12 @@ impl FilenameTemplate {
...
-      if let Some(start) = tmp.find("[hash") {
-        if let Some(end) = tmp[start + 5..].find(']') {
-          tmp.replace_range(start..=start + end + 5, hash);
+    if let Some(hashes) = &options.hashes {
+      for hash in *hashes {
+        if let Some(start) = tmp.find("[hash") {
+          if let Some(end) = tmp[start + 5..].find(']') {
+            tmp.replace_range(start..=start + end + 5, hash.as_str());
+          }
         }
```
IWANABETHATGUY, line 39:
```rs
    options.hashes.into_iter().flatten().for_each(|hash| {
      if let Some(start) = tmp.find("[hash") {
        if let Some(end) = tmp[start + 5..].find(']') {
          tmp.replace_range(start..=start + end + 5, hash.as_str());
        }
      }
    });

```

IWANABETHATGUY (approved):

hyf0:
Nice. Thank you and @alexzhang1030.

PR #3587, opened 2025-02-15
chore(rolldown): oxc v0.51.0

Boshen:


crates/rolldown/tests/rolldown/misc/require_deconflict/artifacts.snap
```diff
@@ -11,7 +11,7 @@ snapshot_kind: text
 
 //#region main.js
 const __require = "test";
-var main_default = () => __require$1("test-dep") || __require$1("test-dep666");
+var main_default = () => __require$1("test-dep") || __require$1(`test-dep${666}`);
```
Boshen, line 14: We can recover the behavior with a bit of more code. @sapphi-red

packages/rolldown/package.json
```diff
@@ -103,15 +103,15 @@
     "dtsHeader": "type MaybePromise<T> = T | Promise<T>\ntype Nullable<T> = T | null | undefined\ntype VoidNullable<T = void> = T | null | undefined | void\nexport type BindingStringOrRegex = string | RegExp\n\n"
   },
   "dependencies": {
-    "@oxc-project/types": "0.48.2",
+    "@oxc-project/types": "0.51.0",
```
Boshen, line 106: I'll add a `just` command for updating oxc next time, we need to keep all the versions in sync.

IWANABETHATGUY (approved):

crates/rolldown/tests/rolldown/misc/require_deconflict/artifacts.snap
IWANABETHATGUY: I am not sure what do you mean, but this is expected.

PR #3598, opened 2025-02-16
feat(rust): remove unused `import(..)` if importee doesn't have side-effects

### LINKED ISSUE #2827: [Feature Request]: remove unused `import(...)` if importee doesn't have side-effects

1. Open https://stackblitz.com/edit/github-xv4yss-lweea2?file=rolldown.config.mjs,src%2Fdynamic.js,src%2Fentry.js
3. `dist/dynamic-B5a49wDx.js` is an empty file and `dist/entry.js` imports it.

### What is expected?

The empty `dist/dynamic-B5a49wDx.js` is removed and the `import("./dynamic-B5a49wDx.js");` is replaced with `Promise.resolve()`. Even further, `Promise.resolve()` can be removed completely in this case because there's no side effect.

`dist/dynamic-B5a49wDx.js` is not removed and the `dist/entry.js` contains `import("./dynamic-B5a49wDx.js");`.

### System Info

```Shell
Stackblitz

  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    rolldown: https://pkg.pr.new/rolldown@01cedb9 => 0.14.0
```

### Any additional comments?

related to #2604, #2826

shulaoda:
### Description

closes #2827 

The core idea is that if a dynamic import module has no included statements and is not referenced, it‘s `module.is_included()` should be `false`.

~~I’m facing an issue with how to implement the polyfill.~~ There are two options: one is to declare `var init_foo = Object.freeze(..)` at the `importer` side, and the other is to introduce a global empty dynamic module (`runtime`). 

~~I'm constantly unable to correctly link the symbol reference for `init_foo` or `__dynamicEmpty`. I’m unsure how to proceed correctly, so I’m just sharing my thoughts and seeking help on this.~~


crates/rolldown/src/runtime/runtime-base.js
```diff
@@ -40,6 +40,10 @@ var __toESM = (mod, isNodeMode, target) => (
     mod)
 )
 var __toCommonJS = mod => __copyProps(__defProp({}, '__esModule', { value: true }), mod)
+var __dynamicEmptyModule = Object.freeze({
+	__proto__: null,
+  default: {}
+});
```
shulaoda: `__DynamicEmptyModule` may not be good because there may be the following situations:

```js
let a = await('./a.js')
let b = await('./b.js')

console.log(a == b)
```

shulaoda: We should align with `rollup` and let me continue to improve.

shulaoda: It is difficult to achieve functions similar to `rollup`.

```js
// entry.js
import('./abc.js')
import('./dynamic.js').then(() => console.log("1"))

// abc.js
import('./dynamic.js').then(() => console.log("2"))

// dynamic.js
```

The current implementation result is:

```js
// entry.js
import "./entry--jpnM2Ll.js";

// entry--jpnM2Ll.js
import("./abc-BIQEauab.js");
Promise.resolve().then(function() {
	return dynamic;
}).then(() => console.log("1"));
var dynamic = Object.freeze({
	__proto__: null,
	default: {}
});

export { dynamic };

// abc-BIQEauab.js
import { dynamic } from "./entry--jpnM2Ll.js";
Promise.resolve().then(function() {
	return dynamic;
}).then(() => console.log("2"));
```


Rollup:
```js
// entry.js
import('./abc-qNCBTG9J.js');

Promise.resolve().then(function () { return dynamic; }).then(() => console.log("1"));

var dynamic = /*#__PURE__*/Object.freeze({
	__proto__: null
});

export { dynamic as d };

// abc-qNCBTG9J.js
import('./main.js').then(function (n) { return n.d; });
```

PR #3613, opened 2025-02-18
fix(cli): cli default option

### LINKED ISSUE #3612: `inlineDyanmicImports: true` doesn't work as expected
> At this point, I would say the behavior of `advancedChunks` is expected. It's not its responsibility to inlining dynamic chunks. It should be done by `inlineDyanmicImports`.
> 
> But I test `inlineDyanmicImports: true, format: 'esm'` without using `advancedChunks` and it's not working. @underfin cc
> 
> `format: 'iffe'` manages to bundle all files into a single chunk. I guess there's a bug for using `inlineDyanmicImports: true` seperately. 

 _Originally posted by @hyf0 in [#3610](https://github.com/rolldown/rolldown/issues/3610#issuecomment-2665008157)_

Repro: https://stackblitz.com/edit/github-nci4o6nk-xkum7nvz

underfin:
### Description
close https://github.com/rolldown/rolldown/issues/3612

### DIFF
--- a/packages/rolldown/tests/cli/__snapshots__/cli-e2e.test.ts.snap
+++ b/packages/rolldown/tests/cli/__snapshots__/cli-e2e.test.ts.snap
@@ -81,6 +81,23 @@ NOTES
   * For more information, please visit https://rolldown.rs/."
 `;
 
+exports[`cli options for bundling > cli default options 1`] = `
+"input cli default options {
+  input: 'index.js',
+  plugins: [
+    {
+      name: 'test-input-plugin',
+      options: [Function: options],
+      outputOptions: [Function: outputOptions]
+    }
+  ]
+}
+output cli default options {}
+<DIR>/index.js  chunk │ size: 0.13 kB
+
+"
+`;
+
 exports[`cli options for bundling > should handle negative boolean options 1`] = `
 "(function(exports, node_fs) {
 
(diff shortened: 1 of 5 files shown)


IWANABETHATGUY:
Can you add a test case related to https://stackblitz.com/edit/github-nci4o6nk-xkum7nvz?file=src%2Findex.js ?

packages/rolldown/tests/cli/__snapshots__/cli-e2e.test.ts.snap
```diff
@@ -81,6 +81,23 @@ NOTES
...
+  plugins: [
+    {
+      name: 'test-input-plugin',
+      options: [Function: options],
+      outputOptions: [Function: outputOptions]
+    }
+  ]
+}
+output cli default options {}
+<DIR>/index.js  chunk │ size: 0.13 kB
```
underfin, line 96: cc @hyf0 please notice the default cli option snapshot changes if some pr changed it, it shouldn't be changed.

hyf0: Could you give a example PR? I don't get what's going on here.

underfin:
> Can you add a test case related to https://stackblitz.com/edit/github-nci4o6nk-xkum7nvz?file=src%2Findex.js ?

The root reason is cli has unless default value, here only add a cli default test could avoid the regression. If add the special option case, here need to many test cases.

hyf0:
Let me try to understand here: The problem of #3612 is caused by the cli's default options overwrites users' options. Is that right?

underfin:
> Let me try to understand here: The problem of #3612 is caused by the cli's default options overwrites users' options. Is that right?

Yeah.

hyf0 (approved):

PR #3614, opened 2025-02-18
fix: glob import from parent dir

jrmajor:
### Description

If `import.meta.glob` pattern starts with `../`, it's currently resolved relative to current dir instead of parent dir.


underfin (approved):

IWANABETHATGUY:
let me have a look result of vite

jrmajor:
@IWANABETHATGUY Here's a test in Vite that also has parent glob:
https://github.com/vitejs/vite/blob/021443c5a11eedfb176f2c65d5fab65e9c21adfb/packages/vite/src/node/__tests__/plugins/importGlob/fixture-a/index.ts#L107-L111
https://github.com/vitejs/vite/blob/021443c5a11eedfb176f2c65d5fab65e9c21adfb/packages/vite/src/node/__tests__/plugins/importGlob/__snapshots__/fixture.spec.ts.snap#L90

PR #3621, opened 2025-02-19
test: port glob import test from vite

jrmajor:
### Description

Not sure whether we want to port this test or create new ones. Feel free to reject this PR.

### DIFF
--- a/packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/_config.ts
+++ b/packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/_config.ts
@@ -1,0 +1,34 @@
+import { importGlobPlugin } from 'rolldown/experimental'
+import { RolldownOutput } from 'rolldown'
+import { defineTest } from 'rolldown-tests'
+import { expect } from 'vitest'
+import * as fs from 'node:fs'
+import * as path from 'node:path'
+
+// ported from https://github.com/vitejs/vite/tree/021443c5a11eedfb176f2c65d5fab65e9c21adfb/packages/vite/src/node/__tests__/plugins/importGlob
+export default defineTest({
+  config: {
+    input: './fixture-a/index.ts',
+    output: {
+      chunkFileNames: '[name].js',
+    },
+    plugins: [
+      importGlobPlugin({
+        root: path.resolve(import.meta.dirname),
+      }),
+      {
+        name: 'load-file-with-query',
+        load(id: string) {
+          if (!id.endsWith('?raw')) return
+          const res = fs.readFileSync(id.slice(0, -4), 'utf-8')
+          return `export default ${JSON.stringify(res)}`
+        },
+      },
+    ],
+  },
+  async afterTest(output: RolldownOutput) {
+    await expect(output.output[0].code).toMatchFileSnapshot(
+      path.resolve(import.meta.dirname, 'fixture-a/index.ts.snap'),
+    )
+  },
+})
--- a/packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/fixture-a/index.ts.snap
+++ b/packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/fixture-a/index.ts.snap
@@ -1,0 +1,74 @@
+import { a_exports, name as name$1 } from "./a2.js";
+import { b_exports, name$1 as name$2 } from "./b2.js";
+import { modules_exports, name$2 as name$3 } from "./modules2.js";
+
+//#region fixture-a/index.ts
+const basic = {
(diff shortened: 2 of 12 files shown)


packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/fixture-a/index.ts.snap
```diff
@@ -0,0 +1,74 @@
...
+	"./modules/a.ts": () => import("./a-DAJhtQJU.js").then((m) => m.default),
+	"./modules/b.ts": () => import("./b-BQKRc37x.js").then((m) => m.default),
+	"./modules/index.ts": () => import("./modules-Ietgo5aO.js").then((m) => m.default)
+};
+const namedDefaultWithObjectKeys = Object.keys({
+	"./modules/a.ts": () => import("./a-DAJhtQJU.js").then((m) => m.default),
+	"./modules/b.ts": () => import("./b-BQKRc37x.js").then((m) => m.default),
+	"./modules/index.ts": () => import("./modules-Ietgo5aO.js").then((m) => m.default)
+});
+const namedDefaultWithObjectValues = Object.values({
```
underfin, line 61: Could you port the snapshot assert to runtime assert, it should be more trustworthy, like https://github.com/rolldown/rolldown/blob/main/packages/rolldown/tests/fixtures/builtin-plugin/glob-import/basic/assert.mjs

jrmajor: In case of tests with `Object.values` and `Object.keys`, like this one:
```js
// vite output: Object.values([__vite_glob_11_0,__vite_glob_11_1,__vite_glob_11_2]);
export const namedEagerWithObjectValues = Object.values(
  import.meta.glob<string>('./modules/*.ts', {
    eager: true,
    import: 'name',
  }),
)
```
it would not be possible to verify whether the values/keys were omitted. Should I do it anyway and just remove these tests?

underfin: The keys and values could be test like https://github.com/rolldown/rolldown/blob/8346be737ee073f321e799ba68225ca79a49bda0/packages/rolldown/tests/fixtures/builtin-plugin/glob-import/basic/assert.mjs#L5-L8

jrmajor: Consider the following code:
```js
export const foo = Object.keys(import.meta.glob('./modules/*.ts'))
```
Vite currently produces this:
```js
Object.keys({
	"./modules/a.ts": 0,
	"./modules/b.ts": 0,
	"./modules/index.ts": 0
});
```
but Rolldown produces this:
```js
Object.keys({
	"./modules/a.ts": () => import("./a-DAJhtQJU.js"),
	"./modules/b.ts": () => import("./b-BQKRc37x.js"),
	"./modules/index.ts": () => import("./modules-Ietgo5aO.js")
});
```
As you can see, Rolldown produced three unused chunks, while Vite knew that we're only interested in keys, so it ommited the imported chunks entirely.

If I import `foo` in the `assert.mjs` file, it will be equal to `['./modules/a.ts', './modules/b.ts', './modules/index.ts']`, which is correct, but it will not be possible to verify whether the `() => import(...)` functions were replaced with `0`.

underfin: OK. I get it. It is reasonable.

packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/_config.ts
```diff
@@ -0,0 +1,31 @@
...
+import { defineTest } from 'rolldown-tests'
+import { expect } from 'vitest'
+import * as fs from 'node:fs'
+import * as path from 'node:path'
+
+// ported from https://github.com/vitejs/vite/tree/021443c5a11eedfb176f2c65d5fab65e9c21adfb/packages/vite/src/node/__tests__/plugins/importGlob
+export default defineTest({
+  config: {
+    input: './fixture-a/index.ts',
+    plugins: [
```
underfin, line 15: Could you add `output.chunkFilenames: '[name].js'`, it could avoid generate hash at snapshot and it avoid update it if hash changed.

jrmajor: Nice, changed it.

underfin (approved):

IWANABETHATGUY (approved):

PR #3626, opened 2025-02-19
chore: fix vitest warnings

jrmajor:
### Description

Fixes the following warning:
`Promise returned by expect(actual).toMatchFileSnapshot(expected) was not awaited. Vitest currently auto-awaits hanging assertions at the end of the test, but this will cause the test to fail in Vitest 3. Please remember to await the assertion.`


IWANABETHATGUY (approved):
thanks!

PR #3639, opened 2025-02-20
fix: don't enable drop_console when compress is true

### LINKED ISSUE #3637: Don't set `drop_console: true` for minifier by default
> @IWANABETHATGUY Can we set `drop_console: false`? It breaks the tests in Vite as it expects the console calls to be preserved. Also I think it's not expected to be set by default.
> _Originally posted by @sapphi-red in https://github.com/rolldown/rolldown/pull/3616#discussion_r1963419743_

IWANABETHATGUY:
### Description
1. Closed https://github.com/rolldown/rolldown/issues/3637

### DIFF
--- a/crates/rolldown_common/src/inner_bundler_options/types/minify_options.rs
+++ b/crates/rolldown_common/src/inner_bundler_options/types/minify_options.rs
@@ -84,7 +84,7 @@ impl MinifyOptionsObject {
   pub fn to_oxc_minifier_options(&self, target: ESTarget) -> oxc::minifier::MinifierOptions {
     oxc::minifier::MinifierOptions {
       mangle: self.mangle.then_some(MangleOptions { top_level: true, debug: false }),
-      compress: Some(CompressOptions { target, drop_debugger: true, drop_console: true })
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY:
* **#3639** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3639?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3639?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sapphi-red (approved):
Yeah, I prefer to set `drop_debugger: false` by default. But terser sets it `true` by default, so one can argue it is fine to be set by default (esbuild does not drop it by default).

IWANABETHATGUY:
> Yeah, I prefer to set `drop_debugger: false` by default. But terser sets it `true` by default, so one can argue it is fine to be set by default (esbuild does not drop it by default).

LGTM, we could expose more options to user when oxc-minifier got more stable.

PR #3645, opened 2025-02-20
refactor(rust): improve `rename_non_root_symbol`

shulaoda:
### Description

See the comments

### DIFF
--- a/crates/rolldown/src/utils/renamer.rs
+++ b/crates/rolldown/src/utils/renamer.rs
@@ -180,9 +180,8 @@ impl<'name> Renamer<'name> {
               break;
             }
           },
-          Entry::Occupied(slot) => {
+          Entry::Occupied(_) => {
             // The symbol is already renamed
-            used_canonical_names_for_this_scope.insert(slot.get().clone(), 0);
           }
         }
       });
--- a/crates/rolldown/src/utils/renamer.rs
+++ b/crates/rolldown/src/utils/renamer.rs
@@ -5,6 +5,7 @@ use rolldown_common::{
   SymbolRefDb,
 };
 use rolldown_rstr::{Rstr, ToRstr};
+use rolldown_utils::rustc_hash::FxHashMapExt;
 use rolldown_utils::{
   concat_string,
   rayon::{IntoParallelIterator, IntoParallelRefIterator, ParallelIterator},
@@ -154,9 +155,9 @@ impl<'name> Renamer<'name> {
       ast_scope: &'name AstScopes,
     ) {
       let mut bindings = ast_scope.get_bindings(scope_id).iter().collect::<Vec<_>>();
+      let mut used_canonical_names_for_this_scope = FxHashMap::with_capacity(bindings.len());
+
       bindings.sort_unstable_by_key(|(_, symbol_id)| *symbol_id);
-      let mut used_canonical_names_for_this_scope = FxHashMap::default();
-      used_canonical_names_for_this_scope.shrink_to(bindings.len());
       bindings.iter().for_each(|(binding_name, &symbol_id)| {
         let binding_ref: SymbolRef = (module.idx, symbol_id).into();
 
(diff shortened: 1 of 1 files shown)


crates/rolldown/src/utils/renamer.rs
```diff
@@ -154,9 +155,9 @@ impl<'name> Renamer<'name> {
       ast_scope: &'name AstScopes,
     ) {
       let mut bindings = ast_scope.get_bindings(scope_id).iter().collect::<Vec<_>>();
+      let mut used_canonical_names_for_this_scope = FxHashMap::with_capacity(bindings.len());
+
       bindings.sort_unstable_by_key(|(_, symbol_id)| *symbol_id);
-      let mut used_canonical_names_for_this_scope = FxHashMap::default();
-      used_canonical_names_for_this_scope.shrink_to(bindings.len());
```
shulaoda, line 159: The behavior of using `shrink_to` after calling `FxHashMap::default()` is meaningless.

crates/rolldown/src/utils/renamer.rs
```diff
@@ -179,11 +180,11 @@ impl<'name> Renamer<'name> {
               break;
             }
           },
-          Entry::Occupied(_) => {
+          Entry::Occupied(slot) => {
             // The symbol is already renamed
+            used_canonical_names_for_this_scope.insert(slot.get().clone(), 0);
           }
         }
-        used_canonical_names_for_this_scope.insert(binding_name.to_rstr(), 0);
```
shulaoda: We should only consider the `candidate_name` corresponding to the `binding_ref`, as for other symbols to be renamed in this scope, the `binding_name` originally associated with the renamed `symbol_id` is no longer valid.

shulaoda: BTW, I can't think of how a binding in the `non-top-scope` would be prematurely deconflicted into `canonical_names`, so I believe `Entry::Occupied(_)` is `unreachable`. However, to account for any potential edge cases, let's do it anyway.

IWANABETHATGUY: Can you provide related esbuild code and see if the code movement is correct?

crates/rolldown/src/utils/renamer.rs
```diff
@@ -179,11 +180,11 @@ impl<'name> Renamer<'name> {
               break;
             }
           },
-          Entry::Occupied(_) => {
+          Entry::Occupied(slot) => {
             // The symbol is already renamed
+            used_canonical_names_for_this_scope.insert(slot.get().clone(), 0);
```
IWANABETHATGUY: this operation seems just insert itself again, you could just remove this line.

IWANABETHATGUY: https://github.com/rolldown/rolldown/pull/3645/files#diff-c87caec33470b59eac43cd6ac05bef999e6a4fa010e716fd3e546971148f01d2R178

IWANABETHATGUY:
Rest part looks good to me

crates/rolldown/src/utils/renamer.rs
shulaoda: This was introduced by #2673, but it doesn't seem necessary.

shulaoda: https://github.com/evanw/esbuild/blob/main/internal/renamer/renamer.go#L580-L628

You could see `s.nameCounts[name] = 1`, `esbuild` only counts the final `name`, and if the `name` is calculated as new, it only counts the new `name`.

shulaoda: Therefore, the `used_canonical_names_for_this_scope.insert` in the following logic is sufficient.

If unchanged, it will record the original `candidate_name`, if changed, it will record the new `candidate_name`.

https://github.com/rolldown/rolldown/blob/397fdafe98f47cafc6f9b8044d3ddfef0ebe73b1/crates/rolldown/src/utils/renamer.rs#L167-L181

IWANABETHATGUY (approved):

PR #3647, opened 2025-02-21
fix(minify): disable top_level option for iife format

### LINKED ISSUE #3646: `top_level` option for mangler should not be enabled for `format: iife`
`top_level` option for mangler should not be enabled for `format: iife` here.
https://github.com/rolldown/rolldown/blob/61bbd9474da045d4c44d92bfb0bd5ac224474a6c/crates/rolldown_common/src/inner_bundler_options/types/minify_options.rs#L86

IIFE exposes a variable on top level scope, so the top-level variable should not be mangled.
```js
var MyLib = (function () {
  // content
  return {}
}());
```

underfin:
### Description

close https://github.com/rolldown/rolldown/issues/3646.


hyf0 (approved):
LGTM.

PR #3652, opened 2025-02-21
feat(mf): add pre order to the resolve_id_meta

underfin:

### DIFF
--- a/crates/rolldown_plugin_module_federation/src/lib.rs
+++ b/crates/rolldown_plugin_module_federation/src/lib.rs
@@ -263,6 +263,11 @@ impl Plugin for ModuleFederationPlugin {
     Ok(())
   }
 
+  // Make sure the plugin resolve_id hook called at first, avoid vite resolver resolved the shared module.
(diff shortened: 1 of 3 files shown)


crates/rolldown_plugin_module_federation/src/lib.rs
```diff
@@ -263,6 +263,11 @@ impl Plugin for ModuleFederationPlugin {
     Ok(())
   }
 
+  // Make sure the plugin resolve_id hook called at first, avoid vite resolver resolved the shared module.
+  fn resolve_id_meta(&self) -> Option<rolldown_plugin::PluginHookMeta> {
+    Some(rolldown_plugin::PluginHookMeta { order: Some(rolldown_plugin::PluginOrder::Pre) })
```
hyf0, line 268: Does rollup have this concept `PluginOrder: 'pre'`?

hyf0 (approved):

crates/rolldown_plugin_module_federation/src/lib.rs
underfin: see https://rollupjs.org/plugin-development/#build-hooks

PR #3657, opened 2025-02-21
feat: omit keys/values in glob import if they're unused

jrmajor:
### Description

Feel free to make changes directly to my branch if it's easier for you than leaving review comments.

### DIFF
--- a/crates/rolldown_plugin_import_glob/src/lib.rs
+++ b/crates/rolldown_plugin_import_glob/src/lib.rs
@@ -5,7 +5,7 @@ use oxc::{
     AstBuilder, NONE, VisitMut,
     ast::{
       Argument, ArrayExpressionElement, Expression, FormalParameterKind, ImportOrExportKind,
-      ObjectPropertyKind, PropertyKey, PropertyKind, Statement,
+      NumberBase, ObjectPropertyKind, PropertyKey, PropertyKind, Statement,
     },
     visit::walk_mut,
   },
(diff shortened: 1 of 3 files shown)


crates/rolldown_plugin_import_glob/src/lib.rs
```diff
@@ -82,52 +82,91 @@ pub struct GlobImportVisit<'ast, 'a> {
...
+      Expression::from(self.ast_builder.member_expression_static(
+        callee.span,
+        self.ast_builder.expression_identifier(i.span, "Object"),
+        callee.property.clone(),
+        false,
+      )),
+      NONE,
+      self.ast_builder.vec1(Argument::from(new_expr)),
+      false,
+    ))
```
jrmajor: Is there a way to only mutate `expr.arguments` instead of recreating the entire `CallExpression`? I can't figure it out.

jrmajor: Never mind, I figured it out :)

crates/rolldown_plugin_import_glob/src/lib.rs
```diff
@@ -82,52 +82,74 @@ pub struct GlobImportVisit<'ast, 'a> {
...
+
+    let mut opts = ImportGlobOptions::default();
+    match call_expr.arguments.as_slice() {
+      [first] => self.eval_glob_expr(first, &mut files),
+      // import.meta.glob('./dir/*.js', { import: 'setup' })
+      [first, second] => {
+        self.eval_glob_expr(first, &mut files);
+        extract_import_glob_options(second, &mut opts);
       }
+      [first, second, _rest @ ..] => {
```
IWANABETHATGUY: the `_rest` does not seem to be needed

jrmajor: Removed the redundant branch.

jrmajor:
Test failure seems unrelated, force pushing to trigger CI re-run.

jrmajor:
Still the same failure, is it possible that it's related to this PR? I can't see how these changes would affect that test.

IWANABETHATGUY:
> Still the same failure, is it possible that it's related to this PR? I can't see how these changes would affect that test.

You could ignore the watcher test failure

IWANABETHATGUY (approved):

PR #3658, opened 2025-02-22
chore: Rust v1.85.0

Boshen:


hyf0 (approved):

crates/rolldown/tests/rolldown/issues/3438/artifacts.snap
```diff
@@ -9,7 +8,7 @@ snapshot_kind: text
 ```js
 import { clear } from "./repro12.js";
 import { clear$1, clear$1$1 } from "./repro22.js";
-import { clear$1$3 as clear$1$2, clear$2 as clear$2$1, clear$3 as clear$2 } from "./repro32.js";
```
hyf0: Ideally, this shouldn't be changed. I guess this is still due to some logic rely on order of `HashMap`.

Boshen: There's an update in hash value https://github.com/rust-lang/rustc-hash/commit/177c5665743ab200cc7620caee88cad922ff574c after bumping rustc-hash.

PR #3672, opened 2025-02-24
perf: use non box type for module to avoid heap allocation

xusd320:
As title.

### DIFF
--- a/crates/rolldown_common/src/module/mod.rs
+++ b/crates/rolldown_common/src/module/mod.rs
@@ -11,8 +11,8 @@ use crate::{
 
 #[derive(Debug)]
 pub enum Module {
-  Normal(Box<NormalModule>),
-  External(Box<ExternalModule>),
+  Normal(NormalModule),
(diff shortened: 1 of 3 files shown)


crates/rolldown_common/src/module/mod.rs
```diff
@@ -11,8 +11,8 @@ use crate::{
 
 #[derive(Debug)]
 pub enum Module {
-  Normal(Box<NormalModule>),
-  External(Box<ExternalModule>),
```
jaredleechn, line 15: whats the Box<> for?

jrmajor: I think this section from Oxc docs explains it well: https://oxc.rs/docs/learn/parser_in_rust/ast.html#enum-size. And here's a more general explanation of `Box`: https://doc.rust-lang.org/book/ch15-01-box.html.

xusd320: Ok, I see why.

PR #3678, opened 2025-02-25
refactor(rust): apply `clippy::needless_pass_by_ref_mut`

shulaoda:
### Description

Related to #3676

### DIFF
--- a/crates/rolldown/src/bundler.rs
+++ b/crates/rolldown/src/bundler.rs
@@ -133,7 +133,7 @@ impl Bundler {
     Ok(output)
   }
 
-  #[allow(clippy::missing_transmute_annotations)]
+  #[allow(clippy::missing_transmute_annotations, clippy::needless_pass_by_ref_mut)]
   async fn bundle_up(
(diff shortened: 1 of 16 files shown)


Brooooooklyn (approved):

crates/rolldown/src/bundler.rs
```diff
@@ -133,7 +133,7 @@ impl Bundler {
     Ok(output)
   }
 
-  #[allow(clippy::missing_transmute_annotations)]
+  #[allow(clippy::missing_transmute_annotations, clippy::needless_pass_by_ref_mut)]
   async fn bundle_up(
     &mut self,
```
shulaoda, line 138: Removing the unnecessary `mut` causes issues with the tests below, as well as with the `Send` trait in the rolldown napi binding. Therefore, I decided to keep it.

https://github.com/rolldown/rolldown/blob/6a7faa0c3951c64df312fb66f5c7841ee78211a2/crates/rolldown/src/bundler.rs#L193-L202

PR #3682, opened 2025-02-25
feat: exclude self from glob import

jrmajor:
### Description

Do not include importing module in `import.meta.glob`.

### DIFF
--- a/packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/fixture-a/index.ts
+++ b/packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/fixture-a/index.ts
@@ -73,7 +73,6 @@ export const rawImportModule = import.meta.glob(
   { query: '?raw', import: '*' },
 )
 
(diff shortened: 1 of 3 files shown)


packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/fixture-a/index.ts
```diff
@@ -91,14 +91,14 @@ export const rawImportModule = import.meta.glob(
...
 
-// todo: shouldn't contain './index.ts'
 export const excludeSelf = import.meta.glob(
   './*.ts',
   // for test: annotation contain ")"
   /*
    * for test: annotation contain ")"
    * */
 )
+export const excludeSelfRaw = import.meta.glob('./*.ts', { query: '?raw' })
```
jrmajor, line 83: Added this to Vite too: vitejs/vite#19516.

IWANABETHATGUY (approved):

PR #3688, opened 2025-02-26
chore: silence rolldown warnings in tests

jrmajor:
### Description

Added `onLog` with assertions to tests that emit warnings. This will ensure that warnings are also covered by tests, while also silencing them so that they're not written to stderr (it makes Vitest output harder to read).

### DIFF
--- a/packages/rolldown/tests/fixtures/builtin-plugin/build-import-analysis/basic/_config.ts
+++ b/packages/rolldown/tests/fixtures/builtin-plugin/build-import-analysis/basic/_config.ts
@@ -1,6 +1,8 @@
 import { buildImportAnalysisPlugin } from 'rolldown/experimental'
 import { defineTest } from 'rolldown-tests'
-import { expect } from 'vitest'
+import { expect, vi } from 'vitest'
+
+const onLogFn = vi.fn()
 
 export default defineTest({
   skipComposingJsPlugin: true,
@@ -32,9 +34,20 @@ export const __vitePreload = (v) => {
         isRelativeBase: false,
       }),
     ],
+    onLog(level, log) {
+      expect(level).toBe('warn')
+      expect(log.code).toBe('UNRESOLVED_IMPORT')
+      expect(log.message).toContain(
+        "Could not resolve 'node:assert' in main.js",
+      )
+      onLogFn()
+    },
   },
   async afterTest(output) {
+    expect(onLogFn).toHaveBeenCalledTimes(1)
+
     await import('./assert.mjs')
+
     output.output.forEach((item) => {
       if (item.type === 'chunk' && item.name === 'main') {
         expect(item.code).to.not.includes('import.meta.url')
--- a/packages/rolldown/tests/fixtures/builtin-plugin/build-import-analysis/basic/_config.ts
+++ b/packages/rolldown/tests/fixtures/builtin-plugin/build-import-analysis/basic/_config.ts
@@ -1,8 +1,6 @@
 import { buildImportAnalysisPlugin } from 'rolldown/experimental'
 import { defineTest } from 'rolldown-tests'
-import { expect, vi } from 'vitest'
-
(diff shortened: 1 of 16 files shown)


packages/rolldown/tests/fixtures/jsx/refresh/_config.ts
```diff
@@ -1,16 +1,28 @@
...
 
+const onLogFn = vi.fn()
+
 export default defineTest({
   config: {
     input: 'main.jsx',
     jsx: {
       refresh: true,
     },
     external: ['react'],
```
IWANABETHATGUY: you could add an extra external `react/jsx-runtime`

packages/rolldown/tests/fixtures/function/define/_config.ts
```diff
@@ -1,14 +1,26 @@
...
 export default defineTest({
   config: {
     define: {
       'process.env.NODE_ENV': '"production"',
     },
+    onLog(level, log) {
+      expect(level).toBe('warn')
+      expect(log.code).toBe('UNRESOLVED_IMPORT')
+      expect(log.message).toContain(
+        "Could not resolve 'node:assert' in main.js",
```
IWANABETHATGUY: add a external

packages/rolldown/tests/fixtures/builtin-plugin/build-import-analysis/basic/_config.ts
```diff
@@ -32,9 +34,20 @@ export const __vitePreload = (v) => {
         isRelativeBase: false,
       }),
     ],
+    onLog(level, log) {
+      expect(level).toBe('warn')
+      expect(log.code).toBe('UNRESOLVED_IMPORT')
+      expect(log.message).toContain(
+        "Could not resolve 'node:assert' in main.js",
```
IWANABETHATGUY: add to external

packages/rolldown/tests/fixtures/misc/keep-names/basic/_config.ts
```diff
@@ -1,10 +1,23 @@
...
+const onLogFn = vi.fn()
 
 export default defineTest({
   config: {
     keepNames: true,
+    onLog(level, log) {
+      expect(level).toBe('warn')
+      expect(log.code).toBe('UNRESOLVED_IMPORT')
+      expect(log.message).toContain(
+        "Could not resolve 'node:assert' in main.js",
```
IWANABETHATGUY: add external

packages/rolldown/tests/fixtures/plugin/load/guess-module-type/_config.ts
```diff
@@ -16,6 +18,16 @@ export default defineTest({
         },
       },
     ],
+    onLog(level, log) {
+      expect(level).toBe('warn')
+      expect(log.code).toBe('UNRESOLVED_IMPORT')
+      expect(log.message).toContain(
+        "Could not resolve 'react/jsx-runtime' in main.jsx",
```
IWANABETHATGUY: add to external

packages/rolldown/tests/fixtures/plugin/context/cycle-load-error/_config.ts
```diff
@@ -8,19 +8,19 @@ export default defineTest({
     plugins: [
       {
         name: 'test-plugin-context',
-        onLog: (level, log) => {
```
jrmajor, line 11: `onLog` in plugin does not prevent the log from being written to `stderr`, but top level `onLog` does.

packages/rolldown/tests/fixtures/jsx/refresh/_config.ts
jrmajor: Done.

IWANABETHATGUY (approved):

PR #3690, opened 2025-02-26
perf(rolldown_plugin_replace): replace static Regex with pure function

Brooooooklyn:
### Benchmark

```
running 2 tests
test tests::pure_function ... bench:          80.56 ns/iter (+/- 12.09)
test tests::regex         ... bench:         201.07 ns/iter (+/- 6.74)

test result: ok. 0 passed; 0 failed; 0 ignored; 2 measured; 0 filtered out; finished in 1.20s
```

```rust
#![feature(test)]

#[cfg(test)]
extern crate test;

use std::sync::LazyLock;

use regex::Regex;

// A long JavaScript snippet to test variable declaration detection
const VARIABLE_DECLARATION_PREFIXES: &[&str] = &[
    // JavaScript code with various variable declarations
    "process.env.NODE_ENV = 'development';",
    "const process = { env: { NODE_ENV: 'development' } };",
    "if (process.env.NODE_ENV === 'development') { }",
    "for (let i = 0; i < 10; i++) { }",
    "class Example { constructor() { } }",
    "const v = 'development';",
    "let v = 'development';",
    "var v = 'development';",
    "function process() { }",
    "const obj = { method() { } }",
    "// This is a comment with const ",
    "/* Multi-line comment with let ",
    "console.log('not a declaration'); const ",
    "export const ",
    "import { useState } from 'react'; const ",
    "async function fetch() { }",
    "try { }",
    "catch (error) { }",
    "finally { }",
    "const v = \"development\";",
];

static NON_ASSIGNMENT_MATCHER: LazyLock<Regex> =
    LazyLock::new(|| Regex::new("\\b(?:const|let|var)\\s+$").expect("Should be valid regex"));

// Checks if the given string ends with a variable declaration keyword (const, let, var)
// followed by whitespace, which would indicate the start of a variable declaration.
fn is_variable_declaration_prefix(s: &str) -> bool {
    // First check if there's any whitespace at the end
    if !s.ends_with(|c: char| c.is_whitespace()) {
        return false;
    }

    // Trim the trailing whitespace
    let s = s.trim_end();

    // Check for word boundary before the keywords
    (s.ends_with("const")
        && (s.len() == 5 || !s.chars().nth(s.len() - 6).unwrap_or(' ').is_alphanumeric()))
        || (s.ends_with("let")
            && (s.len() == 3 || !s.chars().nth(s.len() - 4).unwrap_or(' ').is_alphanumeric()))
        || (s.ends_with("var")
            && (s.len() == 3 || !s.chars().nth(s.len() - 4).unwrap_or(' ').is_alphanumeric()))
}

fn is_variable_declaration_prefix_regex(s: &str) -> bool {
    NON_ASSIGNMENT_MATCHER.is_match(s)
}

#[cfg(test)]
mod tests {
    use test::{black_box, Bencher};

    use super::*;

    #[bench]
    fn pure_function(b: &mut Bencher) {
        b.iter(|| {
            for prefix in VARIABLE_DECLARATION_PREFIXES {
                black_box(is_variable_declaration_prefix(prefix));
            }
        });
    }

    #[bench]
    fn regex(b: &mut Bencher) {
        b.iter(|| {
            for prefix in VARIABLE_DECLARATION_PREFIXES {
                black_box(is_variable_declaration_prefix_regex(prefix));
            }
        });
    }
}

```

### DIFF
--- a/crates/rolldown_plugin_replace/src/plugin.rs
+++ b/crates/rolldown_plugin_replace/src/plugin.rs
@@ -1,11 +1,6 @@
 use std::ops::Range;
-use std::{
-  cmp::Reverse,
-  sync::{Arc, LazyLock},
-};
+use std::{cmp::Reverse, sync::Arc};
 
-// use fancy_regex::Regex;
-use regex::Regex;
 use rolldown_plugin::{HookRenderChunkOutput, HookTransformOutput, Plugin};
 use rustc_hash::FxHashMap;
 use string_wizard::{MagicString, SourceMapOptions};
@@ -38,8 +33,25 @@ pub struct ReplacePlugin {
   sourcemap: bool,
 }
 
-static NON_ASSIGNMENT_MATCHER: LazyLock<Regex> =
-  LazyLock::new(|| Regex::new("\\b(?:const|let|var)\\s+$").expect("Should be valid regex"));
+// Checks if the given string ends with a variable declaration keyword (const, let, var)
+// followed by whitespace, which would indicate the start of a variable declaration.
+fn is_variable_declaration_prefix(s: &str) -> bool {
+  // First check if there's any whitespace at the end
+  if !s.ends_with(|c: char| c.is_whitespace()) {
+    return false;
+  }
+
+  // Trim the trailing whitespace
+  let s = s.trim_end();
+
+  // Check for word boundary before the keywords
+  (s.ends_with("const")
+    && (s.len() == 5 || !s.chars().nth(s.len() - 6).unwrap_or(' ').is_alphanumeric()))
+    || (s.ends_with("let")
+      && (s.len() == 3 || !s.chars().nth(s.len() - 4).unwrap_or(' ').is_alphanumeric()))
+    || (s.ends_with("var")
+      && (s.len() == 3 || !s.chars().nth(s.len() - 4).unwrap_or(' ').is_alphanumeric()))
+}
 
 impl ReplacePlugin {
   pub fn new(values: FxHashMap<String, String>) -> Self {
@@ -112,7 +124,7 @@ impl ReplacePlugin {
   fn look_around_assert(&self, code: &str, matched_range: Range<usize>) -> bool {
     if self.prevent_assignment {
       let before = &code[..matched_range.start];
-      if NON_ASSIGNMENT_MATCHER.is_match(before) {
+      if is_variable_declaration_prefix(before) {
         return true;
       }
     }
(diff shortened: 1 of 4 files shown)


Brooooooklyn:
* **#3690** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3690?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3690?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/rolldown_plugin_replace/src/utils.rs
```diff
@@ -1,18 +1,74 @@
...
+        return false;
+      }
+    }
+  }
+
+  true
+}
+
+/// Checks if a character is valid as the first character of a JavaScript identifier
+fn is_valid_identifier_start(c: char) -> bool {
```
IWANABETHATGUY: https://github.com/rolldown/rolldown/blob/f3e6933176440ab845a1e3ecec80294275cfbbed/crates/rolldown_utils/src/ecmascript.rs?plain=1#L41-L45

crates/rolldown_plugin_replace/src/plugin.rs
```diff
@@ -38,8 +33,25 @@ pub struct ReplacePlugin {
...
+  if !s.ends_with(|c: char| c.is_whitespace()) {
+    return false;
+  }
+
+  // Trim the trailing whitespace
+  let s = s.trim_end();
+
+  // Check for word boundary before the keywords
+  (s.ends_with("const")
+    && (s.len() == 5 || !s.chars().nth(s.len() - 6).unwrap_or(' ').is_alphanumeric()))
```
IWANABETHATGUY, line 49: This seems  inefficient, as chars().nth() has a time complexity of O(N).

Brooooooklyn: I believe it's okay here, the logic will only enter one of the branches.
Anyway, let me run benchmark

Brooooooklyn:
```
running 2 tests
test tests::pure_function ... bench:          80.56 ns/iter (+/- 12.09)
test tests::regex         ... bench:         201.07 ns/iter (+/- 6.74)
``` 
also updated the pr description

IWANABETHATGUY (approved):

Issue #3619, opened 2025-02-19
[Bug]: Paniced when bundling mermaid


ikkz:
### Reproduction link or steps

https://stackblitz.com/edit/github-u37xaqxr?file=rolldown.config.mjs

npm run build

### What is expected?

Build success

### What is actually happening?

```
~/projects/github-u37xaqxr 3s
❯ npm run build

> build
> RD_LOG=error rolldown -c ./rolldown.config.mjs

(node:16) ExperimentalWarning: WASI is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:16) ExperimentalWarning: WASI is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
thread '<unnamed>' panicked at /home/runner/work/rolldown/rolldown/crates/rolldown_std_utils/src/option_ext.rs:34:19:
Got `None` value when calling `OptionExt::unpack()` on `Option<&EsmNamespaceInCjs>`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
worker (tid = 43) sent an error! unreachable
[Error [RuntimeError]: unreachable]

Node.js v18.20.3
```

### System Info

```Shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.20.3 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    rolldown: 1.0.0-beta.3-commit.93d8731 => 1.0.0-beta.3-commit.93d8731
```

IWANABETHATGUY:
seems only happen when format is `iife`

Boshen:
@shulaoda can you check whether this is resolved.

shulaoda:
The latest branch version (`1.0.0-beta.8-commit.e5c11c6`) seems to have solved this issue.

ikkz:
Thanks for fixing this! 🙏

Issue #3529, opened 2025-02-09
[Bug]: Importing CJS with circular deps in ESM results in a runtime error


thecrypticace:
### Reproduction link or steps

1. Create two CJS files with circular dependencies
2. Export stuff from both files so the deps can't be tree-shaken
3. Import both of these files into an ESM file
4. Build with Rolldown
5. Run the output file with Node

Reproduction here:
https://github.com/thecrypticace/rolldown-cjs-bug

Aside: I discovered a crash while working on this bug report. I've included details about that as well in the repo above. I can open a separate issue about this if you prefer!

### What is expected?

After building I expect the code to run without erroring.

### What is actually happening?

Running the built code produces a runtime error:
```
const a = require_a();
          ^
TypeError: require_a is not a function
```

### System Info

```Shell
- MacBook Pro 16-inch, Nov 2023
- Apple M3 Max
- 64GB RAM
- macOS 15.3
- zsh 5.9 (arm64-apple-darwin24.0)
- Node 22.12.0 (also tested v23.6.0)
- npm 10.9.2
```

### Any additional comments?

I discovered this bug when attempting to compile `prettier-plugin-tailwindcss` with Rolldown. We bundle a "fallback" version of v3 and at least two of the files we import have circular dependencies. I discovered this bug in `1.0.0-beta.3` but have confirmed that it still exists in the latest nightly and canary builds.

IWANABETHATGUY:
[...]

PR #3693, opened 2025-02-26
fix(cjs): ensure esm namespace always exist

hyf0:
### Description

- This partially fix the panic problem of #3619, but the output is still not runnable due to #3529.

### DIFF
--- a/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
+++ b/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
@@ -6,9 +6,9 @@ use oxc::span::CompactStr;
 // TODO: The current implementation for matching imports is enough so far but incomplete. It needs to be refactored
 // if we want more enhancements related to exports.
 use rolldown_common::{
-  EcmaModuleAstUsage, ExportsKind, ImportKind, IndexModules, Module, ModuleIdx, ModuleType,
-  NamespaceAlias, NormalModule, OutputFormat, ResolvedExport, Specifier, SymbolOrMemberExprRef,
-  SymbolRef, SymbolRefDb, WrapKind,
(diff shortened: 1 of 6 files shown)


hyf0:
* **#3693** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3693?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3693?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/rolldown_common/src/types/stmt_info.rs
```diff
@@ -101,6 +101,7 @@ pub struct StmtInfo {
   #[cfg(debug_assertions)]
   pub debug_label: Option<String>,
   pub meta: StmtInfoMeta,
+  pub force_tree_shaking: bool,
```
hyf0, line 104: Still borrowed from esbuild

https://github.com/evanw/esbuild/blob/a97f18b6d2c0ae68c955633aba81a9803d769c71/internal/ast/ast.go#L1092

crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
```diff
@@ -123,53 +123,7 @@ impl LinkStage<'_> {
   /// ```
   ///
   /// Unlike import from normal modules, the imported variable deosn't have a place that declared the variable. So we consider `import { a } from 'external'` in `foo.js` as the declaration statement of `a`.
-  #[expect(clippy::too_many_lines)]
   pub(super) fn bind_imports_and_exports(&mut self) {
-    self.sorted_modules.iter().copied().for_each(|importer_idx| {
```
hyf0, line 128: It turns out analysis here is already done by wrapping analysis.

IWANABETHATGUY (approved):

PR #3699, opened 2025-02-27
feat(glob_plugin): ensure query starts with `?`

jrmajor:
### Description

https://github.com/vitejs/vite/blob/ce0ccc6c33544a14728fff4b84c746c9682f9878/packages/vite/src/node/plugins/importMetaGlob.ts#L209

### DIFF
--- a/packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/fixture-a/index.ts.snap
+++ b/packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/fixture-a/index.ts.snap
@@ -83,7 +83,8 @@ const rawImportModule = {
 	"./modules/b.ts": () => import("./b3.js")
 };
 const excludeSelf = { "./sibling.ts": () => import("./sibling.js") };
-const excludeSelfRaw = { "./sibling.ts": () => import("./sibling2.js") };
+const excludeSelfRaw = { "./sibling.ts": () => import("./sibling3.js") };
+const customQueryString = { "./sibling.ts": () => import("./sibling2.js") };
 const parent = {};
 const cleverCwd1 = { "./node_modules/framework/pages/hello.page.js": () => import("./hello.page.js") };
 const cleverCwd2 = {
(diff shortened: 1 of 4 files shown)


packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/fixture-a/index.ts.snap
```diff
@@ -83,7 +83,8 @@ const rawImportModule = {
 	"./modules/b.ts": () => import("./b3.js")
 };
 const excludeSelf = { "./sibling.ts": () => import("./sibling.js") };
-const excludeSelfRaw = { "./sibling.ts": () => import("./sibling2.js") };
+const excludeSelfRaw = { "./sibling.ts": () => import("./sibling3.js") };
+const customQueryString = { "./sibling.ts": () => import("./sibling2.js") };
```
IWANABETHATGUY, line 87: Can you reference related vite tests?

IWANABETHATGUY: The output seems not correct compare to https://github.com/rolldown/vite/blob/dd0fb27304716203e7d45492f87a4b247c9bc1a1/packages/vite/src/node/__tests__/plugins/importGlob/__snapshots__/fixture.spec.ts.snap, there is no `sibling3.js` and `sibling2.js`

IWANABETHATGUY: It seems caused by our test flow not same as vite, so ignore for now.

IWANABETHATGUY (approved):

packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/fixture-a/index.ts.snap
jrmajor: Yeah, Vite test is for dev mode, so they just leave the query string as-is. We do full bundle, so the same file with different query string produces different chunks.

PR #3701, opened 2025-02-27
fix(glob_import): return absolute paths for absolute globs

jrmajor:
### Description

Vite plugin:
- https://github.com/vitejs/vite/blob/ce0ccc6c33544a14728fff4b84c746c9682f9878/packages/vite/src/node/plugins/importMetaGlob.ts#L312
- https://github.com/vitejs/vite/blob/ce0ccc6c33544a14728fff4b84c746c9682f9878/packages/vite/src/node/plugins/importMetaGlob.ts#L441-L450

Vite test:
- https://github.com/vitejs/vite/blob/ce0ccc6c33544a14728fff4b84c746c9682f9878/packages/vite/src/node/__tests__/plugins/importGlob/__snapshots__/fixture.spec.ts.snap#L84

### DIFF
--- a/crates/rolldown_plugin_import_glob/src/lib.rs
+++ b/crates/rolldown_plugin_import_glob/src/lib.rs
@@ -244,7 +244,7 @@ fn extract_import_glob_options(arg: &Argument, opts: &mut ImportGlobOptions) {
 }
 
 impl<'ast> GlobImportVisit<'ast, '_> {
-  fn eval_glob_expr(&self, arg: &Argument, files: &mut std::vec::Vec<String>) {
+  fn eval_glob_expr(&self, arg: &Argument, files: &mut std::vec::Vec<(String, String)>) {
     let mut positive_globs = vec![];
     let mut negated_globs = vec![];
     match arg {
@@ -282,6 +282,10 @@ impl<'ast> GlobImportVisit<'ast, '_> {
       })
       .collect::<std::vec::Vec<_>>();
 
+    let is_relative = positive_globs.iter().all(|g| g.starts_with('.'));
+
+    let self_path = self.format_path(Path::new(self.id), Some(dir));
+
     for glob_expr in positive_globs {
       let processed_glob_expr = preprocess_glob_expr(glob_expr);
       let absolute_glob = to_absolute_glob(&processed_glob_expr, dir, root).unwrap();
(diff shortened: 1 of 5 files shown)


crates/rolldown_plugin_import_glob/src/lib.rs
```diff
@@ -287,41 +291,55 @@ impl<'ast> GlobImportVisit<'ast, '_> {
...
-        if file == self.id.relative(dir).to_slash_lossy() {
+        let import_path = self.format_path(&file, Some(dir));
+        if import_path == self_path {
           continue;
         }
-        let prefix = if file.starts_with('.') { "" } else { "./" };
-        files.push(format!("{prefix}{file}"));
+        let file_path =
+          if is_relative { import_path.clone() } else { self.format_path(&file, None) };
+        files.push((file_path, import_path));
```
jrmajor: Should I make it a struct instead of a tuple?

IWANABETHATGUY: I think so

jrmajor: Done.

crates/rolldown_plugin_import_glob/src/lib.rs
```diff
@@ -287,41 +296,55 @@ impl<'ast> GlobImportVisit<'ast, '_> {
...
         }
-        let file = file.relative(dir).to_slash_lossy().to_string();
-        if file == self.id.relative(dir).to_slash_lossy() {
+        let import_path = self.format_path(&file, Some(dir));
+        if import_path == self_path {
           continue;
         }
-        let prefix = if file.starts_with('.') { "" } else { "./" };
-        files.push(format!("{prefix}{file}"));
+        let key = if is_relative { import_path.clone() } else { self.format_path(&file, None) };
```
underfin, line 307: Here the `import_path.clone()` could be using `&str` to avoid clone.

jrmajor: I tried doing this:
```rs
let key = if is_relative { &import_path } else { &self.format_path(&file, None) };
```
But then in `files.push(FileData { key, import_path })` it tells me that `key` is `&String` instead of `String`, so I have to change it to `key: key.to_string()`, and my understanding is that it's even worse, because before I would clone only if `is_relative`, and now I'm also cloning the `String` I'm getting from `format_path()`.

I also tried changing `clone()` to `to_string()` and it worked, but my understanding is that it's pretty much the same.

Or should I change struct definition to use `&str`?

I appreciate your patience, I've just begun my journey with Rust.

underfin: > Or should I change struct definition to use &str?

Yeah. It should be work.

jrmajor: [That](https://github.com/jrmajor/rolldown/commit/ebbf7f26f0dc9d4752c7d6ea214b4323206c8f22) doesn't work too, because `import_path` is dropped while still borrowed.

underfin: OK. I get it.

underfin (approved):

IWANABETHATGUY (approved):

PR #718, opened 2024-04-01
chore(node/bench): switch back to tinybench and prettify results disaply

hyf0:
### Description

image

### Test Plan

---

PR #3704, opened 2025-02-27
chore: remove unused vitest config

jrmajor:
### Description

Leftover from #718. Prevents Vitest plugin for VS Code from erroring on startup.


underfin (approved):

underfin:
Thank you.

PR #3706, opened 2025-02-28
perf(rolldown_std_utils): remove Regex and LazyLock

Brooooooklyn:

### DIFF
--- a/crates/rolldown_std_utils/src/pretty_type_name.rs
+++ b/crates/rolldown_std_utils/src/pretty_type_name.rs
@@ -1,17 +1,57 @@
 use std::borrow::Cow;
 
-use regex::Regex;
-use std::sync::LazyLock;
-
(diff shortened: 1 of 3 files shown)


crates/rolldown_std_utils/src/pretty_type_name.rs
```diff
@@ -1,17 +1,57 @@
 use std::borrow::Cow;
 
-use regex::Regex;
-use std::sync::LazyLock;
-
-static MODULE_MATCHER_RE: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"(?:\w+::)").unwrap());
```
hyf0, line 6: This is only used for debugging purpose. Normally this code won't be ran during bundling.

So maybe regex is more easier to maintain?

Brooooooklyn: For sure, I'll close it

PR #3709, opened 2025-02-28
refactor: change HookResolveIdOutput#id to ArcStr

underfin:
### Description

Avoid the builtin plugins has many allocate for `HookResolveIdOutput#id`.

### DIFF
--- a/crates/rolldown_plugin_build_import_analysis/src/lib.rs
+++ b/crates/rolldown_plugin_build_import_analysis/src/lib.rs
@@ -51,10 +51,7 @@ impl Plugin for BuildImportAnalysisPlugin {
     args: &HookResolveIdArgs<'_>,
   ) -> HookResolveIdReturn {
     if args.specifier == PRELOAD_HELPER_ID {
-      return Ok(Some(HookResolveIdOutput {
-        id: args.specifier.to_string(),
-        ..Default::default()
-      }));
+      return Ok(Some(HookResolveIdOutput { id: args.specifier.into(), ..Default::default() }));
     }
     Ok(None)
   }
--- a/crates/rolldown_plugin_dynamic_import_vars/src/lib.rs
+++ b/crates/rolldown_plugin_dynamic_import_vars/src/lib.rs
@@ -33,7 +33,7 @@ impl Plugin for DynamicImportVarsPlugin {
     args: &HookResolveIdArgs<'_>,
   ) -> HookResolveIdReturn {
     if args.specifier == DYNAMIC_IMPORT_HELPER {
-      Ok(Some(HookResolveIdOutput { id: DYNAMIC_IMPORT_HELPER.to_string(), ..Default::default() }))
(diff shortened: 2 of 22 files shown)


IWANABETHATGUY:
Actually the `into` still needs to allocate.

underfin:
> Actually the `into` still needs to allocate.

Yeah. The overhead is more small than clone string.

IWANABETHATGUY:
https://docs.rs/arcstr/1.2.0/src/arcstr/arc_str.rs.html#1182, and it seems has more complicated logic about the 

> > Actually the `into` still needs to allocate.
> 
> Yeah. The overhead is more small than clone string.

I don't see any clone in the diff, you could see https://docs.rs/arcstr/1.2.0/src/arcstr/arc_str.rs.html#1182, if there are no `clone()`,  `arcstr` will not have better performance, also since it has extra `AtomicUsize` it will cost more memory than normal `String`

crates/rolldown_plugin_build_import_analysis/src/lib.rs
```diff
@@ -51,10 +51,7 @@ impl Plugin for BuildImportAnalysisPlugin {
     args: &HookResolveIdArgs<'_>,
   ) -> HookResolveIdReturn {
     if args.specifier == PRELOAD_HELPER_ID {
-      return Ok(Some(HookResolveIdOutput {
-        id: args.specifier.to_string(),
```
underfin, line 55: The `&str` has an extra string clone.

crates/rolldown_plugin_dynamic_import_vars/src/lib.rs
```diff
@@ -33,7 +33,7 @@ impl Plugin for DynamicImportVarsPlugin {
     args: &HookResolveIdArgs<'_>,
   ) -> HookResolveIdReturn {
     if args.specifier == DYNAMIC_IMPORT_HELPER {
-      Ok(Some(HookResolveIdOutput { id: DYNAMIC_IMPORT_HELPER.to_string(), ..Default::default() }))
```
underfin, line 36: The `&'static str` has extra string clone.

crates/rolldown_plugin_module_federation/src/lib.rs
```diff
@@ -279,22 +279,22 @@ impl Plugin for ModuleFederationPlugin {
...
+        id: resolve_id.id,
         ..Default::default()
       }));
     }
     if args.specifier.starts_with(SHARED_MODULE_PREFIX) {
       let resolve_id =
         ctx.resolve(&args.specifier[SHARED_MODULE_PREFIX.len()..], None, None).await??;
       return Ok(Some(rolldown_plugin::HookResolveIdOutput {
-        id: resolve_id.id.to_string(),
+        id: resolve_id.id,
```
underfin, line 297: The `ArcStr` has extar string clone

IWANABETHATGUY (approved):

PR #3712, opened 2025-02-28
feat(mf): detect require module is remote module

underfin:

### DIFF
--- a/crates/rolldown_plugin_module_federation/src/init_modules_visitor.rs
+++ b/crates/rolldown_plugin_module_federation/src/init_modules_visitor.rs
@@ -5,7 +5,7 @@ use oxc::{
   ast::{
     AstBuilder, NONE, VisitMut,
     ast::{Argument, ExportAllDeclaration, ExportNamedDeclaration, Expression, ImportDeclaration},
-    visit::walk_mut::walk_expression,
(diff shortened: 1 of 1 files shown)


crates/rolldown_plugin_module_federation/src/init_modules_visitor.rs
```diff
@@ -113,4 +112,16 @@ impl<'ast> VisitMut<'ast> for InitModuleVisitor<'ast, '_> {
     }
     walk_expression(self, expr);
   }
+
+  // TODO need to semantic analysis at transform_ast hook
+  fn visit_call_expression(&mut self, call_expr: &mut oxc::ast::ast::CallExpression<'ast>) {
+    if let Expression::Identifier(id) = &call_expr.callee {
+      if id.name == "require" && call_expr.arguments.len() == 1 {
```
underfin, line 119: Here need to using semantic analysis to detect the global `require`. It need to add semantic data to `transform_ast ` hook. But it has overhead for the pipeline, @IWANABETHATGUY Other buitlin plugins used semantic data?

IWANABETHATGUY: Let's discuss it in next meeting.

PR #3716, opened 2025-02-28
chore: fix typos

phip1611:
Fix various typos across the source tree. I originally planned to integrate [typos](https://github.com/crate-ci/typos), but the tree has to many commit hashes and similar strings that lead to false positives.

Therefore, I manually picked the actual typos and this is what I've got from that.


IWANABETHATGUY:
We have `cspell` already, no need to add `typos`

packages/rollup-tests/src/ignored-passed-snapshot-different-tests.js
```diff
@@ -112,15 +112,15 @@ module.exports = [
     "rollup@form@exported-class-declaration-conflict: handles exporting class declarations with name conflicts in SystemJS@generates es",
     "rollup@form@external-empty-import-no-global: does not expect a global to be provided for empty imports (#1217)@generates es",
     "rollup@form@external-imports: prefixes global names with `global.` when creating UMD bundle (#57)@generates es",
-    "rollup@form@super-classes@super-class-prototype-assignment: correctly resolves the prototype of the super class when assigning properites",
+    "rollup@form@super-classes@super-class-prototype-assignment: correctly resolves the prototype of the super class when assigning properties",
```
IWANABETHATGUY: These typos came from rollup repository, which we clone from `rollup`, We don't plan to touch it.

IWANABETHATGUY (approved):

PR #3718, opened 2025-02-28
style: apply a few fixes suggested by clippy

phip1611:
Apply a few fixes suggested by clippy.

### DIFF
--- a/crates/rolldown/src/hmr/hmr_manager.rs
+++ b/crates/rolldown/src/hmr/hmr_manager.rs
@@ -18,7 +18,7 @@ impl HmrManager {
           changed_modules.push(*module_idx);
         }
         _ => {
-          dbg!("No corresponding module found for changed file path: {:?}", changed_file_path);
+          tracing::debug!("No corresponding module found for changed file path: {:?}", changed_file_path);
         }
       }
     }
--- a/crates/rolldown/src/hmr/hmr_manager.rs
+++ b/crates/rolldown/src/hmr/hmr_manager.rs
@@ -18,7 +18,7 @@ impl HmrManager {
           changed_modules.push(*module_idx);
         }
         _ => {
-          tracing::debug!("No corresponding module found for changed file path: {:?}", changed_file_path);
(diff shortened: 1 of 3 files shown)


crates/rolldown/src/hmr/hmr_manager.rs
```diff
@@ -18,7 +18,7 @@ impl HmrManager {
           changed_modules.push(*module_idx);
         }
         _ => {
-          dbg!("No corresponding module found for changed file path: {:?}", changed_file_path);
+          tracing::debug!("No corresponding module found for changed file path: {:?}", changed_file_path);
```
phip1611: dbg!() doesn't support format args

IWANABETHATGUY: I am not sure if here is intent to use `dbg` . @hyf0

IWANABETHATGUY: The rest part lgtm, can you revert this?

IWANABETHATGUY: https://github.com/rolldown/rolldown/pull/3718/files/d24844787e34825ab84f683e4648d010cdb12ed8#diff-dca406b959fe927dfee454b84b90e46c0463228652cc2a436ad7192f31664da9L11, seems this part is intentional

IWANABETHATGUY (approved):
Add some tweaks, thanks.

phip1611:
@IWANABETHATGUY Thanks for the merge!

However, switching the behavior back as you did in 45f4499b55e283ed148ab0e95517e102f81d1ff0 is not correct (it doesn't exactly do what you think it does). `dbg!()` does not support format args. 

```rust
    let foo = "bar";
    dbg!("foo = {:?}", foo);
```

prints

```
[src/bin/playground1.rs:4:5] "foo = {:?}" = "foo = {:?}"
[src/bin/playground1.rs:4:5] foo = "bar"
```

The format argument is ignored :)

An alternative would be to use `eprintln!()`.

PR #3724, opened 2025-02-28
test(node): improve `outro` and `intro` test

shulaoda:
### Description

Related to #3256 

I'm unsure whether we should add tests for the function of `intro` and `outro`, as their behavior might be similar to that of `banner`.

Also, should we add tests for `async function` of them, just like we did for `banner`?


hyf0:
> I'm unsure whether we should add tests for the function of intro and outro

Yeah. Passing function should be tested in the nodejs's side.

hyf0 (approved):

PR #3736, opened 2025-03-02
fix(rust): only replace specified-length placeholders for the `hash` type

shulaoda:
### Description

I improved performance in general cases and prevented `[name:8]` from being incorrectly replaced.

### DIFF
--- a/crates/rolldown_utils/src/replace_all_placeholder.rs
+++ b/crates/rolldown_utils/src/replace_all_placeholder.rs
@@ -23,58 +23,81 @@ where
 }
 
 /// Replace all `[placeholder]` or `[placeholder:8]` in the pattern
-pub fn replace_all_placeholder(
+pub trait ReplaceAllPlaceholder {
+  fn replace_all(self, placeholder: &str, replacer: impl Replacer) -> String;
+
+  fn replace_all_with_len(self, placeholder: &str, replacer: impl Replacer) -> String;
+}
+
+impl ReplaceAllPlaceholder for String {
+  #[inline]
+  fn replace_all(self, placeholder: &str, replacer: impl Replacer) -> String {
+    replace_all_placeholder_impl(self, false, placeholder, replacer)
+  }
+
+  #[inline]
+  fn replace_all_with_len(self, placeholder: &str, replacer: impl Replacer) -> String {
+    replace_all_placeholder_impl(self, true, placeholder, replacer)
+  }
+}
+
+fn replace_all_placeholder_impl(
   pattern: String,
-  placeholder: &str,
+  is_len_enabled: bool,
+  mut placeholder: &str,
   mut replacer: impl Replacer,
 ) -> String {
   let offset = placeholder.len() - 1;
-  let mut iter = pattern.match_indices(&placeholder[..offset]).peekable();
+
+  if is_len_enabled {
+    placeholder = &placeholder[..offset];
+  }
+
+  let mut iter = pattern.match_indices(placeholder).peekable();
 
   if iter.peek().is_none() {
     return pattern;
   }
 
-  let mut ending = 0;
+  let mut last_end = 0;
   let mut result = String::with_capacity(pattern.len());
 
   for (start, _) in iter {
-    if start < ending {
+    if start < last_end {
       continue;
     }
 
     let start_offset = start + offset;
-    let pat_temp = &pattern[start_offset..];
-
-    let (end, len) = match pat_temp.as_bytes().first() {
-      Some(b']') => (start_offset, None),
-      Some(b':') => {
-        if let Some(end) = pat_temp.find(']') {
-          let end = start_offset + end;
-          let len = pattern[start_offset + 1..end].parse::<usize>().ok();
-
-          if len.is_none() {
+    let (end, len) = if is_len_enabled {
+      let rest = &pattern[start_offset..];
+      match rest.as_bytes().first() {
+        Some(&b':') => {
+          if let Some(index) = rest.find(']') {
+            match rest[1..index].parse::<usize>() {
+              Ok(len) => (start_offset + index, Some(len)),
+              Err(_) => continue,
+            }
+          } else {
             continue;
           }
-
-          (end, len)
-        } else {
-          continue;
         }
+        Some(&b']') => (start_offset, None),
+        _ => continue,
       }
-      _ => continue,
+    } else {
+      (start_offset, None)
     };
 
     let replacer = replacer.get(len);
 
-    result.push_str(&pattern[ending..start]);
+    result.push_str(&pattern[last_end..start]);
     result.push_str(replacer.as_ref());
 
-    ending = end + 1;
+    last_end = end + 1;
   }
 
-  if ending < pattern.len() {
-    result.push_str(&pattern[ending..]);
+  if last_end < pattern.len() {
+    result.push_str(&pattern[last_end..]);
   }
 
   result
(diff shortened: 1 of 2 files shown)


IWANABETHATGUY:
Could you split the bug fixing and perf next time?

IWANABETHATGUY:
It makes it hard to recognize which part is bug-fixing, and which part is performance improvement.

crates/rolldown_utils/src/replace_all_placeholder.rs
```diff
@@ -23,76 +23,98 @@ where
...
-    let (end, len) = match pat_temp.as_bytes().first() {
-      Some(b']') => (start_offset, None),
-      Some(b':') => {
-        if let Some(end) = pat_temp.find(']') {
-          let end = start_offset + end;
-          let len = pattern[start_offset + 1..end].parse::<usize>().ok();
-
-          if len.is_none() {
+    let (end, len) = if is_len_enabled {
+      let rest = &pattern[start_offset..];
```
shulaoda, line 72: I introduced `replace_all` and `replace_all_with_len`. For `[name]` and `[ext]`, we should not replace `[ext:6]` as well.

Therefore, we can directly use `match_indices` to search for `[name]` instead of finding `]` after matching `[name`, which improves performance.

The previous implementation could lead to potential performance issues. See https://github.com/web-infra-dev/rspack/pull/9304 for details.

@IWANABETHATGUY cc

IWANABETHATGUY: Actually using js bench framework, it could be counted as noise if the result fluctuates within 5%

IWANABETHATGUY: I don't know the related part of rspack, but this replace procedure only processes filenames for us, so I don't think we would have performance issues.

shulaoda: Yes, I agree with that.

However, after running multiple ecosystem benchmarks in `rspack`, I found that in 99% of cases, the change resulted in a performance regression for a specific scenario. When reverted, the performance of that specific case improved.

Therefore, while this function may cause performance overhead when applied to very long strings, the impact on our current use case is likely negligible.

shulaoda: So, should we allow `[name]` and `[ext]` to also replace `[name:5]` and `[ext:3]`?

IWANABETHATGUY: We should align with rollup

crates/rolldown_utils/src/replace_all_placeholder.rs
```diff
@@ -23,76 +23,98 @@ where
...
 }
 
 #[test]
 fn test_replace_all_placeholder() {
-  let result = replace_all_placeholder(
-    "hello-[hash]-[hash:-]-[hash_name]-[hash:1]-[hash:].js".to_string(),
-    "[hash]",
-    "abc",
-  );
+  let result = "hello-[hash]-[hash_name]-[hash:1].js".to_string().replace_all("[hash]", "abc");
```
IWANABETHATGUY, line 108: for such a pattern: 

`assetFileNames: "hello-[hash]-[hash:1].js"`

filename: `res.js`

rollup generate: hello-C5dNXyO6-C.js

So I think the result of new testcase is wrong.

crates/rolldown_utils/src/replace_all_placeholder.rs
shulaoda: `Rollup` only supports replacements for the `hash` type, so the PR aligns the behavior accordingly.

https://github.com/rollup/rollup/blob/caa406ec1f7662922d2449ea18816ce42f711c2c/src/utils/renderNamePattern.ts#L21-L27

```js
if (!replacements.hasOwnProperty(type) || (size && type !== 'hash')) {
	return error(
		logFailedValidation(
			`"[${type}${size || ''}]" is not a valid placeholder in the "${patternName}" pattern.`
		)
	);
}
```

crates/rolldown_utils/src/replace_all_placeholder.rs
shulaoda: The behavior of `replace_all` and `replace_all_with_len` is different. `replace_all` does not replace `[hash:8]`, while `replace_all_with_len` does. I added a test for `replace_all` to ensure their behaviors remain distinct.

IWANABETHATGUY: This is still not correct, use https://github.com/rolldown/rolldown/blob/6aa34970186e934c99f323bd6b257f0a84a3c2a9/examples/basic-vue/rolldown.config.js?plain=1#L1-L34 as an example, 

we don't apply the length `[hash:1]` compare to rollup replaced with only one char.
Our replacement: **hello-Cmvkivlj-Cmvkiv.res**

rollup: **hello-C5dNXyO6-C.res**

YOu could ignore the hash diff.

IWANABETHATGUY: I should pass the modification:
```js

export default defineConfig({
  input: "./index.js",
  resolve: {
    // This needs to be explicitly set for now because oxc resolver doesn't
    // assume default exports conditions. Rolldown will ship with a default that
    // aligns with Vite in the future.
    conditionNames: ["import"],
  },
  output: {
    dir: "dist",
    assetFileNames: "hello-[hash]-[hash:1].[name]",
		plugins: [
      {
        name: "test-plugin",
        outputOptions: function (options) {
          options.banner = "/* banner */";
          return options;
        },
        renderChunk(code) {
          this.emitFile({
            type: "asset",
            name: "res.js",
            source: code,
          });
        },
      },
    ],
  },
  experimental: {
    enableComposingJsPlugins: true,
  },
});
```

shulaoda: I see what you mean. This isn't actually an issue with the functionality, but rather due to the length restriction applied elsewhere in the code.

For example, the `hash_placeholder_generator` by default generates a string with a minimum length of `6` characters, but we haven't applied any length truncation to the generated `hash`.
https://github.com/rolldown/rolldown/blob/4a73e66fafdf7cf31854686736fb6fa23642102a/crates/rolldown_common/src/chunk/mod.rs#L212-L220

For instance, we by default generate a `hash` with at least `6` characters, which was originally aligned with the initial implementation. When I made improvements earlier, I did not alter their behavior.
https://github.com/rolldown/rolldown/blob/4a73e66fafdf7cf31854686736fb6fa23642102a/crates/rolldown_common/src/file_emitter.rs#L187-L191

shulaoda: So, this is incorrect behavior, and we should fix it in the next PR?

IWANABETHATGUY (approved):

crates/rolldown_utils/src/replace_all_placeholder.rs
IWANABETHATGUY: Yeah, let's fix in next pr, can you open an issue to track it?

shulaoda: I'm curious, how did you generate it? 🤔
image

shulaoda: > Yeah, let's fix in next pr, can you open an issue to track it?

Of course, if this is truly the right behavior.

IWANABETHATGUY:
```js
// cSpell:disable
import { defineConfig } from "rollup";
import commonjs from "@rollup/plugin-commonjs";
import json from "@rollup/plugin-json";

export default defineConfig({
	input: ["./src/entry.js"],
	output: {
		dir: "dist",
    // file: "hello-[hash]-[hash_name]-[hash:1].js",
    assetFileNames: "hello-[hash]-[hash:1].[name]"
	},
	plugins: [
		commonjs(),
		{
			name: "custom-json",
			load(id) {
				if (id.endsWith(".json")) {
					return ('export default {200: "ok"}');
				}
			},
      renderChunk(code) {
        this.emitFile({
          type: "asset",
          name: "res.js",
          source: code,
        })
      }
		},
		// json(),
	],
});
```

shulaoda: It might be because it only applies to `assetFileNames`. I'll open an issue to track this.

PR #3740, opened 2025-03-02
fix(rust): generate specified hash length for emitted asset filenames

### LINKED ISSUE #3737: [Bug]: Should generate the correct `hash` placeholder of the specified length.
### Description

              This is still not correct, use https://github.com/rolldown/rolldown/blob/6aa34970186e934c99f323bd6b257f0a84a3c2a9/examples/basic-vue/rolldown.config.js?plain=1#L1-L34 as an example, 

we don't apply the length `[hash:1]` compare to rollup replaced with only one char.
Our replacement: **hello-Cmvkivlj-Cmvkiv.res**

rollup: **hello-C5dNXyO6-C.res**

YOu could ignore the hash diff.

_Originally posted by @IWANABETHATGUY in https://github.com/rolldown/rolldown/pull/3736#discussion_r1976666715_
            

Repo: https://stackblitz.com/edit/github-nmlcmage?file=package.json,rolldown.config.mjs,src%2Fcube.css,src%2Fentry.js

### What is expected?

We should generate `hello-B_SY1GJM-B.res` and `name-l0sNRN-l0sNRN.js`.

For `entryFileNames`, it seems that `rollup` can only generate a hash with a length of `6` characters and does not support multiple hashes with specified lengths.

```js
{
  input: './src/entry.js',
  output: {
    dir: './dist',
    entryFileNames: 'name-[hash:6]-[hash:2].js',
    assetFileNames: 'hello-[hash]-[hash:1].[name]',
  },
  plugins: [
    {
      name: 'test',
      renderChunk(code) {
        this.emitFile({
          type: 'asset',
          name: 'res.js',
          source: code,
        });
      },
    },
  ],
}
```

Since we haven't handled the `hash` length in `entryFileNames`, it causes a debug error in `hash_placeholder_generator`.

Under normal circumstances, we only generate hashes with a minimum length of `6` characters.

```bash
thread '<unnamed>' panicked at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/alloc/src/raw_vec.rs:552:17:
capacity overflow
worker (tid = 43) sent an error! unreachable
[Error [RuntimeError]: unreachable]
```

### Any additional comments?

Should we align the hash behavior for all filenames with that of `assetFileNames`?

shulaoda:
### Description

closes #3737

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/mod.rs
+++ b/crates/rolldown/src/stages/generate_stage/mod.rs
@@ -282,7 +282,7 @@ impl<'a> GenerateStage<'a> {
           let mut hash_placeholder = has_hash_pattern.then_some(vec![]);
           let hash_replacer = has_hash_pattern.then_some({
             |len: Option<usize>| {
-              let hash = hash_placeholder_generator.generate(len.map_or(8, |len| len.max(6)));
+              let hash = hash_placeholder_generator.generate(len.map_or(8, |len| len.clamp(6, 21)));
               if let Some(hash_placeholder) = hash_placeholder.as_mut() {
                 hash_placeholder.push(hash.clone());
               }
--- a/crates/rolldown/src/stages/generate_stage/mod.rs
+++ b/crates/rolldown/src/stages/generate_stage/mod.rs
@@ -282,9 +282,7 @@ impl<'a> GenerateStage<'a> {
           let mut hash_placeholder = has_hash_pattern.then_some(vec![]);
           let hash_replacer = has_hash_pattern.then_some({
             |len: Option<usize>| {
-              // Ensure the generated hash length is within the valid range (6-21).
-              // If `len` is `None`, default to 8.
-              let hash = hash_placeholder_generator.generate(len.map_or(8, |len| len.clamp(6, 21)));
+              let hash = hash_placeholder_generator.generate(len);
               if let Some(hash_placeholder) = hash_placeholder.as_mut() {
                 hash_placeholder.push(hash.clone());
               }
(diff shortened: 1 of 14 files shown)


crates/rolldown/src/stages/generate_stage/mod.rs
```diff
@@ -282,7 +282,7 @@ impl<'a> GenerateStage<'a> {
           let mut hash_placeholder = has_hash_pattern.then_some(vec![]);
           let hash_replacer = has_hash_pattern.then_some({
             |len: Option<usize>| {
-              let hash = hash_placeholder_generator.generate(len.unwrap_or(8));
+              let hash = hash_placeholder_generator.generate(len.map_or(8, |len| len.clamp(6, 21)));
```
underfin: Could you add a comment for 6-21?

shulaoda: Currently, we apply `clamp` externally to `hash_placeholder_generator` to ensure it meets the `debug` requirements. Should we move this logic inside the function instead? @hyf0 cc

underfin: I think it could be moved, The logic could be clear.

crates/rolldown_utils/src/hash_placeholder.rs
```diff
@@ -84,8 +85,10 @@ pub struct HashPlaceholderGenerator {
 }
 
 impl HashPlaceholderGenerator {
-  pub fn generate(&mut self, len: usize) -> String {
-    debug_assert!((HASH_PLACEHOLDER_OVERHEAD..=MAX_HASH_SIZE).contains(&len));
+  pub fn generate(&mut self, len: Option<usize>) -> String {
+    // Ensure the generated hash length is within the valid range (6-21).
+    // If `len` is `None`, default to 8.
+    let len = len.map_or(DEFAULT_HASH_SIZE, |len| len.clamp(MIN_HASH_SIZE, MAX_HASH_SIZE));
```
shulaoda, line 91: Since all current usage cases follow the pattern `len.map_or(8, |len| len.clamp(6, 21))`, I have adjusted the `generate` function accordingly.

underfin (approved):

Issue #3697, opened 2025-02-27
[Bug]: Error using configuration option `watch.chokidar`


hotdogc1017:
### Reproduction link or steps

https://stackblitz.com/edit/github-hxpadn9k?file=rolldown.config.mjs

### What is expected?

Instead of reporting a bad report, give a warning or remove this configuration option directly

### What is actually happening?

When I add `watch.chokidar` to the configuration options, running the build reports an error and prompts:
```text
Error: UNSUPPORTED: The watch chokidar option is deprecated, please use notify options instead of it.
```

### System Info

```Shell
System:
    OS: Linux 6.8 Ubuntu 24.04.1 LTS 24.04.1 LTS (Noble Numbat)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
    Memory: 9.44 GB / 23.16 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.0.0 - ~/.nvm/versions/node/v22.0.0/bin/node
    Yarn: 1.22.22 - ~/.local/share/pnpm/yarn
    npm: 10.5.1 - ~/.nvm/versions/node/v22.0.0/bin/npm
    pnpm: 9.8.0 - ~/桌面/nextui-vue/node_modules/.bin/pnpm
    bun: 1.1.21 - ~/.bun/bin/bun
  Browsers:
    Chrome: 133.0.6943.53
  npmPackages:
    rolldown: 1.0.0-beta.3 => 1.0.0-beta.3
```

jrmajor:
> give a warning or remove this configuration option directly

@underfin Do I understand correctly that we just want to print a warning instead of throwing an error?

underfin:
> [@underfin](https://github.com/underfin) Do I understand correctly that we just want to print a warning instead of throwing an error?

Yeah.

underfin:
After we discuss, the `watch.chokidar` check is moved at validate config phase, it will give an error for it. If you want to turn off the behavior of validation config, you could using `ROLLDOWN_OPTIONS_VALIDATION=loose`.

PR #3748, opened 2025-03-03
feat: validate option

underfin:
### Description
related https://github.com/rolldown/rolldown/issues/3697


hyf0 (approved):
Should we also need to have a `ROLLDOWN_OPTIONS_VALIDATION = 'loose' | 'strict'` to allow users escape from strict options validation? @IWANABETHATGUY cc

PR #3749, opened 2025-03-03
fix(node): allow updating sourcemap.debugId by plugins

sapphi-red:
### Description

Allow updating sourcemap.debugId and sourcemap.x_google_ignoreList by plugins.

It is used here in Vite:
https://github.com/rolldown/vite/blob/40a300ab92acf2548bfeb86b7d720026da8657a8/packages/vite/src/node/plugins/importAnalysisBuild.ts#L722-L736

### DIFF
--- a/crates/rolldown_binding/src/types/binding_sourcemap.rs
+++ b/crates/rolldown_binding/src/types/binding_sourcemap.rs
@@ -27,13 +27,16 @@ pub struct BindingJsonSourcemap {
   pub sources: Option<Vec<Option<String>>>,
   pub sources_content: Option<Vec<Option<String>>>,
   pub names: Option<Vec<String>>,
+  pub debug_id: Option<String>,
+  #[napi(js_name = "x_google_ignoreList")]
+  pub x_google_ignore_list: Option<Vec<u32>>,
 }
 
 impl TryFrom<BindingJsonSourcemap> for rolldown_sourcemap::SourceMap {
   type Error = anyhow::Error;
 
   fn try_from(value: BindingJsonSourcemap) -> Result<Self, Self::Error> {
-    rolldown_sourcemap::SourceMap::from_json(rolldown_sourcemap::JSONSourceMap {
+    let mut map = rolldown_sourcemap::SourceMap::from_json(rolldown_sourcemap::JSONSourceMap {
       file: value.file,
       mappings: value.mappings.unwrap_or_default(),
       source_root: value.source_root,
@@ -45,8 +48,12 @@ impl TryFrom<BindingJsonSourcemap> for rolldown_sourcemap::SourceMap {
         .collect(),
       sources_content: value.sources_content,
       names: value.names.unwrap_or_default(),
-      debug_id: None,
+      debug_id: value.debug_id,
     })
-    .map_err(|e| anyhow::format_err!("Convert json sourcemap error: {:?}", e))
+    .map_err(|e| anyhow::format_err!("Convert json sourcemap error: {:?}", e))?;
+    if let Some(ignore_list) = value.x_google_ignore_list {
+      map.set_x_google_ignore_list(ignore_list);
+    }
+    Ok(map)
   }
 }
--- a/packages/rolldown/src/types/rolldown-output.ts
+++ b/packages/rolldown/src/types/rolldown-output.ts
@@ -20,6 +20,8 @@ export interface SourceMap {
   sources: string[]
   sourcesContent: string[]
   version: number
+  debugId?: string
+  x_google_ignoreList?: number[]
   toString(): string
   toUrl(): string
 }
--- a/packages/rolldown/src/types/sourcemap.ts
+++ b/packages/rolldown/src/types/sourcemap.ts
@@ -1,4 +1,5 @@
 import { BindingSourcemap } from '../binding'
+import type { SourceMap } from './rolldown-output'
 
 export interface ExistingRawSourceMap {
   file?: string | null
@@ -14,7 +15,7 @@ export interface ExistingRawSourceMap {
 export type SourceMapInput = ExistingRawSourceMap | string | null
 
 export function bindingifySourcemap(
-  map?: SourceMapInput,
+  map?: SourceMapInput | SourceMap,
 ): undefined | BindingSourcemap {
   if (map == null) return
   return {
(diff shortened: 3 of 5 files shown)


crates/rolldown_binding/src/types/binding_sourcemap.rs
```diff
@@ -45,8 +48,12 @@ impl TryFrom<BindingJsonSourcemap> for rolldown_sourcemap::SourceMap {
...
-      debug_id: None,
+      debug_id: value.debug_id,
     })
-    .map_err(|e| anyhow::format_err!("Convert json sourcemap error: {:?}", e))
+    .map_err(|e| anyhow::format_err!("Convert json sourcemap error: {:?}", e))?;
+    if let Some(ignore_list) = value.x_google_ignore_list {
+      map.set_x_google_ignore_list(ignore_list);
+    }
+    Ok(map)
   }
```
sapphi-red, line 58: This part would be simple after https://github.com/oxc-project/oxc-sourcemap/pull/30 lands.

packages/rolldown/src/types/rolldown-output.ts
```diff
@@ -20,6 +20,8 @@ export interface SourceMap {
   sources: string[]
   sourcesContent: string[]
   version: number
+  debugId?: string
+  x_google_ignoreList?: number[]
```
sapphi-red, line 24: Rollup does not have this field here, but I think it makes sense to have it here as well.

crates/rolldown_binding/src/types/binding_sourcemap.rs
Boshen: oxc_sourcemap = "3"

sapphi-red: Thanks! But it seems I need https://github.com/oxc-project/oxc/pull/9507 as well.
```
   --> crates\rolldown_ecmascript\src\ecma_compiler.rs:130:16
    |
130 |     (ret.code, ret.map)
    |                ^^^^^^^ expected `SourceMap`, found `oxc_sourcemap::sourcemap::SourceMap`
    |
note: two different versions of crate `oxc_sourcemap` are being used; two types coming from two different versions of the same crate are different types even if they look the same
   --> C:\Users\green\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\oxc_sourcemap-3.0.0\src\sourcemap.rs:12:1
    |
12  | pub struct SourceMap {
    | ^^^^^^^^^^^^^^^^^^^^ this is the expected type `SourceMap`
    |
   ::: C:\Users\green\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\oxc_sourcemap-2.0.2\src\sourcemap.rs:12:1
    |
12  | pub struct SourceMap {
    | ^^^^^^^^^^^^^^^^^^^^ this is the found type `oxc_sourcemap::sourcemap::SourceMap`
    |
   ::: crates\rolldown_ecmascript\src\ecma_compiler.rs:15:5
    |
15  | use oxc_sourcemap::SourceMap;
    |     ------------- one version of crate `oxc_sourcemap` used here, as a direct dependency of the current crate
    |
   ::: crates\rolldown_ecmascript\src\ecma_ast\mod.rs:5:5
    |
5   | use oxc::{
    |     --- one version of crate `oxc_sourcemap` used here, as a dependency of crate `oxc_codegen`
    = help: you can use `cargo tree` to explore your dependency tree
```

Boshen: Ok, wait for my oxc release.

packages/rolldown/src/types/sourcemap.ts
```diff
@@ -31,10 +32,16 @@ export function bindingifySourcemap(
             // we convert it to undefined to skip that error.
             // note that if `sourceRoot: null` is included in a string sourcemap,
             // it will be converted to None by serde-json.
-            sourceRoot: map.sourceRoot ?? undefined,
+            sourceRoot:
+              'sourceRoot' in map ? (map.sourceRoot ?? undefined) : undefined,
```
IWANABETHATGUY, line 36: Curious why new version is better than the old one?

IWANABETHATGUY: Which One Should You Use?
✅ Use the first approach ('sourceRoot' in the map ? ...) if the map might be a Proxy or has unusual property behavior.
✅ The second approach is simpler and generally safe if the map is a plain object.
If you're dealing with a standard JavaScript object, both approaches will mostly behave the same. 🚀

This is the answer to ChatGPT, but It seems our `map` will not be a proxy?

sapphi-red: These `foo in map` exists because
- `sourceRoot` field only exists in `SourceMapInput` type
- `debugId` field only exists in `SourceMap` type

(`x_google_ignoreList` was an artifact, removed)

IWANABETHATGUY (approved):

packages/rolldown/src/types/sourcemap.ts
IWANABETHATGUY: ah, thanks for explaining.

sapphi-red: https://www.typescriptlang.org/play/?#code/C4TwDgpgBAglC8UDeUBmB7dAuKBnYATgJYB2A5gDRQBGAhgTvseVAL4BQokUAQgsjXqNCpSoIBew5mTbtO4aAGF+cAD685AEwgBjADb1oO9CXxQdORXJ0A6DOna26BRzbri5Acnueopc1AA-OZ2mFA4JACuenrsnu6+-jpBIe7hUFEx7EA

IWANABETHATGUY: > https://www.typescriptlang.org/play/?#code/C4TwDgpgBAglC8UDeUBmB7dAuKBnYATgJYB2A5gDRQBGAhgTvseVAL4BQokUAQgsjXqNCpSoIBew5mTbtO4aAGF+cAD685AEwgBjADb1oO9CXxQdORXJ0A6DOna26BRzbri5Acnueopc1AA-OZ2mFA4JACuenrsnu6+-jpBIe7hUFEx7EA

haha, I don't open the code editor to review the code, now I get it.

PR #3750, opened 2025-03-03
fix: load mts config without type module

YangJonghun:
### Description
If developer forces module type via specifing TypeScript extension,
Make JS extension follow TS extension,
Node.js read file's module correctly even if it doesn't match "type" field within package.json

This PR can resolve below Node.js warning (when use `rolldown.config.mts`)
```
(node:885) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of file:///<path_name>/rolldown.config.[hash].js is not specified and it doesn't parse as CommonJS.
Reparsing as ES module because module syntax was detected. This incurs a performance overhead.
To eliminate this warning, add "type": "module" to <path_name>/package.json.
(Use `node --trace-warnings ...` to show where the warning was created)
```


underfin:
Could you give a reproduction for it, I'm not sure what happened at here.

IWANABETHATGUY:
Can not reproduce it either, 
image

Joery-M:
I believe I managed to reproduce:
Remove `"type": "module"` from package.json, and change the config file extension from `.[cm]ts` to just `.ts` (or `.js`).
https://stackblitz.com/edit/github-ja4hyvvr?file=package.json

In stackblitz this errors out since it uses node 18 (tries to parse as cjs), but the exact warning can be reproduced when using node v22.14 (not specifically, just what I had on hand)
```
$ nr build

> build
> rolldown -c ./rolldown.config.js

(node:22952) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of file:///<folder>/rolldown.config.js is not specified and it doesn't parse as CommonJS.
Reparsing as ES module because module syntax was detected. This incurs a performance overhead.
To eliminate this warning, add "type": "module" to <folder>\package.json.
(Use `node --trace-warnings ...` to show where the warning was created)
<DIR>/other-entry.js    chunk │ size: 0.11 kB
<DIR>/cube-DoVzd1lL.js  chunk │ size: 0.17 kB
<DIR>/entry.js          chunk │ size: 0.20 kB

✔ Finished in 18.54 ms
```

Do note that I don't believe this PR fixes the warning, but I could be mistaken in that regard.

YangJonghun:
Reproduction is the same as @Joery-M  mentioned.
Remove "type" field from package.json, and set config file to `.mts` extension, also it reproduces on node v22.14.0

This PR aim for no `"type": "module"` declaration and only removes Node.js warning when using `.mts` file as configuration.

Regardless of the value of the "type" field, .mjs files are always treated as ES modules and .cjs files are always treated as CommonJS.

If `.ts` extension is used without filling out "type" field, developer has not explicitly specified module format, so it is correct behavior for Node.js to output warning

underfin:
Thanks for your description. I updated the pr to add the tests, and also fixed load the cts error, it need to compiled the cts to cjs.

underfin (approved):

packages/rolldown/src/cli/load-config.ts
```diff
@@ -76,14 +81,57 @@ async function findConfigFileNameInCwd(): Promise<string> {
...
 
+export function isFilePathESM(filePath: string): boolean {
+  if (/\.m[jt]s$/.test(filePath)) {
+    return true
+  } else if (/\.c[jt]s$/.test(filePath)) {
+    return false
+  } else {
+    // check package.json for type: "module"
+    try {
+      const pkg = findNearestPackageData(path.dirname(filePath))
```
IWANABETHATGUY: There is no need to try/catch again, since you already ignore error in `findNearestPackageData`

underfin: Fixed it.

PR #3758, opened 2025-03-04
test(node): remove unnecessary footer test

AliceLanniste:
### Description

Description
Related to https://github.com/rolldown/rolldown/issues/3256


IWANABETHATGUY:
Can you pass the CI?

packages/rolldown/tests/fixtures/function/footer/string/_config.ts
```diff
@@ -0,0 +1,15 @@
+import { expect } from 'vitest'
+import { defineTest } from 'rolldown-tests'
+
+const footerTxt = '// footer test\n'
```
IWANABETHATGUY: This did not need to be moved, since it is not function

AliceLanniste: Ok, I will keep output/footer/string file

IWANABETHATGUY (approved):

PR #3761, opened 2025-03-04
test(node): improve tests related to customized output filenames

shulaoda:
### Description

Related to #3256 

I have consolidated all `rust-side` tests related to output `filenames` and removed unnecessary tests on the `node` side.


crates/rolldown/tests/rolldown/function/filenames/basic/_test.mjs
```diff
@@ -0,0 +1,21 @@
+// Since we have customized the output `entry_filenames`, 
+// the `compiled_entries` in the following context is incorrect.
+//
+// /crates/rolldown_testing/src/integration_test.rs#L471-L484
```
shulaoda, line 4: We need to add a `_test.mjs` file to prevent test failures caused by the inability to correctly recognize the entry after customizing the output `entry_filenames`.

IWANABETHATGUY:
I have some different opinions about the test migration, need to discuss it in the next meeting.

IWANABETHATGUY:
I update the principle about testing, https://github.com/rolldown/rolldown/pull/3856

Boshen:
Did we have a consensus for this?

PR #3767, opened 2025-03-05
chore(rolldown): oxc v0.56.0

### LINKED ISSUE #3201: [Feature Request]: pure annotations

When publishing a library, pure annotations such as `/*#__PURE__*/` help minifiers know which function calls are free of side effects.

- [terser docs](https://terser.org/docs/miscellaneous/#annotations)
- [babel docs](https://babeljs.io/docs/babel-preset-react#pure)
- [babel jsx output example](https://babeljs.io/repl#?browsers=&build=&builtIns=false&corejs=false&spec=false&loose=false&code_lz=DwEwlgbgBA9AfEA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=true&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Ctypescript&prettier=true&targets=&version=7.26.4&externalPlugins=&assumptions=%7B%7D)

With the input:
```jsx
export default <div />;
```
Here's the rolldown output:
```js
import { jsx } from "react/jsx-runtime";

//#region src/index.tsx
var src_default = jsx("div", {});

//#endregion
export { src_default as default };
```
I'd expect a `PURE` annotation like the babel output example above.

I'm not sure if this is an issue with Rolldown itself or an underlying library.

I'm not sure. It could be the default for non-minified output.

Boshen:
* Adds `/* @__PURE__ */` annotation to react transform
* Applies `/* @__NO_SIDE_EFFECTS */` in the minifier

closes https://github.com/rolldown/rolldown/issues/3201


IWANABETHATGUY:
https://github.com/rolldown/rolldown/issues/3201 this issue  could be closed.

Boshen:
Raw transfer breaks in wasi. Will fix.

IWANABETHATGUY (approved):

Issue #898, opened 2024-04-17
Improve WASI support


hyf0:
This is a followed up of #832.

With the hard work of @Brooooooklyn, we now are ok to compile to wasi successfully and run wasi rolldown successfully in some simple cases.

However, there are still a lot of hard and wired problems to be solved.

As #894 showed
  - In unbuntu
    - The wasi rolldown could compile the basic example successfully
    - Can't pass tests due to being hang while running
  - In windows/macos
    - `Unsupported Operation` error was throwed in rust even in the basic example case

Other problems

1.
https://github.com/rolldown/rolldown/blob/7b64957632f0942ff69dac7899f74f8d29141b6d/.github/workflows/reusable-wasi-test.yml#L43-L47

2.
https://github.com/rolldown/rolldown/blob/7b64957632f0942ff69dac7899f74f8d29141b6d/crates/rolldown_binding/src/bundler.rs#L35-L37

3.
> Update: oh no! When set RUST_LOG=trace, pnpm --filter=@example/basic build success, fails if unset RUST_LOG, so strange! https://github.com/rolldown/rolldown/issues/898#issuecomment-2061370702

4. Running a playground on the browser is blocked by: https://github.com/tokio-rs/tokio/pull/6510. Because the `fs` polyfill crosses the main thread/ web worker is using [`Atomic.wait`](https://github.com/napi-rs/napi-rs/blob/%40napi-rs/cli%403.0.0-alpha.54/wasm-runtime/fs-proxy.cjs#L143), any `std::fs` operations within the `async fn` will block `tokio` `current_thread` runtime forever.

toyobayashi:
@Brooooooklyn @hyf0 

I can build successfully on Windows by manually make following changes:

`examples/basic-vue/rolldown.config.js`

```diff
  import { defineConfig } from 'rolldown'
  
  export default defineConfig({
    input: './index.js',
+   // set posix path for WASI, not native platform path
+   // map to C:\\Users\\toyobayashi\\Projects\\rolldown\\examples\\basic-vue
+   cwd: '/Users/toyobayashi/Projects/rolldown/examples/basic-vue',
    resolve: {
      // This needs to be explicitly set for now because oxc resolver doesn't
      // assume default exports conditions. Rolldown will ship with a default that
      // aligns with Vite in the future.
      conditionNames: ['import'],
    },
  })
```

`packages/rolldown/dist/shared/rolldown-binding.wasi.cjs`:

```diff
[...]

PR #3788, opened 2025-03-08
feat(wasi): support wasi watch mode

suxin2017:
### Description

block by https://github.com/rolldown/rolldown/issues/898#issuecomment-2708126470

improve WASI support

#898


Cargo.toml
```diff
@@ -203,3 +204,6 @@ debug         = false     # Set to `true` for debug information
 lto           = "fat"
 opt-level     = 3
 strip         = "symbols" # Set to `false` for debug information
+
+[patch.crates-io]
+filetime = { git = "https://github.com/oligamiq/filetime.git" }
```
suxin2017: We should find a stable tag patch to prevent security issues

Cargo.toml
```diff
@@ -167,6 +167,7 @@ string_wizard       = { path = "./crates/string_wizard", features = ["serde"] }
 sugar_path          = { version = "1.2.0", features = ["cached_current_dir"] }
 testing_macros      = "1.0.0"
 tokio               = { version = "1.38.0", default-features = false }
+tokio_with_wasm     = { version = "0.8.2", features = ["rt"] }
```
Brooooooklyn: why need this

suxin2017: tokio_with_wasm code
https://github.com/cunarist/tokio-with-wasm/blob/180d56afc4f1e030b6116ba409de5c0c71ad0e73/tokio_with_wasm/src/glue/task/pool.rs#L75C4-L124C6

It uses webwork to support multiple spawn, and it turns out that one of the reasons wasi doesn't work is because it gets blocked, as you can see in this demo project that I'm recreating
https://github.com/rolldown/rolldown/issues/898#issuecomment-2708132359

IWANABETHATGUY:
We need also test bundling performance with stackblitz cc @underfin

Cargo.toml
suxin2017: Do we have to manage this by ourselves? This is someone else's code repository

IWANABETHATGUY:
https://stackblitz.com/~/github.com/hyf0/stackblitz-rolldown-1000

IWANABETHATGUY:
image
This branch crashed on this case, it works fine before.

Cargo.toml
Brooooooklyn: The wasi is blocked due to we are not using `model=reactor`, see https://github.com/rust-lang/rust/issues/137510#issuecomment-2683759021 for the context.
I fixed this in https://github.com/napi-rs/napi-rs/pull/2480, upgrade napi-build can apply this fix.

Use `tokio_with_wasm` is meaningless here because:
1. we are target to `wasm32-waspi1-threads` which has threads in our wasi polyfill: https://toyobayashi.github.io/emnapi-docs/guide/multithreaded-async.html#initialization
2. `tokio_with_wasm` requires `nightly` toolchain and needs pass `-Z build-std=std,panic_abort` and `RUSTFLAGS="-C target-feature=+atomics,+bulk-memory,+mutable-globals"` to rust but we are using stable Rust.

Cargo.toml
Brooooooklyn: In https://github.com/alexcrichton/filetime/issues/109#issuecomment-2708142283:

> My commits is this, but it's old so I forgot it and no test.

we need to test it and contribute to `filetime` crate, the tests can be done via wasmtime like we've done in oxc: https://github.com/oxc-project/oxc/blob/main/.github/workflows/ci.yml#L99

Cargo.toml
suxin2017: Experts are experts, awesome!

suxin2017: image
image
image

Is there something wrong with my configuration? Why is it still blocked?

https://github.com/suxin2017/napi-bind

suxin2017:
I'm so sorry. I thought we solved the problem. I'm so sorry for the delay

Cargo.toml
Brooooooklyn: You can't call `tokio::spawn` on the Node.js main thread, it will call [`memory.atomic.wait32 instruction`](https://webassembly.github.io/threads/core/syntax/instructions.html#syntax-instr-atomic-memory) under the hood and block the main thread forever. You can use `std::thread::spawn` + `tokio::block_on` as a workaround like this: https://github.com/napi-rs/napi-rs/blob/main/crates/napi/src/tokio_runtime.rs#L216-L221

Note, this limitation can be resolved by using `MultiThread` tokio `Runtime` on wasm32, I'm working on this in https://github.com/napi-rs/napi-rs/pull/2501

suxin2017:
@underfin Can you preview it again?

suxin2017:
Preview this and give it a try? @underfin

underfin:
I tested it agian, it also is oom at https://stackblitz.com/~/github.com/hyf0/stackblitz-rolldown-1000. Look like it create many  tokio runtime caused oom at module_loader.rs. 

I remember before the watcher build is hang at module_loader spawn task, https://github.com/rolldown/rolldown/issues/898#issuecomment-2708132359 it is root reason block it?

I also test watcher at https://stackblitz.com/edit/github-wzayskrd?file=src%2Ffoo.ts,src%2Findex.ts,rolldown.config.mjs,package.json,  it only work twice and the watcher received change is slow. And it panic at after the third time changes. Here is log 
```
thread '<unnamed>' panicked at library/std/src/sys/pal/wasi/thread.rs:169:26:
thread::sleep(): unexpected result of poll_oneoff
worker (tid = 50) sent an error! unreachable
```

suxin2017:
> I tested it agian, it also is oom at https://stackblitz.com/~/github.com/hyf0/stackblitz-rolldown-1000. Look like it create many tokio runtime caused oom at module_loader.rs.
> 
> I remember before the watcher build is hang at module_loader spawn task, [#898 (comment)](https://github.com/rolldown/rolldown/issues/898#issuecomment-2708132359) it is root reason block it?
> 
> I also test watcher at https://stackblitz.com/edit/github-wzayskrd?file=src%2Ffoo.ts,src%2Findex.ts,rolldown.config.mjs,package.json, it only work twice and the watcher received change is slow. And it panic at after the third time changes. Here is log
> 
> ```
> thread '<unnamed>' panicked at library/std/src/sys/pal/wasi/thread.rs:169:26:
> thread::sleep(): unexpected result of poll_oneoff
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> worker (tid = 50) sent an error! unreachable
> ```

It turns out that this seems to work, but there are still some problems. The reason for the slow observation of the file is that when compiling to wasm, notify uses poll to monitor the file changes

suxin2017:
image
One last question, occasional memory access out of bounds

PR #3823, opened 2025-03-11
refactor(rust): simplify `keep_name_statement_to_insert`

shulaoda:
### Description

Since `SymbolId` is currently unnecessary, we can remove it along with its related logic.

### DIFF
--- a/crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
+++ b/crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
@@ -237,20 +237,19 @@ impl<'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'_, 'ast> {
 
   fn visit_statements(&mut self, it: &mut allocator::Vec<'ast, ast::Statement<'ast>>) {
     let previous_stmt_index = self.ctx.cur_stmt_index;
-    let mut previous_keep_name_statement =
-      std::mem::take(&mut self.ctx.keep_name_statement_to_insert);
-
+    let previous_keep_name_statement = std::mem::take(&mut self.ctx.keep_name_statement_to_insert);
     for (i, stmt) in it.iter_mut().enumerate() {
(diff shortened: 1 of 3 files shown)


crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
```diff
@@ -237,20 +237,20 @@ impl<'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'_, 'ast> {
...
-    for (stmt_index, _symbol_id, original_name, new_name) in
-      self.ctx.keep_name_statement_to_insert.iter().rev()
-    {
-      it.insert(*stmt_index, self.snippet.keep_name_call_expr_stmt(original_name, new_name));
+    std::mem::swap(&mut self.ctx.keep_name_statement_to_insert, &mut previous_keep_name_statement);
+    for (stmt_index, original_name, new_name) in previous_keep_name_statement.into_iter().rev() {
+      it.insert(stmt_index, self.snippet.keep_name_call_expr_stmt(&original_name, &new_name));
     }
     self.ctx.cur_stmt_index = previous_stmt_index;
-    self.ctx.keep_name_statement_to_insert = previous_keep_name_statement;
```
IWANABETHATGUY: I think the code before is more readable.

shulaoda: I'm wondering if this change can bring a real performance improvement, as the benchmarks I ran locally were quite unstable. After verifying it, I will revert the change.

IWANABETHATGUY (approved):

crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
IWANABETHATGUY: Don't see any opportunity to improve performance, both are passing reference,s and no clones here.

Issue #3806, opened 2025-03-10
[Feature Request]: Support `external: 'absolute'` in the `resolveId` plugin hook


lansi951:
### What problem does this feature solve?

From Nuxt 3.16, use `external: 'absolute'` in the `resolveId` hook.
https://github.com/nuxt/nuxt/blob/8e33eca57817660b2d3b5941d36bfefd9cedaab4/packages/nuxt/src/core/plugins/resolved-externals.ts#L37

Can't build Nuxt because Rolldown doesn't support it yet.

### What does the proposed API look like?

https://rollupjs.org/plugin-development/#resolveid

underfin:
closed at https://github.com/rolldown/rolldown/pull/3834

PR #3834, opened 2025-03-13
feat: support absoulte/relative external

underfin:
### Description

related https://github.com/rolldown/rolldown/issues/3806. Here enable related rollup tests to test it.


crates/rolldown/tests/esbuild/default/external_module_exclusion_relative_path/artifacts.snap
```diff
@@ -1,15 +1,14 @@
...
 
 ## index.js
 
 ```js
-import foo from "./foo.js";
-import out from "../../../out/in-out-dir.js";
-import sha256 from "../../sha256.min.js";
+import foo from "./nested/folder/foo.js";
+import out from "../out/in-out-dir.js";
+import sha256 from "./sha256.min.js";
```
underfin, line 11: The relative path external module rendered path is not finished, i will resolve it at next.

IWANABETHATGUY: Can you open an issue to track it ?

crates/rolldown/src/module_loader/module_loader.rs
```diff
@@ -186,13 +190,32 @@ impl ModuleLoader {
...
             SymbolRefDbForModule::new(Scoping::default(), idx, ScopeId::new(0)),
           );
-          let symbol_ref = self.symbol_ref_db.create_facade_root_symbol_ref(
-            idx,
-            legitimize_identifier_name(resolved_id.id.as_str()).as_ref(),
-          );
 
-          let ext =
-            ExternalModule::new(idx, resolved_id.id, external_module_side_effects, symbol_ref);
+          let renormalize_render_path = !matches!(resolved_id.external, ResolvedExternal::Absolute)
```
IWANABETHATGUY: A better naming would be more readable, https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members?redirectedfrom=MSDN#names-of-properties, 

NIT: `need_renormalize_render_path` maybe better

IWANABETHATGUY (approved):

crates/rolldown/tests/esbuild/default/external_module_exclusion_relative_path/artifacts.snap
IWANABETHATGUY: alright, since #3806 is still open.

crates/rolldown/src/module_loader/module_loader.rs
underfin: Look good for me.

crates/rolldown/tests/esbuild/default/external_module_exclusion_relative_path/artifacts.snap
underfin: The rolldown resut is already same as rollup, https://stackblitz.com/edit/github-shyenfvo-h8p9hpdz?file=package.json,index.js,nested%2Ffolder%2Ftest.js,rollup.config.js,dist%2Findex.js. 

The esbuild is relative to the current working directory [ref](https://github.com/evanw/esbuild/commit/345f64c1a711e22a96eeea2aae71f7523c8a347a), but the rollup using  the common shared parent directory of all entires modules [ref](https://rollupjs.org/configuration-options/#makeabsoluteexternalsrelative).

underfin: I intend it to follow the esbuild behavior, calculating the common dir of all entires  modules is not a good way, it always is working directory at most cases, it is unnecessary to calculate it. I will have try to it to see the breaked the rollup tests.

Update: using the esbuild behavior make many rollup test failed. so here also using rollup behavior.

PR #3840, opened 2025-03-13
chore(cli): improve the output format of errors

shulaoda:
### Description

Reduce the spacing between the error and the stack trace, and add a blank line between the error summary and its content.
Make the error summary, content, and stack trace clearly hierarchical and distinct.

Before: 
JXRSUCI7MMJWKF@L _7}9MH

After:
WTEHUDYY04UWCT$A@K{2FGS


IWANABETHATGUY:
The white color `Error:` prefix seems redundant

IWANABETHATGUY (approved):

shulaoda:
> The white color `Error:` prefix seems redundant

I agree with this perspective. It’s likely designed with node-side errors in mind, where errors thrown from the rust side would have their own prefixes. Let me think about how to improve it.

PR #3841, opened 2025-03-13
refactor(cli): avoid outputting unnecessary error prefixes

shulaoda:
### Description

              The white color `Error:` prefix seems redundant

_Originally posted by @IWANABETHATGUY in https://github.com/rolldown/rolldown/issues/3840#issuecomment-2720482085_

### DIFF
--- a/packages/rolldown/src/utils/error.ts
+++ b/packages/rolldown/src/utils/error.ts
@@ -42,6 +42,12 @@ export function normalizeErrors(rawErrors: (BindingError | Error)[]): Error {
 }
 
 function getErrorMessage(e: RollupError): string {
+  // If the `kind` field is present, we assume it represents
+  // a custom error defined by rolldown on the rust side.
+  if (Object.hasOwn(e, 'kind')) {
+    return e.message
+  }
+
   let s = ''
   if (e.plugin) {
     s += `[plugin ${e.plugin}]`
--- a/packages/rolldown/tests/error/__snapshots__/error.test.ts.snap
+++ b/packages/rolldown/tests/error/__snapshots__/error.test.ts.snap
@@ -13,12 +13,12 @@ Error: js side error
 exports[`Error output format > should correctly output the custom error defined on the rust side 1`] = `
 "Build failed with 1 error:
 
-[31m[UNRESOLVED_IMPORT] Error:[0m Could not resolve './world' in error.js
-   [38;5;246m╭[0m[38;5;246m─[0m[38;5;246m[[0m error.js:1:19 [38;5;246m][0m
-   [38;5;246m│[0m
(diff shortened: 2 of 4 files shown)


packages/rolldown/src/utils/error.ts
```diff
@@ -21,7 +20,7 @@ export function normalizeErrors(rawErrors: (BindingError | Error)[]): Error {
       summary += '...'
       break
     }
-    summary += getErrorMessage(errors[i])
```
shulaoda, line 24: The `getErrorMessage` function is only used within `normalizeErrors`, and `normalizeErrors` is designed to handle either `napi::Error` thrown from the rust side or custom errors defined by `rolldown`. However, in practice, when `napi` encounters an `napi::Error`, it automatically throws an error itself. This means that `normalizeErrors` effectively only processes `BindingError`. 

As a result, we can safely remove the `getErrorMessage` function since its role is redundant under these circumstances.

IWANABETHATGUY:
Can you provide a demonstration or test of the difference?

shulaoda:
> Can you provide a demonstration or test of the difference?

Yes, I am trying to add new test cases.

packages/rolldown/tests/error/__snapshots__/error.test.ts.snap
```diff
@@ -0,0 +1,24 @@
...
+Error: js side error
+    at PluginContextImpl.buildEnd (D:\\shulaoda\\rolldown\\packages\\rolldown\\tests\\error\\error.test.ts:135:19)
+    at plugin (D:\\shulaoda\\rolldown\\packages\\rolldown\\dist\\shared\\src-DEf1-YJM.mjs:1357:18)
+    at plugin.<computed> (D:\\shulaoda\\rolldown\\packages\\rolldown\\dist\\shared\\src-DEf1-YJM.mjs:1963:18)"
+`;
+
+exports[`should correctly output the custom error defined on the rust side 1`] = `
+"Build failed with 1 error:
+
+[31m[UNRESOLVED_IMPORT] Error:[0m Could not resolve './world' in error.js
```
shulaoda: Previously, there was an `Error: ` prefix before `[UNRESOLVED_IMPORT]`.

IWANABETHATGUY: The Ci will not pass, since all path is absolute, I think it is good enough, let's remove the test if you don't find good way to handle it.

shulaoda: Yes, that's right. I was about to replace these path contents. Your code review was so quick that I hadn't even finished resolving it yet. 😂

shulaoda: @IWANABETHATGUY done

IWANABETHATGUY (approved):

IWANABETHATGUY:
It seem the hash is different in different platform, let's remove the js custom error test and preserve another one

PR #3847, opened 2025-03-13
chore: enable tasks/generator in ci

IWANABETHATGUY:


IWANABETHATGUY:
* **#3847** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3847?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3847?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3846** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3846?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

.github/workflows/ci.yml
```diff
@@ -165,3 +166,9 @@ jobs:
 
       - name: Lint Toml
         run: pnpm lint-toml
+
+      - name: update generated code
+        run: |
+          cargo run --bin generator
```
hyf0, line 172: Would it be better if this process it's part of `cargo build` or `pnpm build`? Is it quite slow?

hyf0 (approved):

.github/workflows/ci.yml
IWANABETHATGUY: It is not slow, but since it needs to compile first, and for now the generated code should not be changed frequently, So I don't want to slow down the normal workflow, Putting it here could ensure the generated code is always up to date, also don't bother developer's normal workflow?

IWANABETHATGUY: We could modify it in the future if we add some generators that change frequently, perhaps like napi binding code getter/setter used for syncing data between Rust and JavaScript.

PR #3848, opened 2025-03-13
fix: generated CSS files contains invalid trailing comment

### LINKED ISSUE #3845: Generated CSS files used wrong symbols to reference to comments
> Looks like rolldown outputs a trailing comment like
> ```
> //# sourceMappingURL=styles.css.map
> ```
> in the CSS file, but `//` comments are not valid in CSS, I believe it should be `/*# sourceMappingURL=styles.css.map */` instead.
> 
> Is it possible to disable sourcemaps for CSS files?
> 
> https://github.com/adazzle/react-data-grid/issues/3736 

 _Originally posted by @nstepien in [#3242](https://github.com/rolldown/rolldown/issues/3242#issuecomment-2721489935)_

nilptr:
### Description

fix #3845

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/render_chunk_to_assets.rs
+++ b/crates/rolldown/src/stages/generate_stage/render_chunk_to_assets.rs
@@ -83,6 +83,7 @@ impl GenerateStage<'_> {
                 &file_dir,
                 rendered_chunk.filename.as_str(),
                 rendered_chunk.debug_id,
+                |comment| format!("\n//# {comment}"),
               )
               .await?;
           }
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY:
Sorry, Maybe I have a better version that is not only used for `sourcemap` reference but also for `debugId`

IWANABETHATGUY:
Maybe you could request for assigning the issue before you get started next time.

nilptr:
ok, sorry for the bug caused

IWANABETHATGUY:
> ok, sorry for the bug caused

Not at all, thank you for your contribution

PR #3855, opened 2025-03-14
feat: calculate relative external module rendered path

underfin:
### Description

relative https://github.com/rolldown/rolldown/issues/3806, follow https://rollupjs.org/configuration-options/#makeabsoluteexternalsrelative

### DIFF
--- a/crates/rolldown/tests/esbuild/default/export_wildcard_fs_node_common_js/artifacts.snap
+++ b/crates/rolldown/tests/esbuild/default/export_wildcard_fs_node_common_js/artifacts.snap
@@ -1,6 +1,5 @@
 ---
 source: crates/rolldown_testing/src/integration_test.rs
-snapshot_kind: text
 ---
 # Assets
 
@@ -22,11 +21,11 @@ Object.keys(fs).forEach(function (k) {
   });
 });
 
-var ___external = require("./external");
-Object.keys(___external).forEach(function (k) {
+var external = require("./external");
+Object.keys(external).forEach(function (k) {
   if (k !== 'default' && !Object.prototype.hasOwnProperty.call(exports, k)) Object.defineProperty(exports, k, {
     enumerable: true,
-    get: function () { return ___external[k]; }
+    get: function () { return external[k]; }
   });
 });
 
--- a/crates/rolldown/tests/rollup/relative-external/artifacts.snap
+++ b/crates/rolldown/tests/rollup/relative-external/artifacts.snap
@@ -1,0 +1,35 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+---
+# Assets
+
+## main.js
+
+```js
+import "./external";
+
+```
+---
+
+Variant: ()
+
+# Assets
+
+## main.js
+
+```js
+import "./external";
+
+```
+---
+
+Variant: ()
+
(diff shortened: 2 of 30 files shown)


crates/rolldown/tests/esbuild/default/export_wildcard_fs_node_common_js/artifacts.snap
```diff
@@ -22,11 +21,11 @@ Object.keys(fs).forEach(function (k) {
   });
 });
 
-var ___external = require("./external");
-Object.keys(___external).forEach(function (k) {
+var external = require("./external");
+Object.keys(external).forEach(function (k) {
```
underfin, line 25: Here the relative external module `resolved id` is a absolute path, the `name` is a relative path, the `name` is empty if the external module is `.`, found the case at rollup test `rollup@function@external-directory-import: handles using ../ as external import`. 
So here using the normalized relative path to generate the name, if the external module is '.', after the normalied it is '.'. Using relative path from root, because here common dir is not correctly, ref https://github.com/rolldown/rolldown/pull/3834#discussion_r1993066008/

crates/rolldown/tests/rollup/relative-external/artifacts.snap
```diff
@@ -0,0 +1,35 @@
...
+---
+
+Variant: ()
+
+# Assets
+
+## chunk/main.js
+
+```js
+import "../external";
```
underfin, line 33: The test you could see the rendered `./external` will changed if set `entryFilename: 'chunk/[name].js'`,

crates/rolldown_common/src/module/external_module.rs
```diff
@@ -11,25 +15,63 @@ pub struct ExternalModule {
   /// - Used for iife format to inject symbol and deconflict.
   /// - Used for for rewrite `import { foo } from 'external';console.log(foo)` to `var external = require('external'); console.log(external.foo)` in cjs format.
   pub namespace_ref: SymbolRef,
+  pub id: ArcStr,
```
IWANABETHATGUY, line 20: Can you add some docs about `name` and `id`?

crates/rolldown_common/src/module/external_module.rs
```diff
@@ -11,25 +15,63 @@ pub struct ExternalModule {
...
+    if !self.need_renormalize_render_path {
+      return self.name.clone();
+    }
+    let mut target = self.name.as_str();
+    let mut importer = chunk
+      .preliminary_filename
+      .as_deref()
+      .expect("importer chunk should have preliminary_filename")
+      .to_string();
+    while target.starts_with("../") {
```
IWANABETHATGUY, line 62: This does not work on windows

crates/rolldown_common/src/module/external_module.rs
underfin: Added.

crates/rolldown_common/src/module/external_module.rs
underfin: The `target` come form the `name`, it could be an absolute path or a normalized relative path. Here only handle the already normalized relative path.

IWANABETHATGUY (approved):

PR #3887, opened 2025-03-17
docs(contribution): add `CMake` install guide

shulaoda:
### Description

We need to guide users to install it in advance.

@hyf0 cc

### DIFF
--- a/docs/contrib-guide/setup-the-project.md
+++ b/docs/contrib-guide/setup-the-project.md
@@ -37,6 +37,22 @@ cargo install just
 
 :::
 
+- Install `cmake`
+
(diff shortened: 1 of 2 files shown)


docs/contrib-guide/setup-the-project.md
```diff
@@ -37,6 +37,22 @@ cargo install just
 
 :::
 
+- Install `cmake`
+
+You could install just quickly by running the following command or by following the official [Download](https://cmake.org/download/):
```
IWANABETHATGUY: You missed linux, since linux has so much variants, I would suggest just left the official download page link should be enough.

IWANABETHATGUY (approved):

PR #3896, opened 2025-03-19
docs: enable deep outline

underfin:
### Description

Enable deep outline will make the doc summary clear to find(the options page always need to it), and it could jump to the content after clicking.

image


hyf0 (approved):
Nicely done! I keep wondering how to do this.

Issue #3892, opened 2025-03-18
Improve tests for `Vite`’s built-in plugins


shulaoda:
Port tests for `Vite`'s built-in plugins, fix potential issues, and improve the plugin code.

- [x] dynamicImportVars
- [x] importGlob
- [x] modulePreloadPolyfill
- [x] json

PR #3912, opened 2025-03-20
test(builtin-plugin): migrate `dynamicImportVars` test cases from `Vite`

shulaoda:
### Description

Related to #3892

Based on the snapshot results, there seem to be some issues with the current `dynamicImportVars` plugin.


packages/rolldown/tests/fixtures/builtin-plugin/dynamic-import-vars/vite/main.js
```diff
@@ -0,0 +1,72 @@
+export function basic(base) {
+  return import(`./mods/${base}.js`)
+}
+
+// export function aliasPath(base) {
+//   return import(`@/${base}.js`)
```
shulaoda, line 6: Currently, this type is not supported.

IWANABETHATGUY (approved):

PR #3916, opened 2025-03-21
perf: avoid chunk content clone

underfin:

### DIFF
--- a/crates/rolldown/src/utils/render_chunks.rs
+++ b/crates/rolldown/src/utils/render_chunks.rs
@@ -30,16 +30,13 @@ pub async fn render_chunks<'a>(
       })
       .collect::<FxHashMap<ArcStr, Arc<RollupRenderedChunk>>>(),
   );
-  let result = try_join_all(assets.iter().enumerate().map(|(index, asset)| {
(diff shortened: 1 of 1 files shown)


crates/rolldown/src/utils/render_chunks.rs
```diff
@@ -30,16 +30,13 @@ pub async fn render_chunks<'a>(
...
       .collect::<FxHashMap<ArcStr, Arc<RollupRenderedChunk>>>(),
   );
-  let result = try_join_all(assets.iter().enumerate().map(|(index, asset)| {
+  let result = try_join_all(assets.iter_mut().enumerate().map(|(index, asset)| {
     let chunks = Arc::clone(&chunks);
     async move {
-      // TODO(hyf0): To be refactor:
-      // - content should use ArcStr
-      // - plugin_driver.render_chunk should return Option<...> to be able to see if there is a return value by the plugin
       if let InstantiationKind::Ecma(ecma_meta) = &asset.kind {
```
underfin, line 36: I remove the todo.
- content should use ArcStr. The String to ArcStr has overhead at here, here using String is better.
- return Option<...> to see if there is a return value by the plugin. It is meaningless at here. If here need to it we could to changed it at here.

hyf0 (approved):

PR #3920, opened 2025-03-21
feat: add HookTransformAstArgs#stable_id

underfin:

### DIFF
--- a/crates/rolldown_plugin_isolated_declaration/src/lib.rs
+++ b/crates/rolldown_plugin_isolated_declaration/src/lib.rs
@@ -35,10 +35,8 @@ impl Plugin for IsolatedDeclarationPlugin {
         visitor.imported
       });
 
-      let importer_path = args.cwd.join(args.id);
-      let importer = importer_path.to_string_lossy();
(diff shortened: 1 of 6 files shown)


crates/rolldown_plugin_isolated_declaration/src/lib.rs
```diff
@@ -58,7 +56,7 @@ impl Plugin for IsolatedDeclarationPlugin {
       let codegen_ret =
         CodeGenerator::new().with_options(CodegenOptions::default()).build(&ret.program);
 
-      let mut emit_dts_path = Path::new(args.id).to_path_buf();
+      let mut emit_dts_path = Path::new(args.stable_id).to_path_buf();
```
hyf0, line 59: Would it be better to calculate itself?

Stable id might be not valid path. It may adds more non-path content to keep uniqueness.

underfin: Yeah. It should check the module path is valid at first, it also need to update the importer vitural module path. i will fixed it at next.

hyf0 (approved):

PR #3921, opened 2025-03-21
docs: remove about and redirect to guide

TheAlexLichter:
### Description

This PR removes the `/about` page as it is not meant to be exposed anymore. Further it ensures redirects are in place to `/guide` where most of the info lives.


hyf0:
Run `just fmt` to fix all format errors.

IWANABETHATGUY (approved):

PR #3928, opened 2025-03-22
fix: determine a module side effects for `require` a module that can't analyze statically

### LINKED ISSUE #3926: [Bug]: Panics when bundling `yargs` mjs entry but not cjs

Run the following in an empty directory:
```
npm init -y && npm i rolldown@nightly yargs && npx rolldown -o build.js ./node_modules/yargs/index.mjs
```

### What is expected?

I'd expect rolldown to not panic and produce a bundle containing the `yargs` NPM package

```
thread 'tokio-runtime-worker' panicked at /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/oxc_index-3.0.0/src/indexing.rs:38:10:
index out of bounds: the len is 39 but the index is 4294967295

 ERROR  Panic in async function                                                                           
```

### System Info

```Shell
System:
    OS: macOS 15.3.2
    CPU: (10) arm64 Apple M1 Max
    Memory: 89.33 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.11.0 - ~/.volta/tools/image/node/22.11.0/bin/node
    Yarn: 1.22.22 - ~/.volta/tools/image/yarn/1.22.22/bin/yarn
    npm: 10.9.0 - ~/.volta/tools/image/node/22.11.0/bin/npm
    pnpm: 9.15.4 - ~/.volta/bin/pnpm
    Watchman: 2025.02.17.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 134.0.6998.118
    Chrome Canary: 136.0.7082.1
    Safari: 18.3.1
  npmPackages:
    rolldown: ^1.0.0-beta.6-commit.47fe675 => 1.0.0-beta.6-commit.47fe675 
    yargs: ^17.7.2 => 17.7.2
```

### Any additional comments?

Note that the following does work and does not panic:
```
npm init -y && npm i rolldown@nightly yargs && npx rolldown -o build.js ./node_modules/yargs/index.cjs
```

For context, I'm attempting to use rolldown to bundle some other scripts, but found that it panicked and discovered this to be the minimal repro of the issue. I'm unblocked by changing my script to import from `yargs/yargs` but figured I'd pass along the panic I ran into and the repro steps since it isn't too hard.

I was also able to repro with rolldown@1.0.0-beta.6 
I tried the nightly to see if it was fixed but it still panicked.

IWANABETHATGUY:
### Description
1. closed #3926

### DIFF
--- a/crates/rolldown/tests/rolldown/tree_shaking/require_module_determined_in_runtime/foo.js
+++ b/crates/rolldown/tests/rolldown/tree_shaking/require_module_determined_in_runtime/foo.js
@@ -1,0 +1,4 @@
+export default function applyExtends(config) {
+  require(config.extends);
+}
+
--- a/crates/rolldown/src/stages/link_stage/tree_shaking/determine_side_effects.rs
+++ b/crates/rolldown/src/stages/link_stage/tree_shaking/determine_side_effects.rs
@@ -18,7 +18,15 @@ impl LinkStage<'_> {
       module_id: ModuleIdx,
       normal_modules: &IndexModules,
     ) -> DeterminedSideEffects {
-      let module = &normal_modules[module_id];
+      let Some(module) = normal_modules.get(module_id) else {
+        // It could be a module that could not be analyzed statically.
+        // ```js
+        // export default function() {
+        //   require(window.something);
+        // }
+        // ```
+        return DeterminedSideEffects::Analyzed(true);
+      };
 
       match &mut cache[module_id] {
         SideEffectCache::None => {
--- a/crates/rolldown/tests/rolldown/tree_shaking/require_module_determined_in_runtime/_config.json
+++ b/crates/rolldown/tests/rolldown/tree_shaking/require_module_determined_in_runtime/_config.json
@@ -1,0 +1,11 @@
+{
+  "config": {
+    "input": [
+      {
+        "import": "main.mjs",
+        "name": "main"
+      }
+    ]
+  },
+  "expectExecuted": false
+}
--- a/crates/rolldown/tests/rolldown/tree_shaking/require_module_determined_in_runtime/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/tree_shaking/require_module_determined_in_runtime/artifacts.snap
@@ -1,0 +1,21 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+---
+# Assets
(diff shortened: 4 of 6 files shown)


IWANABETHATGUY:
* **#3928** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3928?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3928?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/rolldown/tests/rolldown/tree_shaking/require_module_determined_in_runtime/foo.js
```diff
@@ -0,0 +1,4 @@
+export default function applyExtends(config) {
+  require(config.extends);
```
hyf0, line 2: Should this be ignored in the first place? I saw we introduced

https://github.com/rolldown/rolldown/blob/5da7efb1c26441859803337f06aa659537588e9a/crates/rolldown/src/ast_scanner/impl_visit.rs#L318-L323

for handling such a case.

- It generates a `ImportRecord` of nothing.
- It results to a `ModuleIdx` that points to non-exist module.

Does this also means every place of using `ModuleIdx` need to do a second validation to make sure the `ModuleIdx` is valid?

---

I don't known the historical contexts, so there are just some intuitive thoughts that may be wrong.

hyf0 (approved):

crates/rolldown/tests/rolldown/tree_shaking/require_module_determined_in_runtime/foo.js
IWANABETHATGUY: Yeah, feel free to change it if anyone have a better way to to represent a module like `require(window..context)` which could not be analyzed staticlly.

IWANABETHATGUY: Actually not everyplace, in most of the place it did not, e.g. A owner of `SymbolRef` would never be a DummyModule, only those places you need to iterate all modules.

PR #3929, opened 2025-03-22
test(builtin-plugin): migrate `importGlob` test cases from `Vite`

shulaoda:
### Description

Related to #3892 

Based on the snap results, it seems that it is not fully aligned with `Vite`.


packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/virtual-modules/_config.ts
```diff
@@ -0,0 +1,46 @@
...
+      {
+        name: 'virtual:module',
+        resolveId(id) {
+          if (id === 'virtual:module') {
+            return 'virtual:module'
+          }
+        },
+        load(id) {
+          if (id === 'virtual:module') {
+            // TODO: support importGlob in virtual module
```
shulaoda, line 29: This is currently not supported.

IWANABETHATGUY (approved):

packages/rolldown/tests/fixtures/builtin-plugin/glob-import/vite/restore-query-extension/_config.ts
```diff
@@ -0,0 +1,43 @@
...
+)
+
+export default defineTest({
+  config: {
+    input: '../fixtures/a/index.ts',
+    output: {
+      chunkFileNames: '[name].js',
+    },
+    plugins: [
+      importGlobPlugin({ root, restoreQueryExtension: true }),
```
shulaoda, line 19: With `restoreQueryExtension` enabled, the behavior differs significantly from `Vite`.

IWANABETHATGUY: You could fixed it later.

PR #3953, opened 2025-03-24
test(module-preload-polyfill): port relevant cases from `Vite`

shulaoda:
### Description

Related to #3892


packages/rolldown/tests/fixtures/builtin-plugin/module-preload-polyfill/cjs/_config.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export default defineTest({
+  config: {
+    output: {
+      format: 'cjs',
+    },
+    plugins: [modulePreloadPolyfillPlugin()],
+  },
+  async afterTest(output) {
+    // TODO: shouldn't load modulepreload polyfill when format is cjs
```
shulaoda, line 14: I'm not sure at which level this logic is implemented. I suspect it's handled internally by `rollup`, as the plugin itself doesn't seem to process `CJS` specifically.

packages/rolldown/tests/fixtures/builtin-plugin/module-preload-polyfill/cjs/_config.ts
```diff
@@ -0,0 +1,19 @@
...
+    await expect(output.output[0].code).toMatchFileSnapshot(
+      path.resolve(import.meta.dirname, 'main.js.snap'),
+    )
```
shulaoda, line 17: I'm not sure at which level this logic is implemented. I suspect it's handled internally by rollup, as the plugin itself doesn't seem to process CJS specifically.

shulaoda: The most likely reason is related to `__VITE_IS_MODERN__` in `Vite`. In `Vite` plugin, it is used as a boolean value, which might have been tree-shaken in the final output.

packages/rolldown/tests/fixtures/builtin-plugin/module-preload-polyfill/cjs/_config.ts
```diff
@@ -0,0 +1,19 @@
...
+  },
```
shulaoda: It should be related to `__VITE_IS_MODERN__`. The Vite plugin performs the replacement, but the output from the native plugin does not contain it.

shulaoda: Currently, when enabling the native `modulePreloadPolyfillPlugin` in `rolldown-vite`, the `__VITE_IS_MODERN__` variable is not being replaced. This is likely caused by our native `buildImportAnalysisPlugin` not performing the replacement.

Based on my review of how `__VITE_IS_MODERN__` is replaced, it seems to be set to true only when the format is `ESM`. Should we directly incorporate this logic to make it a general-purpose plugin? @sapphi-red cc

sapphi-red: Vite replaces `__VITE_IS_MODERN__` in the renderChunk hook because rollup runs renderChunk for each output format. On the other hand, rolldown also runs the transform hook for each output format. So for rolldown, we can use the define plugin or the replace plugin to achieve the same thing.

shulaoda: So, do you think the replacement for `Vite` flags like `__VITE_IS_MODERN__` should be handled in the `define` or `replace` plugin on the `rolldown-vite` side, or should it align with `buildImportAnalysisPlugin` and be replaced in `renderChunk`?

sapphi-red: I think it should be handled in `define` plugin on rolldown-vite side, as that's probably more performant.

IWANABETHATGUY (approved):

PR #3959, opened 2025-03-25
test(json): port relevant cases from `Vite`

shulaoda:
### Description

Related to #3892

There are many related tests, and I will try to remove unnecessary ones later.


packages/rolldown/tests/fixtures/builtin-plugin/json/vite/1/3.json
```diff
@@ -0,0 +1,3 @@
...
...tevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevitevite",
```
IWANABETHATGUY: Did repeating so many times of `vite` have a special meaning?

sapphi-red: It is to exceed this threshold.
https://github.com/vitejs/vite/blob/c5b71915099cfbc15447a166f35620fa0e05c023/packages/vite/src/node/plugins/json.ts#L89-L91

shulaoda: This test seems to cover cases where the target size is not reached. I just did a simple port, and it can be removed.

IWANABETHATGUY: > It is to exceed this threshold. https://github.com/vitejs/vite/blob/c5b71915099cfbc15447a166f35620fa0e05c023/packages/vite/src/node/plugins/json.ts#L89-L91

Ah, I see.

IWANABETHATGUY: > This test seems to cover cases where the target size is not reached. I just did a simple port, and it can be removed.

I just wondering the reason, no need to remove it.

IWANABETHATGUY (approved):

PR #3082, opened 2024-12-09
fix: dedupe external import binding for format `esm`

IWANABETHATGUY:
### Description
1. fixed https://github.com/rolldown/rolldown/issues/1548

PR #3966, opened 2025-03-26
refactor: make scoping immutable

IWANABETHATGUY:

### DIFF
--- a/crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
+++ b/crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
@@ -55,6 +55,7 @@ pub fn deconflict_chunk_symbols(
         .as_ref()
         .unwrap()
         .ast_scopes
+        .scoping()
         .root_unresolved_references()
         .keys()
         .map(Cow::Borrowed)
@@ -99,9 +100,18 @@ pub fn deconflict_chunk_symbols(
   if matches!(format, OutputFormat::Esm) {
     chunk.imports_from_external_modules.iter().for_each(|(module, _)| {
       let db = link_output.symbol_db.local_db(*module);
-      db.classic_data.iter_enumerated().skip(1).for_each(|(symbol, _)| {
-        renamer.add_symbol_in_root_scope((*module, symbol).into());
+      db.classic_data.iter_enumerated().for_each(|(symbol, _)| {
+        let symbol_ref = (*module, symbol).into();
+        if link_output.used_symbol_refs.contains(&symbol_ref) {
+          renamer.add_symbol_in_root_scope(symbol_ref);
+        }
       });
+      for symbol_id in db.ast_scopes.facade_symbol_classic_data().keys() {
+        let symbol_ref = (*module, *symbol_id).into();
+        if link_output.used_symbol_refs.contains(&symbol_ref) {
+          renamer.add_symbol_in_root_scope(symbol_ref);
+        }
+      }
     });
   }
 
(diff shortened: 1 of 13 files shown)


IWANABETHATGUY:
* **#3966** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3966?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3966?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
```diff
@@ -99,9 +100,18 @@ pub fn deconflict_chunk_symbols(
   if matches!(format, OutputFormat::Esm) {
     chunk.imports_from_external_modules.iter().for_each(|(module, _)| {
       let db = link_output.symbol_db.local_db(*module);
-      db.classic_data.iter_enumerated().skip(1).for_each(|(symbol, _)| {
-        renamer.add_symbol_in_root_scope((*module, symbol).into());
```
IWANABETHATGUY, line 103: I am not sure it is a heck or somethingelse, for this case: 


```js
import * as b from 'b'

console.log(`b: `, b)
```
If you remove `skip(1)` it will generate output: 

```js
import * as b$1 from "b";

//#region entry.js
console.log(`b: `, b$1);

//#endregion
//# sourceMappingURL=entry.js.map
```

crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
```diff
@@ -99,9 +100,18 @@ pub fn deconflict_chunk_symbols(
...
+      db.classic_data.iter_enumerated().for_each(|(symbol, _)| {
+        let symbol_ref = (*module, symbol).into();
+        if link_output.used_symbol_refs.contains(&symbol_ref) {
+          renamer.add_symbol_in_root_scope(symbol_ref);
+        }
       });
+      for symbol_id in db.ast_scopes.facade_symbol_classic_data().keys() {
```
IWANABETHATGUY, line 109: I changed to filter by `used_symbol_refs` because we separate `facade_scoping` and `scoping`, we can't use that trick anymore.

crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
hyf0: Codes are introduced in this PR https://github.com/rolldown/rolldown/pull/3082/files#diff-3e1292ddb548179d38ab0c4ce82cc9a16844334560d9e9c8ef26ebab2b389899R92-R99.

I looked up the PR again and didn't found what does the `skip(1)` do.

hyf0 (approved):

crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
IWANABETHATGUY: It is used to filter unused external symbol, which avoid same symbol name to be added twice, once for external symbol, once for import_stmt declared symbol.

IWANABETHATGUY:
https://codspeed.io/rolldown/rolldown/branches/03-24-try_make_scoping_immutable

There might be a little regression, but it should be a big improvement when incremental build.

I will try to improve it in the futrue.

PR #3974, opened 2025-03-27
feat: optimize dep unchanged

IWANABETHATGUY:
### Description
This pull request can ensure that if a module is modified without changing its dependency, only that module will be rebuilt. Before that, the sub-dependency graph of that module will be appended to the rebuild queue.
1. I  merged `intermediate.importers` into `cache. importers` because the `fetch_modules` will access the module even if the module is not changed, before It is ok because the subgraph will be always rebuild, https://github.com/rolldown/rolldown/blob/4092483d2b287e0f4619544275b2063c8f309fd1/crates/rolldown/src/module_loader/module_loader.rs?plain=1#L409-L413, the `idx` will be always allocated before access. Now, since the idx will not always be allocated if the dependency is not changed, if we use `HybridIndexVec`, it will panic at runtime.
2. Another notable I passed the owned cache instead of `cache.module_id_to_idx`, this is because we now have two global maintained field instead of one, maybe more in the future. Passing them one by one may make the function signature hard to read in the future, 
Now the cache has different semantics in different context,
   - In full build, it is almost identical as before.
   - In HMR and incremental rebuild, it has some globally maintained data structure we could modify on the fly during the module loading.
Taking ownership is fine. The program will immediately panic if you don't set it back when you use it, and it is only used in a few places. I don't use mutable references, because that will propagate the lifetime marking everywhere.

### DIFF
--- a/crates/rolldown/src/hmr/hmr_manager.rs
+++ b/crates/rolldown/src/hmr/hmr_manager.rs
@@ -15,10 +15,8 @@ use rolldown_utils::indexmap::FxIndexSet;
 use rustc_hash::{FxHashMap, FxHashSet};
 
 use crate::{
-  SharedOptions, SharedResolver,
-  hmr::hmr_ast_finalizer::HmrAstFinalizer,
-  module_loader::{ModuleLoader, module_loader::VisitState},
-  type_alias::IndexEcmaAst,
+  SharedOptions, SharedResolver, hmr::hmr_ast_finalizer::HmrAstFinalizer,
+  module_loader::ModuleLoader, type_alias::IndexEcmaAst, types::scan_stage_cache::ScanStageCache,
 };
 
 pub struct HmrManagerInput {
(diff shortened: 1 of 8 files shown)


IWANABETHATGUY:
* **#3974** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3974?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3974?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/rolldown/src/hmr/hmr_manager.rs
```diff
@@ -123,31 +122,20 @@ impl HmrManager {
...
-      .modules
-      .iter_enumerated()
-      .map(|(idx, module)| {
-        if modules_to_invalidate.contains(&idx) {
-          (module.id().into(), VisitState::Invalidate(idx))
-        } else {
-          (module.id().into(), VisitState::Seen(idx))
-        }
-      })
-      .collect::<FxHashMap<ArcStr, VisitState>>();
```
IWANABETHATGUY, line 137: Already did here https://github.com/rolldown/rolldown/blob/f7f0b2dd9ae87d7501752270ed04a27d7c121734/crates/rolldown/src/module_loader/module_loader.rs?plain=1#L347-L356

hyf0 (approved):

PR #3976, opened 2025-03-28
fix!: remove glob filter support for resolveId hook

sapphi-red:
### Description

Removed the string (glob) id filter support for resolveId hook.
Globs are meant to be matched against absolute path. On the other hand, raw ids (specifiers) can be relative paths. The semantics of matching relative paths with globs are not clear and is confusing.

This PR only removes the ability from the types as that was easier and it would not break existing plugins (if exists).

### DIFF
--- a/docs/guide/plugin-development.md
+++ b/docs/guide/plugin-development.md
@@ -91,3 +91,9 @@ interface HookFilter {
   code?: StringFilter
 }
 ````
+
+The following properties are supported by each hooks:
+
+- `resolveId` hook: `id` (only `RegExp`)
+- `load` hook: `id`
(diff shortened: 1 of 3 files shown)


sapphi-red:
* **#3976** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3976?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/3976?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

IWANABETHATGUY (approved):

PR #3985, opened 2025-03-30
refactor: annotate the type to be consistent with all_warnings

ShroXd:
### Description

Annotate type for `errors` to maintain consistent code style with `all_warnings`.


underfin:
Thank you. But the rust compiler could infer the typing, so here don't need to add it. If you using the vscode and enabled the rust anlazyer plugin, it could be give an typing for it.

PR #3995, opened 2025-03-31
docs: less ambiguous hook filter usage

TheAlexLichter:
This PR replaces "filter out" with "include" to ensure users understand easier whether the entries are included or excluded


IWANABETHATGUY (commented):
Thanks

IWANABETHATGUY (approved):

PR #4000, opened 2025-04-01
fix: remove blank line before `#endregion` comment

btea:
### Description

I think it makes sense to keep blank lines between module code, rather than between code and a `#endregion` comment.


crates/rolldown/src/utils/render_ecma_module.rs
```diff
@@ -38,13 +38,13 @@ pub fn render_ecma_module(
             .with_pre_compute_sourcemap_data(options.is_sourcemap_enabled()),
         ));
       } else {
-        sources.push(Box::new(render_output.code));
+        sources.push(Box::new(render_output.code.trim_end().to_string()));
       }
     } else {
-      sources.push(Box::new(render_output.code));
+      sources.push(Box::new(render_output.code.trim_end().to_string()));
```
underfin, line 44: The `to_string` allocate the code again at here. It is not worth to do it. 

Other: Here the line 37 also need to change, it need to make sure the sourcemap is valid.

btea: Thank you for your review. I am not familiar with Rust, how should I modify it here?

underfin: > Thank you for your review. I am not familiar with Rust, how should I modify it here?

It need to update the code structure at here, the source need to owner the `String` and trim_end `&str`.

I think it is not worth doing it because it will increase the complex, just using the oxc codgen code is fine.

btea: You mean this modification is unnecessary and the code structure can remain as it is, right?

underfin: Yes. But also thanks for your contribution.

btea:
@underfin Do you think we can just remove the `#endregion`(This way the blank lines won't look so weird.), it seems to have the same effect with or without it. 👀

PR #4003, opened 2025-04-02
chore(rolldown): oxc v0.62.0

Boshen:
In this release:

* built-in oxc `TakeIn` trait
* https://esbuild.github.io/api/#keep-names


crates/rolldown/src/hmr/hmr_ast_finalizer.rs
```diff
@@ -316,7 +318,10 @@ impl<'ast> VisitMut<'ast> for HmrAstFinalizer<'_, 'ast> {
               class.id = Some(self.snippet.id("__rolldown_default__", SPAN));
               self.exports.insert("default".into(), "__rolldown_default__".into());
             }
-            *node = ast::Statement::ClassDeclaration(class.take_in(self.alloc));
+            *node = ast::Statement::ClassDeclaration(ArenaBox::new_in(
+              class.as_mut().take_in(self.alloc),
```
Boshen, line 323: This is the only difference between the previous rolldown implementation vs the current oxc implementation.

crates/rolldown/tests/esbuild/dce/drop_label_tree_shaking_bug_issue3311/artifacts.snap
```diff
@@ -10,7 +9,7 @@ snapshot_kind: text
 
 //#region entry.js
 const myFunc = () => {
-	;
+	debugger;
```
Boshen: Are these dev or prod build?

crates/rolldown/tests/esbuild/ts/ts_enum_tree_shaking/artifacts.snap
```diff
@@ -23,6 +23,7 @@ var x = /* @__PURE__ */ function(x$1) {
 ```js
 
 //#region namespace-before.ts
+let x;
```
Boshen, line 26: @Dunqing can you double check this?

Boshen:
Test fails:

```
failures:

---- test_tests__esbuild__default__named_function_expression_argument_collision__config_json stdout ----
Input: /Users/boshen/github/rolldown/crates/rolldown/tests/esbuild/default/named_function_expression_argument_collision/_config.json


failures:
    test_tests__esbuild__default__named_function_expression_argument_collision__config_json
```

```
# Reason
1. the name `foo` binding of function expression can't not referenced anywhere, rolldown have same behavior as esbuild
# Diff
## /out/entry.js
### esbuild
```js
let x = function(foo) {
  var foo;
  return foo;
};
```
### rolldown
```js

//#region entry.js
let x = function foo(foo) {
	var foo;
	return foo;
};

//#endregion
```
### diff
```diff
===================================================================
--- esbuild	/out/entry.js
+++ rolldown	entry.js
@@ -1,4 +1,4 @@
-let x = function (foo) {
+let x = function foo(foo) {
     var foo;
     return foo;
 };

```
```

We now remove the function name `foo`, how do I update this?

crates/rolldown/tests/esbuild/ts/ts_enum_tree_shaking/artifacts.snap
Dunqing: Fixed in https://github.com/oxc-project/oxc/pull/10172?notification_referrer_id=NT_kwDOAcKkeLQxNTY0MDQwNTA5NToyOTUzMzMwNA#event-17086285809

crates/rolldown/tests/esbuild/dce/drop_label_tree_shaking_bug_issue3311/artifacts.snap
overlookmotel: I think that's an error in the logic of whatever code is producing this output. `TakeIn` should only be used to *temporarily* replace a node with a dummy, and then the dummy should be replaced with a "real" node later on.

With the old manually-written `TakeIn`, the dummy node generated when replacing a `Statement` was an `EmptyStatement`. But with the codegen-ed `TakeIn`, it's now a `DebuggerStatement` (just because `DebuggerStatement` is an earlier variant of `Statement`, so `TakeIn` codegen uses it).

This change shouldn't matter, because the dummy node should never end up in the final AST anyway.

IWANABETHATGUY (approved):

IWANABETHATGUY:
temp disable the failed test case , tracked https://github.com/rolldown/rolldown/issues/4045

PR #4014, opened 2025-04-02
perf(rolldown): re-implement `renamer` to avoid renaming non-root symbols

### LINKED ISSUE #4045: [Bug]: recover esbuild test case named_function_expression_argument_collision

https://github.com/rolldown/rolldown/pull/4003/files#diff-7ecb341752cb2c832ed9a3901f586bec70e4eecd9fcb08697385e795547e1fb8

### What is expected?

n/a

n/a

### System Info

```Shell
n/a
```

### Any additional comments?

_No response_

Dunqing:
### Description

close: #4045

This PR is to re-implement the renamer logic to get rid of the expensive `rename_non_root_symbol` method, and can also address the #4045 problem.

Whether from a new output or performance perspective, I think this change is the correct direction and worth considering.

### How does it work?

This implementation follows a principle that does NOT rename to a name that has been declared, even declared in child scopes. Plus, we regard the entry module as a base module because bundling will bundle every referenced code into the entry module. In other words, we ensure that all entry module names are maintained as-is.

Let's take an example to see how and why:
```js
// entry.js
import './a.js'
const a = 0;

function func(a$1) {
  console.log(a, a$1)
}

// a.js
const a = 0;
console.log(a)
```

Before output:
```js
//#region a.js
const a$1 = 0;
console.log(a$1);

//#endregion
//#region entry.js
const a = 0;
function func(a$1$1) {
	console.log(a, a$1$1);
}

//#endregion
```

After output:

```js
//#region a.js
const a$2 = 0;
console.log(a$2);

//#endregion
//#region entry.js
const a = 0;
function func(a$1) {
	console.log(a, a$1);
}

//#endregion
```

Let's take a look at the before output first, as you can see, after bundling, the `a` in `a.js` has been renamed to `a$1`, which conflicts with `func`'s param `a$1`, that lead param `a$1` has to rename to `a$1$1` to avoid conflict. And let's look at the after output, the `a` in `a.js` has been renamed to `a$2` because `a$1` has been declared in child scopes, as I mentioned principle earlier, so that param `a$1` doesn't need to be renamed because no conflict here. In this way, we no longer need to call an expensive `rename_non_root_symbol`. 

Another key change is that if the `NamedImport` has an alias name, then use it as its new name

```js
// entry.js
import { y as x } from './y.js'
let y = "ENTRY"
```

Before output:
```js
//#region y.js
const y = "y";
console.log(y);

//#endregion
//#region entry.js
let y$1 = "ENTRY";

//#endregion
```

After output:
```js
//#region y.js
const x = "y";
console.log(x);

//#endregion
//#region entry.js
let y = "ENTRY";

//#endregion
```

Let's compare the before and the after output. Unlike the before implementation, we follow the principle we've mentioned earlier, we won't rename entry module names. So there is no change in the entry file.  Let's see the changed part, the `y` in `y.js` has been renamed to `x`  because `x` is the alias name of `y` they use, so that we don't need to calculate the name for it, just use the alias name!

### Pefromance 
https://codspeed.io/rolldown/rolldown/branches/04-03-refactor_rolldown_simplify_renamer_logic
image

I guess the main performance improvement comes in which we removed `rename_non_root_symbols`.

### Further improvement

After this PR, the assumption that `canonical_names` will store the names of all symbols no longer exists. If `canonical_names` doesn't contain that symbol, that means this symbol doesn't need to be renamed. We should return early to avoid unnecessary steps. In this PR, I avoided introducing too many changes, so I changed [canonical_name_for](https://github.com/rolldown/rolldown/blob/3bd706bfa4421ac3e1d9ab3a7684d09c737f18ed/crates/rolldown_common/src/types/symbol_ref_db.rs#L196-L205) method to return the symbol name if it doesn't exist in `canonical_names`. I guess if we refactor this, that would be a notable performance improvement.


Dunqing:
Five rollup tests failed due to a difference renaming way, how to tweak the rollup test outputs?

underfin:
``` .js
// entry.js
import './a.js'
const a = 0;

function func1(a$1) {
  console.log(a, a$1)
}

// a.js
const a = 0;
function func2(a$1, a$2) { // need to renamer
  console.log(a, a$1, a$2)
}


```

Could you hava a try with the case, Look like the `a.js` function scope identifier need to renamer.

Dunqing:
> ```js
> // entry.js
> import './a.js'
> const a = 0;
> 
> function func1(a$1) {
>   console.log(a, a$1)
> }
> 
> // a.js
> const a = 0;
> function func2(a$1, a$2) { // need to renamer
>   console.log(a, a$1, a$2)
> }
> ```
> 
> Could you hava a try with the case, Look like the `a.js` function scope identifier need to renamed.

Output is:
```js
//#region a.js
const a$3 = 0;
function func2(a$1, a$2) {
	console.log(a$3, a$1, a$2);
}

//#endregion
//#region entry.js
const a = 0;
function func1(a$1) {
	console.log(a, a$1);
}

//#endregion
```

This case is correct in the new implementation. Another key point I forgot to mention is that we also need to avoid renaming a name that has been declared in the current module (a.js). In the `a.js` file, `a$2` has been declared in child scope, so it is renamed to `a$3`. Therefore, no conflict happens here.

underfin:
> Five rollup tests failed due to a difference renaming way, how to tweak the rollup test outputs?

You could move the tests to https://github.com/rolldown/rolldown/blob/main/packages/rollup-tests/src/ignored-passed-snapshot-different-tests.js, it will be ignored.

 "rollup@form@absolute-path-resolver: normalizes absolute ids",
  "rollup@form@deconflict-module-priority: prioritizes entry modules over dependencies when deconflicting",
  "rollup@form@default-identifier-deshadowing: Handles export default identifier reassignment deshadowing@generates es",
  "rollup@form@define-replacement: amd.define@generates es",
  "rollup@form@reexport-star-deshadow: Star reexports scope deshadowing@generates es"

IWANABETHATGUY:
These test outputs are just not expected, 

for case: 


```js
// main.js
import dep from './dep.js';
dep();


// dep.js

export default function a() {
  a = someGlobal;
  return a();
}

```

# This branch

```js
//#region dep.js
function dep() {
	dep = someGlobal;
	return dep();
}

//#endregion
//#region main.js
dep();

//#endregion
```

# Main and rollup output
```js
//#region dep.js
function a() {
  a = someGlobal;
  return a();
}

a();


//#endregion

```

IWANABETHATGUY:
You just rename it even there is no conflict.

crates/rolldown/src/utils/renamer.rs
```diff
@@ -34,65 +27,149 @@ pub struct Renamer<'name> {
...
-              vac.insert(0);
-              break;
-            }
-          }
-        }
-        vacant.insert(candidate_name);
+
+    if !self.canonical_names.contains_key(&canonical_ref) {
+      let name = self.get_unique_name(symbol_ref, &original_name);
+      self.canonical_names.insert(canonical_ref, name);
```
underfin: Here the `insert` `canonical_names` will hash twice.

crates/rolldown/src/utils/renamer.rs
```diff
@@ -34,65 +27,149 @@ pub struct Renamer<'name> {
...
+      });
+
+      if (is_root_binding && !self.manual_reserved.contains(candidate_name.as_str()))
+        || !self.used_names.contains(&candidate_name)
+          && !(self.entry_module_used_names.contains(candidate_name.as_str()))
+          && !self
+            .module_used_names
+            .entry(canonical_ref.owner)
+            .or_insert_with(|| Self::get_module_used_names(self.symbol_db, canonical_ref))
+            .contains(candidate_name.as_str())
```
underfin: How about merge the `manual_reserved /used_names/entry_module_used_names/module_used_names` into one `used_names`. It could reduce many hash.

IWANABETHATGUY:
> // entry.js
> import { y as x } from './y.js'
> let y = "ENTRY"
> Before output:
> 
> //#region y.js
> const y = "y";
> console.log(y);
> 
> //#endregion
> //#region entry.js
> let y$1 = "ENTRY";
> 
> //#endregion
> After output:
> 
> //#region y.js
> const x = "y";
> console.log(x);
> 
> //#endregion
> //#region entry.js
> let y = "ENTRY";
> 
> //#endregion


Can you revert this change? for a bundler responsibility, we should preserve the original definition as much as possible, before 
we change a -> a$1, after you change x -> y, this make it hard to search the definition in source code, AFAIK, there is no bunlder rename like that when conflict.

Dunqing:
> Can you revert this change? for a bundler responsibility, we should preserve the original definition as much as possible, before we change a -> a$1, after you change x -> y, this make it hard to search the definition in source code, AFAIK, there is no bunlder rename like that when conflict.

Good point! Yes, we can. The main reason I did this was to fix https://github.com/rollup/rollup/blob/64870c31062983a741832d8cd2d4bcdf55e4de55/test/function/samples/class-name-conflict-2 test fails. I found this case fails due to the `treeshake` option, after I disable, it works. It really weird.

IWANABETHATGUY:
> > Can you revert this change? for a bundler responsibility, we should preserve the original definition as much as possible, before we change a -> a$1, after you change x -> y, this make it hard to search the definition in source code, AFAIK, there is no bunlder rename like that when conflict.
> 
> Good point! Yes, we can. The main reason I did this was to fix https://github.com/rollup/rollup/blob/64870c31062983a741832d8cd2d4bcdf55e4de55/test/function/samples/class-name-conflict-2 test fails. I found this case fails due to the `treeshake` option, after I disable, it works. It really weird.

Fixing one bug should not introduce another bug, or use a workaround, Let's revert it first, and see if we could solve it in other way.

Dunqing:
> Fixing one bug should not introduce another bug, or use a workaround, Let's revert it first, and see if we could solve it in other way.

Yes, fixed in https://github.com/rolldown/rolldown/pull/4014/commits/fc6683f790cd2efc84dc8d0bf5703cef1134f255. When `treeshake` is enabled, the import statements may be excluded, but the current implementation needs to use them.

Dunqing:
> These test outputs are just not expected,
> 
> for case:
> 
> ```js
> // main.js
> import dep from './dep.js';
> dep();
> 
> 
> // dep.js
> 
> export default function a() {
>   a = someGlobal;
>   return a();
> }
> ```
> 
> # This branch
> ```js
> //#region dep.js
> function dep() {
> 	dep = someGlobal;
> 	return dep();
> }
> 
> //#endregion
> //#region main.js
> dep();
> 
> //#endregion
> ```
> 
> # Main and rollup output
> ```js
> //#region dep.js
> function a() {
>   a = someGlobal;
>   return a();
> }
> 
> a();
> 
> 
> //#endregion
> ```

It no longer happens in the latest commit.

crates/rolldown/src/utils/renamer.rs
Dunqing: `manual_reserved` and `entry_module_used_names` has removed.

Unfortunately, `used_names` and `module_used_names` can't be merged into one due to their different uses.
`used_names` is a `FxHashSet<Rstr>` that is used to store names that are used for renaming.
`module_used_names` is a `FxHashSet<&name Rstr>` that stores the module used names, and is exactly used in what I said in https://github.com/rolldown/rolldown/pull/4014#issuecomment-2788185805.

crates/rolldown/src/utils/renamer.rs
Dunqing: I've changed this.

IWANABETHATGUY:
I am hesitant to merge because, according to the snapshot, some of them are improved, others regressed, also lots of the internal code has been changed. I will try to fix the bug based on main branch, If I can't find a better way then I will merge it.

Dunqing:
> I am hesitant to merge because, according to the snapshot, some of them are improved, others regressed, also lots of the internal code has been changed. I will try to fix the bug based on main branch, If I can't find a better way then I will merge it.

That's ok! But aside from fixing the bug, there is a 1%-4% performance improvement in this PR. I think this is also worth considering!

For the output regression, I know the reason why it's happening, but to fix it, I need to dig deep into Rolldown, which will take forever. If you want to fix it, I’d be happy to pair with you!

Boshen:
It seems like we are in the right direction, but need to be cautious about the changes. Are we able to build this with rolldown-vite and test in ecosystem CI?

PR #4019, opened 2025-04-04
fix(rust): avoid panic when `assetFileNames` hash length exceeds `22`

### LINKED ISSUE #4016: [Bug]: an asset file name with [hash:23] panics rolldown

Set `output.assetFileNames` to `"[hash:23].[ext]"`

### What is expected?

Based on the unit tests and behaviors I could find `output.chunkFileNames` will limit the hash length to 21 characters even if requesting more. While for assets, you can request 22 and get 22, a behavior that's possibly even more buggy than the panic.

For `[hash:22]`, a 22-character hash is created (which is not possible for chunk filenames). For `[hash:23]` rolldown panics:

```
thread '<unnamed>' panicked at crates/rolldown_common/src/file_emitter.rs:190:40:
byte index 23 is out of bounds of `BKe-RU-cO4wpXAsrT9DUL7`
fatal runtime error: failed to initiate panic, error 772515008
error: script "build" was terminated by signal SIGABRT (Abort)
Aborted
```

### System Info

```Shell
System:
    OS: Linux 6.10 Debian GNU/Linux 12 (bookworm) 12 (bookworm)
    CPU: (16) arm64 unknown
    Memory: 26.79 GB / 31.29 GB
    Container: Yes
    Shell: 5.2.15 - /bin/bash
  Binaries:
    Node: 20.19.0 - /usr/local/share/nvm/versions/node/v20.19.0/bin/node
    Yarn: 1.22.22 - /usr/bin/yarn
    npm: 10.8.2 - /usr/local/share/nvm/versions/node/v20.19.0/bin/npm
    pnpm: 10.7.0 - /usr/local/share/nvm/versions/node/v20.19.0/bin/pnpm
    bun: 1.2.6 - /usr/local/bin/bun
```

### Any additional comments?

_No response_

shulaoda:
### Description

closes #4016 

I will improve the related test cases based on https://github.com/rolldown/rolldown/pull/3761 in a follow-up.

### DIFF
--- a/crates/rolldown_common/src/file_emitter.rs
+++ b/crates/rolldown_common/src/file_emitter.rs
@@ -187,7 +187,7 @@ impl FileEmitter {
       let mut filename = filename_template.render(
         name,
         Some(extension.unwrap_or_default()),
-        Some(|len: Option<usize>| &hash[..len.unwrap_or(8)]),
+        Some(|len: Option<usize>| &hash[..len.map_or(8, |len| len.min(21))]),
       );
 
       // deconflict file name
--- a/crates/rolldown/tests/rolldown/function/asset_filenames/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/function/asset_filenames/artifacts.snap
@@ -14,7 +14,7 @@ source: crates/rolldown_testing/src/integration_test.rs
 ```js
 
 ```
-## res-B_SY1GJM-B.js
+## res-B_SY1GJM-B-B_SY1GJMMBYNiYRwHTBqq.js
 
 ```js
 
--- a/crates/rolldown/tests/rolldown/function/asset_filenames/mod.rs
+++ b/crates/rolldown/tests/rolldown/function/asset_filenames/mod.rs
@@ -47,7 +47,7 @@ async fn test() {
         }]),
         cwd: Some(cwd),
         asset_filenames: Some(AssetFilenamesOutputOption::String(
-          "[name]-[hash]-[hash:1].js".into(),
+          "[name]-[hash]-[hash:1]-[hash:23].js".into(),
         )),
         ..Default::default()
       },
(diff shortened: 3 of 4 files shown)


IWANABETHATGUY:
Please add a test case

crates/rolldown_common/src/file_emitter.rs
```diff
@@ -187,7 +187,7 @@ impl FileEmitter {
       let mut filename = filename_template.render(
         name,
         Some(extension.unwrap_or_default()),
-        Some(|len: Option<usize>| &hash[..len.unwrap_or(8)]),
+        Some(|len: Option<usize>| &hash[..len.map_or(8, |len| len.min(21))]),
```
IWANABETHATGUY, line 190: rollup set the minimum length as `22`

shulaoda: Related to https://github.com/rollup/rollup/issues/5720, https://github.com/rolldown/rolldown/pull/2740

This limitation exists because we ported `Rollup`’s `xxhash` logic, and we are aligning with the latest `Rollup` behavior.

shulaoda: Since `assetFileNames` is also generated using `xxhash_with_base`, it inherits the same limitation.

IWANABETHATGUY: https://stackblitz.com/edit/github-xwippssu-vnppmf3n?file=src%2Fentry.js, rollup still generate 22 hash char when set `[hash:23]`

shulaoda: Okay. The `assetFileNames` in `rollup` is generated on the `node` side, so it doesn't use the new `xxxhash` function implemented in `rust`, and therefore doesn't encounter this issue.

shulaoda: However, since we are using the `xxxhash`, the issue does exist on our side. Should we consider enforcing consistency?

IWANABETHATGUY: I think it is ok to merge this first. Please open another issue to track it, and discuss it next week.

IWANABETHATGUY (approved):

PR #4020, opened 2025-04-04
refactor(plugin): replace `impl Future` with `async fn` in async hooks

shulaoda:
### Description

Simplify the code.


hyf0 (approved):
Good job!

PR #4030, opened 2025-04-06
perf(rolldown_plugin_json): use `concat_string!` instead of `format!`

shulaoda:
### Description

Avoid unnecessary `format!`.

### DIFF
--- a/crates/rolldown_plugin_json/src/lib.rs
+++ b/crates/rolldown_plugin_json/src/lib.rs
@@ -167,11 +167,11 @@ fn to_esm(data: &Value, named_exports: bool) -> String {
       default_export_rows.push(Cow::Owned(format!("{key}: {value}",)));
     }
   }
-  let default_export_code: String =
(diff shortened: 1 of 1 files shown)


crates/rolldown_plugin_json/src/lib.rs
```diff
@@ -167,11 +167,11 @@ fn to_esm(data: &Value, named_exports: bool) -> String {
       default_export_rows.push(Cow::Owned(format!("{key}: {value}",)));
     }
   }
-  let default_export_code: String =
+
+  let default_export_code =
```
IWANABETHATGUY, line 177: https://github.com/rolldown/rolldown/blob/c28953703da04f6d6c39694b13e0f03cf8930367/crates/rolldown_utils/src/concat_string.rs?plain=1#L46-L56

prefer `concat_string`

IWANABETHATGUY (approved):

PR #4036, opened 2025-04-06
chore(infra): drop `npm-run-all2`

kazupon:
### Description

`pnpm` can run npm-scripts tasks in parallel by specifying regular expressions.
https://pnpm.io/cli/run

There is no longer any need to use `npm-run-all2`.
we can reduce the number of dependencies in `node_modules`.


kazupon:
Oops, I still missed a few things. I'll try to fix those, too.

packages/rolldown/package.json
```diff
@@ -69,11 +69,11 @@
     "bak_build-node": "unbuild",
     "build-node": "tsx ./build.ts",
     "build-types": "tsc -p ./tsconfig.json",
-    "build-js-glue": "run-s build-types build-node build-types-check",
-    "build-native:debug": "run-s build-binding build-js-glue",
-    "build-native:release": "run-s build-binding:release build-js-glue",
-    "build-wasi:debug": "run-s build-binding build-binding:wasi build-node",
-    "build-wasi:release": "run-s build-binding build-binding:wasi:release build-node",
+    "build-js-glue": "pnpm run --workspace-concurrency=1 \"/^build-(types|node|types-check)$/\"",
```
IWANABETHATGUY, line 72: It seems it uses glob to match the npm scripts, can the order of execution be consistent?

kazupon: Hmm, the order of `pnpm run` with regular expressions seems to depend on the order defined in npm scripts.
https://github.com/pnpm/pnpm/pull/5871#discussion_r1089382952

There are two ways to achieve the same order as before without `npm-run-all2`, using only `pnpm`.

1. Reorder the tasks in the npm scripts
2. Use the shell `&&` operator (e.g. `pnpm task1 && pnpm task2`)

The first method has the disadvantage that it requires the maintainer to have that knowledge, and it is not documented in `pnpm run`. I recommend the last method because it is clearer and easier to maintain than the first method, as it executes the scripts in the npm scripts task as written.

shulaoda: The second method seems unfriendly to `windows` users.

IWANABETHATGUY (approved):

packages/rolldown/package.json
IWANABETHATGUY: > Hmm, the order of `pnpm run` with regular expressions seems to depend on the order defined in npm scripts. [pnpm/pnpm#5871 (comment)](https://github.com/pnpm/pnpm/pull/5871#discussion_r1089382952)
> 
> There are two ways to achieve the same order as before without `npm-run-all2`, using only `pnpm`.
> 
> 1. Reorder the tasks in the npm scripts
> 2. Use the shell `&&` operator (e.g. `pnpm task1 && pnpm task2`)
> 
> The first method has the disadvantage that it requires the maintainer to have that knowledge, and it is not documented in `pnpm run`. I recommend the last method because it is clearer and easier to maintain than the first method, as it executes the scripts in the npm scripts task as written.

The first way should be fine if it could guarantee the order.

Issue #3968, opened 2025-03-26
Improve built-in plugins


shulaoda:
Related to #3892 

Improve the current built-in plugins, focusing on `Vite`'s native plugins, to align snapshot test results as closely as possible with `Vite`.

- [x] https://github.com/rolldown/rolldown/issues/3982

sapphi-red:
I think it is also good to try making the tests in vite repo to pass with native plugins. IIRC it doesn't pass some tests.

sapphi-red:
Also the resolver has the most impact to perf so it'd be great if you can prioritize the resolver plugin over other ones.

PR #4042, opened 2025-04-07
refactor(rust): improve `rolldown_plugin_module_preload_polyfill`

shulaoda:
### Description

Related to #3968


IWANABETHATGUY:
This plugin has some breaking change. Can you have a look @sapphi-red?

crates/rolldown_plugin_module_preload_polyfill/README.md
```diff
@@ -0,0 +1,63 @@
...
+  },
+  plugins: [modulePreloadPolyfillPlugin()],
+});
+```
+
+## 🧪 Compatibility
+
+This plugin is originally designed for `Vite` but is more versatile.
+
+Unlike `Vite`, which uses `__VITE_IS_MODERN__` to control the polyfill, this plugin automatically applies the polyfill only when `output.format` is `esm`, matching `Vite`’s behavior.
```
sapphi-red, line 33: It currently injects the polyfill only when `output.format` is `esm` because the modern output uses ESM and the legacy output uses SystemJS. But in the future, that may change (for example, if https://github.com/vitejs/rolldown-vite/pull/80 lands)

crates/rolldown_plugin_module_preload_polyfill/src/lib.rs
```diff
@@ -1,20 +1,16 @@
...
 
 const MODULE_PRELOAD_POLYFILL: &str = "vite/modulepreload-polyfill";
-
 const RESOLVED_MODULE_PRELOAD_POLYFILL_ID: &str = "\0vite/modulepreload-polyfill.js";
 
-const IS_MODERN_FLAG: &str = "__VITE_IS_MODERN__";
-
 #[derive(Debug)]
-pub struct ModulePreloadPolyfillPlugin {
-  pub skip: bool,
```
sapphi-red, line 16: I think we still need this `skip` option to make this part work.
https://github.com/vitejs/rolldown-vite/blob/5f9204a2df876eb4efb97b270c713a955cbc55b8/packages/vite/src/node/plugins/modulePreloadPolyfill.ts#L25-L26

crates/rolldown_plugin_module_preload_polyfill/README.md
sapphi-red, line 1: Do you expect this plugin to be used directly as well? I guess that could make it difficult to make breaking changes while we iterate and would be better to note that it's unstable yet.

sapphi-red (commented):
Since this plugin wouldn't be used with rollup-vite, relying on rolldown specific behavior (using the outputOptions in transform hook) should be ok.

crates/rolldown_plugin_module_preload_polyfill/src/lib.rs
shulaoda: Currently, this is how we use it in `rolldown-vite`, so the `skip` flag is unnecessary and can help reduce redundant hook invocations. What do you think?
```js
modulePreload !== false && modulePreload.polyfill
  ? enableNativePlugin === true
    ? perEnvironmentPlugin(
        'native:modulepreload-polyfill',
        (environment) => {
          if (
            config.command !== 'build' ||
            environment.config.consumer !== 'client'
          )
            return false
          return nativeModulePreloadPolyfillPlugin({
            skip: false,
          })
        },
      )
    : modulePreloadPolyfillPlugin(config)
  : null,
```

crates/rolldown_plugin_module_preload_polyfill/README.md
shulaoda: The main goal is to directly support `Vite`’s usage scenarios. I'm currently working on improving and porting the native plugins, and I'll incorporate your suggestions to make the necessary adaptations and adjustments.

crates/rolldown_plugin_module_preload_polyfill/README.md
shulaoda: Got it.

shulaoda: So do you think we should keep the `__VITE_IS_MODERN__` flag, use a `skip` option to control the behavior, or leave it entirely to the user to handle externally?

crates/rolldown_plugin_module_preload_polyfill/src/lib.rs
sapphi-red: Ah, now I remember doing so. That's true. Yeah, it should be fine to remove this 👍

crates/rolldown_plugin_module_preload_polyfill/README.md
sapphi-red: Then, I think it's better to document that this plugin's main goal is that and is unstable yet. (or removing the README would serve that purpose as well).

crates/rolldown_plugin_module_preload_polyfill/README.md
sapphi-red: I think it's ok now. I wanted to point out that this may change and that's one part of the instability I mentioned in https://github.com/rolldown/rolldown/pull/4042#discussion_r2030426976.

crates/rolldown_plugin_module_preload_polyfill/README.md
shulaoda: Actually, the `README.md` is intended for developers to clarify the purpose of this plugin and highlight the differences from the `Vite` plugin, making it easier to maintain. I will update it to reflect this point.

sapphi-red (approved):

PR #4056, opened 2025-04-08
fix: avoid hang if load error

### LINKED ISSUE #4054: [Bug]: `this.resolve` does not fulfill when an error happens internally in the resolveId hook

1. Open https://stackblitz.com/edit/github-nci4o6nk-81niyagi?file=rolldown.config.mjs,package.json
2. `npm i` runs automatically

### What is expected?

The build ends and an error is thrown.

The build does not finish because `this.resolve` does not fulfill.

### System Info

```Shell
StackBlitz

  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 20.19.0 - /usr/local/bin/node
  npmPackages:
    @rolldown/wasi:  1.0.0-beta.7-commit.25f3c61
```

### Any additional comments?

The detailed condition is:
- `this.load` load a file
- that file has a dependency
- that dependency is resolved by a plugin
- that plugin throws an error

Original issue: https://github.com/vitejs/rolldown-vite/issues/103

underfin:
### Description

close https://github.com/rolldown/rolldown/issues/4054

### DIFF
--- a/crates/rolldown/src/module_loader/module_task.rs
+++ b/crates/rolldown/src/module_loader/module_task.rs
@@ -79,6 +79,12 @@ impl ModuleTask {
     if let Err(errs) = self.run_inner().await {
       self
         .ctx
+        .plugin_driver
+        .mark_context_load_modules_loaded(&ModuleId::new(&self.resolved_id.id), false)
+        .await
+        .expect("Mark context load modules loaded should not fail");
+      self
+        .ctx
         .tx
         .send(ModuleLoaderMsg::BuildErrors(errs.into_vec()))
         .await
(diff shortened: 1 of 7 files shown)


crates/rolldown/src/module_loader/module_task.rs
```diff
@@ -77,6 +77,12 @@ impl ModuleTask {
   #[tracing::instrument(name="NormalModuleTask::run", level = "trace", skip_all, fields(module_id = ?self.resolved_id.id))]
   pub async fn run(mut self) {
     if let Err(errs) = self.run_inner().await {
+      self
+        .ctx
+        .plugin_driver
+        .mark_context_load_modules_loaded(&ModuleId::new(&self.resolved_id.id), false)
+        .await
+        .expect("Mark context load modules loaded should not fail");
```
IWANABETHATGUY, line 85: Use another panic message or use `unwrap` directly. This message looks weird when it is a `Err` variant; anything wrapped with `Result` could be a `Err`

IWANABETHATGUY (approved):

crates/rolldown/src/module_loader/module_task.rs
underfin: The error only happened at call js side function failed, the function is the internal implement, the message maybe is wired for the users, it only used to the developers, it could be clear for me.

PR #4067, opened 2025-04-09
fix: validate hmr option

underfin:


hyf0 (approved):

IWANABETHATGUY:
I remove it from queue because the netlify deploy failed, let's try it again.

PR #4072, opened 2025-04-09
refactor(rust): improve `rolldown_plugin_import_glob`

shulaoda:
### Description

Related to #3968 

- Refactored the code structure and optimized parts of the logic.
- Enabled the test related to `virtual-module`.


IWANABETHATGUY:
IMO, you should not extract a big chunk of code, and modify it in one pr, it is hard to review which part is changed. You could move the code first, and modify it in the next pr.

IWANABETHATGUY (approved):

IWANABETHATGUY:
But you are the owner of the built plugin now, so let's go ahead.

PR #4081, opened 2025-04-11
refactor(rust): improve `rolldown_plugin_wasm_helper`

shulaoda:
### Description

Related to #3968


packages/rolldown/tests/fixtures/builtin-plugin/wasm-helper/_config.ts
```diff
@@ -0,0 +1,15 @@
+
+import { expect } from 'vitest'
+import { defineTest } from 'rolldown-tests'
+import { wasmHelperPlugin } from 'rolldown/experimental'
+
+export default defineTest({
+  config: {
+    plugins: [wasmHelperPlugin()],
```
IWANABETHATGUY, line 8: Why change the test?

shulaoda: The original test case was ineffective, and the `wasmFallbackPlugin` did not function as intended because it only applies to requests ending in `.wasm`, not to those with query strings such as `.wasm?init`.

IWANABETHATGUY: ok, that looks good to me, I left  @underfin to have a look.

underfin (approved):

PR #4085, opened 2025-04-11
refactor(rust): improve `rolldown_plugin_load_fallback`

shulaoda:
### Description

Related to #3968

### DIFF
--- a/crates/rolldown_plugin_load_fallback/src/lib.rs
+++ b/crates/rolldown_plugin_load_fallback/src/lib.rs
@@ -1,5 +1,4 @@
 use std::borrow::Cow;
-use std::fs::read_to_string;
 
 use rolldown_plugin::{HookLoadArgs, HookLoadOutput, HookLoadReturn, Plugin, PluginContext};
 
@@ -20,7 +19,7 @@ impl Plugin for LoadFallbackPlugin {
     let Some(index) = args.id[start..].find(['?', '#']) else { return Ok(None) };
 
     let path = &args.id[..start + index];
-    let code = read_to_string(path)?;
+    let Ok(code) = std::fs::read_to_string(path) else { return Ok(None) };
 
     ctx.add_watch_file(path);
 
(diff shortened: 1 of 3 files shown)


crates/rolldown_plugin_load_fallback/src/lib.rs
```diff
@@ -1,23 +1,29 @@
...
     }
-    let normalized_id = clean_url(args.id);
-    let code =
-      std::fs::read_to_string(normalized_id).or_else(|_| std::fs::read_to_string(args.id))?;
+
+    let start = args.id.rfind('/').unwrap_or(0);
+    let Some(index) = args.id[start..].find(['?', '#']) else { return Ok(None) };
+
+    let path = &args.id[..start + index];
+    let code = read_to_string(path)?;
```
shulaoda: This is different from `Vite` — I think if it fails here, it should fall back to the native logic to load the source.

https://github.com/vitejs/rolldown-vite/blob/03e6286b52f4c1cf9c3ede2366bff685549b3860/packages/vite/src/node/plugins/loadFallback.ts#L19-L28

IWANABETHATGUY: let's just align with vite, the builtin plugin is already native

crates/rolldown_plugin_load_fallback/src/lib.rs
```diff
@@ -1,23 +1,29 @@
...
-    if parse_data_url(args.id).is_some() {
+  async fn load(&self, ctx: &PluginContext, args: &HookLoadArgs<'_>) -> HookLoadReturn {
+    if args.id.trim_start().starts_with("data:") {
       return Ok(None);
     }
-    let normalized_id = clean_url(args.id);
-    let code =
-      std::fs::read_to_string(normalized_id).or_else(|_| std::fs::read_to_string(args.id))?;
+
+    let start = args.id.rfind('/').unwrap_or(0);
```
IWANABETHATGUY: We don't need to align with vite filter, since this plugin is native, please revert the cleanUrl function call.

IWANABETHATGUY (approved):

PR #4087, opened 2025-04-12
feat: add `withFilter` util function

IWANABETHATGUY:
### Description
1. `withFilter` gives the user the abiliity to filter hook call for that ecosystem plugin that are not ready to migrate.

### DIFF
--- a/packages/rolldown/src/plugin/with-filter.ts
+++ b/packages/rolldown/src/plugin/with-filter.ts
@@ -1,44 +1,42 @@
-import { HookFilterExtension, Plugin } from "..";
+import { HookFilterExtension, Plugin } from '..';
 
 type OverrideFilterObject = {
-	transform?: HookFilterExtension<"transform">["filter"];
-	resolveId?: HookFilterExtension<"resolveId">["filter"];
-	load?: HookFilterExtension<"load">["filter"];
+  transform?: HookFilterExtension<'transform'>['filter'];
+  resolveId?: HookFilterExtension<'resolveId'>['filter'];
+  load?: HookFilterExtension<'load'>['filter'];
 };
 // TODO: maybe we should also change the name as well ?
 // my proposal is `${import.meta.filename}@${name}`
 export function withFilter<A, T extends Plugin<A>>(
-	plugin: T,
-	filter_obj: OverrideFilterObject,
+  plugin: T,
+  filter_obj: OverrideFilterObject,
 ): T {
-  plugin['transform']
-	Object.keys(plugin).forEach((key) => {
-		switch (key) {
-			case "transform":
-      case "resolveId":
-      case "load":
+  plugin['transform'];
+  Object.keys(plugin).forEach((key) => {
+    switch (key) {
+      case 'transform':
+      case 'resolveId':
+      case 'load':
         if (!plugin[key]) {
           return;
         }
-				if (typeof plugin[key] === "object") {
-					plugin[key].filter = filter_obj[key] ?? plugin[key].filter;
-				} else {
+        if (typeof plugin[key] === 'object') {
+          plugin[key].filter = filter_obj[key] ?? plugin[key].filter;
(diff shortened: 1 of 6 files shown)


IWANABETHATGUY:
* **#4087** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4087?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4087?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/rolldown/tests/fixtures/plugin/transform/with-filter/_config.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const transformFn = vi.fn();
+const transformFn2 = vi.fn();
+const transformFn3 = vi.fn();
+
+export default defineTest({
+	skipComposingJsPlugin: true,
+	config: {
+		plugins: [
+			withFilter(
```
hyf0, line 13:
```js
import { withFilter } from 'rolldown'
import commonjs from 'commonjs'


1: plugins: [withFilter(commonjs, { id: /.../ })]
2: plugins: [withFilter({ id: /.../ }, commonjs)]
```

NIT: Does the usage 2 is more fit into the current `filter first` intuiation?

hyf0 (approved):

packages/rolldown/src/plugin/with-filter.ts
```diff
@@ -0,0 +1,42 @@
...
+  transform?: HookFilterExtension<'transform'>['filter'];
+  resolveId?: HookFilterExtension<'resolveId'>['filter'];
+  load?: HookFilterExtension<'load'>['filter'];
+};
+// TODO: maybe we should also change the name as well ?
+// my proposal is `${import.meta.filename}@${name}`
+export function withFilter<A, T extends Plugin<A>>(
+  plugin: T,
+  filter_obj: OverrideFilterObject,
+): T {
```
sapphi-red, line 13: This does not work with nested plugins.
For example, a plugin with a type of `[Plugin, Promise<[Plugin, Promise<Plugin>]>]`.
Also, we need a way to specify which plugin the filter should be set to for those cases.

packages/rolldown/src/plugin/with-filter.ts
```diff
@@ -0,0 +1,42 @@
...
+  plugin['transform'];
```
sapphi-red, line 14: Probably this line is not needed?

packages/rolldown/src/plugin/with-filter.ts
```diff
@@ -0,0 +1,42 @@
+import { HookFilterExtension, Plugin } from '..';
+
+type OverrideFilterObject = {
+  transform?: HookFilterExtension<'transform'>['filter'];
+  resolveId?: HookFilterExtension<'resolveId'>['filter'];
+  load?: HookFilterExtension<'load'>['filter'];
+};
+// TODO: maybe we should also change the name as well ?
+// my proposal is `${import.meta.filename}@${name}`
```
sapphi-red, line 9: I think it's better to keep the name. Because there's some plugins that checks the plugin name.
For example: https://github.com/vitejs/vite-plugin-react/blob/87f7fdd63d36e0317b4fa013bae3e15291153f28/packages/plugin-react-swc/src/index.ts#L120-L131

packages/rolldown/src/plugin/with-filter.ts
IWANABETHATGUY: nice catch.

packages/rolldown/src/plugin/with-filter.ts
IWANABETHATGUY: ok

PR #4108, opened 2025-04-14
fix(build): handle empty dependencies in rolldown configuration

CoderSerio:
### Description

The current implementation assumes `pkgJson.dependencies` always exists and attempts to spread its keys into the external array. 

However, some packages might be dependency-free or only have `devDependencies` and `peerDependencies` without any direct runtime dependencies.

> Note: The issue is straightforward - it occurs when `Object.keys` is called on undefined dependencies, so maybe it's ok that there's no reproduction link provided as the fix is self-explanatory :)

### DIFF
--- a/packages/rolldown/build.ts
+++ b/packages/rolldown/build.ts
@@ -102,7 +102,7 @@ function withShared(
       /\.\/rolldown-binding\.wasi\.cjs/,
       // some dependencies, e.g. zod, cannot be inlined because their types
       // are used in public APIs
-      ...Object.keys(pkgJson.dependencies),
+      ...Object.keys(pkgJson.dependencies ?? {}),
       bindingFileWasi,
       bindingFileWasiBrowser,
     ],


packages/rolldown/build.ts
```diff
@@ -37,7 +37,7 @@ const shared = defineConfig({
     /\.\/rolldown-binding\.wasi\.cjs/,
     // some dependencies, e.g. zod, cannot be inlined because their types
     // are used in public APIs
-    ...Object.keys(pkgJson.dependencies),
+    ...(pkgJson.dependencies ? Object.keys(pkgJson.dependencies) : []),
```
IWANABETHATGUY:
```suggestion
    ...Object.keys(pkgJson.dependencies ?? []),
```

packages/rolldown/build.ts
```diff
@@ -37,7 +37,7 @@ const shared = defineConfig({
     /\.\/rolldown-binding\.wasi\.cjs/,
     // some dependencies, e.g. zod, cannot be inlined because their types
     // are used in public APIs
-    ...Object.keys(pkgJson.dependencies),
+    ...Object.keys(pkgJson.dependencies ?? []),
```
IWANABETHATGUY: Sorry for the misleading, it should be `{}`.

CoderSerio: Haha, it's not a big deal, but it makes the project feel more approachable (especially to developers who are new to open source) 😂

IWANABETHATGUY: Can you rebase the branch and it is good to merge.

CoderSerio: ok, I'll finish it in 5 mins

IWANABETHATGUY (approved):

PR #4109, opened 2025-04-14
chore(rolldown_debug): add necessary features for `tracing-subscriber`

shulaoda:
### Description

This is a rather peculiar bug that typically doesn't occur. However, when you add a test case in any plugin crate and run it, the issue will be exposed.

Add below to `crates/rolldown_plugin_load_fallback/src/lib.rs` and run it:

```rust
#[test]
fn test() {
  println!("hello world");
}
```

Then you will get this error:

```rust
error[E0433]: failed to resolve: could not find `fmt` in `tracing_subscriber`
   --> crates\rolldown_debug\src\debug_formatter.rs:11:3
    |
11  |   fmt::{FmtContext, FormatEvent, FormatFields, format::Writer},
    |   ^^^ could not find `fmt` in `tracing_subscriber`
```

### DIFF
--- a/crates/rolldown_debug/Cargo.toml
+++ b/crates/rolldown_debug/Cargo.toml
@@ -12,7 +12,7 @@ serde = { workspace = true, features = ["derive", "rc"] }
 serde_json = { workspace = true }
 tracing = { workspace = true, features = ["valuable"] }
 tracing-serde = { workspace = true }
-tracing-subscriber = { workspace = true, features = ["env-filter", "json", "valuable"] }
+tracing-subscriber = { workspace = true, features = ["env-filter", "fmt", "json", "registry", "std", "valuable"] }
 
 [lints]
 workspace = true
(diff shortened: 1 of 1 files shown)


crates/rolldown_debug/Cargo.toml
```diff
@@ -12,7 +12,7 @@ serde = { workspace = true, features = ["derive", "rc"] }
 serde_json = { workspace = true }
 tracing = { workspace = true, features = ["valuable"] }
 tracing-serde = { workspace = true }
-tracing-subscriber = { workspace = true, features = ["env-filter", "json", "valuable"] }
+tracing-subscriber = { workspace = true, features = ["env-filter", "fmt", "json", "registry", "std", "valuable"] }
```
hyf0: Just add `fmt` and that would be enough.

shulaoda: See https://docs.rs/tracing-subscriber/latest/tracing_subscriber. But you're right, it's confusing.

`fmt`: Enables the [fmt](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/index.html) module, which provides a subscriber implementation for printing formatted representations of trace events. Enabled by default. **Requires “registry” and “std”**.

hyf0 (approved):

hyf0:
Guess we should have a script to check the availability of each crate as a standalone crate in CI. It would be useful if we decide to publish them to crates.io.

PR #4112, opened 2025-04-15
feat: initial implementation of `@rolldown/browser`

### LINKED ISSUE #4012: Browser Support
- [x] https://github.com/rolldown/rolldown/issues/898
- [ ] Node APIs
  - [x] `node:path` -> `pathe`
  - `import { availableParallelism } from "node:os"`
  - `import { Worker } from "node:worker_threads"`
  - `import { Buffer } from "node:buffer"`

sxzz:
### Description

- Rename `@rolldown/wasi` to `@rolldown/browser`
- Add pure browser build `dist/browser.js` and `dist/browser-bundler.js`
  - Disabled parallel js plugin

closes #4012 

## TODO

- Make all node builtin modules optional
- Fix loading binding of `dist/browser.js`: patch `?url`


packages/browser/package.json
```diff
@@ -1,13 +1,13 @@
 {
-  "name": "@rolldown/wasi",
```
hyf0, line 2: Please keep this package. We are using it in stackblitz: https://stackblitz.com/github/rolldown/rolldown-starter-stackblitz?file=package.json.

packages/rolldown/build.ts
```diff
@@ -1,34 +1,41 @@
...
+
 const shared = defineConfig({
   input: {
     index: './src/index',
-    cli: './src/cli/index',
-    'parallel-plugin': './src/parallel-plugin',
-    'parallel-plugin-worker': './src/parallel-plugin-worker',
-    'experimental-index': './src/experimental-index',
-    'parse-ast-index': './src/parse-ast-index',
+    ...isBrowserBuild ? {} : {
```
hyf0: If the browser build doesn't require the exactly same structure as `rolldown` package, you might add new a `build.mts` file in `browser/src` to do the browser build only.

IWANABETHATGUY:
Do we really need this package?
1. `browser` is ambiguous since `wasi` could also run in the browser.
2. What is this package used for other than playgrounds? (If we create an official one, what is this package used for then?) 
3. Adding new package will increase the time of the release CI workflow.

hyf0:
> What is this package used for other than playgrounds? (If we create an official one, what is this package used for then?)

Yeah. If `@rolldown/browser` is ony used in playground and actually does the same thing as `@rolldown/wasi`, maybe `@rolldown/playground` could use `@rolldown/wasi` as the dependency directly.

sxzz:
The term `wasi` is too technical for end-users, and it applies not only to our playground but also to others who want to run Rolldown in a browser (https://x.com/kazu_pon/status/1911730941724131514).

I chose the package name `browser` because Rollup has a similar one: https://www.npmjs.com/package/@rollup/browser.

IWANABETHATGUY:
> The term `wasi` is too technical for end-users, and it applies not only to our playground but also to others who want to run Rolldown in a browser (https://x.com/kazu_pon/status/1911730941724131514).
> 
> I chose the package name `browser` because Rollup has a similar one: https://www.npmjs.com/package/@rollup/browser.

Ah, I see

sxzz:
The time consumption of the release CI workflow is acceptable. We only need to compile a single WASI binary file and copy it to two separate packages, the only difference is uploading on network.

hyf0:
I and @sxzz have a quick meeting:

> browser is ambiguous since wasi could also run in the browser.

We gonna rename @rolldown/wasi to @rolldown/browser

> What is this package used for other than playgrounds? (If we create an official one, what is this package used for then?)

We gonna reuse @rolldown/wasi instead of a new package. But @rolldown/wasi gonna be renamed to @rolldown/browser.

> Adding new package will increase the time of the release CI workflow.

So no new packages.

hyf0:
image
You could trigger this action to test if publish CI is ok.

sxzz:
@hyf0 I have no write permission to trigger it

hyf0:
> @hyf0 I have no write permission to trigger it

Let me do it.

hyf0:
> > @hyf0 I have no write permission to trigger it
> 
> Let me do it.

I can't too. This is forked branch.

hyf0:
Let's merge it first once you think the PR is ready to review and then see if the publish process is ok.

hyf0:
CI failed due to unstable watch tests and trivial reason.

hyf0 (approved):

Inquerito:
Cl is not used for experiments like this, try to compensate for this deficiency with another

Fd929c2CE5fA:
> - Rename `@rolldown/wasi` to `@rolldown/browser`

It would be best to add this migration note in the Readme:
https://www.npmjs.com/package/@rolldown/wasi

It would also be nice if you could add some instructions for use in the new package Readme:
https://www.npmjs.com/package/@rolldown/browser

PR #4113, opened 2025-04-15
chore(rolldown_plugin_load_fallback): add `README.md`

shulaoda:
### Description

Related to #3968


crates/rolldown_plugin_load_fallback/README.md
```diff
@@ -0,0 +1,29 @@
...
+  },
+  plugins: [loadFallbackPlugin()],
+});
+```
+
+## 🧪 Porting Differences
+
+In Vite’s plugin, if reading the cleaned path fails, it falls back to reading the original full ID (including the `query/hash`).
+
+In contrast, Rolldown’s plugin only attempts to read the cleaned path and does not retry the full ID, leaving such cases to be handled by the native bundling logic.
```
IWANABETHATGUY: Both of them are native, I think `rolldown core` maybe less ambiguous.

IWANABETHATGUY (approved):

PR #4123, opened 2025-04-15
refactor(rust): simplify `visit_object_property`

shulaoda:
### Description

When `prop.shorthand` is `true`, `prop.value` must be `ast::Expression::Identifier`.


crates/rolldown/src/module_finalizers/scope_hoisting/impl_visit_mut.rs
```diff
@@ -339,16 +339,14 @@ impl<'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'_, 'ast> {
...
-    match &mut prop.value {
-      ast::Expression::Identifier(id_ref) if prop.shorthand => {
-        if let Some(expr) = self.generate_finalized_expr_for_reference(id_ref, false) {
-          prop.value = expr;
-          prop.shorthand = false;
-        } else {
-          id_ref.reference_id.get_mut().take();
-        }
+    if prop.shorthand {
+      let ast::Expression::Identifier(id_ref) = &mut prop.value else { unreachable!() };
```
hyf0:
```suggestion
      let ast::Expression::Identifier(id_ref) = &mut prop.value else { unreachable!("add reason here") };
```

hyf0 (approved):

PR #4133, opened 2025-04-16
refactor(rolldown_plugin_transform): align `filter` with `rolldown-vite`

shulaoda:
### Description

Related to #3968

PR #4139, opened 2025-04-16
fix(rust): normalize path joining with glob on Windows

shulaoda:
### Description

After submitting PR #4133, I discovered an issue with the `get_matcher_string` function in `pattern_filter`. Upon further investigation of other usages of `fast-glob`, it appears that this issue exists across those as well.

The issue stems from normalizing both the `path` and the `glob` by replacing `\\` with `/` on windows, but `\\` in a `glob` has special meaning, causing potential errors in the merged glob.

If you agree with this, I will apply the fix to other areas as well.

### DIFF
--- a/crates/rolldown_utils/src/join_path_with_glob.rs
+++ b/crates/rolldown_utils/src/join_path_with_glob.rs
@@ -1,0 +1,59 @@
+use std::{borrow::Cow, path::Path};
+
+use crate::pattern_filter::normalize_path;
+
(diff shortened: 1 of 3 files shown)


IWANABETHATGUY:
Did you try rollup behavior?

shulaoda:
> Did you try rollup behavior?

Good point — I hadn't considered that. `Rollup` uses [picomatch](https://github.com/micromatch/picomatch), which indeed doesn't support this feature. I makes sense now.

PR #4158, opened 2025-04-17
fix: clear js side cache at bundle fininsh

underfin:
### Description

The PluginContextData has some cache.
- The `renderChunk#meta#chunks` binding cache
- The `loadModulePromiseMap` stored load successful module

The cache need to clear at hmr or watch rebuild.
- The hmr need to call js side `bundle.write()`.
- The watch rebuild need to call rust `bundle.bundle_write`.

Here using a binding function to do it at bundle finished.

### DIFF
--- a/crates/rolldown/src/bundler.rs
+++ b/crates/rolldown/src/bundler.rs
@@ -213,6 +213,10 @@ impl Bundler {
       .generate_bundle(&mut output.assets, is_write, &self.options, &mut output.warnings)
       .await?;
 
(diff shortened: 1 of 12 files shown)


hyf0 (approved):
Would it be more straightforward to clear them in js side?

underfin:
> Would it be more straightforward to clear them in js side?

It is diffcult because the watch rebuild is trigger at rust side, it hasn't a better way to do it.

PR #4165, opened 2025-04-17
fix: remove `#endregion` comment

btea:
### Description

image

The `//#endregion` comment looks like it is grouped with the following module from the structure, but is grouped with the previous `// #region` comment.

I think a more reasonable approach would be to remove the blank line before the `//#endregion` comment and add a blank line after it. However, removing `#endregion` does not seem to affect the effect of adding the `# region` comment.

I am not familiar with the overall structure. I updated some snap files according to the instructions on the official website, which may lead to some incorrect modifications. Reviewers are welcome to modify them at any time.


hyf0:
Do you know that `#region` is a IDE supported feature to fold code?

btea:
Thanks for the clarification. I had always thought that `#region` comments were just handy for quick side-by-side thumbnails.

hyf0:
Yeah. We use it to let ide fold code of each module. It allows users to inspect the code in a more friendly way.

btea:
Thanks, I learned more. 👍

PR #4168, opened 2025-04-17
feat(debug): add lifecycle-related events

### LINKED ISSUE #4146: [Feature Request][DevTools]: Log Events for Rolldown Lifecycles

Other than plugin hooks like `buildStart` and `buildEnd`, I want a global-level log event like `RolldownBuildStart` and `RolldownBuildEnd` (that only emit once in a build). So I can 

- Know the time it takes to finish the whole process
- If the build is finished or not (in the case of dev mode, for example).
- Also if it would be great to provide additional info like einput ntires, configs / emitted bundle chunks etc, if it would be hard to do in `meta.json` #4134 

Emit log events like `RolldownBuildStart` and `RolldownBuildEnd`

hyf0:
### Description

Closes https://github.com/rolldown/rolldown/issues/4146.

For the need to distinguish if the build is triggered by watch, let's not be bothered untill we have the real need.


hyf0:
* **#4168** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4168?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4168?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/rolldown/src/bundler.rs
```diff
@@ -51,19 +52,25 @@ impl Bundler {
...
 
-    self.bundle_write(scan_stage_output).await
+    let ret = self.bundle_write(scan_stage_output).await;
+    trace_action!(action::BuildEnd { kind: "BuildEnd" });
+    ret
   }
 
   #[tracing::instrument(level = "debug", skip_all, parent = &self.build_span)]
   pub async fn generate(&mut self) -> BuildResult<BundleOutput> {
+    trace_action!(action::BuildStart { kind: "BuildStart" });
```
hyf0, line 66: I realize that these could be abstract as  the start and end of a `Span`. Would optimize them in the future.

IWANABETHATGUY (approved):

PR #4179, opened 2025-04-18
fix: avoid empty line at the beginning of emitted chunks caused by file without imports

### LINKED ISSUE #4150: The bundle file's first line is a blank line.
## description

[reproduction](https://bundler.sxzz.dev/#eNqrVkpWslJS0lHKB1G1ABxkA3s=)

When using `rolldown`, the bundle file's first line is a blank line, but not when using `rollup` and `esbuild`.

I think it's not a functional bug, but an enhancement that's better to have.

Image

marcalexiei:
### Description

Fixes #4150

I updated `render_esm_chunk_imports` to return an `Option<String>` so we can distinguish whether the processed context includes any imports. If it doesn't, no empty line is added to the source output.

This change also removes the `- 1` operation from `source_joiner` `join` function when calculating line length.

Most of the changes in the files are related to snapshot updates, 
so I split them into separate commits to make the review process easier 😅.

- [ ] Check whether this bug might also affect other formats (e.g., CommonJS)


IWANABETHATGUY:
Can you run `just ued` to update snapshot ?

IWANABETHATGUY (approved):
Thanks

marcalexiei:
Hey @IWANABETHATGUY,
I marked the PR as draft since I have checked only ESM code. 
I assume that there is no need to check other formats, but just checking to be sure.

PR #4194, opened 2025-04-18
perf: aovid unnecessary allocate at PreliminaryFilename new

underfin:

PR #4193, opened 2025-04-18
refactor: deconflict asset file names

underfin:
### Description
using `make_unique_name` to deconflict asset file names.

PR #4191, opened 2025-04-18
fix: deconfilct chunk name after name generated

underfin:
### Description
close https://github.com/rolldown/rolldown/issues/4177.

Here follow rollup deconfilct chunk name after name generated, it is used with `bundle` as `used_name_counts`, so here merge the js/css to one map.ref https://github.com/rollup/rollup/blob/master/src/Chunk.ts#L588

Extract it to an isolated function because the asset also could be using it.

PR #4195, opened 2025-04-18
refactor: should deconflict filename if using hash

underfin:

### DIFF
--- a/crates/rolldown_common/src/chunk/mod.rs
+++ b/crates/rolldown_common/src/chunk/mod.rs
@@ -172,12 +172,7 @@ impl Chunk {
 
     let filename = filename_template.render(Some(chunk_name), None, hash_replacer).into();
 
-    let name = if has_hash_pattern {
-      make_unique_name(&filename, used_name_counts);
-      filename
-    } else {
-      make_unique_name(&filename, used_name_counts)
-    };
+    let name = make_unique_name(&filename, used_name_counts);
 
     Ok(PreliminaryFilename::new(name, hash_placeholder))
   }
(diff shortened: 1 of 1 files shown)


underfin:
* **#4195** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4195?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4195?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4194** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4194?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4193** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4193?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4191** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4191?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

crates/rolldown_common/src/chunk/mod.rs
```diff
@@ -172,12 +172,7 @@ impl Chunk {
 
     let filename = filename_template.render(Some(chunk_name), None, hash_replacer).into();
 
-    let name = if has_hash_pattern {
-      make_unique_name(&filename, used_name_counts);
-      filename
```
underfin, line 177: Here the using `make_unique_name` to deconflict filename, but not using it retured value as filename, it is an unexpected error cc @hyf0

hyf0: I remember I did it on purpose, but I can't think of why.

> but not using it retured value

Because rollup works in this way. It makes the expected `name-xxxxx.js` becomes `name2-xxxx.js`. With hash, the final filename should not be conflicted and no need to deconflict chunk name.

hyf0 (approved):

crates/rolldown_common/src/chunk/mod.rs
underfin: let's always deconflict to aovid conflict if enable hash. I think it maybe happend.

PR #4218, opened 2025-04-20
refactor(normal_module): refactor ESM namespace generation

ShroXd:
### Description

Refactor `generate_esm_namespace_in_cjs_stmt` and `generate_esm_namespace_in_cjs_node_mode_stmt`

- Introduces `EsmNamespaceKind` enum to replace magic strings for debug labels, improving type safety and maintainability
- Extracts common logic into the private helper `generate_esm_namespace_in_cjs_internal` to reduce code duplication

### DIFF
--- a/crates/rolldown_common/src/module/normal_module.rs
+++ b/crates/rolldown_common/src/module/normal_module.rs
@@ -363,7 +363,7 @@ pub enum EsmNamespaceKind {
 }
 
 impl EsmNamespaceKind {
-  pub fn debug_label(&self) -> &'static str {
+  pub fn debug_label(self) -> &'static str {
     match self {
       EsmNamespaceKind::Default => "esm_namespace_ref_derived_from_module_exports",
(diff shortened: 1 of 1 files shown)


crates/rolldown_common/src/module/normal_module.rs
```diff
@@ -356,3 +355,18 @@ impl DerefMut for NormalModule {
 pub enum ModuleRenderArgs<'any> {
   Ecma { ast: &'any EcmaAst },
 }
+
+#[derive(Debug, Clone, Copy)]
+pub enum EsmNamespaceKind {
```
hyf0: Could you keep it as `bool`?

ShroXd: > Could you keep it as `bool`?

Nice point! I can refactor this part to avoid over-engineering. Thanks for pointing it out!

hyf0 (approved):
Thank you!

PR #4262, opened 2025-04-23
feat: validate hoistTransitiveImports

### LINKED ISSUE #4250: Allow `hoistTransitiveImports: false` in option validation
Rolldown's behavior is same with `hoistTransitiveImports: false` so there's no need to warn users when this option is set (probably `hoistTransitiveImports: true` should still be warned).
Allowing this without a flag would help users using Vite through a framework to try Rolldown.

underfin:
### Description
close https://github.com/rolldown/rolldown/issues/4250

### DIFF
--- a/packages/rolldown/tests/validate-option.test.ts
+++ b/packages/rolldown/tests/validate-option.test.ts
@@ -42,12 +42,32 @@ test('validate output option', async () => {
     await bundle.write({
       // @ts-ignore  invalid key
       foo: 'bar',
+      hoistTransitiveImports: false
     })
     expect.unreachable()
   } catch (error: any) {
     expect(error.message).toMatchInlineSnapshot(`
         "Failed validate output options.
         - For the "foo". Invalid key: Expected never but received "foo". "
+      `)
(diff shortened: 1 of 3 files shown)


packages/rolldown/tests/validate-option.test.ts
```diff
@@ -51,3 +52,22 @@ test('validate output option', async () => {
...
+    await bundle.write({
+      // @ts-ignore  invalid value
+      hoistTransitiveImports: true
+    })
+    expect.unreachable()
+  } catch (error: any) {
+    expect(error.message).toMatchInlineSnapshot(`
+        "Failed validate output options.
+        - For the "hoistTransitiveImports". The 'true' value is not supported. "
+      `)
```
underfin, line 71: @sapphi-red Give a error could be acceptable?

sapphi-red: Yes, it's ok.

sapphi-red (approved):
LGTM 👍

PR #4316, opened 2025-04-25
Revert "ci: publish to pkg.pr.new on every main commit"

IWANABETHATGUY:
Reverts rolldown/rolldown#4312


IWANABETHATGUY:
This would block the ci when more than 2 or 3 active pr, let's find a better solution

PR #4331, opened 2025-04-26
chore: remove `preinstall` script

btea:
### Description

Remove the `preinstall` script in the rolldown package, otherwise, the following warning message will appear when using pnpm to install rolldown, and relevant configuration is required.

image

In fact, this script has no logical modification, it is only used to verify the package management tool during development, so I think it can be removed. The corresponding script can be included in the package.json file in the project root directory.

### DIFF
--- a/packages/rolldown/package.json
+++ b/packages/rolldown/package.json
@@ -71,8 +71,7 @@
     "build-browser:debug": "cross-env BROWSER_PKG=1 pnpm run --sequential \"/^build-(binding|binding:wasi|node)$/\"",
     "build-browser:release": "cross-env BROWSER_PKG=1 pnpm run --sequential \"/^build-(binding|binding:wasi:release|node)$/\"",
     "# Scrips for docs #": "_",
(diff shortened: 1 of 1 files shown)


packages/rolldown/package.json
```diff
@@ -71,8 +71,7 @@
     "build-browser:debug": "cross-env BROWSER_PKG=1 pnpm run --sequential \"/^build-(binding|binding:wasi|node)$/\"",
     "build-browser:release": "cross-env BROWSER_PKG=1 pnpm run --sequential \"/^build-(binding|binding:wasi:release|node)$/\"",
     "# Scrips for docs #": "_",
-    "extract-options-doc": "typedoc",
-    "preinstall": "npx only-allow pnpm"
```
hyf0, line 75: Should made it as a part of `just setup`.

btea: Sorry, I don't know how to do this. Could you please continue?

hyf0 (approved):

PR #4346, opened 2025-04-27
feat: implement filter expr filtering

IWANABETHATGUY:
### Description
The limitation of current hook filter: 
1. You can't specify test `include` before `exclude`, here is an example: 
```js

function hookHandler(id) {
  if (aRegexp.test(id)) {
    executeSomething();
  }
  if (bRegexp.test(id)) {
    return;
  }
}
```
You may write such filter: 
```js
{
  include: [aRegexp],
  exclude: [bRegexp]
}
```
This is correct if and only if `aRegexp` doesn't intersect with `bRegexp` because `exclude` is always executed before `include` and will return false if it matches.
real-world example: 
https://github.com/vitejs/rolldown-vite/blob/91a494cb70c0fe21222261878770b23500e9ef41/packages/vite/src/node/plugins/asset.ts?plain=1#L199-L213
2. Express  `or` pattern with multiple different field in transform filter with `include`, e.g.
```js
function hookHandler(id, moduleType) {
  if (moduleType == "js" || /\.js$/.test(id)) {
    doSomething();
  }
}
``` 
You may write filter like: 
```js
{
  moduleType: ['js'],
  id: /\.js$/
}
```
But this did not work at all, copy from documentation
> If multiple filter properties are specified, the filter matches when all of the specified properties match. In other words, if even one property fails to match, it is excluded, regardless of the other properties.

Another real-world example would be: 
https://github.com/vitejs/rolldown-vite/blob/fef84b75dbb35a6ec27debdc0dced1d0f1250eb8/packages/vite/src/node/plugins/importAnalysisBuild.ts?plain=1#L242-L244
3. Don't support preprocessor: for example: 
it is commonly in vite to `cleanUrl` before filter the id: 
  - https://github.com/vitejs/vite-plugin-react/blob/7517103485081b26004e79f169efdd2d12a60946/packages/plugin-react/src/index.ts#L189
  - https://github.com/vitejs/rolldown-vite/blob/91a494cb70c0fe21222261878770b23500e9ef41/packages/vite/src/node/plugins/asset.ts?plain=1#L211
  But for now, you need to convert it by your self, this is how vite did: 
  https://github.com/vitejs/rolldown-vite/pull/118/files#diff-91776e7c6039d23a070162f02a69cd46046a2095bd5ecb384ae9e27f2ea5288fR185-R191
  Another example: https://github.com/sveltejs/vite-plugin-svelte/blob/3589433cd19464c484f560516d41e670e5d40710/packages/vite-plugin-svelte/src/index.js#L166-L168, 
  It would be harder to write a regular expression to represent that filter logic.
  
Another benefit of `filterExpr` is the way you write `filterExpr` is  more likely you write javascript, e.g.

```js
  if (aReg.test(id) && bReg.test(id)) {
    return;
  }

  if (cReg.test(code) && !dReg.test(code)) {
    doSomething();
  }

```

the related `filterExpr` would be: 

```js
filter: {
  custom: [
    exclude(and(id(aReg), id(bReg))),
    include(and(code(cReg), not(code(dReg))))
  ]
}
```
They are isomorphic in some sense!


IWANABETHATGUY:
* **#4346** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4346?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4346?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4343** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4343?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4340** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4340?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4338** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4338?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4323** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4323?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/rolldown/tests/fixtures/tree-shake/filter-expression/_config.ts
```diff
@@ -0,0 +1,25 @@
...
+	config: {
+		plugins: [
+			{
+				name: "filter-expr",
+				transform: {
+					handler() {
+						transformHookFunction();
+					},
+					filter: {
+						custom: [exclude(and(id(/src/), not(code(/import\s+{/))))],
```
hyf0, line 16: This looks like so powerful and unlocks much potential abilities. Let's raise it up in the meeting and see how it goes.

packages/rolldown/tests/fixtures/tree-shake/filter-expression/_config.ts
```diff
@@ -0,0 +1,25 @@
+import { defineTest } from "rolldown-tests";
+import { and, code, exclude, id, not } from "rolldown/filterExpr";
```
hyf0:
```suggestion
import { and, code, exclude, id, not } from "rolldown/filter-expr";
```

kebab case is more commonly used for package name.

crates/rolldown_binding/src/options/plugin/js_plugin.rs
```diff
@@ -32,10 +35,18 @@ pub struct JsPluginContextResolveCustomArgId;
...
+  ResolveId,
+  Transform,
+  Load,
+}
 
 #[derive(Debug)]
 pub struct JsPlugin {
   pub(crate) inner: BindingPluginOptions,
+  /// Since there at most three key in the cache, use vec should always faster than hashmap
+  pub(crate) filter_expr_cache: Vec<(FilterExprCacheKey, Vec<FilterExprKind>)>,
```
hyf0, line 46: Not from performance aspect. Would it be more straight if you just put them like

``` rust
struct JsPlugin {
  transform_filter: ...,
  load_filter: ...,
  resolve_id_filter: ...,
}
```

since you are already do the enumerating in `FilterExprCacheKey`.

packages/rolldown/tests/fixtures/tree-shake/filter-expression/_config.ts
hyf0: You want to be careful about these names. Using short names would be concise, but some of them are really contextual.

Wether to use the same name such as `id()` to represent separate meaning depending on the context is tricky.

```js
{
  resolveId: {
    filter: [include(id(...))]
  },
  transform: {
    filter: [include(id(...))]
  },
  renderChunk: {
    filter: [include(id(...))]
  },
}
```

```js
{
  resolveId: {
    filter: [include(importSpecifier(...))],
    // or
    filter: [include(moduleRequest(...))],
  },
  transform: {
    filter: [include(moduleId(...))]
  },
  renderChunk: {
    filter: [include(chunkId(...))]
  },
}
```

hyf0: We might also introduce new concept like `full_id` for module when trying to support type assertions `from '...' with { type: 'css' }`.

IWANABETHATGUY: https://github.com/rolldown/rolldown/blob/9722d32fec5f053a8433a614505610ee1360fb4e/packages/rolldown/package.json?plain=1#L49-L52, I just refer this `exports`, maybe we should talk about the naming convention for `exports` field

hyf0: On the second thought, `rolldown/filter` would be concise.

hyf0 (approved):

IWANABETHATGUY:
All api design and refactoring related, I left it in another pr.

Issue #4249, opened 2025-04-22
[Feature Request]: Support symlinked paths for loading rolldown config


Q16solver:
Hi, I wanted to use shared rolldown config paths across my monorepo the same way that rollup does using symlinks, but I found that this doesn't support it yet, it would be amazing if we can add this, perhaps this may also be a good chance to use unconfig? since it was just fixed as well there along with tsdown

https://github.com/antfu-collective/unconfig/commit/f86b9eae2df3fd0ebfbc179c0b4ca75a4692dd39
https://github.com/rolldown/tsdown/issues/131

> Tried to create this with the feature request template, but for some reason I kept getting unable to create issue error, so had to create it as a blank issue :c

### Proposed API

Using loadConfig from unconfig unless there are blockers/other reasons for not doing so.
It should be able to automatically search for the config file like in tsdown as explained here https://tsdown.dev/guide/config-file for all different types of .js/.ts etc files

hyf0:
Since our config loding logic is purely on js side, this could be a good first issue.

shuga2704:
Tried to contribute, but found that issue is gone. Works as expected with my monorepo setup.

Q16solver:
Yeah, I think it works with --config, now the issue is just automatically loading it without needing to specifically specify the path

shuga2704:
> Yeah, I think it works with --config, now the issue is just automatically loading it without needing to specifically specify the path

So there is no chance to reproduce the original issue, if I am not mistaken?

Q16solver:
Yeah, I think we can close this one, the automatic config discovery support is probably a separate issue

sapphi-red:
Closing as this seems to be already fixed.

PR #4357, opened 2025-04-28
feat: CLI supports automatic search for the config file (#4249)

kingsword09:
### Description

CLI supports automatic search for the config file (#4249)

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,8 @@
     "lint-staged": "^15.2.5",
     "oxlint": "^0.16.0",
     "rolldown": "workspace:*",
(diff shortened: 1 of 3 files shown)


underfin:
The `unconfig` is using [jiti](https://github.com/unjs/jiti) to excute config file, here has a mini problem the `jiti` look like is using `bable`, ref https://github.com/unjs/jiti/blob/main/src/babel.ts. It will pollute the rolldown dependencies. cc @hyf0

hyf0:
Seems not worth it.

Boshen:
Close as not planned.

PR #4375, opened 2025-04-29
perf(rolldown_plugin_json): change `register_hook_usage` to return `HookUsage::Transform`

shulaoda:
### Description

Related to #3968

## Summary by CodeRabbit

- **Refactor**
  - Updated internal plugin behavior to use only the Transform hook, optimizing how plugins are registered. No visible changes to user-facing features.

### DIFF
--- a/crates/rolldown_plugin_json/src/lib.rs
+++ b/crates/rolldown_plugin_json/src/lib.rs
@@ -75,7 +75,6 @@ impl Plugin for JsonPlugin {
   }
 
   fn register_hook_usage(&self) -> rolldown_plugin::HookUsage {
-    // TODO: get hook usage from option
-    HookUsage::all()
+    HookUsage::Transform
   }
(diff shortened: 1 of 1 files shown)


crates/rolldown_plugin_json/src/lib.rs
```diff
@@ -75,7 +75,6 @@ impl Plugin for JsonPlugin {
   }
 
   fn register_hook_usage(&self) -> rolldown_plugin::HookUsage {
-    // TODO: get hook usage from option
-    HookUsage::all()
```
shulaoda, line 79: Was there any particular concern or consideration for not using `HookUsage::Transform` initially?

hyf0: I think it's just in case.

hyf0 (approved):

PR #4377, opened 2025-04-29
chore: bundle `valibot`

beeequeue:
### Description

Valibot is a library with quite a lot of surface area (2MB)

However, it is excellent for treeshaking and bundling - so this PR bundles `valibot`.

This makes its size on disk for rolldown go from 2MB -> 33kB~

## Summary by CodeRabbit

- **Chores**
  - Updated dependencies by removing unused packages from the browser package and shifting certain dependencies in the rolldown package to development-only status. No changes to user-facing functionality.

### DIFF
--- a/packages/rolldown/build.ts
+++ b/packages/rolldown/build.ts
@@ -105,8 +105,6 @@ function withShared(
       /rolldown-binding\..*\.wasm/,
       /@rolldown\/binding-.*/,
       /\.\/rolldown-binding\.wasi\.cjs/,
-      // some dependencies, e.g. zod, cannot be inlined because their types
-      // are used in public APIs
       ...Object.keys(pkgJson.dependencies ?? {}),
     ],
     define: {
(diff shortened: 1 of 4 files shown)


packages/rolldown/build.ts
```diff
@@ -105,8 +105,6 @@ function withShared(
       /rolldown-binding\..*\.wasm/,
       /@rolldown\/binding-.*/,
       /\.\/rolldown-binding\.wasi\.cjs/,
-      // some dependencies, e.g. zod, cannot be inlined because their types
-      // are used in public APIs
```
beeequeue, line 109: I saw this comment, but I couldnt find any types that use valibot, so I asssume it no longer applies

underfin (approved):

underfin:
> This makes its size on disk for rolldown go from 2MB -> 33kB~

Nice catching, thanks a lot!

packages/rolldown/build.ts
hyf0: We used to drived types from zod-schema, so we need the zod's types. Now we declare types manually, so it won't be an issue now.

PR #4384, opened 2025-04-30
chore(rolldown_loader_utils): remove unused `json_to_esm`

shulaoda:
### Description

Remove unused function.

## Summary by CodeRabbit

- **Removed Features**
  - Removed the ability to convert JSON strings to ECMAScript module (ESM) source strings. This functionality is no longer available in the application.
- **Chores**
  - Cleaned up dependencies to remove unused packages.


crates/rolldown_loader_utils/src/json_to_esm.rs
```diff
@@ -1,38 +0,0 @@
-use std::fmt::Write as _;
-
-use rolldown_utils::ecmascript::is_validate_assignee_identifier_name;
-use serde_json::{Value, to_string_pretty};
-
-// TODO: handling https://github.com/tc39/proposal-json-superset
-pub fn json_to_esm(json: &str) -> serde_json::Result<String> {
-  // TODO: use zero-copy deserialization
```
hyf0, line 8: Now we use what? Did we achieve this?

hyf0 (approved):

crates/rolldown_loader_utils/src/json_to_esm.rs
shulaoda: It seems that the function `json_object_expr_to_esm` is currently used in `link_stage`'s `generate_lazy_export`, but it is no longer used anywhere.

PR #4388, opened 2025-04-30
feat: add watcher.off method to remove event listener

### LINKED ISSUE #4382: [Feature Request]: `watcher.off` method to remove event listener

Rollup has `watcher.off` method.
https://github.com/rollup/rollup/blob/cb8f81530279f1262d692fde2d4cecc758cadd46/src/rollup/types.d.ts#L1034-L1037

Vite calls this in tests.

Add `watcher.off` method.

hnrq:
### Description
Add `watcher.off` method to `WatcherEmitter`, fixes #4382

## Summary by CodeRabbit

- **New Features**
  - Added the ability to remove previously registered event listeners for watcher events.

- **Tests**
  - Introduced tests to verify correct addition and removal of event listeners.

### DIFF
--- a/packages/rolldown/tests/watch/watch.test.ts
+++ b/packages/rolldown/tests/watch/watch.test.ts
@@ -1,4 +1,4 @@
-import { expect, test, vi } from 'vitest'
+import { expect, Mock, test, vi } from 'vitest'
 import { watch, RolldownWatcher } from 'rolldown'
 import fs from 'node:fs'
 import path from 'node:path'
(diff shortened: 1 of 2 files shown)


ghiscoding:
@hnrq just a side note, if you want your PR to auto-link and auto-close the referenced issue, you need to use specific keywords like "fix [issue number]" or "fixes [issue number]"... it doesn't currently work with the keywords you used 😉

hnrq:
> @hnrq just a side note, if you want your PR to auto-link and auto-close the referenced issue, you need to use specific keywords like "fix [issue number]" or "fixes [issue number]"... it doesn't currently work with the keywords you used 😉

Thanks for the heads up! I think I got it right now :)

ghiscoding:
@hnrq you do 😸 

image

hnrq:
@underfin can I expose the WatcherEvent type on dist so I don't have to use `any` on tests?

underfin:
> @underfin can I expose the WatcherEvent type on dist so I don't have to use `any` on tests?

Yeah. You could export it.

For the testing, using the internal `WatcherEmitter.listeners` API, i will refactor it make it private.

packages/rolldown/tests/watch/watch.test.ts
```diff
@@ -161,6 +161,59 @@ test.sequential('watch event', async () => {
...
+    change: vi.fn(),
+  }
+  
+  for(let evt in events) watcher.on(evt as any, events[evt])
+
+  fs.writeFileSync(input, 'console.log(12)')
+  await waitUtil(() => {
+    expect(events.event).toHaveBeenCalled()
+    expect(events.restart).toHaveBeenCalled()
+  })
```
hnrq: @underfin I've refactored the test so it doesn't rely on `WatcherEmitter.listeners`, but `expect(events.change).toHaveBeenCalled()` fails here, any ideas on why?

underfin: Because the first build event also will trigger it, but it not emit change event. I updated the test to make the test simply. Thank you.

underfin (approved):

PR #4428, opened 2025-05-06
feat: expose `loadConfig` in api

### LINKED ISSUE #4313: [Feature Request]: Expose `loadConfigFile` to programmatic API

Being able to easily load `.config.ts` files without having to manually transpile them first would be useful!

[Mocha for example](https://github.com/mochajs/mocha/blob/945d6e3bf5a9de19c3aa26fbdac966a721006b58/scripts/karma-rollup-plugin.js#L38) does this with rollup

```ts
import { rolldown, loadConfig } from "rolldown"

const config = await loadConfig("./rolldown.config.ts")

rolldown.build(config)
```

beeequeue:
Closes #4313

### Description

This PR moves `loadConfig` from `cli/` to `utils/`, and exposes it to be imported by users.

It also adds a `rolldown/config` entrypoint which would in theory allow only importing code related to configuration rather than everything, but I'm doubtful it's helpful when `loadConfig` imports the compiler anyways.

Will test if it creates a smaller entrypoint, but if it doesn't or maintainers also don't think it will be helpful I'll remove it


packages/rolldown/src/config.ts
```diff
@@ -0,0 +1,8 @@
+import { version } from '../package.json';
+
+import { defineConfig } from './utils/define-config';
+import { loadConfig } from './utils/load-config';
+
+export { defineConfig, loadConfig };
+
+export const VERSION: string = version;
```
underfin, line 8: The `packages/rolldown/src/index.ts` already export `VERSION `

packages/rolldown/build.ts
```diff
@@ -87,6 +87,7 @@ function withShared(
   return {
     input: {
       index: './src/index',
+      ...isBrowserBuild ? { config: './src/config' } : {},
       ...!isBrowserBuild
```
underfin, line 92: Why only add the input to browser build?

beeequeue: I think I assumed the browser build would be UMD and wouldn't support subpath imports, and had a brainfart and added the opposite...

packages/rolldown/src/config.ts
beeequeue: I added it to this file since it's intended as a separate entrypoint (`rolldown/config`) to not have to import the entire `rolldown`, but like I said I'm not sure it would be helpful since `loadConfig` imports the compiler anyways

packages/rolldown/build.ts
sxzz: Browser build is ESM for bundlers like Vite, and without node built-in modules.

packages/rolldown/src/index.ts
```diff
@@ -97,8 +97,9 @@ import type {
 import type { ExistingRawSourceMap, SourceMapInput } from './types/sourcemap';
 import type { PartialNull } from './types/utils';
 import { defineConfig } from './utils/define-config';
+import { loadConfig } from './utils/load-config';
 
-export { build, defineConfig, rolldown, watch };
+export { build, defineConfig, loadConfig, rolldown, watch };
```
sxzz: Remove `VERSION` export, and add
```ts
export * from './config';
```

packages/rolldown/build.ts
```diff
@@ -88,6 +88,7 @@ function withShared(
   return {
     input: {
       index: './src/index',
+      config: './src/config',
```
sxzz: `node:fs` is used in `loadConfig`, which cannot run on browser.

packages/rolldown/src/index.ts
```diff
@@ -96,10 +95,9 @@ import type {
 } from './types/rolldown-output';
 import type { ExistingRawSourceMap, SourceMapInput } from './types/sourcemap';
 import type { PartialNull } from './types/utils';
-import { defineConfig } from './utils/define-config';
 
-export { build, defineConfig, rolldown, watch };
-export const VERSION: string = version;
+export * from './config';
```
sxzz: `loadConfig` is also exported here, which cannot run on browser build.

Maybe we should not export all, but 

```suggestion
export { defineConfig, VERSION } from './config';
```

beeequeue: The initial purpose of the PR is to expose `loadConfig`, I guess doing it through `/config` is easier than doing some conditional based on if it's a browser build or not?

sxzz (approved):

packages/rolldown/src/config.ts
sxzz: @underfin browser build includes `index.ts` entry, and it cannot include fs API (`loadConfig`) from `config.ts`.

underfin (approved):

PR #4462, opened 2025-05-09
feat(rolldown): oxc_resolver v9.0.0

Boshen:
I removed the `modules` option in a previous version, it's now restored.


underfin:
Look like it's should be oxc-resolver 9.0.0?

underfin (approved):

PR #4466, opened 2025-05-09
test(pluginutils): add test for `exactRegex` and `prefixRegex`

sapphi-red:
### Description

test for #4458

### DIFF
--- a/packages/pluginutils/src/simple-filters.test.ts
+++ b/packages/pluginutils/src/simple-filters.test.ts
@@ -1,3 +1,4 @@
+// cSpell:ignore fooa, afoo
 import { describe, expect, test } from 'vitest';
 import { exactRegex, prefixRegex } from './simple-filters.js';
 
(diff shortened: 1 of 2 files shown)


hyf0 (approved):

packages/pluginutils/src/simple-filters.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    expect(regex.test('foo(bar)')).toBe(true);
+    expect(regex.test('foo(bar\\)')).toBe(false);
+    expect(regex.test('foo(bar)a')).toBe(false);
+    expect(regex.test('afoo(bar)')).toBe(false);
+  });
+});
+
+describe('prefixRegex', () => {
+  test('supports without flag parameter', () => {
+    const regex = prefixRegex('foo');
```
IWANABETHATGUY, line 38: Would it be better to name this utility to `startsWithRegex` ?

IWANABETHATGUY (approved):

packages/pluginutils/src/simple-filters.test.ts
sapphi-red: I'll prefer `prefixRegex` as it's shorter while expresses the behavior. `startsWith` is easy to understand since `.startsWith` exists, but I think `prefix` is easy enough.

IWANABETHATGUY: make sense

PR #4469, opened 2025-05-09
feat(pluginutils): add `makeIdFiltersToMatchWithQuery` function

sapphi-red:
### Description

This PR adds `makeIdFiltersToMatchWithQuery` function to `@rolldown/pluginutils`.

This function is useful for codes like:
```js
const cleanId = cleanUrl(id)
if (filter(id) || filter(filepath)) {
  // process
}
```


IWANABETHATGUY:
LTGM, but I am planning to add some atom exprs something called `Query` and `CleanUrl`, which should be more performant than the current one(Complex regex usually has more overhead than a simple one)

IWANABETHATGUY (approved):

PR #4485, opened 2025-05-10
feat: expose oxc-resolver

sxzz:
### Description

Expose oxc-resolver in `rolldown/experimental`

Use case: `rolldown-plugin-dts` needs to use it to resolve dts files. See https://github.com/sxzz/dts-resolver

If we expose it in rolldown, we can reduce bundle size of plugins.


Boshen (approved):
For the record, can you add a PR description on where you intend to use these APIs?

I'll wait for @rolldown/rolldown to merge.

eyestoreoye:
Very clean code with good test coverage. This is a great example of best practices.

shulaoda (approved):

PR #4487, opened 2025-05-10
perf(rename): reduce some string allocations

Boshen:
While profiling I noticed some short lived string allocations

image

The best I could do is reduce 1 allocation before testing for collision.

### DIFF
--- a/crates/rolldown/src/utils/renamer.rs
+++ b/crates/rolldown/src/utils/renamer.rs
@@ -138,21 +138,22 @@ impl<'name> Renamer<'name> {
         let binding_ref: SymbolRef = (module.idx, symbol_id).into();
 
         let mut count = 1;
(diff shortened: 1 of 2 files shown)


crates/rolldown/src/utils/renamer.rs
```diff
@@ -138,21 +138,22 @@ impl<'name> Renamer<'name> {
         let binding_ref: SymbolRef = (module.idx, symbol_id).into();
 
         let mut count = 1;
-        let mut candidate_name = binding_name.to_rstr();
+        let mut candidate_name = Cow::Borrowed(binding_name);
```
Boshen, line 141: Reduce 1 allocation / some cpus by not converting to compact string.

shulaoda (approved):

PR #4488, opened 2025-05-10
fix(browser): fix binding types

### LINKED ISSUE #4753: [Bug]: `transform` option type is referencing an option that doesn't exist for `@rolldown/browser`

`@rolldown/browser/dist/shared/define-config.d-BBxr6_Bs.d.mts` (https://cdn.jsdelivr.net/npm/@rolldown/browser@1.0.0-beta.10/dist/shared/define-config.d-BBxr6_Bs.d.mts) has
```ts
import { /* omit other imports */ TransformOptions } from "./binding.d-Bl5Vi4GH.mjs";
type OxcTransformOption = Omit<TransformOptions, "sourceType" | "lang" | "cwd" | "sourcemap" | "define" | "inject">;
interface InputOptions {
  transform?: OxcTransformOption;
  // omit other properties
}
```
But `@rolldown/browser/dist/shared/binding.d-Bl5Vi4GH.d.mts` is not included in the published package.

### What is expected?

The type is correct.

The type references a missing type.

### System Info

```Shell
@rolldown/browser@1.0.0-beta.10
```

### Any additional comments?

_No response_

sxzz:
### Description

Upgrade rolldown-plugin-dts to the latest version, fixing binding types

~~Wait https://github.com/oxc-project/oxc-node/issues/130~~ Switched to `tsx` for `conditions` support.

closes #4753, closes #4754, closes #4724


pnpm-workspace.yaml
```diff
@@ -23,3 +23,5 @@ catalog:
   'oxc-parser': '0.69.0'
   'oxc-transform': '0.69.0'
   'tsdown': '0.11.1'
+  'rolldown': '1.0.0-beta.8-commit.534fde3'
```
hyf0: This actually changes from referencing local rolldown to npm rolldown. Is this expected?

We expect to reference local rolldown to see any effect immediately.

sxzz: WIP, not finish yet.

eyestoreoye:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

sxzz:
@hyf0 Since https://github.com/oxc-project/oxc-node/issues/130 has not been resolved for some time, I’ve switched to using npm Rolldown solely for the dts plugin. For bundling, we’re still using the source code version imported from `./src/index`.

hyf0:
I'm not sure if we could do that.

1. Every commit must be ensured to be able to compile rolldown itself
2. Local tests need to run rolldown that builded from local with latest changes instead of the npm one.

packages/rolldown/package.json
```diff
@@ -127,8 +127,8 @@
     "glob": "^11.0.0",
     "oxc-parser": "catalog:",
     "remeda": "^2.10.0",
-    "rolldown": "workspace:*",
-    "rolldown-plugin-dts": "0.7.12",
+    "rolldown": "catalog:",
```
hyf0: It seems to be a bad smell on having different semantics for `rolldown: workspace:*` and `rolldown: "catalog"`.

hyf0: Just smell, let's foucsing on the real problem first.

sxzz: Intentionally relying on `npm:rolldown@nightly`.

hyf0 (approved):
Well done!

PR #4499, opened 2025-05-11
feat(rolldown_common): precompute `ModuleId` hash

Boshen:


crates/rolldown_common/src/types/module_id.rs
```diff
@@ -59,6 +73,36 @@ impl From<ArcStr> for ModuleId {
   }
 }
 
+impl Hash for ModuleId {
+  fn hash<H: Hasher>(&self, state: &mut H) {
+    self.precomputed_hash.hash(state);
```
Boshen, line 78: We need to change all hashmaps to use `IdentityHasher`, otherwise this call double hashes.

PR #4501, opened 2025-05-11
chore(Cargo.toml): disable `arcstr` features

Boshen:
`arcstr` enables `substr` feature by default but we don't use it.

https://docs.rs/crate/arcstr/latest/features

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -124,7 +124,7 @@ rolldown_workspace = { version = "0.1.0", path = "./crates/rolldown_workspace" }
 
 anyhow = "1.0.86"
 append-only-vec = { version = "0.1.5" }
-arcstr = "1.2.0"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -124,7 +124,7 @@ rolldown_workspace = { version = "0.1.0", path = "./crates/rolldown_workspace" }
 
 anyhow = "1.0.86"
 append-only-vec = { version = "0.1.5" }
-arcstr = "1.2.0"
+arcstr = { version = "1.2.0", features = [] }
```
hyf0: Does this do the trick? Should we default-features = false?

hyf0 (approved):

Cargo.toml
Boshen: Forgot it's a default feature.

Issue #4508, opened 2025-05-12
[Bug]: Fails to load wasm fallback on WebContainer


sapphi-red:
### Reproduction link or steps

1. Open https://stackblitz.com/edit/github-cbzzyjkp?file=package.json
2. `pnpm i` and `pnpm build` runs automatically
3. `[rolldown] Downloading @rolldown/binding-wasm32-wasi@1.0.0-beta.8-commit.05b3e10 on WebContainer...` is output
4. `Error: Failed to load native binding` error happens

### What is expected?

`Error: Failed to load native binding` error to not happen

### What is actually happening?

The following error happens:

```
Error: Failed to load native binding
    at src/binding.js (file:///home/projects/github-cbzzyjkp/node_modules/.pnpm/rolldown@1.0.0-beta.8-commit.05b3e10_@oxc-project+runtime@0.69.0/node_modules/rolldown/dist/shared/parse-ast-index-C3bLZ2kg.mjs:353:36)
    at Module.eval (file:///home/projects/github-cbzzyjkp/node_modules/.pnpm/rolldown@1.0.0-beta.8-commit.05b3e10_@oxc-project+runtime@0.69.0/node_modules/rolldown/dist/shared/chunk--iN_1bjD.mjs:32:49)
    at src/builtin-plugin/utils.ts (file:///home/projects/github-cbzzyjkp/node_modules/.pnpm/rolldown@1.0.0-beta.8-commit.05b3e10_@oxc-project+runtime@0.69.0/node_modules/rolldown/dist/shared/src-C2ayPem0.mjs:90:78)
    at eval (file:///home/projects/github-cbzzyjkp/node_modules/.pnpm/rolldown@1.0.0-beta.8-commit.05b3e10_@oxc-project+runtime@0.69.0/node_modules/rolldown/dist/shared/chunk--iN_1bjD.mjs:29:55)
    at src/builtin-plugin/constructors.ts (file:///home/projects/github-cbzzyjkp/node_modules/.pnpm/rolldown@1.0.0-beta.8-commit.05b3e10_@oxc-project+runtime@0.69.0/node_modules/rolldown/dist/shared/src-C2ayPem0.mjs:155:2)
    at eval (file:///home/projects/github-cbzzyjkp/node_modules/.pnpm/rolldown@1.0.0-beta.8-commit.05b3e10_@oxc-project+runtime@0.69.0/node_modules/rolldown/dist/shared/chunk--iN_1bjD.mjs:29:55)
[...]

PR #4253, opened 2025-04-22
refactor(debug): use `serde_json` to emit strcutural logs

hyf0:
### Description

- Using `serde_json` gives more flexibility to manapulate data. It's very useful to inject some global data to logs, while `Valueable` is immutable.
- `Valueable` is not a active-maintained crate. https://github.com/tokio-rs/valuable/issues

PR #4510, opened 2025-05-12
chore(rust): remove unused `tracing_unstable` cfg

sapphi-red:
### Description

`RUSTFLAGS` would override the rustflags specified in `.cargo/config.toml`, so I guess this should not be set.
https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags

https://github.com/rolldown/rolldown/blob/513a17ae165b4ff12a3bebf169f159b3355651e9/.cargo/config.toml#L13-L57

refs https://github.com/rolldown/rolldown/issues/4508
This is the only place I found the difference between `@rolldown/browser` and `@rolldown/binding-wasm32-wasi`.

---

`tracing_unstable` is for using `valuable` crate, which is removed at https://github.com/rolldown/rolldown/pull/4253.


sapphi-red:
I just noticed that `https://pkg.pr.new/rolldown@513a17a` (latest pkg.pr.new build for the main branch) works fine, so I cannot confirm by https://github.com/rolldown/rolldown/pull/4510#issuecomment-2871136122 whether this PR fixes #4508 or not...

sapphi-red:
I'll run the canary one.

sapphi-red:
Ah, no, I cannot run it because it's on my fork...

hyf0 (approved):

PR #4520, opened 2025-05-13
feat: make RolldownBuild.watchFiles to async

underfin:
### Description

Making it to sync api maybe cause dead lock if the user call `write` and `watchFiles` at the same time, so here using async for it.


hyf0 (commented):
Is rollup async for this method?

hyf0 (approved):

PR #4530, opened 2025-05-13
feat: better error message for `manualChunks`

sapphi-red:
### Description

This PR makes the following error message to be output if `ouput.manualChunks` was specified.

> manualChunks is not supported. Please use advancedChunks instead

### DIFF
--- a/packages/rolldown/src/utils/validator.ts
+++ b/packages/rolldown/src/utils/validator.ts
@@ -865,5 +865,8 @@ export function getOutputCliKeys(): string[] {
 }
 
 export function getJsonSchema(): ObjectSchema {
-  return toJsonSchema(CliOptionsSchema) as ObjectSchema;
+  // errorMode: 'ignore' is set to ignore `never` schema
+  // there's no way to surpress the error one-by-one
+  // https://github.com/fabian-hiller/valibot/issues/1062
+  return toJsonSchema(CliOptionsSchema, { errorMode: 'ignore' }) as ObjectSchema;
 }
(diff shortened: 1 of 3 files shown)


packages/rolldown/src/utils/validator.ts
```diff
@@ -862,5 +865,10 @@ export function getOutputCliKeys(): string[] {
 }
 
 export function getJsonSchema(): ObjectSchema {
-  return toJsonSchema(CliOptionsSchema) as ObjectSchema;
+  return toJsonSchema(CliOptionsSchema, {
+    // errorMode: 'ignore' is set to ignore `never` schema
+    // there's no way to suppress the error one-by-one
+    // https://github.com/fabian-hiller/valibot/issues/1062
+    errorMode: 'ignore',
+  }) as ObjectSchema;
```
sapphi-red, line 873: I think we need a unit test to ensure that `manualChunks` is the one thing ignored here. But I didn't find a place to write that.

packages/rolldown/src/utils/validator.ts
```diff
@@ -862,5 +865,10 @@ export function getOutputCliKeys(): string[] {
 }
 
 export function getJsonSchema(): ObjectSchema {
-  return toJsonSchema(CliOptionsSchema) as ObjectSchema;
+  return toJsonSchema(CliOptionsSchema, {
```
hyf0, line 868: We're still using json-schema? @shulaoda 

I thought valibot should cover all our cases.

hyf0 (approved):
LGTM @shulaoda cc

packages/rolldown/src/utils/validator.ts
shulaoda: Some of the logic in CLI depends on it.

shulaoda: I will try to see if it can be removed now.

shulaoda (approved):

PR #4535, opened 2025-05-13
feat(pluginutils): `Query` filter impl

IWANABETHATGUY:

### DIFF
--- a/packages/pluginutils/src/composable-filters.test.ts
+++ b/packages/pluginutils/src/composable-filters.test.ts
@@ -1,0 +1,142 @@
+// cSpell:ignore fooa, afoo
+import { describe, expect, test } from 'vitest';
+import {
+  and,
+  exclude,
+  include,
+  interpreter,
+  or,
+  query,
+  QueryFilterObject,
+  queryObjectToFilterExpr,
+} from './composable-filters';
+
+function queryFilter(
+  id: string,
+  queryFilterObject: QueryFilterObject,
+): boolean {
+  let topLevelFilterExpression = include(
+    queryObjectToFilterExpr(queryFilterObject),
+  );
+  return interpreter([topLevelFilterExpression], undefined, id, undefined);
+}
+describe('queryFilter', () => {
+  test('boolean', () => {
+    expect(
+      queryFilter('/foo/bar?a=1111&b=2222', {
+        a: true,
+        b: true,
+      }),
+    ).toBe(true);
+
+    expect(
+      queryFilter('/foo/bar?a=1111&b=2222', {
+        a: true,
+        b: false,
+      }),
+    ).toBe(false);
+
+    expect(
+      queryFilter('/foo/bar?a=1111&b=2222', {
+        a: true,
+      }),
+    ).toBe(true);
+
+    expect(
+      queryFilter('/foo/bar?a=1111&b=2222', {
+        bar: false,
+      }),
+    ).toBe(true);
+  });
+
+  test('string', () => {
+    expect(
+      queryFilter('/foo/bar?a=1111&b=2222', {
+        a: '1111',
+      }),
+    ).toBe(true);
+
+    expect(
+      queryFilter('/foo/bar?a=1111&b=bar', {
+        a: '1111',
+        b: 'bar',
+      }),
+    ).toBe(true);
+
+    expect(
+      queryFilter('/foo/bar?a=1111&b=bar', {
+        d: '1111',
+      }),
+    ).toBe(false);
+  });
+
+  test('regex', () => {
+    expect(
+      queryFilter('/foo/bar?a=1111&b=2222', {
+        a: /[\d]+/,
+      }),
+    ).toBe(true);
+
+    expect(
+      queryFilter('/foo/bar?a=1111&b=bar', {
+        b: /bar/,
+      }),
+    ).toBe(true);
+
+    expect(
+      queryFilter('/foo/bar?a=1111&b=bar', {
+        d: /1111/,
+      }),
+    ).toBe(false);
+  });
+
+  test('custom', () => {
+    // https://github.com/sveltejs/vite-plugin-svelte/blob/3589433cd19464c484f560516d41e670e5d40710/packages/vite-plugin-svelte/src/utils/id.js#L35-L40
+    let filterExpr = or(
+      query('url', true),
+      and(
+        query('svelte', false),
+        or(
+          query('raw', true),
+          query('direct', true),
+        ),
+      ),
+    );
+    // include `url`, should return `false`
+    expect(interpreter(
+      [exclude(
+        filterExpr,
+      )],
+      undefined,
+      '/foo/bar?url=1',
+      undefined,
+    )).toBe(false);
+    // don't have `svelte` and has `raw`, should return `false`
+    expect(interpreter(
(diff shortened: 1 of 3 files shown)


IWANABETHATGUY:
* **#4537** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4537?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4536** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4536?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4535** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4535?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4535?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

hyf0 (dismissed):

IWANABETHATGUY:
Current impl can't represent this pattern, https://github.com/sveltejs/vite-plugin-svelte/blob/main/packages/vite-plugin-svelte/src/utils/id.js#L35-L40, I will rewrite it.

packages/pluginutils/src/composable-filters.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    expect(
+      queryFilter('/foo/bar?a=1111&b=bar', {
+        d: /1111/,
+      }),
+    ).toBe(false);
+  });
+
+  test('custom', () => {
+    // https://github.com/sveltejs/vite-plugin-svelte/blob/3589433cd19464c484f560516d41e670e5d40710/packages/vite-plugin-svelte/src/utils/id.js#L35-L40
+    let filterExpr = or(
```
IWANABETHATGUY, line 95: Now, the new `query` FilterExpr could express the same semantics as JavaScript.

packages/pluginutils/src/composable-filters.ts
```diff
@@ -148,6 +160,26 @@ export function exclude(expr: FilterExpression): Exclude {
...
+ * ```
+ * @param queryFilterObject The query filter object needs to be matched.
+ * @returns a `And` FilterExpression
+ *
+ * There are three kinds of conditions are supported:
+ * 1. `boolean`: if the value is `true`, the key must exist and be truthy. if the value is `false`, the key must not exist or be falsy.
+ * 2. `string`: the key must exist and be equal to the value.
+ * 3. `RegExp`: the key must exist and match the value.
+ */
+export function queryObjectToFilterExpr(queryFilter: QueryFilterObject): And {
```
IWANABETHATGUY, line 176: This helper function could be used for some simply scenario like https://github.com/vitejs/vite-plugin-vue/blob/main/packages/plugin-vue/src/index.ts#L362

hyf0 (approved):

IWANABETHATGUY:
cc @sapphi-red, I will merge it first, but feel free to comment if you find anything that is not proper, I will fix it in the following pr.

hyf0 (approved):

packages/pluginutils/src/composable-filters.ts
```diff
@@ -148,6 +160,26 @@ export function exclude(expr: FilterExpression): Exclude {
...
+ * ```js
+ *   and(query(k1, v1), query(k2, v2))
+ * ```
+ * @param queryFilterObject The query filter object needs to be matched.
+ * @returns a `And` FilterExpression
+ *
+ * There are three kinds of conditions are supported:
+ * 1. `boolean`: if the value is `true`, the key must exist and be truthy. if the value is `false`, the key must not exist or be falsy.
+ * 2. `string`: the key must exist and be equal to the value.
+ * 3. `RegExp`: the key must exist and match the value.
```
sapphi-red, line 174: I think this explanation should be put on `query` function.

packages/pluginutils/src/composable-filters.ts
sapphi-red, line 176: Maybe name it `queries` to align with other functions? I think this function can be seen similar to other functions, rather than a separate category.

packages/pluginutils/src/composable-filters.ts
```diff
@@ -231,8 +273,29 @@ export function exprInterpreter(
         ? code.includes(expr.pattern)
         : expr.pattern.test(code);
     }
+    case 'query': {
+      if (id === undefined) {
+        throw new Error('`id` is required for `Query` expression');
+      }
+      if (!ctx.urlSearchParamsCache) {
+        let [_, queryString = ''] = id.split('?', 2);
```
sapphi-red, line 281: This should be `const id = s.indexOf("?"); const queryString = pos >= 0 ? s.slice(pos) : ''`.
The second parameter of `split` works differently from rust in JS. For example, `"foo?bar?baz".split("?", 2)` returns `['foo', 'bar']` and not `['foo', 'bar?baz']`.

(I think the implementation in vite-plugins-svelte is slightly wrong because `new URL("https://example.com?foo?bar").search` returns `?foo?bar`.)

packages/pluginutils/src/composable-filters.test.ts
```diff
@@ -0,0 +1,142 @@
...
+        b: /bar/,
+      }),
+    ).toBe(true);
+
+    expect(
+      queryFilter('/foo/bar?a=1111&b=bar', {
+        d: /1111/,
+      }),
+    ).toBe(false);
+  });
```
sapphi-red, line 91: It would be nice to test `/foo/bar?a` and `/foo/bar?a=`. `query('bar', true)` should probably match these.
Maybe `query('bar', null)` should exist that only matches these cases?

packages/pluginutils/src/composable-filters.ts
```diff
@@ -140,6 +148,10 @@ export function code(pattern: StringOrRegExp): Code {
   return new Code(pattern);
 }
 
+export function query(key: string, pattern: StringOrRegExp | boolean): Query {
+  return new Query(key, pattern);
```
sapphi-red, line 152: Maybe this should be part of Id? (this only matches the id right?) In the future, we might add a way to match `importer` and probably good to take that into consideration.

packages/pluginutils/src/composable-filters.ts
```diff
@@ -231,8 +273,29 @@ export function exprInterpreter(
...
+        }
+      } else if (typeof expr.pattern === 'string') {
+        return urlParams.get(expr.key) === expr.pattern;
+      } else {
+        return expr.pattern.test(urlParams.get(expr.key) ?? '');
+      }
+    }
     default: {
-      throw new Error(`Expression kind ${expr.kind} is not expected.`);
+      throw new Error(`Expression kind ${expr} is not expected.`);
```
sapphi-red, line 298: I think graphite is correct here. I think it should be ``throw new Error(`Expression kind ${(expr as any).kind} is not expected.`);`` or ``throw new Error(`Expression ${JSON(expr)} is not expected.`);``.

IWANABETHATGUY: ts will complain: 
image

IWANABETHATGUY: I use option two

PR #4547, opened 2025-05-14
perf: reduce wasm release binary size

sapphi-red:
### Description

This PR reduces the wasm binary size by tweaking some build options. I guess these were not set to iterate the initial build setup, but we can set it now.

- Original: 9.7MB
- + lto='thin' => lto=true: 8.3MB
- + codegen-units=16 => codegen-units=1: 7.9MB
- + debug='full' => debug=false: 7.9MB
- + strip='none' => strip='symbols': 7.9MB

The `debug` and `strip` did not change the binary size. The output binary did not include debug information, so maybe it is stripped by napi-rs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,12 +13,8 @@ debug = true
 inherits = "release"
 
 [profile.release-wasi]
-codegen-units = 16
-debug = 'full'
 inherits = "release"
-lto = "thin"
 opt-level = "s"
-strip = "none"
 
 [workspace.lints.clippy]
 # Guidelines


IWANABETHATGUY (approved):

Brooooooklyn:
> The debug and strip did not change the binary size. The output binary did not include debug information, so maybe it is stripped by napi-rs.

NAPI-RS will remove debug information and symbols, but before doing so, it will first generate a rolldown-debug.wasm32-wasi.wasm file. This is for future debugging purposes. For example, if a crash occurs on a user's machine, the correct error stack can be viewed simply by downloading and replacing the original wasm file with this one. We'll optimize this part of the process when we have more time in the future.

sapphi-red:
@Brooooooklyn Ah, I see. I'll put back them with a comment then 👍

PR #4566, opened 2025-05-16
refactor(utils): optimize the logic for calculating parallelism and enhance compatibility processing

wChenonly:
Used on nodes below availableParallelism, report an error

image


IWANABETHATGUY (approved):

IWANABETHATGUY:
https://nodejs.org/en/blog/release/v18.20.8, BTW  node18 has reached EOF we don't guarantee to support it.

IWANABETHATGUY (approved):

PR #4638, opened 2025-05-22
feat: enable loose mode validate option and give warning

### LINKED ISSUE #4606: [Feature Request]: Validate options in `loose` mode when using rolldown from API or config file

There're a lot of options rollup supports, but rolldown don't support them. Some of these options don't really affect the bundle process, so there's no need to do a hard validation that blocks the build process.

.

### LINKED ISSUE #4608: [Feature Request]: `loose` options validation should emit warnings instead of behaving like `silent`

.

.

underfin:
### Description
close https://github.com/rolldown/rolldown/issues/4606
close https://github.com/rolldown/rolldown/issues/4608

### DIFF
--- a/packages/rolldown/src/utils/validator.ts
+++ b/packages/rolldown/src/utils/validator.ts
@@ -805,8 +805,6 @@ const inputHelperMsgRecord: HelperMsgRecord = {
 const outputHelperMsgRecord: HelperMsgRecord = {};
 
 export function validateOption<T>(key: 'input' | 'output', options: T): void {
-  if (globalThis.process?.env?.ROLLDOWN_OPTIONS_VALIDATION === 'loose') return;
-
   let parsed = v.safeParse(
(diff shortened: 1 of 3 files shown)


hyf0 (approved):

packages/rolldown/src/utils/validator.ts
```diff
@@ -844,7 +843,7 @@ export function validateOption<T>(key: 'input' | 'output', options: T): void {
       })
       .filter(Boolean);
     if (errors.length) {
-      throw new Error(`Failed validate ${key} options.\n` + errors.join('\n'));
+      console.warn(`Warning validate ${key} options.\n` + errors.join('\n'));
```
hyf0, line 846: If `ROLLDOWN_OPTIONS_VALIDATION` is `strict`, should we throw errors?

underfin: The strict  mode is droped, it is unnecessary after loose mode + warning

PR #4664, opened 2025-05-24
perf(rolldown): prevent duplicate text embedding with `include_str!`

### LINKED ISSUE #4354: `include_str` includes multiple copies of the same text
A quick search shows that we are using `include_str` to copy texts, but sometimes the same file is included multiples.

https://github.com/search?q=repo%3Arolldown%2Frolldown%20include_str&type=code

e.g https://github.com/rolldown/rolldown/blob/2d88918fb68896b68d7e48d2f02b5624ee2c1ec9/crates/rolldown/src/module_loader/runtime_module_task.rs#L61-L83

adds 3 copies of `../runtime/runtime-base.js` into the final binary.

re-taro:
### Description

Consolidates multiple `include_str!` calls into one static reference,
eliminating redundant string data and trimming the binary size.

close #4354

### DIFF
--- a/crates/rolldown/src/module_loader/runtime_module_task.rs
+++ b/crates/rolldown/src/module_loader/runtime_module_task.rs
@@ -1,4 +1,5 @@
 use arcstr::ArcStr;
+use constcat::concat;
 use oxc::ast_visit::VisitMut;
 use oxc::span::SourceType;
 use oxc_index::IndexVec;
@@ -19,6 +20,9 @@ use crate::{
   ast_scanner::{AstScanner, ScanResult},
   utils::tweak_ast_for_scanning::PreProcessor,
 };
+
+const RUNTIME_BASE_JS: &str = include_str!("../runtime/runtime-base.js");
+const RUNTIME_TAIL_JS: &str = include_str!("../runtime/runtime-tail.js");
 
 pub struct RuntimeModuleTask {
   tx: tokio::sync::mpsc::Sender<ModuleLoaderMsg>,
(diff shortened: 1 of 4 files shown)


IWANABETHATGUY:
Can you show the binary size diff between the current branch and main branch?

Boshen:
`arcstr::literal` is expensive, but it is `const` so we can create it in `const` context.

IWANABETHATGUY:
target triple: **x86_64-linux-gnu**

### Main 
16,510,264 bytes

### This branch
16,510,264 bytes

There is no difference

IWANABETHATGUY:
concat at runtime

16,508,344bytes

crates/rolldown/src/module_loader/runtime_module_task.rs
```diff
@@ -20,6 +20,28 @@ use crate::{
...
+const RUNTIME_TAIL_NODE_JS: &str = include_str!("../runtime/runtime-tail-node.js");
+
+fn get_runtime_js() -> String {
+  let mut ret = String::with_capacity(RUNTIME_BASE_JS.len() + RUNTIME_TAIL_JS.len());
+  ret.push_str(RUNTIME_BASE_JS);
+  ret.push_str(RUNTIME_TAIL_JS);
+  ret
+}
+
+fn get_runtime_js_with_node_platform() -> String {
```
IWANABETHATGUY, line 35: It's fine to concatenate them at runtime, because the runtime module only generated once.

IWANABETHATGUY (approved):
made some little tweak, thanks!

Boshen:
> Can you show the binary size diff between the current branch and main branch?

hmmm ... `constconcat` seems to be copying content into an const array, https://github.com/rossmacarthur/constcat/blob/e3c05e6965eea8cc2b0315e53ed715488e435d78/src/lib.rs#L254-L286

hence no reduction in size.

Boshen (approved):

PR #4681, opened 2025-05-25
refactor(rust): set default es target to the minimum supported version ES6

shulaoda:
### Description

Related to https://github.com/oxc-project/oxc/issues/11247
https://github.com/rolldown/rolldown/blob/3b287679560db12240550fa7cc3bb586f54e4945/crates/rolldown/src/stages/link_stage/generate_lazy_export.rs#L168-L180

Since the current oxc transform only supports a minimum of `ES6`, we should default to the more compatible `ES6` when the user does not explicitly specify an ES target, such as `target: 'chrome52'`.

### DIFF
--- a/crates/rolldown_binding/src/utils/normalize_binding_options.rs
+++ b/crates/rolldown_binding/src/utils/normalize_binding_options.rs
@@ -117,45 +117,43 @@ fn normalize_globals_option(
 }
 
 fn normalize_es_target(target: Option<&Either<String, Vec<String>>>) -> ESTarget {
-  target
-    .map(|target| {
-      let targets = match target {
-        Either::A(target) => {
-          if target.contains(',') {
-            target.split(',').collect::<Vec<&str>>()
-          } else {
-            vec![target.as_str()]
-          }
-        }
-        Either::B(target) => target.iter().map(std::string::String::as_str).collect::<Vec<&str>>(),
-      };
-      for target in targets {
-        if target.len() <= 2 || !target[..2].eq_ignore_ascii_case("es") {
-          continue;
-        }
-        if target[2..].eq_ignore_ascii_case("next") {
-          return ESTarget::default();
-        }
-        if let Ok(n) = target[2..].parse::<usize>() {
-          return match n {
-            5 => ESTarget::ES5,
-            6 | 2015 => ESTarget::ES2015,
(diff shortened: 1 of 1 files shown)


shulaoda:
* **#4681** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4681?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4681?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Dunqing:
I just found out the `ESTarget` can be set to `ESTarget::ES5`, which should be removed or report a warning, I think. Tracking in https://github.com/oxc-project/oxc/issues/11296

crates/rolldown_binding/src/utils/normalize_binding_options.rs
```diff
@@ -117,45 +117,43 @@ fn normalize_globals_option(
...
-            2018 => ESTarget::ES2018,
-            2019 => ESTarget::ES2019,
-            2020 => ESTarget::ES2020,
-            2021 => ESTarget::ES2021,
-            2022 => ESTarget::ES2022,
-            2023 => ESTarget::ES2023,
-            2024 => ESTarget::ES2024,
-            _ => continue,
-          };
+  target.map_or(ESTarget::ES2015, |target| {
```
hyf0: I was first ok with it and then thought that will it be more better to use the latest target by default?

The default behavior shouldn't downgrade code as possible. It could only be done when it's configured by user manually.

shulaoda: This behavior accounts for scenarios where the user specifies only runtime targets (e.g., `target: ["chrome52"]`) but does not explicitly set an es target. If no `target` is provided at all, we default to `ESTarget::ESNext`. However, if the user specifies runtime targets but omits an es target, we fall back to `ESTarget::ES6`. What do you think of this?

hyf0: > If no target is provided at all, we default to ESTarget::ESNext.

This makes sense.

hyf0 (approved):

Issue #4575, opened 2025-05-16
Rewrites `require('external')` call to static `import`


pi0:
### Reproduction link or steps

rolldown: (issue)

https://stackblitz.com/edit/github-h5ibjszy-43e3vno5?file=src%2Findex.mjs,rolldown.config.mjs,package.json

rollup + commonjs plugin (working)

https://stackblitz.com/edit/github-h5ibjszy?file=rollup.config.mjs,package.json,pnpm-lock.yaml

----

lib.cjs

```js
module.exports = function test() {
  const { join } = require('node:path');
  return join('foo', 'bar');
};
```

index.mjs

```js
import lib from './lib.cjs';

console.log(lib());
```

### What is expected?

`node:path` external should be hoisted as an ESM import in output. This is rollup/cjs plugin behavior:

<details>

```js
import require$$0 from 'node:path';

function getDefaultExportFromCjs (x) {
	return x && x.__esModule && Object.prototype.hasOwnProperty.call(x, 'default') ? x['default'] : x;
}

var lib$1;
var hasRequiredLib;

function requireLib () {
	if (hasRequiredLib) return lib$1;
	hasRequiredLib = 1;
	lib$1 = function test() {
	  const { join } = require$$0;
	  return join('foo', 'bar');
	};
	return lib$1;
}

var libExports = requireLib();
var lib = /*@__PURE__*/getDefaultExportFromCjs(libExports);

// Error: Calling `require` for "node:path" in an environment that doesn't expose the `require` function.
console.log(lib());

```

</details>

### What is actually happening?

externals go through commonjs shim that won't work:

```
Error: Calling `require` for "node:path" in an environment that doesn't expose the `require` function.
```

<details>

```js
//#region rolldown:runtime
var __create = Object.create;
var __defProp = Object.defineProperty;
var __getOwnPropDesc = Object.getOwnPropertyDescriptor;
var __getOwnPropNames = Object.getOwnPropertyNames;
var __getProtoOf = Object.getPrototypeOf;
var __hasOwnProp = Object.prototype.hasOwnProperty;
var __commonJS = (cb, mod) => function() {
	return mod || (0, cb[__getOwnPropNames(cb)[0]])((mod = { exports: {} }).exports, mod), mod.exports;
};
var __copyProps = (to, from, except, desc) => {
	if (from && typeof from === "object" || typeof from === "function") for (var keys = __getOwnPropNames(from), i = 0, n = keys.length, key; i < n; i++) {
[...]

Issue #2685, opened 2024-11-11
[Feature Request]: Make `require('external')` runnable in any execution context if possible


sapphi-red:
### What problem does this feature solve?

When a CJS file `require`s an external module commonly builtin modules like `node:path`, `Error: Dynamic require of "*" is not supported` is thrown for the output bundle.

#### Reproduction steps
1. Open https://stackblitz.com/edit/github-xv4yss-2sb9fe?file=rolldown.config.mjs,src%2Fentry.js,src%2Fa.cjs
2. `npm i` runs automatically
3. `npm run build` runs automatically
4. run `node dist/entry.js`. It errors with `Error: Dynamic require of "node:path" is not supported`.
5. run `npm run build:up` to use rollup
6. run `node dist/entry.js`. It runs successfully.

#### Workarounds

I'm setting
```js
banner: "import { createRequire as ___createRequire } from 'module'; const require = ___createRequire(import.meta.url);",
```
like when I encounter https://github.com/evanw/esbuild/issues/1921.

### What does the proposed API look like?

For `platform: 'node'`, it should be safe to use `import module from 'node:module'; const require = module.createRequire(import.meta.url)` for it.
For other `platform`s, I guess converting `require` into `import` is what users want in most cases. This may lead to crash in some cases due to incompatibility of `module.exports` / `export *`, so probably we need an option to control the interop code like [`requireReturnsDefault` option in @rollup/plugin-commonjs](https://github.com/rollup/plugins/blob/master/packages/commonjs/README.md#requirereturnsdefault).

~~For any `platform`s, I guess the node builtins can be treated as it's safe to convert `require` into `import` even if it's externalized.~~ probably this isn't safe due to hoisting?

sapphi-red:
This also happens when you try to bundle react for node.
https://stackblitz.com/edit/github-xv4yss-edjkwn?file=rolldown.config.mjs,src%2Fentry.js (same reproduction steps)

hyf0:
I just test it https://stackblitz.com/edit/github-xv4yss-2sb9fe?file=rolldown.config.mjs,src%2Fentry.js,src%2Fa.cjs. It runs expectedly.

I guess the problem is that when you run `dist/entry.js`, node consider `dist/entry.js` as esm due to the root `package.json#type: "module"`. So there's no builtin `require` in it and that's why a error is throwed.

If you rename the `entry.js` to `entry.cjs`, everything works fine.

hyf0:
```js
[...]

PR #4683, opened 2025-05-26
docs: explain how to inject `require`

sapphi-red:
### Description

Added an explanation how to inject `require` when the runtime does not provide `require` function and have to inject it by yourself.

refs #2685 #4575

### DIFF
--- a/docs/guide/in-depth/bundling-cjs.md
+++ b/docs/guide/in-depth/bundling-cjs.md
@@ -91,7 +91,35 @@ The `__toESM` helper ensures that CommonJS exports are properly converted to ES 
 
 ## `require` external modules
 
-Rolldown will leave it as it is. You need to ensure the running environment will provide a `require` function.
+When [`platform: 'node'`](../features.md#platform-presets) is set, Rolldown will generate a `require` function from [`module.createRequire`](https://nodejs.org/docs/latest/api/module.html#modulecreaterequirefilename).
+
+For other platforms, Rolldown will leave it as it is. You need to ensure the running environment to provide a `require` function or inject it by yourself.
+
+For example, you can inject the `require` function that returns the value obtained by `import` by using [`inject` feature](../features.md#inject).
+
+::: code-group
+
+```js [rolldown.config.js]
+import path from 'node:path'
+export default {
+  inject: {
+    require: path.resolve('./require.js'),
+  },
+};
+```
+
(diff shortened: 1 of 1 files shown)


docs/guide/in-depth/bundling-cjs.md
```diff
@@ -91,7 +91,35 @@ The `__toESM` helper ensures that CommonJS exports are properly converted to ES
 
 ## `require` external modules
 
-Rolldown will leave it as it is. You need to ensure the running environment will provide a `require` function.
+When [`platform: 'node'`](../features.md#platform-presets) is set, Rolldown will generate a `require` function from [`module.createRequire`](https://nodejs.org/docs/latest/api/module.html#modulecreaterequirefilename).
+
+For other platforms, Rolldown will leave it as it is, so ensure that the running environment provides a `require` function or inject one manually.
```
pi0, line 96: It might be not a practical advice in some cases.

Require polyfill has to be sync but ESM-native environments can only have it async unless there is static import or bundled (lazy loaded) modules that require polyfill returns. It is impractical as js only polyfill.

sapphi-red: Yeah, if you cannot import all the externalized modules statically, the example below won't work. In that case, a plugin is needed (at least for now). Probably that plugin should have [`requireReturnsDefault`](https://github.com/rollup/plugins/blob/master/packages/commonjs/README.md#requirereturnsdefault) option to control the behavior.

pi0: Yes exactly. But is rollup cjs plugin currently compatible with rolldown (IIRC it was not) or are you proposing to introduce new rolldown plugin or option?

sapphi-red: I'm proposing to introduce a new rolldown plugin. Extracting this part should work similar to `requireReturnsDefault=false`.
https://github.com/vitejs/rolldown-vite/blob/52af0fcefd10475e13c5c2536dfada3561c5542d/packages/vite/src/node/optimizer/rolldownDepPlugin.ts#L317-L367

pi0: Thnx for link! It would be also useful for Nitro rolldown support if extracted!

hyf0 (approved):
Thank you!

Issue #4289, opened 2025-04-24
Rolldown behaves differently from node for `import('cjs-module')`


sxzz:
### Reproduction link or steps

https://stackblitz.com/edit/github-tavgmypk?file=src%2Findex.js

- `pnpm run build && node dist/index.js`
- Got `{ parse: undefined }`

### What is expected?

Without bundle:
- `node src/index.js`
- Got `{ parse: [Function: parse$1] }`

### What is actually happening?

undefined parse

### System Info

```Shell
N/A
```

sapphi-red:
The named exports from CJS modules are detected by https://github.com/nodejs/cjs-module-lexer in Node.

related: https://github.com/oxc-project/oxc/issues/6381

hyf0:
I compared the result to esbuild. It seems rolldown resolves to the cjs version of `compiler-sfc` while esbuild resolves to the browser version.

hyf0:
With `format: 'cjs'`, rolldown will infer the platform as `node` instead of `browser`, because `platform: 'browser'` + `format: 'cjs'` seems absurd. However, it makes rolldown resolve to the cjs version of `compiler-sfc` instead of the `browser` version. 

Esbuild doesn't has this inferring behavior.

To solve your problem here, you could either change the platform to `browser` or customize the resolve behavior.

---

In summay, __this is not a bug__. But it does show some potential UX problems.

For nodejs, `import(..)` performs the esm resolution even in cjs environment, but, for rolldown, `platform: 'node'` will make the resolver for `import(..)` performs cjs-preferred resolution.

I'll bring this up in the meeting and have a discussion.

hyf0:
Short-term conclusions:

1. Improve document of `platform`
2. Remove the inferring behavior to align with esbuild

hyf0:
A simple test on how `import()` behave in `.cjs` environment. 


```
➜  node-play node -v           
v22.14.0
```

<img width="669" alt="Image" src="https://github.com/user-attachments/assets/bf9f4680-23ed-4f17-8dac-9292a82ba6db" />

---

<img width="262" alt="Image" src="https://github.com/user-attachments/assets/8ff2ad26-d74e-4e01-a83a-8f52b042cbb5" />

<img width="272" alt="Image" src="https://github.com/user-attachments/assets/4a32b647-da03-44f4-9075-5ebe6aad1684" />

---

[...]

PR #4693, opened 2025-05-27
fix: polyfill cjs on any platform

sxzz:
### Description

Always polyfill CJS, regardless of the platform, since we assume Node APIs (such as `__dirname`) are always available on CJS format.

### DIFF
--- a/crates/rolldown/src/module_finalizers/scope_hoisting/mod.rs
+++ b/crates/rolldown/src/module_finalizers/scope_hoisting/mod.rs
@@ -12,7 +12,7 @@ use oxc::{
 };
 use rolldown_common::{
   AstScopes, EcmaModuleAstUsage, ExportsKind, ImportRecordIdx, ImportRecordMeta, Module, ModuleIdx,
-  ModuleType, OutputFormat, Platform, SymbolRef, WrapKind,
+  ModuleType, OutputFormat, SymbolRef, WrapKind,
 };
 use rolldown_ecmascript_utils::{
   AstSnippet, BindingPatternExt, CallExpressionExt, ExpressionExt, StatementExt,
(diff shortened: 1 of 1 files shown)


hyf0:
Good moves! Here are some spontaneous thoughts:

- `Platform::Netural` intends to be used as a option to control if the bundler should have some platform specific behaviors. Doing polyfill on this platform seems conflict its' original intension.
- The combination of `platform: 'browser'` + `format: 'cjs'` itself is already contradictory. Any further action on this assumtion might introduces more undefined behaviors.
- Rolldown, as a bundler, needs to cover all possible cases, which makes it not suitable to make strong assumtions. If a controversial feature could be done in the user-land, I think it's more proper to do it in the user-land and wait for that it gets  tested solidly via time.

sxzz:
We can discuss this issue during the weekly meeting, but for now, merging this PR won't make things any worse.

Boshen:
We need some test additions / changes to catch regression if we ever going to change the behavior again in the future.

Boshen:
Also, do we have an option to opt out of this behaviour just in case?

yyx990803:
The problem here is that the polyfill results in `require('url')` calls - which will simply fail in non-Node platforms, so the usage itself is problematic, as it produces broken output.

This is the problem when you combine `format: 'cjs'` with a platform that is not `node`. This is also the reason why previously `format: 'cjs'` will automatically imply `platform: 'node'`. However, the inference behavior was reverted in #4611 in order to address #4289...

My thoughts:
- Defaulting to `platform: 'node'` when format is `cjs` is actually the desired behavior (since in practice, CJS can only run in Node or Node-compatible environments).
- Using a platform value that is not `node` with `format: 'cjs'` is non-sensical and should be discouraged.
- In `tsdown`, `cjs` output should use `platform: 'node'` even when the user sets a platform in config
- #4611 should be reverted
- Proper fix for #4289 should be following Node.js resolution when platform is node - i.e. `import(...)` should perform ESM-preferred resolution.
- This PR should be closed.

sxzz:
Then should we throw an error/warning if used neutral/browser platform + CJS?

PR #4715, opened 2025-05-28
chore: introduce knip

webpro:
Introduce Knip and remove unused files, exports and dependencies.

The process I followed (and can recommend):

1. Install Knip (e.g. `pnpm create @knip/config`)
2. Run `pnpm knip --files` and configure additional `entry` points (and narrow down `project` files) until no more unused files are found
3. Then the fun starts..
4. Run `knip` and maybe check a few samples of unused dependencies and/or exports if it feels about right
5. Run `knip --fix-type dependencies`
6. Run `pnpm i` (maybe also `git clean -fdx` first)
7. Run QA (tests)
8. Run `knip --fix-type exports`
9. Run `knip --fix-type types`
10. This one might be a bit controversial:

```
oxlint -A all -D 'no-unused-vars' -D '@typescript-eslint/no-unused-vars' -f json | remove-unused-vars
```

11. This is some "aggressive" code removal and pretty experimental. Run `knip --fix` and `oxlint` + [remove-unused-vars](https://github.com/webpro-nl/remove-unused-vars) again until no more unused variables are found.
12. Run QA, revert what went wrong
13. Hope for the best in CI ✌️


packages/browser/package.json
```diff
@@ -61,11 +61,5 @@
     "build-node": "BROWSER_PKG=1 pnpm run --filter rolldown build-node",
     "preinstall": "npx only-allow pnpm"
   },
-  "dependencies": {
-    "@napi-rs/wasm-runtime": "^0.2.8",
```
hyf0, line 65: Why remove this?

packages/bench/package.json
```diff
@@ -7,15 +7,11 @@
     "bench-ci": "oxnode ./benches/ci.js"
   },
   "dependencies": {
-    "react": "^19.0.0",
```
hyf0: I remember we have benchmarks that bundling react.

packages/browser/package.json
Boshen: It's removed if knip cannot detect its usage. This is a hard one actualy ... it's hidden really deep somewhere.

hyf0: Yeah. Here https://github.com/rolldown/rolldown/blob/9602e8ea86547993b7ddc817d70b871b54131078/packages/rolldown/src/wasi-worker-browser.mjs#L1

webpro: Knip checks dependencies per workspace. The one that we need is still here:

https://github.com/webpro/rolldown/blob/8328d20e46d6bf6e126eb411ae3f5ed486e77c46/packages/rolldown/package.json#L118

packages/bench/package.json
webpro: Yes, good catch! I wanted to see whether CI would fail. We'll definitely have to restore it. Knip doesn't understand this reference:

https://github.com/webpro/rolldown/blob/8328d20e46d6bf6e126eb411ae3f5ed486e77c46/packages/bench/src/suites/index.ts#L43

packages/browser/package.json
hyf0: `@rolldown/browser` is using the `package.json` of `packages/browser`. It requires to be in `dependecies`, when users install `@rolldown/browser`.

hyf0: Oh. Yeah. I forget to add this context. The tricky part here is that the dist code of `packages/browser` is generated from `packages/rolldown`.

webpro: Thanks Yunfei! Is it only `@napi-rs/wasm-runtime`? I have restored it now.

hyf0: I think so. We could release test version to see if everything works as expected.

webpro: OK sounds good.

Maybe we could consider doing something for rolldown packages like Knip does: publish with [pkg.pr.new](https://github.com/stackblitz-labs/pkg.pr.new) and then a real installation (just like from the npm registry) in a real repository. Example: https://github.com/webpro-nl/knip/blob/main/.github/workflows/integration.yml

hyf0: Yes. We had set up releasing process for pkg.pr.new. I'll take a look tomorrow, or boshen might do it. It's midnight in China. I need to go to bed.

Boshen:
Let's ship! Given we just did a release.

I assume our infra is broken if anything breaks - because its lacking test to use these unused code.

packages/browser/package.json
Boshen: Let's not block this PR on this issue.

Boshen (approved):

PR #4723, opened 2025-05-29
chore: add back bench-par

sapphi-red:
### Description

Adds back bench-par command in `packages/bench` that was removed in #4715.
It doesn't work anyway because parallel plugins are not working right now. But I guess it's fine to keep it. (or we can remove it completely if that's better)

### DIFF
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -19,11 +19,8 @@
     "packages/bench": {
       "entry": [
         "vue-entry.js",
-        "src/parallel-babel-plugin/impl.ts", // imported from src/parallel-babel-plugin/index.ts
       ],
-      "babel": {
-        "config": "src/parallel-babel-plugin/impl.ts",
-      },
+      "ignoreDependencies": ["@babel/@preset-.+"],
     },
     "packages/rolldown": {
       "entry": [
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -17,7 +17,13 @@
       ],
     },
     "packages/bench": {
-      "entry": ["vue-entry.js"],
+      "entry": [
+        "vue-entry.js",
+        "src/parallel-babel-plugin/impl.ts", // imported from src/parallel-babel-plugin/index.ts
+      ],
+      "babel": {
+        "config": "src/parallel-babel-plugin/impl.ts"
+      }
     },
     "packages/rolldown": {
       "entry": [
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -22,8 +22,8 @@
         "src/parallel-babel-plugin/impl.ts", // imported from src/parallel-babel-plugin/index.ts
       ],
       "babel": {
-        "config": "src/parallel-babel-plugin/impl.ts"
-      }
+        "config": "src/parallel-babel-plugin/impl.ts",
+      },
     },
     "packages/rolldown": {
       "entry": [
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -20,7 +20,7 @@
       "entry": [
         "vue-entry.js",
       ],
-      "ignoreDependencies": ["@babel/@preset-.+"],
+      "ignoreDependencies": ["@babel/preset-.+"],
     },
     "packages/rolldown": {
       "entry": [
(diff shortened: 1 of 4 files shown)


knip.jsonc
```diff
@@ -17,7 +17,13 @@
...
     },
     "packages/bench": {
-      "entry": ["vue-entry.js"],
+      "entry": [
+        "vue-entry.js",
+        "src/parallel-babel-plugin/impl.ts", // imported from src/parallel-babel-plugin/index.ts
+      ],
+      "babel": {
+        "config": "src/parallel-babel-plugin/impl.ts",
+      },
```
sapphi-red: This doesn't seems to be working. Is there a way to make knip detect these?
https://github.com/rolldown/rolldown/blob/1701888199dd59e5a219c0293c68dfc14cc66ea1/packages/bench/src/parallel-babel-plugin/impl.ts#L11-L14

packages/bench/src/parallel-babel-plugin/impl.ts
```diff
@@ -52,3 +56,11 @@ export const babelPlugin = (): Plugin => {
...
   };
 };
+
+const impl: ParallelPluginImplementation = defineParallelPluginImplementation(
+  (_options, _context) => {
+    return babelPlugin();
+  },
+);
+
+export default impl;
```
sapphi-red, line 67: This default export is referenced by 
https://github.com/rolldown/rolldown/blob/ec28db0817ce37b22538f4e606abe60debd38e9d/packages/bench/src/parallel-babel-plugin/index.ts#L5

I added to `knip.jsonc` as an entry, but it would be nice to write this in `src/parallel-babel-plugin/index.ts` directly. Is there a way to declare an entry there? I thinking of a comment like `// @knip-entry ./impl.js`.

knip.jsonc
```diff
@@ -17,7 +17,13 @@
       ],
     },
     "packages/bench": {
-      "entry": ["vue-entry.js"],
+      "entry": [
+        "vue-entry.js",
+        "src/parallel-babel-plugin/impl.ts", // imported from src/parallel-babel-plugin/index.ts
```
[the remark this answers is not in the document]
sapphi-red: intended

knip.jsonc
```diff
@@ -17,7 +17,13 @@
...
+      ],
+      "babel": {
+        "config": "src/parallel-babel-plugin/impl.ts"
```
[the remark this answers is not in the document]
sapphi-red: it's clear that it uses a babel config inside that file from the name

Boshen:
Probably need some help from @webpro-nl

knip.jsonc
webpro: The exported value of `impl.ts` is not a Babel configuration, so Knip's Babel plugin will not handle it properly.

For now, let's ignore them:

```
"ignoreDependencies": ["@babel/preset-env", "@babel/preset-typescript"]
```

or something like this (regex)

```
"ignoreDependencies": ["@babel/preset-.+"]
```

Looks like `impl.ts` is not an unused file, so no need to add it as an `entry` I think?

webpro:
> Probably need some help from @webpro-nl

Looks like I'm not getting notifications for this, please use @webpro

packages/bench/src/parallel-babel-plugin/impl.ts
webpro: We have at least three options:

- Add `impl.ts` as an `entry` file, since Knip by default does not report exports from entry files
- Add `@public` JSDoc tag:

```ts
/** @public */
export default impl;
```

- Add e.g. `@lintignore` JSDoc tag:

```ts
/** @lintignore */
export default impl;
```

But then we have to run like so: `knip --tags=-lintignore`

webpro: I should add support for this in Knip, a literal `import.meta.dirname` should be easy (but we will not have it yet for this pull request). Knip doesn't have support for comments such as you describe.

sapphi-red: Thanks! I've updated to use `@public` here.

`import.meta.dirname` support would be great! I think having a way to declare a reference on the importer side would also be nice. If it's declared on the importer side, it will be updated when changing the code. On the other hand, if it's declared on the importee side, I guess it'll be more easy to forget.

webpro: Agreed, that's a good idea. Then users can help Knip if the import is "too dynamic".

Boshen (approved):

PR #4735, opened 2025-05-29
refactor(bench): improve benchmark

Boshen:
* Removes remapping bench, seems like not a useful benchmark to be recorded on codspeed.
* Removes threejs10x, as codspeed measures CPU instructions, adding things that are the same doesn't help much.
* Removes benchmarking minification, oxc should take care of it.
* Simplifies the benchmark setup code.
* Running the benchmark is reduce from 17 mins to 3mins.


Dunqing:
I have a question: why do we have both `codspeed-benchmark` and `benchmark-rust`? They seem to do exactly the same thing.

Boshen:
> I have a question: why do we have both `codspeed-benchmark` and `benchmark-rust`? They seem to do exactly the same thing.

I didn't even notice they are the same thing ...

Dunqing:
> I didn't even notice they are the same thing ...

I can't imagine how much time it wastes.😅

PR #4779, opened 2025-06-02
feat: add `index_of_one` utils for `Bitset`

IWANABETHATGUY:
### Description
1. A utils used by `preserveEntrySignature`

PR #4800, opened 2025-06-03
feat: `preseveEntrySignature: allow-extension`

### LINKED ISSUE #4783: [Bug]: `should have export item index` panic at `crates/rolldown/src/ecmascript/format/esm.rs:108:12`

1. Open https://stackblitz.com/edit/github-8hpzqgef-dcrwq16f?file=rolldown.config.mjs
2. `pnpm i` runs automatically
3. `pnpm run build` runs automatically

### What is expected?

No panic happens

The following panic happens:

```
thread 'tokio-runtime-worker' panicked at crates/rolldown/src/ecmascript/format/esm.rs:108:12:
should have export item index
```
https://github.com/rolldown/rolldown/blob/73c4f639e9f8c547f5a63131337030d07f5fadaf/crates/rolldown/src/ecmascript/format/esm.rs#L108

### System Info

```Shell
StackBlitz

  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @rolldown/browser:  1.0.0-beta.10-commit.bf212da
```

### Any additional comments?

Originally reported at https://github.com/vitejs/rolldown-vite/issues/196

IWANABETHATGUY:
### Description
I use another pass to merge pending common modules into existing chunks when `allow-extension` is specified, which is non-intrusive for the existing split chunks logic, and makes the merging algorithm more extensible.
The basic idea is pretty simple: 
- Choose a pending common module(shared by multiple chunk)
- If it satisfies the following condition, the pending common module could be inserted into user-defined entry chunk. 
   - Only shared by one user-defined entry and one or more dynamic entries, but none common chunks.
   - All dynamic entry modules are descendants of the entry module
 

-  Also fixed https://github.com/rolldown/rolldown/issues/4783


IWANABETHATGUY:
* **#4800** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4800?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4800?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4779** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4779?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

crates/rolldown/src/module_finalizers/scope_hoisting/mod.rs
```diff
@@ -244,7 +244,7 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
             // instead of keeping `console.log(foo)` as we did in esm output. The reason here is we need to keep live binding in cjs output.
 
             let exported_name = &self.ctx.chunk_graph.chunk_table[chunk_idx_of_canonical_symbol]
-              .exports_to_other_chunks[&canonical_ref];
+              .exports_to_other_chunks[&canonical_ref][0];
```
IWANABETHATGUY: Use the first one is ok, since they both point to the same symbol.

crates/rolldown/src/utils/chunk/render_chunk_exports.rs
```diff
@@ -302,62 +307,29 @@ pub fn render_object_define_property(key: &str, value: &str) -> String {
...
-) -> Vec<(Rstr, SymbolRef)> {
-  let get_exports_items_from_common_chunk = |chunk: &Chunk| {
-    let mut tmp = chunk
-      .exports_to_other_chunks
-      .iter()
-      .map(|(export_ref, alias)| (alias.clone(), *export_ref))
-      .collect::<Vec<_>>();
-
-    tmp.sort_unstable_by(|a, b| a.0.as_str().cmp(b.0.as_str()));
+pub fn get_export_items(chunk: &Chunk) -> Vec<(Rstr, SymbolRef)> {
```
IWANABETHATGUY, line 308: consolidate normalize `export_items` into export_to_other_chunks`, since after we support `allow-extension`, even the entry chunk may export some symbols not belongs to the entry module.

IWANABETHATGUY: see https://github.com/rolldown/rolldown/pull/4800/files#diff-9aa477569700731f0970cded185634da7c763b00fed60a39e45921bfd8acd0beR302-R329

crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
```diff
@@ -389,20 +417,27 @@ impl GenerateStage<'_> {
...
+        .sorted_by_cached_key(|(symbol_ref, _predefined_names)| {
           // same deconflict order in deconflict_chunk_symbols.rs
           // https://github.com/rolldown/rolldown/blob/504ea76c00563eb7db7a49c2b6e04b2fbe61bdc1/crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs?plain=1#L86-L102
           Reverse::<u32>(self.link_output.module_table[symbol_ref.owner].exec_order())
         })
-        .copied()
       {
-        let original_name: rolldown_rstr::Rstr =
-          chunk_export.name(&self.link_output.symbol_db).to_rstr();
+        let original_name: rolldown_rstr::Rstr = match predefined_names.as_slice() {
```
IWANABETHATGUY, line 430: entry module symbols always take higher priority, so they don't need to be deconflicted, they are required to be as is (because as an entry chunk, the symbol name may be consumed by other applications.)

crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
```diff
@@ -20,7 +20,7 @@ use rustc_hash::{FxHashMap, FxHashSet};
 type IndexChunkDependedSymbols = IndexVec<ChunkIdx, FxIndexSet<SymbolRef>>;
 type IndexChunkImportsFromExternalModules =
   IndexVec<ChunkIdx, FxHashMap<ModuleIdx, Vec<NamedImport>>>;
-type IndexChunkExportedSymbols = IndexVec<ChunkIdx, FxHashSet<SymbolRef>>;
+type IndexChunkExportedSymbols = IndexVec<ChunkIdx, FxHashMap<SymbolRef, Vec<Rstr>>>;
```
IWANABETHATGUY, line 23: For a symbol exported by a entry module, the original name will be pushed into `vec`, for a none entry module exported symbol, keep the empty vec. With this trick, we could determine if the symbol came from entry or not.
Also we use `Vec<Rstr>` instead of `Option<Rstr` because, one `SymbolRef` could related to multiple exported name, e.g.
https://github.com/rolldown/rolldown/blob/3aa36870d2f97c9b4056195b67f1bd24caaa0095/crates/rolldown/tests/esbuild/default/export_forms_es6/entry.js?plain=1#L5-L7

the `Class` symbol corresponding to two name: `[Classs, C]`

hyf0 (approved):

hyf0:
With this, does it mean we use `advancedChunks` to move modules into entry chunk and generate a valid output?(currently advancedChunks doesn't support move modules into entry chunk.

IWANABETHATGUY:
> With this, does it mean we use `advancedChunks` to move modules into entry chunk and generate a valid output?(currently advancedChunks doesn't support move modules into entry chunk.

I think yes. But the condition maybe pretty strict, at least for `preserveEntrySignatures: allow-extension` the strategy is pretty consertative.

PR #4816, opened 2025-06-04
feat: interop `import('some-cjs-module.js')`

hyf0:
### Description

Fixes https://github.com/rolldown/rolldown/issues/4289. @sapphi-red

PR #4815, opened 2025-06-04
refactor(rust): extract `try_rewrite_import_expression` logic into a standalone function

hyf0:

PR #4819, opened 2025-06-04
feat: take node esm spec into account for dynamic imports in cjs modules that satisfy nodejs

hyf0:
### Description

Related to #4289.

@sapphi-red pointed out that dynamic imports in cjs is a thing that's quite specfic to nodejs. So to improve compatibilities of `import('some-cjs-module.js')`, now rolldown always considers node esm spec for dynamic imports in cjs modules.

### DIFF
--- a/crates/rolldown/src/module_finalizers/mod.rs
+++ b/crates/rolldown/src/module_finalizers/mod.rs
@@ -153,7 +153,7 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
               self.snippet.builder.vec(),
               false,
             ),
-            self.ctx.module.should_consider_node_esm_spec(),
+            self.ctx.module.should_consider_node_esm_spec_for_static_import(),
           ),
         );
         return false;
@@ -914,7 +914,7 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
                     self.snippet.builder.atom(to_esm_fn_name.as_str()),
                   ),
                   self.snippet.call_expr_expr(importee_wrapper_ref_name),
-                  self.ctx.module.should_consider_node_esm_spec(),
+                  self.ctx.module.should_consider_node_esm_spec_for_dynamic_import(),
                 ),
               ))
             }
@@ -961,7 +961,7 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
               self.finalized_expr_for_runtime_symbol("__toDynamicImportESM");
 
             let mut arguments = self.snippet.builder.vec();
-            if self.ctx.module.should_consider_node_esm_spec() {
+            if self.ctx.module.should_consider_node_esm_spec_for_dynamic_import() {
               arguments.push(ast::Argument::from(self.snippet.builder.expression_numeric_literal(
                 SPAN,
                 1.0,
@@ -1105,7 +1105,7 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
                                 false,
                               ),
                             ),
-                            self.ctx.module.should_consider_node_esm_spec(),
+                            self.ctx.module.should_consider_node_esm_spec_for_static_import(),
                           ),
                         ),
                       }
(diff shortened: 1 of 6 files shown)


hyf0:
* **#4821** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4821?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4819** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4819?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4819?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4817** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4817?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4816** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4816?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4815** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4815?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

crates/rolldown/tests/esbuild/default/top_level_await_allowed_import_without_splitting/artifacts.snap
```diff
@@ -32,8 +32,8 @@ var init_a = __esm({ async "a.js"() {
 var require_entry = __commonJS({ "entry.js"() {
 	init_a().then(() => a_exports);
 	init_b().then(() => b_exports);
-	Promise.resolve().then(() => __toESM(require_c()));
-	Promise.resolve().then(() => __toESM(require_entry()));
+	Promise.resolve().then(() => __toESM(require_c(), 1));
+	Promise.resolve().then(() => __toESM(require_entry(), 1));
```
sapphi-red: It is not directly related to this PR, but this test case has a broken output.

These have `await` inside a non-async function.
```js
var require_c = __commonJS({ "c.js"() {
	await 0;
} });
```
```js
var require_entry = __commonJS({ "entry.js"() {
	init_a().then(() => a_exports);
	init_b().then(() => b_exports);
	Promise.resolve().then(() => __toESM(require_c(), 1));
	Promise.resolve().then(() => __toESM(require_entry(), 1));
	await 0;
} });
```

Also I think if a module contains TLA, it should be treated as ESM (as TLAs are not allowed in CJS).

crates/rolldown/tests/rolldown/issues/4289/node-mode-by-implicit-cjs.js
```diff
@@ -0,0 +1,13 @@
+import nodeAssert from 'node:assert';
+
+async function main() {
+  const exports = await import('./lib.js');
```
sapphi-red: I think node mode should not be used for this file. Because of the static import in the first line, this file won't run directly on Node.
If the `import nodeAssert from 'node:assert';` is replaced with `const nodeAssert = require('node:assert')`, I would expect node mode to be used.

I think the condition has to be `!containsEsmSyntax && containsCjsSyntax`.

crates/rolldown/tests/esbuild/default/top_level_await_allowed_import_without_splitting/artifacts.snap
hyf0: I think the output is already broken.

sapphi-red: Yeah, this is a separate issue.

crates/rolldown/tests/rolldown/issues/4289/node-mode-by-implicit-cjs.js
hyf0: Currently we follwos esbuild's algorithm, Only `export` and `module/exports`  keyword decide if a module is a cjs or esm module.

IMO, I don't want the detection to be very subtle and intend to follow the esbuild's detection.

However, you do make a point that writing static imports in cjs modules is invalid in nodejs. So I guess the most safe way is still checking extension and package.json.

I'm gonna remove enabling node mode for inplicit cjs modules in this PR and ship safe detection first. Meanwhile, we could have more discussions later.

hyf0: Think again, I do feel using non-node mode would be more suitable. Modules writes static imports and use `module/exports` are definitely for bundling. We should perform `__esModule` checking instead of enabling `node mode`.

sapphi-red (approved):

PR #4851, opened 2025-06-06
Add WASI Preview 2 (wasip2) support

### LINKED ISSUE #898: Improve WASI support

Enrique726:
# Add WASI Preview 2 (wasip2) support

## Description

This pull request implements support for WASI Preview 2 in Rolldown, addressing issue #898. The implementation adds proper platform detection, configuration options, and package infrastructure for WASI Preview 2 environments.

## Changes

- Added `WasiP2` variant to the `Platform` enum in `platform.rs`
- Updated `TryFrom` implementation to support string conversion for "wasi", "wasip1", and "wasip2"
- Created `wasi_features.rs` with utility functions for WASI platform detection and configuration
- Created `wasi_config.rs` for WASI-specific configuration options
- Updated platform detection in `normalize_options.rs`
- Modified `BundlerBuilder` to apply WASI-specific configurations
- Created directory structure and files for the wasm32-wasip2 binding
- Implemented @rolldown/wasip2 package for clean API exposure
- Updated Cargo.toml with release-wasip2 profile
- Added tests for platform detection, string conversion, and configuration utilities

## Features

- Detection of WASI environments (both Preview 1 and Preview 2)
- WASI-specific configuration options based on version
- Dedicated package for WASI Preview 2 builds
- Fallback approach for testing without WebAssembly Component Model 
- Support for both WASI versions while maintaining backward compatibility

## Testing

All tests pass successfully, validating:
- Platform detection between WASI Preview 1 and 2
- String conversion for platform types
- WASI-specific configuration options
- Build system functionality

## Future Enhancements

- Deeper WebAssembly Component Model integration as it matures
- WASI Preview 2-specific optimizations
- Extended testing in various WASI runtimes
- Additional documentation and usage examples

Fixes #898


packages/rolldown/target-wasm/.rustc_info.json
Boshen, line 1: Need to `.gitignore` this directory?

Enrique726: Added the target-wasm/ directory to .gitignore as suggested. This will prevent the build artifacts from being committed to the repository. Thanks for the recommendation!

Brooooooklyn:
bugbot run

crates/rolldown/src/wasi_config.rs
```diff
@@ -0,0 +1,93 @@
...
+  if !is_wasi_platform(platform) {
+    return;
+  }
+
+  // Set appropriate target triple
+  if let Some(target) = get_wasi_target_triple(platform) {
+    environment.insert("RUSTUP_TOOLCHAIN".to_string(), "nightly".to_string());
+    environment.insert("TARGET".to_string(), target.to_string());
+    
+    // For Preview 2, we need component model
```
Brooooooklyn: Actually `nightly` and `component model` are not required, as component model has already been forcefully enabled in NAPI-RS, and it takes effect even in stable Rust. If component model is not enabled, `wasip1` cannot work properly either. https://github.com/napi-rs/napi-rs/blob/main/crates/build/src/wasi.rs#L24-L46

Cargo.toml
```diff
@@ -1,3 +1,5 @@
+cargo-features = ["profile-rustflags"]
```
Brooooooklyn: if add this, we can't use stable Rust anymore. And this is seems no need anymore, so let's remove this line

crates/rolldown/src/utils/normalize_options.rs
```diff
@@ -53,6 +53,21 @@ pub fn normalize_options(mut raw_options: crate::BundlerOptions) -> NormalizeOpt
...
   });
 
+  // Check if we're in a WebAssembly environment - if so, override platform to WASI
+  #[cfg(target_family = "wasm")]
+  let platform = if platform == Platform::Node || platform == Platform::Browser {
+    #[cfg(target_feature = "component-model")]
+    {
+      Platform::WasiP2
+    }
+    #[cfg(not(target_feature = "component-model"))]
```
Brooooooklyn: no need here, component-model is enforced

test-platform-strings.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * Test script to verify platform string conversion for WASI platforms
+ */
+import { tryFrom } from './crates/rolldown_common/src/inner_bundler_options/types/platform.js';
```
Brooooooklyn, line 4: These tests should be moved to `./crates/rolldown_common` and run using Vitest.

Brooooooklyn (changes requested):
@Enrique726 Hey, thanks for your contribution. I left some comments on your PR.
The following changes are needed:
1. nightly and component-model are not required, NAPI-RS has already handled this, we need to ensure that rolldown can be developed normally under stable Rust
2. The build scripts and code need to remove the checks for nightly and `component-model`, currently in the `wasi` environment we ensure that `component-model` is already enabled
3. Some scripts need to be rewritten in `TypeScript`, and testing should be unified using Vitest tests

PR #4858, opened 2025-06-07
ci(CHANGELOG.md): use git-cliff to generate changelogs

Boshen:
This PR removes the npm package `conventional-changelog` in favor of git cliff, which produces cleaner changelog, and is more powerful and easier to use.

changelog is also added to the release PR and github release.

CHANGELOG.md is regenerated.


hyf0:
Some notes:

-  commits include breaking changes need to be put in the head and featured.

PR #4901, opened 2025-06-10
Hide runtime code in hmr snapshot

hyfdev:


hyf0:
Done in https://github.com/rolldown/rolldown/pull/5160.

PR #4908, opened 2025-06-11
feat: make native json plugin callable

shulaoda:
> By the way, are there any native plugins that can be used in dev other than the resolver? If there aren't, I wonder if it's better to disable them in `resolveConfig`.

Currently, we only have three native plugins that are potentially used in the dev environment: `resolver`, `wasmFallback`, and maybe this `json` plugin. The rest aren't callable in dev because they either use the `transform_ast` hook, rely on internal context, or are only used in build environment.

_Originally posted by @shulaoda in https://github.com/vitejs/rolldown-vite/issues/246#issuecomment-2958188784_

### DIFF
--- a/crates/rolldown_binding/src/options/plugin/binding_callable_builtin_plugin.rs
+++ b/crates/rolldown_binding/src/options/plugin/binding_callable_builtin_plugin.rs
@@ -1,12 +1,21 @@
 use std::sync::Arc;
 
+use arcstr::ArcStr;
 use napi::Either;
 use napi_derive::napi;
+use rolldown::ModuleType;
 use rolldown_common::{WatcherChangeKind, side_effects};
 use rolldown_plugin::{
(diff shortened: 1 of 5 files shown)


shulaoda:
* **#4908** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4908?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4908?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

shulaoda:
This adaptation is currently only for the JSON plugin, and it might introduce some performance overhead in the dev environment. Is it really worth it — or should we just use the JS JSON plugin during dev instead? @sapphi-red cc

IWANABETHATGUY (dismissed):

sapphi-red (approved):
I tested this native json plugin on https://github.com/sapphi-red/performance-compare and the diff of the native plugin and js plugin was less than 2% (which is within the margin of error). So I think it's worth.

PR #4915, opened 2025-06-11
docs(pluginutils): add example to simple filter functions

sapphi-red:
Added examples to simple filter functions to show how it can be used.

### DIFF
--- a/packages/pluginutils/src/simple-filters.ts
+++ b/packages/pluginutils/src/simple-filters.ts
@@ -5,6 +5,18 @@
  *
  * @param str the string to match.
  * @param flags flags for the RegExp.
+ *
+ * @example
+ * ```ts
+ * import { exactRegex } from '@rolldown/pluginutils';
+ * const plugin = {
+ *   name: 'plugin',
+ *   resolveId: {
+ *     filter: { id: exactRegex('foo') },
+ *     handler(id) {} // will only be called for `foo`
+ *   }
+ * }
+ * ```
  */
 export function exactRegex(str: string, flags?: string): RegExp {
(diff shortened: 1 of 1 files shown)


hyf0 (approved):

packages/pluginutils/src/simple-filters.ts
```diff
@@ -33,6 +57,25 @@ type WidenString<T> = T extends string ? string : T;
...
+ * ```ts
+ * import { makeIdFiltersToMatchWithQuery } from '@rolldown/pluginutils';
+ * const plugin = {
+ *   name: 'plugin',
+ *   transform: {
+ *     filter: { id: makeIdFiltersToMatchWithQuery(['**' + '/*.js', /\.ts$/]) },
+ *     // The handler will be called for IDs like:
+ *     // - foo.js
+ *     // - foo.js?foo
+ *     // - foo.txt?foo.js
```
dominikg, line 71: is `.txt` missing from the example call? reading that code would have me assume that only js/ts extensions are allowed.

Also how does this break apart regexes and join them together again? eg, if a pattern already includes a ? then you'd have a hard time merging the two to match the start of query string.

dominikg: ahh, the regex is probably not ignoring extensions at the end if a ? is before them.

sapphi-red: If the input pattern includes `?`, it probably matches more than it should. But if that's important you can still filter them out by running the filter inside the handler. Maybe we can add a comment that explains about that a bit.

PR #4917, opened 2025-06-11
docs: add link to repl

sxzz:


docs/.vitepress/config.ts
```diff
@@ -95,6 +95,7 @@ export default defineConfig({
       { text: 'User Guide', link: '/guide/' },
       { text: 'Reference', link: '/reference/config-options.md' },
       { text: 'Contribute', link: '/contrib-guide/' },
+      { text: 'REPL', link: 'https://repl.rolldown.rs/' },
```
TheAlexLichter, line 98: Maybe "playground" instead? Wondering what users are most used to.

shulaoda: Let's align with rollup first.

shulaoda (approved):

PR #4918, opened 2025-06-11
feat: allow to configure`experimental#attachDebugInfo: 'none'` to disable generating `#region` comments

### LINKED ISSUE #4668: [Feature Request]: Make `#region` code folding configurable rather than enforced

Allowed the output to come without `#region` and `#endregion` code folding feature.

Based on https://github.com/rolldown/rolldown/commit/ace105cf69965baec8ac1626d5b79e42b66574e3, add options to enable and disable the feature rahter than enforced.

```ts

export default defineConfig({
    output: {
        regions: false,
    },
});
```

TheAlexLichter:
### Description

This PR changes the signature of `experimental.attachDebugData` and the behavior.

Instead of a boolean, it now accepts a string:

* `none` - Don't print any debug information (including region comments)
* `simple` (defaulting to if none given) - Print region comments. This was `false` before.
* `full` - Print region comments and info about chunking. This was `true` before.

Related PR: https://github.com/rolldown/rolldown/pull/4861

---

TODO:

* Some code might be unnecessarily complex - still new to Rust :D Happy for feedback!


hyf0 (approved):
Well done!

PR #4919, opened 2025-06-11
fix: give a warning when target is set to es5 in tsconfig

### LINKED ISSUE #3275: give a warning when target is set to `es5` in tsconfig.
> > Rolldown doesn't aware of tsconfig.json and we gonna keep this behavior
> 
> I'm using my a rollup plugin to convert to es5 that respects the tsconfig file, but the shims are es6. It seems a waste to transpile twice, but okay
> 
> Currently our upstream transpiler `oxc` don't have enough bandwidth to support lowering to `es5` and if you want to lower to `es5` you could use `swc` to lower it twice , maybe we should give a warning when target is set to `es5` in tsconfig. 

 _Originally posted by @IWANABETHATGUY in [#3243](https://github.com/rolldown/rolldown/issues/3243#issuecomment-2566088045)_

AliceLanniste:
close #3275

### DIFF
--- a/crates/rolldown/src/bundler_builder.rs
+++ b/crates/rolldown/src/bundler_builder.rs
@@ -1,5 +1,7 @@
 use std::sync::Arc;
 
+use oxc::transformer::ESTarget;
+
 use rolldown_common::{FileEmitter, NormalizedBundlerOptions};
 use rolldown_error::BuildDiagnostic;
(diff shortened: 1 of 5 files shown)


shulaoda:
Thank you! We have dropped support for ES5. If users enable it, an error will be thrown.

PR #4948, opened 2025-06-13
feat(rolldown): oxc v0.73.0

### LINKED ISSUE #2603: [Bug]: rolldown does not treat builtin object constructors like `new Map` as sideeffect free

1. Open https://stackblitz.com/edit/github-xv4yss-xwr4ef?file=src%2Fentry.js,package.json
2. `npm i` runs automatically
4. See `dist/entry.js` (not empty)
5. run `npm run build:up` to use rollup
6. See `dist/entry.js` (empty)

### What is expected?

Builtin object constructors like `new Map` are treated as sideeffect free like rollup does and the output gets empty.

The output is not empty.

### System Info

```Shell
Stackblitz

  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    rolldown: https://pkg.pr.new/rolldown@829d80a => 0.13.2
```

### Any additional comments?

Found while trying to bundle Vite with Rolldown.

The code where rollup defines the sideeffects for each builtin functions / objects are written in https://github.com/sapphi-red/rollup/blob/master/src/ast/nodes/shared/knownGlobals.ts

Boshen:
closes #2603


crates/rolldown/src/ast_scanner/side_effect_detector/mod.rs
```diff
@@ -432,7 +432,7 @@ impl<'a> SideEffectDetector<'a> {
 
       Expression::ArrayExpression(expr) => self.detect_side_effect_of_array_expr(expr),
       Expression::NewExpression(expr) => {
-        let is_pure = maybe_side_effect_free_global_constructor(self.scope, expr) || expr.pure;
+        let is_pure = expr.pure || maybe_side_effect_free_global_constructor(self.scope, expr);
```
Boshen, line 435: `maybe_side_effect_free_global_constructor(self.scope, expr);` is not removed because we need to change the test setup to apply pure.

crates/rolldown/tests/esbuild/dce/no_side_effects_comment_type_script_declare/artifacts.snap
```diff
@@ -1,14 +1,29 @@
...
 
-```js
-//#region entry.ts
-let ns;
-(function(_ns) {})(ns || (ns = {}));
+```text
+[PARSE_ERROR] Error: An implementation cannot be declared in ambient contexts.
+   ╭─[ entry.ts:2:51 ]
+   │
+ 2 │ /* @__NO_SIDE_EFFECTS__ */ declare function f1(y) { sideEffect(y) }
```
Boshen, line 12: This is a ts syntax error, where esbuild don't produce such errors.

hyf0 (approved):

crates/rolldown/tests/esbuild/dce/no_side_effects_comment_type_script_declare/artifacts.snap
hyf0: I think it's ok we raise errors if ts itself does it too.

PR #4955, opened 2025-06-14
feat: support symlink

### LINKED ISSUE #4249: [Feature Request]: Support symlinked paths for loading rolldown config

AliceLanniste:
close: #4249

I add tests for symlink,  I failed to create symlink  via execa ``.
windows:
image
unix:
image
I need help to reslove these


sapphi-red:
Closing as the issue seems to be already fixed.

PR #4966, opened 2025-06-15
docs: improve advanced chunks

hyf0:


hyf0:
* **#4966** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4966?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4966?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/guide/in-depth/advanced-chunks.md
```diff
@@ -1,6 +1,238 @@
 # Advanced Chunks
 
-Advanced chunks are a powerful feature that allows you manually control the chunking of your code. This is useful when you want to optimize the loading of your application by splitting it into smaller, more manageable pieces.
+`AdvancedChunks` is a powerful feature that allows you to do `manual chunking` against the `automatic chunking` done by [code splitting](./code-splitting.md). This is useful when you want to optimize the loading performance of your application by splitting it into smaller, more manageable pieces.
```
sapphi-red:
```suggestion
`AdvancedChunks` is a powerful feature that allows you to do `manual chunking` to complement the `automatic chunking` done by [code splitting](./code-splitting.md). This is useful when you want to optimize the loading performance of your application by splitting it into smaller, more manageable pieces.
```
Since using `manual chunking` does not disable `automatic chunking`, I think "against" is not a correct word here.

docs/guide/in-depth/advanced-chunks.md
```diff
@@ -1,6 +1,238 @@
...
+
+- We used 3 libraries: `react`, `react-dom`, and `ui-lib`.
+- `output-hash0.js` is the output file generated by Rolldown.
+- `hash0` is the hash of the output file, it changes if the content of the file changes.
+
+### Reduce cache invalidation
+
+Let's talk about cache invalidation first. Cache invalidation here means that when you deploy a new version of your application, the browser will need to download the new version of the file. If the file is large, this can lead to a poor user experience.
+
+For example, you change the `app.jsx` file
```
sapphi-red:
```suggestion
For example, if you change the `app.jsx` file:
```

docs/guide/in-depth/advanced-chunks.md
```diff
@@ -1,6 +1,238 @@
...
+
+```jsx [app.jsx]
+function App() {
+  return <Button onClick={() => alert('Button clicked!')} />; // [!code --]
+  return <Button onClick={() => alert('Button clicked!!!')} />; // [!code ++]
+}
+```
+
+and naturally, you get a `output-hash1.js` file with the same content as `output-hash0.js`, except the `App` function is changed.
```
sapphi-red:
```suggestion
then naturally, you get a `output-hash1.js` file with the same content as `output-hash0.js`, except for the change in the `App` function.
```

docs/guide/in-depth/advanced-chunks.md
```diff
@@ -1,6 +1,238 @@
...
+
+In this example,
+
+- We used 3 libraries: `react`, `react-dom`, and `ui-lib`.
+- `output-hash0.js` is the output file generated by Rolldown.
+- `hash0` is the hash of the output file, it changes if the content of the file changes.
+
+### Reduce cache invalidation
+
+Let's talk about cache invalidation first. Cache invalidation here means that when you deploy a new version of your application, the browser will need to download the new version of the file. If the file is large, this can lead to a poor user experience.
```
sapphi-red, line 68: I think we need to mention the `Cache-Control` header somewhere as we assume that is set.

docs/guide/in-depth/advanced-chunks.md
```diff
@@ -1,6 +1,238 @@
...
+  return <Button onClick={() => alert('Button clicked!')} />; // [!code --]
+  return <Button onClick={() => alert('Button clicked!!!')} />; // [!code ++]
+}
+```
+
+and naturally, you get a `output-hash1.js` file with the same content as `output-hash0.js`, except the `App` function is changed.
+
+Now, if you deploy this new version of your application, the browser will need to download the entire `output-hash1.js` file, even though only a small part of it has changed. This is because the hash of the file has changed, and the browser will treat it as a new file.
+
+To solve this problem, we can use `advancedChunks` to split output libraries into separate chunks, because they don't change frequently.
```
sapphi-red:
```suggestion
To solve this problem, we can use `advancedChunks` to split output libraries into separate chunks, because they don't change frequently compared to application code.
```

docs/guide/in-depth/advanced-chunks.md
```diff
@@ -1,6 +1,238 @@
...
+"ReactDOM library code";
+
+export { ... };
+```
+
+:::
+
+- The `libs-hash0.js` file is not changed, so the browser can use the cached version of the file.
+- The `output-hash1.js` file is changed, so the browser will download the new version of the file.
+
```
sapphi-red, line 172: It may be worth adding a note that using a function that was tree-shaken in the previous bundle will invalidate the library chunk.

docs/guide/in-depth/advanced-chunks.md
```diff
@@ -65,3 +297,57 @@ When we run `node ./main.js`, the traversal order of the modules would be `main.
...
+export const value = 'a' + value;
+```
+
+:::
+
+You could see, to make `a.js` work, we have to change the export signature of the entry chunk `entry.js` and add an additional export `value`. This totally violates the original intention of the code, which is to only export `foo` from `entry.js`.
+
+Fortunately, rolldown supports [`InputOptions.preserveEntrySignatures: 'allow-extension'`](/reference/config-options#preserveentrysignatures) to let you inform the bundler that it is allowed to change the export signature of the entry chunk.
+
+Enabling [`InputOptions.preserveEntrySignatures: 'allow-extension'`](/reference/config-options#preserveentrysignatures) will prevent the bundler from capturing the dependencies of the captured module.
```
sapphi-red, line 353: Just to confirm, advanced chunks cannot solve "hash cascading problem", right?

docs/guide/in-depth/advanced-chunks.md
```diff
@@ -65,3 +297,57 @@ When we run `node ./main.js`, the traversal order of the modules would be `main.
 `second.js` tries to call `__esm` function before it gets initialized. This will lead to a runtime error which is trying to call `undefined` as a function.
 
 With forcefully generated `runtime.js`, the bundler ensures any chunk that depends on runtime code would first load `runtime.js` before executing itself. This guarantees that the runtime code is always executed before any other chunks, preventing circular import issues.
+
+### Why the dependencies of the captured module get captured too?
+
+When a module is captured by a group, rolldown will try to capture its dependencies recursively. This is because rolldown is only allowed to mangling exports of non-entry chunks by default.
```
sapphi-red:
```suggestion
When a module is captured by a group, Rolldown will try to capture its dependencies recursively. This is because Rolldown is only allowed to mangle the exports of non-entry chunks by default.
```

docs/guide/in-depth/advanced-chunks.md
```diff
@@ -65,3 +297,57 @@ When we run `node ./main.js`, the traversal order of the modules would be `main.
...
+
+// a.js
+import { value as valueB } from './b.js';
+export const value = 'a' + valueB;
+
+// b.js
+export const value = 'b';
+```
+
+Let's say we want to move the `a.js` module into a separate chunk with keeping the `b.js` in the same chunk as `entry.js`. We get
```
sapphi-red:
```suggestion
Let's say we want to move the `a.js` module into a separate chunk while keeping the `b.js` module in the same chunk as `entry.js`. We get
```

docs/guide/in-depth/advanced-chunks.md
```diff
@@ -65,3 +297,57 @@ When we run `node ./main.js`, the traversal order of the modules would be `main.
...
+
+// a.js
+export const value = 'a' + value;
+```
+
+:::
+
+You could see, to make `a.js` work, we have to change the export signature of the entry chunk `entry.js` and add an additional export `value`. This totally violates the original intention of the code, which is to only export `foo` from `entry.js`.
+
+Fortunately, rolldown supports [`InputOptions.preserveEntrySignatures: 'allow-extension'`](/reference/config-options#preserveentrysignatures) to let you inform the bundler that it is allowed to change the export signature of the entry chunk.
```
sapphi-red:
```suggestion
Fortunately, Rolldown supports [`InputOptions.preserveEntrySignatures: 'allow-extension'`](/reference/config-options#preserveentrysignatures) to let you inform the bundler that it is allowed to change the export signature of the entry chunk.
```

packages/rolldown/src/options/output-options.ts
```diff
@@ -111,32 +111,75 @@ export interface OutputOptions {
...
+       *   groups: [
+       *      {
+       *        name: 'react',
+       *       test: /node_modules\/react/,
+       *       priority: 1,
+       *      },
+       *      {
+       *       name: 'other-libs',
+       *       test: /node_modules/,
+       *      priority: 2,
```
sapphi-red:
```suggestion
       *        name: 'react',
       *        test: /node_modules\/react/,
       *        priority: 1,
       *      },
       *      {
       *        name: 'other-libs',
       *        test: /node_modules/,
       *        priority: 2,
```

packages/rolldown/src/options/output-options.ts
```diff
@@ -111,32 +111,75 @@ export interface OutputOptions {
...
+       *
+       * This is a clearly __incorrect__ example. Though `react` group is defined before `other-libs`, it has a lower priority, so the modules in `react` group will be captured in `other-libs` group.
        */
       priority?: number;
       /**
        * - Type: `number`
        * - Default: `0`
        *
-       * Minimum size of the desired chunk. If accumulated size of captured modules is smaller than this value, this group will be ignored.
+       * Minimum size of the desired chunk. If the final size of this group is smaller than this value, it will be ignored. Modules in this group will fall back to the `automatic chunking` if they are not captured by any other group.
```
sapphi-red:
```suggestion
       * Minimum size of the desired chunk. If the accumulated size of the captured modules by this group is smaller than this value, it will be ignored. Modules in this group will fall back to the `automatic chunking` if they are not captured by any other group.
```
I think "the accumulated size" is more clear here.

packages/rolldown/src/options/output-options.ts
```diff
@@ -150,7 +193,7 @@ export interface OutputOptions {
        * - Type: `number`
        * - Default: `Infinity`
        *
-       * If final size of this group is larger than this value, this group will be spit into multiple groups that each has size closed to this value.
+       * If the final size of this group is larger than this value, this group will be split into multiple groups that each has size closed to this value.
```
sapphi-red:
```suggestion
       * If the accumulated size of the captured modules by this group is larger than this value, this group will be split into multiple groups that each has size closed to this value.
```

docs/guide/in-depth/advanced-chunks.md
hyf0: Yeah. If I understand you correctly, it should be solved by future `importmap` support.

docs/guide/in-depth/advanced-chunks.md
hyf0: Maybe. That will make this guide more academic. I assume people whoe read this have this kind of knowledge. But definitely it's possitive to add such information.

docs/guide/in-depth/advanced-chunks.md
hyf0: Yes. Real-world case might be complicated. I'll add warning here.

hyf0:
Notes:

> I think we need to mention the Cache-Control header somewhere as we assume that is set.

> It may be worth adding a note that using a function that was tree-shaken in the previous bundle will invalidate the library chunk.

shulaoda (approved):

PR #4984, opened 2025-06-17
feat: make require('./foo.json') work as expected even if the json is transformed to JS by a custom plugin

### LINKED ISSUE #3640: [Feature Request]: make `require('./foo.json')` work as expected even if the `json` is transformed to JS by a custom plugin

When using the built-in json feature, both `const foo = require('./foo.json')` and `import foo from './foo.json'` returns a value aligned with Node.js (`foo` has the same value containing the value of `./foo.json`).
But when the json is handled by a plugin (like below),
```js
export const plugin = {
  name: 'my-json-plugin',
  load(id) {
    if (!id.endsWith('.json')) return
    const content = fs.readFileSync(id, 'utf8')
    return `export default ${content}`
  }
}
```
`import foo from './foo.json'` works as expected, but `const foo = require('./foo.json')` does not (`foo` has `{ default: contentOfJson }` instead).

This is a problem for rolldown-vite, as Vite has a built-in JSON plugin that has more advanced options that the builtin json feature in rolldown. ([reproduction that shows that it works in current Vite](https://stackblitz.com/edit/vitejs-vite-whoysxga?file=src%2Fmain.js&terminal=dev), [reproduction that shows that it does not work in Rolldown](https://stackblitz.com/edit/github-nci4o6nk-ivnq49ys?file=rolldown.config.mjs))

The reason why it works in rollup seems to be because commonjs plugin has some json specific code.
https://github.com/rollup/plugins/blob/94951774773a8797ff7e1a93d0a7ebf14f009327/packages/commonjs/src/dynamic-modules.js#L91

#### Option 1
Add support for `module.exports` export, which is supported by Node.js (https://github.com/nodejs/node/pull/54563).
With this the JSON plugin in Vite can add `module.exports` export to control what value `require('./foo.json')` should return.

```js
export default jsonContent
export { jsonContent as "module.exports" }
```
The downside of this is that we have to think about what should happen if the arbitrary module namespace specifier (https://github.com/tc39/ecma262/pull/2154) is not supported.

#### Option 2
Add some special handling for json files like the commonjs plugin does.

### Additional Information

I found this while testing rolldown-vite on ecosystem-ci.
The actual dependency I encountered this is https://github.com/jshttp/statuses.
https://github.com/jshttp/statuses/blob/b7b21ff6ea81746d4eba9b8b9d9abee6565da5be/index.js#L28

IWANABETHATGUY:
1. Compatible with the Rollup CommonJS plugin;
2. Note:  only works if `experimental.viteMode` is enabled cc @sapphi-red, this is because our default behavior is aalignedwith esbuild and webpack already (Both of them prefer interop correctness, so do we), I am hesitant to enable it by default.
Closed #3640


IWANABETHATGUY:
* **#4984** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4984?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/4984?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/rolldown/src/ast_scanner/mod.rs
```diff
@@ -355,6 +355,12 @@ impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
       rec.meta.insert(ImportRecordMeta::SAFELY_MERGE_CJS_NS);
     }
 
+    #[allow(clippy::case_sensitive_file_extension_comparisons)]
+    if self.options.experimental.vite_mode.unwrap_or_default() && module_request.ends_with(".json")
+    {
+      rec.meta.insert(ImportRecordMeta::JSON_MODULE);
```
hyf0, line 361: Should we detect module type too?

hyf0 (approved):

crates/rolldown/src/ast_scanner/mod.rs
IWANABETHATGUY: https://github.com/rolldown/rolldown/blob/3802329087c464c9c89260af98bf41246cbe8ad7/crates/rolldown/src/module_finalizers/mod.rs?plain=1#L755-L758, maybe not, we already handle it if the module type is json.

sapphi-red:
Just to confirm, is this a temporary solution until "Option 1" in https://github.com/rolldown/rolldown/issues/3640 (`module.exports` export) is implemented? (As I believe we will support that one in the future)

IWANABETHATGUY:
> Just to confirm, is this a temporary solution until "Option 1" in #3640 (`module.exports` export) is implemented? (As I believe we will support that one in the future)

After the pr is landed the option one is no longer needed, we already use a different way to handle `require('.json')` even if the moduleType is `js`, although it is a little bit tricky, just like the rollup commonjs plugin. See the test as an example.

IWANABETHATGUY:
I test with https://github.com/rolldown/rolldown/issues/3640#issuecomment-2977747535, it works fine when enable `viteMode`

sapphi-red:
> > Just to confirm, is this a temporary solution until "Option 1" in #3640 (`module.exports` export) is implemented? (As I believe we will support that one in the future)
> 
> After the pr is landed the option one is no longer needed, we already use a different way to handle `require('.json')` even if the moduleType is `js`, although it is a little bit tricky, just like the rollup commonjs plugin. See the test as an example.

Yes, this PR would work for Vite's case. But I think we'll still need to implement Option 1 in the future as it's supported by Node.js. Or do we plan not to support that intentionally?

IWANABETHATGUY:
> > > Just to confirm, is this a temporary solution until "Option 1" in #3640 (`module.exports` export) is implemented? (As I believe we will support that one in the future)
> > 
> > 
> > After the pr is landed the option one is no longer needed, we already use a different way to handle `require('.json')` even if the moduleType is `js`, although it is a little bit tricky, just like the rollup commonjs plugin. See the test as an example.
> 
> Yes, this PR would work for Vite's case. But I think we'll still need to implement Option 1 in the future as it's supported by Node.js. Or do we plan not to support that intentionally?

No, I just mean for this case it is not needed. We could support it.

IWANABETHATGUY:
You could open another issue to track it if it is necessary.

PR #4991, opened 2025-06-17
chore(infra): add knip to ci

webpro:
Now that we've mostly established the `package.json` dependency exceptions in `knip.jsonc` we can consider adding knip to CI, so here's that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,6 +140,9 @@ jobs:
       - name: Lint Code
         run: pnpm lint-code
 
+      - name: Lint exports & dependencies
+        run: pnpm lint-knip
+
       - name: Validate workspace package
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -140,6 +140,9 @@ jobs:
       - name: Lint Code
         run: pnpm lint-code
 
+      - name: Lint exports & dependencies
+        run: pnpm lint-knip
```
hyf0, line 144: Could you also add it here https://github.com/rolldown/rolldown/blob/6702674df1738795aaeb1eaa70653365f0efd76f/justfile#L111

webpro: Sure, done

hyf0 (approved):
legit!

PR #4980, opened 2025-06-17
fix(plugin/vite-resolve): importing an optional peer dep should throw an runtime error

sapphi-red:
Port of https://github.com/vitejs/vite/pull/20029 and https://github.com/vitejs/rolldown-vite/pull/167.

PR #4995, opened 2025-06-18
fix(plugin/vite-resolve): optional peer dep id parse error

sapphi-red:
Follow-up to #4980.
It seems the new tests in Vite repo caught a bug.


sapphi-red:
Vite tests: https://github.com/rolldown/rolldown/actions/runs/15722474091 => passed ✅

shulaoda (approved):

PR #5013, opened 2025-06-19
fix(plugin/vite-resolve): normalize leading slash

sapphi-red:
Resovle ids like `//absolute/path` to align with the Vite's JS resolver.


sapphi-red:
https://github.com/rolldown/rolldown/actions/runs/15752776280

IWANABETHATGUY (approved):

PR #5014, opened 2025-06-19
docs: explain that a chunk may be bigger than `maxSize`

sapphi-red:
Added a section to explain that a chunk may be bigger than `maxSize` option.


hyf0:
Let me add more details

hyf0 (approved):
I create a todo task for myself to add more details on this aspect :).

PR #5046, opened 2025-06-23
fix: remove improper jsdoc of  builtin define plugin

### LINKED ISSUE #4257: Remove improper jsdoc of builtin define plugin
Remove lines at

https://github.com/rolldown/rolldown/blob/d53650c52948c411f2aaa9df535e8774f8bfa661/packages/rolldown/src/builtin-plugin/replace-plugin.ts#L13

https://github.com/rolldown/rolldown/blob/d53650c52948c411f2aaa9df535e8774f8bfa661/packages/rolldown/src/builtin-plugin/replace-plugin.ts#L22

---old

While the [Rollup plugin supports](https://github.com/rollup/plugins/blob/e1a5ef99f1578eb38a8c87563cb9651db228f3bd/packages/replace/types/index.d.ts#L4) not only plain strings but also functions returning a string (id as first argument) *and [the example JSdoc](https://github.com/rolldown/rolldown/blob/d53650c52948c411f2aaa9df535e8774f8bfa661/packages/rolldown/src/builtin-plugin/replace-plugin.ts#L13)* also indicate that, the rolldown plugin will fail when a function is being passed.

Either support it (would be preferred for feature parity I suppose) or make clear that functions are not supported.

[Minimal Reproduction](https://stackblitz.com/edit/rolldown-4257)

AliceLanniste:
close #4257

### DIFF
--- a/packages/rolldown/src/builtin-plugin/replace-plugin.ts
+++ b/packages/rolldown/src/builtin-plugin/replace-plugin.ts
@@ -10,7 +10,6 @@ import { BuiltinPlugin } from './constructors';
  * ```js
  * replacePlugin({
  *   'process.env.NODE_ENV': JSON.stringify('production'),
- *    __buildDate__: () => JSON.stringify(new Date()),
  *    __buildVersion: 15
  * })
  * ```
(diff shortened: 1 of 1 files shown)


hyf0 (approved):
Thank you!

PR #5050, opened 2025-06-24
refactor: passing vec length directly

ShroXd:
Refactor the constructor of ChunkGraph so that it directly accepts the module vector length.


IWANABETHATGUY (approved):
Thanks!

PR #5053, opened 2025-06-24
docs(contrib-guide): add memory profile steps

sapphi-red:
Added memory profile steps that worked for me.


hyf0 (approved):

docs/contrib-guide/heaptrack-gui.png
Boshen, line 1: We probably don't want to include a large binary in git :-/

IWANABETHATGUY: This asset is used for documentation, so maybe it is fine? It is more like an asset rather than a binary.

hyf0: It deserves a compression at least I think.

Boshen: I thought it's pretty well known that git stores files per commit, 600kb per commit will grow your git repository very quickly.

sapphi-red: git doesn't duplicate the content unless it is changed. So unless you change the image in every commit, it won't grow that quickly.

Boshen: Ok, let's this how it goes.

PR #5058, opened 2025-06-25
perf: reduce temporary memory usage by avoiding output chunk clone until needed

sapphi-red:
These two lines clones the whole bundle output. This is because `args.bundle` is `Vec<Output>` where `Output` is a enum that contains a `Box`.
https://github.com/rolldown/rolldown/blob/35d8e1569d820347646a014fca28ed32457f912c/crates/rolldown_binding/src/options/plugin/js_plugin.rs#L483
https://github.com/rolldown/rolldown/blob/35d8e1569d820347646a014fca28ed32457f912c/crates/rolldown_binding/src/options/plugin/js_plugin.rs#L510

While this `.clone` is inevitable to allow referencing the `bundle` parameter of `generateBundle` hook after the hook has end, the content of `bundle` parameter does not need to have a separate value, because it is readonly (the modification is done via `ChangedOutputs` instead). This PR changes `Box` to `Arc` so that we can share the underlying value.

This PR will have a bigger impact if the output is large. For example, `astro build` with https://github.com/withastro/astro.build used ~8GB of memory previously (if the GC does not run frequently enough). With this PR, it now uses ~3GB.

refs #1082

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/render_chunk_to_assets.rs
+++ b/crates/rolldown/src/stages/generate_stage/render_chunk_to_assets.rs
@@ -1,4 +1,4 @@
-use std::ops::Deref;
+use std::{ops::Deref, sync::Arc};
 
 use futures::future::try_join_all;
 use oxc_index::{IndexVec, index_vec};
@@ -88,7 +88,7 @@ impl GenerateStage<'_> {
             )
             .await?
             {
-              output_assets.push(Output::Asset(Box::new(sourcemap_asset)));
+              output_assets.push(Output::Asset(Arc::new(sourcemap_asset)));
             }
           }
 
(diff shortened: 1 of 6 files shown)


sapphi-red:
* **#5058** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5058?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5058?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sapphi-red:
Running Vite tests, just in case: https://github.com/rolldown/rolldown/actions/runs/15867992355

Update: Passed ✅

shulaoda (approved):

IWANABETHATGUY (approved):
Nice

PR #5068, opened 2025-06-25
fix(watch): add paths to watcher in batch

### LINKED ISSUE #3890: Watch mode seems to hang in larger project
### Discussed in https://github.com/rolldown/rolldown/discussions/3889

<sup>Originally posted by **tbrannam** March 18, 2025</sup>
We're seeing impressive build time improvements when porting from Rollup to Rolldown, but when we attempt to use `--watch` mode in the build environment, we see Rolldown not complete building the output.  

I'm sure there's something specific to the environment or codebase, as watch seems to work fine in a trivial example.

The output from the cli
```
Waiting for changes...                                                                                                                                                                                                                                                                                                              
✔ Rebuilt lib/web/esm in 1.04 s.                                                                                                                                                                                                                                                                                                   
Found src/web/fragment.tsx changed, rebuilding...                                                                                                                                                                                                                                                                                   
```
From the console it seems that our project does not finish rebuilding, manually inspecting the output file we can see that source changes are not emitted.

I've attempted to set RD_LOG and the volume of output is very large, so I won't post it here without some guidance of what is helpful to narrow this down.

I've tried adding some exclusions to the watch.exclude property, and I've hacked the source code to remove any Rollup plugins from the build environment.  

Please chime in if you have some suggestions on how to narrow down the problem.

</div>

branchseer:
fixes https://github.com/rolldown/rolldown/issues/3890#issuecomment-2988960197.

 `notify-rs` PR: https://github.com/notify-rs/notify/pull/692

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1465,6 +1465,7 @@ checksum = "610a5acd306ec67f907abe5567859a3c693fb9886eb1f012ab8f2a47bef3db51"
 [[package]]
 name = "notify"
 version = "8.0.0"
+source = "git+https://github.com/branchseer/notify?branch=watch-batch#fc215862bdc328649cb8ff1f15c0917f182450e9"
 dependencies = [
  "bitflags 2.9.1",
  "filetime",
@@ -1482,6 +1483,7 @@ dependencies = [
 [[package]]
 name = "notify-types"
 version = "2.0.0"
+source = "git+https://github.com/branchseer/notify?branch=watch-batch#fc215862bdc328649cb8ff1f15c0917f182450e9"
 
 [[package]]
 name = "nu-ansi-term"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1464,11 +1464,11 @@ checksum = "610a5acd306ec67f907abe5567859a3c693fb9886eb1f012ab8f2a47bef3db51"
 
 [[package]]
 name = "notify"
-version = "8.0.0"
-source = "git+https://github.com/branchseer/notify?branch=watch-batch#fc215862bdc328649cb8ff1f15c0917f182450e9"
+version = "8.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


shulaoda:
Most of the performance issues stem from the `watch` configuration:
```js
watch: {
  skipWrite: false,
  include: files,
}
```
where `include: files` explicitly includes all files. This causes the logic at [watcher_task.rs#L126-L133](https://github.com/rolldown/rolldown/blob/a4ec15eea960f08f8c0596d2c20de85e36e76f78/crates/rolldown/src/watch/watcher_task.rs#L126-L133) to become very slow.

This PR fixes issues related to `notify`, so during testing the include option should not be set and all files are included by default.

You can set the `input` option in the user configuration to `'./src/index.ts'`, which is the correct usage and facilitates reproduction and testing.

branchseer:
@shulaoda Thanks for the suggestion. The repro example is updated accordingly at https://github.com/branchseer/rolldown-repro-3890/commit/7b6870504f278ae6a1d47da1eb9f5b86cc9fc627.

`rolldown@1.0.0-beta.19`:
```
> node rolldown2.mjs
BuildStart in 10 ms
Transform in 14 ms
BuildEnd in 53 ms
RenderStart in 56 ms
- Bundle completed in 65 ms
BuildStart in 1 ms
Watcher started
Bundling started...
Transform in 0 ms
BuildEnd in 23 ms
RenderStart in 2262 ms
Bundle completed in 2265 ms
All bundles completed
^C
```

With this fix:
```
> node rolldown2.mjs
BuildStart in 10 ms
Transform in 16 ms
BuildEnd in 67 ms
RenderStart in 69 ms
- Bundle completed in 78 ms
BuildStart in 1 ms
Watcher started
Bundling started...
Transform in 0 ms
BuildEnd in 23 ms
RenderStart in 55 ms
Bundle completed in 66 ms
All bundles completed

```

shulaoda:
Nice!

branchseer:
https://github.com/notify-rs/notify/pull/692 merged. Let's wait for the release.

shulaoda (approved):

PR #5075, opened 2025-06-26
feat(debug): support to emit `meta.json`

hyf0:
- Closes https://github.com/rolldown/rolldown/issues/4134.
- Closes https://github.com/rolldown/rolldown/pull/4205.

Overview
```
{"timestamp":1750921704146,"action":"SessionMeta","inputs":[{"name":"entry","import":"./index.ts"}],"plugins":[{"name":"builtin:oxc-runtime","index":0},{"name":"builtin:data-uri","index":1}],"cwd":".../rolldown/examples/basic-typescript","platform":"browser","format":"esm","dir":null,"file":null,"build_id":"bid_0_count_1","session_id":"sid_0_1750921704070"}

```

PR #5078, opened 2025-06-26
feat(rolldown): oxc_resolver v11.3.0

Boshen:
For https://github.com/vitejs/vite/pull/20252

### DIFF
--- a/crates/rolldown_resolver/src/resolver.rs
+++ b/crates/rolldown_resolver/src/resolver.rs
@@ -121,6 +121,8 @@ impl<F: FileSystem + Default> Resolver<F> {
       symlinks: raw_resolve.symlinks.unwrap_or(true),
       builtin_modules,
       module_type: true,
+      // This is not part of the spec, but required to align with vite.
+      allow_package_exports_in_directory_resolve: true,
     };
     let resolve_options_with_import_conditions = OxcResolverOptions {
       condition_names: import_conditions,
--- a/crates/rolldown_resolver/src/resolver.rs
+++ b/crates/rolldown_resolver/src/resolver.rs
@@ -121,8 +121,7 @@ impl<F: FileSystem + Default> Resolver<F> {
       symlinks: raw_resolve.symlinks.unwrap_or(true),
       builtin_modules,
       module_type: true,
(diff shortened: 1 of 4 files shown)


Boshen:
* **#5078** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5078?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5078?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/rolldown_resolver/src/resolver.rs
```diff
@@ -121,6 +121,8 @@ impl<F: FileSystem + Default> Resolver<F> {
       symlinks: raw_resolve.symlinks.unwrap_or(true),
       builtin_modules,
       module_type: true,
+      // This is not part of the spec, but required to align with vite.
+      allow_package_exports_in_directory_resolve: true,
```
hyf0: Should we enable this by default? We should only enable this when `experimental#vite_mode` is true.

Or `rolldown-vite` is using `rolldown_plugin_vite_resolve`, which should enable this in that part of code.

@sapphi-red cc

Boshen: I don't know :-) You guys need to configure this correctly for ecosystem to work correctly 😅

sapphi-red: Pushed a commit to only set this in the vite resolve plugin 👍

sapphi-red: @hyf0 I tried to try this PR with pkg.pr.new. But the build seems to be broken by https://github.com/rolldown/rolldown/pull/5075/files#diff-4ab5bf726d0813cd862123b2515e4201cc0c9b67031365c01107c10a0daf6b4bR28.

hyf0: I will fix it in https://github.com/rolldown/rolldown/pull/5080. I think it's ok to merge it without testing it.

hyf0 (approved):

PR #5095, opened 2025-06-29
feat: commonjs tree shaking.

IWANABETHATGUY:
Finish linking and codegen for commonjs tree shaking, adding a testcase


crates/rolldown/tests/esbuild/dce/package_json_side_effects_true_keep_common_js/artifacts.snap
```diff
@@ -9,7 +9,6 @@ source: crates/rolldown_testing/src/integration_test.rs
 
 //#region node_modules/demo-pkg/index.js
 var require_demo_pkg = __commonJS({ "node_modules/demo-pkg/index.js"(exports) {
-	exports.foo = 123;
```
IWANABETHATGUY, line 12: Expected since it is not used.

shulaoda (approved):

IWANABETHATGUY:
Feel free to reply if you found anything not expected, I will fixed  it in next pr.

PR #5147, opened 2025-07-02
feat(rolldown): oxc-resolver v11.5.0 (yarn pnp)

Boshen:
### [Yarn Plug'n'Play](https://yarnpkg.com/features/pnp)

- For node.js, yarn pnp should work without any configuration, given the following conditions:
  - the program is called with the `yarn` command, where the value [`process.versions.pnp`](https://yarnpkg.com/advanced/pnpapi#processversionspnp) is set.
  - `.pnp.cjs` manifest file exists in the closest directory, searched from the current working directory,
  - no multi-project setup, per second bullet point in [FIND_PNP_MANIFEST](https://yarnpkg.com/advanced/pnp-spec#find_pnp_manifest)


Boshen:
It feels like I'm missing a bunch of glue code, @shulaoda can you help me complete this PR?

I'm reluctant to add a integration test because that would require a whole yarn pnp setup. Perhaps just a test case for enabling the option and test for an error thrown by the resolver?

Boshen:
Things aren't working as expected, patching it oxc-resolver:

* https://github.com/oxc-project/oxc-resolver/pull/586

shulaoda (approved):

hyf0 (approved):

PR #5188, opened 2025-07-06
feat: support opt-in `topLevelVar`

### LINKED ISSUE #5116: [Feature Request]: Support opt-in `top-level-var`

https://esbuild.github.io/faq/#top-level-var 

A option to control if opt-in `top-level-var`

7086cmd:
resolves #5116
I'm unsure if my implementation and the result are correct. Please check if the behavior output in the test is accurate.


packages/rolldown/tests/fixtures/function/top-level-var/_config.ts
```diff
@@ -0,0 +1,13 @@
...
+import { defineTest } from 'rolldown-tests'
+
+export default defineTest({
+    config: {
+        output: {
+            topLevelVar: true,
+        },
+    },
+    afterTest(output) {
+        expect(output.output[0].code.includes('var greeting')).toBe(true)
```
IWANABETHATGUY: Also need ensure that none top level `VarDeclaration` should not be rewritten.

packages/rolldown/tests/fixtures/function/top-level-var/_config.ts
```diff
@@ -0,0 +1,13 @@
+import { expect } from 'vitest'
```
IWANABETHATGUY: You don't need to add test in both rust and js side.

packages/rolldown/tests/fixtures/function/top-level-var/_config.ts
7086cmd: In rust side, I have tested it. JavaScript side is used to test whether this option is passed to the rust side correctly because adding an option for Rolldown is tremendously complicated.

7086cmd (commented):
It indicates that the function is handled correctly.

7086cmd:
It appears that I forgot to handle the blocked (but not in function or other scopes) sections.

hyf0:
> It appears that I forgot to handle the blocked (but not in function or other scopes) sections.

You could refer to this method to check if current block is in the top level scope

https://github.com/rolldown/rolldown/blob/c142a18b640042699ad248c82c7681c8659f0dd2/crates/rolldown/src/ast_scanner/mod.rs#L227-L232

packages/rolldown/tests/fixtures/function/top-level-var/_config.ts
hyf0: > JavaScript side is used to test whether this option is passed to the rust side correctly

You cuold reuse this test for your purpose

https://github.com/rolldown/rolldown/blob/main/packages/rolldown/tests/fixtures/verify-options/_config.ts

7086cmd:
> It appears that I forgot to handle the blocked (but not in function or other scopes) sections.

It seems that [esbuild doesn't handle blocked declarations due to scope differences](https://esbuild.github.io/try/#YgAwLjI1LjUALS1idW5kbGUAZQBlbnRyeS5qcwBsZXQgZmlyc3RMZXZlbExldCA9ICJsZXQiCnZhciBmaXJzdExldmVsVmFyID0gInZhciIKY29uc3QgZmlyc3RMZXZlbENvbnN0ID0gImNvbnN0IgpjbGFzcyBGaXJzdExldmVsQ2xhc3Mge30KY29uc29sZS5sb2coZmlyc3RMZXZlbExldCwgZmlyc3RMZXZlbFZhciwgZmlyc3RMZXZlbENvbnN0LCBuZXcgRmlyc3RMZXZlbENsYXNzKCkpOwoKaWYgKHRydWUpIHsKICBsZXQgc2hvdWxkTm90QmVTdWJzdGl0dXRlZExldCA9ICJsZXQiOwogIGNvbnNvbGUubG9nKHNob3VsZE5vdEJlU3Vic3RpdHV0ZWRMZXQpCn0KCmZ1bmN0aW9uIHNlY29uZF9sZXZlbCgpIHsKICAgIGxldCBzZWNvbmRMZXZlbExldCA9ICJsZXQiOwogICAgdmFyIHNlY29uZExldmVsVmFyID0gInZhciI7CiAgICBjb25zdCBzZWNvbmRMZXZlbENvbnN0ID0gImNvbnN0IjsKICAgIGNsYXNzIFNlY29uZExldmVsQ2xhc3Mge30KCiAgICBjb25zb2xlLmxvZyhzZWNvbmRMZXZlbExldCwgc2Vjb25kTGV2ZWxWYXIsIHNlY29uZExldmVsQ29uc3QsIG5ldyBTZWNvbmRMZXZlbENsYXNzKCkpOwp9CgpzZWNvbmRfbGV2ZWwoKTsK), so there may be no need to hold the scenario I demonstrated above.

crates/rolldown/src/module_finalizers/mod.rs
```diff
@@ -1168,6 +1169,49 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
...
+            let class_id = self.snippet.builder.binding_pattern(
+              BindingPatternKind::BindingIdentifier(self.snippet.builder.alloc(class_name)),
+              NONE,
+              false,
+            );
+            let expr = self.snippet.builder.expression_class(
+              SPAN,
+              ClassType::ClassExpression,
+              class_decl.decorators.take_in(self.alloc),
+              None,
```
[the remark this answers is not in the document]
7086cmd: @hyf0, how do you think of this? Here, I suppose "ClassExpression" should not hold a name.

IWANABETHATGUY (approved):

PR #5227, opened 2025-07-09
feat: support `chunkModulesOrder` 

IWANABETHATGUY:
opt-in support sort chunk modules by their moduleId, only if
- All those modules are leaf module (without any dependency)
- Those modules should not have side effects

I have tried other ways: 
- construct a matrix that could `O(1)` search if  module `a` is a descendant of module `b` but the time complexity is `O(n ^ 3)` and `O(n ^2)` space complexity, `n` is the count of module in module graph , one case would be: 
`a` -> `b` -> `c` -> .....  -> `n`, they you need to calc descendant of `n`, `n-1` ... `a`, that simply equals to `1 + 2 + 3 + 4 + ... + n`, this is only for one entry if it is a fully connected module then the overall complexity would be `O(n ^3)`, But since we would bailout for circular reference, a more proper estimitate would be `O(n ^2)` which is still not scale, the worse thing is this sometimes leads a none total order sorting, so I came up such a simple strategy, but it is good enoguh for now.

### DIFF
--- a/crates/rolldown/src/chunk_graph.rs
+++ b/crates/rolldown/src/chunk_graph.rs
@@ -1,7 +1,12 @@
 use arcstr::ArcStr;
+use itertools::Itertools;
 use oxc_index::{IndexVec, index_vec};
-use rolldown_common::{Chunk, ChunkIdx, ChunkTable, ModuleIdx, SymbolRef};
+use rolldown_common::{
+  Chunk, ChunkIdx, ChunkModulesOrderBy, ChunkTable, EcmaViewMeta, ModuleIdx, SymbolRef,
+};
 use rustc_hash::FxHashMap;
+
+use crate::{SharedOptions, stages::link_stage::LinkStageOutput};
 
 #[derive(Debug)]
 pub struct ChunkGraph {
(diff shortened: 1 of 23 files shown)


IWANABETHATGUY:
* **#5227** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5227?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5227?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/rolldown/src/chunk_graph.rs
```diff
@@ -42,4 +47,45 @@ impl ChunkGraph {
...
   }
+
+  pub fn sort_chunk_modules(&mut self, link_output: &LinkStageOutput, options: &SharedOptions) {
+    // Sort modules in each chunk by execution order
+    self.chunk_table.iter_mut().for_each(|chunk| {
+      if matches!(
+        options.experimental.chunk_modules_order.unwrap_or_default(),
+        ChunkModulesOrderBy::ExecOrder
+      ) {
+        chunk.modules.sort_by_cached_key(|idx| link_output.module_table[*idx].exec_order());
```
shulaoda, line 58: Out of curiosity, are there any specific advantages of using `sort_by_cached_key` over `sort_unstable_by_key`?

shulaoda (approved):

crates/rolldown/src/chunk_graph.rs
IWANABETHATGUY: Nice catch, I think it is not necessary, since it is expected that those items would not be duplicated.

PR #5231, opened 2025-07-10
refactor(rolldown_debug,rolldown_tracing): remove `EnvFilter`

Boshen:
`EnvFilter` (enabled through the `env-filter` feature flag) includes an outdated regex engine, which increases the binary size by a significant amount.

With the change in this PR, binary size is decreased from 15375248  to
14862624. i.e. 15.37mb - 14.86mb = 0.51mb.

### Testing

I tested

* `RD_LOG=debug`
* `RD_LOG='oxc_resolver'`
* `debug: true` in rolldown config, observed correct output in the `.rolldown/0000000000000/` directory.

### DIFF
--- a/crates/rolldown_debug/src/init_tracing.rs
+++ b/crates/rolldown_debug/src/init_tracing.rs
@@ -1,10 +1,8 @@
-use std::str::FromStr;
 use std::sync::Arc;
 use std::sync::atomic::AtomicBool;
 
-use tracing_subscriber::EnvFilter;
-use tracing_subscriber::fmt;
-use tracing_subscriber::prelude::*;
+use tracing::Level;
+use tracing_subscriber::{filter::FilterFn, fmt, prelude::*};
 
 use crate::debug_data_propagate_layer::DebugDataPropagateLayer;
 use crate::debug_formatter::DebugFormatter;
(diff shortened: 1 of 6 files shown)


crates/rolldown_debug/src/init_tracing.rs
```diff
@@ -14,21 +12,6 @@ use crate::static_data::OPENED_FILES_BY_SESSION;
 
 static IS_INITIALIZED: AtomicBool = AtomicBool::new(false);
 
-static FILTER_FOR_DEVTOOL: &str = "[{meta}]=trace";
-
-pub fn init_devtool_tracing() {
```
Boshen, line 19: This function is unused.

crates/rolldown_tracing/Cargo.toml
```diff
@@ -21,4 +21,4 @@ test = false
 [dependencies]
 tracing = { workspace = true, features = ["valuable"] }
 tracing-chrome = { workspace = true }
-tracing-subscriber = { workspace = true, features = ["env-filter", "json", "valuable"] }
```
Boshen, line 24: I'm not seeing `json` and `valuable` used any where, disabling them is not breaking anything?

hyf0 (approved):

PR #5238, opened 2025-07-10
chore(deps): do not enable `full` feature for `tokio`

Boshen:


Boshen:
The feature flag is on `rolldown_bench`, it won't affect production.

PR #5239, opened 2025-07-10
feat: add `watch.onInvalidate`

### LINKED ISSUE #5008: [Feature Request]: `watch.onInvalidate`

tsdown needs to watch the config file and, instead of triggering the next rebuild, restart the entire rolldown process.

https://rollupjs.org/configuration-options/#watch-oninvalidate

situ2001:
close #5008 

The basic idea is to create js binding `onInvalidate` and expose it to `watchOption`, then call it when watcher task is marked as invalidate.


IWANABETHATGUY (approved):
Thanks

PR #5273, opened 2025-07-14
perf: use `IndexBitSet<Idx>` instead of `IndexVec<Idx, bool>`

sapphi-red:
Implemented `IndexBitSet<Idx>` and replaced all `IndexVec<Idx, bool>` with that. This should reduce the memory usage a bit in theory.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/advanced_chunks.rs
+++ b/crates/rolldown/src/stages/generate_stage/advanced_chunks.rs
@@ -6,6 +6,7 @@ use rolldown_common::{
   Chunk, ChunkKind, ChunkingContext, MatchGroupTest, Module, ModuleIdx, ModuleTable,
 };
 use rolldown_error::BuildResult;
+use rolldown_utils::index_bitset::IndexBitSet;
 use rustc_hash::{FxHashMap, FxHashSet};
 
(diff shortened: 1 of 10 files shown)


sapphi-red:
* **#5273** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5273?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5273?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

IWANABETHATGUY:
I don't find any observable perf improvement in codespeed and criterion bench. We need to investigate what is the hotpot, if you want reduce memory usage, maybe https://github.com/RoaringBitmap/roaring-rs is a better choice.

sapphi-red:
I'll close this one as it's a change that I came up without any profiling.

IWANABETHATGUY:
> I'll close this one as it's a change that I came up without any profiling.

I don't mean it is not good, reducing memory usage doesn't always mean faster, you could open it, and we could compare different implementations then choose a better data structure to store the boolean flag status.

sapphi-red:
I checked the memory usage reduction and it wasn't significant. So I think it's not worth investigating for now.

PR #5378, opened 2025-07-21
feat: add warning and conversion for `import.meta` for non-`esm` output format

### LINKED ISSUE #5343: [Feature Request]: polyfill `import.meta`

https://github.com/rolldown/tsdown/issues/370#issuecomment-3088926309

- convert `import.meta` to {} in cjs format
- giving a warning if user source code has `import.meta` in cjs format

At least we should finish task two

n/a

7086cmd:
resolves #5343.
It appears that code already exists for normalizing options regarding the correction of `import.meta`. However, we need to detect it during pre-processing of the AST, so let's implement it later via `Visit`.

### DIFF
--- a/crates/rolldown/src/ast_scanner/impl_visit.rs
+++ b/crates/rolldown/src/ast_scanner/impl_visit.rs
@@ -257,6 +257,7 @@ impl<'me, 'ast: 'me> Visit<'ast> for AstScanner<'me, 'ast> {
           let static_name = member_expr.static_property_name().unwrap_or(Atom::from(""));
           static_name == "url" || static_name == "dirname" || static_name == "filename"
         })
+        // Here we need to set it to `false` to emit warnings when leaving `import.meta` alone along with the logic `not` head of this.
         .unwrap_or(false)
         && !self.options.format.keep_esm_import_export_syntax()
         && it.meta.name == "import"
--- a/crates/rolldown/src/ast_scanner/impl_visit.rs
+++ b/crates/rolldown/src/ast_scanner/impl_visit.rs
@@ -1,3 +1,5 @@
+use oxc::ast::ast::MetaProperty;
+use oxc::span::Atom;
 use oxc::{
   ast::{
     AstKind,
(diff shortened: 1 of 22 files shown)


crates/rolldown/src/ast_scanner/impl_visit.rs
```diff
@@ -247,6 +249,32 @@ impl<'me, 'ast: 'me> Visit<'ast> for AstScanner<'me, 'ast> {
...
+        .as_member_expression_kind()
+        .map(|member_expr| {
+          let static_name = member_expr.static_property_name().unwrap_or(Atom::from(""));
+          static_name == "url" || static_name == "dirname" || static_name == "filename"
+        })
+        .unwrap_or(false)
+        && !self.options.format.keep_esm_import_export_syntax()
+        && it.meta.name == "import"
+        && it.property.name == "meta"
+      {
```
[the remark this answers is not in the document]
7086cmd: Disagreed. It didn't understand what I thought, and I would comment on why to use `true` here.

crates/rolldown_error/src/events/empty_import_meta.rs
```diff
@@ -0,0 +1,48 @@
...
+    diagnostic.title = format!(
+      "`import.meta` is unavailable with the \"{}\" output format and will be empty.",
+      self.format
+    );
+
+    diagnostic.add_label(
+      &file_id,
+      self.span.start..self.span.end,
+      String::from(
+        "Avoid using `import.meta` in this context, or set the output format to `esm` instead.",
```
IWANABETHATGUY: I prefer diagnostic message like esbuild https://hyrious.me/esbuild-repl/?version=0.23.0&b=e%00entry.js%00import.meta&b=%00file.js%00&o=%7B%0A++treeShaking%3A+true%2C%0A++external%3A+%5B%22c%22%2C+%22a%22%2C+%22b%22%5D%2C%0A%22bundle%22%3A+true%2C%0Aformat%3A+%22cjs%22%0A%7D

crates/rolldown/tests/esbuild/default/import_meta_common_js/diff.md
```diff
@@ -25,6 +25,6 @@ console.log(require("url").pathToFileURL(__filename).href, void 0);
 @@ -1,2 +1,1 @@
 -var import_meta = {};
 -console.log(import_meta.url, import_meta.path);
-+console.log(require("url").pathToFileURL(__filename).href, void 0);
++console.log(require("url").pathToFileURL(__filename).href, {});
```
IWANABETHATGUY: I prefer to transform `import.meta.path` into `{}.path`, which is similar as esbuild, we should preserve the original user code as much as possible.

7086cmd: Completed. Caused by duplicated implementations btw.

crates/rolldown_error/src/events/empty_import_meta.rs
7086cmd: Completed.

IWANABETHATGUY (approved):

PR #5396, opened 2025-07-22
fix: top level VarDeclaration is not hoisted when wrapped in `__esm`

### LINKED ISSUE #5393: [Bug]: `strictExecutionOrder: true` breaks functions that references variables declared in initializer of `for` blocks

https://repl.rolldown.rs/#eNptkEFPhDAQhf/KpJctLlbYRBNZPerVBI+uBxZahLCdzdCuawj/3SksURMvbd+81+nXGYQR2SC0dfSl2j6crch+dCzKIM9HJAdXYAgPsFI3BpHdFdtaZI68HmMx15YOFzXfN0ggTwVBC4/w9h5Dznuy5e0BNrd3W1iv82hHOwvQqqPvP6TMYR2sSDl8ddTYWqYser/vHck0mtIXLONt6Rq08CwjGOY2JdoeO606rGU7hcewCOYk7LoKP63ijGnqX8z/OH/+X2lT+M4tb3BVU3PgWRVdthQBGLAp3dNZlz5QvVClKYMwpDkxgYxiZJQT905VopLrvXaF2tyL8RsSQ4Dw

### What is expected?

`j` to be declared in a scope that can be referenced from `F` like:
```js
//#region foo.js
function F() {
	console.log(j);
}
var j;
var init_foo = __esm({ "foo.js"() {
	j = [];
	for (var R = 0; R < 256; ++R) j.push((R + 256).toString(16).substr(1));
} });

//#endregion
```

`j` is declared in a closure and cannot be referenced from `F`.
```js
//#region foo.js
function F() {
	console.log(j);
}
var init_foo = __esm({ "foo.js"() {
	for (var j = [], R = 0; R < 256; ++R) j.push((R + 256).toString(16).substr(1));
} });

//#endregion
```

### System Info

```Shell
1.0.0-beta.29 REPL
```

### Any additional comments?

Originally reported at https://github.com/vitejs/rolldown-vite/issues/327

IWANABETHATGUY:
Closed #5393


IWANABETHATGUY:
* **#5397** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5397?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5396** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5396?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5396?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/rolldown/tests/esbuild/splitting/splitting_hybrid_esm_and_cjs_issue617/artifacts.snap
```diff
@@ -19,7 +19,9 @@ export { foo };
 var a_exports = {};
 __export(a_exports, { foo: () => foo });
 var foo;
-var init_a = __esm({ "a.js"() {} });
+var init_a = __esm({ "a.js"() {
+	;
```
IWANABETHATGUY, line 23: This regression is caused by a more general implementation of VarDecl to AssignmentExpr transformation (Unlike before, it could return None, so the EmptyStatement will not be generated at all). However, it should remain the same as before after the minification.

IWANABETHATGUY:
recommend use https://app.semanticdiff.com/gh/rolldown/rolldown/pull/5396/changes#crates/rolldown/src/module_finalizers/finalizer_context.rs?ignore_comments=false, it could reduce some diff noise.

hyf0 (approved):

crates/rolldown/tests/esbuild/splitting/splitting_hybrid_esm_and_cjs_issue617/artifacts.snap
hyf0: I think we should figure out a way to solve this in the long run.

IWANABETHATGUY: It could be solved by storing the emptyStmt Address, and then `filter_out` after `visit_stmts`, but I think that is not deserved to do for now.

PR #5404, opened 2025-07-23
feat: using frequent characters first

### LINKED ISSUE #5270: [Feature Request]: generate internal export names using frequent characters first

Currently `output.minifyInternalExports` generates export names incrementally from `'a'`.
https://github.com/rolldown/rolldown/blob/a3a4228c16289091eccb66451ce47785f7ecadbe/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs#L475-L476
But given that the source code normally contains some characters more than other characters, it is better to use those characters first so that the minified code is smaller when compress by gzip.

oxc_mangler uses this pre-computed character list.
https://github.com/oxc-project/oxc/blob/f717da9ba8449afe5b5e9e6730b45efc8927877b/crates/oxc_mangler/src/base54.rs#L30-L31
I think we can use this list for `output.minifyInternalExports` as well.

Not an API change.

AliceLanniste:
close #5270


IWANABETHATGUY:
Can you try some project with this PR and see if there is any gzip improvement?

sapphi-red (changes requested):
The character frequency should not be calculated here as the code is not minified yet (which will change the frequency a lot).
The pre-computed character list should be used.

AliceLanniste:
> The character frequency should not be calculated here as the code is not minified yet (which will change the frequency a lot). The pre-computed character list should be used.

Do you mean to use the to_base64 function in `compute_corss_chunk`?  I dont understand the meaning of the pre-computed character list.
now I get the character frequency list.  Is this considered a pre-computed character list? If not, how do I get a pre-computed list?

sapphi-red:
I mean this list by "the pre-computed character list".
https://github.com/oxc-project/oxc/blob/f717da9ba8449afe5b5e9e6730b45efc8927877b/crates/oxc_mangler/src/base54.rs#L30-L31

PR #5443, opened 2025-07-26
feat(pluginutils): add `suffixRegex` & support multiple string

btea:
Added `suffixRegex` function to facilitate plugins to filter files with specific suffixes.

The first parameter of `prefixRegex`/`suffixRegex`/`exactRegex` functions is expanded to support string arrays, making it easier to combine multiple conditions.

### DIFF
--- a/packages/pluginutils/README.md
+++ b/packages/pluginutils/README.md
@@ -17,6 +17,7 @@ import {
   exactRegex,
   makeIdFiltersToMatchWithQuery,
   prefixRegex,
+  suffixRegex,
 } from '@rolldown/pluginutils';
 
 // Match exactly 'foo.js'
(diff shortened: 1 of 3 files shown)


IWANABETHATGUY:
@sapphi-red , do you think if we need to expose such a function?

sapphi-red:
I don't find a use case for these. What is the expected usage?

For example, `suffixRegex('.vue')` can simply be written as `/\.vue$/` since it doesn't contain a dynamic input like `exactRegex(VIRTUAL_ID)` does. `exactRegex([a, b])` might be useful, but you can write it as `[a, b].map(v => exactRegex(v))` and I don't find it popular enough.

btea:
> For example, `suffixRegex('.vue')` can simply be written as `/\.vue$/` since it doesn't contain a dynamic input like `exactRegex(VIRTUAL_ID)` does. 

Yes, it is possible to write it like this, but I think it should be the same as `prefixRegex`, because `prefixRegex` is already supported, it would be useful to also support `suffixRegex`. And the function that can convert a string into a regular expression will be helpful for some users who are not familiar with writing regular expressions.

>`exactRegex([a, b])` might be useful, but you can write it as `[a, b].map(v => exactRegex(v))` and I don't find it popular enough.

This can also meet the requirements. But I think expanding function parameters should be more convenient and concise than writing two function calls.

IWANABETHATGUY:
close since it is stale.

PR #5448, opened 2025-07-26
fix(ci): fix DTS generation failure in Vite tests

shulaoda:
### Why the DTS step breaks
`rolldown_plugin_dts` fabricates a piece of JS during the `transform` hook that looks like:

```js
var ${binding} = [${symbolId}, () => ${dep}, /* … */, sideEffect()];
```
When the compressor runs with `unused: CompressOptionsUnused::Remove`, the unused var binding is stripped, leaving only
```js
[${symbolId}, () => ${dep}, /* … */, sideEffect()];
```
Because the later `renderChunk` hook expects the full `var … =` pattern to locate and patch this fake module, it can no longer recognise the code and the DTS generation fails.

I will discussing with @sxzz whether we should instead emit something like:

```js
const __DTS_FAKE__ = {};
__DTS_FAKE__.${binding} = [${symbolId}, () => ${dep}, /* … */, sideEffect()];
```

### Why `join_vars` and `sequences` should be kept disabled
Rolldown currently analyses side effects based on statements. When `join_vars` is enabled, a declaration such as `var a = 1; var b = 2;` would be merged as `var a = 1, b = 2;` but Rolldown always expands combined declarations back to individual statements, so it's unnecessary to enable `join_vars` on scan stage.

With `sequences`, successive simple statements are collapsed with the comma operator:
```js
let keep;
+keep; delete keep; ++keep;
```
->
```js
let keep;
+keep, delete keep, ++keep;
```
Because `++keep` has a side effect, the whole comma expression is flagged as having side effects. In strict mode the sub‑expression delete keep is then illegal (`"Delete of an unqualified identifier in strict mode"`), causing a semantic error.


shulaoda:
* **#5448** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5448?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5448?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Boshen (commented):
I need to understand the purpose of the DCE step, what should be considered as dead, and how they should interact with DTS.

Boshen:
DCE removes dead code. DTS plugin need to insert code that don't get removed.

PR #5460, opened 2025-07-27
perf: reduce runtime symbol memory usage and hashtable lookup

IWANABETHATGUY:
# Benefit
1. Before, each runtime symbol needed 8 bytes, now it only needs 4 bytes(Now we only need to store the StmtInfoIdx)
2. Less hashtable lookup(especially when `strictExecutionOrder` is enabled), before referencing each runtime symbol we need to lookup the real `SymbolRef` according to the `name`, now in the whole linking phase, we just store the `RuntimeHelperKind` and lookup them only once finally.
3. It is more convenient to reference runtime symbol at the chunk level after the linking phase, I will explain more details in the next pr, but even without this benefit, the perf is still good enough.

### DIFF
--- a/crates/rolldown/src/chunk_graph.rs
+++ b/crates/rolldown/src/chunk_graph.rs
@@ -2,7 +2,8 @@ use arcstr::ArcStr;
 use itertools::Itertools;
 use oxc_index::{IndexVec, index_vec};
 use rolldown_common::{
-  Chunk, ChunkIdx, ChunkModulesOrderBy, ChunkTable, EcmaViewMeta, ModuleIdx, SymbolRef,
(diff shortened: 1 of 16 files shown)


IWANABETHATGUY:
* **#5460** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5460?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5460?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

hyf0 (approved):

Issue #5469, opened 2025-07-28
[Tracking] Module Federation


Boshen:
> [!NOTE]
> After discussions with all Module Federation stakeholders, we have decided to converge the ecosystem on [module-federation/vite](https://github.com/module-federation/vite), which provides extensive support.
> Rolldown will no longer provide built-in Module Federation support.

---

~~Module Federation support is currently stale. We plan to reinitiate work on it after full-bundle mode is complete.~~

Current code around Module Federation is removed to reduce maintenance burden, and avoid people testing out the current broken version.

Related:

* https://github.com/rolldown/rolldown/pull/5471
* https://github.com/rolldown/rolldown/discussions/5431

zackarychapple:
I think its important to highlight that after a conversation with @TheAlexLichter that this does not mean that Rolldown is dropping support for Module Federation, but instead is going to re-implement it after adding the full-bundle mode. Is my understanding correct?

Boshen:
> I think its important to highlight that after a conversation with [@TheAlexLichter](https://github.com/TheAlexLichter) that this does not mean that Rolldown is dropping support for Module Federation, but instead is going to re-implement it after adding the full-bundle mode. Is my understanding correct?

Correct.

Also updated the issue description to reflect the current state.

Osbornnnnn:
Is it worth waiting for the first version of MF this year?

zackarychapple:
> Is it worth waiting for the first version of MF this year?

Hey @Osbornnnnn great question. Given that the changes to MF Runtime hitting 1.0 will be marking the stabilized api there is really no reason for Rolldown to wait for us to mark 1.0. If there ends up being any breaking changes we (MF core team) will work across the whole ecosystem to see the adjustments made.  I believe the only blockers on Rolldown at this time are some architecture adjustments that will enable easier support.

umararshad4:
whats the update on this

TheAlexLichter:
@umararshad4 Unchanged, otherwise we would've posted here! Full Bundle Mode is priority at the moment.

devinrhode2:
Sounds like rolldown team will help keep mfe/vite as the official solution

zackarychapple:
@devinrhode2 that is correct. The team has been very helpful. We've been working with them for some time now to make it solid.

zlunchbox:
[...]

PR #5471, opened 2025-07-28
feat(rolldown)!: remove unfinished module federation support

Boshen:
Module Federation support is currently stale. We plan to reinitiate work on it after full-bundle mode is complete.

Current code around Module Federation will be removed to reduce maintenance burden, and avoid people testing out the current broken version.

Please follow [tracking issue](https://github.com/rolldown/rolldown/issues/5469) for updates.


TheAlexLichter:
Let's make sure this is also mentioned in the docs so people aren't confused about the current state 👍🏻

Boshen:
> Let's make sure this is also mentioned in the docs so people aren't confused about the current state 👍🏻

The only mention so far is 

https://github.com/rolldown/rolldown/blob/afa8fc27e933ac10d0e9a3a061b85880602da07e/docs/guide/index.md?plain=1#L47

shulaoda (approved):

Issue #5451, opened 2025-07-27
[Feature Request]: Exposing full oxc `minify` options


IWANABETHATGUY:
### What problem does this feature solve?

https://discord.com/channels/1171783886200643635/1398468547176566916/1398468547176566916
related code https://github.com/rolldown/rolldown/blob/c100a53c6cfc67b4f92e230da072eef8494862ef/packages/rolldown/src/options/output-options.ts?plain=1#L73

### What does the proposed API look like?

just like https://github.com/rolldown/rolldown/blob/c100a53c6cfc67b4f92e230da072eef8494862ef/packages/rolldown/src/options/input-options.ts?plain=1#L17-L25

Boshen:
https://crates.io/crates/oxc_minify_napi is up.

AliceLanniste:
@Boshen hi,   can't import Minifyoptions directly, because of MinifyOptions is private

Boshen:
`MinifyOptions` should be `pub` now.

PR #5472, opened 2025-07-28
feat: support `oxc_minifier` options

AliceLanniste:
related #5451

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -199,19 +199,19 @@ napi-build = { version = "2.2.2" }
 napi-derive = { version = "3.0.0", default-features = false, features = ["type-def"] }
 
 # oxc crates with the same version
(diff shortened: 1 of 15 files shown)


IWANABETHATGUY:
Can you just refer to how `transformOptions` implemented, https://github.com/rolldown/rolldown/blob/4171c7d67d31cac7f643a65d185c0018d112a72c/crates/rolldown_plugin_transform/src/types/transform_options.rs?plain=1#L1-L60 I think they should be pretty similar.

IWANABETHATGUY:
https://github.com/rolldown/rolldown/blob/4171c7d67d31cac7f643a65d185c0018d112a72c/crates/rolldown_binding/src/options/binding_input_options/mod.rs?plain=1#L96, hard coding options is not good for maintenance.

IWANABETHATGUY:
The direction looks good, can you rebase the pr?

Issue #5301, opened 2025-07-16
[Bug]: `__rolldown_runtime__ is not defined` error happens when `advancedChunk` is used and a non sideeffect module exists


sapphi-red:
### Reproduction link or steps

1. Open https://stackblitz.com/edit/rolldown-rolldown-starter-stackblitz-aku3p1di?file=rolldown.config.mjs
2. `pnpm i` runs automatically
3. `pnpm build` runs automatically

### What is expected?

`__rolldown_runtime__` is declared before accessed

### What is actually happening?

`__rolldown_runtime__ is not defined` error happens

### System Info

```Shell
StackBlitz

  System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 20.19.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.8.2 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    rolldown: latest => 1.0.0-beta.27
```

### Any additional comments?

related: #5296

sapphi-red:
I think there's two points here.

First, this line should be checking whether this module relies on the execution order (e.g. reading a global variable), not whether it has a side effect.
https://github.com/rolldown/rolldown/blob/57a4bea49634b8d4f7c868ae2f5e4c9aae389f51/crates/rolldown/src/stages/link_stage/wrapping.rs#L39
This is because even if a module doesn't have a side effect (e.g. writing a global variable, console logging), it can rely on the side effect that is caused by other modules. So those modules should be wrapped to keep the execution order.

Second, since `__rolldown_runtime__` calls are injected in the later stage, `module.meta.contains(EcmaViewMeta::HAS_ANALYZED_SIDE_EFFECT)` returns `false`.

sapphi-red:
Made a separate issue for the first point: #5303

IWANABETHATGUY:
Since the optimization is opt-in and disabled by default, I reprioritize this issue to p2. `StrictExecutionOrder` would always wrap all modules now.

IWANABETHATGUY:
Closed due to https://github.com/rolldown/rolldown/pull/5498

sapphi-red:
[...]

PR #5498, opened 2025-07-29
fix: modules that are sideeffect free but relies on the execution order is not wrapped

### LINKED ISSUE #5303: [Bug]: modules that are sideeffect free but relies on the execution order is not wrapped

1. Open https://stackblitz.com/edit/rolldown-rolldown-starter-stackblitz-kahhkybh?file=rolldown.config.mjs
2. `pnpm i` runs automatically
3. `pnpm build` runs automatically
4. `node src/entry.js` does not throw an error
5. `node dist/entry.js` throws an error

### What is expected?

No error happens after build.

The following error happens after build:
```
ReferenceError: globalValue is not defined
```

### System Info

```Shell
StackBlitz

  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    rolldown: latest => 1.0.0-beta.27
```

### Any additional comments?

`#__PURE__` only means "the code can be safely removed if not used". It does not mean "the code does not rely on the execution order".

Specifically, this line should be checking whether this module relies on the execution order (e.g. reading a global variable), not whether it has a side effect.
https://github.com/rolldown/rolldown/blob/57a4bea49634b8d4f7c868ae2f5e4c9aae389f51/crates/rolldown/src/stages/link_stage/wrapping.rs#L39
This is because even if a module doesn't have a side effect (e.g. writing a global variable, console logging), it can rely on the side effect that is caused by other modules. So those modules should be wrapped to keep the execution order.

That said, in most cases "the code can be safely removed if not used" and  "the code does not rely on the execution order" are equivalent. So this is an edge case. If someone encounters this in real world code, please comment on this issue.

related: #5240, #4684

IWANABETHATGUY:
Closed #5303

### DIFF
--- a/crates/rolldown/src/ast_scanner/impl_visit.rs
+++ b/crates/rolldown/src/ast_scanner/impl_visit.rs
@@ -76,8 +76,12 @@ impl<'me, 'ast: 'me> Visit<'ast> for AstScanner<'me, 'ast> {
       }
 
       self.visit_statement(stmt);
-      if self.current_stmt_info.side_effect.contains(SideEffectDetail::Unknown) {
-        self.result.ecma_view_meta.insert(EcmaViewMeta::HasAnalyzedSideEffect);
+      if self.current_stmt_info.side_effect.intersects(
+        SideEffectDetail::Unknown
+          | SideEffectDetail::GlobalVarAccess
+          | SideEffectDetail::PureAnnotation,
+      ) {
+        self.result.ecma_view_meta.insert(EcmaViewMeta::ExecutionOrderSensitive);
       }
       self.result.stmt_infos.add_stmt_info(std::mem::take(&mut self.current_stmt_info));
     }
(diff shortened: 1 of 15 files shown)


IWANABETHATGUY:
* **#5498** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5498?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5498?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5497** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5497?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5495** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5495?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

IWANABETHATGUY:
I think this pr also fixed https://github.com/rolldown/rolldown/issues/5301, @sapphi-red

hyf0 (approved):

PR #5512, opened 2025-07-30
feat(rolldown): oxc v0.79.0

Boshen:
closes #5506


IWANABETHATGUY (approved):

Boshen:
Decided to ignore rollup error case, I need to think about DCE more to circumvent difference cases without regressing performance.

PR #5524, opened 2025-07-30
feat: use frequent characters first for internal export names

### LINKED ISSUE #5270: [Feature Request]: generate internal export names using frequent characters first

AliceLanniste:
close #5270


IWANABETHATGUY:
I don't think those filenames hash change is expected https://github.com/rolldown/rolldown/pull/5524/files#diff-653f87093199db7b853feff1a79061bbf6e22a75d9ea2c24b1b3c6334e7743d7

AliceLanniste:
I  replaced original list `CHARS` with pre-computed list etnriaoscludfpmhg_vybxSCwTEDOkAjMNPFILRzBVHUWGKqJYXZQ$1024368579

crates/rolldown_utils/src/hash_placeholder.rs
```diff
@@ -68,7 +68,7 @@ pub fn find_hash_placeholders<'a>(
   results
 }
 
-const CHARS: &[u8] = b"0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_$";
+const CHARS: &[u8] = b"etnriaoscludfpmhg_vybxSCwTEDOkAjMNPFILRzBVHUWGKqJYXZQ$1024368579";
 const BASE: u32 = 64;
 
 pub fn to_base64(mut value: u32) -> String {
```
sapphi-red, line 75: I think we need to split the function into two to avoid changing the filename hash.
```suggestion
pub fn to_base64(value: u32) -> String {
  to_base64_generic::<CHARS>(value)
}

pub fn to_base64_based_on_frequency(value: u32) -> String {
  to_base64_generic::<FREQUENT_CHARS>(value)
}

fn to_base64_generic<const CHARS: &[u8]>(mut value: u32) -> String {
```

AliceLanniste: https://github.com/rolldown/rolldown/blob/a3a4228c16289091eccb66451ce47785f7ecadbe/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs#L475-L476
replace `to_base64` with `to_base64_frequent`  or keep it

sapphi-red: we should use `to_base64_based_on_frequency` there

sapphi-red:
I noticed that we cannot use `to_base64` any more because `if export_name.starts_with('1')` won't work. So I created a separate function for it.

sapphi-red:
I tested this PR with https://github.com/oxc-project/playground and got the following results with `find dist -type f -name "*.js" -exec gzip -c {} \; | wc -c`:

- without this PR (total gzip size of JS files): 3,402,400B
- with this PR (total gzip size of JS files): 3,402,309B

So it's slightly smaller 😀

hyf0:
> I noticed that we cannot use `to_base64` any more because `if export_name.starts_with('1')` won't work. So I create a separate function for it.

There's extra logic to jump over invalid ones.

hyf0 (approved):

PR #5590, opened 2025-08-04
perf(hmr): do codegen in parallel

hyf0:


hyf0:
* **#5590** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5590?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5590?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/rolldown/src/hmr/hmr_manager.rs
```diff
@@ -317,59 +322,88 @@ impl HmrManager {
...
+        let Module::Normal(affected_module) = affected_module else {
+          unreachable!("HMR only supports normal module");
+        };
 
-    for affected_module_idx in affected_modules {
-      let affected_module = &self.input.module_db.modules[affected_module_idx];
-      let Module::Normal(affected_module) = affected_module else {
-        unreachable!("HMR only supports normal module");
-      };
+        let ast = self.input.index_ecma_ast[affected_module.idx]
```
sapphi-red: If `affected_module.idx` and `affected_module_idx` gets different (if there's a bug in the future), the assumption for the safety comment will no longer hold. I guess it'd be nice to have a check somewhere that ensures these two values are the same.

sapphi-red (approved):

crates/rolldown/src/hmr/hmr_manager.rs
hyf0: `affected_modules` is a `HashSet`. Even two modules got the same `ModuleIdx`, logic here is still safe.

And there's indeed a way without using `unsafe` to write such code, but it's less performant due to need to go through the whole vec

```rust
let affected_ast_vec =  index_ecma_ast.iter_mut().filter_map(|(idx, ast)| affected_modules.contains(idx)).collect::<Vec<)>>();
```

sapphi-red: > Even two modules got the same `ModuleIdx`, logic here is still safe.

`affected_module_idx` would never be the same value. **But** `affected_module.idx` can be a different value if there's a bug. `IndexModules` type does not check whether the key (`ModuleIdx`) is the same value with the one in the value (`Module.idx`).

hyf0: I get what you mean here.

Issue #5176, opened 2025-07-05
[Bug]: higher memory usage with prebuilt binaries than local built binaries


sapphi-red:
### Reproduction link or steps

1. Clone https://github.com/brielov/vite-memory-example
2. Run `pnpm i`
3. Run `node ./node_modules/vite/bin/vite.mjs build`
4. Add overrides to `package.json` to link local rolldown build
5. Run `pnpm i`
6. Run `node ./node_modules/vite/bin/vite.mjs build`

### What is expected?

The memory usage is same for prebuilt binaries and local built binaries.

### What is actually happening?

The prebuilt binary uses more memory (~252MB) than local built one (~168MB).

### System Info

```Shell
System:
    OS: Linux 6.6 Ubuntu 24.04.2 LTS 24.04.2 LTS (Noble Numbat)
    CPU: (24) x64 AMD Ryzen 9 5900X 12-Core Processor
    Memory: 22.64 GB / 23.47 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.16.0 - ~/.local/share/asdf/data/installs/nodejs/22.16.0/bin/node
    npm: 10.9.2 - ~/.local/share/asdf/data/plugins/nodejs/shims/npm
    pnpm: 10.12.4 - ~/.local/share/asdf/data/installs/nodejs/22.16.0/bin/pnpm
    bun: 1.2.17 - ~/.local/share/asdf/data/shims/bun
```

### Any additional comments?

Originally reported at https://github.com/vitejs/rolldown-vite/discussions/297

The binary difference seems to be coming from napi-rs and probably caused by GCC or clang version difference.

Boshen:
https://github.com/rolldown/rolldown/pull/5167#issuecomment-3036431719

```
$ readelf -p .comment ./rolldown-binding.linux-x64-gnu.node # the build on CI
String dump of section '.comment':
  [     0]  GCC: (crosstool-NG 1.24.0.388_94b98a4) 4.8.5
  [    2d]  rustc version 1.88.0 (6b00bc388 2025-06-23)
  [    59]  Ubuntu clang version 18.1.3 (1ubuntu1)
```
```
$ readelf -p .comment ./packages/rolldown/dist/rolldown-binding.linux-x64-gnu.node # the local build
String dump of section '.comment':
  [     0]  GCC: (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
[...]

PR #5613, opened 2025-08-05
wip: try CC instead of TARGET_CC

sapphi-red:
refs #5194 , #5176


sapphi-red:
It seems this isn't the reason

PR #5641, opened 2025-08-06
fix(hmr): rewrite `this` to `exports` for cjs modules

### LINKED ISSUE #5616: [Bug]: `this` in CJS module should be `module.exports` but is `undefined` in HMR patch file
See #5615

hyf0:
- Fixes https://github.com/rolldown/rolldown/issues/5616
- Closes https://github.com/rolldown/rolldown/pull/5615

### DIFF
--- a/crates/rolldown/src/hmr/impl_traverse_for_hmr_ast_finalizer.rs
+++ b/crates/rolldown/src/hmr/impl_traverse_for_hmr_ast_finalizer.rs
@@ -1,0 +1,39 @@
+use oxc::ast::ast;
+use oxc_traverse::Traverse;
+use rolldown_ecmascript_utils::quote_expr;
+
+use crate::hmr::hmr_ast_finalizer::HmrAstFinalizer;
+
+impl<'ast> Traverse<'ast, ()> for HmrAstFinalizer<'_, 'ast> {
+  fn exit_expression(
+    &mut self,
+    node: &mut oxc::ast::ast::Expression<'ast>,
+    ctx: &mut oxc_traverse::TraverseCtx<'ast, ()>,
+  ) {
+    if ctx.is_current_scope_valid_for_tla() && matches!(node, ast::Expression::ThisExpression(_)) {
+      // Rewrite this to `undefined` or `exports`
+      if self.module.exports_kind.is_commonjs() {
+        // Rewrite this to `exports`
+        *node = quote_expr(self.alloc, "exports");
+      } else {
+        // Rewrite this to `undefined`
+        *node = quote_expr(self.alloc, "void 0");
+      }
+    }
+  }
+}
+
+trait TraverseCtxExt<'ast> {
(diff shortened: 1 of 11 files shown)


hyf0:
* **#5647** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5647?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5646** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5646?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5641** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5641?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5641?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

IWANABETHATGUY (approved):

crates/rolldown/src/hmr/impl_traverse_for_hmr_ast_finalizer.rs
```diff
@@ -0,0 +1,39 @@
...
+  fn exit_expression(
+    &mut self,
+    node: &mut oxc::ast::ast::Expression<'ast>,
+    ctx: &mut oxc_traverse::TraverseCtx<'ast, ()>,
+  ) {
+    if ctx.is_current_scope_valid_for_tla() && matches!(node, ast::Expression::ThisExpression(_)) {
+      // Rewrite this to `undefined` or `exports`
+      if self.module.exports_kind.is_commonjs() {
+        // Rewrite this to `exports`
+        *node = quote_expr(self.alloc, "exports");
```
sapphi-red, line 17: Would this work for cases like below? (not a blocker for this PR as it's an edge case though)
```js
const exports = {}
this.foo = 'foo'
```

sapphi-red (approved):

crates/rolldown/src/hmr/impl_traverse_for_hmr_ast_finalizer.rs
hyf0: What's the expectation here?

I probably thinking about following webpack and rename `exports` to `__rolldown_exports__`. And it will solve this case like

```js
function (__rolldown_exports__, __rolldown_module__) {
const exports = {}
__rolldown_exports__.foo = 'foo'
}
```

Declaring `__rolldown_exports__` manually is UB and rolldown won't handle such case.

sapphi-red: Actually the code I wrote didn't work with `SyntaxError: Identifier 'exports' has already been declared`. A valid example would be:
```js
{
  const exports = {}
  this.foo = 'foo'
}
```

The expectation is to work the same with Node. Your suggestion looks good to me.

PR #5645, opened 2025-08-06
test(hmr): run patch exectution after the normal execution

sapphi-red:
Alternative to #5639. This worked for me.

Probably, `--import` tries to run not just the first macro task but also some macro tasks.

@situ2001 Would you try if this one works for you as well?

close #5639 close #5648

PR #5639, opened 2025-08-06
fix(rolldown_test): refine injection logic

situ2001:
Injection should poll for `__rolldown_runtime__` before acutally running test. Otherwise, an error will occur in my Linux x64 machine.

```shell
failures:

---- fixture_with_config_tests__rolldown__topics__hmr__dynamic_import__config_json stdout ----
Input: /home/situ/Codes/rolldown/crates/rolldown/tests/rolldown/topics/hmr/dynamic_import/_config.json
⬇️⬇️ Failed to execute command  ⬇️⬇️
"node" "--import" "data:text/javascript,globalThis.__testPatches%20%3D%20%5B%22.%2F1754486755140.js%22%5D%3B%0Aimport%20url%20from%20%27node%3Aurl%27%3B%0Aimport%20path%20from%20%27node%3Apath%27%3B%0A%0Aconst%20dir%20%3D%20%27%2Fhome%2Fsitu%2FCodes%2Frolldown%2Fcrates%2Frolldown%2Ftests%2Frolldown%2Ftopics%2Fhmr%2Fdynamic_import%2Fhmr-temp%2Fdist%27%3B%0AsetTimeout%28async%20%28%29%20%3D%3E%20%7B%0A%20%20for%20%28const%20patchChunk%20of%20globalThis.__testPatches%29%20%7B%0A%20%20%20%20const%20file%20%3D%20path.join%28dir%2C%20patchChunk%29%3B%0A%20%20%20%20try%20%7B%0A%20%20%20%20%20%20await%20import%28url.pathToFileURL%28file%29%29%3B%0A%20%20%20%20%7D%20catch%20%28error%29%20%7B%0A%20%20%20%20%20%20console.error%28%27Error%20executing%20a%20patch%3A%27%2C%20error%29%3B%0A%20%20%20%20%20%20process.exitCode%20%3D%201%3B%0A%20%20%20%20%20%20break%3B%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%2C%2010%29%3B%0A%20%20%20%20%20%20" "--import" "/home/situ/Codes/rolldown/crates/rolldown/tests/rolldown/topics/hmr/dynamic_import/hmr-temp/dist/main.js" "--eval" "\"\"" "--input-type=module"
⬆️⬆️ end  ⬆️⬆️

thread 'fixture_with_config_tests__rolldown__topics__hmr__dynamic_import__config_json' panicked at crates/rolldown_testing/src/integration_test.rs:678:7:
[...]

PR #5648, opened 2025-08-07
chore(dev): update node version

situ2001:
Related to #5645, #5639

I found that the issues is solved if we update node version to 22.18.0. It is really weird. 😅

Notes: I edited justfile for repeatedly test for 10 times in order to verify the edit on CI.


sapphi-red:
This one worked for me as well. That said, I'll prefer #5645 as it doesn't rely on the Node version.

PR #5717, opened 2025-08-13
feat(rolldown): oxc v0.82.1

Boshen:
Changelog: https://github.com/oxc-project/oxc/pull/13049

DCE phase now runs in a loop, resulting output that aligns to Rollup's behavior, with a small sacrifice in performance.


crates/rolldown/tests/esbuild/lower/lower_using_hoisting/artifacts.snap
```diff
@@ -236,14 +236,14 @@ export { a1, ac1, c1, x1, y1, z1 };
 
 ```js
 //#region hoist-export-local-indirect.js
-using a = b;
+b;
```
sapphi-red, line 239: This shouldn't be changed. Made an issue: https://github.com/oxc-project/oxc/issues/13050

Boshen: Given that this is a super rare case, let's not block the merge.

sapphi-red: Yeah, I agree to that.

shulaoda (dismissed):

Brooooooklyn:
cursor review

IWANABETHATGUY:
The regression in the scan phase seems not expected, should iterate only once during scan phase.

Boshen:
> The regression in the scan phase seems not expected, should iterate only once during scan phase.

There's no way to remove dead code cleanly without at least running twice, unless we build a sophisticated algorithm for it, for example https://github.com/google/closure-compiler/blob/67caecbf83855cfd9c56facdcd355640084c87de/src/com/google/javascript/jscomp/RemoveUnusedCode.java#L50

PR #2987, opened 2024-11-29
feat: pass the remaining normalized fields from rust to js

shulaoda:
### Description

closes #2906 

I've mostly aligned `NormalizedOutputOptions` with `rollup`. If `NormalizedInputOptions` isn't needed, we can close this issue.

The following fields are unique to us:
https://github.com/rolldown/rolldown/blob/71e6863f00b9d304a568072d16959082e1e6b17d/packages/rolldown/src/options/normalized-output-options.ts#L40-L41

I haven't handled this because converting `HybridRegex` in rust to js regex is challenging:
https://github.com/rolldown/rolldown/blob/71e6863f00b9d304a568072d16959082e1e6b17d/packages/rolldown/src/options/output-options.ts#L58-L67

PR #5732, opened 2025-08-15
feat: support  source maps without original source code

### LINKED ISSUE #5554: [Feature Request]: Generate source maps without original source code

Currently, generated source maps always include the original untransformed source code in their `sourcesContent` field. This is useful for more in depth debugging, but when only correct line/column references in errors are required, it's unnecessary bloat in the final bundle when inlining the source maps and prevents obscuring the code with minification.

Both esbuild and rollup have options to leave out `sourcesContent`. This allows keeping the position references to the original code without having to include it.

A boolean property should be added to the output options object. To keep the API similar, naming it either `sourcemapExcludeSources` like Rollup or `sourcesContent` like in esbuild would be best.

Of note is that those two existing solutions use opposing values. The Rollup flag removes the sources when `true`, while esbuild's optins needs to be set to `false`.

AliceLanniste:
close #5554


I-Enderlord-I:
You didn't add the options to the TS typings in [binding.d.ts](https://github.com/AliceLanniste/rolldown/blob/sourcemapExcludeContent/packages/rolldown/src/binding.d.ts). They should get generated after a build as demonstrated by the test.

I-Enderlord-I:
I wonder why there is no getter in `BindingNormalizedOptions `. There should be, right?

I-Enderlord-I:
And you also still need to add it in the actually exported TS definitions in [options/output-options.ts](https://github.com/AliceLanniste/rolldown/blob/sourcemapExcludeContent/packages/rolldown/src/options/output-options.ts) and [options/normalized-output-options.ts](https://github.com/AliceLanniste/rolldown/blob/sourcemapExcludeContent/packages/rolldown/src/options/normalized-output-options.ts).

AliceLanniste:
@I-Enderlord-I ，hi I don't understand why must add getter in BindingNormalizedOptions and options in [options/output-options.ts](https://github.com/AliceLanniste/rolldown/blob/sourcemapExcludeContent/packages/rolldown/src/options/output-options.ts),can you explain ?

I-Enderlord-I:
I don't fully understand the project structure myself (thus why I didn't make a pull request), but as far as I see, the typescript files in `packages/rolldown` define the JS API. Here the `binding.d.ts` corresponds to the actual objects transfered from the Rust side, while the other files define all kinds of wrappers around that interface.

Specifically, the `options/normalized-output-options.ts` is basically a one-to-one wrapper of `BindingNormalizedOptions` in the `binding.d.ts` and `options/output-options.ts` defines the OutputOptions interface as it is passed to the user.

As for adding the getter to `BindingNormalizedOptions`, I have no idea why it was not automatically generated. To me it seems like it should.

I-Enderlord-I:
I think I get it now... Check out #2987 to see where passing additional properties to the JS/TS API requires changes.

That still leaves receiving more in `options/output-options.ts` of course...
I guess [crates/rolldown_binding/src/options/binding_output_options/mod.rs](https://github.com/AliceLanniste/rolldown/blob/sourcemapExcludeContent/crates/rolldown_binding/src/options/binding_output_options/mod.rs) is the place to add that.

I-Enderlord-I:
Ok I was correct. See cf09f46 for everything a new options requires.

AliceLanniste:
@I-Enderlord-I THANK YOU SO MUCH

I-Enderlord-I:
Maybe adding a test to `packages/rolldown/tests` for this would be nice as well. Just to be sure you got everything and it's actually useable.

sapphi-red:
Closing as this was implemented in #8828

PR #5734, opened 2025-08-15
perf(rolldown_plugin_reporter): gzip size computation

IWANABETHATGUY:
> [!note]
> - bench case: **apps/10000** 20000+ modules.
> - configuration: advanced-chunks + disabled minify

# Before
I noticed that the gzip size computation costs a lot of time in rolldown-vite, in my local computer, it takes 158ms to compute gzip size(total building time takes about 750ms)
image

```bash
rolldown-vite v7.0.12 building for production...
✓ 20014 modules transformed.
vite-dist/index.html                               0.62 kB │ gzip:   0.34 kB
vite-dist/assets/rolldown-runtime-DWGyDpL3.js      1.12 kB │ gzip:   0.55 kB
vite-dist/assets/react-jrohfdTy.js               207.41 kB │ gzip:  54.06 kB
vite-dist/assets/index-BROXgPrc.js             3,901.62 kB │ gzip: 323.08 kB
vite-dist/assets/other-libs-BVAwKXhJ.js        6,680.16 kB │ gzip: 938.17 kB
✓ built in 755ms
```

# After
image

```bash
rolldown-vite v7.0.12 building for production...
✓ 20014 modules transformed.
vite-dist/index.html                               0.62 kB │ gzip:   0.34 kB
vite-dist/assets/rolldown-runtime-DWGyDpL3.js      1.12 kB │ gzip:   0.56 kB
vite-dist/assets/react-jrohfdTy.js               207.41 kB │ gzip:  54.25 kB
vite-dist/assets/index-BUf1zkMf.js             3,901.62 kB │ gzip: 323.61 kB
vite-dist/assets/other-libs-BVAwKXhJ.js        6,680.16 kB │ gzip: 938.11 kB
✓ built in 648ms
```
The size after gzip is almost identical; it should be alright to have a little difference since it's just a gzip size used for reference.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -137,7 +137,7 @@ dashmap = "6.1.0"
 derive_more = { version = "2.0.1", features = ["debug"] }
 dunce = "1.0.5" # Normalize Windows paths to the most compatible format, avoiding UNC where possible
 fast-glob = "1.0.0"
-flate2 = "1.1.1"
+flate2 = { version = "1.1.1", features = ["zlib-rs"] }
 form_urlencoded = "1.2.1"
 futures = "0.3.31"
 glob = "0.3.2"
--- a/crates/rolldown_plugin_reporter/src/lib.rs
+++ b/crates/rolldown_plugin_reporter/src/lib.rs
@@ -12,6 +12,7 @@ use std::{
 };
 
 use cow_utils::CowUtils;
+use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
 use rolldown_plugin::{HookUsage, Plugin, PluginContext};
 use sugar_path::SugarPath;
 
@@ -26,7 +27,6 @@ pub struct ReporterPlugin {
   chunk_count: AtomicU32,
   compressed_count: AtomicU32,
   report_compressed_size: bool,
-  has_compress_chunk: AtomicBool,
   has_rendered_chunk: AtomicBool,
   has_transformed: AtomicBool,
   transformed_count: AtomicU32,
@@ -52,7 +52,6 @@ impl ReporterPlugin {
       chunk_count: AtomicU32::new(0),
       compressed_count: AtomicU32::new(0),
       report_compressed_size,
-      has_compress_chunk: AtomicBool::new(false),
       has_rendered_chunk: AtomicBool::new(false),
       has_transformed: AtomicBool::new(false),
       transformed_count: AtomicU32::new(0),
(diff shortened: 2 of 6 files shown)


IWANABETHATGUY:
* **#5734** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5734?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5734?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/rolldown_plugin_reporter/src/utils.rs
```diff
@@ -29,12 +29,27 @@ pub fn display_size(size: usize) -> String {
...
+  fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+    self.count += buf.len();
+    Ok(buf.len())
+  }
+
+  fn flush(&mut self) -> std::io::Result<()> {
+    Ok(())
   }
-  encoder.finish().ok().map(|compressed| compressed.len())
+}
```
IWANABETHATGUY, line 45: Using a custom `Writer` to avoid allocation.

Cargo.toml
```diff
@@ -137,7 +137,7 @@ dashmap = "6.1.0"
 derive_more = { version = "2.0.1", features = ["debug"] }
 dunce = "1.0.5" # Normalize Windows paths to the most compatible format, avoiding UNC where possible
 fast-glob = "1.0.0"
-flate2 = "1.1.1"
+flate2 = { version = "1.1.1", features = ["zlib-rs"] }
```
IWANABETHATGUY, line 140: Using `zlib-rs` backend for best performance, https://github.com/rust-lang/flate2-rs?tab=readme-ov-file#backends, they claimed that they are the fastest, also see benchmark https://trifectatech.org/blog/zlib-rs-is-faster-than-c/

crates/rolldown_plugin_reporter/src/lib.rs
```diff
@@ -171,42 +170,53 @@ impl Plugin for ReporterPlugin {
...
-            let compressed_count = self.compressed_count.fetch_add(1, Ordering::SeqCst);
-            if self.should_log_info && self.is_tty {
-              utils::write_line(&format!(
-                "computing gzip size ({})...",
-                itoa::Buffer::new().format(compressed_count)
-              ));
-            }
-          })
+
+      if self.report_compressed_size && self.should_log_info {
```
IWANABETHATGUY, line 174: According to the benchmark, zlib-rs could compress about 1Gb content in 1s, [ref](https://trifectatech.org/blog/zlib-rs-is-faster-than-c/#:~:text=On%20x86_64%20linux%2C%20we%20are%20faster%20for%20some%20of%20the%20compression%20levels%20that%20matter%20most%2C%20about%206%25%20at%20the%20default%20level%20of%206%2C%20and%20over%2010%25%20at%20the%20%22best%20compression%22%20level%209.%20But%20we%27re%20still%20slightly%20slower%20for%20most%20of%20the%20other%20levels%20when%20comparing%20to%20zlib%2Dng.), and also we use rayon to compute the chunk gzip size in parallel now. So in most scenarios, the computation phase would be instant, so it would be unnecessary to report the progress anymore.

Brooooooklyn:
cursor review

shulaoda (approved):

Cargo.toml
sapphi-red: FYI if we want an identical value, using `features = ["zlib"]` can achieve that. But I agree that the small difference doesn't matter here.

sapphi-red (approved):
I'm curious to know how much this makes the Rolldown's binary size bigger (not a blocker).

IWANABETHATGUY:
# Main
image
# Current
image
increase by around 71KB

Issue #5671, opened 2025-08-08
Question on default value of `options.context`


situ2001:
As https://rollupjs.org/configuration-options/#context shown that the default value of `options.context` is `"undefined"`, however, in Rolldown, the default value of options.context is currently an empty string (""), and it uses `void 0` actually to replace the top-level context, leading to a discrepancy in default values (`"void 0"` vs.`"undefined"` vs. `""`).

https://github.com/rolldown/rolldown/blob/beec4a4947183b49177e5842e391fa7a9a909d1d/crates/rolldown/src/utils/normalize_options.rs#L291

https://github.com/rolldown/rolldown/blob/beec4a4947183b49177e5842e391fa7a9a909d1d/crates/rolldown_common/src/inner_bundler_options/types/normalized_bundler_options.rs#L177

Should we align Rolldown with Rollup by setting the default value of options.context to `"void 0"` (in Rollup, default value is `"undefined"`) to ensure consistency?

IWANABETHATGUY:
Our code base by default would transform `undefined` into `void 0`, so I prefer use `void 0` as default value.

IWANABETHATGUY:
But I prefer the current implementation, since the final results are identical, https://github.com/rolldown/rolldown/blob/9a9a1d9b19940db80c5ea0b3bb96e9c0dc03d4ef/crates/rolldown/src/module_finalizers/impl_visit_mut.rs?plain=1#L353-L361, if the user doesn't set `context`, which should be most of the scenarios, the current one does not need to alloc anything, if we explicitly set it to `undefined` you need to either alloc another string in areana or compre if the context equal to `undefined`, both of these solution is slow than current one.

IWANABETHATGUY:
If the user wants to get the context via js api, we could just return `undefined` when the context is a empty string.

PR #5745, opened 2025-08-17
feat(rolldown): align default value of `option.context`

situ2001:
Refer to https://github.com/rolldown/rolldown/issues/5671

And here is how Rollup process `context` when it is undefined.

https://github.com/rollup/rollup/blob/e4082a84063490503a6ee3b4ac74b45be745f111/src/utils/options/normalizeInputOptions.ts#L43


IWANABETHATGUY (approved):

hyf0:
Thank you!

crates/rolldown_binding/src/types/binding_normalized_options.rs
```diff
@@ -278,6 +278,11 @@ impl BindingNormalizedOptions {
 
   #[napi(getter)]
   pub fn context(&self) -> String {
+    // https://github.com/rolldown/rolldown/issues/5671
+    if self.inner.context.is_empty() {
+      return "undefined".into();
```
hyf0, line 283: Not the right place to put the default value.

You should put it here. Please following how other options  assign their default value

https://github.com/rolldown/rolldown/blob/c3cac553351e04fa3860484215299f1d95603518/crates/rolldown/src/utils/normalize_options.rs#L291

packages/rolldown/tests/fixtures/context/default-value/_config.ts
```diff
@@ -0,0 +1,23 @@
+import path from 'node:path'
+import { expect } from 'vitest'
+import { defineTest } from 'rolldown-tests'
+
+const entry = path.join(__dirname, './main.js')
+
+export default defineTest({
```
hyf0, line 7: Could you follow how

https://github.com/rolldown/rolldown/tree/main/packages/rolldown/tests/fixtures/verify-options

does test?

And create one for testing the default value behavior

You could put the test at

`<REPO>/packages/rolldown/tests/fixtures/topics/options-default-value`

---

You only need to add test code related to the `context` option. No need to worry about others.

crates/rolldown_binding/src/types/binding_normalized_options.rs
IWANABETHATGUY: https://github.com/rolldown/rolldown/issues/5671#issuecomment-3168273555，I explain why I don't prefer to set a default value for this option.

hyf0: Then maybe just use `void 0` as the default value? It's more intuitive that bundler's behavior should match the options it gets.

PR #5787, opened 2025-08-19
feat: using MagicString

AliceLanniste:
@shulaoda ,hi ,I use MagicString instead of String


shulaoda:
In fact, I am currently performing string replacement directly in `render_asset_url_in_js` without using `MagicString`, and I am not planning to do so for now. As a result, your approach cannot produce a valid sourcemap.

PR #5790, opened 2025-08-19
fix: don't match CRLF for `/./`

### LINKED ISSUE #5784: [Bug]: `transform.filter.code: /./` matches empty files

[REPL](https://repl.rolldown.rs/#eNqNT8FOwzAM/RXLl25SyYZ2y68QDqFxSiGzJzeFSVX+nZSxjsMOs2TZfn7P8psxop1RJaUg32w64Tj0Jo8LymjvbVrs6oLOJ9EMgaKfUobZMcApTf3Ao4WXZYILuAT7I1lookjTXrGsnscoerQ3HkAcUiatEHQSqmZndlBWDcC755BIN9v/IgClPClDU78cJZFJ0m8cyqfDbXPjlWv71/yWV8e1Ymlx4EBn87F6X+eLY6eOH8lKJ7RZJyr16FdVPpu92T+9UfbmcMDyA0n3co4=)

### What is expected?

`/./.test("\n".repeat(10))` returns `false` so I expect the output to be empty.

The filter is matched and the output is not empty.

### System Info

```Shell
REPL 1.0.0-beta.31
```

### Any additional comments?

On the other hand, `/./s.test("\n".repeat(10))` returns `true` so I expect to have the current behavior.

IWANABETHATGUY:
Closed https://github.com/rolldown/rolldown/issues/5784

### DIFF
--- a/crates/rolldown_utils/src/js_regex.rs
+++ b/crates/rolldown_utils/src/js_regex.rs
@@ -21,22 +21,28 @@ impl From<&str> for HybridRegex {
 }
 impl HybridRegex {
   pub fn new(pattern: &str) -> anyhow::Result<Self> {
-    match regex::Regex::new(pattern).map(HybridRegex::Optimize) {
+    let regex_pattern = Self::get_regex_pattern(pattern, "");
+    match regex::Regex::new(&regex_pattern).map(HybridRegex::Optimize) {
       Ok(reg) => Ok(reg),
       Err(_) => regress::Regex::new(pattern).map(HybridRegex::Ecma).map_err(anyhow::Error::from),
     }
   }
 
   pub fn with_flags(pattern: &str, flags: &str) -> anyhow::Result<Self> {
-    let regex_pattern =
-      if flags.is_empty() { pattern } else { &concat_string!("(?", flags, ")", pattern) };
-
-    match regex::Regex::new(regex_pattern).map(HybridRegex::Optimize) {
+    let regex_pattern = Self::get_regex_pattern(pattern, flags);
+    match regex::Regex::new(&regex_pattern).map(HybridRegex::Optimize) {
       Ok(reg) => Ok(reg),
       Err(_) => regress::Regex::with_flags(pattern, flags)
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY:
* **#5790** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5790?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5790?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sapphi-red:
@IWANABETHATGUY The test case you added was working because it was using LF. REPL was probably using CRLF as I was using Windows.

It seems regex crate does not allow to match LF for `.`.
https://docs.rs/regex/latest/regex/#:~:text=s%20%20%20%20%20allow%20.%20to%20match%20%5Cn
This is the reason your test case passed. I removed the test case as it's difficult to push a file with CRLF line breaks.

I added `R` flag, which will also disallow CRLF for `.`.
https://github.com/rust-lang/regex/blob/cf00423af182b5d02695fe0f047626583c86cc35/regex-syntax/src/hir/translate.rs#L894-L919
https://github.com/rust-lang/regex/blob/cf00423af182b5d02695fe0f047626583c86cc35/regex-syntax/src/hir/mod.rs#L1905-L1908
In JS, there are [two more characters that are disallowed to be matched](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll#description) (LS, PS), but I guess those are really an edge case and leave it for now.

Setting the `R` flag also fixes the `m` flag behavior to be more similar to JS's regex. Previously it didn't match CRLF but with the flag, it now matches correctly.
https://github.com/rust-lang/regex/blob/cf00423af182b5d02695fe0f047626583c86cc35/regex-automata/src/util/syntax.rs#L236-L238

sapphi-red (approved):

PR #5792, opened 2025-08-19
fix(rolldown): run DCE on chunk when `minify: dce-only`

Boshen:
fixes https://github.com/oxc-project/oxc/issues/13176

`minify_options` are unpacked and passed to `dce_or_minify` because we can't import `rolldown_common` in `rolldown_ecmascript`, it would introduce cyclic import.

### DIFF
--- a/packages/rolldown/tests/fixtures/tree-shake/on-chunk/_config.ts
+++ b/packages/rolldown/tests/fixtures/tree-shake/on-chunk/_config.ts
@@ -1,0 +1,32 @@
+import type { OutputChunk as RolldownOutputChunk } from 'rolldown'
+import { defineTest } from 'rolldown-tests'
+import { expect } from 'vitest'
+
+export default defineTest({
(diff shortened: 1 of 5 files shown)


Boshen:
There are no existing tests to cover this part of the pipeline, where should I add a test case?

shulaoda (approved):

packages/rolldown/tests/fixtures/tree-shake/on-chunk/_config.ts
```diff
@@ -0,0 +1,32 @@
...
+import { defineTest } from 'rolldown-tests'
+import { expect } from 'vitest'
+
+export default defineTest({
+  config: {
+    treeshake: {
+      moduleSideEffects: false,
+    },
+    output: {
+      minify: 'dce-only'
```
hyf0, line 11: We need a document for this option

Issue #5788, opened 2025-08-19
[Bug]: panics with `export class {}`


sapphi-red:
### Reproduction link or steps

[REPL](https://repl.rolldown.rs/#eNqrVkpTsqpWysxLSa3QyyoGsfOUrBB8HaVkIDe1oiC/qEQhOSexuFihujamKCYPKJOqZFVSVJpaW6ujVAZUZKhnoGegm5RakqhnbKxUCwAfeRwX)

### What is expected?

No panic happens. A normal syntax error is shown.

### What is actually happening?

The following panic happens:
```
 thread 'tokio-runtime-worker' panicked at crates/rolldown/src/ast_scanner/mod.rs:646:
 called `Option::unwrap()` on a `None` value
```

### System Info

```Shell
REPL 1.0.0-beta.33
```

### Any additional comments?

https://github.com/rolldown/rolldown/blob/3537676a38cbc29caa1dc047fed018b3a6fe1d32/crates/rolldown/src/ast_scanner/mod.rs#L646

hyf0:
Should oxc parser reports an error or should rolldown handles it? @Boshen cc

IWANABETHATGUY:
https://playground.oxc.rs/#eNo9jzFuwzAMRa8ScM7QFujiucjaol270DIdCJBJg6TSBIbvXiqGPUkUxPf+XyBBB3SfRf2UCpqdlvWX4QwC3QJauR32YMc7dK6VzlAy+363JDMdw2PqpeyTK7KNohN0Ixaj9QwzqpE2IpYif9/kVfmzuuWBLpWTZwnftj4rxd8bfaES2/YaiCZ/IuJ+GDamo14pkgHZ28vre3TIJgWdhg+KaooNH6Q9TZKBrvQsSIx9oR+pmmjC+bBNmfOYd18SdpVyieht6Ubai0X7DdgChRzNYf0HGuN6sg==
Oxc reported a syntax error actually.

IWANABETHATGUY:
Probably, I know why. This syntax error seems marked as a recoverable error, so maybe we need to handle it on rolldown side

Boshen:
Because we are not aborting here, which we should.

Semantic errors should be hard errors.

https://github.com/rolldown/rolldown/blob/41528dc2de57d8bd280a75fd2ef9f03daa2a5254/crates/rolldown/src/utils/pre_process_ecma_ast.rs#L47-L54

AliceLanniste:
[...]

PR #5796, opened 2025-08-20
fix: throw a semantic error message instead of panic

AliceLanniste:
related #5788


shulaoda:
You should run `just test-update` to update the snapshots. For rust tests, simply run `just test-rust`.

AliceLanniste:
> You should run `just test-update` to update the snapshots. For rust tests, simply run `just test-rust`.

yeah， I am updating  tests

crates/rolldown/src/utils/pre_process_ecma_ast.rs
```diff
@@ -49,7 +49,7 @@ impl PreProcessEcmaAst {
         semantic_ret.errors,
         &source,
         path,
-        &Severity::Warning,
+        &Severity::Error,
```
hyf0: If it gonna be put in `warnings`, the `Severity` should be warning.

hyf0:
Before we push forward this PR, I think we should first decide whether we want to treat all oxc warnings as rolldown's error or selectively. @Boshen @IWANABETHATGUY cc

IWANABETHATGUY:
> Before we push forward this PR, I think we should first decide whether we want to treat all oxc warnings as rolldown's error or selectively. @Boshen @IWANABETHATGUY cc

It needs to be treated as diagnostic with `Severity::Error` if we want to solve the issue https://github.com/rolldown/rolldown/issues/5788, https://github.com/rolldown/rolldown/issues/5788#issuecomment-3199332197

IWANABETHATGUY:
> Before we push forward this PR, I think we should first decide whether we want to treat all OXC warnings as roll-down > errors or selectively. @Boshen @IWANABETHATGUY cc

Note: this pr only treats all semantic_errors as errors.

hyf0:
> > Before we push forward this PR, I think we should first decide whether we want to treat all OXC warnings as roll-down > errors or selectively. @Boshen @IWANABETHATGUY cc
> 
> Note: this pr only treats all semantic_errors as errors.

Got it.

crates/rolldown/src/utils/pre_process_ecma_ast.rs
hyf0: You should return `BuildDiagnostic::from_oxc_diagnostics(..)` as `Err(..)` instead of add them to warnings.

AliceLanniste: got it

IWANABETHATGUY (approved):

PR #5806, opened 2025-08-20
test(rolldown): add basic test for asset plugin

situ2001:
Add basic test for rolldown_plugin_asset, also for the fix #5749


shulaoda:
We don't need to add tests for asset-related plugins at the moment because they are not fully completed yet. The asset, css, and html plugins need to work together to function properly on the Vite side. Therefore, once all of them have been fully ported, I will verify their functionality through rolldown-vite side test cases.

Issue #5839, opened 2025-08-22
Using transform settings from tsconfig loaded by `resolve.tsconfigFilename` is quite confusing


sapphi-red:
I was expecting that `resolve.tsconfigFilename` only affects resolution because the option is inside `resolve`. To make it worse, even when I set the related transform options, those were overridden without any warning, making the investigation difficult...
https://github.com/rolldown/rolldown/blob/0078e8415034198cf466b7647f81978f812ab5d9/crates/rolldown/src/bundler_builder.rs#L127-L142

My proposal is to:

- rename `resolve.tsconfigFilename` to a top-level `tsconfig` option
- prefer the value set in the transform / resolve options if both values are set, also emit a warning
- (not directly related) port the tsconfig & transform option merging code from https://github.com/rolldown/rolldown/blob/0078e8415034198cf466b7647f81978f812ab5d9/crates/rolldown_plugin_transform/src/utils.rs#L142-L244 to https://github.com/rolldown/rolldown/blob/0078e8415034198cf466b7647f81978f812ab5d9/crates/rolldown/src/bundler_builder.rs, which sets more options

hyf0:
Having a top-level `tsConfig` is indeed more proper.

shulaoda:
When the user sets `resolve.tsconfigFilename`, we will resolve the `tsconfig` through it and then merge it with the `transform` options, giving priority to the `transform` configuration. If any conflicts arise, a warning will be displayed.

PR #5842, opened 2025-08-22
feat!: support top-level `tsconfig`

shulaoda:
Related to #5839

### DIFF
--- a/crates/rolldown_resolver/src/resolver.rs
+++ b/crates/rolldown_resolver/src/resolver.rs
@@ -14,7 +14,7 @@ use sugar_path::SugarPath;
 
 use oxc_resolver::{
   EnforceExtension, ModuleType, PackageJson as OxcPackageJson, PackageType, Resolution,
-  ResolveError, ResolveOptions as OxcResolverOptions, ResolverGeneric, TsConfig, TsconfigOptions,
+  ResolveError, ResolveOptions as OxcResolverOptions, ResolverGeneric, TsConfig,
 };
 
 #[derive(Debug)]
@@ -42,7 +42,13 @@ impl<F: FileSystem> Resolver<F> {
 
 impl<F: FileSystem> Resolver<F> {
   #[allow(clippy::too_many_lines)]
-  pub fn new(raw_resolve: ResolveOptions, platform: Platform, cwd: PathBuf, fs: F) -> Self {
+  pub fn new(
+    fs: F,
+    cwd: PathBuf,
+    platform: Platform,
+    tsconfig: Option<PathBuf>,
+    raw_resolve: ResolveOptions,
+  ) -> Self {
     let mut default_conditions = vec!["default".to_string()];
     let mut import_conditions = vec!["import".to_string()];
     let mut require_conditions = vec!["require".to_string()];
@@ -86,12 +92,9 @@ impl<F: FileSystem> Resolver<F> {
 
     let resolve_options_with_default_conditions = OxcResolverOptions {
       cwd: Some(cwd.clone()),
-      tsconfig: raw_resolve.tsconfig_filename.map(|p| {
-        let path = PathBuf::from(&p);
-        TsconfigOptions {
-          config_file: if path.is_relative() { cwd.join(path) } else { path },
-          references: oxc_resolver::TsconfigReferences::Disabled,
-        }
+      tsconfig: tsconfig.map(|tsconfig| oxc_resolver::TsconfigOptions {
+        config_file: tsconfig,
+        references: oxc_resolver::TsconfigReferences::Disabled,
       }),
       alias: raw_resolve
         .alias
--- a/packages/rolldown/src/options/input-options.ts
+++ b/packages/rolldown/src/options/input-options.ts
@@ -139,7 +139,6 @@ export interface InputOptions {
     mainFiles?: string[];
     modules?: string[];
     symlinks?: boolean;
-    tsconfigFilename?: string;
   };
   cwd?: string;
   /**
@@ -376,6 +375,17 @@ export interface InputOptions {
     | 'exports-only';
   optimization?: OptimizationOptions;
   context?: string;
+  /**
+   * Allows you to specify where to find the TypeScript configuration file.
+   *
+   * You may provide:
+   * - a relative path to the configuration file. It will be resolved relative to cwd.
+   * - an absolute path to the configuration file.
+   *
+   * When a tsconfig path is specified, the module resolver will respect `compilerOptions.paths` from the specified `tsconfig.json`,
+   * and the tsconfig options will be merged with the top-level `transform` options, with the `transform` options taking precedence.
+   */
+  tsconfig?: string;
 }
 
 interface OverwriteInputOptionsForCli {
--- a/crates/rolldown/examples/build_bench_rome_ts.rs
+++ b/crates/rolldown/examples/build_bench_rome_ts.rs
@@ -1,13 +1,12 @@
-use rolldown::{Bundler, BundlerOptions, InputItem, ResolveOptions, SourceMapType};
-use rolldown_workspace as workspace;
+use rolldown::{Bundler, BundlerOptions, InputItem, SourceMapType};
 
 // cargo run --example build_bench_rome_ts
 
 #[tokio::main]
 async fn main() {
   // Make sure that you have already run `just setup-bench`
-  let root = workspace::root_dir();
-  let project_root = workspace::crate_dir("rolldown");
+  let root = rolldown_workspace::root_dir();
+  let project_root = rolldown_workspace::crate_dir("rolldown");
   let mut bundler = Bundler::new(BundlerOptions {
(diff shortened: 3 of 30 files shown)


shulaoda:
* **#5844** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5844?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5842** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5842?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5842?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sapphi-red:
- What is the motivation of this change? Is it to allow specifying `oxc_resolver::TsconfigReferences` in the option? Why is it better to allow specifying that in the option, over just using the values written in `tsconfig.json`?
- Wouldn't this require loading the tsconfig options for each file like the current transform plugin does? Do you plan to fix that in a separate PR?
- Do you plan to rename the option to the top-level `tsconfig` option in a separate PR?

shulaoda:
> * What is the motivation of this change? Is it to allow specifying `oxc_resolver::TsconfigReferences` in the option? Why is it better to allow specifying that in the option, over just using the values written in `tsconfig.json`?

We parse the `tsconfig` through `oxc-resolver`, so we need to support the full `resolve.tsconfig` configuration option. See https://github.com/rolldown/rolldown/blob/2af8a415/crates/rolldown_resolver/src/resolver.rs#L265-L267

shulaoda:
> * Wouldn't this require loading the tsconfig options for each file like the current transform plugin does? Do you plan to fix that in a separate PR?

Unlike the `transform` plugin, Rolldown only respects the user-configured `resolve.tsconfig` and will load it only once by default.

shulaoda:
> * Do you plan to rename the option to the top-level `tsconfig` option in a separate PR?

We don't plan to do that for now because we load it using `oxc-resolver`, and it should be configured through `resolve.tsconfig`.

sapphi-red:
> > * What is the motivation of this change? Is it to allow specifying `oxc_resolver::TsconfigReferences` in the option? Why is it better to allow specifying that in the option, over just using the values written in `tsconfig.json`?
> 
> We parse the `tsconfig` through `oxc-resolver`, so we need to support the full `resolve.tsconfig` configuration option. See https://github.com/rolldown/rolldown/blob/2af8a415/crates/rolldown_resolver/src/resolver.rs#L265-L267

Why do we need to? We can also change the behavior of oxc-resolver if there's a better approach. I want to know the fundamental motivation.

sapphi-red:
> > * Wouldn't this require loading the tsconfig options for each file like the current transform plugin does? Do you plan to fix that in a separate PR?
> 
> Unlike the `transform` plugin, Rolldown only respects the user-configured `resolve.tsconfig` and will load it only once by default.

Yes, that's the current behavior. This PR adds support for loading tsconfig references. That requires applying a different configuration for each module. For example, `paths` can have different values for each file like:

- `tsconfig.json`: has `"references": ["./foo/tsconfig.json", "./bar/tsconfig.json"]`
- `foo/tsconfig.json`: `"paths": { "@/": "./a" }`
- `bar/tsconfig.json`: `"paths": { "@/": "./b" }`

Files under `foo` would use `foo/tsconfig.json` and files under `bar` would use `bar/tsconfig.json`. The same should apply to the transform options as well. But unless we make the builtin transform to work like the transform plugin, the behavior will be inconsistent and would be confusing.

shulaoda:
> > We parse the `tsconfig` through `oxc-resolver`, so we need to support the full `resolve.tsconfig` configuration option. See https://github.com/rolldown/rolldown/blob/2af8a415/crates/rolldown_resolver/src/resolver.rs#L265-L267
> 
> Why do we need to? We can also change the behavior of oxc-resolver if there's a better approach. I want to know the fundamental motivation.

Alright, I made a mistake. I checked the `resolve_tsconfig` code, and it defaults to `TsconfigReferences::Auto`. So this setting only affects the resolver’s behavior. Previously, we defaulted the resolver’s `references` to `TsconfigReferences::Disabled` — should we change that?

shulaoda:
> > Unlike the `transform` plugin, Rolldown only respects the user-configured `resolve.tsconfig` and will load it only once by default.
> 
> Yes, that's the current behavior. This PR adds support for loading tsconfig references. That requires applying a different configuration for each module. For example, `paths` can have different values for each file like:
> 
> * `tsconfig.json`: has `"references": ["./foo/tsconfig.json", "./bar/tsconfig.json"]`
> * `foo/tsconfig.json`: `"paths": { "@/": "./a" }`
> * `bar/tsconfig.json`: `"paths": { "@/": "./b" }`
> 
> Files under `foo` would use `foo/tsconfig.json` and files under `bar` would use `bar/tsconfig.json`. The same should apply to the transform options as well. But unless we make the builtin transform to work like the transform plugin, the behavior will be inconsistent and would be confusing.

Let me take a look at how `esbuild` handles this.

sapphi-red:
> > > We parse the `tsconfig` through `oxc-resolver`, so we need to support the full `resolve.tsconfig` configuration option. See https://github.com/rolldown/rolldown/blob/2af8a415/crates/rolldown_resolver/src/resolver.rs#L265-L267
> > 
> > 
> > Why do we need to? We can also change the behavior of oxc-resolver if there's a better approach. I want to know the fundamental motivation.
> 
> Alright, I made a mistake. I checked the `resolve_tsconfig` code, and it defaults to `TsconfigReferences::Auto`. So this setting only affects the resolver’s behavior. Previously, we defaulted the resolver’s `references` to `TsconfigReferences::Disabled` — should we change that?

I think we can leave it as `TsconfigReferences::Disabled` for now. It should have a consistent behavior with the transform option (https://github.com/rolldown/rolldown/pull/5842#issuecomment-3213375760) and would require more changes.

sapphi-red:
These are the related code in esbuild (esbuild uses the nearest tsconfig and does not read references).
https://github.com/evanw/esbuild/blob/195e05c16f03a341390feef38b8ebf17d3075e14/internal/resolver/resolver.go#L922-L942
https://github.com/evanw/esbuild/blob/195e05c16f03a341390feef38b8ebf17d3075e14/internal/resolver/resolver.go#L1630-L1637

packages/rolldown/src/options/input-options.ts
```diff
@@ -376,6 +375,17 @@ export interface InputOptions {
...
   context?: string;
+  /**
+   * Allows you to specify where to find the TypeScript configuration file.
+   *
+   * You may provide:
+   * - a relative path to the configuration file. It will be resolved relative to cwd.
+   * - an absolute path to the configuration file.
+   *
+   * When a tsconfig path is specified, the module resolver will respect `compilerOptions.paths` from the specified `tsconfig.json`,
+   * and the tsconfig options will be merged with the top-level `transform` options, with the `transform` options taking precedence.
```
shulaoda, line 386: I will fix the current `tsconfig` merge problems in the next PR.

crates/rolldown_resolver/src/resolver.rs
```diff
@@ -86,12 +92,9 @@ impl<F: FileSystem> Resolver<F> {
...
-      tsconfig: raw_resolve.tsconfig_filename.map(|p| {
-        let path = PathBuf::from(&p);
-        TsconfigOptions {
-          config_file: if path.is_relative() { cwd.join(path) } else { path },
-          references: oxc_resolver::TsconfigReferences::Disabled,
-        }
+      tsconfig: tsconfig.map(|tsconfig| oxc_resolver::TsconfigOptions {
+        config_file: tsconfig,
+        references: oxc_resolver::TsconfigReferences::Disabled,
       }),
```
sapphi-red, line 98: IIRC `env::current_dir()` doesn't work properly in wasm, so I guess we should keep resolving with `cwd` manually here.

shulaoda: What do you mean by `"keep resolving with cwd manually"`? Do you mean below?
```js
let tsconfig = raw_options.tsconfig.as_ref().map(|tsconfig| {
  raw_options
    .cwd
    .as_ref()
    .map(|cwd| cwd.join(tsconfig))
    .unwrap_or_else(|| PathBuf::from(tsconfig))
});
```

sapphi-red: Ah, my bad, I overlooked this line.
https://github.com/rolldown/rolldown/blob/d30ccc34faffd70847c66b399de4c92788c5d474/crates/rolldown/src/utils/normalize_options.rs#L203

sapphi-red (approved):

hyf0 (approved):

packages/rolldown/src/options/input-options.ts
shulaoda: See https://github.com/rolldown/rolldown/pull/5882

PR #5869, opened 2025-08-24
fix: don't try to do import(input) -> import(nonexistent output)

KTibow:
required to make a large rolldown vite build (i'm building the home assistant frontend with vite) not panic

gpt-5 wrote the patch and this explanation:
- We hit a panic during finalization when rewriting import() because the importee module had no chunk mapping (it wasn’t included), while the containing export statement was preserved.
- This happens when export shape is preserved (ESM) but the dynamic import’s target isn’t “live,” so no chunk is emitted.
- The fix is to guard import() rewrites: only rewrite when a chunk mapping exists. Otherwise, keep the original specifier unchanged. This prevents crashes and preserves correct behavior.

sorry for the bad diffing and lack of minimal repro / test cases, i don't think i can do anything about them


KTibow:
this is clearly the wrong fix, i'll file an issue soon

TheAlexLichter:
Issue to track is #5871

PR #5873, opened 2025-08-24
fix: Runtime error occurs depending on the lazy import order.

### LINKED ISSUE #5870: [Bug]: Runtime error occurs depending on the lazy import order.

Errors occurred in the Windows and Ubuntu builds.
(Stackblitz uses the WebAssembly build, so the error does not occur?)
https://stackblitz.com/edit/vitejs-vite-nfwwwqq7?file=package.json&terminal=dev

### What is expected?

Execute `npm run build`.
Generate `init_~` method in `dist/assets.index.js` 65 step. (On StackBlitz, it is generated.)

It is not generated `init_~` method in the Windows build, and a runtime error occurred in the browser console.

### System Info

```Shell
System:
    OS: Windows 11 10.0.26100
    CPU: (24) x64 AMD Ryzen 9 9900X 12-Core Processor
    Memory: 34.74 GB / 47.10 GB
  Binaries:
    Node: 22.17.1 - C:\Program Files\nodejs\node.EXE
    npm: 11.5.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (139.0.3405.111)
```

### Any additional comments?

Perhaps the init_~ methods are not being generated when using re-exports?

IWANABETHATGUY:
Closed #5870
### TL;DR

Fix a bug where re-exported modules weren't being properly initialized, causing undefined values when importing from re-export modules.

### What changed?

- Modified the `reference_needed_symbols.rs` file to use `std::ptr::write_volatile` in three locations to prevent Rust's dead store optimization from affecting side effect analysis
- Added a test case that reproduces issue #5870, demonstrating the problem with re-exported modules and CommonJS interoperability

### How to test?

1. Run the new test case in `crates/rolldown/tests/rolldown/issues/5870/`
2. Verify that the constant from the re-exported module is correctly initialized and accessible
3. Check that the test passes with the assertion `assert.strictEqual(constant, 1)`

### Why make this change?

This fixes issue #5870 where Rust's compiler was incorrectly optimizing away important side effect flags during the linking stage. When a module re-exports symbols from another module using `export *`, the importer needs to be marked as having side effects to ensure proper initialization order. The Rust compiler was optimizing away these side effect flags, causing modules to be initialized in the wrong order, resulting in undefined values when importing from re-export modules.

The fix uses `std::ptr::write_volatile` to prevent the compiler from optimizing away these critical side effect flags, ensuring proper module initialization order.


IWANABETHATGUY:
* **#5873** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5873?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5873?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

shulaoda (approved):

PR #5897, opened 2025-08-25
fix: 5802

IWANABETHATGUY:
### TL;DR

Implemented a singleton pattern for `BuiltinPlugin` class to ensure consistent instances across different module systems.

### What changed?

- Modified the `BuiltinPlugin` class to use a singleton pattern with a private constructor
- Added a static `getInstance()` method that replaces direct instantiation
- Introduced a global symbol `Symbol.for('BuiltinPlugin')` to store the class reference
- Updated all plugin factory functions to use `BuiltinPlugin.getInstance()` instead of `new BuiltinPlugin()`

### How to test?

1. Verify that all built-in plugins work correctly in both ESM and CommonJS environments
2. Test with mixed module systems (e.g., `.mjs` and `.cjs` files in the same project)
3. Ensure plugins maintain consistent behavior across different import methods

### Why make this change?

This change addresses potential issues when using different module systems in the same process. Since the CLI uses `.mjs` but configurations might use `.cjs` (via `require` or TypeScript CTS files), we need to ensure the same `BuiltinPlugin` class is used throughout the application. The singleton pattern guarantees consistent plugin instances regardless of how they're imported or which module system is used.

### DIFF
--- a/packages/rolldown/src/builtin-plugin/alias-plugin.ts
+++ b/packages/rolldown/src/builtin-plugin/alias-plugin.ts
@@ -11,5 +11,5 @@ type AliasPluginConfig = {
 };
 
 export function aliasPlugin(config: AliasPluginConfig): BuiltinPlugin {
-  return new BuiltinPlugin('builtin:alias', config);
+  return BuiltinPlugin.getInstance('builtin:alias', config);
 }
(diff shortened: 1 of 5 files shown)


IWANABETHATGUY:
* **#5897** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5897?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5897?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #5912, opened 2025-08-26
feat: implement inlineConst.pass

IWANABETHATGUY:


IWANABETHATGUY:
* **#5912** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5912?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5912?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5911** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5911?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

crates/rolldown/src/stages/link_stage/cross_module_inline_const.rs
```diff
@@ -0,0 +1,161 @@
...
+    // scan phase. This would already cover most of the cases, and the overhead is minimal.
+    // - if `inline_const.pass` is greater than 1, and there is no cycle in module graph,
+    // we could just revisit the ast of module in topological order only once.
+    // - TODO:
+    //  if there is cycle in module graph, and the `inline_const.pass` is greater than `1`, we
+    //  should revisit the ast of the module for `inline_const.pass - 1` time.
+    //  potential optimization:
+    //  - if in one pass there is no new constant export found, we can stop the pass early.
+    //  - if all dependencies of a module has no constant export, we don't need to visit ast at all.
+    // The extra passes only when user enable `inline_const` and set `pass` greater than 1.
```
sapphi-red:
```suggestion
    // The extra passes only run when user enable `inline_const` and set `pass` greater than 1.
```

hyf0 (approved):

crates/rolldown/src/stages/link_stage/cross_module_inline_const.rs
```diff
@@ -0,0 +1,161 @@
...
+    if inline_const_pass < 2 {
+      return;
+    }
+    // Explain `inline_const.pass`:
+    // - if `inline_const.pass` is 1, we don't need the extra visit pass, since we already do it in
+    // scan phase. This would already cover most of the cases, and the overhead is minimal.
+    // - if `inline_const.pass` is greater than 1, and there is no cycle in module graph,
+    // we could just revisit the ast of module in topological order only once.
+    // - TODO:
+    //  if there is cycle in module graph, and the `inline_const.pass` is greater than `1`, we
```
sapphi-red, line 42: Just in case, the cycle may disappear in a later pass if the import in the cycle was importing a constant.

crates/rolldown/src/ast_scanner/const_eval.rs
```diff
@@ -32,6 +35,10 @@ impl<'ast> GlobalContext<'ast> for ConstEvalCtx<'_, 'ast> {
     &self,
     reference_id: oxc::semantic::ReferenceId,
   ) -> Option<oxc_ecmascript::constant_evaluation::ConstantValue<'ast>> {
+    // If there is an override function, return its result whatever.
+    if let Some(res) = self.overrode_get_constant_value_from_reference_id.map(|f| f(reference_id)) {
+      return res;
+    }
```
sapphi-red: Is this meant to be like below? or is the comment wrong?
```suggestion
    // If there is an override function, return its result whatever.
    if let Some(f) = self.overrode_get_constant_value_from_reference_id {
      return f(reference_id);
    }
```

crates/rolldown/src/stages/link_stage/cross_module_inline_const.rs
```diff
@@ -0,0 +1,161 @@
...
+    };
+
+    let Declaration::VariableDeclaration(var_decl) = decl else {
+      return;
+    };
+
+    var_decl.declarations.iter().for_each(|declarator| {
+      if let BindingPatternKind::BindingIdentifier(ref binding) = declarator.id.kind
+        && let Some(value) =
+          declarator.init.as_ref().and_then(|expr| try_extract_const_literal(&self.eval_ctx, expr))
```
sapphi-red, line 133: Do we need to check whether this variable is written in a different place than the initializer?

crates/rolldown/src/stages/link_stage/cross_module_inline_const.rs
sapphi-red, line 42: On second thought, won't the cycle cause TDZ (let, const) or `undefined` value (var)?

crates/rolldown/src/ast_scanner/const_eval.rs
sapphi-red: Maybe using a separate struct makes the code simple?

crates/rolldown/src/stages/link_stage/cross_module_inline_const.rs
IWANABETHATGUY: Maybe, but inline itself will not introduce extra cycle.

IWANABETHATGUY: You are right, but then we need to regenerate the module graph after each iteration. I think the scenario you said should be pretty rare, so the ROI seems pretty low.

crates/rolldown/src/ast_scanner/const_eval.rs
IWANABETHATGUY: You suggesion seems simpler.

PR #5920, opened 2025-08-26
fix: `replace_plugin` does not work as expected with .ts config

### LINKED ISSUE #5802: Replace plugin: Does not work as expected with `.ts` config
Do I use the experimental plugin wrong or is it not working for the case below (compared to the rollup one)

[Example / Demo](https://github.com/TheAlexLichter/rolldown-reproductions/tree/rolldown-replace-plugin-broken)

Config:

```ts
import type { RolldownOptions } from 'rolldown'
import rollupReplacePlugin from '@rollup/plugin-replace'
import { replacePlugin } from 'rolldown/experimental'
const config: RolldownOptions = {
  input: './src/index.ts',
  plugins: [
    replacePlugin({
      __rolldown: '0'
    }),
    rollupReplacePlugin({
      __rollup: '1',
      preventAssignment: true
    })
  ]
}

export default config
```

```ts
// index.ts

// @ts-ignore
console.log(__rollup, __rolldown)

```

Expected output:

`console.log(1, 0);`

Actual:

`console.log(1, __rolldown);`

IWANABETHATGUY:
Closed #5802

### DIFF
--- a/packages/rolldown/src/builtin-plugin/alias-plugin.ts
+++ b/packages/rolldown/src/builtin-plugin/alias-plugin.ts
@@ -1,4 +1,4 @@
-import { BuiltinPlugin } from './utils';
+import { BuiltinPlugin, createBuiltinPlugin } from './utils';
 
 type AliasPluginAlias = {
   find: string | RegExp;
@@ -11,5 +11,5 @@ type AliasPluginConfig = {
 };
 
 export function aliasPlugin(config: AliasPluginConfig): BuiltinPlugin {
-  return new BuiltinPlugin('builtin:alias', config);
+  return createBuiltinPlugin('builtin:alias', config);
 }
--- a/packages/rolldown/src/builtin-plugin/constructors.ts
+++ b/packages/rolldown/src/builtin-plugin/constructors.ts
@@ -15,12 +15,16 @@ import type {
 } from '../binding';
 import type { StringOrRegExp } from '../types/utils';
 import { normalizedStringOrRegex } from '../utils/normalize-string-or-regex';
-import { BuiltinPlugin, makeBuiltinPluginCallable } from './utils';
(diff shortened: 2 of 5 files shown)


IWANABETHATGUY:
* **#5920** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5920?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5920?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

shulaoda (approved):

shulaoda (commented):
I think we should add a related test

hyf0 (approved):

IWANABETHATGUY:
> I think we should add a related test

We can't add test since it may populate the `globalThis`

sapphi-red:
It seems this caused the `packages/vite-tests` to fail if it's run without `DISABLE_GLOBAL_CLASS_CACHE=1`. This would probably cause many downstream projects to fail without the env var.
Instead of adding this workaround, can we make rolldown ESM-only? I suppose that would also fix this.

IWANABETHATGUY:
> It seems this caused the `packages/vite-tests` to fail if it's run without `DISABLE_GLOBAL_CLASS_CACHE=1`. This would probably cause many downstream projects to fail without the env var. Instead of adding this workaround, can we make rolldown ESM-only? I suppose that would also fix this.

Let's discuss it next week.

IWANABETHATGUY:
Currently the **Dual Package Hazard** make our `instanceof BuiltinPlugin` detection invalid, so I am going to adding a special `string_field` to `BuiltinPlugin`

Issue #5923, opened 2025-08-27
[Bug]: Useless `__export` helper usage


sxzz:
### Reproduction link or steps

https://repl.rolldown.rs/#eNptjd0KgzAMhV8l5EYY0t33VeYuik2Ho0ukxk2Qvvsq8wdhd/mSnO/MGNDO2LGnyeiwzIz24BrbgjT1khQiKQSRhle+QEjygspcPfXlu2q4BAitppFyjb/t5lzpZNwMNCkldnE3BBeHRZEkRi8fNq1w6B7muev+XE5qT8GNUWGGTW7hdhTdIZeuXCreJROd0qCYvyVNWkQ=

### What is expected?

No `index_exports`

### What is actually happening?

`index_exports` exists but is never used

### System Info

```Shell
REPL
```

IWANABETHATGUY:
This is a little hard to resolve, a workaround would be to let the user enable `minify: 'dce-only'` 

https://repl.rolldown.rs/#eNptzk0OgyAQBeCrTNiYNNbuuUrtwsjQ2OCMwaHVEO5eTP2JSXc8YL55UVmlo+rI4FTJuJxJ6SOXqs0Rp4G9gEMBy1zTmi9gPfdQVDeDQ/5d1JQHUGnxAVOpfrebuaaTuAk4CXpq3C7Yxo0L4dk5wx+qWibbPavXzv15OdEGbROcQIQN13A/Fj1K4CBDEA2xJoC+o87OGgrT4pXJzblKgpT7pFzjnV3XCI6i0hfEjWa2

sxzz:
This will have a minor impact on `rolldown-plugin-dts`, which cannot be resolved by a minifier.

sxzz:
Also, the unused variable and function call still remain.

```js
var index_exports = {};
__export(index_exports, { foo: () => foo });
```

PointSingularity:
Is this related to #5922?

IWANABETHATGUY:
> Is this related to [#5922](https://github.com/rolldown/rolldown/issues/5922)?

They are not the same issue

ivodolenc:
Hi, I noticed the PR aiming to fix this is still in draft. Could you please share whether it's part of the current fix pipeline, and if there's a chance it might be resolved in the next few days? I'd really appreciate any info.

[...]

PR #5926, opened 2025-08-27
feat: mark `__export` runtime helper as pure

IWANABETHATGUY:
`__export`​ just like `__esm`​ it only take effects when the export object are used, so mark it as `pure`​ could reduce the output size in some scenario see https://github.com/rolldown/rolldown/issues/5923 as an example

### DIFF
--- a/crates/rolldown/tests/esbuild/dce/package_json_side_effects_array_keep_main_implicit_main/artifacts.snap
+++ b/crates/rolldown/tests/esbuild/dce/package_json_side_effects_array_keep_main_implicit_main/artifacts.snap
@@ -9,7 +9,7 @@ source: crates/rolldown_testing/src/integration_test.rs
 // HIDDEN [rolldown:runtime]
 //#region node_modules/demo-pkg/index-module.js
 var index_module_exports = {};
-__export(index_module_exports, { foo: () => foo });
+/* @__PURE__ */ __export(index_module_exports, { foo: () => foo });
 var foo;
 var init_index_module = __esm({ "node_modules/demo-pkg/index-module.js": (() => {
 	foo = 123;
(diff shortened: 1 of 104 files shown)


IWANABETHATGUY:
* **#5926** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5926?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5926?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

shulaoda (approved):

crates/rolldown/tests/esbuild/dce/package_json_side_effects_array_keep_main_implicit_main/artifacts.snap
```diff
@@ -9,7 +9,7 @@ source: crates/rolldown_testing/src/integration_test.rs
 // HIDDEN [rolldown:runtime]
 //#region node_modules/demo-pkg/index-module.js
 var index_module_exports = {};
-__export(index_module_exports, { foo: () => foo });
+/* @__PURE__ */ __export(index_module_exports, { foo: () => foo });
```
sapphi-red, line 12: I don't think this is correct. This means this expression can be removed whenever because the return value is not used.
We have to output something like:
```js
var index_module_exports = /* @__PURE__ */ __export({}, { foo: () => foo });
```
to mark the call expression as pure.

I haven't checked but I assume this is the reason why Vite's tests failed after this PR is merged.
https://github.com/rolldown/rolldown/actions/runs/17257477246/job/48972158671

shulaoda: @IWANABETHATGUY cc

PR #5927, opened 2025-08-27
refactor(dev): replace `BuildStatus` with `BuildStateMachine`

hyf0:


hyf0:
* **#5927** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5927?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5927?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/rolldown/src/dev/bundling_task.rs
```diff
@@ -47,7 +46,7 @@ impl BundlingTask {
       Ok(()) => {}
       Err(_) => {
         let mut build_status = self.dev_data.status.lock().await;
-        build_status.is_building = false;
+        build_status.try_to_idle().expect("FIXME: Should not unwrap here");
```
sapphi-red, line 49: This part is also called if `build_status.try_to_idle()?` in `build_inner` returns an error.

crates/rolldown/src/dev/dev_engine.rs
```diff
@@ -62,7 +62,7 @@ impl<W: Watcher + Send + 'static> DevEngine<W> {
       return;
     }
 
-    if let Some(build_process_future) = self.build_driver.schedule_build(vec![]).await {
+    if let Ok(Some(build_process_future)) = self.build_driver.schedule_build(vec![]).await {
       build_process_future.await;
     } else {
```
sapphi-red, line 69: I guess we should handle error here as well. (it shouldn't happen though)

hyf0: I'm still thinking about how to handle error. It's a watch build, so errors shouldn't break the control flow. I guess the final form is using callbacks like `on_error`.

sapphi-red: I think the callback approach would work. Would you add a comment or call `.expect` or something so that we don't forget to handle the error here?

crates/rolldown/src/dev/bundling_task.rs
hyf0: good catch. Something's wrong here.

crates/rolldown/src/dev/dev_engine.rs
hyf0: Sure.

crates/rolldown/src/dev/bundling_task.rs
hyf0: This one is to make sure if errors happen, the build status could be reset to idel.

sapphi-red (approved):

Issue #4973, opened 2025-06-16
[Tracking] native compatibility alignment for `rolldown-vite` plugins


shulaoda:
### Ported Plugins

We mark a plugin as **aligned** when its behavior generally matches the original Vite plugin. However, some differences may still exist but considered non-blocking:

* Certain edge cases may not yet be fully handled
* AST polyfill logic caused by regex-based matching will not be aligned (see [vitejs/vite#14006](<https://github.com/vitejs/vite/pull/14006>))
* The `info` and `warn` messages may be inconsistent, since `this.info` / `this.warn` are not yet supported in plugin context

.

- [ ] aliasPlugin
  - [ ] ~~Currently, ~~`~~\~\~customResolver\~\~~~` is not supported due to potential performance concerns.   -> The main difficulty is wrapping the Node-side context for standalone plugin hooks, since Rolldown currently wraps all plugins through the JS-side context.
  - [ ] Support `'vite:alias': { noResolved: true }` meta data
- [X] buildImportAnalysisPlugin
  - [X] Port `renderChunk` hook logic
  - [X] Port `generateBundle` hook logic (related to asset and css plugins)
- [X] dataUriPlugin (builtin)
- [X] dynamicImportVarsPlugin
- [X] importGlobPlugin
- [X] jsonPlugin
- [X] loadFallbackPlugin
- [X] manifestPlugin
  - [X] `css` and `assets` options are not yet supported as they depend on `chunk.viteMetadata`. [More details ↗](<https://github.com/vitejs/rolldown-vite/blob/2b3bc5b4bc2ce5b0fd00af07214de7f4e358fd57/packages/vite/src/node/plugins/manifest.ts#L130-L135>)
- [X] modulePreloadPolyfillPlugin
- [X] reporterPlugin
  - [X] [https://github.com/rolldown/rolldown/issues/6043](<https://github.com/rolldown/rolldown/issues/6043>)
  - [X] [https://github.com/rolldown/rolldown/issues/6044](<https://github.com/rolldown/rolldown/issues/6044>)
- [X] transformPlugin
  - [X] [https://github.com/rolldown/rolldown/issues/5867](<https://github.com/rolldown/rolldown/issues/5867>)
- [X] resolvePlugin (aligned by @sapphi-red)
- [X] wasmFallbackPlugin
- [X] wasmHelperPlugin
[...]

PR #5944, opened 2025-08-28
feat: port `render_chunk` logic

AliceLanniste:
related #4973, port `renderChunk` logic to rolldown_plugin_build_import_analysis


shulaoda:
Thank you and sorry for late reply, this logic only used in https://github.com/vitejs/rolldown-vite/blob/rolldown-vite/packages/vite/src/node/plugins/modulePreloadPolyfill.ts#L42-L44 and I have inlined it into native module preload polyfill plugin https://github.com/rolldown/rolldown/blob/main/crates/rolldown_plugin_module_preload_polyfill/src/lib.rs#L36.

PR #5949, opened 2025-08-28
feat(dev): default to not eager rebuild

hyf0:
Eager rebuild might cause inconsist module graph/state between rolldown and browser. It will cause generating wrong hmr updates for the client.

Currently, I'm thinking the process would be:
After the initial build, 
- For each file change, we won't trigger rebuild
- For each file change, we will generate hmr updates and invoke `onHmrUpdates` callback.
- `DevServer` will trigger the rebuild in proper timings like user reloads the page to make sure user could access latest content.

Above are the situations. There'll be some smarter actions to be done for improving UX. Like if the hmr updates result to full reload, we'll trigger a rebuild immediately instead of waiting for being called by `DevServer`.

### DIFF
--- a/crates/rolldown/examples/dev.rs
+++ b/crates/rolldown/examples/dev.rs
@@ -13,9 +13,11 @@ async fn main() {
     experimental: Some(ExperimentalOptions { incremental_build: Some(true), ..Default::default() }),
     ..Default::default()
   });
-  let dev_engine =
-    DevEngine::<rolldown_watcher::NotifyWatcher>::new(bundler_builder, DevOptions::default())
-      .unwrap();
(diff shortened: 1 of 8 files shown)


hyf0:
* **#5949** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5949?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5949?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sapphi-red (approved):

PR #5966, opened 2025-08-29
fix: __toESM function breaking ES module imports

### LINKED ISSUE #5934: [Bug] __toESM function breaking ES module imports
## Link to minimal reproduction
https://github.com/edison1105/rolldownToESM

## Steps to Reproduce
1. clone the repo
2. pnpm install
3. pnpm run test2

## What's expected?
should log:
```bash
Rolldown output (uses __toESM): [ '__esModule', 'bar', 'default', 'foo' ]
```

## What's actually happening?
logs:
```bash
Rolldown output (uses __toESM): [ '__esModule', 'bar', 'default' ]
```

## Additional comments
- uncomment line 33 in `bug-repro.js` to see it works fine without `__toESM`
- the underlying issue is that `Object.defineProperty` with a getter is not working in ES module imports, see `simple-test.cjs` for a minimal test case
- This is just a simple reproduction. The actual issue comes from https://github.com/vuejs/core/pull/13599
  - I'm migrating the rollup build of `vuejs/core` to rolldown and encountered this issue. All ecosystem-ci tests are failing with errors saying `SyntaxError: The requested module 'vue' does not provide an export named 'xxx'`. After comparing rollup and rolldown outputs, I found that `__toESM` is the cause.
  - `bug-repro.cjs` simulates vue's build output `vue.cjs.js`
  - Running `pnpm run test` will reproduce the `vue/ecosystem-ci` error

IWANABETHATGUY:
Closed #5934

es-module-lexer only recognize specific pattern to collect exports binding,  see https://github.com/edison1105/rolldownToESM/blob/ce20866468677b1ac17946b6f8be9c00bf99827f/bug-repro.cjs?plain=1#L27-L46 as an example, 
when the module is imported via `import()` 
`cjs-module-lexer` can't recognize pattern like: 
```js
const __mock_library = __toESM(require("./mock-library.cjs"));
```
And assume it is dynamic, so that `foo` is not reexported, but it could recognize pattern like: 
```js
const __mock_library = require("./mock-library.cjs");

exports.bar = function bar() {}

Object.keys(__mock_library).forEach(function (k) {
    enumerable: true,
    get: function () { return __mock_library[k]; }
  });

  // still not working, due to __toESM processed properties as a getter function
  // exports[k] = __mock_library[k]
});


```

After doing some investigation, I found that it also could recognize pattern like: 
```js
let __mock_library = require("./mock-library.cjs");
__mock_library = __toESM(__mock_library);
// ... snip
```

So I changed the importing `external` module code to make sure we have better compatibility with `cjs-module-lexer`

### DIFF
--- a/crates/rolldown/src/ecmascript/format/cjs.rs
+++ b/crates/rolldown/src/ecmascript/format/cjs.rs
@@ -128,19 +128,27 @@ fn render_cjs_chunk_imports(ctx: &GenerateContext<'_>) -> String {
 
       if ctx.link_output.used_symbol_refs.contains(&importee.namespace_ref) {
         let external_module_symbol_name = &ctx.chunk.canonical_names[&importee.namespace_ref];
-        s.push_str("const ");
-        s.push_str(external_module_symbol_name);
-        s.push_str(" = ");
-
-        let to_esm_fn_name = ctx.finalized_string_pattern_for_symbol_ref(
-          ctx.link_output.runtime.resolve_symbol("__toESM"),
-          ctx.chunk_idx,
-          &ctx.chunk.canonical_names,
+        // generate code like:
+        // let external_module_symbol_name = require("external-module");
+        // external_module_symbol_name = __toESM(external_module_symbol_name);
+        let require_external = concat_string!(
+          "let ",
+          external_module_symbol_name,
+          " = ",
+          require_path_str,
+          ";\n",
+          external_module_symbol_name,
+          " = ",
+          ctx.finalized_string_pattern_for_symbol_ref(
+            ctx.link_output.runtime.resolve_symbol("__toESM"),
+            ctx.chunk_idx,
+            &ctx.chunk.canonical_names,
+          ),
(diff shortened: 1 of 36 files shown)


IWANABETHATGUY:
* **#5966** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5966?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5966?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

IWANABETHATGUY:
@sapphi-red Do you mean any better idea, I am not sure if there exists any other way to solve #5934

hyf0 (approved):
Could add a test here: https://github.com/rolldown/rolldown/tree/main/crates/rolldown/tests/rolldown/topics/cjs_module_lexer_compat

PR #5975, opened 2025-08-30
feat(rolldown): oxc v0.87.0

Boshen:


sapphi-red:
* **#5975** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5975?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/5975?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/rolldown/tests/rolldown/issues/2833/main.js
```diff
@@ -1,4 +1,5 @@
 {
   const path = './foo.js';
+  window.foo // to prevent inlining `path`
```
sapphi-red, line 3: Added this line to fix the "Module not found" error

crates/rolldown/tests/rolldown/topics/deconflict/default_function/artifacts.snap
```diff
@@ -42,11 +42,13 @@ assert.strictEqual(typeof foo, "function");
...
 (3:16) "a$1, " --> (5:13) "a$1$1, "
-(3:21) "a)\n" --> (5:20) "a$1);\n"
+(3:21) "a)" --> (5:20) "a$1)"
+(3:23) "\n" --> (5:24) ";\n"
 - ../bar.js
 (2:0) "export default " --> (10:0) "var bar_default = "
 (2:15) "{ " --> (10:18) "{ "
-(2:17) "foo }" --> (10:20) "foo: foo$1 };\n"
+(2:17) "foo }" --> (10:20) "foo: foo$1 }"
+(2:22) [invalid] --> (10:32)
```
sapphi-red, line 51: Maybe rolldown is generating an invalid span?

sapphi-red:
@Boshen It seems `oxc-transform` is missing `optionalDependencies` since 0.83.0.

crates/rolldown/tests/esbuild/dce/dce_of_iife/artifacts.snap
```diff
@@ -13,8 +13,7 @@ keepMe();
 var someVar;
 (([y]) => {})(someVar);
 (({ z }) => {})(someVar);
-var keepThis = /* @__PURE__ */ (() => stuff())();
-keepThis();
```
hyf0, line 17: The test seem to intent to keep it.

crates/rolldown/tests/esbuild/default/nested_require_without_call/artifacts.snap
hyf0, line 1: suspicious

crates/rolldown/tests/esbuild/default/require_without_call/artifacts.snap
hyf0, line 1: suspicious

crates/rolldown/tests/esbuild/default/require_without_call_inside_try/artifacts.snap
hyf0, line 1: I think esbuild doesn't support aliased require. But somehow our inline functionality makes it work.

crates/rolldown/tests/esbuild/ts/ts_enum_jsx/artifacts.snap
```diff
@@ -61,11 +60,10 @@ let x;
 (function(_x) {
 	let y;
 	(function(_y) {
-		let React$1 = /* @__PURE__ */ function(React$2) {
-			React$2["Fragment"] = "div";
-			return React$2;
+		_y.React = /* @__PURE__ */ function(React$1) {
```
hyf0, line 63: Such inline breaks the behavior of inferring`.name` property of nameless function. Does the `keepNames` handle this case? @IWANABETHATGUY cc

crates/rolldown/tests/esbuild/dce/dce_of_iife/artifacts.snap
sapphi-red: The call is properly kept (`(/* @__PURE__ */ (() => stuff())())();`). It is inlined though.

crates/rolldown/tests/esbuild/default/nested_require_without_call/artifacts.snap
sapphi-red: Probably
```js
const req = __require;
req("./entry");
```
is replaced with
```js
__require("./entry");
```

crates/rolldown/tests/esbuild/default/require_without_call_inside_try/artifacts.snap
sapphi-red: If we want to keep testing the aliased require case, we can add a sideeffect in between these two statements like https://github.com/rolldown/rolldown/pull/5975#discussion_r2311826835.

crates/rolldown/tests/esbuild/ts/ts_enum_jsx/artifacts.snap
sapphi-red: Good catch! I'll make the inlining feature to prevent the function to be inlined when `CompressOptions::keep_names::function` is set (also for classes).

Boshen:
Bad news we have more failures from rollup tests 

```
failures [
  "rollup@form@assignment-to-exports: declares updated variable in ES output (#755)@generates es",
  "rollup@form@deep-properties-access: handles deeply nested property accesses",
  "rollup@form@deep-properties: handles deeply nested properties",
  "rollup@form@handles-async-await: properly handles exporting async functions",
  "rollup@form@recursive-computed-members: handle recursive computed member access",
  "rollup@form@recursive-multi-expressions: handles recursive multi-expressions"
]
```

sapphi-red:
All rollup tests failures are snapshot diffs and I think these are all fine.

crates/rolldown/tests/esbuild/default/nested_require_without_call/artifacts.snap
IWANABETHATGUY: I think it is replaced with `require("./entry")`, and then it self becmae a cjs module.

crates/rolldown/tests/esbuild/ts/ts_enum_jsx/artifacts.snap
IWANABETHATGUY: This case is the only one that seems to break the runtime behavior, but @sapphi-red's solution looks good to me

IWANABETHATGUY:
If @sapphi-red could ensure all those tests are only different in snapshot, then we could add these tests to `│ ignored(behavior passed, snapshot different)` category

sapphi-red:
On second thought, do we expect this inlining to happen for DCE? I wonder if I should skip this optimization for DCE.

crates/rolldown/tests/esbuild/default/require_without_call_inside_try/artifacts.snap
hyf0: I think this feature is nice to have, which means if it breaks, we don't pay efforts to solve it, but if it works, no reason to disable it. IMO, it's ok to just let it be. @IWANABETHATGUY cc

crates/rolldown/tests/esbuild/default/nested_require_without_call/artifacts.snap
hyf0: My opinion: https://github.com/rolldown/rolldown/pull/5975/files/f5510f0e5b5580cbd293648e86003cf27e11de52#r2312351961

crates/rolldown/tests/esbuild/default/require_without_call/artifacts.snap
hyf0: Same as https://github.com/rolldown/rolldown/pull/5975#discussion_r2311839632

hyf0:
> On second thought, do we expect this inlining to happen for DCE? I wonder if I should skip this optimization for DCE.

My thoughts:
- DCE visit is unavoidable, it's positive we could do more things during the visit.
- Inlining seems to reduce overheads of the incoming scan phrase(reducing variables, top level statementts, maybe else)

Overall I think it's positive.

IWANABETHATGUY:
> On second thought, do we expect this inlining to happen for DCE? I wonder if I should skip this optimization for DCE.

I think that is acceptable, the biggest requirement of the extra `dce` pass in the scan phase is for deleting dead branches, e.g.
```js
if (process.env.NODE_ENV === "development") {
  require('./a.js')
} else {
  require ('./b.js')
}
```

without inline, the dce pass can't handle scenario: 
```js

const cond = process.env.NODE_ENV
if (cond === "development") {
  require('./a.js')
} else {
  require ('./b.js')
}
```

Overall, I think it is reasonable to put the inline in DCE?
> [!note]
> the `process.env.NODE_ENV` will be replaced by `define` option

Boshen:
@sapphi-red updated oxc, remaining issues are

```
failures [
  "rollup@form@assignment-to-exports: declares updated variable in ES output (#755)@generates es",
  "rollup@form@deep-properties-access: handles deeply nested property accesses",
  "rollup@form@deep-properties: handles deeply nested properties",
  "rollup@form@handles-async-await: properly handles exporting async functions",
  "rollup@form@recursive-computed-members: handle recursive computed member access",
  "rollup@form@recursive-multi-expressions: handles recursive multi-expressions",
  "rollup@form@reexport-star-deshadow: Star reexports scope deshadowing@generates es"
]
```

sapphi-red:
https://github.com/oxc-project/oxc/pull/13477 should fix the failing test.

sapphi-red (dismissed):

IWANABETHATGUY (approved):

IWANABETHATGUY:
mapping issues tracked https://github.com/rolldown/rolldown/issues/6054

IWANABETHATGUY:
flaky test

PR #6025, opened 2025-09-05
feat!: drop CJS format, increase minimum required node

### LINKED ISSUE #5941: ESM-only publishing
Right now, Rolldown is dual-published - but switching to ESM-only publishing would be beneficial:

* No Dual Publishing Hazard (see https://github.com/rolldown/rolldown/pull/5920#issuecomment-3228407801)
* ESM is supported in all non-EOL Node versions (and can be used via `require(esm)` if cjs is necessary)
* Vite is also ESM-only.

See #6025

sxzz:
- Drop CJS format
- Sync minimum required node version to Vite's (20.19+, 22.12+, `require(esm)`).

closes #5941


package.json
```diff
@@ -5,7 +5,7 @@
   "private": true,
   "packageManager": "pnpm@10.15.0",
   "engines": {
-    "node": ">=18.20.3"
```
hyf0, line 8: Probably need to remove node tests CI fo node18 too.

IWANABETHATGUY: It is already removed, we only run tests for node `20`, `22`, `24`

hyf0 (approved):

packages/rolldown/build.ts
```diff
@@ -12,7 +12,7 @@ const isReleasingCI = !!process.env.RELEASING;
 const __dirname = nodePath.join(fileURLToPath(import.meta.url), '..');
 
 // In `@rolldown/browser`, there will be three builds:
```
sxzz:
```suggestion
// In `@rolldown/browser`, there will be two builds:
```

IWANABETHATGUY (approved):

PR #6037, opened 2025-09-07
fix(watch): only consider files read from disk are able to watch

hyf0:


hyf0:
* **#6037** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/6037?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/6037?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sapphi-red (approved):
I think we should add tests that ensures:

- ids loaded via `load` hook is not watched
- ids loaded by the file read is watched
- ids added via `addWatchFile` is watched

hyf0:
> I think we should add tests that ensures:
> 
> * ids loaded via `load` hook is not watched
> * ids loaded by the file read is watched
> * ids added via `addWatchFile` is watched

We should do that.

I think the thrid one already has tests. Others, emm, I don't want to touch watch related code if possible at current stage :). It   requires too much efforts to understand it.

PR #6177, opened 2025-09-12
feat(rolldown_plugin_oxc_runtime)!: embed helpers to support browser environment

### LINKED ISSUE #4719: Improve oxc runtime resolution
Maybe the long-term solution is to include the oxc runtime code in the binary and resolve them directly without performing resolution.

_Originally posted by @hyf0 in https://github.com/rolldown/rolldown/pull/4708#pullrequestreview-2872710461_

shulaoda:
closes #4719

# 🚀 Breaking Change Alert: Rolldown's OXC Runtime Plugin Now Embeds Helpers Directly!

We've revolutionized how Rolldown handles OXC transform helpers with an even better solution:

## Before:
- Automatic dependency resolution from `@oxc-project/runtime` (~229 KB)
- Runtime module resolution required
- External package dependency in node_modules
- Browser environments struggled with module resolution

## After:
- Helpers embedded directly in Rolldown binary - zero external dependencies!
- Only ~25 KB added to binary size (91% smaller than 229 KB package!)
- Works seamlessly in ALL environments including browsers
- **Lightning-fast**: Memory access instead of file I/O - instant helper loading!
- **Still no manual installation needed** - but now even more efficient!

This enhancement makes Rolldown more self-contained and efficient. The OXC runtime helpers are now truly part of the core binary, providing better performance and reliability while maintaining the same seamless developer experience.

The transition is automatic - Rolldown continues to handle everything for you, just more efficiently! 🎉


shulaoda:
* **#6179** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/6179?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6177** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/6177?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/6177?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=6177).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown_plugin_oxc_runtime/build.rs
```diff
@@ -0,0 +1,119 @@
+use oxc_resolver::ResolveOptions;
+use std::collections::BTreeMap;
+use std::env;
+use std::fs;
+use std::io::Write;
+use std::path::PathBuf;
+
+fn main() {
```
hyf0, line 9: @IWANABETHATGUY cc had much experiences on this aspect.

crates/rolldown/tests/esbuild/lower/lower_async_generator/artifacts.snap
```diff
@@ -6,7 +6,7 @@ source: crates/rolldown_testing/src/integration_test.rs
 ## entry.js
 
 ```js
-//#region ../../../../../../node_modules/.pnpm/@oxc-project+runtime@0.87.0/node_modules/@oxc-project/runtime/src/helpers/esm/usingCtx.js
+//#region @oxc-project/runtime/helpers/usingCtx
```
hyf0: I'm thinking about the name. Do we want to add a `rolldown:` prefix to infer it's an rolldown interal thing/virtual module.

crates/rolldown_plugin_oxc_runtime/src/lib.rs
```diff
@@ -1,43 +1,68 @@
...
 impl Plugin for OxcRuntimePlugin {
   fn name(&self) -> Cow<'static, str> {
     Cow::Borrowed("builtin:oxc-runtime")
   }
 
+  fn register_hook_usage(&self) -> HookUsage {
+    HookUsage::ResolveId | HookUsage::Load
+  }
+
   async fn resolve_id(
```
hyf0, line 24: I remember there's an order concept for hook. You need to  enforce them to be `Pre`, otherwise users might be able to overwrite it.

crates/rolldown/tests/esbuild/lower/lower_async_generator/artifacts.snap
IWANABETHATGUY: Since it become a virtual module, I think it ok to add a extension .js` ?

crates/rolldown_plugin_oxc_runtime/src/embedded_helpers.rs
hyf0, line 1: Should this file be emitted? @IWANABETHATGUY cc

crates/rolldown/tests/esbuild/lower/lower_nullish_coalescing_assignment_issue1493/artifacts.snap
```diff
@@ -6,32 +6,32 @@ source: crates/rolldown_testing/src/integration_test.rs
 ## entry.js
 
 ```js
-//#region ../../../../../../node_modules/.pnpm/@oxc-project+runtime@0.87.0/node_modules/@oxc-project/runtime/src/helpers/esm/checkPrivateRedeclaration.js
```
hyf0, line 9: Do we still want the version info? Might be positive to retain it? @IWANABETHATGUY cc

hyf0:
After all, a feature gate should be required. I guess we could enable it by default.

crates/rolldown/tests/esbuild/lower/lower_async_generator/artifacts.snap
hyf0: I'm not sure. Current internal virtual modules have names like `rolldown:runtime` and `rolldown:hmr`. Better to have the same pattern. Or we could switch to `rolldown:runtime.js` or `rolldown:hmr.js`.

crates/rolldown/tests/esbuild/lower/lower_nullish_coalescing_assignment_issue1493/artifacts.snap
IWANABETHATGUY: Probably we could just add the `oxc-runtime` version in cli stdout, so that we don't need it in the output anymore. If the user wants to file an issue, just use the version in cli stdout should be enough.

hyf0: > Probably we could just add the oxc-runtime version in cli stdout, 

@shulaoda cc

Two ways: either keep the version in the bundle output and show the version in the rolldown cli output. Hard to decide which one is better. Leave it to you to decide.

crates/rolldown_plugin_oxc_runtime/src/embedded_helpers.rs
IWANABETHATGUY: Can you elaborate on the `emitting`? Do you mean include separated `runtime.js` rather than use `static_string`?

hyf0: I mean "should the `embedded_helpers.rs` be submitted to `git`".

IWANABETHATGUY: I think it's fine to submit to git(`oxc` also has a large visit.rs file ), since it is auto-generated.

crates/rolldown/tests/esbuild/lower/lower_async_generator/artifacts.snap
IWANABETHATGUY: Image

Note sure which one is better, but vite polyfill virtual module ends with `.js`

crates/rolldown_plugin_oxc_runtime/src/embedded_helpers.rs
IWANABETHATGUY: Also, if it is not submitted to `git` the binary can not be compiled.

crates/rolldown/tests/esbuild/lower/lower_async_generator/artifacts.snap
IWANABETHATGUY: cc @sapphi-red would you mind giving us some advice?

hyf0: @sapphi-red Any suggestion on the name part? I think it's different style. We can switch vite's style and be like `\0rolldown/runtime.js` and `\0rolldown/hmr.js`.

sapphi-red: For virtual modules, we should have a `\0` prefix to follow [the convention](https://rollupjs.org/plugin-development/#conventions). Having `.js` at the end is not a must, but it would be better as filters like `*.js` would match that.

sapphi-red:
> After all, a feature gate should be required. I guess we could enable it by default.

@hyf0 What is the reason to put it behind a feature?

hyf0:
> > After all, a feature gate should be required. I guess we could enable it by default.
> 
> @hyf0 What is the reason to put it behind a feature?

Just in case. Think again, maybe not. Since it's the future we want, we should just fix possible bugs directly based on this feature.

crates/rolldown/tests/esbuild/lower/lower_async_generator/artifacts.snap
hyf0: Yeah. We should follow the rollup's pattern on this part. A future refactor would be needed.

@shulaoda cc

sapphi-red:
If we want to give a way for users opt-out this in case of bugs, I think we should add a temporary option instead of a feature gate. (or did you mean an option by feature gate?)
But I guess it's fine without an option.

hyf0:
> If we want to give a way for users opt-out this in case of bugs, I think we should add a temporary option instead of a feature gate. (or did you mean an option by feature gate?) But I guess it's fine without an option.

Oh yeah. I mean a option in `experimentalOptions`, not a rust featrue gate.

crates/rolldown/tests/esbuild/lower/lower_nullish_coalescing_assignment_issue1493/artifacts.snap
shulaoda: I’ll implement this in the next PR.

hyf0 (approved):

PR #6233, opened 2025-09-16
refactor(dev): task-queue based deisgn

hyf0:
In the previous design, it’s hard to think about how to handle complex cases. In this design, by abstract each change into a queued task, it makes parallel events sequential, which reduces the complexity of thinking itself.


hyf0:
* **#6233** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/6233?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/6233?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6232** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/6232?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=6233).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown/src/dev/building_task.rs
```diff
@@ -0,0 +1,161 @@
...
+    &self.input
+  }
+}
+
+impl BundlingTask {
+  pub async fn run(mut self) {
+    if let Err(err) = self.run_inner().await {
+      // FIXME: Should handle the error properly.
+      tracing::error!("Failed to run bundling task: {err:?}");
+      self.dev_context.state.lock().await.try_to_idle().expect("FIXME: Should not unwrap here");
```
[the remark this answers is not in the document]
hyf0: I'm thinking how to handle cache if there's an error. But not this way.

crates/rolldown/src/dev/build_driver.rs
```diff
@@ -31,56 +31,55 @@ impl BuildDriver {
...
       tracing::trace!(
-        "Schedule a build to consume pending changed files due to {:?} or {:?}",
-        build_state.require_full_rebuild,
-        build_state.changed_files
+        "Schedule a build to consume pending changed files due to task{task_input:#?}",
       );
-      // Note: Full rebuild and incremental build both clear changed files.
-      let changed_files = mem::take(&mut build_state.changed_files);
+
+      // TODO: hyf0 merge mergeable task inputs into one
```
hyf0, line 64: Do we want the merge behavior? Or at least make it optional? @sapphi-red 

- Not merging ensures users could see newest changes if possible. Merging might cause a task that's able to do a successful hmr update fail.
- If we want to merge, the strategy would be a little sophisticated. If the current task is a [full] rebuild only task, it will merge all subsquent tasks until finding one that require to generate hmr update.

sapphi-red: I think we want it. There won't be multiple hmr tasks queued unless the hmr is slow. This happens when there's a slow plugin. In that case, I think we should merge the tasks to avoid causing the hmr to be lagged behind. I think that has a benefit even with the downsides you described.
But this is an edge case because normally the HMR is instant and the watcher debounce will merge the file change events (i.e. this only happens if the HMR takes longer than the debounce tick rate). So I think we can leave it for now as long as the multiple rebuild tasks are merged (I assume this is done indirectly by the delay state). I think improving the error handling has a higher priority as that would affect more users.

sapphi-red (approved):

crates/rolldown/src/dev/build_driver.rs
hyf0: Got it. I'll start to think a solution for error handling.

PR #6340, opened 2025-09-25
chore(node): use built-in styleText

dumbmatter:
Node's styleText is supported in all the Node versions listed in the main package.json, so you can drop a couple dependencies by switching to it.


shulaoda (approved):
Thank you for your contribution! ❤

PR #6346, opened 2025-09-26
fix(rolldown_binding): use `Weak<OutputChunk>` to prevent memory leak

### LINKED ISSUE #6121: [Bug]: Memory is not released promptly with API usage
Some builds uses `rolldown` through its API:

```js
import * as rolldown from "rolldown";

await rolldown.build({});

// Check with `Activity Monitor` or `top`.
// Memory is not released.

setTimeout(() => {}, 86400000); // don't exit the process
```

Memory is not released after the `await rolldown.build({});` call.

---

This is reported by Framer.
This should be scheduled after Vite conf.

Boshen:
may fix #6121

The js side should not own the Arc.

### DIFF
--- a/crates/rolldown/src/bundler.rs
+++ b/crates/rolldown/src/bundler.rs
@@ -13,7 +13,7 @@ use anyhow::Result;
 
 use arcstr::ArcStr;
 use rolldown_common::{
-  GetLocalDbMut, Module, NormalizedBundlerOptions, ScanMode, SharedFileEmitter, SymbolRefDb,
+  GetLocalDbMut, Module, NormalizedBundlerOptions, Output, ScanMode, SharedFileEmitter, SymbolRefDb,
 };
 use rolldown_debug::{action, trace_action, trace_action_enabled};
 use rolldown_error::{BuildDiagnostic, BuildResult, Severity};
@@ -40,6 +40,8 @@ pub struct Bundler {
   pub(crate) cache: ScanStageCache,
   pub(crate) session: rolldown_debug::Session,
   pub(crate) build_count: u32,
+  pub(crate) assets: Vec<Output>,
+  // pub(crate) profiler: Option<dhat::Profiler>,
 }
 
 impl Bundler {
@@ -105,6 +107,11 @@ impl Bundler {
 
     self.closed = true;
     self.plugin_driver.close_bundle().await?;
+
+    self.assets.clear();
+    self.cache = ScanStageCache::default();
+    self.file_emitter.clear();
+    // self.profiler = None;
 
     Ok(())
   }
@@ -254,6 +261,7 @@ impl Bundler {
     }
 
     let mut output = bundle_output?;
+    self.assets.extend(output.assets.iter().map(Clone::clone));
 
     // Add additional files from build plugins.
     self.file_emitter.add_additional_files(&mut output.assets, &mut output.warnings);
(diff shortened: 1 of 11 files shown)


crates/rolldown_binding/src/types/binding_output_asset.rs
```diff
@@ -1,48 +1,53 @@
...
 
 #[napi]
 impl BindingOutputAsset {
-  pub fn new(inner: Arc<rolldown_common::OutputAsset>) -> Self {
+  pub fn new(inner: Weak<OutputAsset>) -> Self {
     Self { inner }
   }
 
+  fn inner(&self) -> Arc<OutputAsset> {
+    self.inner.upgrade().unwrap()
```
Boshen, line 20: Instead of crashing, the js side can set the value to `undefined`, or we throw a panic message "Rolldown has been shutdown, access to this data is no longer possible".

crates/rolldown/src/bundler.rs
```diff
@@ -40,6 +40,8 @@ pub struct Bundler {
   pub(crate) cache: ScanStageCache,
   pub(crate) session: rolldown_debug::Session,
   pub(crate) build_count: u32,
+  pub(crate) assets: Vec<Output>,
```
Boshen, line 43: `Bundler` needs to own the `Arc`s from `Output`.

crates/rolldown/src/bundler.rs
```diff
@@ -106,6 +108,11 @@ impl Bundler {
     self.closed = true;
     self.plugin_driver.close_bundle().await?;
 
+    self.assets.clear();
+    self.cache = ScanStageCache::default();
+    self.file_emitter.clear();
```
Boshen, line 113: We need to clear everything stored in `Bundler` instead of relying on `Drop`.

hyf0: Not oppose this direction, but would it casue error if the bundler/assets get dropped and user try to access the assets?

This brings up a convention that if users wants to access the asset, they shouldn't make the the js object `Bundler` get GCed, which is hard thing to tell in the js world due to js doesn't have ownership concept.

crates/rolldown/src/bundler.rs
```diff
@@ -254,6 +261,7 @@ impl Bundler {
     }
 
     let mut output = bundle_output?;
+    self.assets.extend(output.assets.iter().map(Clone::clone));
```
hyf0, line 264: This line might get triggered many times due to incremental build.

crates/rolldown/src/bundler.rs
hyf0: Not oppose this direction, but would it casue error if the bundler/assets get dropped and user try to access the assets?

This brings up a convention that if users wants to access the asset, they shouldn't make the the js object `Bundler` get GCed, which is hard thing to tell in the js world due to js doesn't have ownership concept.

crates/rolldown/src/bundler.rs
Boshen: Something else should own the assets, I stored them in bundler to prove my hypothesis.

crates/rolldown/src/bundler.rs
Boshen: Yes, for script usages, use after free will not be allowed. i.e. it will crash when accessing asset data after rolldown has been closed.

crates/rolldown/src/bundler.rs
hyf0: I think this improvement could be appied first. No negative point found.

hyf0: Had applied this one in https://github.com/rolldown/rolldown/pull/6483.

PR #6400, opened 2025-10-02
fix: respect package.json type field in plugin-resolved modules

### LINKED ISSUE #6398: `ModuleDefFormat::*PackageJson` cannot be set by plugins
`ModuleDefFormat::*PackageJson` is only set by `infer_module_def_format` function here:
https://github.com/rolldown/rolldown/blob/7dec74650fce81f054508a60516a5552d40b9a45/crates/rolldown_resolver/src/resolver.rs#L291-L312
This is called in this `resolve_id` call, which happens when no plugin resolves that id.
https://github.com/rolldown/rolldown/blob/7dec74650fce81f054508a60516a5552d40b9a45/crates/rolldown_plugin/src/utils/resolve_id_with_plugins.rs#L95

But `module_def_format` is set by `ModuleDefFormat::from_path` for ids resolved by plugins.
https://github.com/rolldown/rolldown/blob/7dec74650fce81f054508a60516a5552d40b9a45/crates/rolldown_plugin/src/utils/resolve_id_with_plugins.rs#L76
This does not use the `package.json` data and only uses the extensions.

This causes inconsistent behavior when an id is resolved by a plugin and the plugin cannot even fix that on their side as there's no way to set `module_def_format` from a plugin.

This is causing a problem in rolldown-vite as it uses rolldown's resolver for dep optimizer and uses the custom resolver for build (https://github.com/vitejs/rolldown-vite/issues/421). rolldown-vite can use the custom resolver for both dep optimizer and build, but that means rolldown-vite will behave differently from rolldown for https://github.com/rolldown/rolldown/issues/4289.

hyf0:
partially fixes, #6398

When plugins resolve module IDs, ModuleDefFormat now checks both file extensions and package.json "type" field, matching the behavior of the internal resolver. This ensures consistent module format determination regardless of resolution path.

Changes:
- Add infer_module_def_format() helper in resolve_id_with_plugins.rs
- Update resolve_dynamic_import and resolve_id hooks to use helper
- Add test case with ESM package resolved via internal resolver

### DIFF
--- a/crates/rolldown_plugin/src/utils/resolve_id_with_plugins.rs
+++ b/crates/rolldown_plugin/src/utils/resolve_id_with_plugins.rs
@@ -2,7 +2,9 @@ use crate::{
   HookResolveIdArgs, PluginDriver,
   types::{custom_field::CustomField, hook_resolve_id_skipped::HookResolveIdSkipped},
 };
-use rolldown_common::{ImportKind, ModuleDefFormat, ResolvedId, is_existing_node_builtin_modules};
+use rolldown_common::{
+  ImportKind, ModuleDefFormat, PackageJson, ResolvedId, is_existing_node_builtin_modules,
+};
 use rolldown_resolver::{ResolveError, Resolver};
 use std::{
   path::{Path, PathBuf},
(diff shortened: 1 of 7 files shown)


hyf0:
* **#6400** <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/6400?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/rolldown/rolldown/6400?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=6400).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

sapphi-red (approved):
This would fix for plugins that uses `this.resolve`, but this would not fix the case for rolldown-vite because `resolver.package_json_cache()` is populated only when the internal resolver is used.

PR #6430, opened 2025-10-06
chore(just): Skip corepack if not available

untitaker:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] yes (_bugfixes and features will not be merged without tests_)
- [x] no

Breaking Changes?

- [ ] yes (_breaking changes will not be merged unless absolutely necessary_)
- [x] no

List any relevant issue numbers:

### Description

The contribution docs are written as if corepack is not mandatory, but
eventually I get told to run `just setup` which does kinda mandate it.
Let's just skip it if it's not there, I already set up pnpm and node up
via volta and it works fine.

### DIFF
--- a/justfile
+++ b/justfile
@@ -12,7 +12,7 @@ setup:
   cargo install cargo-binstall
   cargo binstall cargo-insta cargo-deny cargo-shear typos-cli -y
   # Node.js related setup
-  corepack enable
+  (! which corepack) || corepack enable
   pnpm install
   just setup-submodule
   just setup-bench
--- a/justfile
+++ b/justfile
@@ -12,11 +12,21 @@ setup:
   cargo install cargo-binstall
   cargo binstall cargo-insta cargo-deny cargo-shear typos-cli -y
   # Node.js related setup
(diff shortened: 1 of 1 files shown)


hyf0:
@shulaoda Could you check if this works on Windows?

shulaoda:
> @shulaoda Could you check if this works on Windows?

It doesn't work on Windows. I will find a better way based on this.

untitaker:
I hacked through it. I didn't test it e2e on windows, only tested the commands in the Ubuntu Powershell snap.

hyf0:
Think again. We should just enfore using corepack with providing guide on how to install it. Any potentially negative reason we don't want to do that?

untitaker:
It's difficult to contribute to many projects that need different node version managers. But I am not a frequent JavaScript contributor and corepack seems "official" nowadays.

hyf0:
Updated ref: https://github.com/rolldown/rolldown/issues/6473

PR #6435, opened 2025-10-07
chore: change node version to lts

iiio2:
Can we change the node version to lts here ? Actually now the lts is 22.20.0. And in the code the node version is 22.19.0. I thought may be if we change to lts then we are always up-to-date with the lts node release.


.node-version
```diff
@@ -1 +1 @@
-22.19.0
+lts/*
```
hyf0, line 1: I prefer to use lts of node22. What version would `lts/*` resolve to?

iiio2: @hyf0 , Oh I see. Actually i was thinking this if suppose after some months node 24 becomes lts, then should we need to change the node version here or not.

hyf0: I just noticed node22 ends lts in two weeks. If `lts/*` resolve to latest lts version, I think the change is fine.

IWANABETHATGUY (approved):

iiio2:
Thanks @hyf0 , @IWANABETHATGUY . :blush:

PR #6441, opened 2025-10-07
chore(examples): add `"type": "module"` in `package.json`

iiio2:
Thought it is in examples folder, node.js recommends to explicitly mention type: module in package.json. Otherwise it incurs performance cost. (https://nodejs.org/en/blog/release/v20.19.0#module-syntax-detection-is-now-enabled-by-default). That's why I thought may be we can include it.


IWANABETHATGUY (approved):

iiio2:
Thanks @IWANABETHATGUY .

PR #6484, opened 2025-10-11
fix: take `resolve.modules` into account

Pickachu:
## Summary

When `resolve.modules` is configured via the JavaScript API, it never reaches the native resolver. This causes module resolution to fall back to the default `["node_modules"]` path set inside oxc-resolver, even when custom module directories are explicitly specified.

## Root Cause

The `BindingResolveOptions` struct in `crates/rolldown_binding/src/options/binding_input_options/binding_resolve_options.rs` includes a `modules` field but fails to forward it in the `From<BindingResolveOptions> for rolldown::ResolveOptions` implementation. All other resolve options (alias, alias_fields, condition_names, etc.) are properly forwarded—`modules` was simply omitted.

## Changes

Added the missing `modules` field assignment in the conversion:

```rust
impl From<BindingResolveOptions> for rolldown::ResolveOptions {
  fn from(value: BindingResolveOptions) -> Self {
    Self {
      // ... other fields
      modules: value.modules,  // ← Added
      // ... remaining fields
    }
  }
}
```

## Questions

1. Was the omission of `modules` intentional? If so, what was the reasoning?
2. I believe this is the only place that needs to change to fix the issue, but I'm not 100% certain—could there be other spots in the binding layer that need updating?

## Reference

- [Rolldown Module Resolution Documentation](https://rolldown.rs/guide/features#module-resolution)

### DIFF
--- a/crates/rolldown_common/src/inner_bundler_options/types/resolve_options.rs
+++ b/crates/rolldown_common/src/inner_bundler_options/types/resolve_options.rs
@@ -20,6 +20,7 @@ pub struct ResolveOptions {
   pub extension_alias: Option<Vec<(String, Vec<String>)>>,
   pub main_fields: Option<Vec<String>>,
   pub main_files: Option<Vec<String>>,
+  pub modules: Option<Vec<String>>,
   pub symlinks: Option<bool>,
   pub yarn_pnp: Option<bool>,
 }
--- a/packages/rolldown/tests/fixtures/resolve/custom-modules/_config.ts
+++ b/packages/rolldown/tests/fixtures/resolve/custom-modules/_config.ts
@@ -1,0 +1,9 @@
+import { defineTest } from 'rolldown-tests'
+
+export default defineTest({
+  config: {
+    resolve: {
+      modules: ['custom-node-modules'],
+    },
+  },
+})
--- a/crates/rolldown_binding/src/options/binding_input_options/binding_resolve_options.rs
+++ b/crates/rolldown_binding/src/options/binding_input_options/binding_resolve_options.rs
@@ -36,6 +36,7 @@ impl From<BindingResolveOptions> for rolldown::ResolveOptions {
       extension_alias: value.extension_alias.map(|alias| {
         alias.into_iter().map(|item| (item.target, item.replacements)).collect::<Vec<_>>()
       }),
+      modules: value.modules,
       main_fields: value.main_fields,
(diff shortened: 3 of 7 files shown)


crates/rolldown_common/src/inner_bundler_options/types/resolve_options.rs
```diff
@@ -20,6 +20,7 @@ pub struct ResolveOptions {
   pub extension_alias: Option<Vec<(String, Vec<String>)>>,
   pub main_fields: Option<Vec<String>>,
   pub main_files: Option<Vec<String>>,
+  pub modules: Option<Vec<String>>,
```
hyf0, line 23: You need to pass it here

https://github.com/rolldown/rolldown/blob/0dd43df3e6361db141b41b118ddde191997400cb/crates/rolldown_resolver/src/resolver.rs#L150

Pickachu: Thanks for catching that! Updated both the binding and resolver—appreciate your help! 🙌

hyf0: Could you have a test in `rolldown` package?

You could put it under `packages/rolldown/tests/fixtures/resolve/custom-modules`.

The test logic could be simple. As long as `import foo from 'foo'; assert.strictEqual(foo, 'foo')` in `main.js` could resolve `foo` package in `custom_node_modules_folder`, that would be fine.

https://rolldown.rs/contrib-guide/testing#node-js-api-tests

packages/rolldown/tests/fixtures/resolve/custom-modules/_config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineTest } from 'rolldown-tests'
+import path from 'node:path'
+
+export default defineTest({
+  config: {
+    resolve: {
+      modules: [path.join(__dirname, 'custom_node_modules')],
```
hyf0: I believe it should be

```suggestion
      modules: ['custom_node_modules'],
```

hyf0 (approved):
Good work. Thank you!

crates/rolldown_common/src/inner_bundler_options/types/resolve_options.rs
Pickachu: Yes! No problem.

PR #6486, opened 2025-10-12
feat(rolldown): support `output.clearDir` to clean up `dir` before build

### LINKED ISSUE #6485: [Feature Request]: Empty Output Directory

Add an option to empty output directory before build, just like the build.emptyOutDir options in [vite](https://vite.dev/config/build-options.html#build-emptyoutdir).

```ts
import { defineConfig } from "rolldown"

export default defineConfig({
  ...,
  output: {
    ...,
    dir: xxx,
    emptyOutDir: true,
 }
})
``` 

#6486

aprosail:
Provide an output option (`OutputOptions`) to empty output directory (`emptyOutDir`): When both `emptyOutDir` and `dir` in the output options are configured, it will empty the configured output directory (remove all files and directories recursively inside). You may configure like this:

```ts
import {defineConfig} from "rolldown"

export default defineConfig({
  ...,
  output: {
    dir: './dist',
    cleanDir: true,
  }
})
```

Inspired by [vite](https://vite.dev/config/build-options.html#build-emptyoutdir).


aprosail:
Need to resolve security risks: https://github.com/rolldown/rolldown/issues/6485#issuecomment-3394055784
Use such API:

```ts
export interface OutputOptions {
  ...
  /**
   * Empty output directory before writing.
   *
   * - `'force'`: clean the out dir even if it's not inside project root.
   * - `true`: only clean the out dir inside project root.
   * - `false` or `undefined`: don't do anything about the out dir.
   */
  emptyOutDir?: boolean | 'force';
}
```

aprosail:
There might still something to improve: Currently it uses `cwd` rather than the `root` option configured inside `input`.

aprosail:
Currently I don't have mode free time to focus on such PR (diving down into rolldown's source code is rather difficult for a beginner like me)... I've made [a plugin](https://www.npmjs.com/package/rolldown-plugin-empty-outdir) to satisfy such demand.

Cargo.toml
```diff
@@ -143,6 +143,7 @@ css-module-lexer = "0.0.15"
 dashmap = "6.1.0"
 derive_more = { version = "2.0.1", features = ["debug"] }
 dunce = "1.0.5" # Normalize Windows paths to the most compatible format, avoiding UNC where possible
+empty_inside_dir = "1.1.0"
```
hyf0: For mataince concern, we don't want to introduce outer dependency unless we have to.

crates/rolldown/src/bundler.rs
```diff
@@ -235,6 +236,21 @@ impl Bundler {
...
     let dist_dir = self.options.cwd.join(&self.options.out_dir);
 
+    if self.options.dir.is_some() {
+      match self.options.empty_out_dir {
+        EmptyOutDirMode::Disabled => {}
+        EmptyOutDirMode::Force => {
+          empty_dir(&dist_dir).expect("cannot empty out dir (force mode)");
+        }
+        EmptyOutDirMode::Normal => {
+          // todo: how to resolve real root?
```
hyf0: `cwd` is the root.

hyf0 (changes requested):
Could you also change the option name to `cleanDir`?

crates/rolldown/src/bundler.rs
hyf0: `dir` supports to receive absolute path, so I think you just compare whether the dir is inside of `cwd`. Raise a warning and don't do the clean up if `dir` isn't inside of `cwd`. @sapphi-red cc

sapphi-red: I think we can remove `EmptyOutDirMode::Normal` completely as it's not enabled by default (https://github.com/rolldown/rolldown/issues/6485#issuecomment-3404613381).

hyf0: @aprosail cc

Cargo.toml
aprosail: ~~Got it, I'll then copy all those code inside the `rolldown_utils` crate.~~

The `rolldown_utils` crate has a structured code style that I don't know how create a `fn clean_dir` inside such crate properly. Maybe I need to modify the encapsulated `rolldown_fs` crate...

@hyf0 you are right, rolldown uses another self-maintained and higher-level file system inside the `rolldown_fs` crate, which may incompatible with 3rd-party utils crates. I'm now trying to implement the `fn empty_dir` inside such crate...

crates/rolldown/src/bundler.rs
aprosail: Got it. So I'll then turned it back to `bool` type, which is also much more simpler in code structure.

Cargo.toml
aprosail: Already removed such deps, and implement `clean_dir` inside the `rolldown_fs` module since [this commit](https://github.com/rolldown/rolldown/pull/6486/commits/bbe19e9b21fa80872f1d15392af41a026a885875).

crates/rolldown/src/bundler.rs
aprosail: Already removed the enum and corresponding detect-inside-cwd logic since [this commit](https://github.com/rolldown/rolldown/pull/6486/commits/e90bd869853aaeaf64ee0a9882a680ab49d1d7be).

aprosail:
> Could you also change the option name to `cleanDir`?

WIP...

aprosail:
> Could you also change the option name to `cleanDir`?

Done.

crates/rolldown/src/bundler.rs
```diff
@@ -235,6 +234,10 @@ impl Bundler {
 
     let dist_dir = self.options.cwd.join(&self.options.out_dir);
 
+    if self.options.clean_dir && self.options.dir.is_some() {
+      self.fs.clean_dir(&dist_dir).expect("cannot clean out dir");
```
hyf0: Don't call `expect`, map the error to unhandleable and return it.

crates/rolldown_fs/src/file_system.rs
```diff
@@ -27,4 +30,52 @@ pub trait FileSystem: Send + Sync + OxcResolverFileSystem {
...
+}
+
+/// Utility trait for file system operations
+pub trait FileSystemUtils: FileSystem {
+  /// Empty the contents of a directory without deleting the directory itself.
+  ///
+  /// 1. When the path is not a directory, it will return `Err`.
+  /// 2. When the path not exist, nothing will happen, it will return `Ok`.
+  /// 3. Only when the path is an existing directory, it will empty inside.
+  fn clean_dir(&self, path: &Path) -> io::Result<()> {
```
hyf0: No need to add trait for this method. You could put in the `rolldown_uitls` crate.

crates/rolldown_fs/src/memory.rs
```diff
@@ -172,4 +196,142 @@ mod tests {
 
     Ok(())
   }
+
+  #[cfg(test)]
```
hyf0: These tests are not needed.

crates/rolldown_fs/src/os.rs
```diff
@@ -61,3 +77,99 @@ impl OxcResolverFileSystem for OsFileSystem {
     self.0.read_link(path)
   }
 }
+
+#[cfg(test)]
```
hyf0: Not needed too.

crates/rolldown_fs/src/memory.rs
hyf0: Having a integration test would be fine. I guess this feature can't be tested by fixture test.

Could you create a test at `packages/rolldown/tests/behaviors/clean-dir.test.ts`?

The test will call the bundler twice and generate output with different names, so you could observe whether the output gets deleted.

crates/rolldown_fs/src/file_system.rs
aprosail: @hyf0 I saw there are many other code calling the `self.fs` APIs besides, which is defined in the `rolldown_fs` crate, and can support both `os` and `memory` mode. **Are you sure just a utils function calling `std::fs` is okay here?** I don't know whether there are potential risks if such function cannot support `vfs` called in `rolldown_fs`.

crates/rolldown_fs/src/memory.rs
aprosail: Okay, I'll move such tests into a more "end-point" position, as detailed tests seems unnecessary here...

crates/rolldown_fs/src/memory.rs
```diff
@@ -85,8 +85,32 @@ impl FileSystem for MemoryFileSystem {
...
+      .read_dir(path_str.as_ref())
+      .map_err(|err| io::Error::new(io::ErrorKind::NotFound, err))?;
+
+    let mut paths = Vec::new();
+    for entry in entries {
+      let entry_path = PathBuf::from(entry);
+      paths.push(entry_path);
+    }
+    Ok(paths)
+  }
```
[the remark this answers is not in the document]
aprosail: Not necessary, resolve manually.

crates/rolldown_fs/src/file_system.rs
hyf0: I mean you could have a function called `clean_dir(fs: &dyn Fs)` like this. You don't need to have a trait to do that. I don't see any benefit on this part.

We intend to treat `rolldown_fs` as a plain crate that only defines the fundumental inteface of fs.

aprosail: The `fn clean_dir` can build on the top of **only one of** the original file system provided by `std::fs`, or the `FileSystem` abstraction provided by the `rolldown_fs` crate, because they are in different structures:

- `std::fs` define `fn`s like `read_dir` or `remove_file` directly at the root of certain `mod` (see the [source code example](https://doc.rust-lang.org/stable/src/std/fs.rs.html#304-313)), there are no `trait` abstraction.
- while crate `rolldown_fs` encapsulate them all inside a `trait`, that once call such `fn`s, we need to call an instance that `impl` such `trait`, which is different.

If I just create a `fn` to handle that, I came up with only theose four proper ways:

1. Put such `fn` inside `rolldown_fs` and it requires a parameter that implement `trait rolldown_fs::FileSystem`. And such `fn` can only use with the `rolldown_fs` system.
2. Put such `fn` inside `rolldown_utils`, but it will depends on the `FileSystem` abstraction defined in `rolldown_fs`, that the crate `rolldown_utils` will depends on `rolldown_fs`, which introduce new dependencies (although inside the workspace).
3. Make such `fn clean_dir` only support `std::fs` file system using direct `fn`s. This might cause potential risks in further development, but currently, it can pass all the rust tests (perhaps because there are not test about that yet).
4. Create another `FileSystem` abstraction in crate `rolldown_utils`, or even let `rolldown_fs` depends on `rolldown_utils`, that a `fn clean_dir` in `rolldown_utils` can compat with such APIs properly.

Please tell me which approach is better in your opinion, or is there any other ways that I didn't find out?

crates/rolldown/src/bundler.rs
aprosail: Oh! Really sorry about that, I made a mistake 😂 (fixed in [this commit](https://github.com/rolldown/rolldown/pull/6486/commits/1afb9eae710216962bd5ea253e6bfe0cb3367627))

crates/rolldown_fs/src/memory.rs
aprosail: Done in [this commit](https://github.com/rolldown/rolldown/pull/6486/commits/26b6e4e210cd12ad6e53c2a0b5b5bffa4fb3ef9e).

crates/rolldown_fs/src/file_system.rs
aprosail: Since [this commit](https://github.com/rolldown/rolldown/pull/6486/commits/7803511f36dc8fd72b0cebf7e2a749f23a3026e0), it can work normally, and all tests passed. The unnecessary tests had also been removed. I'm just wondering how to optimize the trait structure...

hyf0:
Hi @aprosail! 👋

I've created #6577 which implements the refactoring suggested by @hyf0 in https://github.com/rolldown/rolldown/pull/6486#discussion_r2434437626.

The main change is moving `clean_dir` from a `FileSystemUtils` trait to a standalone function in `crates/rolldown/src/utils/fs_utils.rs`. This keeps the `rolldown_fs` crate focused on fundamental filesystem interfaces as intended.

## To integrate these changes into your PR:

```bash
# Add the upstream remote if you haven't already
git remote add upstream https://github.com/rolldown/rolldown.git

# Fetch the refactoring branch
git fetch upstream refactor/clean-dir-standalone-function

# Cherry-pick the commit
git cherry-pick 6fb679093
```

All tests pass with these changes, and the implementation is cleaner with better type safety using generics instead of trait objects.

Let me know if you need any help integrating this!

crates/rolldown_fs/src/file_system.rs
hyf0: See https://github.com/rolldown/rolldown/blob/6fb679093eaf17a81a0550763a4078e0fa89e53c/crates/rolldown/src/utils/fs_utils.rs of https://github.com/rolldown/rolldown/pull/6577.

packages/rolldown/src/options/output-options.ts
```diff
@@ -450,7 +450,13 @@ export interface OutputOptions {
    */
   minifyInternalExports?: boolean;
   /**
-   * Keep function and class names after bundling.
+   * - Type: `boolean`
+   * - Default: `false`
+   *
+   * Clean output directory before writing.
```
hyf0:
```suggestion
   * Clean output directory before emitting output.
```

packages/rolldown/src/utils/validator.ts
```diff
@@ -882,6 +882,10 @@ const OutputOptionsSchema = v.strictObject({
     v.optional(v.boolean()),
     v.description('Rewrite top-level declarations to use `var`.'),
   ),
+  cleanDir: v.pipe(
+    v.optional(v.boolean()),
+    v.description('Clean output directory before writing'),
```
hyf0:
```suggestion
    v.description('Clean output directory before emitting output'),
```

hyf0: Run `just test-node-rolldown --update` to update cli's snapshot.

packages/rolldown/tests/behaviors/clean-dir.test.ts
```diff
@@ -0,0 +1,22 @@
...
+import { join } from 'node:path';
+import { rolldown } from 'rolldown';
+import { expect, test } from 'vitest';
+
+const root = import.meta.dirname
+
+test('clean out dir', async () => {
+  const bundler = await rolldown({ input: 'index.ts', cwd: root })
+  const outdir = 'dist'
+  await bundler.write({ dir: outdir, entryFileNames: 'index1.js', cleanDir: true })
```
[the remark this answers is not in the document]
aprosail: Use `clean`, see [this issue comment](https://github.com/rolldown/rolldown/issues/6485#issuecomment-3404751135).

packages/rolldown/src/utils/validator.ts
aprosail: Done, and all tests passed.

aprosail:
Changes had been compressed from 500+ lines into 150+ lines 😂

hyf0:
Could you run `just lint` locally and fix related errors.

hyf0 (approved):
Thank you for contributing such an amazing feature!

aprosail:
Thank you for so detailed guidance to a newcomer like me contributing to Rolldown for the first time 😄

aprosail:
> Could you run `just lint` locally and fix related errors.

This error might be something about timeout configuration: sometimes it failed, but mostly it will pass. Sorry that I still cannot have a deep understand about the workspace code workflow structures, that maybe I cannot resolve such bug yet...

lazka:
It seems this is being run after `generateBundle` and before `writeBundle` (based on my testing), which results in all files emitted by plugins before `writeBundle` being deleted too. Is that on purpose?

sapphi-red:
That is an expected behavior for me.

hyf0:
> It seems this is being run after `generateBundle` and before `writeBundle` (based on my testing), which results in all files emitted by plugins before `writeBundle` being deleted too. Is that on purpose?

If this's common community plugin pattern, It's ok to advance the timing.

lazka:
I've only had a problem with [rollup-plugin-copy](https://github.com/vladshcherbin/rollup-plugin-copy/), but that allows setting target hook to run in, so easily worked around. I noticed that other plugins that create files in generateBundle via emitFile seem to work fine too. So no big problem for me at least, I was just curious since vite seems to delete in renderStart.

lazka:
Ah, [@rollup/plugin-url](https://github.com/rollup/plugins/tree/master/packages/url) is broken too. It copies files manually in `generateBundle`.

aprosail:
@hyf0 https://github.com/rolldown/rolldown/pull/6647 how about a fix like this?

This PR already clean dir before the `generateBundle` hook,
that all files generated inside the `generateBundle` hook will not be deleted.

hyf0:
@aprosail Thank you for your work. @lazka we had changed the behavior and doc. We'll release a new version Monday including that change.

PR #6647, opened 2025-10-21
fix(clean-dir): advance clean outdir to allow `generateBundle` outputs

aprosail:
See: https://github.com/rolldown/rolldown/pull/6486#issuecomment-3421660516

Advance the "clean dir" code before the `bundle_up` is called inside the `bundle_write` function: `fn generate_bundle` is called in `fn bundle_up`, and the "clean dir" code is advanced before the call of `fn bundle_up` in `fn bundle_write`.

- [x] Fix the source code and pass all current tests.
- [x] Create new behavior tests about the `generateBundle` and `writeBundle` hook.

I also fix another bug about testing timeout configuration, which prevents the previous random failure when testing spent more time than the default 20s (20_000ms).

### DIFF
--- a/crates/rolldown/src/bundler.rs
+++ b/crates/rolldown/src/bundler.rs
@@ -227,8 +227,6 @@ impl Bundler {
     &mut self,
     scan_stage_output: NormalizedScanStageOutput,
   ) -> BuildResult<BundleOutput> {
-    let mut output = self.bundle_up(scan_stage_output, /* is_write */ true).await?;
-
     let dist_dir = self.options.cwd.join(&self.options.out_dir);
 
     if self.options.clean_dir && self.options.dir.is_some() {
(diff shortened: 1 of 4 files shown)


IWANABETHATGUY (dismissed):

hyf0 (approved):
LGTM @IWANABETHATGUY cc

IWANABETHATGUY (approved):

aprosail:
Already update with the main branch, but:

This test ([here](https://github.com/rolldown/rolldown/blob/6327abc53c45bc49d77c29c2241e643fb842d881/packages/rolldown/tests/fixture.test.ts#L18)), always takes a long time, which usually cause timeout that fail the tests. I've specified the timeout from 20s by default to 30s, but it's still not enough. So I modified it to 60s in the latest commit. Sorry that I don't have enough time to dive into the massive code in this repo to figure out why such tests takes so long time. But I guess perhaps there's some bug about loop somewhere in the codebase about those testing.

PR #6654, opened 2025-10-21
docs: set up redirects via netlify

TheAlexLichter:
Improve redirects so people won't get 404s for old pages.


hyf0 (approved):
Thank you! I didn't even notice this before.

PR #6698, opened 2025-10-26
fix(node/options): should validate `transform.jsx` correctly

ocavue:
Correct the `valibot` schema for option `jsx` to match the TypeScript type defined at [`transform-options.ts`](https://github.com/ocavue/rolldown/blob/058627d2fe17b2a729ff698a59c255c31df69ee2/packages/rolldown/src/options/transform-options.ts#L119)

Related: https://github.com/rolldown/tsdown/issues/562


hyf0 (approved):
Thank you!

PR #6746, opened 2025-10-29
feat: introduce stackblitz codeflow

iiio2:
Actually I've seen stackblitz codeflow in vite (https://github.com/vitejs/vite). That is why I thought if we can do similar thing in rolldown also.


hyf0:
Thank you for the contribution,  but I don't feel the need for this.

iiio2:
Ok @hyf0 , thanks.

PR #6782, opened 2025-10-30
fix(plugins): wrap replacePlugin with makeBuiltinPluginCallable

### LINKED ISSUE #473: rolldown replacePlugin doesn't work
### Describe the bug

builtin plugins without JS hooks are skipped in the transform loop. Rolldown replace plugin has no JS hooks when passed as a plain builtin, so it never runs.

https://github.com/nuxt/nuxt/issues/33602

### Reproduction

https://stackblitz.com/edit/vitejs-vite-ce9n83dt?file=package.json,vite.config.js,src%2Fmain.js&terminal=dev

### Steps to reproduce

- npm install && npm run dev
- open inspector, should log `true` instead of having `YOLO` being undefined

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    rolldown: ^1.0.0-beta.45 => 1.0.0-beta.45 
    vite: npm:rolldown-vite@latest => 7.1.12
```

### Used Package Manager

npm

### Logs

_No response_

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitejs/rolldown-vite/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitejs/rolldown-vite/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vite.dev/guide) and the [Rolldown-related guide](https://main.vite.dev/guide/rolldown.html).
- [x] Check that there isn't [already an issue](https://github.com/vitejs/rolldown-vite/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitejs/rolldown-vite/discussions) or join our [Discord Chat Server](https://chat.vite.dev/).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

### LINKED ISSUE #33602: Nuxt with rolldown does not hydrate component after SSR
### Environment

```
------------------------------
- Operating System: Windows_NT
- Node Version:     v24.10.0
- Nuxt Version:     4.1.3
- CLI Version:      3.29.3
- Nitro Version:    2.12.9
- Package Manager:  pnpm@10.18.3
- Builder:          -
- User Config:      modules, $development, components, imports, devtools, app, runtimeConfig, build, routeRules, sourcemap, experimental, compatibilityDate, nitro, vite, typescript, telemetry, eslint, i18n, image, pwa, security
- Runtime Modules:  @nuxt/eslint@1.9.0, @nuxt/image@1.11.0, @nuxt/test-utils/module@3.20.1, @nuxtjs/i18n@10.1.2, @unocss/nuxt@66.5.4, nuxt-security@2.4.0, reka-ui/nuxt@2.6.0
- Build Modules:    -
------------------------------
```

### Reproduction

[Rolldown Reproduction](https://stackblitz.com/edit/github-fqfxufpw?file=app%2Fapp.vue)
1. Review console
2. Server outputs as expected
3. Client has no output and `ClientOnly` component is not visible.

[Working Rollup Reproduction](https://stackblitz.com/edit/github-fqfxufpw-jronfptz?file=app%2Fapp.vue)
1. Review console
2. Server outputs as expected.
3. Client has output and `ClientOnly` component has rendered.

### Describe the bug

It appears that when loading a page from SSR it does not perform hydration.

In the reproduction I added a number of logs to determine the current steps, from the server you can see the logs being outputted but on the client there is no output.

It also looks like the `ClientOnly` component is impacted as a result of `import.meta.client` not being replaced.

This only started occurring during 4.1.3+ for rolldown. It's possible it could be #33258 where the replace plugin was replaced with rolldown's. 

### Additional context

Rendered Rolldown Output
Image

Rendered Rollup Output
Image

### Logs

```shell-script

```

huang-julien:
fix https://github.com/vitejs/rolldown-vite/issues/473
fix https://github.com/nuxt/nuxt/issues/33602

This PR makes the replace plugin callable from rolldown vite

### DIFF
--- a/packages/rolldown/src/builtin-plugin/replace-plugin.ts
+++ b/packages/rolldown/src/builtin-plugin/replace-plugin.ts
@@ -1,5 +1,5 @@
 import type { BindingReplacePluginConfig } from '../binding.cjs';
-import { BuiltinPlugin } from './utils';
+import { BuiltinPlugin, makeBuiltinPluginCallable } from './utils';
 
 /**
  * Replaces targeted strings in files while bundling.
@@ -35,5 +35,7 @@ export function replacePlugin(
     }
   });
 
-  return new BuiltinPlugin('builtin:replace', { ...options, values });
+  return makeBuiltinPluginCallable(
+    new BuiltinPlugin('builtin:replace', { ...options, values }),
+  );
 }


sapphi-red:
@shulaoda Given that the replace plugin can never use `transform_ast` hook, I guess this is safe to do. What do you think?

shulaoda:
> @shulaoda Given that the replace plugin can never use `transform_ast` hook, I guess this is safe to do. What do you think?

Yes. I’m also thinking about how we should handle the plugins we expose from Rolldown when running rolldown-vite in `dev`.
Specifically, what should we do with plugins that use `transform_ast` or `render_chunk`..? Should we add a constraint so that such plugins are only allowed in full bundle mode and during build?

shulaoda (approved):

sapphi-red:
> > @shulaoda Given that the replace plugin can never use `transform_ast` hook, I guess this is safe to do. What do you think?
> 
> Yes. I’m also thinking about how we should handle the plugins we expose from Rolldown when running rolldown-vite in `dev`. Specifically, what should we do with plugins that use `transform_ast` or `render_chunk`..? Should we add a constraint so that such plugins are only allowed in full bundle mode and during build?

I think it's fine to say it's only supported in full bundle mode & build. But it would be better to output a warning message if it's used in dev.

TheAlexLichter:
@sapphi-red What would be the recommendation when used in unbundled dev mode? Fall back to the rollup plugin like in https://github.com/nuxt/nuxt/pull/33615?

sapphi-red:
@TheAlexLichter Yes, that would be the recommendation. The plugins listed in https://rolldown.rs/builtin-plugins/ are the only ones expected to be used by external projects. So with this PR, there won't be any cases that requires doing so though.

shulaoda:
Thank you for your contribution! ❤

PR #6786, opened 2025-10-31
docs: add minification preservation instructions for banner/footer options

ZiuChen:
Addresses https://github.com/vitejs/rolldown-vite/issues/459 where users discovered their configured `banner` and `footer` content was stripped during minification.

## Changes

Added **In-depth** sections to `banner` and `footer` documentation explaining legal comment preservation:

- **Comment formats that survive minification:**
  - `/*!` prefix: `/*! My banner */`
  - `@license` keyword: `/* @license My banner */`
  - `@preserve` keyword: `/* @preserve My banner */`
  - `//!` prefix for single-line comments

- **Minification example** showing proper usage with `minify: true`

- **Cross-reference** to `output.legalComments` option (defaults to `'inline'`)

## Example

```js
export default {
  output: {
    banner: '/*! My Library v1.0.0 | MIT License */',
    minify: true,  // Banner persists due to /*! syntax
  },
};
```

## Screenshot

image

The table of contents now shows the new "In-depth" and "Example with minification" subsections under both `banner` and `footer` options.

### DIFF
--- a/docs/options/output.md
+++ b/docs/options/output.md
@@ -167,6 +167,19 @@ export default {
 };
 ```
 
+:::warning
+
+When using `output.banner` or `output.footer` with minification enabled, the banner content may be stripped out unless it is formatted as a legal comment. To ensure your banner persists through minification, use one of these comment formats:
+
+- Comments starting with `/*!` (e.g., `/*! My banner */`)
+- Comments containing `@license` (e.g., `/* @license My banner */`)
+- Comments containing `@preserve` (e.g., `/* @preserve My banner */`)
+- Comments starting with `//!` (for single-line comments)
+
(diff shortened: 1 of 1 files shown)


hyf0:
I think current rolldown's behavior is wrong. We might want to solve it in rolldown's side.

hyf0:
> I think current rolldown's behavior is wrong. We might want to solve it in rolldown's side.

Just saw https://github.com/vitejs/rolldown-vite/issues/459#issuecomment-3441031606. Adding document is a good move.

hyf0:
I still want to check esbuild's behavior in relation to this.

hyf0 (approved):
Thank you!

PR #6838, opened 2025-11-04
docs: generate api reference

sxzz:


hyf0:
@copilot Why this PR deosn't trigger [netlify](https://github.com/apps/netlify) preview CI?

sxzz:
I squashed the commit and now triggered.

hyf0:
For `Bundler API`, I think we should put well-organized content, but we currently don't have time to do it. So having something is better than having nothing.

Suggestions:

- Still keep `Bundler API` page, but with a link that points to the `reference` page.
- Put `reference` under the `APIs` sidebar with the name `API Reference`
  - For the name, I think `API Reference` is a good name. If it was me, I'll try to emphasize this page actually shows all API, so I will call it `API Catelog` under `APIs`.
- I think we should keep `config` button in the nav bar, since it's the most frequent accessed content.

hyf0:
Ah I see. You are gonna remove the `options` page. Let me add some context here.

sxzz:
- `Bundler API` page is empty currently, so it's meaningless until someone can write a guidance, then we can have the page back.
- We can rename `Reference` to `API`, and full name is `API Reference` (Vue docs also did)
- Config on the navbar is weird, I think.

hyf0:
> Bundler API page is empty currently, so it's meaningless until someone can write a guidance, then we can have the page back.

Yes. We could do this.

> We can rename Reference to API, and full name is API Reference (Vue docs also did)

Hmm. I think having a `API` item under the sidebar `APIs` is a bit of weird.

If you put it only in the nav bar. I think people will confuse about the relationship between `API` nav bar button and sidebar `APIs` area. I personally think that shows we don't think clearly about how to organize the content.

> Config on the navbar is weird, I think.

This comes from vite's doc. https://vite.dev/config/

sxzz:
This is a bit different from Vite. Essentially, Config is just one item in the sidebar, whereas Vite dedicates several pages to explaining config in detail.

Also, I'm not sure what approach the Rolldown team wants to take for introducing config, whether we plan to provide both a reference and handwritten guidance, or choose just one style.

hyf0 (dismissed):
LGTM. Sill let's see if there are opinions from @TheAlexLichter.

docs/.vitepress/scripts/generate-reference.ts
```diff
@@ -0,0 +1,54 @@
...
+import type { PluginOptions } from 'typedoc-plugin-markdown';
+console.log('📚 Generating reference...');
+
+// Generate API documentation
+await runTypedoc();
+console.log('✅ Reference generated successfully!');
+console.log('📚 Beautifying reference structure...');
+
+await rm('reference/api/index.md', { force: true });
+await rm('reference/api/_media', { recursive: true, force: true });
```
mdong1909: Is typedoc supposed to create a `_media` file?

sxzz: It might be necessary if there are external documents involved. For more information, see: https://typedoc.org/documents/External_Documents.html

I have removed this line.

PR #6877, opened 2025-11-07
refactor(node): implement a new `Bundler` that satisfy the usage of `RolldownBuild`

hyf0:
This's a transitive refactor and still some things need to be done in later PRs. So

- watch mode is not affected because I isolated changes and only affect the `RolldownBuild` logic.
- dev mode is not affected too.
- `Bundler` in `crates/rolldown` is not touched to avoid too much changes.
- This PR doesn't seperate the bundler/build level data, but it's a preparation for that purpose.  

---

This PR introduces a `Build`​ concept to represent the process of doing a build. Having this abstraction allows us to implement different bundler for different usages.

In the previous code, `RolldownBuild`​ create a `Bundler`​  underlying for each `generate/write`​ call. Without mentioning potential errors, this abstraction is already wrong and give us buggy mental model when working aounrd these code.

With this PR, `RolldownBuild`​ create only a `Bundler`​, but multiple `builds`​ underlying for each `generate/write`​ call. This gives us a correct mental model when working with code under `rolldown`​ without needing to considering about the usage of `RolldownBuild`​.

### DIFF
--- a/crates/rolldown/src/build/build.rs
+++ b/crates/rolldown/src/build/build.rs
@@ -1,70 +1,45 @@
-use super::stages::{link_stage::LinkStage, scan_stage::NormalizedScanStageOutput};
-use crate::{
-  BundlerOptions, SharedOptions, SharedResolver,
-  bundler_builder::BundlerBuilder,
-  hmr::hmr_stage::{HmrStage, HmrStageInput},
+use crate::build::build_context::BuildContext;
+
+use super::super::{
+  SharedOptions, SharedResolver,
+  bundler::CacheGuard,
   module_loader::deferred_scan_data::defer_sync_scan_data,
   stages::{
     generate_stage::GenerateStage,
-    scan_stage::{ScanStage, ScanStageOutput},
+    link_stage::LinkStage,
+    scan_stage::{NormalizedScanStageOutput, ScanStage, ScanStageOutput},
   },
   types::{bundle_output::BundleOutput, scan_stage_cache::ScanStageCache},
   utils::fs_utils::clean_dir,
 };
-use anyhow::Result;
 use arcstr::ArcStr;
-use rolldown_common::{
-  ClientHmrInput, ClientHmrUpdate, GetLocalDbMut, HmrUpdate, Module, ScanMode, SharedFileEmitter,
-  SymbolRefDb,
-};
+use rolldown_common::{GetLocalDbMut, Module, ScanMode, SharedFileEmitter, SymbolRefDb};
 use rolldown_debug::{action, trace_action, trace_action_enabled};
 use rolldown_error::{BuildDiagnostic, BuildResult, Severity};
 use rolldown_fs::{FileSystem, OsFileSystem};
-use rolldown_plugin::{
-  __inner::SharedPluginable, HookBuildEndArgs, HookRenderErrorArgs, SharedPluginDriver,
-};
+use rolldown_plugin::{HookBuildEndArgs, HookRenderErrorArgs, SharedPluginDriver};
 use rolldown_utils::dashmap::FxDashSet;
-use rustc_hash::FxHashSet;
-use std::{
-  any::Any,
-  sync::{Arc, atomic::AtomicU32},
-};
+use std::sync::Arc;
 
-pub struct Bundler {
-  pub closed: bool,
+pub struct Build {
   pub(crate) fs: OsFileSystem,
   pub(crate) options: SharedOptions,
   pub(crate) resolver: SharedResolver,
   pub(crate) file_emitter: SharedFileEmitter,
   pub(crate) plugin_driver: SharedPluginDriver,
   pub(crate) warnings: Vec<BuildDiagnostic>,
-  pub(crate) _log_guard: Option<Box<dyn Any + Send>>,
   pub(crate) cache: ScanStageCache,
   pub(crate) session: rolldown_debug::Session,
 }
 
-impl Bundler {
-  // --- Public API ---
-
-  pub fn new(options: BundlerOptions) -> BuildResult<Self> {
-    BundlerBuilder::default().with_options(options).build()
-  }
-
-  pub fn with_plugins(
-    options: BundlerOptions,
-    plugins: Vec<SharedPluginable>,
-  ) -> BuildResult<Self> {
-    BundlerBuilder::default().with_options(options).with_plugins(plugins).build()
-  }
-
+impl Build {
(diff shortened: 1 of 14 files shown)


hyf0:
* **#6878** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/6878?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6877** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/6877?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/6877?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=6877).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

packages/rolldown/tests/error/error.test.ts
```diff
@@ -107,8 +107,12 @@ describe('Plugin closeBundle hook', async () => {
       await build.close();
       await build.write();
     } catch (error: any) {
-      expect(error.message).toMatch(
-        `Rolldown internal error: Bundle is already closed, no more calls to 'generate' or 'write' are allowed.`,
+      expect(error.message).toMatchInlineSnapshot(
+        `
```
hyf0, line 111: This's unexpected. But failed to find the reason from previous code. I intend to introduce a builtin error type for this situation intead of the current `UNHANDLEABLE_ERROR` or previous `Rolldown internal error`. They are both wrong.

In short, this will be fixed in the later PR.

packages/rolldown/src/cli/commands/bundle.ts
```diff
@@ -159,18 +159,18 @@ async function bundleInner(
   for (const config of configList) {
     const outputList = arraify(config.output || {});
     const build = await rolldown({ ...config, ...cliOptions.input });
-    for (const output of outputList) {
```
hyf0, line 162: After implemting correct `close` logic. This code raise errors because it tries to use the closed `RolldownBuild` to generate output.

packages/rolldown/src/api/rolldown/rolldown-build.ts
```diff
@@ -57,61 +45,71 @@ export class RolldownBuild {
...
+    this.#stopWorkers = void 0;
   }
 
   async [Symbol.asyncDispose](): Promise<void> {
     await this.close();
   }
 
-  // TODO(shulaoda)
-  // The `watchFiles` method returns a promise, but Rollup does not.
-  // Converting it to a synchronous API might cause a deadlock if the user calls `write` and `watchFiles` simultaneously.
```
sapphi-red, line 112: I think we should keep this comment.

crates/rolldown_binding/src/binding_bundler.rs
```diff
@@ -1,47 +1,225 @@
...
+          Ok(napi::Either::A(BindingErrors::new(errors)))
+        }
+      }
+    };
+    env.spawn_future(fut)
+  }
+
+  #[napi]
+  /// - `.close()` requires acquiring `&mut self`
+  /// - Acquiring `&mut self` in async napi `fn` is unsafe, so we must use a sync `fn` here.
```
sapphi-red: This comment applies to `generate` and `write` too. Maybe put this at the top of the `impl` block?

crates/rolldown/src/build/build.rs
```diff
@@ -0,0 +1,345 @@
...
+  pub(crate) file_emitter: SharedFileEmitter,
+  pub(crate) plugin_driver: SharedPluginDriver,
+  pub(crate) warnings: Vec<BuildDiagnostic>,
+  pub(crate) cache: ScanStageCache,
+  pub(crate) session: rolldown_debug::Session,
+}
+
+impl Build {
+  #[tracing::instrument(level = "debug", skip_all, parent = &self.session.span)]
+  /// This method intentionally to get the ownership of `self` to show that the method can be called multiple times.
```
sapphi-red:
```suggestion
  /// This method intentionally get the ownership of `self` to show that the method cannot be called multiple times.
```
(other places as well)

packages/rolldown/src/api/rolldown/rolldown-build.ts
hyf0: We're actually available to make it become sync, but I don't want to introduce a breaking change here.

crates/rolldown_binding/src/binding_bundler.rs
hyf0: Not really. `Bundler` in `rolldown_binding` doesn't have a `generate`/`write` method. `BindingBundler` will consume `Build` directly.

packages/rolldown/src/api/rolldown/rolldown-build.ts
hyf0: Done.

crates/rolldown_binding/src/binding_bundler.rs
hyf0: I think we should come up a different concept for `Bundler` in `rolldown_binding`. It's too confusing.

sapphi-red: `BindingBundler::generate` and `BindingBundler::write` takes `&mut self` and uses a sync fn + `env.spawn_future`. What is this different from `close`?

hyf0: It's not different. But comments added here is as a complement to https://github.com/rolldown/rolldown/blob/9c5af2f8cfc835b0119390c61344ac59541b22ae/crates/rolldown_binding/src/bundler.rs#L64-L73

if it's just these comments alone, I probably won't add them in the first place, becuase I think there're natural things.

`BindingBundler::generate/write` won't cause `Build#generate/write` being written in a non-intuitive way.

> `.close()` requires acquiring `&mut self`

means `Bundler.close()` not `BindingBundler.close()`

sapphi-red: Ah, I see. I think changing `` `.close()` requires acquiring `&mut self` `` to `` `Bundler::close()` requires acquiring `&mut self` `` would be helpful.

hyf0: I updated the comment to make it more clear.

sapphi-red (approved):

PR #6890, opened 2025-11-08
feat: port getLogFilter helper from Rollup for advanced log filtering

### LINKED ISSUE #5848: [Feature Request]: Missing `getLogFilter` helper for applying advanced log filters

Hi, it seems that the `getLogFilter` helper function is missing in Rolldown.

Official [`docs`](https://rollupjs.org/javascript-api/#applying-advanced-log-filters) from Rollup.

Same as Rollup existing API:

```js
// rollup.config.mjs
import { getLogFilter } from 'rollup/getLogFilter';

const logFilter = getLogFilter(['code:FOO', 'code:BAR']);

export default {
	input: 'main.js',
	output: { format: 'es' },
	onLog(level, log, handler) {
		if (logFilter(log)) {
			handler(level, log);
		}
	}
};
```

taearls:
Closes #5848 

Implement the missing `getLogFilter` helper function to achieve full Rollup API compatibility for advanced log filtering.

The helper function supports:
- Simple property matching (code:FOO)
- Wildcard patterns (code:*A, code:B*, code:*C*)
- Inverted filters (!code:FOO)
- AND conditions (code:FOO&plugin:bar)
- OR conditions (multiple filters in array)
- Nested properties (foo.bar:baz)
- Object and number matching

Usage:
```javascript
import getLogFilter from 'rolldown/getLogFilter';

const logFilter = getLogFilter(['code:FOO', 'code:BAR']);

export default {
  onLog(level, log, handler) {
    if (logFilter(log)) {
      handler(level, log);
    }
  }
};
```

Implementation details:
- Single file implementation in src/get-log-filter.ts
- Comprehensive test suite with 9 tests covering all edge cases


packages/rolldown/package.json
```diff
@@ -58,6 +58,10 @@
       "dev": "./src/parse-ast-index.ts",
       "default": "./dist/parse-ast-index.mjs"
     },
+    "./getLogFilter": {
```
hyf0: Does rollup expose it this from here too?

taearls: yeah. the only difference is they export a CJS version too:

https://github.com/rollup/rollup/blob/master/package.json#L243C5-L247C7

hyf0:
I'm hesitating because rolldown doesn't expose much error details as rollup did, so I'm thinking the value of having this. @sapphi-red cc

sapphi-red:
I think there aren't much downside to have this (the upside is rollup compat).

As a side note, we probably need to add a hook filter to `onLog` as we have to call `onLog` for all logs (we filter out the info messages with `logLevel` beforehand, but `onLog` can raise the severity).

taearls:
> I think there aren't much downside to have this (the upside is rollup compat).
> 
> As a side note, we probably need to add a hook filter to `onLog` as we have to call `onLog` for all logs (we filter out the info messages with `logLevel` beforehand, but `onLog` can raise the severity).

Hi! I'm not sure I understand the ask. Does `onLog` need to be called from within `getLogFilter`?

hyf0 (approved):
LGTM. See if others wanna do some tweaks.

sapphi-red (approved):
LGTM

For attribution, it's worth noting that much of the implementation is derived from Rollup.

PR #6891, opened 2025-11-08
refactor(rust): introduce `PluginDriverFactory` to manage creation of `PluginDriver`

hyf0:
Follow-up of https://github.com/rolldown/rolldown/pull/6883. We will use `PluginDriverFactory` to manage bundler level and bundle level data separately.

With this PR, we're finally able to find and solve potential issues like https://github.com/rolldown/rolldown/pull/6891/files#r2507078414. The new design might encounter many issues, but that's expected. These issues are caused because we don't treat bundler and build/bundle level data properly. It also makes complicated features like `incremental build/hmr/watch mode` vulnerable.

PR #6894, opened 2025-11-08
feat(rust): use `BundleMode` to handle incremental build exhaustively

hyf0:
Several reasons I want to introduce `BundleMode`:

## A mental model footgun

`BundleMode` is used to emphasize there are 3 states when we integrate incremental build into rolldown.

For exmaple, for watch mode:

1. if `incremental_build: false`, file changes should result to full rebuild every time
2. if `incremental_build: true`, the first build is not a plain full build, it needs to be done with saving data into cache.
3. if `incremental_build: true`, files changes should result to incremental rebuilds.

The footgun:

https://github.com/rolldown/rolldown/blob/c19666e8309871688d997729d32f49108604249a/crates/rolldown/src/watch/watcher_task.rs#L53-L67

When `incremental_build: false`, the watch mode will use `ScanMode:Full` in order to trigger full build, which seems to satisfy the above `1.`, but it's not!

> 1. if `incremental_build: false`, file changes should result to full rebuild every time

`watch mode` triggers a full build with extra unnecessary overhead of saving cache! And for each file change.

This PR doesn't fix the behavior but make it clear to be recognized. See https://github.com/rolldown/rolldown/pull/6894/files#r2507176596.

## Reduce bug prone patterns:

- https://github.com/rolldown/rolldown/pull/6894/files#r2507167806
- https://github.com/rolldown/rolldown/pull/6894/files#r2507174518

## Misc

Inside of the `Bundle` logic, we're using `ScanMode` + `is_incremental_build_enabled` to achieve the same effect as `BundleMode`. I restrict the refactor area of this PR, so they are not touched.

The previous design for the bundler hides many issues and cause advance features vulnerable. https://github.com/rolldown/rolldown/pull/6891#discussion_r2507078414 probably won't be the last issue.

### DIFF
--- a/crates/rolldown/src/bundle/bundle_factory.rs
+++ b/crates/rolldown/src/bundle/bundle_factory.rs
@@ -1,7 +1,8 @@
 use std::{any::Any, sync::Arc};
 
 use rolldown_common::{
-  BundlerOptions, FileEmitter, NormalizedBundlerOptions, SharedFileEmitter, SharedModuleInfoDashMap,
+  BundleMode, BundlerOptions, FileEmitter, NormalizedBundlerOptions, SharedFileEmitter,
+  SharedModuleInfoDashMap,
 };
 use rolldown_error::{BuildDiagnostic, BuildResult, EventKindSwitcher};
 use rolldown_fs::OsFileSystem;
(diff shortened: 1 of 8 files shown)


hyf0:
* **#6896** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/6896?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6894** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/6894?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/6894?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6893** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/6893?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6892** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/6892?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6891** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/6891?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=6894).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown/src/bundle/bundle_factory.rs
```diff
@@ -90,39 +91,34 @@ impl BundleFactory {
     ))
   }
 
-  pub fn create_bundle(&mut self) -> Bundle {
+  pub fn create_bundle(
```
hyf0, line 94: I removed the second `create_incremental_bundle` method, and merge the logic into one method, becuase  I found the previous pattern is bug-prone and hard to track the control flow.

It also causes a bug.

```rust
let bundler = Bundler::new(..) // with `incremental: true`

bundler.generate()
bundler.incremental_generate(["main.js"])
```

In the above code, `incremental_generate` will panic because of `generate` will call `create_bundle` instead of `create_incremental_bundle`, so it doesn't create build cache.

crates/rolldown/src/bundle/bundle_factory.rs
```diff
@@ -90,39 +91,34 @@ impl BundleFactory {
...
-    bundle
-  }
 
-  pub fn create_incremental_bundle(&mut self, cache: ScanStageCache) -> Bundle {
-    let bundle_span = self.generate_unique_bundle_span();
-
-    // Reuse module infos for incremental bundle
-    // TODO: hyf0 or should we reset it?
-    let module_infos = Arc::clone(&self.module_infos_for_incremental_build);
+    let module_infos = if bundle_mode.is_full_build() {
```
hyf0: Another footgun: `FullBuild` and `IncrementalFullBuild` are all requires to clear the `module_infos_for_incremental_build`, but the previous `create_incremental_bundle` doesn't handle the `IncrementalFullBuild` situation.

In the scenario of `dev mode`, dev engine might trigger a second `IncrementalFullBuild` if there are too many file changes. In this case, the previous code will mix up the module infos between two builds.

crates/rolldown/src/watch/watcher_task.rs
```diff
@@ -62,15 +61,21 @@ impl WatcherTask {
...
 
     let result = {
+      let is_incremental = bundler.options.experimental.is_incremental_build_enabled();
+
       let scan_mode = if is_incremental && !changed_files.is_empty() {
         rolldown_common::ScanMode::Partial(changed_files.to_vec())
       } else {
         rolldown_common::ScanMode::Full
       };
+      let bundle_mode = match scan_mode {
```
hyf0, line 71: I don't fix the behavior becuase it requires a quite change on related code.

IWANABETHATGUY (approved):

PR #6948, opened 2025-11-12
chore: enable trust policy

iiio2:
Since we are using pnpm 10.21, it has a very good security feature for enabling trust policy. i think maybe we can use it here.
https://pnpm.io/settings#trustpolicy


hyf0 (approved):
LGTM. Thank you!

sapphi-red (approved):

iiio2:
Thank you @sapphi-red , @hyf0 , @Boshen . :blush:

PR #7002, opened 2025-11-14
fix: remove unused module namespace object exporting 

### LINKED ISSUE #6992: undefined named exports are being added
We are experimenting with migrating the Sentry JS SDK codebase from rollup to rolldown, and while it works for the most part, a common issue we are seeing is a named export is getting added without being defined.

You can find one of those [failed tests here](https://github.com/getsentry/sentry-javascript/actions/runs/19328410843/job/55286081012?pr=18156#step:14:113) failing for sveltekit bundling.

For now, these are my observations from taking a look at both rolldown and rollup outputs and linked possibly related issues:

- Unnecessary variables declarations: Some variables get declared but are never used (e.g: `index_server_exports`) https://github.com/rolldown/rolldown/issues/6017 https://github.com/rolldown/rolldown/issues/6158
- Those don't then get exported, but variables with similar names get exported instead, but they are undefined so they break builds.

**server/index.mjs**
```js
// Stuff...

// ⚠️ Was never defined by user code or rolldown, crashes builds
export { server_exports };
```

**index.server.mjs**
```js
// ⚠️ That export isn't defined
import { server_exports } from './server/index.mjs';


// ⚠️ This gets declared, never used
//#region src/index.server.ts
var index_server_exports = /* @__PURE__ */ __export({
  //...
});
//#endregion
```

For our configurations, trying out `preserveEntrySignatures="allow-extension"` or `strict` has no effect, we also set `preserveModules` to `true` and `preserveModulesRoot` to `src`, changing any of those has no effect.

I investigated a little bit and noticed that the second re-export is the one that breaks it, this is because we are working around a Vite issue that puts those exports at the default export rather than named ones.

```ts
// server/index.mjs
export {
  addBreadcrumb,
  addEventProcessor,
  // also everything else
} from '@sentry/node';

// This re-export breaks it!!!
export * from '@sentry/node';
```

Removing the re-export or using `export type * ...` instead works and produces a valid bundle.

I know the Sentry codebase is too big to pinpoint the issue with all of these custom configurations, so here is a minimal example that demonstrates the issue:

https://repl.rolldown.rs/#eNrdU0Fu2zAQ/MqaFyWFI7mHXlS4KArkWLRIeyt7oKVVwkQihSVlOxD09yxFKbbjJA+IIEDC7M5wOYPtRSXyXmhT4j51SFuk1LsAGZGfwUtRMIr71pKHT1CRbSBJs9iQfJWGO1DknjocloJsXZd2Z9LCmkrfHum+UonSuhml/WOL0MPN1Par9doaB8N04kxPpJFmmqbESnW1h4seVKthuMzP6Otv0EsDQOg7MvEfoCUc5782nh7/6FujuIouh0TVtd1d4d6jcSyQLMNxgWI733Y+nxUOGj9t2dWBGxzg9lerN9YyVwpHhRTPTZWlRjGeoGv4XgEapmKYgIyqc/g3d0vx3WEYODO2RCki/j98Bmn4lSbL4K9ueRa2Ix2DjFSA9XoNaqe0h2j3xcHQy5EXn8Bjc5F0w0ep+h1edtwXRMQUf9fOEd+fhH+Cx+jfTvKN4M5S+HgWxsXK4hoe1uclfLKWox2qLH8QqrKgrtmMHjByvWXx32QLdM4So/M+HRsRtjje/E472LDIgwPtF4vFc0Dz4p/SxMADb3mOz+kqXV1t0Kv0y0oMT8//dCA=

Is this a bug or is it something we should be aware of when using rolldown? re-exporting should be fine in esm AFAIK

IWANABETHATGUY:
**Isuses**
- Before, when hmr is enabled, every `module_namespace`​ export object is generated but some of them are not included in `used_symbol_refs`​.
- Before, some module namespace object is not generated but included in `used_symbol_refs`. (e.g. indirect external module reexport)export * from "@sentry/node"
￼

This pr syncs `module_namespace` usage before code generation(both module level and chunk level), so that we could use `used_symbol_ref` to determine if a module namespace ref is used or not.

Closed #6992

### DIFF
--- a/crates/rolldown/src/module_finalizers/mod.rs
+++ b/crates/rolldown/src/module_finalizers/mod.rs
@@ -418,22 +418,7 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
   }
 
   fn generate_declaration_of_module_namespace_object(&self) -> Vec<ast::Statement<'ast>> {
-    let module_namespace_included_reason = self.ctx.linking_info.module_namespace_included_reason;
-    let is_namespace_referenced = matches!(self.ctx.module.exports_kind, ExportsKind::Esm)
-      && if module_namespace_included_reason.contains(ModuleNamespaceIncludedReason::Unknown) {
-        true
-      } else if module_namespace_included_reason
-        .contains(ModuleNamespaceIncludedReason::ReExportExternalModule)
-      {
-        // If the module namespace is only used to reexport external module,
-        // then we need to ensure if it is still has dynamic exports after flatten entry level
-        // external module, see `find_entry_level_external_module`
-        self.ctx.linking_info.has_dynamic_exports
-      } else {
-        false
-      };
-
-    if !is_namespace_referenced {
+    if !self.module_namespace_included {
       return vec![];
     }
 
--- a/crates/rolldown/src/stages/generate_stage/code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/code_splitting.rs
@@ -451,6 +451,54 @@ impl GenerateStage<'_> {
     }
   }
 
+  /// Finalizes which module namespace objects should be included in the output bundle.
+  ///
+  /// This method determines whether each ESM module's namespace object (e.g., `import * as ns from './module'`)
+  /// should be kept in the final bundle based on how it's used. This is an optimization to avoid
+  /// generating unnecessary namespace objects that aren't actually referenced.
+  ///
+  /// The decision is based on three cases:
+  /// 1. **Unknown usage**: Keep the namespace (conservative approach when we can't prove it's unused)
+  /// 2. **Re-exporting external modules**: Keep only if the module has dynamic exports after flattening
+  ///    entry-level external modules (see [`find_entry_level_external_module`]), this is used for
+  ///    indirect external module re-exports optimization.
+  /// 3. **All other cases**: Remove the namespace object
+  ///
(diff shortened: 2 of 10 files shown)


IWANABETHATGUY:
* **#7002** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7002?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7002?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=7002).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyf0:
Feel like we should should remove `used_symbol_refs` and use `statementInfo` to check if a symbol is included. I also encountered several bugs that a symbol showed in `used_symbol_refs` and the corresponding `statementInfo` isn't get included.

hyf0 (approved):

hyf0:
> Closed https://github.com/rolldown/rolldown/pull/6922

is not an issue? It it a correct link?

IWANABETHATGUY:
> > Closed #6922
> 
> is not an issue? It it a correct link?

Fixed

PR #7062, opened 2025-11-16
docs: update top level navigation

mdong1909:
Reorganizes the navigation bar and includes activeMatch for readers to more easily see where they are.

- Separated In-depth as standalone submenu and renamed to `Concepts`
- Removed inconsistent active status
- Turned `Docs` into a nested menu with Guide, Glossary, and Concepts
- Moved `Contribute` under Resources
- Removed social links from Resources
- Renamed `Config` to References

This does NOT make any changes to `References`. This will be done in another PR that auto-generates documentation based typing


hyf0:
> Renamed Config to References

I previewed the doc locally. I felt this's a negative change. `References` doesn't give people a clear intention for those want to inspect documentation about options, which should be the most frequent intention.

hyf0:
> > Renamed Config to References
> 
> I previewed the doc locally. I felt this's a negative change. `References` doesn't give people a clear intention for those want to inspect documentation about options, which should be the most frequent intention.

IMO, we should keep calling it `Config` or `Options` in the navbar area.

hyf0:
Think again, I feel `References` or `API` are more proper to be used to carry auto-gen docs made in https://github.com/rolldown/rolldown/pull/6838. It means something deeper and is for those want to get techenical.

TheAlexLichter:
> Think again, I feel `References` or `API` are more proper to be used to carry auto-gen docs made in #6838. It means something deeper and is for those want to get techenical.

API is tricky as it could be ambiguous for Rolldown.

From @sapphi-red:

> API feels strange here because I think normally people won’t understand “options” as part of the API (even though technically it is). If it were called “API”, I would expect a reference for the programmable interface (the way to use Rolldown from scripts).

mdong1909:
Closing as moved over to: https://github.com/rolldown/rolldown/pull/7092

PR #7234, opened 2025-11-27
fix(cjs): prevent duplicate require declarations for external modules with preserveModules

### LINKED ISSUE #7233: [Bug] CJS Bundles redeclaring duplicate variable names
I tried this on the latest pkgr release to test out the fix to #7115 in #7218 

```
https://pkg.pr.new/rolldown@f4aaab9
```

While it works for simple chains, I tried a longer export chain with multiple re-exports along the path. If exports name the same module twice, the resultant bundle is invalid as it would declare it twice, once as a `let` variable and once as a `var` which would break the bundle.

Again this is with `preserveModules: true`

### Exact Case

We do a combination of start exports along with named exports, the main reason is the behavior that exists in rolldown/rollup that automatically nests all exported names under `default`.

To simplify the scenario, Assume an external named `zod`, export transitively from a module, once as named exports and another as a star exports.

```ts
export { string, object } from 'zod';

// Server: re-exports from external library
export * from 'zod';
```

**Expected Result:** Either deconflict the names by giving each one a unique name or let the star export override the named ones, this is valid ES module stuff.

**Actual Result:** Two variables declared, causing conflicts and a redeclaration error.

```ts
const require_rolldown_runtime = require('./_virtual/rolldown_runtime.js');
require('./server.js');

// first declaration, maps to the named exports
let zod = require("zod");

// ...

// Second declaration, maps to the star exports, breaks the bundle
var zod = require("zod");
Object.keys(zod).forEach(function (k) {
    enumerable: true,
    get: function () { return zod[k]; }
  });
});
```

### Reproduction

I got a simple reproduction here, with just two files:

https://stackblitz.com/edit/c5u9nqeg?file=package.json,index.ts,server.ts,rolldown.config.ts,dist%2Findex.js

Note that it also happens if done at different points of the export chain, which makes it hard to work around, here is an example of a chain:

https://stackblitz.com/edit/c5u9nqeg-rmmyl3p5?file=package.json,index.ts,server.ts,rolldown.config.ts,middle.ts,dist%2Findex.js

Also every file in the chain will contain the two variable declaration, causing each file to be invalid as well.

--- 

Workarounds, we could just remove all named exports we have that get rexported by a start in the same file, but hard to do/guarantee that for long chains.

logaretm:
When using CJS format with `preserveModules`, files that do both named exports AND star exports from the same external module would generate duplicate `require()` statements: one as let and one as var. breaking the bundle with a redeclaration error.

The deduplication logic was only checking `direct_imports_from_external_modules` but missing `import_symbol_from_external_modules`. 

I added a `.chain()` call to check both lists so it actually knows the module was already imported and skips the duplicate declaration.

Feel free to close this PR if the approach is incorrect, just figured to give it a go.

closes #7233

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -180,3 +180,6 @@ crates/rolldown_devtools_action/bindings/**
 # AI config
 .claude
 CLAUDE.local.md
+
+# workaround for https://github.com/anthropics/claude-code/issues/4928
+nul
--- a/crates/rolldown/src/utils/chunk/render_chunk_exports.rs
+++ b/crates/rolldown/src/utils/chunk/render_chunk_exports.rs
@@ -217,6 +217,12 @@ pub fn render_chunk_exports(
                 .expect("Should be external module here");
               external.namespace_ref
             })
+            .chain(ctx.chunk.import_symbol_from_external_modules.iter().map(|idx| {
+              let external = &ctx.link_output.module_table[*idx]
+                .as_external()
+                .expect("Should be external module here");
+              external.namespace_ref
+            }))
             .collect();
           external_modules.iter().for_each(|idx| {
           let external = &ctx.link_output.module_table[*idx].as_external().expect("Should be external module here");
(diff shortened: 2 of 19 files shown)


IWANABETHATGUY:
You don't need to write `mod.rs` since those tests can be written with a static JSON config.

logaretm:
Got it, I reverted the `mod.rs` changes 👍

IWANABETHATGUY (approved):

PR #7364, opened 2025-12-07
feat(plugin): collect plugin hook execution timings

shulaoda:
When enabled, Rolldown measures time spent in each plugin hook. If plugins significantly impact build performance, a warning is emitted with a breakdown of plugin timings.

> [!NOTE]
> For hooks using `ctx.resolve()` or `ctx.load()`, the reported time includes waiting for other plugins, which may overestimate that plugin's actual cost.

image


shulaoda:
* **#7428** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7428?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7427** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7427?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7364** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7364?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7364?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=7364).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

sapphi-red (commented):
Have you measured the overhead?

shulaoda:
> Have you measured the overhead?

Yes, according to my tests, the performance impact is negligible. I didn’t observe any measurable overhead.

packages/rolldown/src/options/input-options.ts
```diff
@@ -314,6 +314,20 @@ export interface InputOptions {
...
+     * and emit a warning if plugins are significantly impacting build performance.
+     * The warning lists slow plugins with their percentage of total plugin time.
+     *
+     * > [!NOTE]
+     * > For hooks using `ctx.resolve()` or `ctx.load()`, the reported time includes
+     * > waiting for other plugins, which may overestimate that plugin's actual cost.
+     *
+     * @default false
+     */
+    warnSlowPlugins?: boolean;
```
sapphi-red: I think we can remove this option and solely rely on `checks.slowPlugins`.

crates/rolldown/src/bundle/bundle.rs
```diff
@@ -49,13 +50,16 @@ impl Bundle {
       trace_action!(action::BuildEnd { action: "BuildEnd" });
       ret
     }
-    .await
+    .await;
+    self.plugin_driver.set_total_build_time(start);
+    self.emit_slow_plugins_warning(result)
```
sapphi-red: Can we have a function like this so that we don't forget to call all three functions?
```rust
  pub async fn wrap_timing<T, F: Future<Output = T>>(&self, cb: impl FnOnce() -> F) -> T {
    let start = self.start_timing();
    let result = cb().await;
    self.set_total_build_time(start);
    self.emit_slow_plugins_warning(result)
    result
  }
```
(probably need to move things around to make the ownership checker happy)

crates/rolldown/src/bundle/bundle.rs
```diff
@@ -342,4 +350,34 @@ impl Bundle {
       });
     }
   }
+
+  /// Emit warning if plugins are slow
```
sapphi-red: Would you add a comment here what the threshold is?

crates/rolldown/src/bundle/bundle.rs
shulaoda: I ran into the issue below. Maybe we could use a macro for this? These parts only exist in `write` and `generate`, so perhaps we don’t need to worry about it too much?
image

packages/rolldown/src/options/input-options.ts
shulaoda: Does this mean we should enable it by default?

shulaoda: I’ll keep it disabled by default for now, since it seems to have an impact on CI.

sapphi-red: > it seems to have an impact on CI.

What kind of impact does it have?

crates/rolldown/src/bundle/bundle.rs
sapphi-red: That is the reason I said "(probably need to move things around to make the ownership checker happy)". But yeah, probably ok.

Would you add `#[must_use]` to `start_timing` so that we at least don't forget to call `record_timing` or `set_total_build_time`?

packages/rolldown/src/options/input-options.ts
shulaoda: > What kind of impact does it have?

We can update its snapshot, but I’m not sure whether it’s stable on CI or whether every build will consistently include the slow plugins.

sapphi-red: I see. I think we can disable the warning in our tests.

sapphi-red (commented):
LGTM, but I'd like another member to take a look

hyf0 (approved):

PR #7544, opened 2025-12-17
fix(dev/hmr): ensure cjs modules with no exports reference correct `module` identifier

leegeunhyeok:
## Description

Fixes an issue where CommonJS modules with no exports reference an incorrect `module` identifier when registering with the HMR runtime.

**Problem**

[StackBlitz](https://stackblitz.com/edit/github-qel99geu)

When CommonJS modules without exports are registered with the HMR runtime, they reference the wrong `module` identifier:

```js
// other_1.js
console.log('other_1 loaded');

// main.js
require('./other_1.js');
```

```js
var require_other_1 = /* @__PURE__ */ __commonJSMin(() => {
  const other_1_hot = __rolldown_runtime__.createModuleHotContext("src/other_1.js");
  __rolldown_runtime__.registerModule("src/other_1.js", module); // Invalid `module` identifier
                                                          ^^
  console.log("other_1 loaded");
});
```

In Node.js, this references the global `module` object in CommonJS environment. In runtime environments like React Native, the `module` identifier doesn't exist, causing an error.

Even if a `module` identifier exists, CJS modules with no exports always share the same export reference, leading to incorrect behavior:

```js
const mod0 = __rolldown_runtime__.loadExports('src/other_1.js');
const mod1 = __rolldown_runtime__.loadExports('src/other_2.js');

// Different modules but identical `exports` reference
mod0 === mod1; // true
```

---

**After this changes**

References the `module` identifier from `__commonJSMin` scope

```js
var require_other_1 = /* @__PURE__ */ __commonJSMin((exports, module) => {
                                                                ^^
  const other_1_hot = __rolldown_runtime__.createModuleHotContext("src/other_1.js");
  __rolldown_runtime__.registerModule("src/other_1.js", module);
                                                          ^^
  console.log("other_1 loaded");
});
```

### DIFF
--- a/crates/rolldown/src/ast_scanner/mod.rs
+++ b/crates/rolldown/src/ast_scanner/mod.rs
@@ -370,7 +370,9 @@ impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
     // Following cases need to be registered:
     // - `ExportsKind::None` (no exports)
     // - `ExportsKind::CommonJs` (commonjs module)
-    if self.immutable_ctx.options.is_dev_mode_enabled() && !exports_kind.is_esm() {
+    if self.immutable_ctx.options.is_dev_mode_enabled()
+      && matches!(exports_kind, ExportsKind::None | ExportsKind::CommonJs)
+    {
       self.result.ast_usage.insert(EcmaModuleAstUsage::ModuleRef);
     }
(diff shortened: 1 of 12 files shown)


hyf0 (approved):
Thank you!

Issue #7541, opened 2025-12-17
[Feature Request]: add INEFFECTIVE_DYNAMIC_IMPORT warning in core


sapphi-red:
### What problem does this feature solve?

It is implemented in `rolldown_plugin_vite_reporter`. But this is not specific to Vite, so it would be nice to be implemented in core.
https://github.com/rolldown/rolldown/blob/e966919c981c80824622ef8b34a4906b5d7c0728/crates/rolldown_plugin_vite_reporter/src/lib.rs#L122-L153

Tried copilot in https://github.com/rolldown/rolldown/pull/7530 but didn't have time to verify the code.

### What does the proposed API look like?

Not an API change

PR #7600, opened 2025-12-19
feat: add ineffective dynamic import warning

luke358:
This PR implements the changes described in #7541  
- Implement warning for dynamic imports that won't create separate chunks when a module is both statically and dynamically imported.
- Detect modules with both static and dynamic importers in same chunk and filter out HMR hot.accept to avoid false positives
- Add some cases covering various scenarios


crates/rolldown/tests/rolldown/warnings/ineffective_dynamic_import/basic/main.js
IWANABETHATGUY, line 1: This test case is redundant, as there already exists a similar test case in our codebase.

crates/rolldown/tests/rolldown/warnings/ineffective_dynamic_import/common_chunk/entry1.js
IWANABETHATGUY, line 1: I think there also exists a similar testcase for this one as well.

IWANABETHATGUY:
Could you review your test cases again and remove those that are redundant or already covered by existing tests in our codebase?

crates/rolldown/src/stages/generate_stage/code_splitting.rs
```diff
@@ -831,6 +835,75 @@ impl GenerateStage<'_> {
...
+  /// This method iterates through all chunks and their modules to detect cases where:
+  /// - A module has both static importers and dynamic importers
+  /// - A dynamic importer is in the same chunk as the imported module
+  /// - This makes the dynamic import ineffective (it won't create a separate chunk)
+  fn collect_ineffective_dynamic_import_warnings(&mut self, chunk_graph: &ChunkGraph) {
+    if self.options.inline_dynamic_imports {
+      return;
+    }
+
+    for chunk in chunk_graph.chunk_table.iter() {
```
sapphi-red: I assume this has a perf impact because we are adding a new loop here.

luke358: I noticed that there are duplicated loops at
- https://github.com/rolldown/rolldown/blob/bd3a07a78e22395d563425331cab56189c00863e/crates/rolldown/src/stages/generate_stage/mod.rs#L606
- https://github.com/rolldown/rolldown/blob/bd3a07a78e22395d563425331cab56189c00863e/crates/rolldown/src/stages/generate_stage/mod.rs#L565

These loops look very similar and might be merged. Since the chunks have already been processed at this point, it seems reasonable to collect warnings here.

luke358:
> Could you review your test cases again and remove those that are redundant or already covered by existing tests in our codebase?

Thanks for the feedback!  
I’m reviewing the test cases now and will remove the redundant ones shortly.

shulaoda:
Closed by #8284

PR #7610, opened 2025-12-21
fix: dts navigation on output plugin names

KazariEX:


hyf0:
Good job! Could you add some comments in the code so others wouldn't break it in the future?

packages/rolldown/src/plugin/index.ts
```diff
@@ -319,10 +319,14 @@ type AddonHookFunction = (
...
 type AddonHook = string | AddonHookFunction;
 
-interface OutputPlugin
-  extends
-    Partial<{ [K in OutputPluginHooks]: PluginHooks[K] }>,
-    Partial<{ [K in AddonHooks]: ObjectHook<AddonHook> }>
+interface OutputPlugin extends
+  Partial<
+    {
+      // use remapping keys to support IDE features
```
hyf0:
```suggestion
      // Use key remapping pattern to provide better  "go to definition" experience.
      // https://github.com/rolldown/rolldown/pull/7610
```

hyf0 (approved):
Thank you!

Issue #7608, opened 2025-12-21
Follow up issues of `postBanner/Footer`


hyf0:
Caused by https://github.com/rolldown/rolldown/pull/7583

- [ ] Incorrect sourcemap https://github.com/rolldown/rolldown/pull/7583#discussion_r2637206125
- [x] Warning for duplicate shebang https://github.com/rolldown/rolldown/pull/7583#issuecomment-3674790202

btea:
I tried the latest version and found that there were extra blank lines between `postBanner` and shebang in the output. [example](https://stackblitz.com/edit/vitejs-vite-zl8hy4bz?file=package.json)

hyf0:
> I tried the latest version and found that there were extra blank lines between `postBanner` and shebang in the output. [example](https://stackblitz.com/edit/vitejs-vite-zl8hy4bz?file=package.json)

Feel free to open a PR to fix it.

PR #7643, opened 2025-12-24
fix: remove the blank line between shebang and postBanner

btea:
rel https://github.com/rolldown/rolldown/issues/7608#issuecomment-3681231299


hyf0:
Please add a test case.

btea:
> Please add a test case.

Okay.

hyf0 (approved):
Thank you!

PR #7728, opened 2026-01-01
fix(linker): resolve race condition in side effects computation for export-star

### LINKED ISSUE #7646: Rolldown creates a different CJS bundle in ~50% runs
## Describe the bug
Rolldown creates a different CJS bundle on some runs when using AWS SDK for JavaScript

## Steps to reproduce
Follow instructions in https://github.com/trivikr/test-rolldown-aws-sdk-js-v3-bundle/tree/b3c43980ee3305cba689f361ec664029435f0ab1
At the time of comment, it's tested with `rolldown@1.0.0-beta.57` and `@aws-sdk/client-sts@3.946.0`

#### Observed behavior

A different bundle is created on some runs, leading to git diff
```console
$ ./test-build.sh
...

---------------

Iteration 2

> build
> rolldown -c

<DIR>/bundle.js  chunk │ size: 381.24 kB

✔ rolldown v1.0.0-beta.57 Finished in 93.30 ms
No changes in bundle.js

---------------

Iteration 3

> build
> rolldown -c

<DIR>/bundle.js  chunk │ size: 381.23 kB

✔ rolldown v1.0.0-beta.57 Finished in 92.95 ms
Git diff detected in bundle.js - failing at iteration 3
```

#### Expected behavior

The loop should run forever, and bundle created shouldn't change.

## Additional context

We noticed this issue when testing [@aws-sdk/find-v2](https://www.npmjs.com/package/@aws-sdk/find-v2), where we search for AWS SDK for JavaScript (v2) bundled in applications using different popular bundlers.

Over 90+ runs of the all bundler configurations, we noticed that the artifact difference only appears when we attempt to create CJS bundle using rolldown, and it happens around 30-40% of the times. The artifact doesn't differ for ESM bundle created with rolldown, or CJS/ESM bundle created with other bundlers. We tested webpack, rollup, esbuild, parcel and rspack.

camc314:
NOTE: I used AI for this, as I'm not familiar with the rolldown codebase. So feel free to close if this is the wrong fix. (That said, I manually validate each of the commits, including making sure that this fixes the issue, so I think it's OK.)

I've split this change into three commits

Firstly, I got AI to identify the cause of the issue, I cloned the repro as a submodule and got claude to investigate it. Eventually, this lead to identifying a race condition in reference_needed_symbols.  This function runs in parallel across all modules and was both reading importee.side_effects and writing to importer.side_effects simultaneously. When module A reads module B's side effects while another thread is upgrading B's side effects, A could see stale data.

**Commit 1: Validating the race condition**

After working out what the problem was, I added debug assertions to validate that the race condition was real and reproducible. This involved adding assertions in the relevant parts of the code to check for inconsistent states.

I then ran this against the reproduction repo, and was able to consistently trigger the panics that I added.

**Commit 2: First fix attempt** 

Given that the cause (race condition) was identified, I made an initial attempt to fix it (in that same file), the approach being:

1. Pre-compute which modules needed side_effects upgraded to `true` **before** entering the parallel section
2. Build a snapshot of all side effects values (side_effects_snapshot: Vec<bool>)
3. Use the snapshot for reads during parallel processing instead of reading live data
4. Defer the writes until after the parallel section completed

This eliminates the race condition.

However, this doensn't really correctly handle `transitive side effects e.g:

`A.js`: `export * from 'B'`   // B has WrapKind::None
`B.js`: `export * from 'C'`   // C is WrapKind::Cjs
`C.js`: `module.exports = {}`

B gets upgraded because of C, but A's snapshot of B was captured before B was upgraded. The fix worked for direct dependencies but could miss chains.

**Commit 3: Better solution**

Rather than adding complexity to handle transitive cases in the snapshot approach (commit 2), I refactored for a cleaner architecture.

During the link stage, there is a dedicated `determine_side_effects`. This runs before `reference_needed_symbold`, already handling transitive effects via import.

---------------

I've got no idea how to write a test for this given how flaky it is, so suggestions are appreciated.

fixes #7646

### DIFF
--- a/crates/rolldown/src/stages/link_stage/reference_needed_symbols.rs
+++ b/crates/rolldown/src/stages/link_stage/reference_needed_symbols.rs
@@ -1,4 +1,5 @@
 use std::ptr::addr_of;
+use std::sync::atomic::{AtomicU8, Ordering};
 
 use rolldown_common::{
   ExportsKind, ImportKind, ImportRecordIdx, ImportRecordMeta, Module, ModuleIdx, ModuleTable,
@@ -14,6 +15,14 @@ use rolldown_utils::{
 
 use super::LinkStage;
 use crate::utils::external_import_interop::import_record_needs_interop;
+
+// Debug assertions to detect race conditions on side_effects field.
+// During parallel processing, one thread may write to importer.side_effects while
+// another thread reads from importee.side_effects (which could be the same module).
+#[cfg(debug_assertions)]
+const SIDE_EFFECTS_READ: u8 = 1;
+#[cfg(debug_assertions)]
+const SIDE_EFFECTS_WRITTEN: u8 = 2;
 
 fn is_external_dynamic_import(
   table: &ModuleTable,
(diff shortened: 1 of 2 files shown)


IWANABETHATGUY (commented):
Thanks

IWANABETHATGUY (approved):

Issue #7784, opened 2026-01-07
Investigate and resolve incompatibility with `output.dynamicImportInCjs`


shulaoda:
See #7776

PR #7823, opened 2026-01-09
fix: transform non-static dynamic imports when `dynamicImportInCjs` is `false`

shulaoda:
Related to #7784

Previously, non-static dynamic imports like `import(foo)` or imports with `@vite-ignore` comment were not being transformed to require() calls in CJS output format when dynamicImportInCjs was false.

This change:
- Tracks statement indices for non-static/ignored dynamic imports
- Transforms them to `Promise.resolve().then(() => __toESM(require(...)))`
- Ensures __toESM runtime helper is properly included

### DIFF
--- a/crates/rolldown/src/module_finalizers/mod.rs
+++ b/crates/rolldown/src/module_finalizers/mod.rs
@@ -1722,11 +1722,25 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
     if expr.options.is_some() {
       return false;
     }
-    let Some(rec_idx) = self.ctx.module.imports.get(&expr.span) else {
-      return false;
-    };
-    // Make sure the import expression is in correct form. If it's not, we should leave it as it is.
-    let Some(str) = expr.source.as_static_module_request() else {
+
+    let (Some(str), Some(rec_idx)) =
+      (expr.source.as_static_module_request(), self.ctx.module.imports.get(&expr.span))
+    else {
+      if matches!(self.ctx.options.format, OutputFormat::Cjs)
+        && !self.ctx.options.dynamic_import_in_cjs
+      {
+        // Transform `import(expr)` to `Promise.resolve().then(() => __toESM(require(expr)))`
+        let source = expr.source.take_in(self.alloc);
+        let require_call = self.snippet.call_expr_with_arg_expr_expr("require", source);
+        let to_esm_fn_name = self.finalized_expr_for_runtime_symbol("__toESM");
+        let wrapped = self.snippet.wrap_with_to_esm(
+          to_esm_fn_name,
+          require_call,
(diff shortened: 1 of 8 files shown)


shulaoda:
* **#7823** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7823?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7823?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=7823).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown/src/stages/link_stage/reference_needed_symbols.rs
```diff
@@ -289,6 +289,14 @@ impl LinkStage<'_> {
           }
         });
 
+        // Handle non-static dynamic imports like `import(foo)` or `import('a' + 'b')`
+        // When format is CJS and dynamicImportInCjs is false, we need __toESM
+        if matches!(self.options.format, OutputFormat::Cjs) && !self.options.dynamic_import_in_cjs {
+          for &stmt_idx in &importer.non_static_dynamic_import_stmt_idxs {
```
hyf0: I guess you could have a `StmtMeta` to do this? Need to create extra `non_static_dynamic_import_stmt_idxs`

IWANABETHATGUY (approved):

crates/rolldown/src/module_finalizers/mod.rs
```diff
@@ -1722,11 +1722,25 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
     if expr.options.is_some() {
       return false;
     }
-    let Some(rec_idx) = self.ctx.module.imports.get(&expr.span) else {
-      return false;
-    };
-    // Make sure the import expression is in correct form. If it's not, we should leave it as it is.
-    let Some(str) = expr.source.as_static_module_request() else {
+
```
IWANABETHATGUY, line 1725: I assume both of two pattern genereate similar asm, but the previous one seems more readable.

shulaoda: Because we need to reuse the code that generates the `require` expression, it was written this way. Do you think the following is acceptable? 
```js
let rec_idx = self.ctx.module.imports.get(&expr.span);
let str = expr.source.as_static_module_request();
let (Some(str), Some(rec_idx)) = (str, rec_idx) else { .. }
```

PR #7879, opened 2026-01-14
fix(plugins): add moduleType to transform output for vite builtin plu…

richavos:
…gins

Several builtin vite plugins were returning `HookTransformOutput` without setting the `module_type` field. According to the Rollup plugin spec, when a transform hook changes the module content to a different type (e.g., CSS/HTML to JS), it should specify the resulting `moduleType`.

The NAPI binding layer expects `moduleType` to be present when the output is passed to JavaScript, causing "Missing field moduleType" errors when these plugins were invoked through rolldown-vite.

Fixed plugins:
- vite-react-refresh-wrapper: wraps React components with HMR refresh code
- vite-css: transforms CSS files to JS modules
- vite-css-post: transforms CSS proxy files to JS exports
- vite-html: transforms HTML entry files to JS modules

All these plugins output JavaScript regardless of input type, so they now explicitly set `module_type: Some(ModuleType::Js)`.

Plugins that preserve the original module type (replace, import-glob, dynamic-import-vars, asset-import-meta-url) correctly leave module_type as None to preserve the input type.


packages/rolldown/tests/fixtures/builtin-plugin/vite-react-refresh-wrapper/_config.ts
sapphi-red, line 1: This test doesn't fail even if I revert the changes in `crates/rolldown_plugin_vite_react_refresh_wrapper/src/lib.rs`. Would you create a reproduction that I can test?

sapphi-red (changes requested):

packages/rolldown/tests/fixtures/builtin-plugin/vite-react-refresh-wrapper/_config.ts
richavos: Apologies, updated now!

packages/rolldown/tests/fixtures/builtin-plugin/vite-react-refresh-wrapper/_config.ts
```diff
@@ -0,0 +1,42 @@
...
+        name: 'assert-module-type-updated',
+        transform(code, id, meta) {
+          if (id.endsWith('main.jsx')) {
+            // The vite-react-refresh-wrapper plugin outputs JS code, so moduleType
+            // should be 'js', not the original 'jsx' from the file extension.
+            // This check catches the bug where HookTransformOutput.module_type is None,
+            // causing the moduleType to incorrectly remain as the input type.
+            if (code.includes('RefreshRuntime')) {
+              expect(meta.moduleType, 'moduleType should be updated to js after transform').toBe('js');
+            }
```
sapphi-red, line 29: rolldown_plugin_vite_react_refresh_wrapper outputs JSX code if the input is JSX (and outputs JS code if the input is JS), so this description doesn't make sense to me

sapphi-red:
Closing as stale

PR #7967, opened 2026-01-20
test: add lazy barrel optimization test cases

shulaoda:

PR #7933, opened 2026-01-16
feat: support lazy barrel optimization

shulaoda:

PR #7907, opened 2026-01-15
refactor(import_record): make resolved_module optional

shulaoda:

PR #7969, opened 2026-01-20
docs: add in-depth documentation for lazy barrel optimization

shulaoda:
Needs https://github.com/rolldown/benchmarks/pull/97

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -83,6 +83,7 @@ const sidebarForInDepth: DefaultTheme.SidebarItem[] = [
         link: '/in-depth/why-plugin-hook-filter.md',
       },
       { text: 'Directives', link: '/in-depth/directives.md' },
+      { text: 'Lazy Barrel Optimization', link: '/in-depth/lazy-barrel-optimization.md' },
     ],
   },
 ];
@@ -197,6 +198,7 @@ const sidebarForGlossary: DefaultTheme.SidebarItem[] = [
   {
     text: 'Glossary',
     items: [
+      { text: 'Barrel Module', link: '/glossary/barrel-module.md' },
       { text: 'Entry', link: '/glossary/entry.md' },
       { text: 'Entry Chunk', link: '/glossary/entry-chunk.md' },
       { text: 'Entry Name', link: '/glossary/entry-name.md' },
--- a/docs/in-depth/lazy-barrel-optimization.md
+++ b/docs/in-depth/lazy-barrel-optimization.md
@@ -1,0 +1,235 @@
+# Lazy Barrel Optimization
+
+Lazy barrel is an optimization feature that enhances build performance by avoiding compilation of unused re-export modules in side-effect-free [barrel modules](/glossary/barrel-module).
+
+## Why use Lazy Barrel
+
+Large component libraries like [Ant Design](https://ant.design/) use barrel modules extensively. When you import just one component, the bundler traditionally compiles thousands of modules, most of which are unused.
+
+Here's a real-world example importing only `Button` from antd:
+
+```js
+import { Button } from 'antd';
+Button;
+```
+
+| Metric               | Without lazy barrel | With lazy barrel |
+| -------------------- | ------------------- | ---------------- |
+| Modules compiled     | 2986                | 300              |
+| Build time (macOS)   | ~65ms               | ~28ms            |
+| Build time (Windows) | ~210ms              | ~50ms            |
+
+By enabling lazy barrel, Rolldown reduces the number of compiled modules by **90%** and speeds up the build by **2-4x**.
+
+::: tip
+You can reproduce this benchmark using the [lazy-barrel example](https://github.com/rolldown/benchmarks/tree/main/examples/lazy-barrel).
+:::
+
+## How Lazy Barrel works
+
+When enabled, Rolldown analyzes which exports are actually used and only compiles those modules. The unused re-export modules are skipped, significantly improving build performance for large codebases with many barrel modules.
+
+### Basic example
+
+```js
+// barrel/index.js
+export { a } from './a';
+export { b } from './b';
+
+// main.js
+import { a } from './barrel';
+console.log(a);
+```
+
+With lazy barrel optimization:
+
+- `barrel/index.js` is loaded and analyzed
+- Only `a.js` is compiled since `a` is imported
+- `b.js` is **not** compiled since `b` is not used
+
+## Supported export patterns
+
+Lazy barrel optimization works with various export patterns:
+
+### Named re-exports
+
+```js
+export { Component } from './Component';
+export { helper as utils } from './helper';
+export { default as Button } from './Button';
+```
+
+### Star re-exports
+
+```js
+export * from './components';
+```
+
+### Mixed exports
+
+```js
+export { a } from './a';
+export * from './others';
+export * from './more';
+```
+
+When an import can be found in named exports, star exports are not searched, avoiding unnecessary module loading.
+
+However, if the import is not found in named exports, all star re-exports will be loaded to resolve it. If those star re-exported modules are also barrel modules, only the specific import specifier will be loaded from them.
+
+## Advanced scenarios
+
+### Self re-export
+
+Lazy barrel correctly handles barrel modules that re-export from themselves:
+
+```js
+// barrel/index.js
+export { a } from './a';
+export { a as b } from './index'; // self re-export
+```
+
+### Circular exports
+
+Lazy barrel correctly handles circular export relationships between barrel modules:
+
+```js
+// barrel-a/index.js
+export { a } from './a';
+export * from '../barrel-b';
+
+// barrel-b/index.js
+export { b } from './b';
+export { a as c } from '../barrel-a'; // circular reference
+```
+
+### Dynamic import entry
+
+When a barrel module is dynamically imported, it becomes an entry point and all its exports must be available:
+
+```js
+// barrel/a.js
+export const a = 'a';
+import('./index.js'); // makes barrel an entry point
+
+// barrel/index.js
+export { a } from './a';
+export { b } from './b'; // b.js will be loaded
+```
+
+However, if `b.js` is also a barrel module, its unused exports will still be optimized.
+
+### Non-re-export imports
+
+When a barrel module contains non-re-export imports, those imported modules will always be loaded:
+
+```js
+// barrel/index.js
+import { a } from './a'; // a.js is loaded (actual import)
+export { b } from './b'; // b.js may or may not be loaded
+export function helper() {
+  return a;
+}
+```
+
+However, if `./a` is also a barrel module, lazy barrel optimization still applies to it, only loading the exports that are actually used (`a` in this case).
+
+### Side-effect imports
+
+The following patterns are treated as having side effects, so the target module will always be loaded:
+
+```js
+import './module'; // side-effect import
+import {} from './module'; // empty import
+export {} from './module'; // empty re-export
+```
+
+If the target module is a barrel module, the barrel itself is loaded but its re-exported modules are **not** loaded.
+
+### Unused import specifiers
+
+By default, even if an imported specifier is not used, its corresponding module will still be loaded:
+
+```js
+// barrel/index.js
+export { a } from './a';
+export { b } from './b';
+
+// main.js
+import { a } from './barrel'; // a.js is loaded even if `a` is never used
+```
+
+To automatically remove unused import specifiers and avoid loading their modules, set `treeshake.invalidImportSideEffects` to `false`:
+
+```js
+// rolldown.config.js
+export default {
+  treeshake: {
+    invalidImportSideEffects: false,
+  },
+};
+```
+
+## Configuration
+
+Enable lazy barrel optimization in your Rolldown configuration:
+
+```js
+// rolldown.config.js
+export default {
+  experimental: {
+    lazyBarrel: true,
+  },
+};
+```
+
+## Requirements
+
+For lazy barrel optimization to work effectively, barrel modules should be side-effect-free. Rolldown enables tree-shaking by default and performs smart side-effect analysis. For cases where automatic detection isn't sufficient, you can explicitly declare side-effect-free modules through:
+
+1. **Package declaration**: Adding `"sideEffects": false` to `package.json`
+
+2. **Bundler configuration**: Using the `treeshake.moduleSideEffects` option
+
+```js
+// rolldown.config.js
+export default {
+  treeshake: {
+    moduleSideEffects: [
+      // Mark barrel modules as side-effect-free using regex
+      { test: /\/barrel\//, sideEffects: false },
+      // Or mark specific paths
+      { test: /\/components\/index\.js$/, sideEffects: false },
+    ],
+  },
+};
+```
+
+You can also use a function for more complex logic:
+
+```js
+// rolldown.config.js
+export default {
+  treeshake: {
+    moduleSideEffects: (id) => {
+      // Mark all index.js files as side-effect-free
+      if (id.endsWith('/index.js')) return false;
+      return true;
+    },
+  },
+};
+```
+
+## When to use
+
+Lazy barrel optimization is particularly beneficial when:
+
+- Your codebase has many barrel modules (common in component libraries)
+- Barrel modules re-export many modules but consumers typically use only a few
+- You're importing from large component libraries like antd or material-ui
+
+## Limitations
+
+- Barrel modules with side effects cannot be optimized
+- Unmatched named imports require loading all star re-exports to resolve
+- Entry files, `import * as ns`, `import('..')`, `require('..')`, etc. will cause the barrel module to load all its exports
--- a/docs/in-depth/lazy-barrel-optimization.md
+++ b/docs/in-depth/lazy-barrel-optimization.md
@@ -185,7 +185,7 @@ export default {
 
 ## Requirements
 
-For lazy barrel optimization to work effectively, barrel modules should be side-effect-free. Rolldown enables tree-shaking by default and performs smart side-effect analysis. For cases where automatic detection isn't sufficient, you can explicitly declare side-effect-free modules through:
+For lazy barrel optimization to work, barrel modules needs to be marked as side-effect-free explicitly. While Rolldown performs side-effect analysis, it requires to read the modules imported from the barrel files. By marking the barrel modules as side-effect-free, Rolldown can rely on that fact and apply the optimization:
 
 1. **Package declaration**: Adding `"sideEffects": false` to `package.json`
 
--- a/docs/glossary/barrel-module.md
+++ b/docs/glossary/barrel-module.md
@@ -1,0 +1,20 @@
+# Barrel Module
+
+A barrel module is a module that re-exports functionality from other modules, commonly used to create a cleaner public API for a package or directory:
+
+```js
+// components/index.js (barrel module)
+export { Button } from './Button';
+export { Card } from './Card';
+export { Modal } from './Modal';
+export { Tabs } from './Tabs';
+// ... dozens more components
+```
+
+This allows consumers to import from a single entry point:
+
+```js
+import { Button, Card } from './components';
+```
+
+However, barrel modules can cause performance issues because bundlers traditionally need to compile all re-exported modules, even if only a few are actually used. See [Lazy Barrel Optimization](/in-depth/lazy-barrel-optimization) for how Rolldown addresses this.
--- a/docs/glossary/index.md
+++ b/docs/glossary/index.md
@@ -2,6 +2,10 @@
 
 Common terms and concepts used in Rolldown documentation.
 
+## B
+
+- [Barrel Module](./barrel-module.md)
+
 ## E
 
 - [Entry](./entry.md)
--- a/packages/rolldown/src/options/input-options.ts
+++ b/packages/rolldown/src/options/input-options.ts
@@ -617,6 +617,10 @@ export interface InputOptions {
     /**
      * Control whether to enable lazy barrel optimization.
      *
+     * Lazy barrel optimization avoids compiling unused re-export modules in side-effect-free barrel modules,
+     * significantly improving build performance for large codebases with many barrel modules.
+     *
+     * @see {@link https://rolldown.rs/in-depth/lazy-barrel-optimization | Lazy Barrel Documentation}
      * @default false
      */
     lazyBarrel?: boolean;


shulaoda:
* **#8009** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8009?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7986** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7986?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7969** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7969?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7969?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7967** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7967?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#7979](https://github.com/rolldown/rolldown/pull/7979) <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7979?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#7933** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7933?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7908** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7908?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7907** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/7907?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=7969).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

docs/in-depth/lazy-barrel-optimization.md
```diff
@@ -0,0 +1,256 @@
+# Lazy Barrel
+
+Lazy barrel is an optimization feature that enhances build performance by avoiding compilation of unused re-export modules in side-effect-free barrel modules.
```
TheAlexLichter: note: can we add a [glossary](https://rolldown.rs/glossary/#glossary) entry for barrel modules?

docs/in-depth/lazy-barrel-optimization.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Why lazy barrel
+
+Large component libraries like [Ant Design](https://ant.design/) use barrel modules extensively. When you import just one component, the bundler traditionally compiles thousands of modules, most of which are unused.
```
TheAlexLichter, line 7: is there another example (just so we can name more than one)?

docs/in-depth/lazy-barrel-optimization.md
```diff
@@ -0,0 +1,256 @@
+# Lazy Barrel
+
+Lazy barrel is an optimization feature that enhances build performance by avoiding compilation of unused re-export modules in side-effect-free barrel modules.
+
+## Why lazy barrel
```
TheAlexLichter: casing (also below)

```suggestion
## Why Lazy Barrel
```

docs/in-depth/lazy-barrel-optimization.md
```diff
@@ -0,0 +1,256 @@
...
+| Build time (macOS)   | ~65ms               | ~28ms            |
+| Build time (Windows) | ~210ms              | ~50ms            |
+
+By enabling lazy barrel, Rolldown reduces the number of compiled modules by **90%** and speeds up the build by **2-4x**.
+
+::: tip
+You can reproduce this benchmark using the [lazy-barrel example](https://github.com/rolldown/benchmarks/tree/main/examples/lazy-barrel).
+:::
+
+## What is a barrel module
```
TheAlexLichter: Ah, this could go into the glossary entry mentioned above (as it would be helpful to know in general)

docs/in-depth/lazy-barrel-optimization.md
```diff
@@ -0,0 +1,256 @@
...
+
+When an import can be found in named exports, star exports are not searched, avoiding unnecessary module loading.
+
+However, if the import is not found in named exports, all star re-exports will be loaded to resolve it. If those star re-exported modules are also barrel modules, only the specific import specifier will be loaded from them.
+
+## Advanced scenarios
+
+### Self re-export
+
+Barrel modules can re-export from themselves:
```
TheAlexLichter: Is this also "properly" handled by lazy barrel? Adding a note how the special case is dealt with (any caveats, otherwise note that this is correctly handled).

docs/.vitepress/config.ts
```diff
@@ -83,6 +83,7 @@ const sidebarForInDepth: DefaultTheme.SidebarItem[] = [
         link: '/in-depth/why-plugin-hook-filter.md',
       },
       { text: 'Directives', link: '/in-depth/directives.md' },
+      { text: 'Lazy Barrel', link: '/in-depth/lazy-barrel.md' },
```
TheAlexLichter: I feel like there is a word missing, e.g. "Lazy Barrel Compilation". This is just a hunch and might be too close to Lazy Compilation itself.

docs/in-depth/lazy-barrel-optimization.md
shulaoda: Why would you be confused about whether it is handled correctly? It has a subsection titled **Advanced scenarios** that lists the advanced use cases it supports. 🤔

docs/in-depth/lazy-barrel-optimization.md
shulaoda: Do you mean moving it to the glossary? Does it still need to be kept here? I think it should also remain here; otherwise, users would have the additional cognitive overhead of having to look it up in the glossary.

docs/in-depth/lazy-barrel-optimization.md
shulaoda: Do you mean all the second-level headings below should be capitalized? I noticed that the other `In-depth` sections do not do this. Should we only capitalize the top-level heading instead? I actually based this on the documentation style used by Rspack. https://rspack.rs/guide/optimization/lazy-barrel#lazy-barrel

docs/in-depth/lazy-barrel-optimization.md
shulaoda: See below https://github.com/rolldown/rolldown/pull/7969#discussion_r2708094270

docs/in-depth/lazy-barrel-optimization.md
shulaoda: Do you mean real-world project scenarios? I haven’t found any yet — this may require users to try it out and report back to us. For now, I don’t think it’s necessary to introduce more cases in this document; it’s mainly meant to help users understand the use cases and benchmark results for lazy barrels.

By the way, this example is the same one Rspack used when they posted their tweet related with lazy barrel.

docs/.vitepress/config.ts
shulaoda: Could this be confused with `Lazy Compilation`? Perhaps we should use `Lazy Barrel Optimization` instead?

TheAlexLichter: Yes. Lazy Barrel Optimization is good too 👍

docs/in-depth/lazy-barrel-optimization.md
shulaoda: But make sense for me, did it.

docs/in-depth/lazy-barrel-optimization.md
shulaoda: Maybe `"Why use lazy barrel optimization"`?

docs/in-depth/lazy-barrel-optimization.md
TheAlexLichter: This is not urgent as Ant Design is a great example.

But if there are more libraries that come to our (or our users) mind it'd great to list them.

docs/in-depth/lazy-barrel-optimization.md
TheAlexLichter: I'm prefer title casing for all headings but we don't have to and keep them mostly "sentence case".

As "Lazy Barrel"  / "Lazy Barrel Optimization" could be considered a "Proper Noun", I'd capitalize here in any case.

docs/in-depth/lazy-barrel-optimization.md
TheAlexLichter: For now we can copy it and leave it here as well. The user still might have a cognitive overhead as the term is introduced quite late though, while liking to the glossary when the term is used the first time could help.

docs/in-depth/lazy-barrel-optimization.md
TheAlexLichter: > Why would you be confused about whether it is handled correctly? It has a subsection titled Advanced scenarios that lists the advanced use cases it supports. 🤔

I'd clear the doubt for each. Advanced scenarios can mean a lot of things. Special cases to consider, caveats, etc.
That's why confirming with a sentence will make sure the user (and also LLM) knows that this is no problem for our optimization, or if it is, under which constraints.

docs/in-depth/lazy-barrel-optimization.md
shulaoda: Got it. It looks like moving it to the glossary is the better choice, so we won’t keep it here.

TheAlexLichter (approved):

docs/in-depth/lazy-barrel-optimization.md
```diff
@@ -0,0 +1,235 @@
...
+
+The following patterns are treated as having side effects, so the target module will always be loaded:
+
+```js
+import './module'; // side-effect import
+import {} from './module'; // empty import
+export {} from './module'; // empty re-export
+```
+
+If the target module is a barrel module, the barrel itself is loaded but its re-exported modules are **not** loaded.
```
sapphi-red, line 147: Since we require the barrel file to be marked with `moduleSideEffects: false`, I guess we should remove these imports to be consistent.

docs/in-depth/lazy-barrel-optimization.md
```diff
@@ -0,0 +1,235 @@
...
+export default {
+  experimental: {
+    lazyBarrel: true,
+  },
+};
+```
+
+## Requirements
+
+For lazy barrel optimization to work effectively, barrel modules should be side-effect-free. Rolldown enables tree-shaking by default and performs smart side-effect analysis. For cases where automatic detection isn't sufficient, you can explicitly declare side-effect-free modules through:
```
sapphi-red: > For cases where automatic detection isn't sufficient

To determine the sideeffect of the imported module automatically, we have to load the module itself. This mean we cannot apply lazy barrel optimization. Thus, the explicit declaration is a requirement for lazy barrel optimization.

shulaoda: I’d like to clarify one point: we only decide whether Lazy Barrel optimization can be applied **after the module is loaded**, based on its `moduleSideEffects`. Lazy Barrel is predicated on `moduleSideEffects` being `false` and is specifically an optimization for **re-exports**.

What it determines is whether we need to eagerly load the modules referenced by the current module’s re-exports. If the module has no re-exports, we ultimately only load the module itself.

For example, if there is an `import { b } from 'a'`, the `a` module is always loaded eagerly, because this is a direct `import` rather than a `re-export`. The specifier `b` is then passed to `a`. After `a` is loaded, if `a` has `moduleSideEffects: false`, contains re-exports, and `b` matches one of the re-exported export names, we only load the corresponding re-exported module and propagate the `b` specifier to it.

shulaoda: Lazy Barrel is an optimization specifically for the **re-exports** of a module. During the module loading process, if the current module is **side-effect-free**, we can selectively load its re-exports **based on the exports that are actually referenced**.

shulaoda: For example, starting from an entry module, suppose the entry contains `import { d } from 'a'; export { b } from 'b';`. Even if the entry module is side-effect-free, we still need to load `a` as well as the re-exported module `b`.

After a is loaded, assume `a` contains `import { c } from 'c'; export { d } from 'd'; export { e } from 'e';`. If `a` is side-effect-free and we only need the export `d`, we will selectively load the re-exported module `d`, while `e` will not be loaded. Note that `c` will still be loaded and propagated with the `c` specifier, since it is a direct import rather than a re-export.

shulaoda: Since the entry module eagerly loads all of its imports, including re-exported dependencies, the required set of exports is known at load time. After a module is loaded, if it is side-effect-free, its re-exports can be selectively resolved to enable the Lazy Barrel optimization.

sapphi-red: I'm not sure if I get your point.
This is the example where the current implementation is not working as expected.
https://github.com/rolldown/rolldown/pull/7979/files#diff-609121ac68017fbd3f49593f162ca8cf529395f26043819b86f081e8c9e42f08R32-R39
`console.log('b')` should be kept because `barrel/index.js` is not marked as side effect free explicitly (it is `DeterminedSideEffects::Analyzed(_)` and not `DeterminedSideEffects::UserDefined(false)`).

shulaoda: I understand your point. You’re saying that we shouldn’t rely on inferred analysis, but instead depend on explicit signals — such as `moduleSideEffects: false` returned by a plugin hook, `"sideEffects": false` in `package.json`, or a configuration provided via `treeshake.moduleSideEffects`. Is that correct?

sapphi-red: Yes, because I don't think there's a way to analyze whether the imported module has a sideeffect without actually reading that module.

shulaoda: > there's a way to analyze whether the imported module has a sideeffect without actually reading that module

I’d like to dig into this statement. In our current implementation, our `“analysis”` only happens after the module has been loaded.

sapphi-red:
```suggestion
For lazy barrel optimization to work, barrel modules needs to be marked as side-effect-free explicitly. While Rolldown performs side-effect analysis, it requires to read the modules imported from the barrel files. By marking the barrel modules as side-effect-free, Rolldown can rely on that fact and apply the optimization:
```

docs/in-depth/lazy-barrel-optimization.md
shulaoda: Rspack will eagerly load the first two cases by default, but it will eliminate the last one. Since `export {} from '..'` is invalid syntax and very uncommon, I didn’t handle it and treated it as `import '..'` instead.

That said, your point also makes sense. We don’t necessarily need to match Rspack’s behavior, but I need think through whether removing these could introduce any potential issues. 🤔

sapphi-red: Rollup works like:

- [bundler explorer 1](https://bundler.sxzz.dev/#eNptkUtuhDAQRK/S8gaQEFFmdkhZ5gRZZrJwcBs58gf5I0VC3D02BsYzgQ1dXVT7uZkJJ/1MFBW6+3Gp1KQ/ZEuGqISajPVQdS/f1FqU0alu9qajjaT3NuDSEm7MazFhl3nCYLQzEjtpxrpKVtWQnLk8ZlZ5nrkcmetjZpXnmWvOHNh76t54vuHGvd5v684LcGvU5ibC1cXfEzexlG4C8hANeIMElPcWkayRMkxd9LkYC7J//Qy4jWPIaZDx0DQGYJJhFNr18Jk17EZ6NFXY50Pbe9dbqh03VtWDYdiCYE0ZAhAcasE6oQcZGLq6yruqmqfvACz6YDXMoAwLEj8Ew3fOcfARiFPpEJYyUIij3Iqv9FprsqTfdeyBLH8HgNT9)
- [bundler explorer 2](https://bundler.sxzz.dev/#eNptUstuhDAM/BUrF0BCVN29IfXYL+ix2wMlDkoVEpSHVAnx782DR+jCBY/H44ydzISRdiZjx2XzY0IoSbvDmvQe8XFS2sIMTClYgGk1QtG8fHdao/BlxUM/pK9F0lrtcKmJL3zN2m0wteuVNEpgI9RQFoEqKpI0t7MmwmvNbdfcz5oIrzX3pNltb6ojcRrXT7j6jvNtSzjmXx1GFn8v2OAlZ4MhG5f4BsFQ2pu3pJUQbmo8z/iQOXvKJ4NrO4qsc8IfGtoATMINXJoWPhOGjQif7EZs06H1kbW6k4YpPZa9olgDp1UuAuAMSk4bLnvhKJqySLsqqn91ABqt09K/kVFRJ/CDU3xnDHvrDbFOGIQlF2RgD9fgK/xiTJZwXfseyPIHu+HZbQ==)
- [bundler explorer 3](https://bundler.sxzz.dev/#eNptUstqhTAQ/ZUhGxXE0nt3Qpf9gi57u7BmIikxkTygIP578/ARexXEOXNyZs6MmQkj7UzGjsvmx4RQknaHNek94uOktIUZmFKwANNqhKJ5+e60RuGPFQ/9kOHtlTRKYCPUUPqzVcj5Gkhaqx0uNfHJ16zNBlObXF0EqqhI0tzOmgivNbddcz9rIrzW3JNmH2dTHYnTGvzkq+8497acYy+rw8ji7wUbvORsMGTjct8gGIpcsKSVEG5qPM/4kDl7yieDazmKrHPCNw1lACbhBi5NC58Jw0aER3YjtqlpfWSt7qRhSo9lryjWwGmViwA4g5LThsteOIqmLNKuiurfOQCN1mnp786oqBP4wSm+M4a99YZYJwzCkgsysIdr8BU+MSZL+F37HsjyB0pr4pU=)

shulaoda: That makes sense. I’ve been hesitant to remove them because our behavior would differ from Rspack’s, but this is actually the correct behavior and we should implement it. I’ll address this in the next PR — it shouldn’t take too long.

shulaoda: Wait, I think this falls outside the scope of Lazy Barrel, since Lazy Barrel only targets re-exports.

We could potentially avoid loading `import '..';`, `import {} from '..';`, and `export {} from '..'` by default when they appear in a side-effect-free module. Because we make `resolved_module` optional up front, we now are able to apply this optimization.

shulaoda: For now, I’ll implement this under Lazy Barrel and we can discuss it further afterward.

shulaoda: > Rollup works like:
> 
> * [bundler explorer 1](https://bundler.sxzz.dev/#eNptkUtuhDAQRK/S8gaQEFFmdkhZ5gRZZrJwcBs58gf5I0VC3D02BsYzgQ1dXVT7uZkJJ/1MFBW6+3Gp1KQ/ZEuGqISajPVQdS/f1FqU0alu9qajjaT3NuDSEm7MazFhl3nCYLQzEjtpxrpKVtWQnLk8ZlZ5nrkcmetjZpXnmWvOHNh76t54vuHGvd5v684LcGvU5ibC1cXfEzexlG4C8hANeIMElPcWkayRMkxd9LkYC7J//Qy4jWPIaZDx0DQGYJJhFNr18Jk17EZ6NFXY50Pbe9dbqh03VtWDYdiCYE0ZAhAcasE6oQcZGLq6yruqmqfvACz6YDXMoAwLEj8Ew3fOcfARiFPpEJYyUIij3Iqv9FprsqTfdeyBLH8HgNT9)
> * [bundler explorer 2](https://bundler.sxzz.dev/#eNptUstuhDAM/BUrF0BCVN29IfXYL+ix2wMlDkoVEpSHVAnx782DR+jCBY/H44ydzISRdiZjx2XzY0IoSbvDmvQe8XFS2sIMTClYgGk1QtG8fHdao/BlxUM/pK9F0lrtcKmJL3zN2m0wteuVNEpgI9RQFoEqKpI0t7MmwmvNbdfcz5oIrzX3pNltb6ojcRrXT7j6jvNtSzjmXx1GFn8v2OAlZ4MhG5f4BsFQ2pu3pJUQbmo8z/iQOXvKJ4NrO4qsc8IfGtoATMINXJoWPhOGjQif7EZs06H1kbW6k4YpPZa9olgDp1UuAuAMSk4bLnvhKJqySLsqqn91ABqt09K/kVFRJ/CDU3xnDHvrDbFOGIQlF2RgD9fgK/xiTJZwXfseyPIHu+HZbQ==)
> * [bundler explorer 3](https://bundler.sxzz.dev/#eNptUstqhTAQ/ZUhGxXE0nt3Qpf9gi57u7BmIikxkTygIP578/ARexXEOXNyZs6MmQkj7UzGjsvmx4RQknaHNek94uOktIUZmFKwANNqhKJ5+e60RuGPFQ/9kOHtlTRKYCPUUPqzVcj5Gkhaqx0uNfHJ16zNBlObXF0EqqhI0tzOmgivNbddcz9rIrzW3JNmH2dTHYnTGvzkq+8497acYy+rw8ji7wUbvORsMGTjct8gGIpcsKSVEG5qPM/4kDl7yieDazmKrHPCNw1lACbhBi5NC58Jw0aER3YjtqlpfWSt7qRhSo9lryjWwGmViwA4g5LThsteOIqmLNKuiurfOQCN1mnp786oqBP4wSm+M4a99YZYJwzCkgsysIdr8BU+MSZL+F37HsjyB0pr4pU=)

@sapphi-red I’ve thought about this again, and we can’t simply infer whether this form should be removed based on how the entry module uses it. We need to look at the behavior inside the **side-effect-free barrel module** itself. 

See [here](https://bundler.sxzz.dev/#eNptUstqhTAQ/ZUhGxXE0nvBhaXLfkGXvV1YM5GUmEhMoCD+e/PwEVsFcc7MHOfMSWbCSDOToeWy+p58KEmzw5J0DvFhVNrADEwpWIBpNUBWPX21WqNwbdlDP6R/OyUnJbASqs9db+Fz7h9IGqMtLiVxyedkzAbjmJSd+VJWkMi5nTkBXnNuO+d+5gR4zblHzr7OxjoSJxvc5qvusPdmzuHLqjBU8eei6rWkVS/IBHNfwQsKtVRlXdfFSzSStWLyTmolhB0r18V4n4j+l4/a10kUWWuF0+MnAIzC9lxODXxEDFvBP7IdsIl6yiNrdCsnpvSQd4piCZwWKQmAM8g5rbjshKU45Vm0MSv+9AFoNFZLd60GRa3Ad07xjTHsjBMU1oQlJSRgD9fg039CTBZ/krsPZPkF2CDrQA==).

shulaoda: However, Lazy Barrel is a new optimization that is explicitly built around the concept of side-effect-free modules. It allows us to honor the user’s explicit side-effect declarations and trust their configuration. In that sense, it represents a different optimization model from Rollup, and we can’t simply use its bundling results as a reference.

shulaoda: It’s not just `Rspack` — I should also look at how `Turbopack` and `Parcel` handle the constraints and behavior around Lazy Barrel. I’ll share my conclusions afterward.

PR #7986, opened 2026-01-21
feat(lazy-barrel): skip plain imports in side-effect-free modules

shulaoda:
Since we require the barrel file to be marked with `moduleSideEffects: false`, I guess we should remove these imports to be consistent.

_Originally posted by @sapphi-red in https://github.com/rolldown/rolldown/pull/7969#discussion_r2710810334_

PR #8005, opened 2026-01-22
docs(in-depth): add dead code elimination page

sapphi-red:


sapphi-red:
* **#8005** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8005?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8005?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=8005).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

docs/in-depth/dead-code-elimination.md
```diff
@@ -0,0 +1,251 @@
...
+const button = createComponent('button');
+// This call will also be removed if `input` is unused
+const input = createComponent('input');
+```
+
+This can be more convenient than adding `@__PURE__` to every call site when you know the function itself is always pure.
+
+### Marking Entire Modules
+
+While annotations mark individual expressions or functions, you can also mark entire modules as side-effect-free. This is equivalent to marking every statement in the module as side-effect-free.
```
sapphi-red, line 109: > This is equivalent to marking every statement in the module as side-effect-free.

This is what I was expecting. But it seems this isn't how it works in webpack ([related docs](https://webpack.js.org/guides/tree-shaking/#clarifying-tree-shaking-and-sideeffects:~:text=To%20tackle%20this,for%20side%20effects.%22.)) and Rollup ([related docs](https://rollupjs.org/plugin-development/#transform:~:text=If%20false%20is%20returned%20for%20moduleSideEffects%20and%20no%20other%20module%20imports%20anything%20from%20this%20module%2C%20then%20this%20module%20will%20not%20be%20included%20even%20if%20the%20module%20would%20have%20side%20effects.)) because [the sideeffect field in the example can be `"sideEffects": ["**/*.css"]`](https://webpack.js.org/guides/tree-shaking/#the-correct-configuration-for-this-package)...
But this definition would not allow us to do #7986. We can skip the import if the imported module is side-effect-free, instead of "if the importer module is side-effect-free".

sapphi-red:
Note: this PR describes how *I understood*. It's not correct as pointed out in https://github.com/rolldown/rolldown/pull/8005#discussion_r2715351317

docs/in-depth/dead-code-elimination.md
shulaoda: Great point. I was misunderstanding too🤔

PR #8042, opened 2026-01-24
refactor(rust): use `is_data_url` more consistently

valadaptive:
I noticed this when fixing another issue.

While many places were using `s.trim_start().starts_with("data:")`, some others were *just* doing `s.starts_with("data:")` with no leading whitespace trim. I believe the correct behavior is to always trim leading whitespace, but this *may* change the behavior in some edge cases.

Note also that `resolve_id_with_plugins` had its own `is_data_url` helper, which is completely identical to the one that *already* existed in `rolldown_utils`! I've removed it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3347,6 +3347,7 @@ version = "0.1.0"
 dependencies = [
  "memchr",
  "rolldown_plugin",
+ "rolldown_utils",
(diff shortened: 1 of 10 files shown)


hyf0 (approved):
Thank you!

PR #8043, opened 2026-01-24
fix: use string literal span for `new URL` error diagnostic

valadaptive:
Previously, the diagnostic span was taken from the *entire* `new URL(...)` expression. However, the `UNRESOLVED_IMPORT` diagnostic assumes all spans refer to string literals, and unconditionally strips the quotes. This results in the first and last characters of the `new URL(...)` expression being stripped:

image

I *think* dynamic imports' diagnostics already refer to just the string literal's span, so the `new URL` import should too.

### DIFF
--- a/crates/rolldown/src/ast_scanner/new_url.rs
+++ b/crates/rolldown/src/ast_scanner/new_url.rs
@@ -48,8 +48,13 @@ impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
       return;
     }
 
-    let idx =
-      self.add_import_record(path, ImportKind::NewUrl, expr.span, ImportRecordMeta::empty(), None);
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY:
> Previously, the diagnostic span was taken from the _entire_ `new URL(...)` expression. However, the `UNRESOLVED_IMPORT` diagnostic assumes all spans refer to string literals, and unconditionally strips the quotes. This results in the first and last characters of the `new URL(...)` expression being stripped:
> 
> image
> I _think_ dynamic imports' diagnostics already refer to just the string literal's span, so the `new URL` import should too.

Can you provide the screenshots before and after this PR?

IWANABETHATGUY (approved):

PR #8110, opened 2026-01-28
build: remove warnings  for building rolldown with `not(feature = "experimental")`

coolreader18:
The `rolldown` crate fails to build with its default featureset, since these two function calls aren't gated by `#[cfg(feature = "experimental")]`. Though it seems like it'd probably make more sense to get rid of that feature entirely, since there's a lot of experimental features that are available even without it.


hyf0:
I wonder if we could add a check to prevent future regressions

hyf0 (approved):
Thank you!

PR #8112, opened 2026-01-29
feat(lazy-barrel): support incremental build mode

shulaoda:


shulaoda:
* **#8112** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8112?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8112?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=8112).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown/src/module_loader/module_loader.rs
```diff
@@ -911,11 +904,40 @@ impl<'a> ModuleLoader<'a> {
...
+              .unwrap()
+          }
+        }
+      };
+
       tracked_records.retain(|&rec_idx, (import_record_state, resolved_id)| {
         if !needed_records.contains_key(&rec_idx) {
           return true;
         }
+        let barrel_normal_module = unsafe { &mut *barrel_module_ptr };
```
hyf0, line 940: Is this unsafe necessary?

shulaoda: Yes, dereference of raw pointer is unsafe and requires an unsafe function or block

hyf0: I think I knew that. What I'm asking is does the logic have to be implemented with using unsafe?

shulaoda: Yes, this is necessary. As shown below, `barrel_normal_module` and `self.try_spawn_new_task` conflict due to a `&mut` vs `&` borrow. In addition, `barrel_normal_module` is immutably borrowed both before and after the call to `self.try_spawn_new_task` and also mutably borrowed after in full build.
```rust
let new_idx = self.try_spawn_new_task(
  resolved_id.clone(),
  Some(ModuleTaskOwnerRef::new(barrel_normal_module, import_record_state.span)),
  false,
  import_record_state.asserted_module_type.as_ref(),
  user_defined_entries,
)
```

shulaoda: Using a raw pointer is the simplest solution. Otherwise, we would need to refactor and add a significant amount of extra code to work around this. And they ultimately achieve the same result.

hyf0: Don't feel so. I looked into the code and I think the current design has issues.

shulaoda: Another safe implementation: #8114

PR #8149, opened 2026-02-01
refactor: filter empty module_groups before sorting

ShroXd:
Move filtering logic before sorting to avoid unnecessary sorting of empty module groups.


hyf0 (approved):
LGTM

PR #8210, opened 2026-02-05
Add new filter: matchExt

NullVoxPopuli:
It's not easy to match _just the extension_ with the existing utilities, afaict.

Hoping others may find this useful

### DIFF
--- a/packages/pluginutils/src/filter/simple-filters.test.ts
+++ b/packages/pluginutils/src/filter/simple-filters.test.ts
@@ -47,6 +47,7 @@ describe('matchExt', () => {
     expect(regex.test(`app/foo.js.js?f=foo.js`)).toBeTruthy();
     expect(regex.test(`app/js/foo.js`)).toBeTruthy();
     expect(regex.test(`app/js/foo.js`)).toBeTruthy();
+    expect(regex.test(`app/js/foo.js#t=ts`)).toBeTruthy();
   })
 
   test('non-matches: js', () => {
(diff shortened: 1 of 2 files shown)


sapphi-red (commented):
This behavior differs from the common behavior in the ecosystem so I'm hesitant to have this.

IWANABETHATGUY:
> This behavior differs from the common behavior in the ecosystem so I'm hesitant to have this.

Yeah, The functionality of `matchExt` is already covered by `moduleType`.

NullVoxPopuli:
modeleType? Where's that?
- rollup's plugin utils doesn't have it's, which made me think I'm missing out on filter config, but it's not there either: https://rollupjs.org/plugin-development/#conventions
- i did find a maduleType function/class here: https://github.com/rolldown/rolldown/blob/3c8ecb4032bb37a4d5c3541f3246079444631c93/packages/pluginutils/src/filter/composable-filters.ts#L160 but does that is js and wouldn't benefit from the rust speed improvements for filtering, ya? (Or maybe filtering isn't a big deal and it's ok to use js?)

NullVoxPopuli:
is moduleType compatible with rollup?

Asked in discord: https://discord.com/channels/1171783886200643635/1469345267986993242/1469345416654098608

sapphi-red:
To clarify, I think it's totally fine to use your regex in your plugin. I thought it won't be a good idea to add your new function to the pluginutils package unless it's a common behavior. AFAIK many Vite plugins has a behavior aligned with `makeIdFiltersToMatchWithQuery` function.

I think `moduleType` won't cover this function as we need to filter the module by the id to set the `moduleType` to filter we a `moduleType`.

PR #8227, opened 2026-02-07
fix: prevent circular chunk dependencies in manualCodeSplitting

### LINKED ISSUE #7449: [Bug]: incorrect execution order with CJS + advancedChunks

[REPL](https://repl.rolldown.rs/#eNptUE1vwyAM/SsWl7RSlN4zdZf9gEnTbqUHWkjKRkzER9Yp4r/PkKbTtEYBmffsZ/vNrGPtzJw1RtovbM4WO903wWcUWfuIqdmZCD2M1gUI36OCGd5uaa9j0BY9JOicHaBayyuOHEnCB1iE2n8le5g5AmgcY2hhkW8GFURToDqTNobClkwAISeBZyVfLhE//R0G6J2NIwGHFQCaEcWgWqgmhdK6qoagPEntDuJ05Lz58DtIpUn+jrdoQRJH+jmqa9lZqk5Es27CkaWaaZTqSiKrb/f3H7eqZicIJDd+gdMCUKJibXBRkVpOWpVKvKj0xp6Eeb9oT57oQJZttrB/pr1dHjP7a41qjO03VU6otplIT/nOZ7AyEr+sURxPGc7z5ynWjiV+2HGzLVosUcVEtBHZRJZ+AG48wTU=)

### What is expected?

`init` is output when the bundle is executed.

`TypeError: globalThis.init is not a function` error happens when the bundle is executed.

### System Info

```Shell
REPL 1.0.0-beta.53
```

### Any additional comments?

It works correctly when `experimental.strictExecutionOrder: true` is set.

mldangelo:
## Summary

When `includeDependenciesRecursively: false` splits a non-wrapped module into a separate chunk, it can create bidirectional (circular) imports between chunks, causing TDZ `ReferenceError` for `const`/`let` bindings at runtime.

This PR adds **cycle prevention** during manual group formation and keeps a **const/let→var safety net** as defense-in-depth:

### Layer 1: Prevention (new)

After populating module groups, detect if splitting a module would create a circular chunk dependency. If so, keep it in the entry chunk and emit a `MANUAL_CODE_SPLITTING_SKIPPED` warning with actionable suggestions (`strictExecutionOrder: true` or `includeDependenciesRecursively: true`).

Wrapped modules (`WrapKind::Esm`/`Cjs`) are excluded since their lazy initialization (`__esmMin`/`__cjsMin`) already prevents TDZ.

**Scope**: The prevention heuristic catches direct A↔B cycles at group formation time. Transitive chains (A→B→C→A) and cycles emerging after chunk optimization still rely on the safety net.

### Layer 2: Safety net (existing)

For any circular chunk deps that slip through, convert `const`/`let` to `var` in affected chunks. Scoped to ESM format and non-wrapped modules only.

### How cycle detection works

For each non-wrapped module in a manual splitting group:
1. Check if it has dependencies outside the group
2. Check if something outside the group depends on it AND shares a chunk (same `BitSet` bits) with one of its outside deps
3. If both → splitting would create a circular chunk import → remove from group

### Changes

| File | Change |
|------|--------|
| `manual_code_splitting.rs` | Cycle prevention with WrapKind guard, `PreventedManualSplit` tracking, warning emission |
| `chunk_graph.rs` | `chunks_with_circular_deps` + `detect_circular_chunk_deps()` (safety net) |
| `module_finalizers/mod.rs` | const/let→var for ESM non-wrapped chunks with circular deps (safety net) |
| `generate_stage/mod.rs` | Call `detect_circular_chunk_deps()` |
| `event_kind.rs` | New `ManualCodeSplittingSkipped` event kind (separate from `CircularDependency`) |
| `manual_code_splitting_circular_chunk_dependency.rs` | Warning diagnostic with actionable suggestions |
| `cross_chunk_tdz_esm/` | Validates cycle prevention: single chunk, `const` preserved |
| `cross_chunk_tdz_esm_strict_execution_order/` | Validates WrapKind guard: split proceeds with lazy init |
| `cross_chunk_tdz_esm_safety_net/` | Validates const→var safety net fires independently |
| `manual_splitting_no_circular/` | Validates leaf modules still split correctly |

### Test results

- **`cross_chunk_tdz_esm`**: Single chunk, `const` preserved, `MANUAL_CODE_SPLITTING_SKIPPED` warning ✓
- **`cross_chunk_tdz_esm_strict_execution_order`**: Vendor chunk created, wrapped modules split correctly ✓
- **`cross_chunk_tdz_esm_safety_net`**: Two chunks, `var` conversion in affected chunks ✓
- **`manual_splitting_no_circular`**: Vendor chunk created for leaf module ✓
- **`issues/3650`**: Snapshot unchanged ✓
- **Full integration suite**: 733 passed, 0 failed ✓

Fixes #7449


IWANABETHATGUY:
Let's discuss it at https://github.com/rolldown/rolldown/issues/7449 before making a pull request.

crates/rolldown/src/module_finalizers/mod.rs
```diff
@@ -1543,7 +1543,19 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
...
+        // 2. This is an ESM non-wrapped module in a chunk with circular chunk dependencies
+        //    (safety net to prevent TDZ ReferenceErrors from circular chunk imports).
+        //
+        // For case 2: This runs AFTER export stripping (above), so it catches declarations
+        // that were originally `export const` and are now plain `const`. Wrapped modules
+        // (strictExecutionOrder/on-demand wrapping) are immune to TDZ via lazy initializers.
+        let needs_top_level_var = self.ctx.options.top_level_var
+          || (self.ctx.options.format.is_esm()
+            && self.ctx.linking_info.wrap_kind().is_none()
+            && self.ctx.chunk_graph.chunks_with_circular_deps.contains(&self.ctx.chunk_idx));
```
IWANABETHATGUY: I prefer to warn the user and let them decide whether to enable `topLevelVar`. (Because there is not only one way to solve circular references)

IWANABETHATGUY: This is also how rollup handle circularChunks 
https://github.com/rollup/rollup/blob/ae846957f109690a866cc3e4c073613c338d3476/src/Bundle.ts#L234-L273

hyf0:
Just took a rough look to manual code splitting part, I have to say I'm not convinced about what this PR tries to solve.

Feel free to reach me in email or discord for deeper discussions. I will also check this PR again when I got time.

mldangelo-oai:
Thanks again for the earlier review and feedback here. I opened #9226 with a narrower fix for the same Promptfoo/tsdown failure mode: it keeps the generated static chunk graph acyclic in `chunk_optimizer` by checking merges as graph contractions, instead of using the `const`/`let` to `var` TDZ mitigation explored in this PR.

I think #9226 is the better review path now because it addresses the cycle at the optimizer level and includes a reduced regression fixture for #7449 plus the Promptfoo `tsdown@0.21.10` proof. I’m going to treat this PR as superseded by #9226.

PR #8229, opened 2026-02-07
feat: unify `comments` and `legalComments` into a single granular `comments` option

### LINKED ISSUE #5849: [Feature Request]: Allows disabling `comments` in generated modules

Hi, currently when bundling code that contains comments, those comments are preserved in the generated modules. In some cases this is fine, but when generating code for `npm` packages, we typically provide type definitions as well.

Since the `rolldown-plugin-dts` handles type generation and preserves comments in the `.d.ts` files, duplicating those comments in the modules is unnecessary. In fact, they only contribute to increased file size without providing additional value.

It would be great to have an option to disable comment preservation in the generated modules.

Just to clarify, I'm not referring to region comments like `//#region` and `//#endregion`, but rather to `JSDoc` comments.

As a reference, `esbuild` removes comments from output by default.

### comments file

```ts
// comments.ts

/**
 * This is a simple comment example.
 *
 * @default string
 */
export const foo = 'bar'
```

### esbuild

```ts
// build.ts

import { build } from 'esbuild'

await build({
  entryPoints: ['comments.ts'],
  bundle: true,
  format: 'esm',
  outdir: 'dist',
})
```

### output

```js
// comments.ts
var foo = 'bar'
export { foo }
```

---

### rolldown

```ts
import { build } from 'rolldown'

await build({
  input: 'comments.ts',
  output: {
    dir: 'dist',
    format: 'esm',
  },
})
```

### output

```js
//#region comments.ts
/**
 * This is a simple comment example.
 *
 * @default string
 */
const foo = 'bar'

//#endregion
export { foo }
```

As mentioned, the `rolldown-plugin-dts` automatically preserves comments in the generated type definitions, so there's no need to include them in the production version of the module.

Feel free to add the API you think would be most suitable for this, here is the simple example:

```ts
await build({
  input: 'comments.ts',
  output: {
    comments: 'none' // Removes all comments 
  },
})
```

- `'none'` - removes all comments from the output
- `'inline'` - preserves comments marked with special tags (e.g., /*! ... */)
- `'all'` - keeps all comments, including JSDoc and regular inline comments (default value)

IWANABETHATGUY:
closed #5849


IWANABETHATGUY:
* **#8254** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8254?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8229** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8229?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8229?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=8229).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

packages/rolldown/src/options/output-options.ts
```diff
@@ -600,6 +600,18 @@ export interface OutputOptions {
...
+   *
+   * - `'all'`: Preserve all comments including JSDoc and annotation comments (default)
+   * - `'none'`: Strip all non-legal comments (JSDoc, annotations, etc.)
+   *
+   * Legal comments (`@license`, `@preserve`, `//!`, `/*!`) are controlled separately
+   * by the {@linkcode legalComments | output.legalComments} option.
+   *
+   * @default 'all'
+   */
+  comments?: 'none' | 'all';
```
hyf0: @sapphi-red What's your take on having two comment options?

Without looking into the details, I intend to only have one.

packages/rolldown/src/binding.d.cts
```diff
@@ -2316,14 +2323,15 @@ export interface BindingOutputOptions {
...
   legalComments?: 'none' | 'inline'
-  polyfillRequire?: boolean
-  preserveModules?: boolean
-  virtualDirname?: string
-  preserveModulesRoot?: string
-  topLevelVar?: boolean
-  minifyInternalExports?: boolean
-  cleanDir?: boolean
-  strictExecutionOrder?: boolean
+comments?: boolean | { legal?: boolean, annotation?: boolean, other?: boolean }
```
IWANABETHATGUY: The format issue is related to the code generator.

IWANABETHATGUY: fixed

hyf0 (approved):

Issue #8236, opened 2026-02-08
[Panic]: `deferred_scan_data.rs:45` - "Should have resolved id: NotFound"


steffanhalv:
### Panic message

```Shell
# Rolldown Panic Report: `deferred_scan_data.rs:45` - "Should have resolved id: NotFound"

`@rolldown/browser@1.0.0-rc.2` panics during the Vite build process with:


thread 'tokio-runtime-worker' (2) panicked at
  /home/runner/work/rolldown/rolldown/crates/rolldown/src/module_loader/deferred_scan_data.rs:45:12:
Should have resolved id: NotFound("/ecommerce/src/entry-client.js")


The panic occurs in the `defer_sync_scan_data` function when `resolver.resolve()` returns `NotFound` for a module that was **successfully resolved and scanned** during the initial scan phase. The `.expect("Should have resolved id")` on line 45 crashes instead of handling the error gracefully.


### Full console output (in order)


wasi-worker-browser.mjs:19  Rolldown panicked. This is a bug in Rolldown, not your code.

wasi-worker-browser.mjs:19  thread 'tokio-runtime-worker' (2) panicked at
  /home/runner/work/rolldown/rolldown/crates/rolldown/src/module_loader/deferred_scan_data.rs:45:12:

wasi-worker-browser.mjs:19  Should have resolved id: NotFound("/ecommerce/src/entry-client.js")

wasi-worker-browser.mjs:19  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

wasi-worker-browser.mjs:19  Please report this issue at:
  https://github.com/rolldown/rolldown/issues/new?template=panic_report.yml

wasi-threads.esm-bundler.js:257  worker (tid = 44) sent an error!
  Uncaught RuntimeError: unreachable


### WASM stack trace


$wasi_thread_start @ 02a212da:1
ThreadMessageHandler._start @ wasi-threads.esm-bundler.js:890
ThreadMessageHandler.handleAfterLoad @ wasi-threads.esm-bundler.js:930
ThreadMessageHandler.handle @ wasi-threads.esm-bundler.js:847
MessageHandler.handle @ emnapi-core.esm-bundler.js:7609
globalThis.onmessage @ wasi-worker-browser.mjs:38


## Root Cause Analysis

Looking at [`deferred_scan_data.rs`](https://github.com/rolldown/rolldown/blob/main/crates/rolldown/src/module_loader/deferred_scan_data.rs), line 45:


[...]

PR #8238, opened 2026-02-08
fix: deferred scan data resolve panic

steffanhalv:
https://github.com/rolldown/rolldown/issues/8236


sapphi-red (changes requested):
I haven't read though the code, but I don't think this is the right fix. This feels like we're hiding the underlying issue.

Issue #6677, opened 2025-10-24
empty chunk generated and imported when using advancedChunks


hyf0:
### Discussed in https://github.com/rolldown/rolldown/discussions/6657

<div type='discussions-op-text'>

<sup>Originally posted by **windsonR** October 22, 2025</sup>
reproduction: https://repl.rolldown.rs/#eNqVVMFu4yAQ/ZURF28l11GurPaQZO8r9breA8HjrFsCloG0lZV/L4NDbdK0UpEP4xnmzWN4zMhaxkfW6QZfKmfJ1ozP/yWT9HvszeBg3JS7M7SDOUJRraQJXo3a2erRFvVQa2m0NQorZQ4/NnfXnl25Xq/vAiIy7gaP55INRqnGPOsq7Gu7w4LAjUhGxb32CCM8XLb96V0XSkFil9KLWk8kHExA/EPKLxhrDdDp3jsOE3x1RCeq6CopaLyjaNxISzQnoSU2u/9eP9nZD7BaBSSpfIO/scfQRS07tA8o/WC7E6pXDq1QFiPstA6D8b3lf2fPAg/AoXW8ENTjRRKAFkfkxd7rRuG9CAdNgXO27VPE/deI++8jyq8R5RIxmf+SkWqESPhqjS/xnhtshVfp9mrNgmoy4SXB5M5JKxeMcTOLNvbxJ0kzBbdzMLYkCy7lTsFJvfEGAxECS/WjnZWdZLcJAisErXD89I6WsMGdvZLLC0k1iFOqEe0bNbZUY0/riiBxfm/QJ8k7Spa0ApM8PU6B+/X1XJg8+WTYlt+YDdsPs+HGtFhOisDlFIopQUJj5zfdoHcU

It has two question.
1. generate a empty chunk(in project,it maybe not empty) and it is imported.
2. maybe `includeDependenciesRecursively` shoule respect `advancedChunks.group`?</div>

windsonR:
Has there been any progress on this issue?

hyf0:
Haven't checked. Let me make a note of it.

hyf0:
I get the reason why, but I need to fix https://github.com/rolldown/rolldown/issues/7114?issue=rolldown%7Crolldown%7C7117 first as it might potentially fix the issue.

<details>

<summary> For the record </summary>

Normal Code Splitting

  In normal code splitting without advancedChunks:

  Entry: index.js → components.js → a.js, b.js, c.js
[...]

PR #8251, opened 2026-02-09
fix: skip redundant side-effect imports for transitively reachable chunks

Dunqing:
I found this bug while working on https://github.com/tailwindlabs/prettier-plugin-tailwindcss/blob/125a8bc77639529a5a0c7e4e8a02174d7ed2d70b/tsdown.config.ts#L67.

## Summary

- When `manualChunks` moves an entry module into a separate chunk, the entry becomes a facade. The facade was incorrectly getting a direct side-effect import of transitive dependencies already loaded via the symbol import chain (e.g. `import "./resolve.js"` when `sorter-chunk` already imports it).
- Before adding side-effect imports, compute transitively reachable chunks from symbol-based imports by following module-level import records, and skip any chunk already in that transitive set.

**Before:**
```js
// dist/sorter.js (facade)
import "./resolve.js";  // <-- redundant
import { t as createSorter } from "./sorter-chunk.js";
export { createSorter };
```

**After:**
```js
// dist/sorter.js (facade)
import { t as createSorter } from "./sorter-chunk.js";
export { createSorter };
```

## Test plan

- [x] Added test case `code_splitting/redundant_facade_side_effect_import` reproducing the scenario
- [x] All existing Rust tests pass (`just test-rust`)
- [x] Clippy passes

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
+++ b/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
@@ -333,6 +333,7 @@ impl GenerateStage<'_> {
 
   /// - Filter out depended symbols to come from other chunks
   /// - Mark exports of importee chunks
+  #[expect(clippy::too_many_lines)]
   fn compute_chunk_imports(
     &self,
     chunk_graph: &ChunkGraph,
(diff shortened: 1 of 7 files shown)


IWANABETHATGUY:
Can we file an issue first?

Dunqing:
> Can we file an issue first?

I will do it later

IWANABETHATGUY:
https://github.com/rolldown/rolldown/issues/6677 Not sure if they are related. Probably we should treat those issues as a whole and solve them in one go rather than solve them separately, if possible.

Dunqing:
> #6677 Not sure if they are related. Probably we should treat those issues as a whole and solve them in one go rather than solve them separately, if possible.

I can see the same issue in the reproduction of #6677, but we are reporting different bugs. I've created an issue at https://github.com/rolldown/rolldown/issues/8252.

PR #8263, opened 2026-02-10
chore(oxc): prepare next oxc version

Dunqing:
Not ready. Testing https://github.com/oxc-project/oxc/pull/19196, the open state is for triggering the benchmark.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,7 +15,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
 dependencies = [
  "cfg-if",
- "getrandom",
+ "getrandom 0.3.4",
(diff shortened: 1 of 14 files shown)


Dunqing:
https://codspeed.io/rolldown/rolldown/branches/Dunqing%3Aprepare-next-oxc?utm_source=github&utm_medium=check&utm_content=button

image


I believe the change at https://github.com/oxc-project/oxc/pull/19196 should improve performance; however, it shows a slight regression. I don't think this is related to the change, but rather due to other modifications that I am unaware of.

Additionally, some minify-related tests failed or changed the snapshot, cc @sapphi-red

sapphi-red:
I guess you need the change from #8194 for `crates/rolldown/src/utils/pre_process_ecma_ast.rs`.

Boshen:
-1% should be fine, we should be able to get more perf from the Ident change - I'll make some changes in Rolldown to accommodate the changes.

PR #8279, opened 2026-02-11
docs(manual-code-splitting): clarify minSize constraints

ShroXd:
According to the source code of the manual code splitting logic, `minSize` may also cause a chunk's size to exceed `maxSize` when the split chunk is smaller than that threshold. I have clarified this scenario in the documentation.


hyf0 (approved):
Thank you!

PR #8312, opened 2026-02-13
feat(rust/mcs): support `entriesAwareMergeThreshold`

hyf0:
## Summary

This PR adds `entriesAwareMergeThreshold` for advanced chunk groups (`manualCodeSplitting.groups[]`).

It is designed specifically for `entriesAware` mode, where we want to keep the correctness and fetch-precision benefits of entry-aware splitting, but avoid exploding into too many tiny chunks.

---

## What `entriesAware` does

When `entriesAware: true` is enabled on a group, matched modules are split by **entry reachability** (entry-bitset).

That means:
- modules reached by exactly the same entry set go to the same subgroup/chunk
- modules with different reachability bits go to different subgroups/chunks

Example:
- `{A,B,C}`-reachable modules -> one subgroup
- `{A,B}`-reachable modules -> another subgroup
- `{A}`-reachable modules -> another subgroup

### Benefits

- **Most precise loading behavior**: each entry loads only the shared pieces it actually needs.
- **Less over-fetching** compared with coarse “one shared vendor chunk” strategies.
- Works naturally with rolldown’s singleton module model.

### Downsides

- Can generate many small subgroups/chunks, especially in multi-entry apps.
- Too many tiny chunks increase request overhead and scheduling overhead.
- Chunk graph may become noisy/harder to reason about for users.

---

## What `entriesAwareMergeThreshold` does

New per-group option:

```js
manualCodeSplitting: {
  groups: [
    {
      name: 'vendor',
      test: /.../,
      entriesAware: true,
      entriesAwareMergeThreshold: 28000
    }
  ]
}
```

Behavior:
- only applies to `entriesAware` subgroups
- if subgroup `size < entriesAwareMergeThreshold`, it is considered **unqualified**
- unqualified subgroups are **merged** into a better sibling subgroup (same origin group), instead of being left as standalone tiny chunks

---

## How merge chooses a target

For each unqualified subgroup candidate, choose target by:
1. **lowest extra-entry count** (`|candidate-target| + |target-candidate|` on entry bits)
2. if tie: **smaller target size**
3. if tie: stable deterministic key order

This is intentional: the merge heuristic is not only about reducing chunk count, but also about keeping **unnecessary extra download** as low as possible after merging.

By preferring targets with fewer extra entries, we try to merge into the closest reachability neighbor so fewer entries pay for modules they do not strictly need.

---

## Recommended splitting pattern with `maxSize`

`entriesAwareMergeThreshold` and `maxSize` are intended to work together:

1. Use `entriesAwareMergeThreshold` to merge tiny entries-aware subgroups and reduce micro-chunk fragmentation.
2. Keep/use `maxSize` to split any merged result that becomes too large.

So even if merging creates a big group, `maxSize` can re-split it into smaller chunks in the existing splitting phase.

This is the recommended chunk-splitting pattern in practice:
- **merge tiny chunks first** (reduce request overhead)
- **then cap large chunks** (control payload size)
- while still minimizing additional over-fetch / unnecessary download introduced by merge

---

## Relationship with existing size options

- `minSize` and `maxSize` semantics remain unchanged.
- `entriesAwareMergeThreshold` runs as an entries-aware pre-merge step.
- `maxSize` is not a hard constraint in merge target selection.

So this is additive behavior targeted at entries-aware subgroup consolidation.

---

## Why this helps

`entriesAware` gives the best fetch precision but can be too fragmented.

`entriesAwareMergeThreshold` provides a practical middle ground:
- keep entry-aware structure
- reduce tiny subgroup count
- improve chunk graph practicality for real-world apps

In short: **preserve most of entries-aware correctness/precision while reducing micro-chunk overhead**.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/manual_code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/manual_code_splitting.rs
@@ -1,9 +1,14 @@
-use std::{cmp::Reverse, path::Path, sync::Arc};
+use std::{
+  cmp::{Ordering, Reverse},
+  collections::BinaryHeap,
+  path::Path,
+  sync::Arc,
+};
 
 use arcstr::ArcStr;
 use oxc_index::IndexVec;
 use rolldown_common::{
-  Chunk, ChunkKind, ChunkingContext, MatchGroupTest, Module, ModuleIdx, ModuleTable,
+  Chunk, ChunkKind, ChunkingContext, MatchGroup, MatchGroupTest, Module, ModuleIdx, ModuleTable,
 };
 use rolldown_error::BuildResult;
 use rolldown_utils::{BitSet, xxhash::xxhash_with_base};
@@ -25,6 +30,7 @@ struct ModuleGroup {
   modules: FxHashSet<ModuleIdx>,
   priority: u32,
   sizes: f64,
+  entries_aware_bits: Option<BitSet>,
 }
 
 /// Used to track what created the module group
@@ -84,6 +90,8 @@ impl GenerateStage<'_> {
 
     let mut index_module_groups: IndexVec<ModuleGroupIdx, ModuleGroup> = IndexVec::new();
     let mut group_idx_by_id: FxHashMap<ModuleGroupId, ModuleGroupIdx> = FxHashMap::default();
+    let mut entries_aware_groups_by_origin: FxHashMap<ModuleGroupOrigin, Vec<ModuleGroupIdx>> =
+      FxHashMap::default();
     let metas = &self.link_output.metas;
     for normal_module in self.link_output.module_table.modules.iter().filter_map(Module::as_normal)
     {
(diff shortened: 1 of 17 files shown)


hyf0:
* **#8313** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8313?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8312** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8312?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8312?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8311** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8311?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=8312).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

IWANABETHATGUY (approved):

PR #8317, opened 2026-02-13
feat(rust/register_hook): add RegisterHook macro

JayFate:
feat(rust/macros): add RegisterHook proc-macro to auto-generate plugin hook usage

## Summary

Introduce `rolldown_plugin_register_macro`, a procedural macro that generates `register_hook_usage` for Rolldown plugins.

It reduces boilerplate and prevents mismatches between implemented hooks and declared hook usage.

---

## Problem

Rolldown plugins must manually maintain `register_hook_usage` to declare which hooks they implement.

This is:
- easy to forget when adding/removing hook methods
- tedious across many plugins
- a common source of incorrect hook registration

---

## What `#[RegisterHook]` does

When applied to an `impl Plugin for T` block, `#[RegisterHook]`:
- analyzes which hook methods are implemented in the impl
- generates `fn register_hook_usage(&self) -> HookUsage` automatically
- returns a bitflag union of the detected hooks (e.g. `HookUsage::BuildStart | HookUsage::Transform`)

---

## Usage

Add the attribute to your plugin impl:

```rust
#[RegisterHook]
impl Plugin for MyPlugin {
  async fn build_start(...) -> HookNoopReturn { ... }
  async fn transform(...) -> HookTransformReturn { ... }
}
```

Generated output:

```rust
fn register_hook_usage(&self) -> HookUsage {
  HookUsage::BuildStart | HookUsage::Transform
}
```


IWANABETHATGUY:
**Positive:**
- Removes ~277 lines of boilerplate
- Prevents hook registration bugs

**Negative:**
- Adds `syn`, `quote`, `proc-macro2` to the dependency graph
- Every clean build pays the cost of compiling the proc-macro crate
- The boilerplate it removes was simple and mechanical — not a major maintenance burden

The built-in plugin is already pretty stable, so there will not be too much work needed on the plugin side. 

Overall, the reward did not outweigh the cost.

JayFate:
> register_hook_usage



> **Positive:**
> 
> * Removes ~277 lines of boilerplate
> * Prevents hook registration bugs
> 
> **Negative:**
> 
> * Adds `syn`, `quote`, `proc-macro2` to the dependency graph
> * Every clean build pays the cost of compiling the proc-macro crate
> * The boilerplate it removes was simple and mechanical — not a major maintenance burden
> 
> The built-in plugin is already pretty stable, so there will not be too much work needed on the plugin side.
> 
> Overall, the reward did not outweigh the cost.

Compile time BEFORE  this commit:

```
sh -c 'cargo clean && CARGO_INCREMENTAL=0 just build-rolldown'  359.48s user 59.40s system 411% cpu 1:41.84 total
```

AFTER  this commit:

```
sh -c 'cargo clean && CARGO_INCREMENTAL=0 just build-rolldown'  361.31s user 60.35s system 404% cpu 1:42.82 total
```

No noticeable increase in compile time was observed.

Leveraging the compiler’s code generation capability is superior to manual coding by developers.

It avoids having to manually modify `register_hook_usage` every time hooks are added or removed from existing code.

It also removes the need to write `register_hook_usage` from scratch for newly added code.

crates/rolldown_plugin_esm_external_require/src/lib.rs
```diff
@@ -14,21 +14,12 @@ pub struct EsmExternalRequirePlugin {
...
 
-  fn register_hook_usage(&self) -> HookUsage {
-    if self.external.is_empty() {
-      HookUsage::empty()
-    } else if self.skip_duplicate_check {
-      HookUsage::ResolveId | HookUsage::Load
-    } else {
-      HookUsage::BuildStart | HookUsage::ResolveId | HookUsage::Load
-    }
-  }
```
IWANABETHATGUY, line 30: @shulaoda I suspect whether the macro could generate correct code for this scenario.

hyf0:
Appreciate your efforts. The reason we didn't do this is that current mechanism works well and we don't want add compleixity on this part.

PR #8438, opened 2026-02-24
refactor(bindingify-build-hook): extract helper for PluginContextImpl

ShroXd:
Create a shared `createPluginContext` helper to encapsulate repeated `PluginContextImpl` construction and update all build hooks to use it.


hyf0 (approved):
Thank you!

PR #8487, opened 2026-02-27
perf(rust/tla): skip compute_tla if there is no module use TLA

ShroXd:
A todo in `compute_tla`  indicates that this process could be skipped when there is no module that uses TLA — which is also the common case for many real-world projects.

This PR tracks the count of TLA-using modules during the scan stage and stores it in `tla_module_count`. Link stage will check this counter and return early if the count is zero.

### DIFF
--- a/crates/rolldown/src/module_loader/module_loader.rs
+++ b/crates/rolldown/src/module_loader/module_loader.rs
@@ -11,11 +11,11 @@ use oxc_allocator::Address;
 use oxc_index::IndexVec;
 use rolldown_common::dynamic_import_usage::DynamicImportExportsUsage;
 use rolldown_common::{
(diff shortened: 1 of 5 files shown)


hyf0:
Is count a bit of unnecessary?

I would go for having a big flag under module scanner and return a `has_module_used_tla` bitflag/boolean in the final ModuleLoaderOutput.

ShroXd:
> Is count a bit of unnecessary?
> 
> I would go for having a big flag under module scanner and return a `has_module_used_tla` bitflag/boolean in the final ModuleLoaderOutput.

Hi, thanks for the feedback.

I think the counter is necessary because it makes it easier to check if any other TLA modules __remain__ in the `scan_stage_cache.rs:merge` method. A bool or bit flag cannot tell us whether other modules still have TLA after one TLA is removed.

hyf0:
Well done!

hyf0 (approved):

PR #8493, opened 2026-02-27
ci: add s390x and ppc64le linux gnu targets

Brooooooklyn:
> [!NOTE]
> **Low Risk**
> Low risk: CI/workflow and packaging metadata changes only; main risk is potential release-build failures for the new architectures.
> 
> **Overview**
> Expands the release build matrix to produce `@rolldown/binding` artifacts for **Linux `s390x-unknown-linux-gnu`** and **`powerpc64le-unknown-linux-gnu`** (using `--use-napi-cross`, with `s390x` forcing `CFLAGS=-fuse-ld=lld`).
> 
> Updates tooling versions used in CI by bumping Zig for musl builds to `0.15.2` and the FreeBSD VM version to `15.0`, and adds the two new targets to `packages/rolldown/package.json` `napi.targets` so they are included in published binaries.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 73b60668b3775c9a5378744d38f6561f462f707a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #8495, opened 2026-02-27
docs: list s390x/ppc64le prebuilt binaries

crusty-voidzero:
Adds the new Linux prebuilt binary targets to the "Using a minor platform" installation docs.

Depends on #8493.


hyf0 (approved):
Thank you!

PR #8497, opened 2026-02-28
perf(plugin): use IndexBitSet to optimize skipped plugins checking

ShroXd:
Replace `Vec<PluginIdx>` used to track skipped plugins in the `resolve_idx` and `resolve_dynamic_import` with the `IndexBitSet`, turning $O(n)$ scans into $O(1)$ bit lookups.


hyf0 (approved):
Well done!

PR #8527, opened 2026-03-02
test: module graph

Dunqing:
Open PR for triggering Benchmark CI


Dunqing:
Verifying with a general module graph would not significantly impact performance. However, since Rolldown's module graph is closely tied to its specific fields, adopting a general module graph could require extensive refactoring. Given that Rolldown is still rapidly evolving, maintaining an external module graph may prove challenging.

PR #8536, opened 2026-03-03
chore(vscode): use oxc formatter for markdown files

minsoo-web:
## Summary
- Configure workspace Markdown formatting to use `oxc.oxc-vscode`.
- Enable format-on-save for Markdown in workspace settings.
- Ensure Markdown uses the workspace OXC formatter on save so user-level default formatter preferences (including Prettier) do not rewrite quote style in this repo.

## Test plan
- [x] Save a Markdown file in this workspace and confirm the active formatter is OXC.
- [x] Verify `.vscode/settings.json` includes `[markdown]` override with `editor.defaultFormatter` and `editor.formatOnSave`.
- [x] Confirm no other file changes are introduced by this update.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -19,6 +19,10 @@
     "editor.formatOnSave": true,
     "editor.defaultFormatter": "oxc.oxc-vscode"
   },
+  "[markdown]": {
+    "editor.defaultFormatter": "oxc.oxc-vscode",
+    "editor.formatOnSave": true
+  },
   "files.associations": {
     "*.snap": "markdown",
     "*.snap.new": "markdown",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -20,8 +20,8 @@
     "editor.defaultFormatter": "oxc.oxc-vscode"
   },
   "[markdown]": {
-    "editor.defaultFormatter": "oxc.oxc-vscode",
-    "editor.formatOnSave": true
+    "editor.formatOnSave": true,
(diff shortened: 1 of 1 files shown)


minsoo-web:
Thanks for calling this out. The intent of this change is not to disable Prettier globally.

By setting `[markdown].editor.defaultFormatter` to `oxc.oxc-vscode` and enabling `[markdown].editor.formatOnSave`, Markdown files in this workspace use OXC on save. That prevents user-level default formatter preferences (including Prettier) from taking over Markdown formatting in this repo.

I'll also update the PR description to make this behavior explicit.

hyf0:
Could you check if `oxfmt` command will format markdown files? If we're using to use oxfmt to format markdown, it should be enabled always, not just from vscode.

minsoo-web:
@hyf0 Thanks for the suggestion. I verified that formatting is re-applied by oxfmt during commit via lint-staged, so formatting consistency is already enforced at commit time.

My intention with this PR was mainly editor-side ergonomics: when someone uses a different formatter in their editor (e.g. Prettier), Markdown formatting in this repo can look inconsistent before commit. The VS Code workspace setting helps align the local editing experience with the repo formatter.

That said, since commit-time enforcement is already in place, I’m also okay to drop this change if we prefer to avoid editor-specific config.

What do you think?

hyf0:
Having IDE support is good. I like it. Thanks for your effrots.

hyf0 (approved):

PR #8537, opened 2026-03-03
fix(binding): export BindingResult in generated dts header

minsoo-web:
## Summary
- export `BindingResult` in `packages/rolldown/package.json` `napi.dtsHeader`
- regenerate `packages/rolldown/src/binding.d.cts` so the generated type is also exported
- remove `[MISSING_EXPORT] \"BindingResult\" is not exported by \"src/binding.d.cts\"` warning during `just test-node`

## Test plan
- [x] `just build-rolldown-binding`
- [x] `SKIP_PASS_DIFF=1 just test-node`
- [x] `pnpm --filter rolldown run build-js-glue`

### DIFF
--- a/packages/rolldown/package.json
+++ b/packages/rolldown/package.json
@@ -144,7 +144,7 @@
   },
   "napi": {
     "binaryName": "rolldown-binding",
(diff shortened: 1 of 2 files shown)


hyf0 (approved):
Thank you!

PR #8540, opened 2026-03-04
refactor(rust): remove FacadeScoping, use Scoping::create_symbol for facade symbols

Dunqing:
## Summary

- Remove the `FacadeScoping` struct and `minimum_symbol_id` boundary that separated "real" AST symbols from synthetic "facade" symbols using reverse-counting IDs (`u32::MAX - 1` downward)
- Use oxc's `Scoping::create_symbol()` API instead, giving facade symbols contiguous forward-counting IDs in the same symbol table
- Add `SymbolRefFlags::IsFacade` flag to identify facade symbols, replacing the old boundary-based check
- Simplify `get_classic_data` from boundary-check + HashMap fallback to direct `IndexVec` indexing
- Delete `SymbolIdExt` trait (hardcoded `u32::MAX - 2` for namespace detection), add `Module::namespace_object_ref()` API
- Add `SymbolRefDbForModule::merge_from_build()` for correct incremental build cache handling

## Test plan

- [x] `cargo clippy -p rolldown -p rolldown_common` — no errors
- [x] esbuild integration tests — 834 passed, 0 failed
- [x] rollup integration tests — 47 passed, 0 failed
- [x] rolldown integration tests — 744 passed, 1 failed (pre-existing `util_deprecate`)
- [x] HMR incremental build test `no_accept_outside_circular` — passes


crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
```diff
@@ -406,32 +405,19 @@ impl GenerateStage<'_> {
             {
               for dynamic_entry_module in set {
                 let meta = &self.link_output.metas[*dynamic_entry_module];
-                match meta.wrap_kind() {
```
IWANABETHATGUY, line 409: I prefer the previous code, since it provides more context.

crates/rolldown_common/src/ecmascript/symbol_id_ext.rs
```diff
@@ -1,28 +0,0 @@
...
-  fn module_namespace_symbol_ref(module_idx: ModuleIdx) -> SymbolRef;
-}
-
-impl SymbolIdExt for SymbolId {
-  /// Returns true if the symbol id is the special module namespace symbol.
-  /// `namespace_object_ref` is always created in second times with `create_facade_root_symbol_ref`,
-  ///  see: https://github.com/rolldown/rolldown/blob/8bc7dca5a09047b6b494e3fa7b6b7564aa465372/crates/rolldown/src/ast_scanner/mod.rs?plain=1#L156-L160
-  #[inline]
-  fn is_module_namespace(&self) -> bool {
-    *self == SymbolId::from_raw_unchecked(u32::MAX - 2)
```
IWANABETHATGUY, line 18: Actually this is a trick to speed up checking if a symbol is a module namespace..., and can verify it is correct.

crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
IWANABETHATGUY: It looks fine to merge except for this one.

Dunqing:
This PR aims to standardize the handling of facade symbols, making it easier to understand. 

### Before:
The old `FacadeScoping` used reverse-counting IDs (`u32::MAX - 1` downward) to keep synthetic "facade" symbols separate from real AST symbols. This was fragile — it relied on a hardcoded `u32::MAX - 2` to identify namespace symbols, stored facade data in a HashMap instead of the contiguous IndexVec, and required boundary checks on every `get_classic_data` call.

### After:
This PR switches to oxc's `Scoping::create_symbol()`, which gives facade symbols normal forward-counting IDs in the same table. A `SymbolRefFlags::IsFacade` flag replaces the boundary-based detection. The result is simpler data access (direct indexing instead of boundary check + HashMap fallback), no magic constants, and proper use of upstream APIs.


image

Additionally, we achieved a 0.5% performance improvement. Also, this change allows us to use `IndexVec` instead of `FxHashMap`, since it no longer has a Max::u32-sized Vec.

crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
Dunqing: Thank you for reviewing! I've reverted.

IWANABETHATGUY (approved):

hyf0 (approved):

PR #8544, opened 2026-03-04
fix(link_stage): only determine exports kind for js/jsx/ts/tsx modules

ShroXd:
Resolve the todo in `determine_module_exports_kind` by filtering out non-JS modules (only `Js | Jsx | Ts | Tsx` are processed now). I didn't find any similar usage in the codebase, therefore writing an inline filter instead of creating helper function.

Most of the line changes in this PR are due to auto-formatting, the real logic changes only include L19 to L21.


hyf0:
Thanks for you effrot. The todo is added when supporting builtin css, but now css had been removed. There's no need to check if module is js anymore, all modules would be js definitely.

ShroXd:
> Thanks for you effrot. The todo is added when supporting builtin css, but now css had been removed. There's no need to check if module is js anymore, all modules would be js definitely.

Hi, thanks for the explanation. I will close this PR.

Issue #8553, opened 2026-03-05
Improve CI timings


Boshen:
## Goal
Reduce required CI wall-clock time for faster PR merges, while preserving current required coverage in the first pass.

## Current Required Merge Flow
```mermaid
flowchart TD
  A[Detect Changes] --> B[Rust Validation]
  A --> C[Repo Validation]
  A --> D[Spell Check]
  A --> E[cargo-test matrix: ubuntu/macos/windows]
  A --> F[build-rolldown-native matrix: ubuntu/macos/windows]
  A --> I[Node Validation]

  F --> J[node-test: windows/macos/ubuntu]
  F --> K[node-dev-server-test: windows/macos/ubuntu]
  F --> L[Type Check]
```

## Workflow Files (Hotspots)
- `.github/workflows/ci.yml`
- `.github/workflows/reusable-node-test.yml`
- `.github/workflows/reusable-node-dev-server-test.yml`
- `.github/workflows/reusable-cargo-test.yml`
- `.github/workflows/reusable-native-build.yml`

## Timing Baseline (Pre-optimization)
Measured from runs:
- `22699519602`
- `22678505235`

Total required CI wall-clock:
- ~11.23 min
- ~11.38 min

### Job Family Averages (Pre-optimization)
| Job family | Avg duration | Max duration |
|---|---:|---:|
| `node-dev-server-test` | 5.69m | 7.32m |
| `cargo-test` | 4.98m | 6.23m |
| `node-test` | 4.38m | 4.75m |
| `Node Validation` | 4.35m | 4.43m |
| `Build @rolldown/browser` | 4.23m | 4.60m |
| `build-rolldown-wasi` | 4.01m | 4.22m |
| `build-rolldown-native` | 2.83m | 3.88m |

## Progress

| PR | Title | Status | Slow steps addressed | Estimated savings |
|---|---|---|---|---|
| #8554 | ci: dedupe type-check from dev server workflow | ✅ Merged | `node-dev-server-test` (avg 5.69m, worst 7.32m) — removed redundant `pnpm type-check` step | Reduces the slowest CI lane; type-check already covered in `node-test-*` jobs |
[...]

PR #8566, opened 2026-03-06
ci: skip redundant native binding build for browser and remove standalone job

Boshen:
Part of https://github.com/rolldown/rolldown/issues/8553

## Summary

- Remove `binding|` from `build-browser-pkg:debug` and `build-browser-pkg:release` scripts so browser builds only run WASI + node steps, skipping ~60-90s of unnecessary native Rust compilation (the `.node` file is marked as `external` in `copy-addon-plugin.ts` and never bundled)
- Remove standalone `build-browser` CI job since `node-validation` already builds and validates `@rolldown/browser` with publint/knip

https://github.com/rolldown/rolldown/blob/36b69d8676c5a5960551b7af3f06957144c80bd9/.github/workflows/ci.yml#L262-L263

**Note:** `build-browser` may need to be removed from GitHub branch protection required status checks after merging.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -182,31 +182,6 @@ jobs:
       changed: ${{ needs.changes.outputs.node-changes == 'true' }}
       os: ubuntu-latest
 
-  build-browser:
-    name: Build `@rolldown/browser`
-    needs: [changes]
-    if: ${{ needs.changes.outputs.node-changes == 'true' }}
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-        with:
-          submodules: true # Pull submodules for additional files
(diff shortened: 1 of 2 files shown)


Boshen:
@hyf0 we need to fix required jobs if this makes sense.

hyf0:
I just recalled. The purpose of this job is ensuring each commit doesn't hurt browser build.

hyf0:
I just found WASI test already builds the required browser binary. ~~Maybe we could follow what we did for node tests, making~~ It seems we already it, `.wasi` build is shared by multiple platform tests, no duplicate build happens, we could apply the same for the `browser-build` too.

https://github.com/rolldown/rolldown/blob/36b69d8676c5a5960551b7af3f06957144c80bd9/.github/workflows/reusable-wasi-build.yml#L37-L44

I think we should keep it and reuse the effort of build wasi(This still requires some cumbersome change to the CI)

PR #8576, opened 2026-03-07
test(hmr): fix infinite loop in dev server test retry logic

hyf0-agent:
## Summary

Fix an infinite loop bug in the dev server fixture test retry mechanism.

The retry directory naming used a fixed `-retry` suffix:
```js
while (await isDirectoryExists(tmpProjectPath)) {
  tmpProjectPath = nodePath.join(tmpFixturesPath, fixtureName + '-retry');
}
```

When both the original directory and `-retry` directory existed (after 2 failed attempts), the while loop would repeatedly resolve to the same existing path → infinite loop → test times out at 90s.

This made vitest's `retry: 3` config effectively useless beyond the first retry attempt.

### Fix

Use an incrementing counter: `${fixtureName}-retry-${retryIndex}`


hyf0 (approved):
Small change, gonna just merge it.

PR #8602, opened 2026-03-09
feat(plugin): support addWatchGlob

### LINKED ISSUE #8491: [Feature Request]: Add glob support for watcher

`this.addWatchFile` does not currently support globs. While tsdown supports glob entries, the watcher has no way of detecting changes to files matched by those globs when new entries are added.

Relate issue https://github.com/rolldown/tsdown/issues/709, https://github.com/rolldown/tsdown/issues/720

`this.addWatchFile(globPatterns: string | string[])` or `this.addWatchGlob(globPatterns: string | string[])`

ShroXd:
Resolve #8491 

## Context

Although Rolldown provides `addWatchFile` api, like the original issue and related issues mentioned, it would be useful to add a new API `addWatchGlob` allows glob watching.

## Implementation

Add a new API: `addWatchGlob(globPattern: string)`. Follow the existing pattern of `addWatchFile`, if multiple globs need to be watched, users need to add them one by one.

`watch_task.rs` holds core logic. For all passed globs, extract the static base directory and register those directories to `fs_watcher` after each build.

When Notify emits a file system event, `matches_watch_globs` check the event path against `watch_globs`. If matched, a rebuild will be triggered.

## Test

Follow the existing pattern, add some tests in `watch.test.ts` to cover:

- Modify existing file triggers rebuild
- Create new file matching glob triggers rebuild
- Modifying non-matching file should not trigger rebuild
- Deleting a matching file triggers rebuild

## Following work

Original issue and [API usage in tsdown](https://github.com/rolldown/tsdown/blob/2b1be7e0accdb1d6e903cd42642666431fc5e93d/src/features/watch.ts#L24-L26) show it maybe worth to support users provide multiple globs or files one time. But if we really want to support it, it would be better to do in a separated PR.

Documentation updating would be included in a separated PR once maintainer think this PR is good to merged, so that we have enough information about the API behavior to write the document.

### DIFF
--- a/crates/rolldown/src/bundle/bundle.rs
+++ b/crates/rolldown/src/bundle/bundle.rs
@@ -143,6 +143,10 @@ impl Bundle {
     &self.plugin_driver.watch_files
   }
 
+  pub fn get_watch_globs(&self) -> &Arc<FxDashSet<ArcStr>> {
+    &self.plugin_driver.watch_globs
+  }
+
   pub fn context(&self) -> BundleHandle {
     BundleHandle {
       options: Arc::clone(&self.options),
--- a/crates/rolldown/src/bundle/bundle_handle.rs
+++ b/crates/rolldown/src/bundle/bundle_handle.rs
@@ -60,6 +60,14 @@ impl BundleHandle {
     &self.plugin_driver.watch_files
   }
 
+  /// Get the watch globs collected during this bundle.
+  ///
+  /// These patterns (already normalized to absolute form) are matched against
+  /// every path that `notify` emits. Any match triggers a rebuild.
+  pub fn watch_globs(&self) -> &Arc<rolldown_utils::dashmap::FxDashSet<arcstr::ArcStr>> {
+    &self.plugin_driver.watch_globs
+  }
+
   /// Get the plugin driver used in this bundle.
   ///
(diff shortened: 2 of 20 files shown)


hyf0:
We haven't discussed if we want to add this API. I feel like this could be done in the userland, I don't see huge benefits of writting it in rust. @sapphi-red cc

ShroXd:
> We haven't discussed if we want to add this API. I feel like this could be done in the userland, I don't see huge benefits of writting it in rust. @sapphi-red cc

Hi, thanks for the feedback!

I agree with you. Users could support this functionality on their side, and that is exactly what tsdown is doing now. My thought is that this could provide a clean and easy-to-use API, avoiding the need to transform globs into watched files in TypeScript. The fast-glob and notify combo could also provide faster processing speed.

That said, I respect your design decision regarding Rolldown's API. If the Rolldown team thinks `addWatchGlob` is unnecessary, please let me know and I will close this PR.

sapphi-red:
While I haven't thought about the API, I'd point out that it's not possible to do this on userland for incremental builds. For example, there's no way to tell that a module depends on `**/foo/*.js`. When `foo/a.js` is created, without this API or other kinds of API, it's not possible to trigger the rebuild.

PR #8609, opened 2026-03-09
perf: replace FxHashMap with IndexVec for sequential index keys

Dunqing:
## Summary

- Replace `FxHashMap` with `IndexVec` where keys are sequential index types (`ModuleIdx`, `ChunkIdx`, `SymbolId`, `ImportRecordIdx`) to eliminate hash overhead and improve cache locality
- **`SymbolRefDbForModule::flags`**: `FxHashMap<SymbolId, u8>` → `IndexVec<SymbolId, SymbolRefFlags>` — highest impact change, accessed in hot union-find paths. The `u8` bitflags cost 1 byte/symbol in IndexVec vs ~40+ bytes/entry in FxHashMap
- **`ChunkGraph`**: 4 `FxHashMap<ChunkIdx, ...>` fields → `IndexVec<ChunkIdx, ...>` — chunk count is small, trivial wins
- **`LinkStageOutput`**: 3 `FxHashMap<ModuleIdx, ...>` → `IndexVec<ModuleIdx, Option<...>>` — module count known at link stage
- **`ScanResult`**: 3 per-module `FxHashMap` fields → `IndexVec<_, Option<_>>` — sized from import record/symbol counts

## Test plan

- [x] `cargo check -p rolldown_common -p rolldown` passes
- [x] `cargo clippy -p rolldown_common -p rolldown` passes with no new warnings
- [ ] CI tests pass

### DIFF
--- a/crates/rolldown/src/ast_scanner/mod.rs
+++ b/crates/rolldown/src/ast_scanner/mod.rs
@@ -107,15 +107,15 @@ pub struct ScanResult {
   pub hashbang_range: Option<Span>,
   /// we don't know the ImportRecord related ModuleIdx yet, so use ImportRecordIdx as key
   /// temporarily
-  pub dynamic_import_rec_exports_usage: FxHashMap<ImportRecordIdx, DynamicImportExportsUsage>,
+  pub dynamic_import_rec_exports_usage: IndexVec<ImportRecordIdx, Option<DynamicImportExportsUsage>>,
   /// `new URL('...', import.meta.url)`
   pub new_url_references: FxHashMap<Span, ImportRecordIdx>,
   pub this_expr_replace_map: FxHashMap<Span, ThisExprReplaceKind>,
   pub hmr_info: HmrInfo,
   pub hmr_hot_ref: Option<SymbolRef>,
   pub directive_range: Vec<Span>,
-  pub constant_export_map: FxHashMap<SymbolId, ConstExportMeta>,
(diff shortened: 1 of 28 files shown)


hyf0:
Is there a way to verify these changes' impact? Or they're just better in common sense.

crates/rolldown/src/ast_scanner/mod.rs
```diff
@@ -711,10 +714,9 @@ impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
         );
       });
       if let Some(ref with_clause) = decl.with_clause {
-        self
-          .result
-          .import_attribute_map
-          .insert(record_idx, ImportAttribute::from_with_clause(with_clause));
+        self.result.import_attribute_map.resize(record_idx.index() + 1, None);
```
hyf0, line 717: Is this using `IndexVec` to simulate Map structure? Wouldn't this causes a lot of empty slots?

Dunqing: It would. This is an unexpected change. My intent is to replace those FxHashMaps that are always full.

Dunqing:
Considering there is no performance change, so close it to avoid breaking anything.

Dunqing:
> Is there a way to verify these changes' impact? Or they're just better in common sense.

It would be better at accessing since there is no need for a hash lookup.

Issue #7522, opened 2025-12-15
Native Magic String Support


sxzz:
- missing properties
  - [x] clone https://github.com/rolldown/rolldown/pull/7813
  - [x] move https://github.com/rolldown/rolldown/pull/7796
  - [x] reset https://github.com/rolldown/rolldown/pull/7828
  - [x] **slice** (important) https://github.com/rolldown/rolldown/pull/7807
  - [x] snip https://github.com/rolldown/rolldown/pull/7818
  - [x] trim / trimStart / trimEnd / trimLines https://github.com/rolldown/rolldown/pull/7800
  - [x] lastChar / lastLine https://github.com/rolldown/rolldown/pull/7819
  - [x] original
  - [x] offset
- [x] Add a way to detect whether a magic-string instance is native or not, without importing rolldown (since rolldown may not be installed). Current I use `s.constructor.name === 'BindingMagicString'` in [rolldown-string](https://github.com/sxzz/rolldown-string/blob/23b2270d3cb9da8b9cc705b3e148c20888724f24/src/index.ts#L74), which is fragile https://github.com/rolldown/rolldown/pull/8614.
  - Consider set `Symbol.toStringTag`, see [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag)
- ~~Support native magic-string on `renderChunk` hook~~ => #7934

PR #8614, opened 2026-03-10
feat(rolldown): add isRolldownMagicString property for reliable native detection

IWANABETHATGUY:
## Summary

External packages like `rolldown-string` need to detect native `BindingMagicString` instances, but the existing `obj.constructor.name` check breaks under minification or bundling. This PR adds a reliable `isRolldownMagicString` boolean property to the `BindingMagicString` prototype via `Object.defineProperty` (non-writable, non-configurable). A new `binding-magic-string.ts` module centralizes the patching and re-exports the class with proper TypeScript typings. All internal imports now go through this module to ensure consistency. Tests cover prototype presence, instance access, and detection without importing rolldown.

## Related

https://github.com/rolldown/rolldown/issues/7522

### DIFF
--- a/packages/rolldown/src/binding-magic-string.ts
+++ b/packages/rolldown/src/binding-magic-string.ts
@@ -1,0 +1,23 @@
+import { BindingMagicString as NativeBindingMagicString } from './binding.cjs';
+
+// Set `isRolldownMagicString` so external packages (e.g. rolldown-string) can
+// detect native BindingMagicString instances without importing rolldown:
+//   obj.isRolldownMagicString === true
+// This replaces the fragile `obj.constructor.name` check which breaks with
+// minification or bundling.
+Object.defineProperty(NativeBindingMagicString.prototype, 'isRolldownMagicString', {
+  value: true,
+  writable: false,
+  configurable: false,
+});
+
+export interface BindingMagicString extends NativeBindingMagicString {
+  readonly isRolldownMagicString: true;
+}
+
+type BindingMagicStringConstructor = Omit<typeof NativeBindingMagicString, 'prototype'> & {
+  new (...args: ConstructorParameters<typeof NativeBindingMagicString>): BindingMagicString;
+  prototype: BindingMagicString;
+};
+
+export const BindingMagicString = NativeBindingMagicString as BindingMagicStringConstructor;
--- a/packages/rolldown/tests/magic-string/magic-string-to-string-tag.test.ts
+++ b/packages/rolldown/tests/magic-string/magic-string-to-string-tag.test.ts
@@ -1,0 +1,29 @@
+import assert from 'node:assert';
+import { BindingMagicString as MagicString } from 'rolldown';
+import { describe, it } from 'vitest';
+
+describe('MagicString isRolldownMagicString', () => {
+  it('should have isRolldownMagicString on prototype', () => {
+    assert.strictEqual(MagicString.prototype.isRolldownMagicString, true);
+  });
+
+  it('should be accessible on instances', () => {
+    const s = new MagicString('hello');
+    assert.strictEqual(s.isRolldownMagicString, true);
+  });
+
+  it('should allow detection without importing rolldown', () => {
+    const s = new MagicString('hello');
+    // This is the primary use case: external packages can detect native
+    // BindingMagicString instances using isRolldownMagicString instead of
+    // the fragile `s.constructor.name === 'RolldownMagicString'`
+    const isNative = (obj: unknown): boolean =>
+      typeof obj === 'object' &&
+      obj !== null &&
+      (obj as Record<string, unknown>).isRolldownMagicString === true;
+    assert.ok(isNative(s));
+    assert.ok(!isNative({}));
+    assert.ok(!isNative('string'));
+    assert.ok(!isNative(null));
+  });
+});
--- a/packages/rolldown/src/index.ts
+++ b/packages/rolldown/src/index.ts
@@ -3,6 +3,7 @@ import { build, type BuildOptions } from './api/build';
 import { rolldown } from './api/rolldown';
 import type { RolldownBuild } from './api/rolldown/rolldown-build';
 import { watch } from './api/watch';
+import { BindingMagicString } from './binding-magic-string';
 import type {
   RolldownWatcher,
   RolldownWatcherEvent,
@@ -118,7 +119,7 @@ import type { BundleError } from './utils/error';
 
 export { RUNTIME_MODULE_ID, VERSION } from './constants';
 export { build, defineConfig, rolldown, watch };
-export { BindingMagicString } from './binding.cjs';
+export { BindingMagicString };
 export type {
   AddonFunction,
   BundleError,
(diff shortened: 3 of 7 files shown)


IWANABETHATGUY:
* **#8614** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8614?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8614?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=8614).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

IWANABETHATGUY:
### Merge activity

* **Mar 10, 5:03 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Mar 10, 2:54 PM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Mar 10, 3:00 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

packages/rolldown/tests/magic-string/magic-string-to-string-tag.test.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  it('should allow detection without importing rolldown', () => {
+    const s = new MagicString('hello');
+    // This is the primary use case: external packages can detect native
+    // BindingMagicString instances using Symbol.toStringTag instead of
+    // the fragile `s.constructor.name === 'BindingMagicString'`
+    const isNative = (obj: unknown): boolean =>
+      typeof obj === 'object' &&
+      obj !== null &&
+      (obj as Record<symbol, unknown>)[Symbol.toStringTag] === 'BindingMagicString';
```
hyf0: If we're gonna make this like an official contract, could we chose some more serious name than `BindingMagicString`?

IWANABETHATGUY: Ah yeah, did not think too much about it because this is really a niche usage. So directly use suggestion from https://github.com/rolldown/rolldown/issues/7522 

What about `NativeMagicString`?

IWANABETHATGUY: CC recommended `RolldownMagicString` because it clearly shows the origin. It's unambiguous and won't conflict with anything else. This is my suggestion. I believe that `RolldownMagicString` is better than `NativeMagicString` for library developers to identify whether the constructor is from rolldown or the original `magic-string`.

sapphi-red: If the motivation is to add a way to detect the constructor is from rolldown, I think it is better to add `isRolldownMagicString` property. `[Symbol.toStringTag]` is harder to access and using `Object.prototype.toString.call(s)` is even harder.

Putting that aside, I think we should avoid exposing `Binding*` names. `RolldownMagicString` sounds good to me.

IWANABETHATGUY: > If the motivation is to add a way to detect the constructor is from rolldown, I think it is better to add `isRolldownMagicString` property. `[Symbol.toStringTag]` is harder to access and using `Object.prototype.toString.call(s)` is even harder.
> 
> Putting that aside, I think we should avoid exposing `Binding*` names. `RolldownMagicString` sounds good to me.

I see

packages/rolldown/src/binding-magic-string.ts
```diff
@@ -0,0 +1,23 @@
+import { BindingMagicString as NativeBindingMagicString } from './binding.cjs';
+
+// Set `isRolldownMagicString` so external packages (e.g. rolldown-string) can
+// detect native BindingMagicString instances without importing rolldown:
+//   obj.isRolldownMagicString === true
+// This replaces the fragile `obj.constructor.name` check which breaks with
+// minification or bundling.
+Object.defineProperty(NativeBindingMagicString.prototype, 'isRolldownMagicString', {
```
hyf0, line 8: Could we use other method? The current way relies on the module execution order, which isn't a quite realiable way.

---

Wait. What does this code does? I just saw the following code already has other logic to ensure this behavior.

Image

packages/rolldown/src/binding-magic-string.ts
```diff
@@ -0,0 +1,23 @@
...
+  value: true,
+  writable: false,
+  configurable: false,
+});
+
+export interface BindingMagicString extends NativeBindingMagicString {
+  readonly isRolldownMagicString: true;
+}
+
+type BindingMagicStringConstructor = Omit<typeof NativeBindingMagicString, 'prototype'> & {
```
hyf0, line 18: And what's the purpose of this? 

What's wrong with `export const BindingMagicString = NativeBindingMagicString` directly?

packages/rolldown/tests/magic-string/magic-string-to-string-tag.test.ts
```diff
@@ -0,0 +1,29 @@
+import assert from 'node:assert';
+import { BindingMagicString as MagicString } from 'rolldown';
```
hyf0, line 2: I thought we're gonna rename this as `RolldownMagicString` instead of keeping `BindingMagicString`.

hyf0 (commented):
Image

FYI: Your commits are not verified.

---

Sorry about the question tone, I talked to CC too much and got used to ask questions.

packages/rolldown/tests/magic-string/magic-string-to-string-tag.test.ts
IWANABETHATGUY: > I thought we're gonna rename this as `RolldownMagicString` instead of keeping `BindingMagicString`.

Then that is a breaking change, will do it in another pr.

packages/rolldown/src/binding-magic-string.ts
IWANABETHATGUY: > And what's the purpose of this?
> 
> What's wrong with `export const BindingMagicString = NativeBindingMagicString` directly?

Make sure the type check passes.

packages/rolldown/src/binding-magic-string.ts
IWANABETHATGUY: One is type extension, and another one is implementation.

IWANABETHATGUY: > 

As for the issue of execution order. For our internal codebase, we did not use the prop at all, so it would not be affected by the execution order. For an external user, it would always be initialized before exporting.

hyf0 (approved):

IWANABETHATGUY:
> Image
> FYI: Your commits are not verified.
> 
> Sorry about the question tone, I talked to CC too much and got used to ask questions.

This is because I sent a PR with a laptop that is not used often; the key is stale. will fix it

IWANABETHATGUY:
> > Image
> > FYI: Your commits are not verified.
> > Sorry about the question tone, I talked to CC too much and got used to ask questions.
> 
> This is because I sent a PR with a laptop that is not used often; the key is stale. will fix it

image

There is a pro of graphite, it would auto-sign the commit when it is merged into the main branch .

PR #8637, opened 2026-03-11
refactor(rust): filter external modules from entries instead of mapping bit positions

Dunqing:
## Summary

Follow-up to #8596. Instead of using `bit_to_chunk_idx` to map bit positions to chunk indices (a band-aid that every consumer must remember), this filters external modules at the source so they never get entry bit positions.

- **`module_loader.rs`**: Skip external modules when collecting `dynamic_import_entry_ids`
- **`chunk_graph.rs`**: Remove `bit_to_chunk_idx` field
- **`code_splitting.rs`**: Remove `bit_to_chunk_idx` pre-allocation/recording, add `debug_assert` for entry module normality
- **`chunk_optimizer.rs`**: Use direct `ChunkIdx::from_raw(bit)` instead of `bit_to_chunk_idx` lookup
- **`code-splitting.md`**: Update design doc to reflect new approach

This matches esbuild's approach where external modules never enter the entry list, ensuring bit positions directly equal chunk indices.

## Test plan

- [x] `cargo test -p rolldown --test integration -- "8595"` passes
- [x] `cargo test -p rolldown --test integration -- "circular_chunk_from_external"` passes
- [ ] Full CI

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -350,7 +350,7 @@ impl GenerateStage<'_> {
         if !temp_chunk.needs_creation {
           return None;
         }
-        let chunk_idxs: Vec<_> = bits.index_of_one().map(|bit| ChunkIdx::from_raw(bit)).collect();
+        let chunk_idxs: Vec<_> = bits.index_of_one().map(ChunkIdx::from_raw).collect();
 
         let merge_target = Self::try_insert_into_existing_chunk(
           &chunk_idxs,
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -352,10 +352,7 @@ impl GenerateStage<'_> {
         }
         let chunk_idxs: Vec<_> = bits
           .index_of_one()
-          // Use bit_to_chunk_idx to correctly map bit positions to chunk indices.
-          // Bit positions may not match chunk indices when external module entries
-          // are skipped during chunk creation.
-          .filter_map(|bit| chunk_graph.bit_to_chunk_idx.get(bit as usize).and_then(|idx| *idx))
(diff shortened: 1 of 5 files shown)


crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
```diff
@@ -350,13 +350,7 @@ impl GenerateStage<'_> {
         if !temp_chunk.needs_creation {
           return None;
         }
-        let chunk_idxs: Vec<_> = bits
-          .index_of_one()
-          // Use bit_to_chunk_idx to correctly map bit positions to chunk indices.
-          // Bit positions may not match chunk indices when external module entries
-          // are skipped during chunk creation.
-          .filter_map(|bit| chunk_graph.bit_to_chunk_idx.get(bit as usize).and_then(|idx| *idx))
-          .collect();
```
IWANABETHATGUY, line 359: https://github.com/rolldown/rolldown/pull/8596/changes#diff-e11d3ed0b62f1520cb0548f0bc4d47661026081bb72771f84e548eefda7b40faL355-L360

Previously, the chunk_idxs only collected entry chunk indexes. Both before and after #8596, the condition was true; now it breaks.

Dunqing: The old `entry_chunk_idx` filter existed because external entries created gaps between bit positions and chunk indices. After this PR, external modules are filtered out in `module_loader.rs` before they get entry bit positions, so every bit `0..N-1` maps directly to `ChunkIdx(0..N-1)` — all valid entry chunks. No gaps, no filter needed.

Additionally, `try_insert_into_existing_chunk` already guards against invalid indices at [L525-526](https://github.com/rolldown/rolldown/blob/61d181719/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs#L525-L526) and returns `None` for common chunks at [L536](https://github.com/rolldown/rolldown/blob/61d181719/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs#L536).

IWANABETHATGUY (approved):

PR #8672, opened 2026-03-13
refactor(rust): simplify `cross_module_optimization` by removing redundant scope tracking

Dunqing:
Remove `scope_stack`, `traverse_state`, and unused `side_effects_free_function_optimization` field from `CrossModuleOptimizationRunnerContext`. The `TopLevel` guard on `visit_call_expression` was redundant because the second branch `!stmt_idx_to_dynamic_import_expr_addr.is_empty()` already ensured the check ran for modules with dynamic imports, and for the top-level case the side-effect-free function detection should run unconditionally anyway.

### DIFF
--- a/crates/rolldown/src/stages/link_stage/cross_module_optimization.rs
+++ b/crates/rolldown/src/stages/link_stage/cross_module_optimization.rs
@@ -8,23 +8,19 @@ use oxc::{
     },
   },
   ast_visit::{Visit, walk},
-  semantic::ScopeFlags,
 };
 use rolldown_common::{
   AstScopes, ConstExportMeta, EcmaViewMeta, FlatOptions, GetLocalDb, ModuleIdx,
   SharedNormalizedBundlerOptions, SideEffectDetail, StmtInfoIdx, SymbolRef, SymbolRefDb,
   SymbolRefFlags,
 };
-use rolldown_ecmascript_utils::{ExpressionExt, is_top_level};
+use rolldown_ecmascript_utils::ExpressionExt;
 use rolldown_utils::rayon::{IntoParallelRefIterator, ParallelIterator};
 use rustc_hash::{FxHashMap, FxHashSet};
 
-use crate::{
-  ast_scanner::{
-    const_eval::{ConstEvalCtx, try_extract_const_literal},
-    side_effect_detector::SideEffectDetector,
-  },
-  module_finalizers::TraverseState,
+use crate::ast_scanner::{
+  const_eval::{ConstEvalCtx, try_extract_const_literal},
+  side_effect_detector::SideEffectDetector,
 };
 
 use super::LinkStage;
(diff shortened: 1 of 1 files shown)


crates/rolldown/src/stages/link_stage/cross_module_optimization.rs
```diff
@@ -405,28 +371,24 @@ impl<'a, 'ast: 'a> Visit<'ast> for CrossModuleOptimizationRunnerContext<'a, 'ast
...
+        let ref_id = item.reference_id.get()?;
+        let symbol_id = self.immutable_ctx.eval_ctx.scope.get_reference(ref_id).symbol_id()?;
+
+        let symbol_ref = self
+          .immutable_ctx
+          .symbols
+          .canonical_ref_for((self.immutable_ctx.module_idx, symbol_id).into());
+        Some(self.immutable_ctx.global_side_effect_free_function_symbols.contains(&symbol_ref))
+      })
+      .unwrap_or(false);
```
IWANABETHATGUY, line 387: For the inner scope CallExpr heavy scenario, this will regress.

IWANABETHATGUY:
Analysis generated by C/C: 

**Old code's guard was cheap:**
- `self.traverse_state.contains(TraverseState::TopLevel)` is a single bitflag check — essentially free
- It immediately rejects all non-top-level call expressions with zero work

**Old code's scope tracking was also cheap:**
- `Vec::push`/`pop` of a `ScopeFlags` (a small integer) — sequential, cache-friendly
- Bitflag update — trivial

**New code now runs the full check on every call expression:**
- `as_identifier()` — enum match
- `reference_id.get()` — Cell read
- `scope.get_reference().symbol_id()` — indexed lookup
- `canonical_ref_for(...)` — pointer chasing through symbol ref chain
- `HashSet::contains(...)` — hash + probe

**the vast majority of call expressions live inside function bodies, not at the top level**. A typical module might have 10 top-level calls and 200+ nested ones. The old bitflag guard rejected those 200 calls with a single branch. Now all 200 go through the full lookup chain.

So the real trade-off is:

| | Old | New |
|---|---|---|
| Scope tracking | O(scopes) cheap Vec ops | 0 |
| Side-effect check | ~top-level calls only | ALL call expressions |

The 190+ extra hash lookups (with cache-unfriendly pointer chasing in `canonical_ref_for`) likely outweigh the savings from removing cheap, cache-friendly Vec push/pops.

Performance is likely a **small net negative**, not a win. The readability gain is real, but calling the scope tracking "redundant" understates the fact that it was also serving as an efficient early-exit optimization.

IWANABETHATGUY (approved):

IWANABETHATGUY:
I don't have a strong opinion on this because I haven't had time to analyze the actual distribution of `CallExpr`, which requires many real-world projects. Feel free to merge

Dunqing:
Considering this becomes much clearer, so merging it.

shulaoda (approved):

PR #8680, opened 2026-03-14
refactor(bench): benchmark individual bundler phases

Boshen:
## Summary

- Refactor `Bundle` public API to expose `scan()`, `link()`, and `generate_from_link()` as composable steps, re-exporting `NormalizedScanStageOutput` and `LinkStageOutput` from the crate root
- Benchmark each bundler phase independently (scan, link, generate) using `iter_batched` to exclude setup work from timing
- Merge `scan.rs` benchmark into `bundle.rs` and enable minify variants for generate/bundle groups


crates/rolldown/src/bundle/bundle.rs
```diff
@@ -80,11 +80,19 @@ impl<Fs: FileSystem + Clone + 'static> Bundle<Fs> {
...
+  pub async fn scan(&mut self) -> BuildResult<NormalizedScanStageOutput> {
+    self.scan_modules(ScanMode::Full).await
+  }
+
+  pub fn link(&self, scan_output: NormalizedScanStageOutput) -> LinkStageOutput {
+    LinkStage::new(scan_output, &self.options).link()
+  }
 
-    Ok(())
+  pub async fn generate_from_link(
```
hyf0, line 91: Can we hide them under feature gates like `internal` or `bench`?

hyf0 (dismissed):

crates/rolldown/src/bundle/bundle.rs
```diff
@@ -80,11 +80,19 @@ impl<Fs: FileSystem + Clone + 'static> Bundle<Fs> {
   }
 
   #[tracing::instrument(level = "debug", skip_all, parent = &*self.bundle_span)]
-  /// This method intentionally get the ownership of `self` to show that the method cannot be called multiple times.
-  pub async fn scan(mut self) -> BuildResult<()> {
-    self.scan_modules(ScanMode::Full).await?;
+  pub async fn scan(&mut self) -> BuildResult<NormalizedScanStageOutput> {
```
hyf0, line 83:
```suggestion
  pub async fn scan(self) -> BuildResult<NormalizedScanStageOutput> {
```

`Bundle` is not execpted to be used multiple times.

Can we distinguish API that we exposed for normal usage and benchmark usage?

Boshen:
hmm let me think about this.

PR #8694, opened 2026-03-14
perf(bench): remove scan benchmark binary to halve LTO link time

Boshen:
## Summary

- Remove the redundant `scan` benchmark binary — since `bundle` (which calls `generate()`) already includes scanning, it was duplicating work
- Each benchmark binary triggers a separate fat LTO link pass (~57s locally, ~7-8 min on CI), so removing one binary cuts that cost in half
- Rename `bundle.rs` → `bench.rs` and consolidate to a single `[[bench]]` entry in Cargo.toml


Boshen:
Why do we have scan anyway when bundle includes scan?

IWANABETHATGUY (approved):

Boshen:
The `scan` benchmark was originally added in #1410 (June 2024) when the monolithic `threejs.rs` benchmark was split into separate `bundle.rs` and `scan.rs` binaries, alongside adding the `rome_ts` preset. It was split to measure scan and bundle performance independently, but since `bundle` already performs scanning as part of its pipeline, the separate binary was redundant and doubled the fat LTO link cost.

PR #8701, opened 2026-03-15
perf(plugin): reduce monomorphization bloat by merging Plugin and Pluginable traits

Boshen:
## Summary

- Merge the two-trait plugin architecture (`Plugin` + `Pluginable`) into a single `#[async_trait]` `Plugin` trait
- Previously each of ~34 plugin types generated unique future types for all 22 async hooks via RPITIT, even for default no-ops — the blanket impl then boxed each one, producing ~748 monomorphized wrappers
- Now default methods compile once (shared vtable entry), only overridden methods generate per-type code (~102 wrappers instead of 748)
- Net deletion of ~420 lines (the entire `Pluginable` trait and blanket impl)
- Re-exports `async_trait` from `rolldown_plugin` so plugin crates don't need a direct dependency


Boshen:
## Result: Net negative on binary size

After benchmarking, this approach **increases** the napi binary by **+1.8 MiB (+9.6%)** instead of reducing it.

| | `.node` binary size |
|---|---|
| `main` | 19,422,240 bytes (18.5 MiB) |
| this PR | 21,277,792 bytes (20.3 MiB) |
| **diff** | **+1,855,552 bytes (+1.8 MiB)** |

While `rolldown_plugin` itself shrank (379 KiB → 219 KiB), the `#[async_trait]` desugaring on each of the ~52 `impl Plugin for` blocks generates more total boxing/lifetime code than the old blanket impl + RPITIT approach eliminated. The monomorphization cost was effectively replaced by a larger `async_trait` overhead spread across all plugin crates.

Closing as the approach doesn't achieve its goal.

PR #8702, opened 2026-03-15
perf: disable napi tracing feature to reduce binary size

Boshen:
## Summary

- Remove the `"tracing"` feature from both `napi` and `napi-derive` workspace dependencies
- This eliminates ~491 generated `tracing::debug!()` call sites in napi wrapper functions, saving **~99 KiB** in the release binary
- Rolldown's own tracing infrastructure (`rolldown_tracing` / `RD_LOG`) is unaffected

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1526,7 +1526,6 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
- "tracing",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -229,9 +229,9 @@ quote = "1"
 syn = { version = "2", default-features = false }
 
 # napi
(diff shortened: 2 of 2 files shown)


sapphi-red:
Hmm, is it possible to reduce the size without completely removing this feature? If we do this, we'll have to create a build with this feature enabled and give it to users when these kinds of bug were found.

Boshen:
> Hmm, is it possible to reduce the size without completely removing this feature? If we do this, we'll have to create a build with this feature enabled and give it to users when these kinds of bug were found.

What useful tracing information do the N-API crates provide?

sapphi-red:
It helps debugging deadlock issues like https://github.com/rolldown/rolldown/pull/7289. It will log for each napi function call and we can know which function is pending.

Boshen:
ahh ok

PR #8708, opened 2026-03-15
perf: replace TypedDashMap with TypedMap in CustomField

Boshen:
## Summary

- Replace `TypedDashMap` (DashMap-backed, 64 shards × 8 KB = 512 KB per instance) with `TypedMap` (plain HashMap-backed) in `CustomField`
- `CustomField` is write-once-read-many: populated at creation, wrapped in `Arc`, then only read — concurrent write access is never needed
- Removes `dashmap` feature dependency from `typedmap` crate

  Before (from earlier profiling):
  - DashMap shard allocations from CustomField: ~24.2 MB in 3,104 blocks
  - Total allocated: ~311 MB

  After (with TypedMap):
  - DashMap shard overhead (remaining, from other DashMaps like resolver/file emitter): 0.17 MB in ~17 blocks
  - TypedMap/CustomField allocations: 0.15 MB
  - Total allocated: 301.2 MB

  The CustomField DashMap shard overhead dropped from ~24 MB to ~0.15 MB — a ~24 MB reduction (~99.4% elimination). The remaining 0.17 MB of
  DashMap shard allocations come from other legitimate concurrent DashMaps (resolver cache, file emitter, etc.), not from CustomField.

  The total allocation reduction of ~10 MB (311 → 301 MB) is less than the full 24 MB because this is a single-threaded dhat run (fewer shards
  per DashMap than the 64-shard production scenario). The savings on a 10+ core production machine would be the full ~24 MB.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4318,9 +4318,6 @@ name = "typedmap"
 version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "63278e72ed4f207eb3216c944cbafb35bdb656d2eab97ef73c0c165a1cd3e319"
-dependencies = [
- "dashmap",
-]
 
 [[package]]
 name = "typenum"
--- a/crates/rolldown/tests/rolldown/plugin/plugin_context/custom_arg_in_resolve/mod.rs
+++ b/crates/rolldown/tests/rolldown/plugin/plugin_context/custom_arg_in_resolve/mod.rs
@@ -29,7 +29,7 @@ impl Plugin for TestPluginCaller {
     args: &HookResolveIdArgs<'_>,
   ) -> HookResolveIdReturn {
     if args.specifier == "foo" {
-      let custom = CustomField::default();
+      let mut custom = CustomField::default();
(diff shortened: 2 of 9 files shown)


hyf0:
@copilot+gpt-5.4 add a design note `plugin-context-resolve.md`, no need to add detailed content, just record content like becuase "concurrent write access is never needed" so we're using TypedMap directly, once we encounter features that required write in multiple places, we could switch back to `TypedDashMap`.

hyf0 (approved):

hyf0:
### Merge activity

* **Mar 15, 3:40 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Mar 15, 3:43 PM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Mar 15, 3:48 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

Issue #8732, opened 2026-03-16
[Bug]: `resolve.tsconfigPaths` fails on Windows during SSR build — tsconfig not found with mixed path separators


benpsnyder:
## Environment

- Vite 8.0.0
- Rolldown 1.0.0-rc.9
- Windows (GitHub Actions `windows-latest`)
- Node 24.11.0
- pnpm 10.32.0

## Description

`resolve: { tsconfigPaths: true }` works correctly on Linux but fails on Windows during the SSR build phase. The client build succeeds; the failure occurs when `@analogjs/vite-plugin-nitro` triggers a separate Vite SSR build via `closeBundle`.

The error is:

```
Tsconfig not found D:/a/analog\tsconfig.base.json.
```

The path has mixed separators (`D:/a/analog\tsconfig.base.json`) and resolves to the **parent** of the repository root (`D:/a/analog/`) instead of the repository root itself (`D:/a/analog/analog/`). The actual `tsconfig.base.json` is at `D:/a/analog/analog/tsconfig.base.json`.

## Reproduction

Repository structure:

```
D:/a/analog/analog/              ← workspace root, contains dagger.json
├── tsconfig.base.json           ← the file that should be found
├── apps/
│   └── analog-app/
│       ├── vite.config.ts       ← has resolve: { tsconfigPaths: true }
│       └── tsconfig.json        ← extends ../../tsconfig.base.json
└── packages/
    └── vite-plugin-nitro/       ← triggers SSR build in closeBundle hook
```

`apps/analog-app/vite.config.ts`:

```ts
export default defineConfig(() => ({
  root: __dirname,
  resolve: { tsconfigPaths: true },
  plugins: [analog()],
}));
```

`apps/analog-app/tsconfig.json`:

```json
{
  "extends": "../../tsconfig.base.json"
}
```

Steps:

1. Clone repo on Windows (or run on GitHub Actions `windows-latest`)
2. `pnpm install`
3. `pnpm exec nx build analog-app`
4. Client build succeeds
5. SSR build (triggered by `vite-plugin-nitro` in `closeBundle`) fails with:

```
[plugin rollup-plugin-replace-files]
Error: Errored while resolving "@angular/core" in `this.resolve`.
Caused by:
    Tsconfig not found D:/a/analog\tsconfig.base.json.
```

## Analysis

[...]

PR #8733, opened 2026-03-16
fix(vite-resolve): normalize importer paths to native separators on Windows

benpsnyder:
## Summary

- On Windows, Vite passes importer paths with forward slashes (`D:/a/project/src/main.ts`), but when these are joined with OS paths or used in oxc_resolver cache lookups, mixed separators (`D:/a/project\tsconfig.json`) cause tsconfig path resolution to fail.
- Added `normalize_path_to_native()` utility that converts path separators to the OS-native format (backslashes on Windows, no-op elsewhere).
- Normalize importer paths before passing to `oxc_resolver` and normalize the root path at plugin initialization to ensure consistent separators throughout resolution.
- Fixed optional peer dep detection to use `Path`-based comparison (`Path::eq` compares by components) instead of string comparison, preventing separator mismatches between the importer directory and root path on Windows.
- Enabled lib tests for the crate and added unit tests for `normalize_path` and `normalize_path_to_native` with platform-conditional assertions.

## Test plan

- [x] Unit tests for `normalize_path` and `normalize_path_to_native` (platform-conditional)
- [ ] Verify Windows CI passes with tsconfig paths resolution
- [ ] Verify no regression on Linux/macOS (the normalization is a no-op on non-Windows platforms)

### DIFF
--- a/crates/rolldown_plugin_vite_resolve/Cargo.toml
+++ b/crates/rolldown_plugin_vite_resolve/Cargo.toml
@@ -8,7 +8,6 @@ description = "Rolldown plugin for Vite module resolution"
 
 [lib]
 doctest = false
-test = false
 
 [dependencies]
 anyhow = { workspace = true }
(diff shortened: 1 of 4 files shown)


sapphi-red (changes requested):
We need a minimal reproduction or a test that fails without this change and passes with this change.
https://github.com/rolldown/rolldown/issues/8732#issuecomment-4066564976

Boshen:
🤔 I'd rather fix this properly in the resolver

PR #8741, opened 2026-03-16
fix(renamer): handle existing bindings in nested scopes when finding unique names

### LINKED ISSUE #8739: [Bug]: Produced JS contains conflicting identifiers

When there is an import of kind `import * as dep from "./dep.js"` which is later used as `dep.clashingIdentifier()` then Rolldown adds that `clashingIdentifier` into scope as is + renames already existing local `clashingIdentifier` into `clashingIdentifier$1`. The problem is that it does not check whether there already was another local `clashingIdentifier$1`. That results into two declarations of the same identifier which is an error.

[REPL with minimal repro](https://repl.rolldown.rs/#eNptUsFOwzAM/RUrQmKDqhPXIk6cOHDjmEvUuFsgTarEGaCp/46dbgO0SZFax8/P9ss7qEF1B+WCxa/2Pct/UN1v3KhewnGKieAOTAaLEwwpjqBVu+GAUVo96qDDUEJPLgYgzLTaobENkHF+DQcdADwS9N7knQvbF4uB3OAwwRMIkgkWxOis9ci3Qn0JX63PyMvkzQPXSceKSUglhWst709d7q+ScPUs+9Q1tJLxtGp4X6HWSiZgXVB1lArOjUrRexs/Q9vHMLhtS2cZr2T+CXrgLQcX8LmmYV6EvT2V3coU+FWhDDTF1++5YFV1dWEq1MFC2Y5Ipq1XjSRjoZqtSIDNBt52LgOfqST03zDEJCIkR99gS2IpuMUY2z94ZHguCBOm7DJlwD0G+HS0Y3Yg5juip4QZ0x5foy0ecwcikKRmYEnXLBurNZn+w2yRbRNZokWnf3eLQnVgrdgFyF4MvUO2Ge/BL8NHiBbvnSiO0VJ81OxsyGs+WiQ5mkSrvfE8qxi50jP/nom8EROo+QfuuRXq)

*This is not a synthetic theoretical issue, I discovered it in real world project.*

Context: I use [Gleam](https://github.com/gleam-lang/gleam) + [Lustre](https://github.com/lustre-labs/lustre) for web development and Vite for bundling my projects. After upgrading to Vite 8.0 I discovered that produced JS code started failing with errors like `Uncaught SyntaxError: Identifier 'v' has already been declared`. 

When Gleam is used with JavaScript as a target, Gleam compiler generates `*.mjs` files which I then import in HTML and/or other JS/TS modules (and vice versa) and bundle the whole project using Vite 8.0. There are a couple of things that come together to cause the issue.

Firstly, let's say there is a Gleam module that imports another Gleam module `import lustre/vdom/path` which is then `path.child(some_path)`. Gleam compiler transpiles such import to `import * as $path from "../../lustre/vdom/path.mjs"` and the function call to `$path.child(some_path)`.

Secondly, Gleam has shadowing and if in some scope there is a declaration like `let child = "shadowed"` which is at some later point followed by `let child = "value"` — that is a totally valid (and still immutable) Gleam code. However Gleam compiler would produce following JS for those: `let child = "shadowed";` for the first assignment and `let child$1 = "value";` for the second assignment. You can see this in [Gleam Playground](https://playground.gleam.run/#N4IgbgpgTgzglgewHYgFwEYA0IDGyAuES+aIcAtgA4JT4AEA5gDYQCG5A9IgDpK+UBXAEZ0AZkjrlWcJAAoAlHWC86dFvRwALOEwAmdALx1uIGJta6EAdwi6TKtRA3a9h4+FZMBEexLqIAOkooGXwmOS0dXXleAF9eEFigA=) without having to install Gleam: just click "Compiled JavaScript" tab on the right side.

And as you might've guessed, if I happen to have both of those things in the same module — the import of a Gleam package with the `child` function, and a shadowed local variable with the same name — that leads to the exact situation that was shown in the Rolldown REPL and Rolldown generates invalid JavaScript.

### What is expected?

When deconflicting(?) I expect Rolldown to check if identifier name like `identifier$1` is already taken before it chooses to use it.

When rewriting imports of valid JavaScript code Rolldown creates name conflicts that result in errors like `Uncaught SyntaxError: Identifier '...' has already been declared`. For example, valid JS like:

```js
import * as dep from "./dep.js";

function test(head, tail) {
  let clashingIdentifier = head; // <- this was OK
  let middle = dep.clashingIdentifier();
  let clashingIdentifier$1 = tail; // <- this was OK
  return clashingIdentifier + middle + clashingIdentifier$1;
}

test("head", "tail");
```

gets transformed into

```js
import { clashingIdentifier } from "./dep.js";
//#region index.js
function test(head, tail) {
	let clashingIdentifier$1 = head; // <- this is a conflict now
	let middle = clashingIdentifier();
	let clashingIdentifier$1 = tail; // <- this is a conflict now
	return clashingIdentifier$1 + middle + clashingIdentifier$1;
}
test("head", "tail");
//#endregion

```

### System Info

```Shell
System:
    OS: macOS 15.5
    CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
    Memory: 1.32 GB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 25.8.0 - /Users/idea-list/.local/share/mise/installs/node/25.8.0/bin/node
    npm: 11.11.0 - /Users/idea-list/.local/share/mise/installs/node/25.8.0/bin/npm
    pnpm: 10.28.0 - /etc/profiles/per-user/idea-list/bin/pnpm
    bun: 1.3.3 - /etc/profiles/per-user/idea-list/bin/bun
  Browsers:
    Brave Browser: 145.1.87.188
    Safari: 18.5
  npmPackages:
    rolldown: ^1.0.0-rc.9 => 1.0.0-rc.9
```

### Any additional comments?

_No response_

drewolson:
Fixes #8739

### DIFF
--- a/crates/rolldown/src/utils/renamer.rs
+++ b/crates/rolldown/src/utils/renamer.rs
@@ -257,16 +257,27 @@ impl<'name> Renamer<'name> {
     }
 
     // Find unique name: skip candidates that conflict with top-level symbols
+    // or with existing bindings in nested scopes of the same module.
     for count in 1u32.. {
       let name: CompactStr =
         concat_string!(original_name, "$", itoa::Buffer::new().format(count)).into();
 
-      if let Entry::Vacant(entry) = self.used_canonical_names.entry(name) {
-        let candidate_name = entry.key().clone();
-        entry.insert(0);
-        self.canonical_names.insert(symbol_ref, candidate_name);
-        return;
+      if self.used_canonical_names.contains_key(&name) {
+        continue;
       }
+
+      // Also skip if the candidate name conflicts with an existing binding in
+      // a nested scope of the same module. Without this check, renaming `child`
+      // to `child$1` could collide with an existing `child$1` binding in the
+      // same scope (e.g. from Gleam's variable shadowing convention).
+      if self.has_nested_scope_binding(symbol_ref.owner, &name) {
+        self.used_canonical_names.insert(name, 0);
+        continue;
+      }
+
+      self.used_canonical_names.insert(name.clone(), 0);
+      self.canonical_names.insert(symbol_ref, name);
+      return;
     }
   }
 
--- a/crates/rolldown/tests/rolldown/issues/8739/_config.json
+++ b/crates/rolldown/tests/rolldown/issues/8739/_config.json
@@ -1,0 +1,5 @@
+{
+  "config": {
+    "minify": false
+  }
+}
--- a/crates/rolldown/tests/rolldown/issues/8739/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/issues/8739/artifacts.snap
@@ -1,0 +1,27 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+---
+# Assets
+
+## main.js
+
+```js
+import assert from "node:assert";
+
+//#region dep.js
+function clashingIdentifier() {
+	return "middle";
+}
+
+//#endregion
+//#region main.js
+function test(head, tail) {
+	let clashingIdentifier$2 = head;
+	let middle = clashingIdentifier();
+	let clashingIdentifier$1 = tail;
+	return clashingIdentifier$2 + middle + clashingIdentifier$1;
+}
+assert.strictEqual(test("head", "tail"), "headmiddletail");
+
+//#endregion
+```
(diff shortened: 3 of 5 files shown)


Dunqing (approved):
LGTM!

PR #8778, opened 2026-03-18
docs: update reference to transform function in plugin API documentation

zOadT:
Minor fix for the docs: `transform` was moved to `rolldown/utils` in #8214, but the docs still reference `rolldown/experimental` currently.


IWANABETHATGUY (approved):
Thanks

PR #8780, opened 2026-03-19
fix: external.md on Windows OS

bddjr:


hyf0 (approved):
Thank you!

bddjr:
There is a bug in the documentation framework where `\\` is being incorrectly rendered as `%5C%5C`.

https://rolldown.rs/reference/InputOptions.external

image

PR #8784, opened 2026-03-19
chore: upgrade oxc to 0.121.0

shulaoda:
## Summary
- Upgrade oxc Cargo crates from 0.120.0 to 0.121.0
- Upgrade `@oxc-project/runtime`, `@oxc-project/types`, `oxc-parser`, `oxc-minify`, `oxc-transform` npm packages from 0.120.0 to 0.121.0

## Test plan
- [x] `cargo check` passes
- [x] `just test-update` passes (0 failures)
- [x] `just roll` passes (full build, lint, and test suite)

### DIFF
--- a/crates/rolldown/tests/esbuild/dce/dce_of_experimental_decorators/artifacts.snap
+++ b/crates/rolldown/tests/esbuild/dce/dce_of_experimental_decorators/artifacts.snap
@@ -23,7 +23,13 @@ var Method = class {
 };
 __decorate([fn], Method.prototype, "method", null);
 var Accessor = class {
-	accessor accessor;
+	#_accessor_accessor_storage;
+	get accessor() {
+		return this.#_accessor_accessor_storage;
+	}
+	set accessor(value) {
+		this.#_accessor_accessor_storage = value;
+	}
 };
 __decorate([fn], Accessor.prototype, "accessor", null);
 var Parameter = class {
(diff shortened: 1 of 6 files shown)


crates/rolldown/tests/esbuild/dce/dce_of_experimental_decorators/artifacts.snap
```diff
@@ -23,7 +23,13 @@ var Method = class {
 };
 __decorate([fn], Method.prototype, "method", null);
 var Accessor = class {
-	accessor accessor;
+	#_accessor_accessor_storage;
```
hyf0, line 26: Is this expected

hyf0 (approved):

crates/rolldown/tests/esbuild/dce/dce_of_experimental_decorators/artifacts.snap
shulaoda: @copilot take a look at this snapshot change

shulaoda: I think so.

The snapshot change in `dce_of_experimental_decorators` is expected. OXC 0.121.0 now properly lowers the `accessor` keyword in experimental decorators mode to a private field + getter/setter pair (e.g `#_accessor_accessor_storage` + `get accessor()` / `set accessor()`), instead of leaving the raw `accessor` syntax in the output.

Issue #3650, opened 2025-02-21
[Bug]: Circular deps caused by advancedChunks would throw error of referenced value is `undefined`


hyf0:
### Reproduction link or steps

I met this at a large project. I'm trying to make a smaller reproduction.

### What is expected?

.

### What is actually happening?

.

### System Info

```Shell
.
```

hyf0:
Chunk_main.js -> Chunk_sub.js(includes `rolldown:runtime`) -> Chunk_deep.js -> Chunk_sub.js

hyf0:
`rolldown:runtime` doesn't get wrapped, which means it doesn't get ensured to be executeed first and on demand.

Issue #2654, opened 2024-11-07
[Feat]: Inline `rolldown:runtime` for each chunk instead of share it


thesmartwon:
### Reproduction link or steps

When using `advancedChunks.groups` the highest priority chunk will get the `rolldown:runtime` and others will import from it.

### What is expected?

```js
//#region rolldown:runtime
var __esmMin = (fn, res) => () => (fn && (res = fn(fn = 0)), res);
```

```js
//#region rolldown:runtime
var __esmMin = (fn, res) => () => (fn && (res = fn(fn = 0)), res);
```

### What is actually happening?

```js
//#region rolldown:runtime
var __esmMin = (fn, res) => () => (fn && (res = fn(fn = 0)), res);
```

```js
import { __esmMin } from "./chunk-one.js";
```

### System Info

```Shell
Rolldown 0.14.0
```

IWANABETHATGUY:
Shared `runtime` helper in multiple chunks will reduce amount of code size, can you elaborate why do you need to inline all the runtime helper function?

hyf0:
`rolldown:runtime` is like a normal module. There will be only one instance during the build. So if any chunk rely on it, it will imported it from where it exist.

thesmartwon:
> Shared runtime helper in multiple chunks will reduce amount of code size, can you elaborate why do you need to inline all the runtime helper function?

You're creating new dependencies between chunks that nullifies my chunking strategy because I use `[name]-[hash]` for chunk names.

> rolldown:runtime is like a normal module. There will be only one instance during the build. So if any chunk rely on it, it will imported it from where it exist.

Yes, this is exactly the issue. Whenever a chunk containing the implementation of a rollup runtime helper changes, the filename changes, and then ALL other chunks have to change to update their import path since ALL chunks depend on the runtime code. This happens regardless of my chunk's dependencies.

A workaround for now is to extract the runtime chunk. However, this is rather silly because the chunk is so small. I'd rather just have it inlined into each of my other chunks, and I'd argue this should be rollup's default strategy up until `(runtime size x number of chunks) > a single HTTP request overhead`.

I suppose I'm asking for a generic feature to inline chunks, and for rolldown's runtime to use that feature by default.

thesmartwon:
Ran into this again today. I'd love to eliminate one more tiny HTTP request!

FWIW, `tsc` and `babel` by default resort to inlining their helpers and require manual configuration to import them instead. I find this good behavior.

PR #8810, opened 2026-03-20
fix: extract runtime module from entry chunk to prevent circular dependencies

### LINKED ISSUE #8809: Runtime helper circular dependency when __exportAll is in entry chunk during code-splitting
## Bug Description

When code-splitting, rolldown may place the runtime module (containing `__exportAll`, `__commonJSMin`, etc.) inside the entry chunk. If facade elimination later adds `__exportAll` dependency to a common chunk, that common chunk imports the helper from the entry chunk. When the entry chunk also imports from the common chunk (e.g., through static `import` chains), a circular dependency is created, causing the helper to be `undefined` at ESM evaluation time.

## Reproduction

Project structure:
```
entry.js  →  gateway.js  →  daemon.js (barrel: export * from './impl.js')
                              ↑
completion.js  ──── dynamic import('./daemon.js')
```

Where `entry.js` also uses a CJS dependency (triggering `__commonJSMin` runtime helper, placing the runtime in the entry chunk).

**Expected:** `__exportAll` is imported from a standalone `rolldown-runtime` chunk or the chunk that defines it.
**Actual:** `__exportAll` is imported from `index.js` (entry chunk), creating a circular dependency:
```
daemon-cli-XXXX.js  →  import { __exportAll } from "./index.js"
index.js            →  import { ... } from "./daemon-cli-XXXX.js"
```

At runtime: `TypeError: __exportAll is not a function`

## Root Cause

In `chunk_optimizer.rs`, `optimize_facade_entry_chunks`:
1. The runtime module is assigned to the entry chunk during `split_chunks` (because only the entry's CJS path references runtime helpers)
2. Facade elimination for dynamic entries adds `__exportAll` to common chunks
3. The circular dependency check at line ~736 may be bypassed when `to_temp_idx()` returns `None` for newly created common chunks
4. The existing code at line ~1005 only handles the case where the runtime is NOT yet assigned (`module_to_chunk[runtime_module_idx].is_none()`), missing the case where it's already in the entry chunk

## Impact

- rc.3 → rc.10 fixed most instances (12 → 1 in our project), but edge cases remain
- The remaining case doesn't crash on startup but causes `undefined` errors when the affected dynamic import path is executed

## Related Issues

- #3650 — Same class of bug (closed, but not fully fixed)
- #2654 — Proposes inlining runtime per-chunk as a more thorough solution

Afee2019:
## Summary

- When the runtime module is in an entry chunk and facade elimination adds runtime helper dependencies (e.g. `__exportAll`) to other chunks, extract the runtime to a dedicated `rolldown-runtime` chunk to prevent circular dependencies
- Only triggers when the runtime is in an entry chunk (entry chunks depend on common chunks, creating cycle risk) — does not affect manual code splitting which already extracts the runtime
- Adds regression test with code-splitting scenario involving CJS deps + barrel modules + dynamic imports

Fixes #8809
Related: #3650, #2654

## Test plan

- [x] All existing fixture tests pass (748 passed, 4 pre-existing failures from missing `cjs-module-lexer` npm dep)
- [x] New test `runtime_helper_circular_dependency` verifies no circular imports between entry and common chunks
- [x] Manual code splitting tests unaffected (runtime already extracted by `extract_runtime_chunk`)


IWANABETHATGUY:
Thanks for your contribution; I would prefer to fix it rather than bail out the optimization directly.

Afee2019:
Thanks for the feedback! That makes total sense — fixing the root cause in the circular dependency detection (e.g. ensuring `to_temp_idx()` covers newly created common chunks) would be a better approach than bailing out.

I'll keep an eye on #8809 for the proper fix. Happy to help test it against our project when it's ready.

IWANABETHATGUY:
One more thing, so your test case is a minimum reproduction, right?

Afee2019:
Yes, it's a minimum reproduction that covers the 5 conditions needed to trigger the bug:

1. **CJS dependency** in entry — forces the runtime module (`__commonJSMin`) into the entry chunk's bit set
2. **Barrel module** (`daemon.js`) using `export *` — requires `__exportAll` helper for namespace construction
3. **Both static and dynamic imports** of the same barrel — static via `gateway.js`, dynamic via `completion.js`
4. **Multi-entry code splitting** — the barrel lands in a common chunk shared between entries
5. **Facade elimination** — the dynamic entry's facade is empty, so it gets eliminated, adding `__exportAll` dependency to the common chunk

The cycle: common chunk needs `__exportAll` from entry chunk (where runtime lives), and entry chunk imports business code from common chunk.

The test files in the PR (`entry.js`, `cjs-dep.cjs`, `gateway.js`, `daemon.js`, `daemon-impl.js`, `completion.js`) are the minimal set to reproduce this. Removing any one of these conditions prevents the bug from triggering.

IWANABETHATGUY:
I am sorry, but I can't reproduce the circular reference on main branch.

PR #8813, opened 2026-03-20
chore(node): migrate oxlint and oxfmt to Vite+

Boshen:


justfile
```diff
@@ -8,13 +8,10 @@ _default:
   just --list -u
 
 setup:
-  just check-setup-prerequisites
-  # Rust related setup
+  curl -fsSL https://vite.plus | bash
```
jakebailey, line 11: This effectively forces an install of `vp`, even if its already installed; can this be conditional on whether or not `vp` is present on PATH or something? (I have it installed, but not via this "pipe to bash" thing.)

camc314: https://github.com/rolldown/rolldown/pull/8862

PR #8816, opened 2026-03-20
fix: relax test utils path assertion to support git worktrees

younggglcy:
## Summary
- The `testsDir()` assertion in test utils checked the last 4 path segments, including the repo root directory name (`rolldown`). This fails when running in git worktrees or custom clone directories where the root name differs.
- Changed `slice(-4)` to `slice(-3)` so only `['packages', 'rolldown', 'tests']` is verified, which is sufficient to validate the path correctness regardless of the repo root name.

## Test plan
- Run existing tests in both the main repo and a git worktree to confirm the assertion passes in both cases.


hyf0 (approved):
Thank you!

PR #8817, opened 2026-03-20
fix: use sanitized filename for chunk name with preserveModulesRoot

### LINKED ISSUE #8761: [Bug]: sanitizeFileName is ignored for JS chunks when preserveModules: true with object-form entry

https://github.com/ef81sp/rolldown-sanitize-filename-repro

```
pnpm install
pnpm build
```

### What is expected?

JS chunk filenames should be sanitized (`?` and `&` replaced with `_`):

```
dist/components/TestComp.vue_vue_type_script_setup_true_lang.js
```

```
dist/components/TestComp.vue?vue&type=script&setup=true&lang.js
```

Custom `sanitizeFileName` in output options is also ignored for these chunks. CSS filenames are correctly sanitized in the same build.

This occurs when `entry` is object-form (`{ index: './src/index.ts' }`). With string-form entry (`'./src/index.ts'`), `preserveModules` does not preserve the directory structure at all, so the issue does not surface.

### System Info

```Shell
System:
    OS: macOS 26.3.1
    CPU: (8) arm64 Apple M1
    Memory: 102.55 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 25.5.0 - /opt/homebrew/bin/node
    pnpm: 10.28.0
  npmPackages:
    vite: 8.0.0 (rolldown 1.0.0-rc.9)
```

### Any additional comments?

#7554 fixed `sanitizeFileName` for `preserveModules`, but only when `entry` is string-form. With object-form `entry`, `sanitizeFileName` is still not applied to JS chunk filenames.

Workaround: using `entryFileNames` as a function to manually sanitize works.

younggglcy:
## Summary

- When `preserveModulesRoot` is set, the chunk name was derived from the unsanitized absolute filename, causing query string characters (`?`, `&`) to leak into `chunk.name` and subsequently into `fileName` when `entryFileNames` is a function
- This affected Vue SFC virtual modules (e.g. `Comp.vue?vue&type=script&setup=true&lang.ts`) in Vite lib mode, producing invalid filenames like `Comp.vue?vue&type=script&setup=true&lang.js`
- Fix: use the already-sanitized filename (`sanitized_absolute_filename`) instead of the raw one (`absolute_chunk_file_name`) when stripping the `preserveModulesRoot` prefix

Closes #8761

## Test plan

- [x] Added test fixture `sanitize-filename-query-string` that reproduces the bug with `preserveModulesRoot` + function-form `entryFileNames` + query string module IDs
- [x] Verified existing `sanitize-filename` test still passes
- [x] Verified with the [reproduction repo](https://github.com/ef81sp/rolldown-sanitize-filename-repro)

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/mod.rs
+++ b/crates/rolldown/src/stages/generate_stage/mod.rs
@@ -178,8 +178,8 @@ impl<'a> GenerateStage<'a> {
                 let p = PathBuf::from(sanitized_absolute_filename.as_str());
                 let relative_path = if p.is_absolute() {
                   if let Some(ref preserve_modules_root) = preserve_modules_root {
-                    if absolute_chunk_file_name.starts_with(preserve_modules_root.as_str()) {
-                      absolute_chunk_file_name[preserve_modules_root.len()..]
+                    if sanitized_absolute_filename.starts_with(preserve_modules_root.as_str()) {
+                      sanitized_absolute_filename[preserve_modules_root.len()..]
                         .trim_start_matches(['/', '\\'])
                         .to_string()
                     } else {
--- a/crates/rolldown/src/stages/generate_stage/mod.rs
+++ b/crates/rolldown/src/stages/generate_stage/mod.rs
@@ -148,8 +148,8 @@ impl<'a> GenerateStage<'a> {
 
     let mut index_chunk_id_to_representative_name = FxHashMap::default();
 
-    // Sanitize preserveModulesRoot so that a custom sanitizeFileName (e.g. '+' → '__')
-    // doesn't break the starts_with prefix comparison against sanitized absolute filenames.
+    // Keep the sanitized preserveModulesRoot for preliminary filename generation,
+    // which still operates on sanitized chunk filenames later in this stage.
     let sanitized_preserve_modules_root =
       if let Some(ref preserve_modules_root) = self.options.preserve_modules_root {
         Some(self.options.sanitize_filename.call(preserve_modules_root).await?)
@@ -159,7 +159,7 @@ impl<'a> GenerateStage<'a> {
 
     let index_pre_generated_names_futures = chunk_graph.chunk_table.iter().map(|chunk| {
       let sanitize_filename = self.options.sanitize_filename.clone();
-      let preserve_modules_root = sanitized_preserve_modules_root.clone();
+      let preserve_modules_root = self.options.preserve_modules_root.clone();
       let input_base = chunk.input_base.clone();
       let virtual_dirname = self.options.virtual_dirname.clone();
       async move {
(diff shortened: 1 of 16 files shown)


hyf0 (dismissed):
LGTM. Looks like related to `preserveModules`, so @IWANABETHATGUY cc

crates/rolldown/src/stages/generate_stage/mod.rs
```diff
@@ -178,8 +178,8 @@ impl<'a> GenerateStage<'a> {
                 let p = PathBuf::from(sanitized_absolute_filename.as_str());
                 let relative_path = if p.is_absolute() {
                   if let Some(ref preserve_modules_root) = preserve_modules_root {
-                    if absolute_chunk_file_name.starts_with(preserve_modules_root.as_str()) {
-                      absolute_chunk_file_name[preserve_modules_root.len()..]
+                    if sanitized_absolute_filename.starts_with(preserve_modules_root.as_str()) {
+                      sanitized_absolute_filename[preserve_modules_root.len()..]
```
sapphi-red: I think this doesn't work for a case like: https://github.com/rolldown/rolldown/pull/8762/changes#diff-9526602ae70389844a9f246bfcbad27aed83602b9b1d86a5c5b22dd6675d4f39
(I closed my copilot PR as it's wasn't easy enough to do it as a side task)

younggglcy: Thanks for pointing this out, will invesigate deeper on it further.

crates/rolldown/src/stages/generate_stage/mod.rs
```diff
@@ -178,8 +187,8 @@ impl<'a> GenerateStage<'a> {
                 let p = PathBuf::from(sanitized_absolute_filename.as_str());
                 let relative_path = if p.is_absolute() {
                   if let Some(ref preserve_modules_root) = preserve_modules_root {
-                    if absolute_chunk_file_name.starts_with(preserve_modules_root.as_str()) {
-                      absolute_chunk_file_name[preserve_modules_root.len()..]
+                    if sanitized_absolute_filename.starts_with(preserve_modules_root.as_str()) {
```
sapphi-red: I think this will incorrectly match when two directories are sanitized to the same value:

- `preserve_modules_root: './somewhere/foo+bar'` -> (sanitized) `./somewhere/foo__bar`
- chunk 1 filename: `./somewhere/foo+bar/baz.js` -> (sanitized) `./somewhere/foo__bar/baz.js`
- chunk 2 filename: `./somewhere/foo#bar/baz.js` -> (sanitized) `./somewhere/foo__bar/baz.js`

crates/rolldown_common/src/chunk/mod.rs
```diff
@@ -244,8 +258,8 @@ impl Chunk {
 
     let p = PathBuf::from(chunk_name);
     let p = if p.is_absolute() {
-      if let Some(ref preserve_modules_root) = options.preserve_modules_root {
-        if chunk_name.starts_with(preserve_modules_root) {
```
sapphi-red, line 248: I think this has the same problem with https://github.com/rolldown/rolldown/pull/8817#discussion_r3007850866

younggglcy:
Thanks a lot for the patient reviews here. This needs a broader
design pass than I can take on, so I'll close in favor of leaving
it open for someone else to pick up.

PR #8824, opened 2026-03-21
ci: speed up CodSpeed benchmark build by disabling LTO

Boshen:
## Summary

- The CodSpeed benchmark build takes ~18min, with ~14min spent on the final `bench` crate's fat LTO linking phase
- Since CodSpeed uses **simulation mode**, LTO optimizations have zero effect on measurement accuracy
- Split the build/run steps: build with `cargo build --release` using RUSTFLAGS that disable LTO (`-C lto=false`), increase codegen parallelism (`-C codegen-units=16`), and add debug info needed by CodSpeed (`-C debuginfo=1 -C strip=none -g --cfg codspeed`), then place the binary where `cargo codspeed run` expects it
- Expected build time: ~3-4min (down from ~18min)

### DIFF
--- a/.github/workflows/benchmark-rust.yml
+++ b/.github/workflows/benchmark-rust.yml
@@ -51,9 +51,18 @@ jobs:
       - name: Setup Benchmark Input
         run: just setup-bench
 
+      - name: Build benchmark
+        env:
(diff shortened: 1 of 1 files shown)


Boshen:
Down to 7 minutes. Codspeed measures relative perf, so the current regression is fine.

PR #8847, opened 2026-03-22
perf: reduce timing of `dervie_entries_aware_chunk_name`

### LINKED ISSUE #8276: `derive_entries_aware_chunk_name` currently iterates over *all* entry points
`derive_entries_aware_chunk_name` currently iterates over *all* entry points and calls `has_bit` for each one, every time an entries-aware chunk is created. If there can be many manual-code-splitting groups, this becomes O(num_chunks × num_entries). Consider deriving the set entry indices from `bits` (e.g., `BitSet::index_of_one()`) and/or caching a flattened entry list once, so chunk name generation scales better on large projects.

_Originally posted by @Copilot in https://github.com/rolldown/rolldown/pull/8275#discussion_r2792195374_

AliceLanniste:
close  #8276

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/manual_code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/manual_code_splitting.rs
@@ -7,8 +7,8 @@ use std::{
 
 use arcstr::ArcStr;
 use rolldown_common::{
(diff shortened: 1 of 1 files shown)


hyf0 (approved):
Thank you!

PR #8848, opened 2026-03-22
fix: enable arbitrary_precision for serde_json to fix JSON float parsing

elderapo:
## Summary

- Fix floating-point precision loss when bundling JSON files
- Enable `arbitrary_precision` feature for `serde_json` in `rolldown_common`
- Add regression test with 17-significant-digit floats from the Vite bug report

## Problem

When rolldown converts a `.json` file to a JS module, numbers with 17 significant
digits can silently change value. For example:

| JSON source | After bundling (before fix) | Correct f64 |
|---|---|---|
| `114.35143799257997` | `114.35143799257996` | `114.35143799257997` |
| `406.31486713248995` | `406.3148671324899` | `406.31486713248995` |

The root cause is `serde_json`'s default float parser, which has off-by-one
ULP (unit in the last place) errors for certain inputs. The parsed f64 differs
by 1 bit of mantissa from the correct nearest value.

## Fix

Enable `serde_json`'s `arbitrary_precision` feature. This changes the internal
representation of `serde_json::Number` from `f64` to a raw string, and `as_f64()`
then delegates to Rust's standard `str::parse::<f64>()` — which uses the
Eisel-Lemire algorithm and is correctly rounded.

## Test plan

- Added `test_float_17_significant_digits` covering the 4 values from the
  original Vite issue
- All existing `json_to_program` tests pass
- Full `cargo check` passes with no compilation errors

Fixes: https://github.com/vitejs/vite/issues/21982


IWANABETHATGUY (approved):
Thanks, I simplified the test case a little.

PR #8870, opened 2026-03-23
fix: forward test filters through vp run

younggglcy:
## Summary
- stop `vp run` from consuming forwarded test filters by inserting `--` before passthrough args in the affected `just` recipes
- add a `just --dry-run` regression test covering rolldown, rollup, and HMR test command forwarding

## Test Plan
- `vp run --filter rolldown-tests test:main -- justfile.test.ts`
- `just test-node-rolldown -t resolve/alias`

### DIFF
--- a/justfile
+++ b/justfile
@@ -100,7 +100,7 @@ test-node-hmr *args: build build-test-dev-server
   just test-node-hmr-only {{ args }}
 
 test-node-hmr-only *args:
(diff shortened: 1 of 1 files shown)


younggglcy:
After migrating to vp in #8813, commands like `just test-node-rolldown -t <testname>` do not work correctly because `{{ args }}` aren't being forwarded correctly

PR #8895, opened 2026-03-25
fix(lazy-barrel): load import-then-export specifiers when barrel has local exports

### LINKED ISSUE #8799: [Bug]: `lazyBarrel` can emit bare `forwardRef(...)` from a React component imported through a barrel, causing `forwardRef is not defined` at runtime

[rolldown-lazy-barrel-bug 2.zip](https://github.com/user-attachments/files/26157189/rolldown-lazy-barrel-bug.2.zip)

Reproduction steps:
```sh
npm install
npm run reproduce
```

The repro script builds, starts preview, opens the page in Playwright, and verifies the runtime error automatically.

Disclaimer: 🤖 This reproduction was AI-generated, but every line was reviewed by a human and reproduction was manually tested.

### What is expected?

Build succeeds, preview starts, and the page renders without crashing.

The bundle should not reference React APIs as bare identifiers (e.g. `forwardRef`), so there is no runtime failure.

Production build succeeds and preview starts, but the page crashes at runtime with:

```
forwardRef is not defined
```

The emitted bundle contains code shaped like:

```js
u = forwardRef(function (e, t) { ... })
```

In the broken output, forwardRef is emitted as a bare identifier instead of being referenced via a bound React import/namespace.

Minimal trigger conditions observed:

1. build.rolldownOptions.experimental.lazyBarrel = true enabled.
2. Importing a React component through a package barrel entry.
3. Barrel pattern that has _both_ named and default exports in the same barrel:
   ```js
   import AsyncButton from './AsyncButton.js';

   export { AsyncButton };
   export default AsyncButton;
   ```
   If the barrel has only a default export, or only a named export, the bug stops reproducing.
4. sideEffects: false in the package package.json appears to matter (as lazyBarrel does not apply in such case).

### System Info

```Shell
System:
    OS: macOS 26.3
    CPU: (10) arm64 Apple M1 Max
    Memory: 149.84 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.11.1 - /usr/local/bin/node
    Yarn: 1.22.22 - /Users/wojciech.maj/.yarn/switch/bin/yarn
    npm: 11.6.2 - /usr/local/bin/npm
    pnpm: 10.30.1 - /usr/local/bin/pnpm
  Browsers:
    Chrome: 146.0.7680.80
    Firefox: 141.0.3
    Safari: 26.3
```

### Any additional comments?

If the app is changed to import the concrete file directly instead of the barrel, for example:

```js
import AsyncButton from 'react-async-button-bug/dist/AsyncButton.js';
```

the runtime error goes away. This suggests the issue is tied to the `lazyBarrel` optimization path for this barrel pattern rather than to the component module itself that imports forwardRef.

shulaoda:
## Summary

Fixes #8799

When a barrel module uses the `import { a } from './a'; export { a }` pattern (import-then-export) alongside a local export like `export default b`, importing the local export requires the barrel to execute. However, the lazy barrel optimization incorrectly treated the shared import record for `./a` the same as a dedicated re-export record (`export { a } from './a'`), giving it an empty specifier set. This caused `a` to not be properly loaded, breaking the barrel's execution.

The fix introduces an `IsReExportOnly` flag on `ImportRecordMeta` to distinguish dedicated re-export records (`export { .. } from '..'`, `export * as ns from '..'`) from shared import records (`import { a } from '..'; export { a }`). When a barrel has local exports and needs to execute, only dedicated re-export records use empty specifiers — shared import records are loaded with their full specifiers since we cannot statically determine whether their bindings are used by the barrel's own code.


shulaoda:
* **#8895** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8895?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8895?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=8895).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown_common/src/types/lazy_barrel.rs
```diff
@@ -226,7 +229,8 @@ impl BarrelInfo {
         }
         if has_local_export {
           let mut reexports = FxHashSet::with_capacity(self.named.len() + self.star.len());
-          reexports.extend(self.named.values().map(|v| v.record_idx));
+          reexports
+            .extend(self.named.values().filter(|v| v.is_direct_reexport).map(|v| v.record_idx));
```
hyf0, line 233: So the real fix happens here? could you or let cc explain why we need to add `is_direct_reexport`?

hyf0: Also I think you could start to commit design note about related features, especially around these logic to provide context for others.

hyf0 (approved):
LGTM.

crates/rolldown_common/src/types/lazy_barrel.rs
shulaoda: Yes, I am very clear about the current Lazy Barrel Optimization feature, so I asked CC to help me quickly analyze it and fix it through my idea.

shulaoda: I will improve the performance of the current Lazy Barrel Optimization after this.

shulaoda: > Also I think you could start to commit design note about related features, especially around these logic to provide context for others.

I will do it after I finish improving the current implementation.

PR #8922, opened 2026-03-26
fix: preserve init calls for named re-exports under strict execution order

### LINKED ISSUE #8777: [Bug]: strictExecutionOrder + sideEffects: false drops init calls for named re-exports, leaving variables undefined

https://stackblitz.com/edit/rolldown-rolldown-starter-stackblitz-aeizewzc?file=dist%2Findex.mjs

### What is expected?

Foo and Bar should be defined objects. The `init_Foo` wrapper (for Foo/index.js) should call `init_Foo$1()` (for Foo.js) so that the Foo variable gets assigned:                                       
                                                                                                                                                                                                     
```js
  var init_Foo = __esmMin((() => {                                                                                                                                                                   
      init_Foo$1();        // ← initialize Foo.js (assigns the Foo variable)                                                                                                                         
      init_fooClasses();                                                                                                                                                                             
      init_fooClasses();                                                                                                                                                                             
  })); 
```

Foo and Bar are undefined. The Foo.js module wrapper becomes anonymous and is never called. The Foo/index.js wrapper only initializes fooClasses, skipping Foo.js:

```js
  // Foo.js — anonymous wrapper, never called! Foo stays undefined.
  var Foo;
  __esmMin((() => {
      init_fooClasses();
      Foo = { name: "Foo", root: fooClasses.root };
  }));
```

```js
  // Foo/index.js — missing init_Foo$1() call
  var init_Foo = __esmMin((() => {
      init_fooClasses();   // ← only classes — no Foo.js init!
      init_fooClasses();
  }));
```

### System Info

```Shell
rolldown v1.0.0-rc.9
Node.js v22.21.1
Linux x86_64
```

### Any additional comments?

A lot of iterating found that we needed the following to be true to hit this:

1. strictExecutionOrder: true
2. "sideEffects": false in a parent package.json
3. Nested barrel re-export pattern (barrel → sub-barrel → component)

This is true seemingly in our usage of the mui material-ui package which is how we first encountered it.

A very very rough and novice look suggests that this is because in https://github.com/rolldown/rolldown/blob/e17f131a39d4fa8891cefa8f5a61ff0ee690505b/crates/rolldown/src/stages/link_stage/reference_needed_symbols.rs#L187-L194

For a named re-export like `export { default } from './Foo.js'`:

- is_reexport_all is false (it's not export *)
- importee.side_effects is UserDefined(false) (from package.json)

So stmt_info.side_effect is set to false

https://github.com/rolldown/rolldown/blob/e17f131a39d4fa8891cefa8f5a61ff0ee690505b/crates/rolldown/src/stages/link_stage/tree_shaking/include_statements.rs#L647-L665

The `export { default } from './Foo.js' `re-export in Foo/index.js is excluded from stmt_info_included. When `module_finalizers/mod.rs:remove_unused_top_level_stmt()` later processes Foo/index.js, it never sees this re-export statement, never calls `transform_or_remove_import_export_stmt` for it, and never
  generates the `init_Foo$1() `call that would initialise the Foo variable.

Then, UserDefined(false) from package.json short-circuits
https://github.com/rolldown/rolldown/blob/e17f131a39d4fa8891cefa8f5a61ff0ee690505b/crates/rolldown/src/stages/link_stage/tree_shaking/determine_side_effects.rs#L54-L56
while normally `determine_side_effects` would detect the wrapped ESM importee via the `export * from / WrapKind::Esm` check and return `Analyzed(true)`.

The semantics here seem kinda deliberate so I wanted to discuss but I can raise a change that marks re-export statements as always having side_effect: true when strictExecutionOrder is enabled, importee has WrapKind:Esm? But I'm open to any solutions.

Apologies again if this overview is totally in the wrong direction, first forays into rolldown and the bundler internals so feel free to course correct

teee32:
## Summary
- keep named re-export records side-effectful when strict execution order wraps an ESM importee, so their generated init calls are not tree-shaken away just because the package is marked sideEffects: false
- add an issue 8777 regression fixture that snapshots and executes an export-default re-export under `strictExecutionOrder: true`
- verify the fix with `cargo test -p rolldown`

## Root cause
`reference_needed_symbols.rs` only treated wrapped ESM import records as side-effectful for `export *` or for importees whose package metadata reported side effects. Under `strictExecutionOrder`, named re-exports still need their generated `init_*()` call to run so wrapped export bindings are assigned before use, even when the importee is otherwise side-effect-free.

## Testing
- `cargo test -p rolldown`

Fixes #8777.

### DIFF
--- a/crates/rolldown/src/stages/link_stage/reference_needed_symbols.rs
+++ b/crates/rolldown/src/stages/link_stage/reference_needed_symbols.rs
@@ -190,11 +190,10 @@ impl LinkStage<'_> {
                           self.options.is_strict_execution_order_enabled()
                             && rec.meta.contains(ImportRecordMeta::IsReExportOnly)
                             && !is_reexport_all;
-                        stmt_info.side_effect =
-                          (is_reexport_all
-                            || importee.side_effects.has_side_effects()
-                            || needs_init_for_named_reexport)
-                            .into();
+                        stmt_info.side_effect = (is_reexport_all
+                          || importee.side_effects.has_side_effects()
+                          || needs_init_for_named_reexport)
+                          .into();
                         // Reference to `init_foo`
                         stmt_info
                           .referenced_symbols
--- a/crates/rolldown/src/stages/link_stage/reference_needed_symbols.rs
+++ b/crates/rolldown/src/stages/link_stage/reference_needed_symbols.rs
@@ -186,8 +186,15 @@ impl LinkStage<'_> {
                       }
                       WrapKind::Esm => {
                         // Turn `import ... from 'bar_esm'` into `init_bar_esm()`
+                        let needs_init_for_named_reexport =
+                          self.options.is_strict_execution_order_enabled()
+                            && rec.meta.contains(ImportRecordMeta::IsReExportOnly)
+                            && !is_reexport_all;
                         stmt_info.side_effect =
-                          (is_reexport_all || importee.side_effects.has_side_effects()).into();
+                          (is_reexport_all
+                            || importee.side_effects.has_side_effects()
+                            || needs_init_for_named_reexport)
+                            .into();
                         // Reference to `init_foo`
                         stmt_info
                           .referenced_symbols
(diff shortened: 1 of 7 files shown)


IWANABETHATGUY:
Your tests have already passed in the main branch.

Issue #8345, opened 2026-02-16
`platform: "node"` causes CJS module exports to be silently dropped (immer)


silverwind:
*Filed by [@silverwind](https://github.com/silverwind) via Claude Code*

## Bug Report

When `platform: "node"` is set, rolldown silently drops all `exports.xxx = ...` assignments from immer's CJS development build (`immer.cjs.development.js`). The wrapped module returns an empty `{}` object at runtime. The production build (`immer.cjs.production.min.js`) is unaffected.

Removing `platform: "node"` fixes the issue.

## Reproduction

```bash
mkdir rolldown-repro && cd rolldown-repro
npm init -y
npm install rolldown@1.0.0-rc.4 immer@9.0.21
```

**entry.js:**
```js
const immer = require("immer");
console.log("default:", typeof immer.default, "produce:", typeof immer.produce);
console.log("keys:", Object.keys(immer).slice(0, 5));
```

**rolldown.config.mjs:**
```js
import { defineConfig } from "rolldown";

export default defineConfig({
  input: "entry.js",
  platform: "node",
  output: { format: "esm", file: "dist/bundle.js" },
});
```

```bash
npx rolldown -c rolldown.config.mjs && node dist/bundle.js
```

## Expected

```
default: function produce: function
keys: [ 'Immer', 'applyPatches', 'castDraft', 'castImmutable', 'createDraft' ]
```

## Actual

```
default: undefined produce: undefined
keys: []
```

## Analysis

In the bundle, `require_immer_cjs_development` wraps the development build's source but the `exports.xxx = ...` assignments at the end of the file are completely removed. The wrapped callback only contains local variable declarations, ending with `var produce = immer.produce;` — no exports.

The production build (`require_immer_cjs_production_min`) retains its exports and works correctly. The key difference: the production build uses comma-separated assignments on a single expression statement (`exports.Immer = nn, exports.default = tn, ...`), while the development build uses separate `exports.xxx = ...;` statements.

Removing `platform: "node"` causes both builds to bundle and work correctly.

## Workaround

Define `process.env.NODE_ENV` as `"production"` at build time so the broken development build is tree-shaken:

```js
defineConfig({
  // ...
  env: { NODE_ENV: "production" },
});
```

## Versions

- rolldown: 1.0.0-rc.3 and 1.0.0-rc.4
- immer: 9.0.21
- Node.js: v22+
- OS: macOS and Linux

silverwind:
[...]

PR #8925, opened 2026-03-26
fix: track CJS re-export import records to fix inline const and tree-shaking

h-a-n-a:
## Summary

When a CJS module conditionally re-exports from multiple sources (e.g., React's `jsx-dev-runtime.js`), the linker previously only followed the first import record. This could lead to incorrect inline const values and incomplete tree-shaking.

This PR:
- Records which import records correspond to `module.exports = require(...)` during scanning, rather than relying on `import_records.first()`
- Follows all CJS re-export targets when building `resolved_exports`
- Tracks conflicting CJS exports when multiple targets provide the same name
- Skips inline const for exports with CJS conflicts, since the runtime value depends on which branch executes
- Uses the recorded import record indices to simplify the tree-shaking bailout heuristic

## Background

```js
// jsx-dev-runtime.js
if (process.env.NODE_ENV === 'production') {
  module.exports = require('./production');   // exports.jsxDEV = void 0
} else {
  module.exports = require('./development');  // exports.jsxDEV = function() {...}
}
```

Previously, `add_exports_for_export_star` used `import_records.first()` to find the re-export target. This meant only the first `require()` target's exports entered `resolved_exports` — the rest were silently dropped. This caused `jsxDEV` to be incorrectly inlined as `void 0`.

## Related issues

- https://github.com/vitejs/vite/issues/21884 — the `void 0` inlining case fixed by this PR
- https://github.com/rolldown/rolldown/issues/8345#issuecomment-3908370251 — related discussion on the underlying issue
- https://github.com/vitejs/vite/issues/21843 — related Vite issue, worked around by https://github.com/vitejs/vite/pull/21865 (disabling `inlineConst` for affected packages). This PR fixes the root cause in rolldown so we can reenable `inlineConst` in Vite. cc @sapphi-red 

## Test plan

- [x] New `inline_const_cjs_reexport` - validates inline const works with multiple cjs re-exports
- [x] New `tree_shaking/cjs_reexport_conditional` — validates ESM+CJS mixed re-export (ESM first, CJS second)
- [x] `just test-rust` all green

### DIFF
--- a/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
+++ b/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
@@ -62,6 +62,7 @@ impl PartialEq for MatchImportKindNormal {
 }
 
 #[derive(Debug, PartialEq, Eq)]
+#[expect(clippy::box_collection)]
 pub enum MatchImportKind {
   /// The import is either external or not defined.
   _Ignore,
@@ -81,7 +82,7 @@ pub enum MatchImportKind {
   // The import resolved to multiple symbols via "export * from"
   Ambiguous {
     symbol_ref: SymbolRef,
-    potentially_ambiguous_symbol_refs: Vec<SymbolRef>,
+    potentially_ambiguous_symbol_refs: Box<Vec<SymbolRef>>,
   },
   NoMatch,
 }
@@ -145,12 +146,7 @@ impl LinkStage<'_> {
         .named_exports
         .iter()
         .map(|(name, local)| {
-          let resolved_export = ResolvedExport {
-            symbol_ref: local.referenced,
-            potentially_ambiguous_symbol_refs: None,
-            came_from_cjs: local.came_from_commonjs,
-          };
-          (name.clone(), resolved_export)
+          (name.clone(), ResolvedExport::new(local.referenced, local.came_from_commonjs))
         })
         .collect::<FxHashMap<_, _>>();
 
@@ -222,7 +218,7 @@ impl LinkStage<'_> {
         {
           let main_ref = self.symbols.canonical_ref_for(resolved_export.symbol_ref);
 
-          for ambiguous_ref in potentially_ambiguous_symbol_refs {
+          for ambiguous_ref in potentially_ambiguous_symbol_refs.iter() {
             let ambiguous_ref = self.symbols.canonical_ref_for(*ambiguous_ref);
             if main_ref != ambiguous_ref {
               continue 'next_export;
@@ -319,12 +315,19 @@ impl LinkStage<'_> {
       return;
     };
 
-    let is_cjsreexports = module.ast_usage.contains(EcmaModuleAstUsage::IsCjsReexport);
+    let cjs_reexport_modules: Vec<ModuleIdx> =
+      if module.ast_usage.contains(EcmaModuleAstUsage::IsCjsReexport) {
+        module
+          .ecma_view
+          .cjs_reexport_import_record_ids
+          .iter()
+          .filter_map(|&rec_idx| module.import_records[rec_idx].resolved_module)
+          .collect()
+      } else {
+        vec![]
+      };
 
-    let cjs_reexport_module =
-      is_cjsreexports.then(|| module.import_records.first().unwrap().into_resolved_module());
-
-    for dep_id in module.star_export_module_ids().chain(cjs_reexport_module) {
+    for dep_id in module.star_export_module_ids().chain(cjs_reexport_modules) {
       let Module::Normal(dep_module) = &normal_modules[dep_id] else {
         continue;
       };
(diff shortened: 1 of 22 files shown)


h-a-n-a:
* **#8925** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8925?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/8925?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=8925).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown/tests/rolldown/tree_shaking/commonjs_reexport_conditional/_config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "config": {
+    "treeshake": true
+  }
+}
```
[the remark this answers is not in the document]
h-a-n-a: `tree_shaking/cjs_reexport_conditional_mixed` is removed as duplication.

crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
```diff
@@ -348,20 +350,26 @@ impl LinkStage<'_> {
...
+              resolved_export
+                .cjs_conflicting_symbol_refs
+                .get_or_insert(Vec::default())
+                .push(named_export.referenced);
+            } else {
+              resolved_export
+                .potentially_ambiguous_symbol_refs
+                .get_or_insert(Vec::default())
+                .push(named_export.referenced);
+            }
```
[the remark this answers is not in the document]
h-a-n-a: Have changed to `resolved_export.came_from_cjs || named_export.came_from_commonjs`. It tracks if either side is from commonjs.

crates/rolldown_common/src/types/resolved_export.rs
```diff
@@ -25,10 +25,20 @@ pub struct ResolvedExport {
   pub potentially_ambiguous_symbol_refs: Option<Vec<SymbolRef>>,
   pub symbol_ref: SymbolRef,
   pub came_from_cjs: bool,
+  /// When multiple CJS sources (conditional re-exports) provide the same export name,
+  /// this tracks the alternative symbols. Unlike ESM ambiguity (which is an error),
+  /// CJS conflicts are expected — only one branch runs at runtime, but statically
+  /// we don't know which.
+  pub cjs_conflicting_symbol_refs: Option<Vec<SymbolRef>>,
```
h-a-n-a: Implementing this is actually not our final goal here. This is only an intermediate solution to unblock things. We'll implement conditional inline const in the end.

crates/rolldown/tests/rolldown/tree_shaking/commonjs_reexport_conditional/_test.mjs
```diff
@@ -0,0 +1,3 @@
+import assert from 'node:assert';
+import './dist/main.js';
+assert.strictEqual(globalThis.foo, 'from-cjs');
```
IWANABETHATGUY: YOu could inline the assertion in `main.js`

crates/rolldown/tests/rolldown/topics/hmr/inline_const_cjs_reexport/main.js
```diff
@@ -0,0 +1,11 @@
+// Reproduces a bug where inline const incorrectly inlines `void 0` for a CJS
+// named export when the CJS entry conditionally re-exports from two files,
+// and one of them (production) sets the export to `void 0`.
+// This mimics the react/jsx-dev-runtime pattern.
+import { jsxDEV } from './jsx-dev-runtime.js';
+
+const show = true;
+const result = show && jsxDEV('span', { children: 'hello' });
+console.log(result);
```
IWANABETHATGUY: Can we use `assert` instead of `console`?

crates/rolldown_common/src/types/resolved_export.rs
IWANABETHATGUY: Since most of `ResolvedExport` don't have `cjs_conflicting_symbol_refs` and `potentially_ambiguous_symbol_refs`, consider using `Box<Vec<>>` instead.

h-a-n-a: Maybe we should use this for both.

h-a-n-a: Size dropped from 64 -> 32

crates/rolldown/tests/rolldown/tree_shaking/commonjs_reexport_conditional/_test.mjs
```diff
@@ -0,0 +1 @@
+import './dist/main.js';
```
IWANABETHATGUY: The generated entry will be automatically executed, so the `_test.mjs` is redundant. see https://github.com/rolldown/rolldown/blob/114437ef596276bae9ec777fe794857c0d4270fa/crates/rolldown_testing/src/integration_test.rs#L602-L609

IWANABETHATGUY (approved):

crates/rolldown/tests/rolldown/tree_shaking/commonjs_reexport_conditional/_test.mjs
h-a-n-a: Thanks for pointing out!

crates/rolldown_common/src/types/resolved_export.rs
IWANABETHATGUY: > Size dropped from 64 -> 32

Nice

PR #8927, opened 2026-03-26
refactor: cleanup and simplify in dynamic_import.rs

ulrichstark:
- extract duplicate code out of match arms
- bail out earlier if it's not a `StaticMemberExpression` with property name `then` to shorten code and avoid `unwrap`
- refactor out dummy return value of `init_dynamic_import_binding_usage_info`

### DIFF
--- a/crates/rolldown/src/ast_scanner/dynamic_import.rs
+++ b/crates/rolldown/src/ast_scanner/dynamic_import.rs
@@ -1,6 +1,6 @@
 use oxc::{
   ast::{
-    AstKind, MemberExpressionKind,
+    AstKind,
     ast::{self, Argument, IdentifierReference},
   },
   span::CompactStr,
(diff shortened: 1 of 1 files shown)


ulrichstark:
I can't reproduce the CI failure locally and it's probably not related to my change. Or is there anything I can do?

crates/rolldown/src/ast_scanner/dynamic_import.rs
```diff
@@ -150,26 +132,17 @@ impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
...
+  fn init_dynamic_import_usage_with_static_member_expr(
     &mut self,
-    parent: &MemberExpressionKind<'ast>,
-    ancestor_len: usize,
     import_record_idx: ImportRecordIdx,
   ) -> Option<FxHashSet<CompactStr>> {
-    let MemberExpressionKind::Static(parent) = parent else {
-      return None;
-    };
-    if parent.property.name != "then" {
```
hyf0, line 162: Why remove this?

ulrichstark: It's now on line 62: https://github.com/ulrichstark/rolldown/blob/817866e26fe51f1ff8a42cfe9eeffc5de1efdac8/crates/rolldown/src/ast_scanner/dynamic_import.rs#L62

IWANABETHATGUY (approved):

hyf0:
Thank you!

PR #8915, opened 2026-03-26
fix: always check circular deps in chunk optimizer

IWANABETHATGUY:
## Summary
- Removes the strict_execution_order bypass in would_create_circular_dependency, which incorrectly skipped circular dependency detection when strict execution order was enabled
- The assumption that init_() wrappers make circular chunk dependencies safe was wrong
- Adds a regression test for issue #8910 and removes the now-invalid prior test

closed https://github.com/rolldown/rolldown/issues/8910

## Test plan
- [x] New integration test at strict_execution_order/issue_8910
- [x] CI passes

Issue #8910, opened 2026-03-26
[Bug]: (regression) circular import generated when using `require` and dynamic import


sapphi-red:
### Reproduction link or steps

https://stackblitz.com/edit/jwjzzvna?file=rolldown.config.js
[REPL](https://repl.rolldown.rs/#eNqNkl1v0zAUhv/Kka9AatImbOsWxGWFuOIC7gjKXOck8/Ds4o/Qrup/5zjuQkBFmhRFfo9fv+fxx5F1rDoybVoscvHoRsGqWWHBBOlemS1XXx+ky5uGi4NQUjQm+F3wTReen5sCPkDxvtaDkS1Y/BmkxTc1y5cxqcwfXc3e0jTud8Z6l8fqtIZ6IKu8DXhajJ1X5J+TjPqVICsKXb2AyKfY7gLHwC2MDNF+PFElqXzgKuA5IsHC8cUZbeyMGKPmiKN+JWJJ+eV/EONW/0GM9j+I5YQYI/5GLCdEa5RqzS+dC6M72c9gL8wk7OUSPqJGyz22sD3AveC2N2CDhmwHF/YBWbaVGvrzoqaTex8sUpk+h5RSXBfl9d1dsb5Zl7frq3c3V7fl/QTdYseDIvhaA0hNsVUaA9QMtbeHrKhZRSIdzfgeaxYdp0X8496j1VxV8I0WnEW2qtn3cdpbRPfAf2AFHVcOx+LOokM74Cbmf5G95pHZzS1pixOL81YKv9mjCF4a/dm2aCsKD8kN8CS17A6fdOq/SS98Hhhp08XQzQx01IqOy3l2+g3l4S0c) (same content, not working due to https://github.com/rolldown/rolldown/issues/8888)

### What is expected?

When running `node dist/entry-1.js`, no output is shown.

### What is actually happening?

When running `node dist/entry-1.js` the following error is shown.
```
TypeError: _entry_1_js.__exportAll is not a function
    at eval (file:///home/projects/wvuoppb3--run/dist/node2-CpG0Mwi-.js:19:49)
    at ModuleJob.run (node:internal/modules/esm/module_job:170:4963)
```

### System Info

```Shell
StackBlitz

  System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
[...]

Issue #8361, opened 2026-02-17
[Bug]: __commonJSMin is not a function since rc4


lazka:
### Reproduction link or steps

I've created an example here:

https://github.com/lazka/rolldown-commonJSMin
I couldn't find a way to make it smaller, I hope it's still useful.

### What is expected?

Things run without errors.

### What is actually happening?

```
TypeError: __commonJSMin is not a function
```

### System Info

```Shell
rolldown 1.0.0-rc.4
```

sapphi-red:
A bit more smaller reproduction: ~~[REPL](https://repl.rolldown.rs/#eNqFUFtugzAQvMrKPyQSIlLVL6p+9QC9AD8G1o1Ts47wui/E3bs2IQ+1UiTwY2ZndryTMqqe1Oid6/0nVZ0nY98qDgklVf/HlKoTwg5HPzJM0KOxhC+ZhhnM6AcoVlnRUEP4lUulUEeX97NgMzUEYOkYuYbFshqQdZWhEnY74D2CsQ4DBHTYMfbQfme0jcyeQD7U3R5Yt8lMJNIQRzsgsXY15BaCAnT7SO+vR7aD/dFsPdVgtAtYnirmhuZtQ2ouVSTL1eE8hvV68/ai2rUCFk/NKKJSoap5jChqfSXN53u6HEOECV+F+bwITxM0kboUG4z3m608TPQyvey6EdfUqthm22s4N1vgOS3pv7DdIVz42zQLteY53ZZEg++jw2oJFuAZHh/+GIjDh5Q6zRhYzb8rMsxx)~~

It seems this isn't reproducing.

sapphi-red:
[...]

PR #8960, opened 2026-03-30
fix: skip circular dependency check for pure ESM chunks in chunk optimizer

AlonMiz:
## Summary

The circular dependency check added by #8371 was correct for chunks containing the runtime module or runtime-dependent modules (e.g., `__commonJSMin`, `__toESM`), but overly conservative for regular ESM chunks.

ESM handles circular imports fine — modules are initialized lazily and circular references are resolved at access time. Only runtime helpers need to be available synchronously before the importing module executes, making circular chunk imports harmful only when runtime is involved.

### The Fix

This scopes the circular dependency check to only apply when the source chunk contains:
1. The runtime module itself, OR
2. Modules that depend on runtime helpers (`depended_runtime_helper`)

For all other pure ESM chunks, the merge is allowed regardless of whether it would create circular chunk imports.

### Impact

In large applications (tested with a Large Angular app with ~2,000 files), the over-conservative cycle detection was causing:
- **+53% more JS files** (1,958 → 3,206)
- **+8% bundle size** (58.72 MB → 63.17 MB)

Bisect data from vitejs/vite#22007:

| Checkpoint | Files | Total Size | Status |
|------------|-------|------------|--------|
| rc.4 baseline | 1,958 | 58.72 MB | ✅ |
| before #8371 | 1,982 | 58.57 MB | ✅ |
| rc.5 (with #8371) | 3,206 | 63.17 MB | ❌ |

### Test Plan

- [x] Existing test `issues/8361` passes (CJS runtime cycle still detected)
- [x] Existing test `issues/8361_2` passes
- [x] Existing test `strict_execution_order/issue_8910` passes
- [x] New regression test `issues/8371_regression` passes
- [x] Full test suite: 1649 passed, 5 failed (all pre-existing failures)

### Related Issues

- Fixes regression from #8371
- Refs: #8361, #8886, #8910, #8915
- Refs: vitejs/vite#22007


AlonMiz:
Closing this PR — verified locally that it doesn't actually reduce chunk count in our large Angular monorepo. 

The Rolldown team has already explored similar approaches (#8886 skipping for strictExecutionOrder, reverted by #8915). The fundamental issue is that the `would_create_circular_dependency` algorithm in #8371 is too conservative, not that it should be skipped for certain chunk types.

The bisect data from vitejs/vite#22007 remains valid — the regression is definitively in #8371. A proper fix would need to make the cycle detection more precise rather than bypassing it.

IWANABETHATGUY:
> Closing this PR — verified locally that it doesn't actually reduce chunk count in our large Angular monorepo.
> 
> The Rolldown team has already explored similar approaches (#8886 skipping for strictExecutionOrder, reverted by #8915). The fundamental issue is that the `would_create_circular_dependency` algorithm in #8371 is too conservative, not that it should be skipped for certain chunk types.
> 
> The bisect data from [vitejs/vite#22007](https://github.com/vitejs/vite/issues/22007) remains valid — the regression is definitively in #8371. A proper fix would need to make the cycle detection more precise rather than bypassing it.

https://github.com/rolldown/rolldown/pull/8960/changes#diff-4b7c73536b8f9beae921ab0f1bca2e7db9e795ffb04446d45e825780c0340650
Is this your reproduction?

PR #8982, opened 2026-04-01
Addressing PR comments

shulaoda:


Brooooooklyn (changes requested):
Instead of using pinned versions, you should run `pnpm dedupe --check` in Rolldown CI to avoid any duplication of packages.
See https://github.com/voidzero-dev/vite-plus/blob/main/.github/workflows/ci.yml#L181

sapphi-red:
I think `pnpm dedupe` won't help here. What we want to do here is to sync the versions of `@emnapi/core`, `@emnapi/runtime` with the one in `@rolldown/bindings-wasm32-wasi`. Because `@rolldown/bindings-wasm32-wasi` is generated at build-time, I think we cannot sync them by `pnpm dedupe`.

Brooooooklyn:
@claude[agent] resolve conflict

PR #8993, opened 2026-04-02
fix: emit namespace declaration for empty modules in manual chunks

privatenumber:
## Summary

- Fix undefined export binding when an empty module is dynamically imported and routed to a named chunk via `manualChunks`
- Root cause: `finalized_module_namespace_ref_usage` gated on `ExportsKind::Esm` before checking `SimulateFacadeChunk`, so empty modules (`ExportsKind::None`) had their namespace removed despite the chunk optimizer marking it as needed

## Problem

When all three conditions are met:
1. **Empty module** (0 bytes, e.g. CSS extracted by Vite)
2. **`manualChunks`** routing the module to a named chunk
3. **Non-top-level dynamic import** (inside a function — top-level `await import()` gets inlined)

The output contained `export { empty_exports as t }` without ever defining `empty_exports`, causing a `SyntaxError` at runtime.

## Fix

Check `SimulateFacadeChunk` before the `ExportsKind` gate in `finalized_module_namespace_ref_usage`, since the chunk optimizer's decision is authoritative regardless of exports kind.

## Test plan

- [x] Added regression test `packages/rolldown/tests/fixtures/issues/empty-module-manual-chunks/`
- [x] 721 Node.js tests pass
- [x] 1662 Rust tests pass (1 pre-existing test262 submodule failure)

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/code_splitting.rs
@@ -534,11 +534,17 @@ impl GenerateStage<'_> {
         let meta = &self.link_output.metas[module_idx];
 
         let module_namespace_included_reason = &meta.module_namespace_included_reason;
-        let is_namespace_referenced = matches!(m.exports_kind, ExportsKind::Esm)
-          && if module_namespace_included_reason.intersects(
-            ModuleNamespaceIncludedReason::Unknown
-              | ModuleNamespaceIncludedReason::SimulateFacadeChunk,
-          ) {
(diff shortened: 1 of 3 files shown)


hyf0:
@IWANABETHATGUY Could you check this one? The title mention manual chunks but I saw the fix isn't quite related to manual chunks.

hyf0:
Test failed without the fix. Change is small. LGTM.

hyf0 (approved):

IWANABETHATGUY (approved):

hyf0:
Thank you!

PR #9017, opened 2026-04-07
docs(meta/design): introduce module tags

hyf0:
## Summary

Design doc for module tags — a mechanism to attach string labels to modules based on their loading context, enabling smarter manual code splitting.

## Key design decisions

- **Built-in tags**: `$initial` (statically imported by user-defined entries or part of their dependency chain) and `$lazy` (reachable only via `import()`) — `$` prefix distinguishes built-in from user-defined tags
- **Mutually exclusive**: a module imported both statically and dynamically gets `$initial` only — it's already a dependency of an initial entry
- **`tags` filter**: AND semantics, `string[]` for Phase 1, extensible to `{ include, exclude }` object form in the future
- **Reserved names**: names starting with `$` are reserved for built-in tags

## Implementation phases

- **Phase 1**: `$initial` tag + `tags` filter in manual code splitting groups
- **Phase 2**: `$lazy` tag
- **Future**: per-entry tags (`$entry:<name>`), user-defined tags, magic comments, plugin API, `tags` object form

## Test plan

- [ ] Review design doc for completeness and correctness

### DIFF
--- a/meta/design/module-tags.md
+++ b/meta/design/module-tags.md
@@ -1,0 +1,106 @@
+# Module Tags
+
+## Summary
+
+Module tags attach a set of string labels to modules, allowing them to be identified and filtered in other processes — most importantly in manual code splitting groups. Tags can be user-defined (via entry config) or built-in (computed by rolldown based on the module graph).
+
+## Motivation
+
+Several real-world scenarios require knowing _how_ a module is loaded, not just _what_ it is:
+
+1. **Splitting initial-loading code from async code** — webpack's `chunks: 'initial' | 'async'`. In large apps, hundreds or thousands of modules in main.js's static import tree create a render-blocking request waterfall. Capturing them into a manual code splitting group and using `maxSize` to split into parallel chunks reduces initial requests with zero over-fetch.
+
+2. **Identifying static import chains** — VitePress needs to know which modules are statically reachable from an entry to create a framework chunk.
+
+3. **Separating server and browser code** — Framer bundles server + browser entries in the same build. Chunking rules must avoid putting server-only code into browser chunks.
+
+All three are solved by tagging modules with metadata about their loading context, then filtering on those tags in code splitting groups.
+
+## Design
+
+### Built-in tags
+
+Built-in tags use a `$` prefix to distinguish them from user-defined tags. Names starting with `$` are reserved — user-defined tags must not use the `$` prefix.
+
+- **`$initial`** — the module is statically imported by at least one user-defined entry point, or is part of its static dependency chain. These modules are render-blocking: the browser must fetch and execute them all before the entry can run. A module that is both statically and dynamically imported still gets `$initial`, since it's already a dependency of an initial entry regardless of the dynamic `import()`.
+- **`$lazy`** — the module is reachable only via `import()` and is NOT in any entry's static import chain. `$initial` and `$lazy` are mutually exclusive.
+
+### Using tags in manual code splitting
+
+Groups can filter by tags using `tags`:
+
+```js
+output: {
+  codeSplitting: {
+    groups: [
+      {
+        name: 'initial-deps',
+        tags: ['$initial'],
+        maxSize: 1048576,
+        priority: 10,
+      },
+    ],
+  },
+}
+```
+
+`tags` uses AND semantics: only modules that have **all** specified tags are captured by the group. It combines with other filters (`test`, `minShareCount`, etc.) — a module must match all criteria.
+
+### Implementation priority
+
+**Phase 1:**
+
+- Built-in `$initial` tag — computed from the module graph during the generate stage
+- `tags` filter in manual code splitting groups
+- This alone enables initial-load parallelization: capture initial-loading modules into reasonably-sized parallel chunks
+
+**Phase 2:**
+
+- Built-in `$lazy` tag
+
+### Future directions
+
+These are not planned for initial implementation but could be added based on real-world needs:
+
+- **Per-entry tags** — `$entry:<name>` to identify which entry a module is reachable from. With `{ input: { main: './main.ts', admin: './admin.ts' } }`, a module reachable from both gets `$entry:main` and `$entry:admin`. Combined with `$initial`, this enables per-entry grouping:
+
+  ```js
+  groups: [
+    // Initial deps shared by both entries
+    { name: 'shared-initial', tags: ['$initial', '$entry:main', '$entry:admin'] },
+    // Initial deps only for admin
+    { name: 'admin-initial', tags: ['$initial', '$entry:admin'] },
+  ];
+  ```
+
+  This also works for dynamic entries by naming them via magic comments:
+
+  ```js
+  import(/* rolldownEntryName: 'admin' */ './admin-panel.js');
+  // rolldown automatically tags reachable modules with $entry:admin
+  ```
+
+  ```js
+  { name: 'admin-deps', tags: ['$entry:admin'] }
+  ```
+
+- **User-defined tags on entries** — `tags: ['browser']` in input config, propagating to reachable modules. Enables server/browser code separation in multi-entry builds.
+- **Magic comments** — tag modules at import sites via magic comments (e.g., `import(/* rolldownTag: 'heavy' */ './chart.js')`). Semantics for conflicting tags at different import sites need further design (see Unresolved Questions).
+- **Plugin API** — expose tags in hooks like `moduleParsed` for plugin-driven tagging.
+- **Object form for `tags`** — extend `tags` to accept `{ include, exclude }` for negation:
+
+  ```js
+  { name: 'initial-deps', tags: { include: ['$initial'], exclude: ['$entry:admin'] } }
+  ```
+
+  `tags: string[]` would be shorthand for `{ include: [...] }`. Type: `string[] | { include?: string[], exclude?: string[] }`.
+
+## Unresolved Questions
+
+- Should tags be exposed in the plugin API (e.g., in `moduleParsed` hook)? Useful for plugins but adds API surface.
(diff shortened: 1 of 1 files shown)


meta/design/module-tags.md
```diff
@@ -0,0 +1,96 @@
...
+codeSplitting: {
+  groups: [
+    {
+      name: 'initial-deps',
+      matchTags: ['$initial'],
+      maxSize: 1048576,
+      priority: 10,
+    },
+  ],
+}
```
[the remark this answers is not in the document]
hyf0: Fixed — added the `output: { ... }` wrapper to match the full config shape.

meta/design/module-tags.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+`matchTags` is a filter: only modules that have ALL specified tags are captured by the group. It combines with other filters (`test`, `minShareCount`, etc.) — a module must match all criteria.
```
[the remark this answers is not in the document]
hyf0: Good catch. Clarified that `matchTags` uses AND semantics as the decided behavior. Moved the OR variant (`matchAnyTag`) to unresolved questions as a deferred consideration.

meta/design/module-tags.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Implementation priority
+
+**Phase 1:**
+- Built-in `$initial` tag — computed from the module graph during the link stage
+- `matchTags` filter in manual code splitting groups
+- This alone enables the Layer 1 optimization: capture initial-loading modules into reasonably-sized parallel chunks
```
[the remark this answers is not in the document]
hyf0: Agreed — replaced with "initial-load parallelization" which is self-explanatory.

meta/design/module-tags.md
```diff
@@ -0,0 +1,103 @@
...
+All three are solved by tagging modules with metadata about their loading context, then filtering on those tags in code splitting groups.
+
+## Design
+
+### Built-in tags
+
+Built-in tag names (`initial`, `lazy`) are reserved. User-defined tags must not use reserved names.
+
+- **`initial`** — the module is in the static import chain of at least one entry. These modules are render-blocking: the browser must fetch them all before the entry can execute.
+- **`lazy`** — the module is reachable only via `import()` and is NOT in any entry's static import chain. `initial` and `lazy` are mutually exclusive: a module that is both statically and dynamically imported gets `initial` only — it's already in the initial load, so the dynamic `import()` is a no-op for chunking.
```
cpojer: Can we use a "Symbol" or a more complex string token, like `$rolldown$initial` instead of using common names such as `initial` or `lazy` here?

meta/design/module-tags.md
```diff
@@ -0,0 +1,103 @@
...
+  This also works for dynamic entries by naming them via magic comments:
+
+  ```js
+  import(/* rolldownEntryName: 'admin' */ './admin-panel.js');
+  // rolldown automatically tags reachable modules with entry:admin
+  ```
+
+  ```js
+  { name: 'admin-deps', byTags: ['entry:admin'] }
+  ```
```
sapphi-red, line 85: Can we mention that `initial` is an alias of specifying all static entrypoints for future reference? (also `lazy` is all dynamic entrypoints)

meta/design/module-tags.md
```diff
@@ -0,0 +1,103 @@
...
+
+Groups can filter by tags using `byTags`:
+
+```js
+output: {
+  codeSplitting: {
+    groups: [
+      {
+        name: 'initial-deps',
+        byTags: ['initial'],
```
cpojer: Why not just `tags`?

sapphi-red (approved):

meta/design/module-tags.md
```diff
@@ -0,0 +1,103 @@
+# Module Tags
+
+## Summary
+
+Module tags attach a set of string labels to modules, allowing them to be identified and filtered in other processes — most importantly in manual code splitting groups. Tags can be user-defined (via entry config or magic comments) or built-in (computed by rolldown based on the module graph).
```
cpojer: It says there will be magic comment support, but it is not yet clear what that will look like (besides potential future direction). Is it correct that it is out of scope for the initial implementation?

meta/design/module-tags.md
```diff
@@ -0,0 +1,103 @@
...
+    { name: 'shared-initial', byTags: ['initial', 'entry:main', 'entry:admin'] },
+    // Initial deps only for admin
+    { name: 'admin-initial', byTags: ['initial', 'entry:admin'] },
+  ];
+  ```
+
+  This also works for dynamic entries by naming them via magic comments:
+
+  ```js
+  import(/* rolldownEntryName: 'admin' */ './admin-panel.js');
```
cpojer, line 79: The RFC is called module tags, but this proposal allows tagging a module at the import site. What if two import sites tag the same module with a different name?

meta/design/module-tags.md
hyf0: I was thinking about `rolldown:initial` and `$initisl` and chose `initial` for its simplity. The previous idea is using `$initial` to prevent being conflict with users' custom tag. If it feels right for you, we could go back to that. `$rolldown$initial` feels too lonh.

meta/design/module-tags.md
hyf0: Two reasons:

1. `byTags` is a filter mechanism, `tags` doesn't give people that impression.
2. https://github.com/rolldown/rolldown/blob/0110dc1277b56e29218967961e8fa37edf2da6fe/meta/design/module-tags.md?plain=1#L92

---

I also was thinking using `tags` too. Other than above reasons, I think it's a perfect name.

meta/design/module-tags.md
hyf0: Not really, but you're right I should explain `initial` means modules that would be loaded initially (include those modules both imported by initial and dynamic entries).

meta/design/module-tags.md
hyf0: magic comment is in the direction part, won't be implemented before discussion. I will make the clarification more clear.

meta/design/module-tags.md
hyf0: I'll move this to unresolved questions section. Currently I think maybe just do a conflict detection. Webpack has similar feature, we could refer to it.

hyf0:
* **#9050** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9050?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9045** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9045?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9017** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9017?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9017?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9017).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

meta/design/module-tags.md
mdong1909: Adding a symbol makes sense to help easily differentiate it as a reserved name & built in.

meta/design/module-tags.md
mdong1909: Sharing Vitest's use of `tags` (https://vitest.dev/guide/test-tags). It's also for filtering, but with some difference in devs using it (script vs config object)

If there's no huge issue or strong opinions, maybe we can use `tags` to be consistent?

meta/design/module-tags.md
cpojer: I think `$initial` and `$lazy` is a good default.

hyf0: Done — renamed to `$initial` and `$lazy` with the `$` prefix. Added a note that names starting with `$` are reserved for built-in tags.

meta/design/module-tags.md
hyf0: Done — expanded the `$initial` description to clarify: "the module is in the static import chain of at least one user-defined entry point. Initial means the module will be loaded as part of the initial page load — these are render-blocking because the browser must fetch and execute them all before the entry can run." Also clarified that a module both statically and dynamically imported still gets `$initial`.

meta/design/module-tags.md
hyf0: Done — renamed `byTags` → `tags` throughout the doc. Agreed it's cleaner and consistent with Vitest's precedent.

meta/design/module-tags.md
hyf0: Clarified — removed "magic comments" from the summary (Phase 1 scope). Magic comments are now listed only under "Future directions" with a note that semantics need further design.

meta/design/module-tags.md
hyf0: Good question — moved magic comments to "Future directions" and added conflicting tags as an unresolved question: "if magic comments are supported in the future, what happens when two different import sites tag the same module with different names? Options include merging (union of all tags), erroring on conflicts, or last-write-wins."

hyf0:
### Merge activity

* **Apr 9, 7:03 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Apr 9, 7:03 AM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Apr 9, 7:06 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9032, opened 2026-04-08
fix(rolldown_plugin_vite_import_glob): normalize resolved alias path to prevent double slashes

shulaoda:
Related to https://github.com/vitejs/vite/issues/22124, I will add a test case in vite side.

### DIFF
--- a/crates/rolldown_plugin_vite_import_glob/src/utils.rs
+++ b/crates/rolldown_plugin_vite_import_glob/src/utils.rs
@@ -300,10 +300,10 @@ impl GlobImportVisit<'_> {
         }),
       );
 
-      let resolved_id = rolldown_utils::futures::block_on(future)
-        .ok()
-        .and_then(Result::ok)
-        .map(|resolved| resolved.id.to_string());
+      let resolved_id =
+        rolldown_utils::futures::block_on(future).ok().and_then(Result::ok).map(|resolved| {
+          PathBuf::from(resolved.id.as_str()).normalize().to_slash_lossy().into_owned()
+        });
 
       if let Some(ref id) = resolved_id
         && Path::new(id.as_str()).is_absolute()


IWANABETHATGUY (approved):

h-a-n-a (approved):

crates/rolldown_plugin_vite_import_glob/src/utils.rs
```diff
@@ -300,10 +300,10 @@ impl GlobImportVisit<'_> {
         }),
       );
 
-      let resolved_id = rolldown_utils::futures::block_on(future)
-        .ok()
-        .and_then(Result::ok)
-        .map(|resolved| resolved.id.to_string());
+      let resolved_id =
+        rolldown_utils::futures::block_on(future).ok().and_then(Result::ok).map(|resolved| {
+          PathBuf::from(resolved.id.as_str()).normalize().to_slash_lossy().into_owned()
```
sapphi-red, line 305: Vite v7 code:
https://github.com/vitejs/vite/blob/cc383e07b66d4c5a9768fcb570e0af812cb8d999/packages/vite/src/node/plugins/importMetaGlob.ts#L662-L669
https://github.com/vitejs/vite/blob/cc383e07b66d4c5a9768fcb570e0af812cb8d999/packages/vite/src/node/plugins/resolve.ts#L212-L219

v8 code:
https://github.com/rolldown/rolldown/blob/56d4f66e2e054d18729ff5baec21ebfee749fbc5/crates/rolldown_plugin_vite_resolve/src/vite_resolve_plugin.rs#L260-L269
https://github.com/vitejs/vite/blob/94fd9e7fe1a7d7970ba5e154d588f88318a72302/packages/vite/src/node/plugins/resolve.ts#L354-L360

shulaoda: Is this just an explanation of the code, or do I need to make further changes?

sapphi-red:
I'm not sure why this would fix https://github.com/vitejs/vite/issues/22124. Isn't it already normalized at https://github.com/rolldown/rolldown/blob/56d4f66e2e054d18729ff5baec21ebfee749fbc5/crates/rolldown_plugin_vite_resolve/src/vite_resolve_plugin.rs#L276 ?
This `.normalize` call is aligned with https://github.com/vitejs/vite/blob/cc383e07b66d4c5a9768fcb570e0af812cb8d999/packages/vite/src/node/plugins/resolve.ts#L212-L219

crates/rolldown_plugin_vite_import_glob/src/utils.rs
sapphi-red: It was for me to understand why this change is needed. Would you elaborate why this change is the way to fix https://github.com/vitejs/vite/issues/22124 (https://github.com/rolldown/rolldown/pull/9032#issuecomment-4205364736)?

shulaoda: > It was for me to understand why this change is needed. Would you elaborate why this change is the way to fix [vitejs/vite#22124](https://github.com/vitejs/vite/issues/22124) ([#9032 (comment)](https://github.com/rolldown/rolldown/pull/9032#issuecomment-4205364736))?

This is because the normalization only targets subpath imports, and it doesn’t fix alias imports. I suspect it might be an issue with the alias plugin, although I’ve aligned the behavior with Vite’s JS import glob implementation.

shulaoda: There is a `normalizePath` here that fixes the issue where alias returns a non-standard path.
https://github.com/vitejs/vite/blob/cc383e07b66d4c5a9768fcb570e0af812cb8d999/packages/vite/src/node/plugins/importMetaGlob.ts#L662

sapphi-red: Ah, I see. I overlooked that `normalizePath` outside.

PR #9047, opened 2026-04-09
fix: correct circular dependency check in facade elimination

h-a-n-a:
## Summary

This PR fixes an case of false-positive circular dependency check in facade chunk elimination that caused unnecessary facade chunks. The following is from a minimal reproduction case that is derived from https://github.com/vitejs/vite/issues/21884. You may refer to the test case in this PR for details.

`common.js` is a shared CJS dependency of both `ar_EG.js` and `en_US.js`. Due to
`includeDependenciesRecursively`, `common.js` gets pulled into whichever group is processed first
(`i18n-ar`). This makes `i18n-en-US` cross-chunk-depend on `i18n-ar` to access `require_common`:

i18n-en-US → i18n-ar → runtime

Since `i18n-ar` depends on the runtime, the old check — simulating merging runtime into `i18n-en-US` —
falsely constructs a cycle:

i18n-en-US → i18n-ar → (runtime = i18n-en-US)  ← cycle here

## Fix

Replaced `would_create_circular_dependency` with `is_reachable`. The old check is to simulate merging runtime into the target (i18n-en-US) instead of just checking if runtime has any reference to the target chunk. This causes a false positive on the case where there're multiple chunks that transitively or directly imports runtime. In this case, it's i18n-en-US that imports i18n-ar that imports runtime, which will be considered as cyclic deps. 

Changing to `is_reachable` fixes this as it only checks if runtime chunk has any imports to the target chunk. This makes more sense because we are not going to merge runtime into any of those chunks.

## Test plan

- [x] New fixture `facade_elimination_with_shared_cjs_dep` — 4 chunks, no facades
- [x] Existing snapshot updates from improved facade elimination

## Related issues

Reduces facade chunks in Lobehub's case. Check out the report made by claude here: https://github.com/vitejs/vite/issues/21884#issuecomment-4212470028

### DIFF
--- a/crates/rolldown/tests/rolldown/code_splitting/facade_elimination_with_shared_cjs_dep/main.js
+++ b/crates/rolldown/tests/rolldown/code_splitting/facade_elimination_with_shared_cjs_dep/main.js
@@ -1,0 +1,22 @@
+import assert from 'node:assert';
+
+// This mimics the way import.meta.glob works in Vite.
+const loaders = {
+  ar: () => import('./vendor/antd/locale/ar_EG.js'),
+  'en-US': () => import('./vendor/antd/locale/en_US.js'),
+};
+
+async function main() {
+  const ar = await loaders.ar();
+  const enUS = await loaders['en-US']();
+
+  assert.strictEqual(ar.locale, 'ar');
+  assert.strictEqual(ar.DatePicker.locale, 'ar');
+  assert.strictEqual(ar.DatePicker.yearFormat, 'YYYY');
+
+  assert.strictEqual(enUS.locale, 'en');
+  assert.strictEqual(enUS.DatePicker.locale, 'en_US');
+  assert.strictEqual(enUS.DatePicker.yearFormat, 'YYYY');
+}
+
+main();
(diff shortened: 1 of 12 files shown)


h-a-n-a:
* **#9047** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9047?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9047?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9047).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown/tests/rolldown/code_splitting/facade_elimination_with_shared_cjs_dep/main.js
hyf0, line 1: Would be nice to check if it's possible to add assertion logic.

IWANABETHATGUY (approved):

PR #9057, opened 2026-04-10
fix: prevent circular runtime helper imports during facade elimination (#8989)

### LINKED ISSUE #8989: [Bug]: circular import generated when using dynamic import and namespace exports

[REPL](https://repl.rolldown.rs/#eNqNVEuP2jAQ/itWTm3FI68KmmqPqOqph/bWVFmTOKxbYyPbYWER/70zdsgD0AqJg8f+5nuMTU5BHWSnQKqKhbO/xq2DrK8nQQnlRqg1Fb9euJkVBS2PpeBloRq7a2xRN29vRUieSPg1l3uqCfa6jdMZdnw121PRsBbEDjulLTldkAgDIRZkVjfsPHHq0ZUbVz/oJgKhCEj51gl9ItQQv2ZVERUxqbXakjyYzZE5BuY8aM03hjnE06gBDtuDmWZ1y97FuPRgjnG2BJU185uASch5rJ0MtV0Leu8HF3WDGym2yLuDwzjDwbn6wcFhbgz7Xor4kRRI1KeIuxS35H5sNymQcpjC1Q+mSEAouRFKRykSqK9SpF0Kxav2+j+MDz8OM6JMnzHpMt5Kw8DuZUTOYUZXP5gxBaF06AY3ejdp5wZB14NwbsCEVkJU6lXOSiVrvhnYuXPijc3n5BuTTFP4Z5D1kTyXVG8U0Y0k0x2545RMp2suyaZtKmp+sI1msA0/w4AljtJwufy8WH6JwyheLBdJ8tx5rlhNGwHec0kIl8Ca+TWBe2PS6uM0zIOsu0T8aOXBZIyIhgj8kNwg4iHCfxCuEMkQ4R88As4Oxg6WaUlFRn4Dvi3Q2AT7LzXYGNUgOqpB4o+js/BMzQv9xzJYNszt7TQzTO/ZCt385BtJcYwGPVEh1OsUaaThSrbW/R100zJW89KuDqxsLIB+6IrpjNRUGM9PyJZLXh+/S+9m5eYP/D0Es/qnA29nD49BwIUaG5z/A/YpFpk=)

### What is expected?

No error happens when executing `dist/entry-2.js`.

The following error happens when executing `dist/entry-2.js`:
```
TypeError: __exportAll is not a function
    at dist/node4-DpoBIw4R.js:3:37
    at ModuleJob.run (node:internal/modules/esm/module_job:430:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:661:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)
```
In other words, the cyclic import "entry-2 -> entry-3 -> node4-hash -> entry-2" is not output.

### System Info

```Shell
System:
    OS: Windows 11 10.0.26200
    CPU: (32) x64 AMD Ryzen 9 9950X 16-Core Processor
    Memory: 10.68 GB / 31.06 GB
  Binaries:
    Node: 24.14.1 - C:\Users\green\.vite-plus\js_runtime\node\24.14.1\node.EXE
    Yarn: 1.22.22 - C:\Users\green\.vite-plus\js_runtime\node\24.14.1\yarn.CMD
    npm: 11.11.0 - C:\Users\green\.vite-plus\js_runtime\node\24.14.1\npm.CMD
    pnpm: 10.33.0 - C:\Users\green\.vite-plus\js_runtime\node\24.14.1\pnpm.CMD
    bun: 1.3.11 - C:\Users\green\AppData\Local\Microsoft\WinGet\Links\bun.EXE
    Deno: 2.7.8 - C:\Users\green\AppData\Local\Microsoft\WinGet\Links\deno.EXE
  Browsers:
    Chrome: 146.0.7680.177
    Edge: Chromium (146.0.3856.84)
    Firefox: 148.0.2 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.26100.7309
  npmPackages:
    rolldown: 1.0.0-rc.13 => 1.0.0-rc.13
```

### Any additional comments?

This seemed to be working with rc.4.
Found at https://github.com/sapphi-red/zarara/issues/5#issuecomment-4062163102

IWANABETHATGUY:
Fixes #8989, a circular import that arises when facade chunk elimination adds new runtime-helper consumers (`__exportAll`, `__toESM`, etc.) to a chunk that already has a transitive forward path back to the chunk currently hosting the runtime module.

### Root cause

`optimize_facade_entry_chunks` runs after the merge phase has already placed the runtime module into some chunk based on bitset assignment. When facade elimination later calls `include_symbol(namespace_object_ref)` for an eliminated facade, it sets `target_chunk.depended_runtime_helper.insert(ExportAll)` and adds the target to `runtime_dependent_chunks`. If the target chunk transitively reaches the chunk currently hosting the runtime, the new helper-import edge closes a cycle.

For example, in the regression test the merge phase puts runtime in `chunk(node2)` which has a forward path `node2 -> node3 -> node4`, and facade elimination then makes `chunk(node4)` need `__exportAll` from `chunk(node2)` — closing `entry2 -> entry3 -> node4 -> entry2`.

### Algorithm

The fix replaces the original "place runtime once and never move it" branch in `optimize_facade_entry_chunks` with a two-step decision that runs whenever `runtime_dependent_chunks` is non-empty (i.e., facade elimination has actually added new consumers).

**Step 1 — Peel decision (cycle prevention):**

The runtime is peeled out of its current host chunk only when both predicates hold:

- `host_has_other_modules` — the host chunk contains modules besides the runtime. If runtime is alone in its chunk, that chunk is already a leaf with no outgoing imports and cannot participate in a cycle. Skipping this case avoids leaving an empty chunk that would crash downstream code expecting `chunk.modules[0]` to exist.

- `has_external_consumer` — `runtime_dependent_chunks` contains a chunk that is *not* the host. If every facade-elim consumer IS the host itself, then helpers don't have to cross any chunk boundary and no new edges are introduced.

When both are true, the implementation removes `runtime_module_idx` from the host chunk's `modules` vec via `swap_remove` (ordering doesn't matter — `sort_chunk_modules` re-establishes it later) and clears `module_to_chunk[runtime_module_idx]`.

**Step 2 — Placement decision (consumer-set count):**

When the runtime is unplaced (either because Step 1 just peeled it, or because the merge phase never placed it), the implementation computes the full set of consumer chunks as:

```
consumer_chunks = (non-removed chunks with non-empty depended_runtime_helper) ∪ runtime_dependent_chunks
```

The first term picks up chunks that already required helpers from the linking stage; the second term picks up chunks that facade elimination just announced. Deduplication is automatic via `FxHashSet`. Then:

- `consumer_chunks.len() == 1` → runtime moves into that single consumer chunk. No extra chunk is created; the lone consumer hosts both its own modules and the runtime.
- `consumer_chunks.len() > 1` → runtime is placed in a fresh `rolldown-runtime.js` leaf chunk created with the runtime's bitset. Every consumer imports from it; the dedicated chunk has zero outgoing edges so cycles are structurally impossible.

### Why both steps are needed

The peel gate exists because relying on the consumer-count alone over-triggers — many tests have multiple consumers that don't actually form cycles (e.g., `import_missing_neither_es6_nor_common_js`, where runtime lives in `foo.js` and `require.js` imports `__toCommonJS` from it without any back-edge). The peel gate keeps these untouched.

The consumer-set count exists because relying on `runtime_dependent_chunks.len()` alone (the original code's optimization) undercounts: it ignores chunks that already required helpers from the linking stage. After peeling, the unioned set correctly identifies whether runtime can piggyback on a single consumer or needs its own home.

### Test results and snapshot impact

Net change against `main`: only `chunk_optimizer.rs` and a new regression fixture at `crates/rolldown/tests/rolldown/issues/8989/`. **Zero existing snapshots are modified.** All 1677 integration tests pass.

The 8989 fixture covers the cycle: 4 entries (`node0`–`node3`), `node4` dynamically imported from `node3`, `node1` namespace-importing `node2` (forcing `__exportAll` for `node2`'s namespace materialization). The output places runtime into `node4.js` (the leaf), which both `entry2` and `entry3` reach through forward-only edges. Acyclic ✓.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -1061,15 +1061,50 @@ impl GenerateStage<'_> {
       include_runtime_symbol(context, runtime, RuntimeHelper::ExportAll);
     }
 
-    // Ensure runtime module is properly assigned to chunk graph
-    if chunk_graph.module_to_chunk[runtime_module_idx].is_none()
-      && !runtime_dependent_chunks.is_empty()
-    {
-      // If only one common chunk was appended with dynamic entry module, we just put runtime module into that chunk.
-      // Else create a new common chunk to store runtime module.
-      let runtime_chunk_idx = match runtime_dependent_chunks.len() {
-        1 => runtime_dependent_chunks.into_iter().next().unwrap(),
-        _ => {
+    // Ensure runtime module is properly assigned to chunk graph.
+    //
+    // Facade elimination above can introduce new runtime-helper consumers
+    // (`runtime_dependent_chunks`). The runtime module may also already be
+    // co-located with other modules in some host chunk from the merge phase.
+    // When the host has other modules AND a facade-elim consumer that is not
+    // the host exists, we have ≥2 distinct consumers that both need helpers
+    // from the host. If the host also has a forward path back to the other
+    // consumer, the dependency graph closes a cycle. Peel the runtime out in
+    // that case and let the placement step below re-home it.
+    if !runtime_dependent_chunks.is_empty() {
+      if let Some(host_idx) = chunk_graph.module_to_chunk[runtime_module_idx] {
+        let host_chunk = &chunk_graph.chunk_table[host_idx];
+        let host_has_other_modules = host_chunk.modules.len() > 1;
+        let has_external_consumer = runtime_dependent_chunks.iter().any(|&c| c != host_idx);
+        if host_has_other_modules
+          && has_external_consumer
+          && let Some(pos) = host_chunk.modules.iter().position(|m| *m == runtime_module_idx)
+        {
+          let host_chunk = &mut chunk_graph.chunk_table[host_idx];
+          host_chunk.modules.swap_remove(pos);
+          chunk_graph.module_to_chunk[runtime_module_idx] = None;
+        }
+      }
+
+      if chunk_graph.module_to_chunk[runtime_module_idx].is_none() {
+        // Pick a placement based on the full set of consumer chunks: every
+        // non-removed chunk with a non-empty `depended_runtime_helper`,
+        // unioned with `runtime_dependent_chunks`. Single consumer → reuse it;
+        // multiple consumers → dedicated leaf `rolldown-runtime` chunk.
+        let removed_chunks = &chunk_graph.post_chunk_optimization_operations;
+        let mut consumer_chunks: FxHashSet<ChunkIdx> = chunk_graph
+          .chunk_table
+          .iter_enumerated()
+          .filter_map(|(idx, chunk)| {
+            (!removed_chunks.contains_key(&idx) && !chunk.depended_runtime_helper.is_empty())
+              .then_some(idx)
+          })
+          .collect();
+        consumer_chunks.extend(runtime_dependent_chunks.iter().copied());
+
+        let runtime_chunk_idx = if consumer_chunks.len() == 1 {
+          consumer_chunks.into_iter().next().unwrap()
+        } else {
           let runtime_chunk = Chunk::new(
             Some("rolldown-runtime".into()),
             None,
@@ -1080,14 +1115,14 @@ impl GenerateStage<'_> {
             None,
           );
           chunk_graph.add_chunk(runtime_chunk)
-        }
-      };
-      chunk_graph.add_module_to_chunk(
-        runtime_module_idx,
-        runtime_chunk_idx,
-        self.link_output.metas[runtime_module_idx].depended_runtime_helper,
-      );
-      module_is_assigned.set_bit(runtime_module_idx);
+        };
+        chunk_graph.add_module_to_chunk(
+          runtime_module_idx,
+          runtime_chunk_idx,
+          self.link_output.metas[runtime_module_idx].depended_runtime_helper,
+        );
+        module_is_assigned.set_bit(runtime_module_idx);
+      }
     }
 
     // Restore the included info back to metas
--- a/crates/rolldown/tests/rolldown/issues/8989/_config.json
+++ b/crates/rolldown/tests/rolldown/issues/8989/_config.json
@@ -1,0 +1,24 @@
+{
+  "config": {
+    "input": [
+      {
+        "import": "./node0.js",
+        "name": "entry0"
+      },
+      {
+        "import": "./node1.js",
+        "name": "entry1"
+      },
+      {
+        "import": "./node2.js",
+        "name": "entry2"
+      },
+      {
+        "import": "./node3.js",
+        "name": "entry3"
+      }
+    ],
+    "minify": false,
+    "preserveEntrySignatures": "allow-extension"
+  }
+}
--- a/crates/rolldown/tests/rolldown/issues/8989/_test.mjs
+++ b/crates/rolldown/tests/rolldown/issues/8989/_test.mjs
@@ -1,0 +1,51 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import assert from 'node:assert';
+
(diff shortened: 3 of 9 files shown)


IWANABETHATGUY:
* **#9062** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9062?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9057** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9057?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9057?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9057).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

h-a-n-a:
I though that [this](https://github.com/rolldown/rolldown/blob/bdf2fcf6edc7f65b565d66fbad33536280b83bda/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs#L798-L805) check in chunk_optimizer already does this job. Is this PR considered as an optimization for the case where runtime module is not inserted to any chunk yet?

IWANABETHATGUY:
> I though that [this](https://github.com/rolldown/rolldown/blob/bdf2fcf6edc7f65b565d66fbad33536280b83bda/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs#L798-L805) check in chunk_optimizer already does this job. Is this PR considered as an optimization for the case where runtime module is not inserted to any chunk yet?

I ran the test case on the main branch, and it still failed. This is a bug fix, not an optimization. You can find more details in the PR description.

IWANABETHATGUY:
> I though that [this](https://github.com/rolldown/rolldown/blob/bdf2fcf6edc7f65b565d66fbad33536280b83bda/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs#L798-L805) check in chunk_optimizer already does this job. Is this PR considered as an optimization for the case where runtime module is not inserted to any chunk yet?

 For #8989, with the PR's fix reverted:

  runtime in chunk_graph = Some(2)        ← runtime is in entry2 chunk
  temp_runtime_chunk_idx = Some(2)        ← translates via identity (entry chunk)

  facade-elim from=4 target=5
    temp(2) → temp(6)
    is_reachable(2, 6) = false            ← check passes, elimination is allowed

  is_reachable(2, 6) returns false in the temp graph even though the real chunk graph has entry2 → entry3 → node4 (chunk_graph indices 2 → 3 → 5). The temp
  graph dependencies are computed once via calc_chunk_dependencies from the per-module metas[m].dependencies snapshot; they reflect what the merge phase
  saw, not what the post-elimination chunk graph looks like.

  Why the check is structurally insufficient

  The check only knows two things at decision time:
  1. Where the runtime currently lives (temp_runtime_chunk_idx)
  2. The temp graph's pre-elimination reachability

  What it can't know:
  - Which chunks will newly depend on the runtime after the elimination loop runs target_chunk.depended_runtime_helper.insert(ExportAll) and
  runtime_dependent_chunks.insert(to_chunk_idx). Those new helper-import edges are exactly what closes the cycle.
  - The post-elimination chunk graph topology, because chunk indices and dependency edges shift once eliminations are committed.

  In #8989, no single in-progress elimination by itself sees a forward path from runtime to target in the temp graph. The cycle only emerges after facade
  elimination has added all the new helper-import edges and the placement step has decided where the runtime ends up. That's a post-condition that a
  pre-condition check cannot model — the check is operating on stale graph state.

shulaoda (approved):

PR #9065, opened 2026-04-10
feat: skip __exportAll for dynamic import facades with static exports

h-a-n-a:


h-a-n-a:
* **#9065** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9065?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9065?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9065).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown/tests/rolldown/code_splitting/facade_elimination_with_shared_cjs_dep/artifacts.snap
```diff
@@ -61,8 +54,8 @@ export { en_US_exports as t };
 import assert from "node:assert";
 //#region main.js
 const loaders = {
-	ar: () => import("./i18n-ar.js").then((n) => n.t),
```
IWANABETHATGUY, line 64: This is incorrect; directly run `main.js`:

Image

after bundling: 

Image

IWANABETHATGUY:
I am not too clear what this pr is going to optimize. Can we add some description or open an issue?

IWANABETHATGUY:
`__exportAll` is also a key point for preserving the module’s original semantics; you can also see the difference in `Object.toStringTag `.

h-a-n-a:
> I am not too clear what this pr is going to optimize. Can we add some description or open an issue?

Sure thing! I will add more information as soon as it finishes. For now, I'm still figuring a few things out. The reason I marked this PR as draft is it's not complete yet.

h-a-n-a:
Closing this — the gate is proxying for the wrong invariant.

  **Why it's wrong**

  The direct-exports path reuses the target chunk's native ESM namespace as the dynamically-imported
  module's namespace. That's only correct when the chunk's top-level export set *is exactly* A's export
  set. The current gate (`dynamic_facade_count_by_target ≤ 1`) only counts *other eliminated facades* — it
   says nothing about whether the target chunk already contains other content modules whose exports are
  live cross-chunk.

  So for `C = {A dynamic, B static}`:

  1. **Broken namespace identity.** `import('./C.js')` has `[Symbol.toStringTag]='Module'` but contains
  B's exports — it's not A's namespace.
  2. **No way to keep both working.** The correct shape is `var a_exports = __exportAll({…}); export {
  a_exports, b_named }` — the existing namespace path. Flat direct exports loses it.
  3. **Check pile grows without hitting the real invariant.** None of the five gates express "chunk's
  native namespace == A's namespace."

  **To work on later**

  - Gate on the real invariant: target chunk contains only the facade's entry module (or A is the sole
  contributor to cross-chunk exports). Everything else falls back to the existing `__exportAll` /
  `a_exports` path.
  - Once the gate is correct, `facade_count ≤ 1` is subsumed and both pre-pass counters go away.
  - Decide between the strict "one content module" check vs. a precise "one cross-chunk exporter" analysis
   — the latter needs data only populated later in `compute_cross_chunk_links`.
  - Add a mixed-chunk regression fixture (A dynamic + B static in the same chunk). The suite only covers
  the single-module case today, which is how this slipped through.
  - Reconsider whether the optimization is worth it at all — it saves one `__exportAll` call per
  qualifying dynamic import; the namespace path is the safe default.

PR #9066, opened 2026-04-10
fix(runtime): release `cb` reference after `__commonJS` factory initialization

### LINKED ISSUE #9063: __commonJS runtime helper retains factory reference after initialization, preventing GC in long-lived processes
## Description

The `__commonJS` runtime helper retains a reference to the factory function (`cb`) after the module has been initialized. In long-lived Node.js processes (e.g. SSR servers), the executed factory can never be garbage collected, causing memory to grow monotonically.

## Context

When loading a Rolldown-built CJS bundle via `vm.createContext` + `vm.runInContext` (a common SSR pattern), module exports are cached in memory for request reuse. Because the `__commonJS` closure keeps holding the factory reference even after execution, the factory cannot be GC'd. This means each time a new version of the bundle is loaded, memory only increases and never drops back.

## Reproduction

A runnable demo is available at:
https://stackblitz.com/edit/vitejs-vite-pn1msfza?file=rolldown-issue.md

Run `node --expose-gc demo.mjs` to reproduce.

## Expected Behavior

After the module factory has been executed and exports cached, the factory reference should be released (e.g. set `cb = null`) so it becomes eligible for garbage collection.

## Suggested Fix

In the `__commonJS` helper, null out the `cb` reference after the factory has been called:

```js
// After calling the factory and caching the result:
cb = null;
```

This is a one-line fix but has significant impact for SSR and other long-lived process use cases where bundles are loaded/unloaded over time.

hyf0-agent:
## Summary

After the first `__require()` call, `mod` is set and `cb` (the factory function) is never accessed again. Without an explicit release, `cb` is permanently retained in the closure and cannot be garbage collected.

This is a memory leak in long-lived Node.js processes (e.g. SSR servers loading bundles via `vm.createContext`): each factory can contain thousands of lines of compiled library code, and a typical bundle has hundreds of CJS modules — all kept in heap indefinitely after initialization.

**Fix:** set `cb = null` after the factory call via comma operator.

```js
// Before
var __commonJS = (cb, mod) => function __require() {
  return mod || (0, cb[__getOwnPropNames(cb)[0]])((mod = { exports: {} }).exports, mod), mod.exports;
};

// After
var __commonJS = (cb, mod) => function __require() {
  return mod || ((0, cb[__getOwnPropNames(cb)[0]])((mod = { exports: {} }).exports, mod), cb = null), mod.exports;
};
```

Note: `__esm` already handles this correctly via the `(fn = 0)` trick. `__commonJS` cannot use the same approach because the factory signature is `(exports, module)` — argument slots are occupied.

Also adds `meta/design/runtime-helpers.md` documenting the design decision.

Fixes #9063

### DIFF
--- a/crates/rolldown/src/runtime/runtime-base.js
+++ b/crates/rolldown/src/runtime/runtime-base.js
@@ -13,11 +13,11 @@ export var __esmMin = (fn, res) => () => (fn && (res = fn((fn = 0))), res);
 export var __commonJS = (cb, mod) =>
   function () {
     return (
-      mod || (0, cb[__getOwnPropNames(cb)[0]])((mod = { exports: {} }).exports, mod), mod.exports
+      mod || ((0, cb[__getOwnPropNames(cb)[0]])((mod = { exports: {} }).exports, mod), cb = null), mod.exports
     );
   };
 export var __commonJSMin = (cb, mod) => () => (
-  mod || cb((mod = { exports: {} }).exports, mod), mod.exports
+  mod || (cb((mod = { exports: {} }).exports, mod), cb = null), mod.exports
 );
 export var __exportAll = (all, no_symbols) => {
   let target = {};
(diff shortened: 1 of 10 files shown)


hyf0-agent:
Replaced by a cleaner PR rebased on upstream/main — no unrelated file changes.

PR #9071, opened 2026-04-11
fix: emit REQUIRE_TLA error when require() loads a module with top-level await

jaehafe:
## Summary                                                                                                    
- Add `REQUIRE_TLA` build error when `require()` loads a module with top-level await
- This was a TODO in `compute_tla.rs:46`: "require TLA module should give a error"
- Rolldown docs state "require TLA module is forbidden" but no error was emitted
                                                                                                                
## Changes                                                                                                                                                                                                                    
- Add `RequireTlaError` event kind and `RequireTla` diagnostic
- After computing TLA status, second pass checks if any `ImportKind::Require` points to a TLA module
- Add integration test in `errors/require_tla/`
                                                                                                                
## Note                                                                                                                                                                                                                       
The existing esbuild test `top_level_await_forbidden_require` may fail because it expects require+TLA to bundle without error. This PR intentionally enforces the restriction documented in `docs/in-depth/tla-in-rolldown.md`. Please advise how to handle that test.


crates/rolldown/src/stages/link_stage/compute_tla.rs
```diff
@@ -61,5 +60,22 @@ impl LinkStage<'_> {
       self.metas[module.idx].is_tla_or_contains_tla_dependency =
         is_tla(module.idx, &self.module_table, &mut visited);
     });
+
+    // Check for require() of TLA modules — this is forbidden.
+    for module in &self.module_table.modules {
```
hyf0: Don't wanna pay an extra traverse cost on emitting such error.

jaehafe: Thanks for the feedback. 

Moved the require+TLA check into the existing `for_each` loop

IWANABETHATGUY:
The diagnosis is neither useful nor accurate; please refer to how esbuild handles this scenario, https://hyrious.me/esbuild-repl/?version=0.23.0&b=%00a.js%00import+%27.%2Fsomething%27+%2F%2F+Deliberately+offset+the+import+record+index%0Aimport+%27.%2Fb%27&b=%00b.js%00import+%27.%2Fc%27&b=%00c.js%00await+0&b=e%00entry.js%00require%28%27.%2Fa%27%29%0Arequire%28%27.%2Fb%27%29%0Arequire%28%27.%2Fc%27%29%0Arequire%28%27.%2Fentry%27%29%0Aawait+0&b=%00something.js%00&o=%7B%0A++bundle%3A+true%2C%0A++format%3A+%22esm%22%0A%7D

crates/rolldown/tests/esbuild/default/top_level_await_forbidden_require/artifacts.snap
```diff
@@ -1,34 +1,32 @@
...
-	await 0;
-}));
-//#endregion
-export default require_entry();
+# Errors
+
+## REQUIRE_TLA
+
+```text
+[REQUIRE_TLA] Error: Module 'entry.js' cannot require 'a.js' because it contains a top-level await.
```
IWANABETHATGUY: `a.js` did not contain any top-level await.
instead, transitive dependency "c.js" contains a top-level await

jaehafe: Updated the diagnostic to match esbuild's behavior

- Added `tla_keyword_span: Option<Span>` to `EcmaView` to capture the await location during scanning.

IWANABETHATGUY:
The overall direction seems good! I'll make a few tweaks to the code directly on your branch to ensure everything fits perfectly.

IWANABETHATGUY (approved):

crates/rolldown/src/ast_scanner/impl_visit.rs
```diff
@@ -219,6 +222,9 @@ impl<'me, 'ast: 'me> Visit<'ast> for AstScanner<'me, 'ast> {
     }
     if is_top_level_await {
       self.result.ast_usage.insert(EcmaModuleAstUsage::TopLevelAwait);
+      if self.result.tla_keyword_span.is_none() {
+        self.result.tla_keyword_span = Some(it.span());
+      }
```
fz6m, line 227:
```suggestion
      self.result.tla_keyword_span.get_or_insert_with(|| it.span());
```

PR #9072, opened 2026-04-11
chore: remove outdated TODO in determine_module_exports_kind

jaehafe:
## Summary
- Resolves TODO comment by `hyf0`: "should check if importer is a js module"


hyf0:
All modules are js modules by default now. You could just remove the todo, the logic change is not necessary.

IWANABETHATGUY (approved):

Issue #8842, opened 2026-03-21
[Bug]: cjs build: `__toESM` fails to resolve import default `jsonc-parser` correctly


Sysix:
### Reproduction link or steps

https://stackblitz.com/edit/ney29w29?file=index.ts

Run:
```
pnpm install
pnpm build
node dist/index.cjs
```

### What is expected?

The program runs successfully and can find `jsonc-parser` parse function.
This was supported in vite7 and got broken with vite8. 

`eslint-plugin-oxlint@v1.55.0` with vite7: https://npmx.dev/package-code/eslint-plugin-oxlint/v/1.55.0/dist%2Fbuild-from-oxlint-config%2Futilities.cjs
`eslint-plugin-xolint@v1.56.0` with vite8: https://npmx.dev/package-code/eslint-plugin-oxlint/v/1.56.0/dist%2Fbuild-from-oxlint-config%2Futilities.cjs

I expect that the 2 following code blocks behave the same:
```
// fails to run build cjs
import Parser from 'jsonc-parser';
Parser.parse();
```

```
// runs build cjs
import {parse} from 'jsonc-parser';
parse();
```

### What is actually happening?

```
❯ node dist/index.cjs
TypeError: can't access property "parse", jsonc_parser.default is undefined
    at <anonymous> (/home/projects/ney29w29/dist/index.cjs:26:1)
    at Module.prototype._compile (node:internal/modules/cjs/loader:160:18351)
    at Module._extensions..js (node:internal/modules/cjs/loader:160:19092)
    at Module.prototype.load (node:internal/modules/cjs/loader:160:16623)
    at Module._load (node:internal/modules/cjs/loader:160:14286)
    at traceSync (node:diagnostics_channel:26:3678)
    at wrapModuleLoad (node:internal/modules/cjs/loader:160:3949)
    at executeUserEntryPoint (node:internal/modules/run_main:179:2180)
    at run_main_module (node:internal/main/run_main_module:155:484)
```

### System Info

```Shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 22.22.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.8.2 - /usr/local/bin/npm
[...]

PR #8987, opened 2026-04-01
fix: emit separate __toESM bindings for mixed ESM/CJS external imports

IWANABETHATGUY:
## Summary
- When a CJS/IIFE/UMD chunk has both ESM and non-ESM importers of the same external module, emits two __toESM bindings — one with node-mode (__toESM(x, 1)) for ESM importers and one without for non-ESM importers
- Previously, all importers shared a single binding, which broke interop when the external has __esModule: true
- Also fixes IIFE/UMD format which never passed the node-mode flag at all

Fixes #8842

## Test plan
- [x] Added 8842 test (single-mode: ESM importer of external in CJS output)
- [x] Added 8842_mixed_chunk test (mixed-mode: both .mjs and .js importers of same external)
- [x] All 1665 integration tests pass

PR #9076, opened 2026-04-12
fix(runtime): prevent `__toESM` double-wrapping for native ESM packages in CJS output

JellyBrick:
When outputting CJS format and the source is ESM (`"type": "module"` or `.mjs`):

(Generated code by rolldown)

```js
let mod = require("esm-package");
mod = __toESM(mod, 1);
new mod.default(...); // TypeError: mod.default is not a constructor (caused by double-wrapping)
```

ref #8842

### DIFF
--- a/crates/rolldown/src/runtime/index.js
+++ b/crates/rolldown/src/runtime/index.js
@@ -90,9 +90,9 @@ export var __toESM = (mod, isNodeMode, target) => (
       // file that has been converted to a CommonJS file using a Babel-
       // compatible transform (i.e. "__esModule" has not been set), then set
       // "default" to the CommonJS "module.exports" for node compatibility.
-      isNodeMode || !mod || !mod.__esModule
-        ? __defProp(target, 'default', { value: mod, enumerable: true })
-        : target,
+      (isNodeMode ? mod && mod[Symbol.toStringTag] === 'Module' : mod && mod.__esModule)
+        ? target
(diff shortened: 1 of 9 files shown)


IWANABETHATGUY:
I don't think it is related to #8842, also this is output generated by esbuild for the newly added test case, same as rolldown, I will investigate it later.


https://hyrious.me/esbuild-repl/?version=0.23.0&b=%00node_modules%2Fesm-pkg%2Findex.js%00export+default+class+Foo+%7B%0A++constructor%28%29+%7B%0A++++this.ok+%3D+true%3B%0A++%7D%0A%7D%0A&b=%00node_modules%2Fesm-pkg%2Fpackage.json%00%7B%0A++%22name%22%3A+%22esm-pkg%22%2C%0A++%22version%22%3A+%221.0.0%22%2C%0A++%22type%22%3A+%22module%22%0A%7D%0A&b=e%00main.mjs%00import+Foo+from+%27esm-pkg%27%3B%0Aimport+assert+from+%27assert%27%3B%0Aconst+instance+%3D+new+Foo%28%29%3B%0Aassert.strictEqual%28instance.ok%2C+true%29%3B%0A&o=%7B%0A++bundle%3A+true%2C%0A++format%3A+%22cjs%22%2C%0A++external%3A+%5B%22assert%22%2C+%22esm-pkg%22%5D%0A%7D

sapphi-red:
Fixed by #8987

PR #9077, opened 2026-04-12
fix(rolldown_plugin_vite_import_glob): use POSIX path join/normalize for glob resolution

shulaoda:
## Summary

- Replace `std::path::Path::join` + `SugarPath::normalize` with `path_posix::join` / `path_posix::normalize` so that backslash-escaped glob syntax (e.g. `\(`, `\)`, `\?`, `\[`) is never mistakenly treated as a Windows path separator on Windows.
- Add `\\` to glob metachar detection in `find_glob_syntax`
- Replace `MAIN_SEPARATOR` checks with literal `b'/'` in `split_path_and_glob_inner` and `get_common_base`
- Slash-convert `dir`, `root`, and walkdir entry paths via `to_slash_lossy` for consistent `/` (only matching)

Related to vitejs/vite#22166, I will add a test case on Vite side.


crates/rolldown_plugin_vite_import_glob/src/utils.rs
```diff
@@ -270,21 +268,25 @@ impl GlobImportVisit<'_> {
...
     };
     let absolute_glob = if let Some(glob) = glob.strip_prefix('/') {
-      root.join(glob)
+      path_posix::join(&[root, glob])
     } else if glob.starts_with("**") {
-      root.join(glob)
+      path_posix::join(&[root, glob])
     } else if glob.starts_with("./") || glob.starts_with("../") {
-      dir.join(glob)
+      path_posix::join(&[&dir, glob])
```
sapphi-red, line 289: Not directly related to this PR, but I think `globSafePath` calls are missing.

sapphi-red (approved):

crates/rolldown_plugin_vite_import_glob/src/utils.rs
shulaoda: > Not directly related to this PR, but I think `globSafePath` calls are missing.

Since we split the path and the glob, and then directly apply `strip_prefix` before performing the glob match, there is no need to apply `globSafePath` to the static path part.

sapphi-red: Ah, makes sense.

PR #8698, opened 2026-03-14
fix: use try_init instead of init for tracing subscriber to avoid panic

cal-gooo:
## Summary

- Replace `.init()` with `.try_init()` for tracing subscriber initialization in `rolldown_tracing` and `rolldown_devtools`
- `.init()` calls `set_global_default()` which panics if a global subscriber is already set; `.try_init()` returns a `Result` instead
- Fixes concurrent build scenarios (e.g. multiple `build()` calls via `Promise.all`) where the second initialization would panic

Closes #8656

## Test plan

- [x] `cargo check -p rolldown_tracing -p rolldown_devtools` passes
- Concurrent builds via vite-plus no longer panic with `SetGlobalDefaultError`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9078, opened 2026-04-12
fix: use try_init() to avoid SetGlobalDefaultError panic

fengmk2:
## Summary

- Replace `.init()` with `.try_init().ok()` in `rolldown_devtools::DebugTracer::init()` and all 3 call sites in `rolldown_tracing::try_init_tracing()`
- Prevents `SetGlobalDefaultError` panic when rolldown is embedded in a host binary that has already installed a global tracing subscriber

## Context

When rolldown is bundled into a host binary (e.g., vite-plus bundles `rolldown_binding` via `pub extern crate`), the host may install a global tracing subscriber during NAPI module init. Later, when `DebugTracer::init()` runs (triggered by `build.rolldownOptions.devtools: {}`), it calls `.init()` which panics:

```
thread '<unnamed>' panicked at tracing-subscriber-0.3.23/src/util.rs:94:14:
failed to set global default subscriber: SetGlobalDefaultError("a global default trace dispatcher has already been set")
```

`.try_init()` returns `Result` instead of panicking, so a previously-installed subscriber is silently accepted.

Ref: voidzero-dev/vite-plus#1356

## Test plan

- [x] `cargo check -p rolldown_devtools -p rolldown_tracing` passes
- [ ] CI passes

### DIFF
--- a/crates/rolldown_devtools/src/init_tracing.rs
+++ b/crates/rolldown_devtools/src/init_tracing.rs
@@ -38,7 +38,8 @@ impl DebugTracer {
     tracing_subscriber::registry()
       .with(DevtoolsLayer.with_filter(devtools_event_filter))
       .with(fmt::layer().event_format(DevtoolsFormatter))
-      .init();
+      .try_init()
+      .ok();
 
     tracer
   }
--- a/crates/rolldown_tracing/src/lib.rs
+++ b/crates/rolldown_tracing/src/lib.rs
@@ -51,7 +51,8 @@ pub fn try_init_tracing() -> Option<Box<dyn Any + Send>> {
       tracing_subscriber::registry()
         .with(Targets::from_str(&env_var).unwrap())
         .with(chrome_layer.with_filter(filter_for_removing_devtools_event))
-        .init();
+        .try_init()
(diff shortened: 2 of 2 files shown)


sapphi-red:
refs https://github.com/rolldown/rolldown/pull/8698#issuecomment-4103651424

sapphi-red:
I guess this would disable the debug file output is used by the `devtools` feature and will break the feature.

fengmk2:
@sapphi-red Indeed, we need to think of a better solution. Do you have any suggestions?

sapphi-red:
I'm not familiar with the code around here and not sure how it's working in details. @hyf0 Do you have any ideas?

fengmk2:
After communicating with @hyf0, I will handle it at the Vite+ side for now to avoid simultaneous registration. The long-term solution needs @hyf0 to take another look.

hyf0:
For the future plan:

https://github.com/rolldown/rolldown/blob/0af953355ba76ccc151e46634a58c4d3143d4512/meta/design/devtools.md?plain=1#L263

Should be able to solve it without causing conflict with other crates.

PR #9101, opened 2026-04-14
fix: await async export-star init wrappers

### LINKED ISSUE #9083: [Bug]: barrel init missing await on async dependency makes export wildcard + transitive top-level await broken
## Description

Issue #8986 was closed as fixed in rc.13 and the fix addressed the *direct* case where a module with TLA is directly re-exported via `export * from`. However, the **transitive** case is still broken in rc.15: when a barrel re-exports a module whose `__esmMin` init function is async *only because of a transitive TLA dependency*, the barrel's init still calls it without `await`.

## Reproduction

```typescript
// deep.ts — has top-level await
export const value = 'hello'
await Promise.resolve()  // TLA here

// middle.ts — transitively depends on deep.ts (no own TLA)
import { value } from './deep.ts'
export const manager = { value, ready: false }
export function setup() { manager.ready = true }

// barrel.ts — pure barrel re-export
export * from './middle.ts'

// main.ts
async function run() {
  const { manager, setup } = await import('./barrel.ts')
  setup()
  console.log(manager.ready) // expected: true, actual: crashes (manager is undefined)
}
run()
```

[REPL](https://repl.rolldown.rs/#eNptUstupDAQ/JUWF5gV60kOe0Gaw2q1t5U2hxy5eEwzODI28mOS0Yh/T9sGQqJICPCjqquq+170RXMvRi41e3HxVxfNtqwLQSvublpAH7Tw0miwQVcHuLcaQBjtPNxh5Jpf0Nbg0IcJZjgBf+XSgxwnY31VsuOZW4uqPERYulWl38hgFDJlLtXCwizy7naA4xHwbULhsWvA24A1cOEDVw0Iy92ADlYISAdBd9hLjR3xzq1OKltNFrBoInqui6xh5/NjIzuleiQXfkBvzQikepRdp7BsLTERQV7ug9o2MkH2S4lcuQpIQaxEHeKUaJYSOblV/mlF1JDMN9Bz5Qi/Q2z5L+l9xJ4DI5JoM2Oi2FhyJ3VdfnKaZWSxJygHVMokmbl9T6ReOqQC1KQrUtHYled/v2FAi2sha5TqzKtmxNbLy67mNydfcso9+5OOt7hWWNnqzT5d5EGl7wao0hBKPQXfLKPGRvScpa06avUDQi8VzYpDlWYJzre0ew7eU5r0IBcDeH6OZCb4xJaYaVK9lcL/fUMRYvb/bYd2GcZ4Yab3HKdsphiu5OuRPbCHn1awx1/F/A4kFyU1)

## What rolldown rc.15 emits

```javascript
var init_middle = __esmMin((async () => {
  await init_deep()        // ✅ correct — middle depends on deep (async)
  manager = { ... }
}))

var init_barrel = __esmMin((async () => {
  init_middle()            // ❌ missing await! init_middle is async
}))
```

Because `init_barrel` does not `await init_middle()`, the Promise from `init_middle` is discarded. When the dynamic import of `barrel.ts` resolves, `init_middle` has not finished, so `manager` is still `undefined`.

## Difference from #8986

#8986 was the case where the module **directly containing TLA** is re-exported:
```
barrel.ts  --export * from-->  deep.ts (has TLA)
```

This issue is the **transitive** case:
```
barrel.ts  --export * from-->  middle.ts (no TLA)  --import-->  deep.ts (has TLA)
```
`middle.ts` has no TLA of its own, but its `__esmMin` init becomes `async` because it must `await init_deep()`. The barrel's init function still does not await this async dependency.

## Real-world impact

Encountered in [secflow](https://git.pku.edu.cn/cclab/secflow), a Node.js SEA bundle:
- `oem/index.ts` has `await enforceLicense()` (TLA)  
- `gateway/routes/index.ts` transitively imports `oem/` via config/middleware, making `init_routes` async  
- `gateway/index.ts` is a pure barrel (`export * from './routes/index.ts'`)
- Bundle output: `init_gateway` calls `init_routes()` without `await`  
- `authManager` (exported from routes) is `undefined` at runtime despite `await import('gateway/index.ts')`

## System Info

```
rolldown: 1.0.0-rc.15 (pinned via yarn resolutions)
tsdown: 0.21.7
Node.js: v25.9.0
```

thezzisu:
## Summary
- await async init wrappers when lowering `export * from` for ESM-wrapped modules with transitive TLA dependencies
- add a regression fixture for issue 9083 covering the transitive barrel case under `strictExecutionOrder`

## Testing
- `INSTA_UPDATE=always cargo run -p rolldown_testing --bin run-fixture crates/rolldown/tests/rolldown/issues/9083/_config.json`
- `INSTA_UPDATE=always cargo run -p rolldown_testing --bin run-fixture crates/rolldown/tests/rolldown/issues/9028/_config.json`
- `cargo fmt --all --check`

Partially fixes #9083 (Bug1 only).

### DIFF
--- a/crates/rolldown/tests/rolldown/issues/9083/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/issues/9083/artifacts.snap
@@ -1,0 +1,59 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+---
+# Assets
+
+## main.js
+
+```js
+//#region deep.js
+const value = "hello";
+await Promise.resolve();
+//#endregion
+//#region middle.js
+const manager = {
+	value,
+	ready: false
+};
+function setup() {
+	manager.ready = true;
+}
+//#endregion
+export { manager, setup };
+
+```
+
+# Variant: [strict_execution_order: true]
+
+## Assets
+
+### main.js
+
+```js
+// HIDDEN [\0rolldown/runtime.js]
+//#region deep.js
+var value;
+var init_deep = __esmMin((async () => {
+	value = "hello";
+	await Promise.resolve();
+}));
+//#endregion
+//#region middle.js
+function setup() {
+	manager.ready = true;
+}
+var manager;
+var init_middle = __esmMin((async () => {
+	await init_deep();
+	manager = {
+		value,
+		ready: false
+	};
+}));
+//#endregion
+await __esmMin((async () => {
+	await init_middle();
+}))();
+export { manager, setup };
+
+```
--- a/crates/rolldown/tests/rolldown/issues/9083/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/issues/9083/artifacts.snap
@@ -7,8 +7,10 @@ source: crates/rolldown_testing/src/integration_test.rs
 
 ```js
 //#region deep.js
-await Promise.resolve();
-await Promise.resolve();
+function sleep(ms) {
+	return new Promise((resolve) => setTimeout(resolve, ms));
+}
+await sleep(100);
 //#endregion
 //#region middle.js
 const manager = {
@@ -35,10 +37,12 @@ export { manager };
 ```js
 // HIDDEN [\0rolldown/runtime.js]
 //#region deep.js
+function sleep(ms) {
+	return new Promise((resolve) => setTimeout(resolve, ms));
+}
 var value;
 var init_deep = __esmMin((async () => {
-	await Promise.resolve();
-	await Promise.resolve();
+	await sleep(100);
 	value = "hello";
 }));
 //#endregion
--- a/crates/rolldown/tests/rolldown/issues/9083/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/issues/9083/artifacts.snap
@@ -7,19 +7,22 @@ source: crates/rolldown_testing/src/integration_test.rs
 
 ```js
 //#region deep.js
-const value = "hello";
+await Promise.resolve();
 await Promise.resolve();
 //#endregion
 //#region middle.js
 const manager = {
-	value,
+	value: "hello",
 	ready: false
 };
 function setup() {
 	manager.ready = true;
 }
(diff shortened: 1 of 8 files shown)


crates/rolldown/tests/rolldown/issues/9083/artifacts.snap
```diff
@@ -0,0 +1,59 @@
...
+	manager = {
+		value,
+		ready: false
+	};
+}));
+//#endregion
+await __esmMin((async () => {
+	await init_middle();
+}))();
+export { manager, setup };
```
IWANABETHATGUY: This test case already passed on the main branch. Can you write a more appropriate test?

crates/rolldown/tests/rolldown/issues/9083/deep.js
```diff
@@ -0,0 +1,3 @@
+export const value = 'hello';
+
+await Promise.resolve();
```
IWANABETHATGUY: Replace `await Promise.resolve()` with
```js
function sleep(ms) {
  return new Promise((resolve) => setTimeout(resolve, ms));
}
await sleep(100)
```
failed on main branch

crates/rolldown/tests/rolldown/issues/9083/artifacts.snap
thezzisu: I reworked the fixture so it now goes through the export-star barrel path and reproduces the async init ordering issue. I also verified that this case fails on main and passes with this fix.

crates/rolldown/tests/rolldown/issues/9083/deep.js
thezzisu: Updated to `sleep(100)` as suggested. I had first used extra microtask hops to keep the fixture a bit lighter, but your method is much more readable and clearer. If test speed is a concern, I’m happy to switch it back.

IWANABETHATGUY (approved):

PR #9104, opened 2026-04-14
fix: await async init in transitive strict-order reexports

### LINKED ISSUE #9083: [Bug]: barrel init missing await on async dependency makes export wildcard + transitive top-level await broken

thezzisu:
## Summary
- partially fixes #9083 (bug 2) in `generate_transitive_esm_init`
- await async ESM init wrappers when strict execution order emits fallback init calls for excluded re-export statements
- keep Bug1 out of scope here; the direct included `export *` path is handled separately in #9101

## Tests
- add `crates/rolldown/tests/rolldown/issues/9083_gen_transitive`
- `cargo test -p rolldown 9083_gen_transitive --test integration`
- `cargo run -p rolldown_testing --bin run-fixture crates/rolldown/tests/rolldown/issues/9083_gen_transitive/_config.json`

### DIFF
--- a/crates/rolldown/src/module_finalizers/mod.rs
+++ b/crates/rolldown/src/module_finalizers/mod.rs
@@ -164,7 +164,17 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
           self.snippet.builder.vec(),
           false,
         );
-        body.push(self.snippet.builder.statement_expression(SPAN, init_call));
+        let init_stmt = if importee_linking_info.is_tla_or_contains_tla_dependency {
+          self.snippet.builder.statement_expression(
+            SPAN,
+            ast::Expression::AwaitExpression(
+              self.snippet.builder.alloc_await_expression(SPAN, init_call),
+            ),
+          )
+        } else {
+          self.snippet.builder.statement_expression(SPAN, init_call)
+        };
+        body.push(init_stmt);
       } else {
         // Importee is not included (barrel module) — traverse its import records
         // to find included importees transitively.
--- a/crates/rolldown/tests/rolldown/issues/9083_gen_transitive/_config.json
+++ b/crates/rolldown/tests/rolldown/issues/9083_gen_transitive/_config.json
@@ -1,0 +1,7 @@
+{
+  "configVariants": [
+    {
+      "strictExecutionOrder": true
+    }
+  ]
+}
--- a/crates/rolldown/tests/rolldown/issues/9083_gen_transitive/_test.mjs
+++ b/crates/rolldown/tests/rolldown/issues/9083_gen_transitive/_test.mjs
@@ -1,0 +1,5 @@
+import assert from 'node:assert';
+import { ready, value } from './dist/main.js';
+
+assert.strictEqual(value, 42);
+assert.strictEqual(ready, true);
--- a/crates/rolldown/tests/rolldown/issues/9083_gen_transitive/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/issues/9083_gen_transitive/artifacts.snap
@@ -1,0 +1,68 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+---
+# Assets
+
+## main.js
+
+```js
+//#region lib.js
+const value = 42;
+//#endregion
+//#region deep.js
+const value$1 = "hello";
+await Promise.resolve();
+//#endregion
+//#region middle.js
+globalThis.__bug2_ready = value$1 === "hello";
+//#endregion
+//#region main.js
+const ready = globalThis.__bug2_ready === true;
+//#endregion
+export { ready, value };
+
+```
+
+# Variant: [strict_execution_order: true]
+
+## Assets
+
+### main.js
+
+```js
+// HIDDEN [\0rolldown/runtime.js]
+//#region lib.js
+var value;
+var init_lib = __esmMin((() => {
+	value = 42;
+}));
+//#endregion
+//#region deep.js
+var value$1;
+var init_deep = __esmMin((async () => {
+	value$1 = "hello";
+	await Promise.resolve();
+}));
+//#endregion
+//#region middle.js
+var init_middle = __esmMin((async () => {
+	await init_deep();
+	globalThis.__bug2_ready = value$1 === "hello";
+}));
+//#endregion
+//#region barrel.js
+var init_barrel = __esmMin((async () => {
+	init_lib();
+	await init_middle();
+}));
+//#endregion
+//#region main.js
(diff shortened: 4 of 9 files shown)


IWANABETHATGUY:
Can we add some assertion? The test seems worked fine on main branch.

thezzisu:
Thanks for the feedback. I had the same concern and am currently auditing all code paths to confirm if this change is necessary. I’ll move this to a draft for now to avoid unnecessary pings while I investigate.

thezzisu:
Closing this one. After tracing the tree-shaking and finalization paths end to end, I don't think the new `await` in `generate_transitive_esm_init()` is reachable for TLA chains. Any re-export that actually leads to TLA is marked side-effectful and stays on the included path, which is already covered by #9101.

If someone finds a real reachable case here, I'm happy to reopen and fix it properly.

thezzisu:
<details>
<summary>Why this looks unreachable</summary>

I traced the path all the way from `generate_transitive_esm_init()` back to tree-shaking.

`generate_transitive_esm_init()` only runs from `remove_unused_top_level_stmt()` for source re-export statements that are already excluded, i.e. `!is_stmt_included`.

For ESM re-exports with a source, `reference_needed_symbols` marks the statement as side-effectful whenever either it is `export * from ...` or the direct importee has side effects:

`stmt_info.side_effect = is_reexport_all || importee.side_effects.has_side_effects()`

That matters because top-level `await` is already treated as a side effect by the side-effect detector, and module side effects are propagated recursively through imports and re-exports in `determine_side_effects`.

So if a re-export chain really leads to a TLA module, the direct importee on that chain will already be side-effectful, which makes the re-export statement side-effectful too. Then `include_statements` includes it before finalization, so it never reaches the excluded-statement branch that calls `generate_transitive_esm_init()`.

That leaves two cases:

1. `export * from ...`: always included because `is_reexport_all` is true.
2. `export { ... } from ...`: only excluded when the direct importee has no side effects, which rules out any chain that actually reaches TLA.

So the new `await` branch in `generate_transitive_esm_init()` appears to be dead for the exact class of cases this PR is trying to fix.

The relevant code is:

- `crates/rolldown/src/module_finalizers/mod.rs` in `remove_unused_top_level_stmt()` and `generate_transitive_esm_init()`
- `crates/rolldown/src/stages/link_stage/reference_needed_symbols.rs`
- `crates/rolldown/src/stages/link_stage/tree_shaking/determine_side_effects.rs`
- `crates/rolldown/src/stages/link_stage/tree_shaking/include_statements.rs`
- `crates/rolldown/src/ast_scanner/side_effect_detector/mod.rs`

I also tried several fixture shapes to disprove this, including direct barrels, bridge modules, proxy chains, and cycle-based chains. Every time a chain genuinely reached TLA, the statement was kept on the included path and the output collapsed back to the path already fixed by #9101.

</details>

PR #9108, opened 2026-04-14
fix: keep browser:false imports ignored when plugins return this.resolve()

### LINKED ISSUE #9107: [Bug]: Returning this.resolve() from resolveId misresolves browser: false imports

## Reproduction link or steps

Minimal repro:

`package.json`

```json
{
  "private": true,
  "type": "module",
  "scripts": {
  },
  "devDependencies": {
    "rolldown": "1.0.0-rc.12"
  }
}
```

`src/main.js`

```js
import "../demo/index.js";
```

`demo/package.json`

```json
{
  "name": "demo",
  "type": "module",
  "browser": {
    "buffer": false
  }
}
```

`demo/index.js`

```js
import "buffer";
```

`rolldown.config.mjs`

```js
export default {
  input: "./src/main.js",
  platform: "browser",
  plugins: [
    {
      name: "probe",
          return null;
        }

          ...options,
          skipSelf: true,
        });

        console.log(id, resolved);
        return resolved ?? null;
      },
    },
  ],
};
```

Run:

```bash
npm install
npm run build
```

Control:

## What is expected?

The build fails with:

```text
demo/index.js:1:8
import "buffer";
       ^^^^^^^^
Is a directory (os error 21)
```

## System Info

```shell
System:
  OS: macOS 26.4
  CPU: (12) arm64 Apple M4 Pro
  Memory: 1020.94 MB / 24.00 GB
Binaries:
Browsers:
  Chrome: 147.0.7727.56
  Firefox: 149.0.2
  Safari: 26.4
npmPackages:
```

AlexanderKolberg:
Closes #9107

This fixes a case where `browser: false` handling was lost if a plugin returned the result of `this.resolve(..., { skipSelf: true })` from `resolveId`.

In normal browser resolution, Rolldown correctly treats those imports as ignored. But in the plugin-return path, the returned `ResolvedId` could lose that ignored state. In the repro, that caused `buffer` to resolve to the package directory and later fail during loading with `Is a directory`.

## Fix

When a plugin returns a resolved id, Rolldown now checks whether the same request would be treated as ignored by the internal resolver. If so, the returned `ResolvedId` is marked as ignored as well.

This keeps the plugin-return path consistent with normal browser resolution.

## Test

- adds a regression fixture for a `browser: false` import resolved through a plugin that returns `this.resolve()`

### DIFF
--- a/crates/rolldown_plugin/src/utils/resolve_id_with_plugins.rs
+++ b/crates/rolldown_plugin/src/utils/resolve_id_with_plugins.rs
@@ -14,6 +14,28 @@ fn is_http_url(s: &str) -> bool {
   s.starts_with("http://") || s.starts_with("https://") || s.starts_with("//")
 }
 
+fn should_preserve_ignored<Fs: FileSystem>(
+  resolver: &Resolver<Fs>,
+  specifier: &str,
+  importer: Option<&str>,
+  import_kind: ImportKind,
+  is_user_defined_entry: bool,
+  resolved_id: &str,
+) -> bool {
+  // `this.resolve()` may surface browser:false mappings as the ignored path itself.
+  // If a plugin returns that object from `resolveId`, recover the ignored bit so
+  // module loading still produces an empty module instead of trying to read the path.
+  match resolver.resolve(
+    importer.map(Path::new),
+    specifier,
+    import_kind,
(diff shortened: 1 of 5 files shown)


shulaoda:
Fixed in #9192

Issue #9005, opened 2026-04-06
The `Value is non of these types` error thrown for invalid return value in function options is difficult to figure the reason


sapphi-red:
```js
import { defineConfig } from 'rolldown'

export default defineConfig({
  input: "./index.js",
  // should return boolean, but returns a number
  external: id => ['foo'].find(name => id === name)
})
```
errors with
```
<anonymous_script>:0


[Error: Value is non of these types `bool`, `UnknownReturnValue`, ] {
  code: 'InvalidArg'
}
```
.
[REPL](https://repl.rolldown.rs/#eNptT81qwzAMfhXhSzrI3JbdAtllj7Hs4Njy6uHIxVG2jOB3n+3RQKEgIfT9CW3Cim4TMXhvwg9JHci6T8lzQUl0j5hW6Ey46RoiwwYGrSN8qzQksDFM0NxszUAD4VqlWagWX+duOGwDATi6LtzBf6SckJWsUAvHI/AFwTqPM8zoUTMaGH8rOi7MgSAXKn0BVmMJw5UxkvI5z0D/Cu+NDaH5kPmoOZCasICF6nso69NAKbdIrcgKXOXX/vy+371c84aYLa1A0XFcMGXzd1ac5UmenqOW5xeR/gDVDnUi)

This error message is difficult to understand. It would be better to show the stacktrace if possible. If that's difficult, we should show the option name.

shulaoda:
Here's my take on this issue:

The root cause isn't an inconsistency between the Rust and JS sides — it's that the user wrote plain JavaScript (no TypeScript), accidentally returned a value that doesn't match the option's expected type, and `bindingify` didn't catch it. The bad value then slipped through to the NAPI boundary, where Rust rejected it with a generic `Value is non of these types `bool`, ...` error that carries no option name and no stack trace.

Fixing this on the Rust side isn't appropriate. Because we run a `bindingify` step in JS, the return type Rust expects is not necessarily the same as the return type the JS-facing API accepts (e.g. Rust sees `bool` after `bindingify` has already normalized a `string | boolean | null`). A Rust-side check would validate the post-`bindingify` shape, which is not the contract we want to report errors against — and it has no access to the JS option name or the user's stack.

The right place for this is `bindingify` itself. For the options whose expected JS-level return type we know, we should add a per-option validation step there. That lets us surface a clear error with the option name, the offending value, and a real JS stack trace pointing back to the user's function — which is exactly the context missing from the current NAPI error.

[...]

PR #9125, opened 2026-04-16
fix: improve error message for invalid return values in function options

shulaoda:
Related to #9005

## Problem

When a user's JS function option (plugin hook, `external`, `treeshake.moduleSideEffects`, etc.) returned a value whose type did not match what Rust expected, the error was unhelpful in two ways:

1. **The message was vague.** Errors read like:
   ```
   UNKNOWN_RETURN_VALUE. Cannot convert unknown to `Option<String>` in ThreadsafeFunction<..., Either<..., UnknownReturnValue>, ...>.
   ```
   The "received" side was always the literal string `"unknown"` — we never inspected the actual JS value — and the "expected" side was a Rust type name (`Option<String>`) that plugin authors writing JavaScript do not recognise.

2. **In some paths the mismatch bypassed our error channel entirely** and surfaced as an uncatchable `napi_fatal_exception`, because NAPI's built-in `UnknownReturnValue` fallback validates as `Unknown` and lets the underlying type check fail further up the stack.

## Fix

Replace NAPI's `UnknownReturnValue` with a custom `InvalidReturnValue` type in `crates/rolldown_binding/src/types/js_callback.rs`:

- `InvalidReturnValue` captures the **actual `ValueType`** of whatever JS returned (`string`, `number`, `object`, `function`, …) by inspecting the napi value in its `FromNapiValue` impl.
- Its `ValidateNapiValue` impl always succeeds, so type mismatches are consistently caught on the Rust side as a clean `napi::Error` instead of becoming a fatal exception.
- `JsCallback<Args, Ret>` is retyped from `Either<Ret, UnknownReturnValue>` to `Either<Ret, InvalidReturnValue>`. When the `Ret` branch fails to validate, we now hit the `Either::B(invalid)` arm with real information.
- The new error message is:
  ```
  The function returned `string`, but expected `number` in `ThreadsafeFunction<(String, bool), ..., ...>`.
  ```
  - Both `received` and `expected` are printed as **JS type names** (via a `js_type_name(ValueType)` helper) so the text reads naturally to JS plugin authors.
  - `fn_type` keeps `pretty_type_name::<Self>()` from the previous implementation so the callback's `Args`/`Ret` signature is still visible for debugging — this is what tells you *which* hook threw.
- `invoke_async` / `await_call` are simplified into plain `async fn`s; the `#[expect(clippy::manual_async_fn)]` workaround is no longer needed.

### DIFF
--- a/crates/rolldown_binding/src/types/js_callback.rs
+++ b/crates/rolldown_binding/src/types/js_callback.rs
@@ -1,12 +1,46 @@
-use std::sync::Arc;
+use std::{ptr, sync::Arc};
 
 use futures::Future;
 use napi::{
-  Either, Status,
-  bindgen_prelude::{FromNapiValue, JsValuesTupleIntoVec, Promise},
-  threadsafe_function::{ThreadsafeFunction, UnknownReturnValue},
+  Either, Status, ValueType,
+  bindgen_prelude::{FromNapiValue, JsValuesTupleIntoVec, Promise, TypeName, ValidateNapiValue},
+  sys,
+  threadsafe_function::ThreadsafeFunction,
 };
-use rolldown_utils::debug::pretty_type_name;
+
+/// Used as the fallback branch in `Either<Ret, InvalidReturnValue>` to catch
+/// type mismatches from JS function options. Always passes NAPI validation so
+/// that when `Ret` validation fails, the error is handled in Rust with a clear
+/// message instead of becoming an uncatchable `napi_fatal_exception`.
+pub struct InvalidReturnValue {
+  pub value_type: ValueType,
+}
+
+impl TypeName for InvalidReturnValue {
+  fn type_name() -> &'static str {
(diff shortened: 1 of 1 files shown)


shulaoda:
* **#9125** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9125?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9125?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9125).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyf0 (approved):

PR #9136, opened 2026-04-16
Allow undefined in sourcesContent type

### LINKED ISSUE #9135: [Bug]: BindingSourceMap is incompatible to ExistingRawSourceMap

If i return source map from the RolldownMagicString["generateMap"];
Type check fails.

```ts
export default {
   transform: {
      order: "post",
      handler(this, code, id) {
        const ms = new RolldownMagicString(code);

        // Do Something with the string

        // Error Here
        const result = { code: ms, map: ms.generateMap({ hires: true }) } satisfies TransformResult;
        return result;
      },
    },
}
```

### What is expected?

tsc does not fail

Image

This because BindingSourceMap has `get sourcesContent(): Array<string | undefined | null>;``

And ExistingRawSourceMap has `sourcesContent?: (string | null)[];`

i would like to avoid using `strictNullChecks: false`

### System Info

```Shell
System:
    OS: macOS 26.4
    CPU: (10) arm64 Apple M2 Pro
    Memory: 77.16 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.11.0 - /Users/jurij.zahn/.local/state/fnm_multishells/35855_1776071072676/bin/node
    npm: 11.6.1 - /Users/jurij.zahn/.local/state/fnm_multishells/35855_1776071072676/bin/npm
    pnpm: 10.30.3 - /Users/jurij.zahn/.local/state/fnm_multishells/35855_1776071072676/bin/pnpm
  Browsers:
    Brave Browser: 147.1.89.132
    Chrome: 147.0.7727.56
    Edge: 147.0.3912.60
    Firefox: 149.0.2
    Safari: 26.4
```

### Any additional comments?

_No response_

jurijzahn8019:
Updated sourcesContent type to allow undefined values.

This makes it compatible to BindingSourceMap

Fixes: #9135

### DIFF
--- a/packages/rolldown/src/types/sourcemap.ts
+++ b/packages/rolldown/src/types/sourcemap.ts
@@ -7,7 +7,7 @@ export interface ExistingRawSourceMap {
   mappings: string;
   names?: string[];
   sources?: (string | null)[];
-  sourcesContent?: (string | null | undefined)[];
+  sourcesContent?: (string | null | undefined)[];
   sourceRoot?: string;
   version?: number; // make it optional to compat { mappings: '' }
   x_google_ignoreList?: number[];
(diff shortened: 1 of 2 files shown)


IWANABETHATGUY (approved):
Thanks

PR #9139, opened 2026-04-16
fix(plugin/copy-module): honor external resolutions from other plugins

TheAlexLichter:
When another plugin resolves a specifier and marks it `external: true`, `CopyModulePlugin::resolve_id` was ignoring the flag and attempting to red the id, producing `Failed to read copy module <id>`.

This surfaces with `rolldown-plugin-dts`, which returns ambient-module glob specifiers such as `typeof import("*.jpg")` as external. If `.jpg` is also configured as `moduleTypes: copy`, the copy plugin  invokes `ctx.resolve`, which runs the dts resolver, gets back `{ id: "*.jpg", external: true }`, and then crashes trying to read `*.jpg` off disk.

Real world example: https://github.com/alveusgg/data/pull/265

### DIFF
--- a/crates/rolldown_plugin_copy_module/src/lib.rs
+++ b/crates/rolldown_plugin_copy_module/src/lib.rs
@@ -74,6 +74,16 @@ impl Plugin for CopyModulePlugin {
       Err(_) => return Ok(None),
     };
 
+    // Honor external resolutions from other plugins. `rolldown-plugin-dts`,
+    // for instance, marks TypeScript ambient-module glob specifiers like
+    // `typeof import("*.jpg")` as `{ id, external: true }` so they pass
+    // through untouched. Without this check the copy plugin would ignore the
+    // `external` flag and try to read `*.jpg` from disk, emitting a
+    // confusing `Failed to read copy module *.jpg` error.
+    if resolved_id.external.is_external() {
+      return Ok(None);
+    }
+
     // Strip query/fragment (e.g. `file.txt?url`) before extension check and file read
     let clean_id = clean_url(resolved_id.id.as_str());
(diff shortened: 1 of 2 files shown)


IWANABETHATGUY:
@shulaoda @hyf0 I am not familiar with this plugin. Could you help review this PR?

crates/rolldown_plugin_copy_module/src/lib.rs
```diff
@@ -74,6 +74,16 @@ impl Plugin for CopyModulePlugin {
       Err(_) => return Ok(None),
     };
 
+    // Honor external resolutions from other plugins. `rolldown-plugin-dts`,
+    // for instance, marks TypeScript ambient-module glob specifiers like
+    // `typeof import("*.jpg")` as `{ id, external: true }` so they pass
+    // through untouched. Without this check the copy plugin would ignore the
+    // `external` flag and try to read `*.jpg` from disk, emitting a
+    // confusing `Failed to read copy module *.jpg` error.
+    if resolved_id.external.is_external() {
```
hyf0, line 83: This is where I wanna introduce `postResolveId`. @sapphi-red cc

hyf0 (approved):
Well done!

crates/rolldown_plugin_copy_module/src/lib.rs
sapphi-red: I don't think this should be the reason to introduce `postResolveId`.

hyf0: I don't think it's the only reason too. It's just one usage I noticed.

Issue #8849, opened 2026-03-22
[Bug]: Dynamic common chunk imports entry module


SebastianStamm:
### Reproduction link or steps

https://repl.rolldown.rs/#eNp9ksFugzAMhl/FyoVWQlSadqLaaY+x7JCCodkSB0HYOiHefU6yUpi6SSBwYjvf/zuTaEQ5CU01Xgo/hH8S5S3ORcVh5WjwYOEJtO1c73dSFIfKWetIiv1R9pLCG9KcwcK4dpf1yD16TS1YpSlLWesMuz9yexSl70ecc6Fqq2nFsMSJIZ0MH8qMCE3vLASIGEqxQdiQZlfQH4J/OOOB90DzdCrzMmbvjKndJxWc0uh2BXxnZ4M+QY2NJnyO2zAnFdm1LGM4wktM5UQ1mvhdCnaTJABN3ehLeJHLjKTI2YqrW1K85nA4gD+zS9rgAAMarDzWcPqKq6fRe0fAD6rqDF6dQl8ucZ3XLFvSvJcUpHaqelctFm+DY31J5GYtyYtcUtTYBSup0sgUJUwzd0qNon0rp5Y4Nfglekp2l/D4APP2jqRRrjrdFv66JXwBYpTdHf8y2EDJ9UZ5HLyYvwE1dg2T

### What is expected?

[...]

PR #7194, opened 2025-11-22
fix: optimize chunk merging for shared entry points

IWANABETHATGUY:
related to #4904

PR #9085, opened 2026-04-13
fix: relax overly conservative side-effect leak check in chunk optimizer

IWANABETHATGUY:
## Summary
- Remove redundant `is_reachable` check in `try_insert_into_existing_chunk` — all dynamic entries in `chunk_idxs` already depend on the pending common chunk by construction, so filtering them via `is_reachable` was a no-op.
- Refine the side-effect leak guard: instead of unconditionally blocking the merge when any dynamic entry depends on the common chunk, only block it when a *different* user-defined entry can also reach those dynamic entries. This allows more merges that are actually safe (the dynamic entry is only reachable from the merge target, so its side effects have already run).
- Remove the now-unused `temp_chunk_idx` parameter from `try_insert_into_existing_chunk`.

## Test plan
- All 1677 integration tests pass.
- Clippy and fmt pass.

PR #9154, opened 2026-04-18
fix: avoid extra shared chunks for statically owned lazy-entry modules

schiller-manuel:
## Summary
Fix two chunk-merging false negatives in shared entry/lazy-entry graphs:
- compute static-entry reachability using only static imports
- avoid splitting out an extra shared chunk when the target entry already statically owns those modules

Related issue:
- https://github.com/TanStack/router/issues/7197

## Background
While investigating [TanStack Router issue `#7197`](https://github.com/TanStack/router/issues/7197), I found a more general Rolldown chunk-merging problem.

The relevant graph shape is:
- a startup entry statically imports an intermediate module
- that intermediate module statically imports some shared module
- that same intermediate module also exposes a lazy import to another chunk
- the lazy chunk imports that same shared module

In this topology, the shared module is already statically owned by the startup-entry side of the graph, but Rolldown can still bail out and emit it as a separate common chunk. That output shape is suboptimal for this case: the shared module still needs to be loaded eagerly, but it should stay merged into the startup entry chunk instead of being split out as its own extra shared chunk.

## Cause
#7194 added `all_dynamic_entries_reachable` as a safety check before merging shared modules into a user-defined entry chunk.
That guard is still correct in general, but it has two false negatives for this class of graphs:
- static-entry reachability was traversing dynamic imports, which over-approximated what a static entry actually reaches
- the merge bailout was too strict for shared modules that are already transitively statically owned by the target entry chunk

At the same time, the fallback cannot be too broad, or it causes real regressions by making dynamic chunks import side-effectful entry chunks.

## Fix
This PR makes two functional changes in the chunk optimizer:
- only traverse static imports when computing static-entry reachability
- allow merging into a user-defined entry when the pending shared modules are already transitively statically owned by that target chunk

The second part is intentionally narrower than a simple "used by target chunk" check. During validation, that broader form caused regressions in cases like:
- #8849
- #9049
- `optimization/chunk_merging/transitive_dynamic_dep_should_not_merge_into_side_effectful_entry`

So the final rule keeps the safety guarantees from #7194:
- dynamic-entry importer checks still have to pass
- side-effect leaks through unrelated dynamic-entry paths are still blocked
- the fallback only applies when the target chunk already statically owns the pending modules through its own static dependency chain

## Test
Added a regression test here:
`crates/rolldown/tests/rolldown/optimization/chunk_merging/startup_entry_shared_module_not_eager_chunked/`
The test is intentionally modeled after the TanStack Router failure:

- `main.js` statically imports `route-graph.js`
- `route-graph.js` statically imports `route-shared.js`
- `route-graph.js` lazily imports `route-component.js`
- `route-component.js` also imports `route-shared.js`

Expected behavior with the fix:
- `route-shared` is merged into `main.js`
- `route-component.js` imports from `main.js`
- no separate `route-shared.js` common chunk is emitted

This regression was validated by temporarily removing the second fix:
- without the fix, Rolldown emits a separate `route-shared.js` chunk imported by both `main.js` and `route-component.js`
- with the fix restored, that extra shared chunk disappears

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -353,7 +353,12 @@ impl GenerateStage<'_> {
           continue;
         };
 
-        for dep_idx in module.import_records.iter().filter_map(|r| r.resolved_module) {
+        for dep_idx in module
+          .import_records
+          .iter()
+          .filter(|rec| rec.kind.is_static())
+          .filter_map(|r| r.resolved_module)
+        {
           // Can't put it at the beginning of the loop,
           if let Some(chunk_idx) = dynamic_entry_modules.get(&dep_idx) {
             ret.entry(entry_chunk_idx).or_default().insert(*chunk_idx);
@@ -620,9 +625,6 @@ impl GenerateStage<'_> {
       // the dynamic entries would not be able to access the shared modules.
       let all_dynamic_entries_reachable =
         dynamic_entry.iter().all(|idx| reached_dynamic_chunk.is_some_and(|set| set.contains(idx)));
-      if !all_dynamic_entries_reachable {
-        return None;
-      }
       // If any dynamic entry chunk depends on the pending common chunk,
       // merging into the user-defined entry would make the dynamic entry
       // import the entry chunk. This is only safe if the entry chunk
@@ -649,6 +651,39 @@ impl GenerateStage<'_> {
           return None;
         }
       }
+      let candidate_modules =
+        chunk.modules.iter().copied().chain(info.modules.iter().copied()).collect::<FxHashSet<_>>();
+      let mut statically_owned_modules = chunk.modules.iter().copied().collect::<FxHashSet<_>>();
+      let mut changed = true;
+      while changed {
+        changed = false;
+        for &module_idx in &info.modules {
+          if statically_owned_modules.contains(&module_idx) {
+            continue;
+          }
+          let Some(module) = module_table[module_idx].as_normal() else {
+            continue;
+          };
+
+          let is_statically_owned = module.importers_idx.iter().any(|importer_idx| {
+            candidate_modules.contains(importer_idx)
+              && statically_owned_modules.contains(importer_idx)
+              && module_table[*importer_idx].as_normal().is_some_and(|importer| {
+                importer
+                  .import_records
+                  .iter()
+                  .any(|rec| rec.kind.is_static() && rec.resolved_module == Some(module_idx))
+              })
+          });
+
+          if is_statically_owned {
+            statically_owned_modules.insert(module_idx);
+            changed = true;
+          }
+        }
+      }
+      let modules_are_statically_used_by_target_chunk =
+        info.modules.iter().all(|module_idx| statically_owned_modules.contains(module_idx));
       let modules_set = chunk
         .modules
         .iter()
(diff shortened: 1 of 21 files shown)


schiller-manuel:
Closing - bisecting the TanStack Router repro between v1.0.0-rc.15 (157 root preloads, bug present) and v1.0.0-rc.16 (6 root preloads, bug fixed) lands on #9085 ("relax overly conservative side-effect leak check in chunk optimizer") as the fix. That PR addresses the same root cause from a different angle and has already shipped in rc.16, so this PR is no longer needed.

crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
```diff
@@ -353,7 +353,12 @@ impl GenerateStage<'_> {
           continue;
         };
 
-        for dep_idx in module.import_records.iter().filter_map(|r| r.resolved_module) {
+        for dep_idx in module
+          .import_records
+          .iter()
+          .filter(|rec| rec.kind.is_static())
```
hyf0, line 359: @IWANABETHATGUY Though this PR is closed, could you still check if the original code of this line is expected?

IWANABETHATGUY: I took a closer look — applying only `rec.kind.is_static()` on this BFS alone is not correct, and the `4895` snapshot churn is the evidence.

**Why the filter alone breaks things**

The BFS populates `ret[user_entry]` with a dynamic entry chunk only when the dynamic entry module is visited. In `4895`, `main.js` reaches `dynamic.js` only through `import('./dynamic.js')` — a dynamic import record. Filter that out and `dynamic.js` is never visited, so `dynamic_entry_modules.get(&dep_idx)` never fires and `ret[main_chunk]` ends up empty.

Downstream, in `try_insert_into_existing_chunk`:

```rust
let all_dynamic_entries_reachable =
    dynamic_entry.iter().all(|idx| reached_dynamic_chunk.is_some_and(|set| set.contains(idx)));
if !all_dynamic_entries_reachable { return None; }
```

`reached_dynamic_chunk` becomes empty for `main.js`, the guard rejects the merge, and `lib.js` can no longer fold into `main.js` — it gets emitted as its own common chunk. That's the `lib.js` / `lib2.js` / `lib3.js` / `lib4.js` split you'll see in the snapshot:

~~~diff
 ## allow-extension.js

 ```js
-//#region allow-extension/lib.js
-const shared = "shared";
-//#endregion
+import { t as shared } from "./lib.js";
 //#region allow-extension/main.js
 console.log(shared);
 import("./dynamic.js");
 const unused = 42;
 //#endregion
-export { shared as t, unused };
+export { unused };
 ```

 ## dynamic.js

 ```js
-import { t as shared } from "./allow-extension.js";
+import { t as shared } from "./lib.js";
 //#region allow-extension/dynamic.js
 console.log(shared);
 //#endregion
 ```

 ## lib2.js

+```js
+//#region false/lib.js
+const shared = "shared";
+//#endregion
+export { shared as t };
+```
+
+## lib3.js
+
+```js
+//#region not-specified/lib.js
+const shared = "shared";
+//#endregion
+export { shared as t };
+```
+
+## lib4.js
+
 ```js
 //#region strict/lib.js
 const shared = "shared";
~~~

(same pattern for `false` / `not-specified` / `strict`.)

**Why PR #9154 doesn't regress `4895`**

The PR doesn't rely on the filter alone — it also adds a fallback in `try_insert_into_existing_chunk`:

```rust
all_dynamic_entries_reachable
  || (modules_are_statically_used_by_target_chunk && all_dynamic_entries_importers_valid)
```

`modules_are_statically_used_by_target_chunk` recovers `4895` via a different path — `lib.js` is transitively, statically owned by `main.js`, so `main.js` is still chosen as the merge target even though the reachability map is now empty.

**Summary**

The two changes are load-bearing together: the `is_static()` filter narrows the reachability set to match its name, and the `modules_are_statically_used_by_target_chunk` fallback covers the merges the narrowing would otherwise drop. Applying only the filter is semantically too strict, which is exactly what the `4895` snapshot diff shows.

PR #9159, opened 2026-04-19
docs: add feedback form

TheAlexLichter:
This PR adds the newly created `vitepress-plugin-feedback-tracker` plugin that I was working on and off over the past weeks.

It allows users to send feedback via the sidebar or on the bottom of the screen.

image

image

**TODO:**

- [x] Test on the preview deployment
- [] Test when live


sapphi-red:
Nice! Here are the things I noticed:

- I cannot change :-1: to :+1: if I missclicked.
- When I click :+1: or :-1: on the side bar, the page is scrolled, but it's a bit laggy. It takes a bit until the scroll happens. It is a bit unclear why the scroll happened.
- After selecting the reason on the sidebar, the bottom one still shows the reason selection. (i.e. the state is not synced)
- In the source code, I see that `Arrow*` keys are handled. But I cannot select other categories by pressing those buttons.

mdong1909:
@TheAlexLichter - to make things simpler, how about we only have the feedback form at the bottom of the page? Remove the sidebar form. 

Was looking at other docs sites like nuxt and almost everyone only has one. Either bottom or sidebar.

TheAlexLichter:
@sapphi-red Good call on these!

> I cannot change 👎 to 👍 if I missclicked.

> When I click 👍 or 👎 on the side bar, the page is scrolled, but it's a bit laggy. It takes a bit until the scroll happens. It is a bit unclear why the scroll happened.
 
> In the source code, I see that Arrow* keys are handled. But I cannot select other categories by pressing those buttons.

Will fix these!

> After selecting the reason on the sidebar, the bottom one still shows the reason selection. (i.e. the state is not synced)

That was originally on purpose but I see how this is confusing

@mdong1909 

> to make things simpler, how about we only have the feedback form at the bottom of the page? Remove the sidebar form. Was looking at other docs sites like nuxt and almost everyone only has one. Either bottom or sidebar.

What is the main concern? I think we can do that so we don't overload users but only for "short pages" so longer pages like https://rolldown.rs/reference/OutputOptions.cleanDir still have both 🤔

mdong1909:
1. Maintenance. We want this plugin/feature to be lightweight. Even if it won't be hard to fix issues in the future, it's still a context switch. Err towards simplicity. 
2. Ux. Half of sapphis comments are related to sidebar-bottom interaction. As a user, I'm thrown when inputting sidebar jumps me to the bottom. Real world experience is I'll have to then scroll back up to the part I was just at. Syncing state when a user can see both forms is another ux exercise. 

The question is "what benefit do we get from including both the sidebar & bottom?" 

It's possible we get a few more responses but I'd argue it doesn't justify the added complexity and maintenance surface. Looking ahead, I don't want you to spend much, if any, time maintaining this plugin. There's more impactful projects to work on. For an initial version, the simpler the better.

TheAlexLichter:
Okay, then lets keep it only at the bottom. That way we don't have weird scrolling and can still show the email + option field.

TheAlexLichter:
Updated this PR:

* Addressed all the feedback from before.
* Switched the docs integration to vitepress-plugin-feedback-tracker@0.2.0-alpha.0
* Removed the sidebar feedback widget so feedback only renders at the bottom of doc pages

sapphi-red:
[repro-change.webm](https://github.com/user-attachments/assets/be219ad7-7adc-4f5c-958b-610a9ffcf25b)

It seems it breaks when changing the "Yes" / "No".

TheAlexLichter:
Fixed in the latest version (but needs to wait for the min release age guard, adding a workaround)

h-a-n-a (approved):

sapphi-red:
It seems it's still broken for me 🤔

TheAlexLichter:
> It seems it's still broken for me 🤔

Just double checked. Yes -> Change (And same for no) seems fine now for me. The "validation changed" as in the video didn't happen.

Maybe there was something cached?

sapphi-red:
Hmm, it seems it's working now. I did reload with cache clear, but maybe I somehow got the old one 🤔
Sorry for the false flag.

sapphi-red (approved):

PR #9162, opened 2026-04-20
ci: verify wasm32-wasi binding deps match @rolldown/browser before publish

### LINKED ISSUE #8974: `@rolldown/browser` must have `@emnapi/runtime` and `@emnapi/core` as a dep
`@emnapi/runtime` and `@emnapi/core` is a peer dep of `@napi-rs/wasm-runtime` and it must be a dep.
The version must be aligned with the version of `emnapi` used by `@napi-rs/cli`.

The reason for it is https://github.com/toyobayashi/emnapi/issues/202

refs https://github.com/rolldown/rolldown/pull/8943

shulaoda:
closes #8974

## Summary

- Add `scripts/misc/check-wasi-binding-deps.mjs` to verify that the runtime dependencies (`@napi-rs/wasm-runtime`, `@emnapi/core`, `@emnapi/runtime`) declared in the generated `@rolldown/binding-wasm32-wasi/package.json` match the specifiers used by `@rolldown/browser` (resolved through the pnpm catalog when applicable).
- Wire the check into both publish workflows (`publish-to-npm.yml` and `publish-to-pkg.pr.new.yml`) right after `napi create-npm-dirs` runs, so a mismatch fails the job before anything is published.

## Why

`napi create-npm-dirs` pins the wasm32-wasi binding's runtime dependencies from its own sources — `@napi-rs/wasm-runtime` is fetched as the latest published version on npm, while `@emnapi/core` / `@emnapi/runtime` follow the version of `emnapi` resolved locally. These can silently drift from the specifiers used by `@rolldown/browser`, which bundles the binding's glue code and must load the same runtime versions at execution time.

The generated binding is the source of truth: its glue code was built against those exact runtime versions, so `@rolldown/browser` must stay aligned. When this check fails, the fix is to bump the corresponding entries in `pnpm-workspace.yaml` (for deps referenced via `catalog:`) or in `packages/browser/package.json` (for deps pinned directly) to match what the binding declares.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -18,7 +18,7 @@ catalog:
   '@babel/preset-env': ^7.24.7
   '@babel/preset-typescript': ^7.24.7
   '@napi-rs/cli': ^3.6.1
-  '@napi-rs/wasm-runtime': ^1.1.3
+  '@napi-rs/wasm-runtime': ^1.1.4
   '@oxc-node/cli': ^0.1.0
   '@oxc-node/core': ^0.1.0
   '@oxc-project/runtime': '=0.126.0'
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -17,8 +17,6 @@ catalog:
   '@babel/core': ^7.24.7
   '@babel/preset-env': ^7.24.7
   '@babel/preset-typescript': ^7.24.7
-  '@emnapi/core': ^1.10.0
-  '@emnapi/runtime': ^1.10.0
   '@napi-rs/cli': ^3.6.1
   '@napi-rs/wasm-runtime': ^1.1.3
   '@oxc-node/cli': ^0.1.0
@@ -51,7 +49,6 @@ catalog:
   consola: ^3.4.2
   dedent: ^1.5.3
   diff: ^9.0.0
-  emnapi: ^1.10.0
   esbuild: ^0.28.0
   estree-toolkit: ^1.7.8
   execa: ^9.2.0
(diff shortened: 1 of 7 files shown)


pnpm-workspace.yaml
```diff
@@ -17,10 +17,8 @@ catalog:
   '@babel/core': ^7.24.7
   '@babel/preset-env': ^7.24.7
   '@babel/preset-typescript': ^7.24.7
-  '@emnapi/core': ^1.10.0
-  '@emnapi/runtime': ^1.10.0
```
sapphi-red, line 21: Why were these removed from the catalog?

shulaoda: Since it is only used in `@rolldown/browser`, and the versions of `@emnapi/core` and `@emnapi/runtime` generated by napi-rs in [@rolldown/binding-wasm32-wasi](https://www.npmjs.com/package/@rolldown/binding-wasm32-wasi?activeTab=code) are fixed, I declared it directly in `@rolldown/browser`.

scripts/misc/check-wasi-binding-deps.mjs
```diff
@@ -0,0 +1,98 @@
...
+const TRACKED = ['@napi-rs/wasm-runtime', '@emnapi/core', '@emnapi/runtime'];
+const BINDING_PKG = path.join(REPO_ROOT, 'packages/rolldown/npm/wasm32-wasi/package.json');
+const BROWSER_PKG = path.join(REPO_ROOT, 'packages/browser/package.json');
+const WORKSPACE_YAML = path.join(REPO_ROOT, 'pnpm-workspace.yaml');
+
+function readJson(p) {
+  return JSON.parse(fs.readFileSync(p, 'utf8'));
+}
+
+function readCatalog() {
```
sapphi-red: Can we use `pnpm --dir ./packages/browser ls @emnapi/core --json` instead of parsing the catalog manually?

sapphi-red (approved):

Issue #9114, opened 2026-04-15
Replace all `tokio::sync::Mutex` with `std::sync::Mutex` where appropriate


sapphi-red:
According to the docs of `tokio::sync::Mutex`, we should be using `std::sync::Mutex` instead in most places:

> Contrary to popular belief, it is ok and often preferred to use the ordinary [`Mutex`](https://doc.rust-lang.org/nightly/std/sync/poison/mutex/struct.Mutex.html) from the standard library in asynchronous code.
>
> The feature that the async mutex offers over the blocking mutex is the ability to keep it locked across an `.await` point. This makes the async mutex more expensive than the blocking mutex, so the blocking mutex should be preferred in the cases where it can be used. The primary use case for the async mutex is to provide shared mutable access to IO resources such as a database connection. If the value behind the mutex is just data, it’s usually appropriate to use a blocking mutex such as the one in the standard library or [`parking_lot`](https://docs.rs/parking_lot).
> https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

There also a warning in that section:

> Note that, although the compiler will not prevent the std `Mutex` from holding its guard across `.await` points in situations where the task is not movable between threads, this virtually never leads to correct concurrent code in practice as it can easily lead to deadlocks.
> https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

But these cases can be caught by clippy: https://rust-lang.github.io/rust-clippy/master/index.html#await_holding_lock

related: https://github.com/rolldown/rolldown/pull/9031

PR #9176, opened 2026-04-21
refactor: replace tokio::sync::Mutex with std::sync::Mutex for non-IO data

shulaoda:
## Summary

- Replace `tokio::sync::Mutex` with `std::sync::Mutex` for two places that only guard plain data and never hold the lock across an `.await` point:
  - `PluginDriver.tx` / `NativePluginContextImpl.tx` — `Arc<Mutex<Option<UnboundedSender<ModuleLoaderMsg>>>>`
  - `BundleCoordinator.watcher` — `Mutex<DynFsWatcher>`
- De-async `PluginDriver::set_context_load_modules_tx` as a consequence, and drop the redundant `.await` at its two call sites in `ScanStage`.

## Why

Per tokio's own guidance, the async mutex should be reserved for protecting IO resources where the lock genuinely needs to be held across `.await`. For plain data the blocking `std::sync::Mutex` is both cheaper and less prone to latent deadlocks when used through sync NAPI bindings. This continues the migration started in #9031 (which collapsed the `FileEmitter.tx` path to sync for the `emit_chunk` deadlock fix).

Scope of this change:

- `PluginDriver.tx` is only ever acquired to **clone an `Option<UnboundedSender>`** and to swap a new sender in at scan boundaries. The existing code in `NativePluginContextImpl::load` already drops the guard before awaiting `sender.send(...)`; the only reason it used `tokio::sync::Mutex` was incidental. Switching to `std::sync::Mutex` makes that invariant syntactically enforced (no `.await` possible while the guard is live) and lets `set_context_load_modules_tx` shed its unnecessary `async` marker.
- `BundleCoordinator.watcher` is used exclusively inside `update_watch_paths`, where the guard is held only over synchronous `paths_mut.add(...)` / `paths_mut.commit()` calls. No `.await` exists in that critical section, so async mutex is pure overhead here.

No behavioral change for plugin authors or bundler consumers, the JS APIs and message flows are unchanged.

## Refs

- Closes part of #9114 (tokio → std mutex migration)
- Related: #9031 (sync-NAPI deadlock fix for `emit_chunk`; established the `std::sync::Mutex` pattern for the tx slot on `FileEmitter`)

### DIFF
--- a/crates/rolldown_dev/src/bundle_coordinator.rs
+++ b/crates/rolldown_dev/src/bundle_coordinator.rs
@@ -1,9 +1,10 @@
 use std::{
   collections::VecDeque,
   path::PathBuf,
-  sync::{Arc, atomic::AtomicU32},
+  sync::{Arc, Mutex as StdMutex, atomic::AtomicU32},
 };
 
+use anyhow::Context;
 use arcstr::ArcStr;
 use futures::FutureExt;
 use notify::EventKind;
@@ -35,7 +36,7 @@ pub struct BundleCoordinator {
   ctx: SharedDevContext,
   next_hmr_patch_id: Arc<AtomicU32>,
   rx: CoordinatorReceiver,
-  watcher: Mutex<DynFsWatcher>,
+  watcher: StdMutex<DynFsWatcher>,
   watched_files: FxDashSet<ArcStr>,
   /// Tracks the state of the initial build
   state: CoordinatorState,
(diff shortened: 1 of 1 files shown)


shulaoda:
* **#9176** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9176?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9176?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9176).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

h-a-n-a (approved):

PR #9185, opened 2026-04-21
fix: preserve `ignored` flag through plugin `this.resolve()` round-trip

### LINKED ISSUE #9107: [Bug]: Returning this.resolve() from resolveId misresolves browser: false imports

shulaoda:
## Summary

When a plugin's `resolveId` hook calls `this.resolve(..., { skipSelf: true })` and returns the result directly, the `ignored` flag (set when `browser: false` maps a module to nothing) was silently dropped. The bundler then tried to load the raw directory path returned by the resolver, failing with `Is a directory`.

**Root cause**: `HookResolveIdOutput` had no `ignored` field, so `ResolvedId.ignored = true` was lost when passing through the plugin/NAPI boundary and back.

Closes #9107

## Changes

- **`HookResolveIdOutput`** — added `ignored: bool`; `from_resolved_id()` now copies it
- **`resolve_id_with_plugins.rs`** — propagates `ignored` from plugin output into the final `ResolvedId` (both regular and dynamic import paths)
- **NAPI bindings** — `BindingPluginContextResolvedId` (return of `this.resolve()`) and `BindingHookResolveIdOutput` (return of `resolveId` hook) both carry `ignored`
- **TypeScript** — `ResolvedId`, `PartialResolvedId`, `bindingify-build-hooks.ts`, and `plugin-context.ts` all thread `ignored` through; `binding.d.cts` regenerated

## Test

Added regression fixture under `plugin/context/resolve-browser-false`: reproduces the exact scenario — `browser: { "buffer": false }` in a local `package.json`, a plugin that forwards the `this.resolve()` result, and `platform: "browser"`.

The test fixture (plugin shape and the `code === ''` empty-output assertion) is adapted from @AlexanderKolberg's #9108, which proposed an alternative fix for the same issue. This PR takes a different fix approach — threading `ignored` through the plugin/NAPI boundary instead of re-querying the internal resolver to recover the bit — but reuses the clean regression case from #9108, so credit for the test case goes to them.

---

Co-authored-by: Alexander Kolberg <74478295+AlexanderKolberg@users.noreply.github.com>

### DIFF
--- a/packages/rolldown/src/plugin/index.ts
+++ b/packages/rolldown/src/plugin/index.ts
@@ -71,6 +71,8 @@ export interface ModuleOptions {
 export interface ResolvedId extends ModuleOptions {
   external: boolean | 'absolute';
   id: string;
+  /** @internal When true, the module is ignored (e.g. browser: false mapping). */
+  ignored?: boolean;
 }
 
 // Separate interface to dedupe the JSDoc comment
@@ -111,6 +113,8 @@ export interface PartialResolvedId
    */
   external?: boolean | 'absolute' | 'relative';
   id: string;
+  /** @internal When true, the module is ignored (e.g. browser: false mapping). */
+  ignored?: boolean;
 }
 
 /** @category Plugin APIs */
--- a/packages/rolldown/tests/fixtures/plugin/context/resolve-browser-false/_config.ts
+++ b/packages/rolldown/tests/fixtures/plugin/context/resolve-browser-false/_config.ts
@@ -1,0 +1,31 @@
+import { defineTest } from 'rolldown-tests';
+import { expect } from 'vitest';
+
+// Regression test: Returning the result of `this.resolve()` from `resolveId` should
+// preserve `browser: false` handling (the module should be ignored/empty).
+export default defineTest({
+  config: {
+    input: './main.js',
+    platform: 'browser',
+    plugins: [
+      {
+        name: 'probe',
+        async resolveId(source, importer, options) {
+          if (options?.isEntry || !importer) {
+            return;
+          }
+          const resolved = await this.resolve(source, importer, {
+            ...options,
+            skipSelf: true,
+          });
+          return resolved;
+        },
+      },
+    ],
+  },
+  afterTest(output) {
+    expect(output.output).toHaveLength(1);
+    expect(output.output[0].type).toBe('chunk');
+    expect(output.output[0].code).toBe('');
+  },
+});
--- a/packages/rolldown/tests/fixtures/plugin/context/resolve-browser-false/_config.ts
+++ b/packages/rolldown/tests/fixtures/plugin/context/resolve-browser-false/_config.ts
@@ -1,8 +1,8 @@
 import { defineTest } from 'rolldown-tests';
-import { expect } from 'vitest';
+import { expect, vi } from 'vitest';
 
-// Regression test: Returning the result of `this.resolve()` from `resolveId` should
-// preserve `browser: false` handling (the module should be ignored/empty).
+const onResolved = vi.fn();
+
 export default defineTest({
   config: {
     input: './main.js',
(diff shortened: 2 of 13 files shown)


packages/rolldown/tests/fixtures/plugin/context/resolve-browser-false/_config.ts
```diff
@@ -0,0 +1,31 @@
...
+    input: './main.js',
+    platform: 'browser',
+    plugins: [
+      {
+        name: 'probe',
+        async resolveId(source, importer, options) {
+          if (options?.isEntry || !importer) {
+            return;
+          }
+          const resolved = await this.resolve(source, importer, {
```
hyf0, line 17: Should we do a assertion here? `resolved.ignored` must be true.

packages/rolldown/tests/fixtures/plugin/context/resolve-browser-false/_config.ts
```diff
@@ -0,0 +1,31 @@
...
+          });
+          return resolved;
+        },
+      },
+    ],
+  },
+  afterTest(output) {
+    expect(output.output).toHaveLength(1);
+    expect(output.output[0].type).toBe('chunk');
+    expect(output.output[0].code).toBe('');
```
hyf0: I think these 3 are meaningless.

 I would suggest having a mock function inside of `resolveId` and assert it must be called 1 time.

hyf0 (approved):

packages/rolldown/src/plugin/index.ts
```diff
@@ -71,6 +71,8 @@ export interface ModuleOptions {
 export interface ResolvedId extends ModuleOptions {
   external: boolean | 'absolute';
   id: string;
+  /** @internal When true, the module is ignored (e.g. browser: false mapping). */
+  ignored?: boolean;
```
shulaoda, line 75: Do you think it is reasonable for us to expose this to users here? @sapphi-red

sapphi-red: Why is this a separate flag from `id`?

shulaoda: This happens because internally we use an `ignore` flag to represent modules with `browser: false`, see the implementation here:
- https://github.com/rolldown/rolldown/blob/main/crates/rolldown_common/src/types/resolved_id.rs#L40-L41  
- https://github.com/rolldown/rolldown/blob/main/crates/rolldown/src/utils/load_source.rs#L37-L39  

This leads to an issue: when a user calls `this.resolve` inside `resolveId` and directly returns the result, the `ignore` field gets lost, which causes the problem described in #9107.

On the other hand, `@rollup/plugin-node-resolve` handles `browser: false` by mapping it to a virtual module id:
- https://github.com/rollup/plugins/blob/master/packages/node-resolve/src/index.js#L119-L121  

I’ve also aligned with this approach in https://github.com/rolldown/rolldown/pull/9192.

Which of these approaches do you think is more reasonable and correct?

sapphi-red: I think using a special id is better because the user won't have to be aware of this special flag.

PR #9188, opened 2026-04-21
ci: remove pull_request_target workflows

Boshen:
## Summary
- Remove the stale PR draft conversion automation.
- Switch PR label/title workflows from `pull_request_target` to `pull_request`.
- Scope issue and PR label jobs to their intended events.

## Why
`pull_request_target` runs in the context of the base repository and can receive write-capable tokens even for PR-related events. That makes it worth avoiding unless a workflow genuinely needs privileged access for fork PRs.

After reviewing the jobs that used it:
- PR title linting only validates metadata, so it can run on `pull_request`.
- `needs-reproduction` is issue-label automation and does not need any PR trigger.
- documentation issue creation only needs to run for same-repo PRs, so it can run on `pull_request` with a same-repo guard.
- stale PR draft conversion was the only job that still required privileged PR mutation for fork PRs, and we are removing that automation instead of keeping `pull_request_target`.

### DIFF
--- a/.github/workflows/automation-labeled.yml
+++ b/.github/workflows/automation-labeled.yml
@@ -3,7 +3,7 @@ name: 'Automation Labeled'
 permissions: {}
 
 on:
-  pull_request_target: # zizmor: ignore[dangerous-triggers]
+  pull_request:
     types: [labeled]
   issues:
     types: [labeled]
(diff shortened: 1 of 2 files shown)


.github/workflows/lint-pr-title.yml
```diff
@@ -3,7 +3,7 @@ name: Lint Pr Title
 permissions: {}
 
 on:
-  pull_request_target: # zizmor: ignore[dangerous-triggers]
```
sapphi-red, line 6: I think this one needs to be `pull_request_target`. Otherwise it won't work with PRs from forks.
https://github.com/amannn/action-semantic-pull-request#event-triggers:~:text=pull_request_target%3A%20This%20allows,the%20main%20branch.

.github/workflows/automation-labeled.yml
```diff
@@ -67,7 +53,7 @@ jobs:
           edit-mode: replace
 
   create_issue_if_doc_labeled_on_pr:
-    if: github.event_name == 'pull_request_target' && github.event.label.name == 'documentation'
```
sapphi-red, line 70: I guess this won't work for PRs from forks for the same reason above.

sapphi-red: > documentation issue creation only needs to run for same-repo PRs, so it can run on `pull_request` with a same-repo guard.

Ah, it is ok if this is the case. I personally don't use this label trigger

PR #9192, opened 2026-04-22
fix: use sentinel id for `browser: false` ignored modules

### LINKED ISSUE #9107: [Bug]: Returning this.resolve() from resolveId misresolves browser: false imports

shulaoda:
closes #9107

### DIFF
--- a/crates/rolldown_plugin/src/utils/resolve_id_with_plugins.rs
+++ b/crates/rolldown_plugin/src/utils/resolve_id_with_plugins.rs
@@ -161,9 +161,7 @@ fn resolve_id<Fs: FileSystem>(
         ..Default::default()
       }),
       ResolveError::Ignored(p) => Ok(ResolvedId {
-        //(hyf0) TODO: This `p` doesn't seem to contains `query` or `fragment` of the input. We need to make sure this is ok
-        id: ModuleId::new(p.to_str().expect("Should be valid utf8")),
-        ignored: true,
+        id: ModuleId::new_empty(p.to_str().expect("Should be valid utf8")),
         ..Default::default()
       }),
       _ => Err(err),
--- a/crates/rolldown/src/utils/load_source.rs
+++ b/crates/rolldown/src/utils/load_source.rs
@@ -21,25 +21,20 @@ pub async fn load_source<Fs: FileSystem + 'static>(
   is_read_from_disk: &mut bool,
   module_idx: ModuleIdx,
 ) -> anyhow::Result<(StrOrBytes, ModuleType)> {
-  let (maybe_source, mut maybe_module_type) = match plugin_driver
-    .load(&HookLoadArgs { id: &resolved_id.id, module_idx, asserted_module_type })
-    .await?
-  {
-    Some(load_hook_output) => {
-      sourcemap_chain.extend(load_hook_output.map.map(SourcemapChainElement::Load));
-      if let Some(v) = load_hook_output.side_effects {
-        *side_effects = Some(v);
-      }
-
-      (Some(load_hook_output.code.to_string()), load_hook_output.module_type)
-    }
-    _ => {
-      if resolved_id.ignored {
-        (Some(String::new()), Some(ModuleType::Empty))
-      } else {
-        (None, None)
-      }
-    }
+  let (maybe_source, mut maybe_module_type) = if resolved_id.id.is_empty_module() {
+    (Some(String::new()), Some(ModuleType::Empty))
+  } else {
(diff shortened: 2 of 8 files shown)


crates/rolldown/tests/esbuild/packagejson/package_json_browser_map_module_disabled/artifacts.snap
```diff
@@ -8,12 +8,12 @@ source: crates/rolldown_testing/src/integration_test.rs
 ```js
 // HIDDEN [\0rolldown/runtime.js]
-//#region (ignored) node_modules/demo-pkg
-var require_demo_pkg$1 = /* @__PURE__ */ __commonJSMin((() => {}));
+//#region \0rolldown/empty.js
```
hyfdev: This might cause an extra chunk chunk when multiple entries exist.

crates/rolldown_plugin/src/utils/resolve_id_with_plugins.rs
```diff
@@ -160,12 +162,7 @@ fn resolve_id<Fs: FileSystem>(
         external: true.into(),
         ..Default::default()
       }),
-      ResolveError::Ignored(p) => Ok(ResolvedId {
-        //(hyf0) TODO: This `p` doesn't seem to contains `query` or `fragment` of the input. We need to make sure this is ok
-        id: ModuleId::new(p.to_str().expect("Should be valid utf8")),
-        ignored: true,
-        ..Default::default()
-      }),
+      ResolveError::Ignored(_) => Ok(ResolvedId { id: EMPTY_MODULE_ID, ..Default::default() }),
```
shulaoda: Here are several possible approaches to handle the debug ID issue:

One option is to introduce an additional field such as `ignore: Option<ModuleId>`. However, this approach may still run into problems where the object returned by `resolveId` is missing certain fields.

Another option is to concatenate `EMPTY_MODULE_ID` with the ignored module ID. The downside is that, similar to the current behavior, this would still result in different objects being generated for modules marked as `browser: false`.

A third approach is to store the ignored module IDs in a global context. This relies on assumptions about global state and may introduce uncertainty or unintended side effects.

crates/rolldown/tests/esbuild/packagejson/package_json_browser_map_module_disabled/artifacts.snap
shulaoda: @sapphi-red Do you think we should use a unified `EMPTY_MODULE_ID`, or concatenate `EMPTY_MODULE_ID` with the ignored id?

sapphi-red: I think we should concatenate with the ignored id.

sapphi-red (approved):

hyf0 (approved):
Why there's no snapshot change?

I assume the comment should print the id directly

shulaoda:
> Why there's no snapshot change?
> 
> I assume the comment should print the id directly

I think it may because we construct something like `Path::new("\0rolldown/empty?C:\\a\\b\\c.js")` and then use `file_stem()` to get the file name `c`, which keeps the behavior somewhat consistent with the original. Do you think we should adjust this?

PR #9194, opened 2026-04-22
chore: upgrade oxc to 0.127.0

### LINKED ISSUE #8248: [Bug]: `/*@vite-ignore*/` is incorrectly stripped

https://repl.rolldown.rs/#eNqFUs1OhDAQfpWxF7q7LGg8SbLqyZMHY2K8cGnKsNSUlpRhdxPCu9suIrjRyIG233w/zXR6VrKsZ43Y401Cbdgbls3nmEl/xFNjHUHZGUnKGiit5Svoc5cbAGlNSxAUT0rji6AKdhBFY9Ehdc6AqoMBT9ePB0W4VXtjHa7TH6pVIgXJinNcwe5+ch/9rcYEnbPOF5c4Qee0jzN4hLfXZ770u4ohqoiaLE1FIm0dXSib800nDIB7qySARtQ+zRTtu6KKR6kyBZ6SupgNwvcQomd+q5VEfh3D9m4FG4gCf0nP/qLfnumLENjMuqBpUThZXaKVaKsJ+bfJvrkjdzivQ/j5p0WWketwiJmzWhf2aHyfTKn2ycf3JPxSGWdibOKI+hfoh9zk5mtQCixFp6eqzxp8xsGLtCBsiQ2fOxzLyg==

### What is expected?

`/*@vite-ignore*/` should be preserved in the second dynamic import too.

`/*@vite-ignore*/` is being stripped.

### System Info

Running in web REPL.

### Any additional comments?

Working fine till beta.35.

### LINKED ISSUE #9121: [Bug]: Static property initialization with computed Symbol keys fails when class has a constructor

You may see the repository with reproduction at:

https://github.com/volkanceylan/vite8-static-init-issue

I changed the sample source so that only rolldown is used.

### What is expected?

When bundling TypeScript classes with static properties that use computed keys (e.g., static [Symbol.toPrimitive] = "test"), rolldown incorrectly omits the initialization code for the computed key variable if the class also has a constructor. This results in the static property not being set at runtime.

### Steps to Reproduce
Create a TypeScript file with a class that has a static property using a computed Symbol key and a constructor:

#### with-constuctor.ts:
```ts
export class SampleClass {
    static [Symbol.toPrimitive] = "test"

    constructor() {
    }
}

console.log("SampleClass[Symbol.toPrimitive]:", SampleClass[Symbol.toPrimitive]);
```

Bundle the code using rolldown and the output is like following:

#### dist/with-constructor.js
```js
//#region with-constructor.ts
let _Symbol$toPrimitive;
var SampleClass = class {
	static {
		this[_Symbol$toPrimitive] = "test";
	}
	constructor() {}
};
console.log("SampleClass[Symbol.toPrimitive]:", SampleClass[Symbol.toPrimitive]);
//#endregion
export { SampleClass };
```

Note that `let _Symbol$toPrimitive` is not initialized anywhere.

If you simply remove the constructor:

#### without-constuctor.ts:
```ts
export class SampleClass {
    static [Symbol.toPrimitive] = "test"
}

console.log("SampleClass[Symbol.toPrimitive]:", SampleClass[Symbol.toPrimitive]);
```

The output is now correct:

```js
//#region without-constructor.ts
let _Symbol$toPrimitive;
var SampleClass = class {
	static {
		_Symbol$toPrimitive = Symbol.toPrimitive;
	}
	static {
		this[_Symbol$toPrimitive] = "test";
	}
};
console.log("SampleClass[Symbol.toPrimitive]:", SampleClass[Symbol.toPrimitive]);
//#endregion
export { SampleClass };
```

Note the first static initialization block which was missing from the constructor case.

Static initialization block that sets `_Symbol$toPrimitive`  variable is missing if the class has a constructor so the property is `undefined`.

### System Info

```Shell
rolldown 1.0.0-rc.15
node 24.10.0
nom 11.1.1
```

### Any additional comments?

I noticed this bug while trying vite8 with vitest, and thought the problem is with vitest or vite8 ssr, and opened an issue there, but seems like it is at rolldown level, so I closed that one and reopening here.

Dunqing:
- Closes https://github.com/rolldown/rolldown/issues/8248
- Closes https://github.com/rolldown/rolldown/issues/9121

## Summary

Upgrade OXC from `0.126.0` to `0.127.0` across the Rust workspace and JS workspace packages.

This updates the pinned OXC crates and packages, refreshes the embedded runtime helper version, and records the snapshot / fixture changes caused by the new OXC output.

## What Changed

- bump Rust OXC workspace deps in `Cargo.toml` to `0.127.0`
- bump JS OXC packages in `pnpm-workspace.yaml` to `0.127.0`
- refresh `Cargo.lock` and `pnpm-lock.yaml`
- regenerate `crates/rolldown_plugin_oxc_runtime/src/generated/embedded_helpers.rs`
- update affected snapshots and sourcemap expectations

## Why

Rolldown tracks OXC closely, and this brings the repo onto the latest `crates_v0.127.0` release.

The snapshot churn here is expected from upstream minifier / DCE behavior changes rather than local Rolldown logic changes.

## Upstream Behavior Behind The Snapshot Changes

- OXC `#21567` preserves the replaced identifier span when inlining single-use variables during DCE.
  - This explains the sourcemap provenance changes in `packages/rolldown/tests/fixtures/output/sourcemap/true` and `sourcemap-base-url`.
  - It also explains `issue_6263`, where `/* @vite-ignore */` is preserved during inlining.
  - The same DCE path is the best explanation for `ts_import_equals_undefined_import` and `transform-empty-mappings` becoming more aggressively folded / mapped to use sites.
- OXC `#21568` and `#21571` explain `this_outside_function` dropping `this` when it is side-effect free.

## Impact

- OXC version alignment stays current on both the Rust and JS sides.
- Sourcemap expectations now match upstream OXC use-site span preservation during DCE.
- A few esbuild / rolldown snapshots become smaller or structurally different due to stronger folding.

## Validation

- `just roll`
  - This ran through Rust tests, clippy, cargo checks, Node tests, rollup tests, and `vp check`.
  - It then stopped at `vp run lint-knip` with unlisted-dependency warnings from generated test outputs under `packages/rolldown/tests/.../dist`.
  - That blocker is repo-level validation noise, not a failure in the OXC upgrade itself.


crates/rolldown/tests/esbuild/ts/ts_import_equals_undefined_import/artifacts.snap
```diff
@@ -6,10 +6,7 @@ source: crates/rolldown_testing/src/integration_test.rs
 ## entry.js
 
 ```js
-// HIDDEN [\0rolldown/runtime.js]
-//#region entry.ts
-let foo = (/* @__PURE__ */ __exportAll({ value: () => 123 })).value;
-console.log(foo);
```
Dunqing, line 12: I do not exactly know why this change occurred; it may be related to https://github.com/oxc-project/oxc/pull/21567.

Dunqing:
Most of the snapshot changes are introduced by https://github.com/oxc-project/oxc/pull/21567, which even corrected the sourcemap.

shulaoda (approved):

crates/rolldown/tests/esbuild/ts/ts_import_equals_undefined_import/artifacts.snap
sapphi-red: It seems it works because the inlining works because the span is preserved

PR #9195, opened 2026-04-22
refactor(rolldown_plugin_vite_import_glob): do not rewrite import path for absolute base

shulaoda:
Related to #9145, https://github.com/rolldown/rolldown/pull/9145#discussion_r3122326775

## Summary

- Stop overriding `import_path` when `options.base` starts with `/`. Previously the code re-computed `import_path = self.relative_path(file, None)`, which was intended to emulate Vite's `/${relative(root, file)}` shape but produced a broken result for files **outside** `root` (no leading `/`, e.g. `../basic/dir/a.js`).
- Keep `import_path` as `self.relative_path(file, Some(dir))` — a plain path relative to the importer's directory — regardless of whether `base` is absolute. This yields stable, directly-resolvable relative imports that don't depend on any resolver-specific behavior.
- Extend the `import-glob/parent` fixture with two new cases covering `base: '/'`, including one where the glob reaches a sibling directory (`../basic/dir/*.js`) outside the importer's folder.

## Why not align with Vite's `importPath = \`/${relative(root, file)}\``?

Vite's `importMetaGlob.ts` at [`L495-L497`](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/plugins/importMetaGlob.ts#L495-L497) rewrites `importPath` to an absolute-looking form (e.g. `/../external/a.js`) and relies on its resolver to turn that back into a real file. Concretely, `rolldown_plugin_vite_resolve` sets:

```rust
roots: if base_options.as_src { vec![base_options.root.clone()] } else { vec![] }
```

oxc-resolver's `roots` option says: for requests starting with `/`, resolve them against each root (plus an absolute-path attempt on non-Windows). That is what lets Vite recover `/../external/a.js` — it falls through to `${root}/../external/a.js` and normalizes.

This only works while `asSrc` is `true`. For `oxcResolvePlugin` that is effectively always the case, **except inside `createIdResolver`**, where `asSrc` is `false` and `roots` becomes empty. In that path, a generated `/../external/a.js` would fail to resolve.

Using an importer-relative path (`../basic/dir/a.js`) side-steps this entirely: it's a plain relative specifier that any resolver handles the same way, independent of `asSrc`/`roots`. The trade-off is an intentional shape divergence from Vite — the keys in the returned `import.meta.glob` object differ when `base: '/'` is used with out-of-dir matches — but resolution becomes correct and resolver-agnostic.


shulaoda:
Do you think this change is reasonable? To some extent, it diverges from Vite rather than strictly aligning with it, and instead introduces a more generalized extension. @sapphi-red

hyf0 (approved):

sapphi-red:
Is this a fix? or is this a refactor?

shulaoda:
> Is this a fix? or is this a refactor?

Oh, sorry, this is more of a refactor aimed at making it extensible for third-party usage.

sapphi-red (approved):
LGTM let's update the code in Vite as well.

PR #9197, opened 2026-04-22
fix(link): error on missing export between TS modules

### LINKED ISSUE #9141: [Bug]: preserveModules emits dangling export reference for TypeScript interface re-exported without `type` keyword

https://repl.rolldown.rs/ (or minimal repo below)

**Source files:**

`src/types.ts`:
```ts
export interface Foo {
  name: string;
}
```

`src/impl.ts`:
```ts
export class Bar {
  value = 42;
}
```

`src/index.ts`:
```ts
export { Foo } from './types';
export { Bar } from './impl';
```

**Rolldown config:**
```js
module.exports = {
  input: { index: 'src/index.ts' },
  output: {
    dir: 'dist',
    preserveModules: true,
    preserveModulesRoot: 'src',
    sourcemap: true,
  },
  external: [],
};
```

### What is expected?

Since `Foo` is a TypeScript `interface` (type-only, no runtime value), rolldown should fully elide it from the JS output — both the import and the export reference should be removed. The output `dist/index.js` should be:

```js
import { Bar } from "./impl.js";
export { Bar };
```

This is how Rollup handles the same scenario.

Rolldown correctly strips the **import** of `Foo` (since there's no runtime binding to import), but **leaves `Foo` in the export statement**, producing:

```js
import { Bar } from "./impl.js";
export { Bar, Foo };
```

`Foo` is never imported or defined, so this is invalid JavaScript. Downstream consumers (e.g., webpack) fail with:

```
Module parse failed: Export 'Foo' is not defined
```

### System Info

```Shell
rolldown: 1.0.0-rc.11
Node: v22.x
OS: Linux
```

### Important notes

- This only affects `export { X } from './file'` where `X` is a **pure type** (interface or type alias) and the `type` keyword is **not** used.
- Using `export type { Foo } from './types'` works correctly — rolldown fully strips both the import and export.
- The issue occurs with `preserveModules: true`. It may also occur without `preserveModules` but that was not tested.
- TypeScript does **not** require the `type` keyword on re-exports of type-only symbols (unless `isolatedModules` + `verbatimModuleSyntax` are both enabled). Bundlers that process TypeScript (including Rollup, esbuild, and tsc) handle this correctly without the `type` keyword.

### Workaround

Add the `type` keyword to re-exports of type-only symbols:

```ts
export type { Foo } from './types';  // add `type` keyword
export { Bar } from './impl';
```

### Any additional comments?

Related but distinct from:
- #6992 — was about `export *` from externals producing undefined namespace objects (fixed in beta.51)
- #3133 — was about `export { type Foo, Bar }` (mixed type/value in one statement) being entirely ignored (fixed in #3150)
- #9122 — about runtime re-exports being dropped with `preserveModules`

This bug is specifically about rolldown's TypeScript type elision being incomplete: it strips the import side of a type-only re-export but not the export side, producing invalid output.

IWANABETHATGUY:
Rolldown previously downgraded `MISSING_EXPORT` to a warning when both importer and importee were TypeScript modules, while still emitting the dangling re-export in the output — producing syntactically valid but semantically broken JavaScript that crashed downstream bundlers like webpack.

This change removes the TS-to-TS severity downgrade so a missing export is always an error, matching Rollup's behavior. The existing `If you meant to import a type...` hint is preserved.

Fixes #9141.

### DIFF
--- a/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
+++ b/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
@@ -855,7 +855,7 @@ impl BindImportsAndExportsContext<'_> {
               importee.as_normal().map(|m| &m.module_type),
               Some(ModuleType::Ts | ModuleType::Tsx)
             ) && matches!(module.module_type, ModuleType::Ts | ModuleType::Tsx);
-          let mut diagnostic = BuildDiagnostic::missing_export(
+          let diagnostic = BuildDiagnostic::missing_export(
             module.id.to_string(),
             module.stable_id.to_string(),
             importee.id().to_string(),
@@ -865,12 +865,7 @@ impl BindImportsAndExportsContext<'_> {
             named_import.span_imported,
             is_ts_like_importing_ts_like.then(|| format!("If you meant to import a type rather than a value, make sure to add the `type` modifier (e.g. `import {{ type Foo }} from '{}'`).", rec.module_request))
           );
-          if is_ts_like_importing_ts_like {
-            diagnostic = diagnostic.with_severity_warning();
-            self.warnings.push(diagnostic);
-          } else {
-            self.errors.push(diagnostic);
-          }
+          self.errors.push(diagnostic);
         }
       }
     }
--- a/crates/rolldown/tests/esbuild/ts/ts_export_default_type_issue316/_config.json
+++ b/crates/rolldown/tests/esbuild/ts/ts_export_default_type_issue316/_config.json
@@ -7,5 +7,5 @@
       }
     ]
   },
-  "expectExecuted": false
+  "expectError": true
 }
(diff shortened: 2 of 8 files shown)


IWANABETHATGUY:
* **#9197** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9197?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9197?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9197).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

h-a-n-a (approved):

shulaoda (approved):

silverwind:
Seems this exposed bug https://github.com/sxzz/rolldown-plugin-dts/issues/230.

IWANABETHATGUY:
> Seems this exposed bug [sxzz/rolldown-plugin-dts#230](https://github.com/sxzz/rolldown-plugin-dts/issues/230).

We aligned with rollup, which emits an error by default when an export is missing.

silverwind:
Yeah, my take is that this is that plugin's fault, it was previously suppressing the warnings.

PR #9218, opened 2026-04-23
fix(docs): bypass TypeDoc for code blocks injected via 'include'

### LINKED ISSUE #8792: doc: backslashes in code blocks rendered as URL-encoded sequences on docs site
## Description

Backslashes inside fenced code blocks in `packages/rolldown/src/options/docs/*.md` files are rendered as URL-encoded sequences (`%5C`) on [rolldown.rs](https://rolldown.rs).

**Example:** [rolldown.rs/reference/InputOptions.external](https://rolldown.rs/reference/InputOptions.external)

The regex example:
```js
// Source (.md file)
external: /^[^./](?!:\\)/,

// Rendered on site
external: /^[^./](?!:%5C%5C)/,
```

`%5B` = `[`, `%5C` = `\`, `%5D` = `]` — the characters are being URL-encoded instead of rendered literally.

Originally reported by @bddjr in https://github.com/rolldown/rolldown/pull/8780#issuecomment-2983429547.

## Root Cause

The docs flow is:

1. `external.md` contains correct markdown with fenced code blocks
2. `{@include ./docs/external.md}` injects the content into a JSDoc comment in `input-options.ts`
3. **TypeDoc's comment parser processes the injected content as JSDoc markdown, including the contents of fenced code blocks** — backslashes get treated as escape sequences
4. `extract-options-plugin.ts` extracts the (already mangled) content
5. VitePress renders the mangled output

The issue is that TypeDoc's `{@include}` does not preserve code block boundaries when processing escape sequences in its block lexer (see `blockLexer.ts` — it handles `\{`, `\}`, `\@`, `\/`, `` \` `` as escape pairs).

**Why `codeSplitting` docs are not affected:** The regex `[\\/]` in `output-code-splitting-example.md` survives because `\/` is a recognized escape pair that resolves to `/` — which happens to be correct. But `\\` followed by `]` does not match any escape pair, so it goes through a different path and gets URL-encoded.

## Possible Fixes

1. **Best: bypass TypeDoc for doc content** — `extract-options-plugin.ts` already knows which property maps to which `docs/*.md` file. Instead of extracting the content from TypeDoc's processed output, read the original `.md` file directly and splice it in. This avoids the TypeDoc parser entirely.

2. **Upstream: fix TypeDoc** — `{@include}` should skip escape processing inside fenced code blocks. Correct but depends on TypeDoc team.

3. **Workaround: post-process URL decoding** — Add `section.replace(/%5B/g, "[").replace(/%5C/g, "\\").replace(/%5D/g, "]")` in `extract-options-plugin.ts`. Quick but fragile.

## Affected Pages

Any `docs/*.md` file with backslashes in code blocks that is `{@include}`'d into a JSDoc comment. Currently:
- `external.md` — regex with `\\`

Kyujenius:
Fenced code blocks injected into JSDoc comments via `{@include}` are rewritten by TypeDoc's block lexer, which URL-encodes `[`, `\`, and `]`. The regex `/^[^./](?!:[/\\])/` from `packages/rolldown/src/options/docs/external.md` therefore renders as `/^[^./](?!:%5B/%5C%5C%5D)/` on [`/reference/InputOptions.external`](https://rolldown.rs/reference/InputOptions.external); the same mangling appears on [`Interface.RolldownOptions`](https://rolldown.rs/reference/Interface.RolldownOptions) and [`Interface.WatchOptions`](https://rolldown.rs/reference/Interface.WatchOptions), which share the option's `{@include}` content.

`extract-options-plugin.ts` now resolves each property's `{@include}` targets from its JSDoc and replaces the rendered fenced blocks with the original `.md` content before TypeDoc writes the page. Prose and heading levels remain TypeDoc's output. The replacement is skipped when the fenced block count on the page differs from the count in the source `.md`, so a property that combines `{@include}` with `@example` is left untouched.

### Affected pages

- [`/reference/InputOptions.external`](https://rolldown.rs/reference/InputOptions.external)
- [`/reference/Interface.InputOptions`](https://rolldown.rs/reference/Interface.InputOptions)
- [`/reference/Interface.RolldownOptions`](https://rolldown.rs/reference/Interface.RolldownOptions)
- [`/reference/Interface.WatchOptions`](https://rolldown.rs/reference/Interface.WatchOptions)

Fixes #8792.

### DIFF
--- a/docs/.vitepress/scripts/extract-options-plugin.ts
+++ b/docs/.vitepress/scripts/extract-options-plugin.ts
@@ -6,6 +6,117 @@ function escapeRegex(s: string) {
   return s.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 }
 
+const sourceFileCache = new Map<string, string>();
+function readSourceFile(filePath: string): string {
+  let contents = sourceFileCache.get(filePath);
+  if (contents === undefined) {
+    contents = fs.readFileSync(filePath, 'utf8');
+    sourceFileCache.set(filePath, contents);
+  }
+  return contents;
+}
+
+/**
+ * Walk upward from a property's source line to find the `/** ... *\/` JSDoc
+ * block that immediately precedes it. Returns the raw text including the
+ * opening and closing markers, or undefined if no adjacent block exists.
+ */
+function findJSDocAbove(sourceText: string, propertyLine: number): string | undefined {
+  const lines = sourceText.split('\n');
+  let end = propertyLine - 2;
+  while (end >= 0 && lines[end].trim() === '') end--;
+  if (end < 0 || !lines[end].trimEnd().endsWith('*/')) return undefined;
+  let start = end;
+  while (start >= 0 && !lines[start].trimStart().startsWith('/**')) start--;
+  if (start < 0) return undefined;
+  return lines.slice(start, end + 1).join('\n');
+}
+
+/**
+ * Resolve the {@link include} targets of a property's JSDoc to absolute paths.
+ * Mirrors TypeDoc's `{@include}` inline-tag semantics, but gives us a handle
+ * on the original `.md` content before TypeDoc's block lexer rewrites escape
+ * sequences — see https://github.com/rolldown/rolldown/issues/8792.
+ */
+function resolveIncludePaths(property: td.Reflection): string[] {
+  const source = property.sources?.[0];
+  if (!source) return [];
+  const filePath =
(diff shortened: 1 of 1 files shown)


sapphi-red:
I think this is a bug in typedoc and don't think we need to have a patch in rolldown.

Kyujenius:
@sapphi-red 
Thanks for your comments! 
I agree this should be fixed upstream. I'll look into opening an issue/PR on TypeDoc's side to handle 'include' content inside fenced code blocks properly. Closing this for now.

PR #9164, opened 2026-04-20
fix(chunk-optimizer): pick dominator for runtime placement to avoid cycles

IWANABETHATGUY:
Fixes a chunk-splitting issue where the runtime module could be placed in a chunk that introduced a cyclic dependency between chunks. The optimizer now selects the dominator of all runtime consumers for placement, ensuring runtime-containing chunks load before their dependents. Includes a regression test for issue #8920 and updates to the code-splitting design doc.

closes #9161
closes https://github.com/rolldown/rolldown/issues/9224

PR #9226, opened 2026-04-25
fix: prevent chunk optimizer from creating import cycles

### LINKED ISSUE #7449: [Bug]: incorrect execution order with CJS + advancedChunks

mldangelo-oai:
## Summary

- Fix chunk-optimizer cycle detection by modeling merges as graph contractions instead of special-casing source/target dependency walks.
- Track aliases for common chunks already merged into existing chunks, so later cycle and reachability checks resolve stale dependency edges through the final merge target.
- Normalize target dependency sets after merges, preventing stale intra-group or alias-chain edges from lingering in the temporary graph.
- Apply safe same-target common-chunk merge groups atomically, with an individual fallback when the whole group would create a static import cycle.
- Keep alias-aware facade reachability compatible with runtime rehoming, so #8989-style safe facade elimination is not blocked by the more accurate graph walk.
- Add a regression fixture for the Promptfoo/tsdown circular-chunk failure and document the merge-contraction/runtime-rehome behavior in the code-splitting design note.

Related to rolldown/tsdown#760 and the Promptfoo/tsdown reproduction discussed in #7449. This does not close #7449's original CJS execution-order reproduction.

## Why

`tsdown@0.21.10` builds Promptfoo with `rolldown@1.0.0-rc.17` into a cyclic static chunk graph. At runtime the generated server imports `fetch-*.js`, which imports back into `server/index.js` before `state$2` is initialized:

```text
ReferenceError: Cannot access 'state$2' before initialization
    at getEnvString (.../dist/src/server/index.js:125:2)
    at .../dist/src/fetch-*.js:44:28
```

The previous cycle check only explored `source.deps` when the source chunk had dependencies. That misses cases where the target already reaches the source through another chunk, so merging the source into the target creates `target -> ... -> target`.

## Related Work

- Supersedes #8227 with a narrower graph-level fix. That PR explored a defensive TDZ mitigation in `manualCodeSplitting`; this PR keeps the generated static chunk graph acyclic instead.
- Related to #7449, but does not close it. The original issue body is an execution-order repro; this PR covers the later Promptfoo/tsdown circular-chunk case discussed there.
- Related to #9164. That draft fixes runtime placement cycles. This patch applies cleanly on top of #9164 and keeps the #8989 runtime-rehome path working with alias-aware reachability.
- Related to #8625 / #6677 in the broader `advancedChunks` area. That PR fixes empty chunk generation and also touches `chunk_optimizer.rs`, but the issue class is different.

## Validation

- `cargo test -p rolldown chunk_optimizer::tests -- --nocapture`
- `just t-run crates/rolldown/tests/rolldown/issues/tsdown_760/_config.json`
- `just t-run crates/rolldown/tests/rolldown/issues/8989/_config.json`
- `just t-run crates/rolldown/tests/rolldown/issues/9049/_config.json`
- `just t-run crates/rolldown/tests/rolldown/code_splitting/facade_elimination_with_shared_cjs_dep/_config.json`
- `just t-run crates/rolldown/tests/rolldown/code_splitting/issue_8184/_config.json`
- `just t-run crates/rolldown/tests/rolldown/code_splitting/issue_8809/_config.json`
- `just lint-rust && just lint-repo && git diff --check`
- `just build-rolldown`
- Earlier full-branch validation: `just test-rust`, `just test-node-rolldown`, `just test-node-rollup`, `vp check`, `vp run lint-publint`.
- Promptfoo PR #8878 proof:
  - unpatched `tsdown@0.21.10` / `rolldown@1.0.0-rc.17` reproduces the server startup TDZ crash.
  - local Promptfoo override to this patched Rolldown package clean-builds, starts `dist/src/server/index.js`, runs the CI `--share` eval, and `/api/results` returns exactly one uploaded result.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -969,6 +969,7 @@ impl GenerateStage<'_> {
       if is_target_manual_chunk && is_emitted_entry_chunk {
         emitted_chunk_groups.entry(target_chunk_idx).or_default().push(from_chunk_idx);
       } else if !is_emitted_entry_chunk {
+        // See meta/design/code-splitting.md: facade reachability must account for runtime rehoming.
         // Check if merging would create a runtime-helper circular dependency.
         // Translate chunk_graph indices to temp_chunk_graph indices, since the two graphs may
         // diverge once new common chunks are materialised. Preserve the existing conservative
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -63,6 +63,8 @@ pub struct ChunkOptimizationGraph {
   /// Initial chunks share identical indices; newly-created common chunks
   /// in chunk_graph are registered here so callers can translate.
   chunk_idx_to_temp_chunk_idx: FxHashMap<ChunkIdx, ChunkIdx>,
+  /// Tracks common chunks that have already been merged into existing chunks.
+  merged_chunk_aliases: FxHashMap<ChunkIdx, ChunkIdx>,
 }
 
 impl ChunkOptimizationGraph {
@@ -106,7 +108,13 @@ impl ChunkOptimizationGraph {
       bits_to_chunk_idx.iter().map(|(k, v)| (k.clone(), *v)).collect();
     bits_to_chunk_idx.sort_unstable_keys();
 
-    Self { chunks, bits_to_chunk_idx, module_to_chunk, chunk_idx_to_temp_chunk_idx }
+    Self {
+      chunks,
+      bits_to_chunk_idx,
+      module_to_chunk,
+      chunk_idx_to_temp_chunk_idx,
+      merged_chunk_aliases: FxHashMap::default(),
+    }
   }
 
   /// Assigns a module to a temporary chunk based on its reachability bits.
@@ -199,9 +207,8 @@ impl ChunkOptimizationGraph {
   /// that previously depended on `source` now depends on `target`. A cycle exists iff `target`
   /// is reachable from `source.deps ∪ target.deps` in the resulting graph.
   ///
-  /// We approximate this by doing a BFS from `source.deps ∪ target.deps`, treating any chunk
-  /// that depends on `source` as also pointing to `target` (since after the merge they would).
-  /// If `target` is reachable, the merge would create a cycle.
+  /// We detect this by contracting `source` into `target`, then checking whether the merged
+  /// target can reach itself through any external chunk.
   ///
   /// This is similar to Rollup's check in `getAdditionalSizeIfNoTransitiveDependencyOrNonCorrelatedSideEffect`.
   pub fn would_create_circular_dependency(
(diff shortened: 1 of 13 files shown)


IWANABETHATGUY:
The original issue is caused by incorrect execution order, and I don't see any circular reference in the related reproduction. It seems you closed the wrong related issue.

IWANABETHATGUY:
I have verified that you didn't solve the issues in #7449.

mldangelo-oai:
Thanks, I agree with the closure as scoped. I incorrectly framed this as fixing #7449.

I re-checked #7449's original repro and agree it is an execution-order issue. This branch is targeting a different failure mode that came from the Promptfoo tsdown upgrade: `tsdown@0.21.10` / `rolldown@1.0.0-rc.17` emitted a circular static chunk graph, which crashed at startup with:

```text
ReferenceError: Cannot access 'state$2' before initialization
```

That downstream failure is the concrete issue I was trying to fix, not the original #7449 CJS/advancedChunks repro.

I've updated the branch to remove the #7449 closing claim, renamed the fixture away from `issues/7449`, and hardened the regression test to assert the emitted static chunk graph is acyclic.

I'll split this into a dedicated issue + new PR for the Promptfoo/tsdown static import cycle instead of asking to reopen this PR. Before I do that, is there anything specific you'd want the new PR to include or avoid to make the scope acceptable? For example: a different fixture shape, stronger before/after generated chunk evidence, a narrower optimizer change, or a different place in the pipeline to enforce the acyclic chunk-graph invariant.

mldangelo-oai:
Small update after checking for an existing issue: I found #9225, which appears to cover the same static chunk-cycle failure class.

I verified it locally against the reported EVMcrispr repro:

- with Vite 8.0.10 / Rolldown rc.17, the output has the reported static cycle: `index-*.js -> secp256k1-*.js -> index-*.js`
- with this branch's rebuilt Rolldown package swapped in, the same build completes and the static cycle is gone: `index-*.js` still imports `secp256k1-*.js`, but `secp256k1-*.js` no longer imports back from `index-*.js`

I also found the downstream Promptfoo failure that led me here:

- Promptfoo PR: https://github.com/promptfoo/promptfoo/pull/8878 (`chore(deps): update dependency tsdown to v0.21.10`)
- failing run/job: https://github.com/promptfoo/promptfoo/actions/runs/24916015596/job/72968106530 (`Share Test`)
- that build logs `tsdown v0.21.10 powered by rolldown v1.0.0-rc.17`, then the server exits before readiness with:

```text
ReferenceError: Cannot access 'state$2' before initialization
    at getEnvString (file:///home/runner/work/promptfoo/promptfoo/dist/src/server/index.js:125:2)
    at getEnvInt$1 (file:///home/runner/work/promptfoo/promptfoo/dist/src/server/index.js:152:14)
    at file:///home/runner/work/promptfoo/promptfoo/dist/src/fetch-BqOTcZcw.js:44:28
```

So I won't open a duplicate issue. If I open a replacement PR, I'll scope it to #9225 plus the Promptfoo/tsdown downstream case, and keep the explicit note that it does not fix #7449's original execution-order repro.

IWANABETHATGUY:
> Small update after checking for an existing issue: I found #9225, which appears to cover the same static chunk-cycle failure class.
> 
> I verified it locally against the reported EVMcrispr repro:
> 
> * with Vite 8.0.10 / Rolldown rc.17, the output has the reported static cycle: `index-*.js -> secp256k1-*.js -> index-*.js`
> * with this branch's rebuilt Rolldown package swapped in, the same build completes and the static cycle is gone: `index-*.js` still imports `secp256k1-*.js`, but `secp256k1-*.js` no longer imports back from `index-*.js`
> 
> So I won't open a duplicate issue. If I open a replacement PR, I'll scope it to #9225 plus the Promptfoo/tsdown downstream case, and keep the explicit note that it does not fix #7449's original execution-order repro.

I think you overcomplicated the issue, I have already prepare a patch that is clearer

IWANABETHATGUY:
I just verified locally, the issue is regressed by https://github.com/rolldown/rolldown/pull/9049

PR #9228, opened 2026-04-25
fix: prevent chunk optimizer from creating import cycles

### LINKED ISSUE #9225: Circular chunk dependency causes TDZ error with @noble/curves + @noble/hashes (viem/wagmi)
## Bug Description

Rolldown (1.0.0-rc.17 via Vite 8.0.10) splits `@noble/hashes` and `@noble/curves` into separate chunks that form a **static** circular dependency. This causes a runtime error because `@noble/curves/secp256k1` reads `sha256` from `@noble/hashes` at **module evaluation time** (top-level curve construction), but the live binding is still `undefined` due to the circular chunk initialisation order.

This works correctly with Vite 7 / Rollup.

## Error

```
Error: param hash is invalid. Expected hash, got undefined
    at r (utils-BeVtMw_R.js:1)
    at k (utils-BeVtMw_R.js:1)
    at we (secp256k1-DjV6mbXe.js:1)
    at Te (secp256k1-DjV6mbXe.js:1)
    at ke (secp256k1-DjV6mbXe.js:1)
    at <anonymous> (secp256k1-DjV6mbXe.js:1)
```

## What happens

Rolldown produces three relevant chunks:

| Chunk | Contains |
|---|---|
| `index-*.js` (entry) | `@noble/hashes/sha256`, viem core, React, app code |
| `secp256k1-*.js` | `@noble/curves/secp256k1` + abstract curve utilities (Field, weierstrass, hash-to-curve) |
| various lazy chunks | EVMcrispr modules (dynamic imports) |

The dependency graph between them:

```
index-*.js ──static──▶ secp256k1-*.js ──static──▶ index-*.js
     │                                                 ▲
     └──────────dynamic import()───────────────────────┘
```

1. `index-*.js` **statically** imports `{a, i, o, r, t}` from `secp256k1-*.js` (the abstract curve utilities: Field, weierstrass, hash-to-curve, mapToCurveSimpleSWU, and the secp256k1 curve)
2. `secp256k1-*.js` **statically** imports `{$, U}` from `index-*.js` (sha256 hash function and hmac)
3. `index-*.js` also **dynamically** `import()`s `secp256k1-*.js` (viem's lazy signature recovery)

When the browser evaluates `index-*.js`, it must first evaluate `secp256k1-*.js` (static dep). But `secp256k1-*.js` reads sha256 from `index-*.js` at **module top-level** — and `index-*.js` hasn't finished initializing yet. So sha256 is `undefined`, and the `@noble/curves` parameter validator throws.

## Reproduction

```bash
git clone https://github.com/EVMcrispr/evmcrispr.git
cd evmcrispr
git checkout c67bc95c84cc7afbe7f8e383f3a3815c1bceec2a
bun install
bun run --filter='evmcrispr-terminal' build
```

Then verify the circular dependency in the output:

```bash
# index statically imports from secp256k1
grep -oP 'import\{[^}]*\}from"./secp256k1[^"]*"' apps/evmcrispr-terminal/dist/assets/index-*.js

# secp256k1 statically imports from index
grep -oP 'import\{[^}]*\}from"./index[^"]*"' apps/evmcrispr-terminal/dist/assets/secp256k1-*.js
```

Serve with `npx vite preview` inside `apps/evmcrispr-terminal` and open the browser console to see the error.

## Expected behaviour

The bundler should either:
- Keep modules with top-level cross-package dependencies in the same chunk, or
- Ensure correct evaluation order so that statically imported bindings are initialised before dependent chunks execute

## Workaround

Force both packages into the same chunk:

```js
// vite.config.ts
build: {
  rollupOptions: {
    output: {
      manualChunks(id) {
        if (id.includes("@noble/hashes") || id.includes("@noble/curves")) {
          return "noble-crypto";
        }
      },
    },
  },
},
```

## Impact

This affects any project using **viem**, **wagmi**, or **ethers** with Vite 8 when the dependency graph is complex enough that Rolldown's chunk optimizer splits `@noble/hashes` and `@noble/curves` into separate chunks with a static circular dependency.

## Environment

- Vite: 8.0.10
- Rolldown: 1.0.0-rc.17
- `@noble/curves`: 1.8.2 (via viem 2.46.3)
- `@noble/hashes`: 1.7.2 (via viem 2.46.3)

## Related issues

- #3650 — Circular deps caused by advancedChunks
- #8812 — TinyMCE module execution order bug (same class of issue)
- vitejs/rolldown-vite#597 — optimizeDeps circular chunk regression

IWANABETHATGUY:
## Summary

Fixes a regression introduced by #9049: when the chunk optimizer merges a shared module into an entry chunk, it can produce a static import cycle between that entry chunk and a manual-split chunk that also depends on the shared module.

## Problem

`would_create_circular_dependency` decides whether merging `source` (the temp chunk holding shared modules) into `target` (an existing entry chunk) is safe. The merge is unsafe iff, post-merge, `target` would reach itself in the dependency graph.

#9049 split the algorithm into two cases based on whether `source` has dependencies, in order to reduce false positives that were blocking legitimate merges (the PR reported ~1,046 blocked merges and +12.5% bundle size on a 2,200-entry application). However, the `source has deps` branch only does a forward BFS from `source.deps` and drops the post-merge alias step that the original algorithm used to model new back-edges introduced by the merge:

```rust
// pre-#9049 step (removed in the source-has-deps branch)
if self.chunks[chunk_idx].dependencies.contains(&source_chunk_idx) {
  queue.push_back(target_chunk_idx);
}
```

Without that step, the BFS cannot see cycles where the closing edge is one of the back-edges created by retargeting `source`'s importers onto `target`.

## Concrete repro shape (issue #9225)

- `main.js` (entry) statically imports `api.js` and `env.js`, plus dynamic imports `lazy.js`, `env-user.js`, `dep-user.js`.
- `api.js` is captured by a manual-split group `{ name: "api", test: "api\\.js" }`.
- `api.js` statically imports `env.js`; `env-user.js` statically imports `env.js`.
- `env.js` is a shared module not captured by any group (`includeDependenciesRecursively: false`).

The optimizer selects `main` as the static dominator for `env.js`'s consumers and merges `env.js` into it. After the merge:

- `api.js` and `env-user.js` rewrite their `./env.js` imports to `./main.js`.
- `main.js` still statically imports `./api.js` (pre-existing edge).
- → static cycle `main.js ⇄ api.js` in the output.

In the `source has deps` branch, the BFS starts from `source.deps = { dep.js's chunk }`. `dep.js` has no further deps and does not reach `main`, so the BFS returns `false` and the merge proceeds. The dropped alias step would have queued `target` when the BFS visited `api`'s chunk (whose `dependencies` contain `source`), correctly detecting the cycle.

## Fix

Revert `would_create_circular_dependency` to the pre-#9049 unified algorithm:

- BFS from `source.deps ∪ target.deps`.
- For each visited chunk `c`, when `source ∈ c.dependencies`, also enqueue `target` (modeling the post-merge edge `c → target`).
- If `target` is dequeued, the merge would create a cycle.

This is sound: it catches both the forward path (`source.deps → target`) and the back-edge path (`target → … → c → source-becomes-target`).

## Why reverting #9049 is safe: the `issues/9049` fixture is already covered by #9085

The existing `crates/rolldown/tests/rolldown/issues/9049/` fixture continues to pass after this revert, because **#9049's `would_create_circular_dependency` change was never load-bearing for that fixture**. Bisecting between rc.15 and rc.16 (rc.15 produces 4 chunks, rc.16 produces 3) pinpoints the actual fix as **#9085** — `fix: relax overly conservative side-effect leak check in chunk optimizer` (commit `5ff1a7265`). #9085 narrows the side-effect leak guard in `try_insert_into_existing_chunk`: the merge is now blocked only when a *different* user-defined entry can also reach the dynamic entries depending on the common chunk, instead of unconditionally blocking on any such dependency. That relaxation is what allows `services/svc{0,1}.js` to merge into the `route0/route1` chunks for this fixture.

The fixture's name (`issues/9049`) is therefore historically misleading: it tests behavior fixed by #9085, not #9049. The false-positive regressions #9049 was addressing in real-world bundles remain a concern but are not exercised by any in-tree fixture; a follow-up can re-tighten the check (e.g. by tracking dependent aliases more precisely) without losing the regression coverage this PR adds.

## Regression test

`crates/rolldown/tests/rolldown/issues/9225/` reproduces the cycle that #9049 was masking.

- `_config.json` sets up `manualCodeSplitting.groups = [{ name: "api", test: "api\\.js" }]` with `includeDependenciesRecursively: false`.
- `_test.mjs` walks `dist/*.js`, parses the static-import edges, and asserts the graph has no cycle. It additionally executes the bundled entry, checks `globalThis.__rolldown_issue_7449_value === 300000`, and verifies that the side-effect module `dep.js` runs exactly once even when the dynamic chunks (which all transitively depend on `dep.js`) load.

Without this fix, `_test.mjs` fails on the cycle assertion before any runtime check runs.

## Testing

- `cargo test -p rolldown --test integration` — **1692 passed, 0 failed, 70 ignored**.
- `just test-node` (Rolldown vitest `test:main` + `test:watcher` + Rollup compatibility suite) — **722 + 31 + 916 passed, 0 failed**.
- New `issues/9225` fixture passes both the cycle assertion and the runtime checks.

Closes #9225.
`


IWANABETHATGUY:
* **#9228** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9228?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9228?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9228).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

h-a-n-a (approved):

PR #9231, opened 2026-04-25
fix: add optional chaining for getOrder method call in plugin normalization

zfogg:
## Summary

Fix TypeError in `makeBuiltinPluginCallable` when a plugin object doesn't have a `getOrder` method.

## Problem

When normalizing builtin plugins, the code attempted to call `getOrder()` without checking if the method exists. This causes a TypeError when a plugin lacks this method:

```
TypeError: callablePlugin.getOrder is not a function
```

This issue occurs in monorepo environments where multiple plugins with varying interfaces are used.

## Solution

Use optional chaining (?.) to safely attempt the method call, returning `undefined` if the method doesn't exist:

```typescript
const order = callablePlugin.getOrder?.(key);
```

This maintains backward compatibility while handling plugins with optional `getOrder` methods.

## Changed Files

- `packages/rolldown/src/builtin-plugin/utils.ts` - Added optional chaining to line 55

### DIFF
--- a/packages/rolldown/src/builtin-plugin/utils.ts
+++ b/packages/rolldown/src/builtin-plugin/utils.ts
@@ -52,7 +52,7 @@ export function makeBuiltinPluginCallable(
       }
     };
 
-    const order = callablePlugin.getOrder(key);
+    const order = callablePlugin.getOrder?.(key);
     if (order == undefined) {
       // @ts-expect-error
       wrappedPlugin[key] = wrappedHook;


zfogg:
## What This Fixes

This resolves a TypeError that occurs in projects using [vite-plus](https://github.com/voidzero-dev/vite-plus) when running `vp check`.

**Error**: `TypeError: callablePlugin.getOrder is not a function`

**Root cause**: The `makeBuiltinPluginCallable` function was calling `getOrder()` on plugin objects without checking if the method exists. This fails when plugins lack the `getOrder` method, which can happen in monorepo environments with heterogeneous plugin interfaces.

**Related issue**: https://github.com/voidzero-dev/vite-plus/issues (vite-plus monorepo integration)

sapphi-red:
I think this is a bug in vite-plus.
`getOrder` method is always defined.
https://github.com/rolldown/rolldown/blob/5a5f8f5d71d42153f21a88eb3ab3630aa86005ec/crates/rolldown_binding/src/options/plugin/binding_callable_builtin_plugin.rs#L58

PR #9236, opened 2026-04-27
feat(experimental/lazy-barrel): advice on oversized barrel modules

shulaoda:
## Summary

Adds a build-time advisory when the experimental `lazyBarrel` flag is enabled and a barrel module exceeds **5000 re-exports** (think large icon packs like `lucide-react` or `@mui/icons-material`). Eagerly resolving every re-export in such a barrel is a known bottleneck, so rolldown now points users at [`@rolldown/plugin-transform-imports`](https://github.com/rolldown/plugins/tree/main/packages/transform-imports), which rewrites imports at the source so the barrel file is never loaded.

Example output:

```
advice[LARGE_BARREL_MODULES]: node_modules/lucide-react/dist/esm/lucide-react.js has 1463 re-exports. Eagerly resolving every entry can significantly slow down the build. Consider using `@rolldown/plugin-transform-imports` to rewrite imports at the source level so the barrel file is never loaded.
  help: See https://github.com/rolldown/plugins/tree/main/packages/transform-imports for usage.
```

## What's in this PR

- **New diagnostic event `LargeBarrelModules`** (`EventKind = 46`) under `rolldown_error`, with a dedicated `BuildEvent` impl that formats the module id, re-export count, and a help link.
- **`Severity::Info` variant** on `BuildDiagnostic`, rendered as `ReportKind::Advice` via ariadne (distinct from the existing `error:` / `warning:` prefixes). A `with_severity()` builder is added so the same diagnostic can be downgraded to an info-level log.
- **Detection in `module_task.rs`**: after lazy-barrel info is computed, if the resolved barrel has more than 5000 import records, an `Info`-level `Log` is dispatched through `on_log` with `code = "LARGE_BARREL_MODULES"`. Gated on the new check flag so users can silence it.
- **New check flag** `checks.largeBarrelModules` (default `true`), propagated through the usual codegen: Rust `EventKindSwitcher`, NAPI binding, TS `ChecksOptions`, validator, JSON schema, and CLI `--checks.large-barrel-modules`.

## How to opt out

```js
export default {
  experimental: { lazyBarrel: true },
  checks: { largeBarrelModules: false },
}
```

Or from the CLI:

```sh
rolldown --no-checks.large-barrel-modules
```

## Notes on the threshold

5000 was chosen so that ordinary component or utility barrels (typically in the low hundreds) stay quiet, and only the real outliers (icon packs, mass re-export shims) trip it. The threshold is intentionally not user-configurable for now; once we have real-world feedback it can be promoted to an option.
image

### DIFF
--- a/packages/rolldown/src/options/generated/checks-options.ts
+++ b/packages/rolldown/src/options/generated/checks-options.ts
@@ -159,4 +159,14 @@ export interface ChecksOptions {
    * @default true
    * */
   ineffectiveDynamicImport?: boolean;
+
+  /**
+   * Whether to emit warnings when a barrel module has a very large number of re-exports (more than 5000).
+   *
+   * Such modules can significantly slow down module resolution. Consider using
+   * [`@rolldown/plugin-transform-imports`](https://github.com/rolldown/plugins/tree/main/packages/transform-imports)
+   * to rewrite barrel imports at the source level so the barrel file is never loaded.
+   * @default true
+   * */
+  lazyBarrelLargeReexports?: boolean;
 }
--- a/packages/rolldown/src/options/generated/checks-options.ts
+++ b/packages/rolldown/src/options/generated/checks-options.ts
@@ -161,12 +161,12 @@ export interface ChecksOptions {
   ineffectiveDynamicImport?: boolean;
 
   /**
-   * Whether to emit warnings when a barrel module has a very large number of re-exports (more than 5000).
+   * Whether to emit info logs when a barrel module has a very large number of re-exports (more than 5000).
    *
    * Such modules can significantly slow down module resolution. Consider using
    * [`@rolldown/plugin-transform-imports`](https://github.com/rolldown/plugins/tree/main/packages/transform-imports)
    * to rewrite barrel imports at the source level so the barrel file is never loaded.
    * @default true
    * */
-  lazyBarrelLargeReexports?: boolean;
+  largeBarrelModules?: boolean;
 }
--- a/crates/rolldown/src/module_loader/module_task.rs
+++ b/crates/rolldown/src/module_loader/module_task.rs
@@ -10,7 +10,8 @@ use rolldown_common::{
   StrOrBytes, try_extract_lazy_barrel_info,
 };
 use rolldown_error::{
-  BuildDiagnostic, BuildResult, UnloadableDependencyContext, downcast_napi_error_diagnostics,
+  BuildDiagnostic, BuildResult, EventKindSwitcher, UnloadableDependencyContext,
+  downcast_napi_error_diagnostics,
 };
 use rolldown_std_utils::PathExt as _;
 use rolldown_utils::{ecmascript::legitimize_identifier_name, indexmap::FxIndexSet};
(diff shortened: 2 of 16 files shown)


shulaoda:
* **#9236** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9236?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9236?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9236).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyfdev:
What's the intention of this warning? Is it worth to have it? @sapphi-red cc

sapphi-red:
It is a "info" level message rather than a warning. I think it is worth to have it as it would be easier to notice the performance overhead caused by it.

shulaoda:
> It is a "info" level message rather than a warning.

Do you mean emitting an info log through `onLog`?

sapphi-red:
What do you mean by "through `onLog`"? I think it should be emitted directly as a info log.

shulaoda:
> What do you mean by "through `onLog`"? I think it should be emitted directly as a info log.

Do you mean logging directly on the Rust side with `print` or writing an info log, instead of going through `options.onLog` or `warnings`?

sapphi-red:
> Do you mean logging directly on the Rust side with `print` or writing an info log, instead of going through `options.onLog` or `warnings`?

Ah, I thought you were thinking of changing the severity in `onLog`. I mean going through `options.onLog`. The input to `onLog` would have `info` severity.

shulaoda:
> It is a "info" level message rather than a warning.

Why shouldn’t this be a performance warning related to large barrel modules? Do you think the switch controlling it should live under `checks`, or under `lazyBarrel: { detectLargeBarrelModules: boolean }`?

It’s worth noting that the current documentation describes `checks` as the place for controlling warnings, and the options under `checks` are automatically generated from diagnostic `EventKinds`.

Maybe we should use the `checks` approach, but expand its documentation to describe it as controlling both infos and warnings?

sapphi-red:
> Why shouldn’t this be a performance warning related to large barrel modules?

Because the user is not doing anything wrong. In hindsight, I think `pluginTimings` and `preferBuiltinFeature` should be an info message as well.

> Do you think the switch controlling it should live under `checks`, or under `lazyBarrel: { detectLargeBarrelModules: boolean }`?

I think it can be under `checks`.

packages/rolldown/src/options/generated/checks-options.ts
```diff
@@ -159,4 +159,14 @@ export interface ChecksOptions {
    * @default true
    * */
   ineffectiveDynamicImport?: boolean;
+
+  /**
+   * Whether to emit info logs when a barrel module has a very large number of re-exports (more than 5000).
+   *
+   * Such modules can significantly slow down module resolution. Consider using
+   * [`@rolldown/plugin-transform-imports`](https://github.com/rolldown/plugins/tree/main/packages/transform-imports)
+   * to rewrite barrel imports at the source level so the barrel file is never loaded.
```
sapphi-red, line 168: (Not a blocker) I think we should add the reasons a bit more here later on.

sapphi-red (approved):

packages/rolldown/src/options/generated/checks-options.ts
shulaoda: I will do it later

PR #9205, opened 2026-04-22
refactor(link_stage): simplify exports-kind filter and clarify safety comments

IWANABETHATGUY:
Collapse the two-step `filter_map` + `for_each` in `determine_module_exports_kind` so the iterator yields `(rec, &NormalModule)` directly via `Module::as_normal`, removing the inline `let-else` inside the closure. Also rewrites the `SAFETY:` and promotion comments to be clearer about what each branch does and why self-import aliasing is safe.

Note: replaces #9178 which was auto-closed as merged after a stack reorder (GitHub detected all its commits were in its former base branch). Same content, now targeting main directly.

PR #9237, opened 2026-04-27
fix(link_stage): restore inline let-else in exports-kind filter

IWANABETHATGUY:
Reverts the iterator restructuring from #9205 in `determine_module_exports_kind`, restoring the original `filter_map(rec.resolved_module)` + inline `let Module::Normal(importee) = ... else { return; }` pattern. The clarified `SAFETY` and promotion comments introduced in that PR are kept intact — only the structural refactor is rolled back.

We're rolling back the structural change because it appears to have introduced a regression that we have not yet been able to reduce to a reliable repro. Restoring the prior shape lets us unblock users while we continue investigating the root cause; once we have a minimal reproduction we can revisit the refactor with a regression test in place.


IWANABETHATGUY:
* **#9237** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9237?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9237?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9237).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

sapphi-red:
I added a test case. You can run it with `cargo test -p rolldown --test integration --release -- cjs_require_esm_none_exports_kind`. Note that you need `--release` as this only happens with release build.

The output of it without fix:
```js
//#region \0rolldown/runtime.js
var __create = Object.create;
var __getOwnPropDesc = Object.getOwnPropertyDescriptor;
var __getProtoOf = Object.getPrototypeOf;
var __hasOwnProp = Object.prototype.hasOwnProperty;
var __commonJSMin = (cb, mod) => () => (mod || (cb((mod = { exports: {} }).exports, mod), cb = null), mod.exports);
var __copyProps = (to, from, except, desc) => {
	if (from && typeof from === "object" || typeof from === "function") for (var keys = __getOwnPropNames(from), i = 0, n = keys.length, key; i < n; i++) {
		key = keys[i];
		if (!__hasOwnProp.call(to, key) && key !== except) __defProp(to, key, {
			get: ((k) => from[k]).bind(null, key),
			enumerable: !(desc = __getOwnPropDesc(from, key)) || desc.enumerable
		});
	}
	return to;
};
var __toESM = (mod, isNodeMode, target) => (target = mod != null ? __create(__getProtoOf(mod)) : {}, __copyProps(isNodeMode || !mod || !mod.__esModule ? __defProp(target, "default", {
	value: mod,
	enumerable: true
}) : target, mod));
//#endregion
//#region aligner.js
var require_aligner = /* @__PURE__ */ __commonJSMin((() => {
	var Aligner = function() {};
	Aligner.prototype.align = function() {};
}));
//#endregion
//#region default_aligner.js
var require_default_aligner = /* @__PURE__ */ __commonJSMin(((exports, module) => {
	var Aligner = (require_aligner(), __toCommonJS(aligner_exports));
	var DefaultAligner = function() {};
	DefaultAligner.prototype = Object.create(Aligner);
	module.exports = DefaultAligner;
}));
require_aligner();
var import_default_aligner = /* @__PURE__ */ __toESM(require_default_aligner());
console.log(import_default_aligner.default);
//#endregion
```
(references `__toCommonJS` and `aligner_exports` which are not declared.

The output with the fix:
```js
//#region \0rolldown/runtime.js
var __create = Object.create;
var __getOwnPropDesc = Object.getOwnPropertyDescriptor;
var __getProtoOf = Object.getPrototypeOf;
var __hasOwnProp = Object.prototype.hasOwnProperty;
var __commonJSMin = (cb, mod) => () => (mod || (cb((mod = { exports: {} }).exports, mod), cb = null), mod.exports);
var __exportAll = (all, no_symbols) => {
	let target = {};
		get: all[name],
		enumerable: true
	});
	if (!no_symbols) __defProp(target, Symbol.toStringTag, { value: "Module" });
	return target;
};
var __copyProps = (to, from, except, desc) => {
	if (from && typeof from === "object" || typeof from === "function") for (var keys = __getOwnPropNames(from), i = 0, n = keys.length, key; i < n; i++) {
		key = keys[i];
		if (!__hasOwnProp.call(to, key) && key !== except) __defProp(to, key, {
			get: ((k) => from[k]).bind(null, key),
			enumerable: !(desc = __getOwnPropDesc(from, key)) || desc.enumerable
		});
	}
	return to;
};
var __toESM = (mod, isNodeMode, target) => (target = mod != null ? __create(__getProtoOf(mod)) : {}, __copyProps(isNodeMode || !mod || !mod.__esModule ? __defProp(target, "default", {
	value: mod,
	enumerable: true
}) : target, mod));
var __toCommonJS = (mod) => __hasOwnProp.call(mod, "module.exports") ? mod["module.exports"] : __copyProps(__defProp({}, "__esModule", { value: true }), mod);
//#endregion
//#region aligner.js
var aligner_exports = /* @__PURE__ */ __exportAll({});
var Aligner;
var init_aligner = __esmMin((() => {
	Aligner = function() {};
	Aligner.prototype.align = function() {};
}));
//#endregion
//#region default_aligner.js
var require_default_aligner = /* @__PURE__ */ __commonJSMin(((exports, module) => {
	var Aligner = (init_aligner(), __toCommonJS(aligner_exports));
	var DefaultAligner = function() {};
	DefaultAligner.prototype = Object.create(Aligner);
	module.exports = DefaultAligner;
}));
//#endregion
//#region main.js
init_aligner();
var import_default_aligner = /* @__PURE__ */ __toESM(require_default_aligner());
console.log(import_default_aligner.default);
//#endregion
```

The explanation by claude:

> In the buggy code, the `filter_map` closure holds `&self.module_table` for the entire iterator chain lifetime. LLVM sees this immutable reference and applies optimizations that exploit the immutability guarantee — effectively caching or eliding reads of exports_kind from self.module_table. The unsafe mutation via `addr_of!(*importee).cast_mut()` violates this guarantee, causing stale reads. The fix moves `as_normal()` back into for_each, so filter_map no longer captures `&self.module_table`. This changes the LLVM IR structure such that the specific optimization exploiting the immutable reference is not applied. The unsafe mutations of exports_kind take effect as intended.
> 
> Note: the fundamental UB still exists (the outer iterator `self.module_table.modules.iter()` holds an immutable reference while for_each mutates through it). A proper fix would eliminate the unsafe mutation entirely, e.g., by collecting mutations and applying them after the iterator completes, or using Cell/interior mutability.

sapphi-red (approved):
I think we should merge this, but I think we should try to avoid the `unsafe` here.

IWANABETHATGUY:
> I think we should merge this, but I think we should try to avoid the `unsafe` here.

agree

PR #9240, opened 2026-04-27
docs: replace `Inter` with system font stack in OG template SVG

### LINKED ISSUE #8427: OG image is using wrong fonts
The OG image template assumes Inter is available. But it isn't in the building environment, so the rendered OG images look off and misaligned:

Image

yvbopeng:
Fixes #8427


sapphi-red (approved):
Thanks!

PR #9245, opened 2026-04-28
chore(ai): add rolldown REPL decode skill

Dunqing:
## Summary

Adds the `rolldown-repl-decode` Claude Code skill into the repo so it ships with the project.

## Why

When users report a Rolldown bug, the most useful context is almost always the REPL share link — it captures the exact source files **and** the Rolldown version that reproduce the issue. But the REPL stores all of that state in `location.hash` as `base64(zlib(JSON))`, which the browser never sends to the server. A naive `fetch` of the URL returns an empty page, so Claude Code has no way to read what the user actually typed.

This skill bridges that gap. With it installed, Claude Code can:

- Detect a Rolldown REPL link (`repl.rolldown.rs/#…`, `rolldown-repl.netlify.app/#…`) anywhere in an issue body or chat.
- Decode the hash locally to recover the source files + Rolldown version.
- Optionally drop the files onto disk so the bug can be reproduced against the matching Rolldown version.

End result: Claude Code can analyze and fix REPL-linked issues with the actual reproduction in hand, instead of asking the reporter to repaste their code.

## What's in this PR

- `.claude/skills/rolldown-repl-decode/SKILL.md` — when-to-trigger metadata + usage.
- `.claude/skills/rolldown-repl-decode/scripts/decode.mjs` — plain Node script (no deps) that prints the decoded file list + version, and writes the sources to disk with `--write <dir>`.
- `.gitignore` — loosen `.claude` ignore so `.claude/skills/` is tracked, while keeping local user state (`settings.local.json`, history, worktrees) ignored.

## Test plan

- [x] `node .claude/skills/rolldown-repl-decode/scripts/decode.mjs` prints usage.
- [x] Decoded a real REPL share URL end-to-end (file list + version + `--write` output).

### DIFF
--- a/.claude/skills/rolldown-repl-decode/SKILL.md
+++ b/.claude/skills/rolldown-repl-decode/SKILL.md
@@ -1,0 +1,57 @@
+---
+name: rolldown-repl-decode
+description: Use whenever a rolldown REPL share URL appears in the conversation — `repl.rolldown.rs/#…`, `rolldown-repl.netlify.app/#…`, or any URL whose hash is a base64-encoded rolldown REPL payload. This commonly happens when investigating GitHub issues in rolldown/rolldown that include a "minimal reproduction" REPL link. Decodes the hash into the actual source files + rolldown version so the issue can be reproduced or understood. Trigger even if the user only pastes the URL without explicitly asking to decode it.
+---
+
+# Decode a rolldown REPL share URL
+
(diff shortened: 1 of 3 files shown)


h-a-n-a (approved):

IWANABETHATGUY (approved):

Dunqing:
image
It works well in my quick test.

shulaoda (approved):

PR #9260, opened 2026-04-29
chore: upgrade oxc to 0.128.0

shulaoda:
## Summary

- Bumps `@oxc-project/*` npm packages and `oxc_*` Rust crates from `0.127.0` to `0.128.0`. `oxc_resolver` / `oxc_resolver_napi` stay at `11.19.1`.
- Adapts to the breaking change in `oxc_ast`: `ArrayExpressionElement::Elision` now wraps `Box<'a, Elision>`, so `binding_pattern_ext.rs` switches from `builder.elision(SPAN)` to `builder.alloc_elision(SPAN)`.
- Regenerates `embedded_helpers.rs` and refreshes esbuild/rolldown snapshots that change due to more aggressive DCE / function inlining in the new oxc minifier.
- Adds `rollup@form@namespace-optimization` to `ignored-passed-snapshot-different-tests.js` (rolldown now folds `a()` into `void 0`, matching the existing `namespace-optimization-b` skip).

## Test plan

- [x] `cargo check`
- [x] `just update-generated-code`
- [x] `just test-update`
- [x] `just ued`
- [x] `just roll`


h-a-n-a (approved):

crates/rolldown/tests/rolldown/optimization/inline_const/multi_pass/artifacts.snap
```diff
@@ -46,19 +46,13 @@ demo();
 ### pagea.js
 
 ```js
-//#region pagea.js
-function demo() {}
-demo();
 //#endregion
 
 ```
```
shulaoda, line 51: Here, the module should be removed, and the `//#region pagea.js` comment attached to this method has also been removed.

crates/rolldown/tests/rolldown/issues/5139/artifacts.snap
```diff
@@ -11,18 +11,16 @@ source: crates/rolldown_testing/src/integration_test.rs
...
 //#endregion
 //#region main.js
 var Foo$1 = class {
 	static {
-		__name$1(this, "Foo");
+		__name(this, "Foo");
 	}
 };
 function foo$1() {}
-__name$1(foo$1, "foo");
```
IWANABETHATGUY, line 24: I think we should preserve this test case. This test is used to ensure the correct deconfliction of the runtime helper function.

IWANABETHATGUY: Although the DCE is correct.

IWANABETHATGUY (approved):

PR #9262, opened 2026-04-29
fix(error): remove severity prefix from diagnostic messages

### LINKED ISSUE #8640: [Bug]: changing the severity to `error` in `onLog` does not cause the build to throw

We're migrating to Rolldown and Vite currently. We have strict policies regarding circular dependencies and previously relied on plugins for Webpack to throw errors and stop builds when they are detected. I found that while Rolldown [has](https://rolldown.rs/reference/InputOptions.checks#circulardependency) ability to emit console logs (requested [here](https://github.com/rolldown/rolldown/issues/831)) for them it doesn't seem to be able to emit errors. 
Plugins for Rollup and Vite can cover this, but they are finding circular dependencies where we don't see them both with Rolldown and with Webpack plugin, so there is worry that dependency resolve algos are too different.

Would be very nice to have ability to emit errors natively without overhead from plugins.

I propose API akin to eslint rules - default action on boolean (silent or warning) and more detailed when exact value passed. So current behavior wouldn't break.

```
checks: {
  circularDependency: undefined | boolean | "none" | "warning" | "error",
},
```

On "Error" after all circular dependencies found (it looks like warning thrown only after whole bundle checked) error is thrown, breaking build, in console with list of all places where they are found and dependency chains. Example in Webpack `circular-dependency-plugin` , rollup `rollup-plugin-circular-dependencies` and Vite `vite-plugin-circular-dependency`.

Kyujenius:
Use ariadne's `with_severity_prefix(false)` so that diagnostic messages no longer contain a hardcoded "Warning:" or "Error:" prefix. This ensures `log.message` in `onLog` callbacks reflects the actual severity level passed to `defaultHandler`, aligning with Rollup's behavior.

Since `console.warn`/`console.error` already convey severity, the prefix is redundant for console output. If a visible text prefix is desired, `defaultPrintLog` could prepend one based on `level` in a follow-up.

Fixes #8640.

### Before (main)
스크린샷 2026-04-29 오후 9 48 02

### After (fix/onlog-defaulthandler-error)
after


IWANABETHATGUY:
Can you give me a real-world example (diagnostic with color)? I need to inspect how it behaves differently before and after this pr.

Kyujenius:
Updated the PR description with before/after screenshots showing the colored diagnostic output.

sapphi-red (approved):

shulaoda (approved):

shulaoda:
Thank you for the contribution!

PR #9270, opened 2026-04-29
fix: implement dynamic dominator merge logic

### LINKED ISSUE #9269: [Bug]: Excessive automatic code splitting when importing a module both statically and dynamically

Consider these two examples:
- [REPL A](https://repl.rolldown.rs/#eNp9UbFugzAQ/ZWTlyQSgixdyNi5U8a6g8EHcWvOFpg2LeLfe9iFKFVVycL43Xvvnu4m0YhyEsr7/HVY/kiU6ysTNT9M510fQIq8GC6qR80lKU6SakeDs5hb1+7lopHiwHDiM8J8o7FSfRQcTuyHogz9iHMmNqu15w1Ibe/dUzU1EIt8c970N+S/3L9j/8hiPvbtnbXafVDOrMa0edj8/6jc9ZlAY2MIH2MZZmh618Fule0kScJrpDJRjTbem2A/SQIw5MdQQrLMOwwqj1AGRQHhgtAYiwMMaLEOqKH6jGg1huAI+KCqLxBUtXQDcGOIftEboHYaz96aEAy1GwrQGXpyerR4Nl9YwsPxmK2ltnejH0p4noBUx0UpDGm8SpFBwIG9i/guYH5Jmnm54ocjOx8Mj1vSfEh786p+Uy3yKhwPNU32DkszjdGk0OiR3ak2yKvjxOzLR8zs9M40q5YMYv4GiaLszw==) (working as expected)
- [REPL B](https://repl.rolldown.rs/#eNp9Ub1OwzAQfpWTF0CKki4sYWRm6ogZ3PiSGpyzFTvQEuXdudhtqkKFFMW57+77yXkSragngRSHY/kelm8S9aUuRMNl4yg4i6V13b3MTSkeniSZ3rshMlZWynsmLDCTUNRxGHEuRIbPuqcqq2YyLOSwVwPqxGfVazvm/DELRuNODVeGrbJhcVy1zqYX4Nbf5G52EAt9lV75F+S/4L9zn2gpIOsOzlrtvqjkqdZ0ZVz1b3SufCbQ2BrC59SGGdrB9XB3pt1JkoSHNMqDarTpXAn3kyQAQ36MNWTJsseoygQVUFUQ9witsRggoMUmoobdMaG7MUZHwA+qZg9R7RY3ADfGpJe0ARqnceutidFQt6IAvaEXp0eLW/ONNTxuNsW51Q1u9KGG1wlI9dyUwpDGgxQFRAysXaW6gvktc+blSC+O7Hw0vG5J80O+N6+aD9UhX4XjpebNXmF5pymaFBo9sjo1BvnqODHr8iNmVvrkMauWDGL+AVsMCkg=) (unexpected behavior)

They both contain these modules:
```js
// app.js
import "./shared.js";
console.log("app");
import("./sidebar.js");
```
```js
// shared.js
console.log("shared");
```
```js
// sidebar.js
import "./shared.js"
console.log("sidebar");
```

In addition, REPL B contains this module:
```js
// entry.js
console.log("entry");
import("./app.js");
```

The entry point of REPL A is `app.js`, while the entry point of REPL B is `entry.js`.

### What is expected?

Rolldown bundles REPL A with two chunks:
- `app`, which contains `app.js` and `shared.js`
- `sidebar`, which contains `sidebar.js`

`app` has a dynamic import for `sidebar` (for the `app.js`->`sidebar.js` dynamic import), and `sidebar` has a static import for `app` (for the `sidebar.js`->`shared.js` static import). Rolldown correctly detects that it does not need to split `shared.js` into its own chunk, because the only way `sidebar.js` is ever loaded is through `app.js`, so it's always `app.js` causing first `shared.js` to be loaded.

**I would expect** REPL B to be similar, given that the only difference is that `app.js` is not directly the entrypoint but it's dynamically imported by `entry.js`:
- `entry`, which contains `entry.js`, and dynamically imports `app`
- `app`, which contains `app.js` and `shared.js`, and dynamically imports `sidebar`
- `sidebar`, which contains `sidebar.js`, and statically imports `app`

Rolldown **should not** extract `shared.js` into its own chunk for the same reason as above.

In REPL B, Rolldown is extracting `shared.js` into its own chunk:
- `entry`, which contains `entry.js`, and dynamically imports `app`
- `app`, which contains `app.js`, statically imports `shared`, and dynamically imports `sidebar`
- `shared`, which contains `shared.js`
- `sidebar`, which contains `sidebar.js`, and statically imports `app`

### System Info

```Shell
This `envinfo` command doesn't print the rolldown version, but it would be 1.0.0-rc.16.
```

### Any additional comments?

I noticed this because Vite 8 (Rolldown) is producing many more chunks than Vite 7 (Rollup), going from about 50 to about 80.

TheAlexLichter:
A bit of context:

## How chunk splitting works 

(To anyone swinging by here, skip this if you already know how the internal workflow is)

1. Each user entry and each dynamic-import target gets an entry index. A module's bits bitset records which entries reach it via static imports (dynamic imports don't propagate bits as they're separate roots).
2. Modules with a unique bitset would normally each form their own chunk. The `chunk_optimization` pass then tries to merge common modules into an existing entry chunk via `try_insert_into_existing_chunk` — picking a "merge target" that's guaranteed to be loaded before any of the entries needing the shared module.

## What is broken right now

When a module is shared between a user entry and one of its descendant dynamic chunks, things work as expected. The optimizer picks the user entry as the merge target.

But it does not handle the case when there is a shared module between **two dynamic chunks** where one imports the other one dynamically.

In the broken example of https://github.com/rolldown/rolldown/issues/9269, `app_chunk` imports `sidebar_chunk` via a dynamic import, so `app_chunk` would be the expected merge target (it always loaded before sidebar runs), but `find_merge_target` only inspects static importer sets and so finds nothing.

## How this PR fixes it

The PR adds a dominator check to `find_merge_target` to allow dynamic importers to be considered as merge targets, as long as they "dominate" the entry chunks that need the shared module, meaning they are guaranteed to be loaded before those chunks. This allows the optimizer to correctly merge shared modules into the appropriate dynamic chunk, even when there are dynamic import relationships between chunks. If no dominator merge target is found, it falls back to the existing logic and emit a separate common chunk.

---

Resolves #9269

### DIFF
--- a/crates/rolldown/tests/rolldown/optimization/chunk_merging/dynamic_dominator_with_exports_no_merge/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/optimization/chunk_merging/dynamic_dominator_with_exports_no_merge/artifacts.snap
@@ -1,0 +1,43 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+---
+# Assets
+
+## dynamic1.js
+
+```js
+import "./sharedDynamic.js";
+//#region dynamic1.js
+console.log(true);
+const promise = import("./dynamic2.js");
+//#endregion
+export { promise };
+
+```
+
+## dynamic2.js
+
+```js
+import { t as sharedDynamic } from "./sharedDynamic.js";
+export { sharedDynamic };
+
+```
+
+## main.js
+
+```js
+//#region main.js
+import("./dynamic1.js").then(({ promise, ...ns }) => ns);
+//#endregion
+
+```
+
+## sharedDynamic.js
+
+```js
+//#region sharedDynamic.js
+const sharedDynamic = true;
+//#endregion
+export { sharedDynamic as t };
+
+```
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -597,6 +597,7 @@ impl GenerateStage<'_> {
             module_table,
             dynamic_entry_to_dynamic_importers,
             entry_chunk_reference,
+            &info.modules,
           )
         },
       )
@@ -749,18 +750,38 @@ impl GenerateStage<'_> {
   /// gain `D`'s chunk as a static dependency, no load can reach them without
   /// `D` already being loaded.
   ///
-  /// The leak guard rejects `D` when any user-defined entry reaches one of the
-  /// covered dynamic entries without also reaching `D` — after the merge that
-  /// other path would pull `D`'s chunk in as a static dep and run `D`'s side
-  /// effects on a load that previously did not touch `D`.
+  /// Two extra guards:
+  /// - **Export-pollution guard:** rejects the merge when any pending module
+  ///   has named exports. Those exports would still be needed cross-chunk by
+  ///   the other dynamic entries, forcing `D`'s chunk file to expose them at
+  ///   the file level — which is what `import('./D.js')` resolves to at
+  ///   runtime, polluting the dynamic-entry namespace observed by callers.
+  /// - **Side-effect leak guard:** rejects `D` when any user-defined entry
+  ///   reaches one of the covered dynamic entries without also reaching `D` —
+  ///   after the merge that other path would pull `D`'s chunk in as a static
+  ///   dep and run `D`'s side effects on a load that previously did not
+  ///   touch `D`.
   fn find_dynamic_dominator(
     dynamic_entry: &[ChunkIdx],
     dynamic_entry_modules: &[ModuleIdx],
     module_table: &ModuleTable,
     dynamic_entry_to_dynamic_importers: &FxHashMap<ModuleIdx, FxHashSet<ModuleIdx>>,
     entry_chunk_reference: &FxHashMap<ChunkIdx, FxHashSet<ChunkIdx>>,
+    pending_modules: &[ModuleIdx],
   ) -> Option<ChunkIdx> {
     if dynamic_entry.len() < 2 {
+      return None;
+    }
+    // Refuse the merge whenever any pending module exposes named exports. The
+    // other dynamic-entry chunks in `chunk_idxs` still need those exports
+    // cross-chunk, so after the merge the dominator's chunk would have to add
+    // them to its file-level export list — and that list is what
+    // `import('./dominator.js')` resolves to at runtime, polluting the
+    // dynamic-entry namespace observed by callers.
+    let exposes_exports = pending_modules
+      .iter()
+      .any(|m| module_table[*m].as_normal().is_some_and(|n| !n.named_exports.is_empty()));
+    if exposes_exports {
       return None;
     }
     dynamic_entry.iter().zip(dynamic_entry_modules.iter()).find_map(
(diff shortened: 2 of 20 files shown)


crates/rolldown/tests/rolldown/optimization/chunk_merging/dynamic_dominator_with_exports_no_merge/artifacts.snap
```diff
@@ -0,0 +1,43 @@
...
+
+## sharedDynamic.js
+
+```js
+//#region sharedDynamic.js
+const sharedDynamic = true;
+//#endregion
+export { sharedDynamic as t };
+
+```
```
IWANABETHATGUY, line 43: The output of this test case still seems suboptimal. At least, it is safe to merge dynamic2.js and sharedDynamic.js. Moreover, rollup generated only two chunks. https://rollupjs.org/repl/?version=4.60.2&shareable=eyJleGFtcGxlIjpudWxsLCJtb2R1bGVzIjpbeyJjb2RlIjoiaW1wb3J0KCcuL2R5bmFtaWMxLmpzJykudGhlbigoeyBwcm9taXNlLCAuLi5ucyB9KSA9PiBucyk7IiwiaXNFbnRyeSI6dHJ1ZSwibmFtZSI6Im1haW4uanMifSx7ImNvZGUiOiJpbXBvcnQgeyBzaGFyZWREeW5hbWljIH0gZnJvbSAnLi9zaGFyZWREeW5hbWljLmpzJztcblxuY29uc29sZS5sb2coc2hhcmVkRHluYW1pYyk7XG5leHBvcnQgY29uc3QgcHJvbWlzZSA9IGltcG9ydCgnLi9keW5hbWljMi5qcycpOyIsImlzRW50cnkiOmZhbHNlLCJuYW1lIjoiZHluYW1pYzEuanMifSx7ImNvZGUiOiJleHBvcnQgY29uc3Qgc2hhcmVkRHluYW1pYyA9IHRydWU7IiwiaXNFbnRyeSI6ZmFsc2UsIm5hbWUiOiJzaGFyZWREeW5hbWljLmpzIn0seyJjb2RlIjoiZXhwb3J0IHsgc2hhcmVkRHluYW1pYyB9IGZyb20gJy4vc2hhcmVkRHluYW1pYy5qcyc7IiwiaXNFbnRyeSI6ZmFsc2UsIm5hbWUiOiJkeW5hbWljMi5qcyJ9XSwib3B0aW9ucyI6e319

IWANABETHATGUY:
Although some test cases still need improvement, I don't see any blockers to merging this PR.

IWANABETHATGUY (approved):

crates/rolldown/tests/rolldown/optimization/chunk_merging/dynamic_dominator_with_exports_no_merge/artifacts.snap
TheAlexLichter: Rollup generates two chunks because it has a heuristic for inlining dynamic imports. But you are right, the test case can be improved. let me open up a follow-up PR to iterate on (so this one can be merged already).

PR #9289, opened 2026-05-03
fix: sanitize preserveModules chunk names from query ids

### LINKED ISSUE #8761: [Bug]: sanitizeFileName is ignored for JS chunks when preserveModules: true with object-form entry

deltama37:
## Summary

Fixes #8761.

When `output.preserveModules` is enabled, Rolldown derives chunk names from module ids. For query-bearing module ids such as Vue SFC submodules:

```text
components/TestComp.vue?vue&type=script&setup=true&lang
```

the generated JS chunk name could keep `?`, `&`, and `=` in the output filename when using object-form entries.

This change applies the existing `sanitizeFileName` logic after the preserveModules chunk name has been assembled, so the emitted filename becomes:

```text
components/TestComp.vue_vue_type_script_setup_true_lang.js
```

## Test

Added a Node fixture covering a preserveModules build with an object-form entry and a Vue-style query module id.

```sh
just test-node-rolldown-only -t topics/preserve-modules/sanitize-filename-query
```


sapphi-red (changes requested):
Please add the tests from https://github.com/rolldown/rolldown/pull/8817 and address the comments there

PR #9290, opened 2026-05-04
feat: re-use dynamic entry if dominator merge bails

TheAlexLichter:
Improvement of https://github.com/rolldown/rolldown/pull/9270

In case the dominator merge bails due to export pollution, the algorithm looks for a dynamic entry that already re-exports every named export of the pending modules and uses it as the merge target.

This is a trade-off between lazy-loading granularity and having fewer chunks.


IWANABETHATGUY:
I saw you plan to merge the pr into main directly, so should we close https://github.com/rolldown/rolldown/pull/9270

TheAlexLichter:
Closed in favor of https://github.com/rolldown/rolldown/pull/9270

PR #9298, opened 2026-05-05
docs(lazy-barrel): clarify own-exports behavior for import-then-export records

shulaoda:
Related to #8895

## Summary

Updates `docs/in-depth/lazy-barrel-optimization.md` to reflect the behavior introduced by #8895.

The "Own exports (non-pure re-export barrels)" section did not previously cover the import-then-export pattern (`import { a } from './a'; export { a }`). After #8895, the loader distinguishes:

- **Dedicated re-export records** (`export { .. } from '..'`, `export * as ns from '..'`): when the barrel must execute, these can still fall back to an empty specifier set if their binding is not requested by `main.js`.
- **Shared import records** (produced by `import { a } from '..'; export { a }`): always keep their full specifiers, because Rolldown cannot statically determine whether the barrel's own code references the binding.

## Changes

- Extended the example in "Own exports" to include an `import { e } from './e'; export { e }` case.
- Added a bullet describing how `e.js` is loaded with `e` requested.
- Added a paragraph explicitly drawing the distinction between dedicated re-export records and shared import records, noting the precondition (barrel must execute because its own exports are loaded by main.js).

### DIFF
--- a/docs/in-depth/lazy-barrel-optimization.md
+++ b/docs/in-depth/lazy-barrel-optimization.md
@@ -213,8 +213,10 @@ When a barrel module has its own exports (not just re-exports), all its import r
 // barrel/index.js
 import './a';
 import { b } from './b';
+import { e } from './e';
 export { c } from './c';
 export { d } from './d';
+export { e };
 
 console.log(b);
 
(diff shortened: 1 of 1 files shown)


hyf0 (approved):

hyfdev:
### Merge activity

* **May 6, 6:58 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **May 6, 6:58 AM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **May 6, 7:00 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9317, opened 2026-05-07
chore: migrate `@rolldown/pluginutils` to `rolldown/plugins`

### LINKED ISSUE #8653: Separate `@rolldown/pluginutils` release from Rolldown's release
`@rolldown/pluginutils` is now released together with Rolldown's release. But this isn't needed. We should be able to make untangled.

Currently, it causes many duplicated packages in `node_modules` because the deduping doesn't work.

We can move this package to https://github.com/rolldown/plugins if that's easier to handle.

shulaoda:
## PR description

Closes #8653.

`@rolldown/pluginutils` has been moved to https://github.com/rolldown/plugins so it can be released independently from `rolldown`. Companion PR: rolldown/plugins#61.

### Why

Previously `@rolldown/pluginutils` shared rolldown's release cadence — every rolldown publish (`1.0.0-beta.53` … `1.0.0-rc.18` … `1.0.0`) produced a new pluginutils version. Because pnpm rewrites `workspace:*` to an **exact** version on publish and prerelease semver does not match across pre-releases (`^1.0.0-rc.16` does **not** match `1.0.0-rc.17`), every plugin pinned to an older rc forced an extra copy into `node_modules`. Users typically saw 3–4 copies of `@rolldown/pluginutils` resolved side by side.

Decoupling the release cadence and depending on `^1.0.0` lets pnpm/npm dedupe across rolldown and any plugin that depends on pluginutils.

### Changes

- `packages/rolldown/package.json`: dependency on `@rolldown/pluginutils` switched from `workspace:*` to `^1.0.0`.
- `packages/pluginutils/`: removed (source now lives in `rolldown/plugins`).
- `Justfile`: dropped the `build-pluginutils` target and removed it as a prereq from `build`, `build-rolldown`, `build-rolldown-wasi`, `build-rolldown-release`, `build-browser`, `build-browser-release`.
- `.github/workflows/`:
  - `ci.yml`: removed the `pluginutils-changes` paths-filter and the `pluginutils-test` job; removed `Build @rolldown/pluginutils` steps in `node-validation` and `docs-validation`.
  - `publish-to-npm.yml`: removed the pluginutils artifact download.
  - `reusable-release-build.yml`: removed the `build-rolldown-pluginutils` job and the dependent download steps in `build-node-packages` / `build-rolldown-browser`.
  - `reusable-native-build.yml`: dropped `packages/pluginutils/dist` from the artifact upload path.
  - `publish-to-pkg.pr.new.yml`: removed the pluginutils artifact download and dropped `./packages/pluginutils` from the `pkg-pr-new` publish list.
- `scripts/misc/published-package-check.mjs`: removed `packages/pluginutils` from `packagesNeedToPublish`.


packages/rolldown/package.json
```diff
@@ -120,7 +120,7 @@
   },
   "dependencies": {
     "@oxc-project/types": "catalog:",
-    "@rolldown/pluginutils": "workspace:*"
+    "@rolldown/pluginutils": "1.0.0"
```
shulaoda: Should we pin the exact version or use `^1.0.0`? If we use `^1.0.0`, we require strict adherence to semantic versioning from users. I suggest using `1.0.0` directly, since we likely will not release upgrades very frequently.

sapphi-red: I think we should follow semver for `@rolldown/pluginutils` regardless of whether we use `^1.0.0` or `1.0.0` here. Thus, I think we can use `^1.0.0`.

sapphi-red (approved):

sapphi-red:
cc @rolldown/rolldown so that you are aware

PR #9322, opened 2026-05-07
fix(chunk-optimizer): refuse asymmetric merge for cyclic dynamic entries (#9320)

### LINKED ISSUE #9320: [Bug]: namespace pollution when two dynamic-import targets form a static-import cycle

Standalone repo: https://github.com/aminpaks/vite8-cycle-import-bug

```bash
git clone https://github.com/aminpaks/vite8-cycle-import-bug
cd vite8-cycle-import-bug
pnpm install
pnpm run all
```

The repo contains two builds of the same 4-file app: one with vite 6 (rollup) and one with vite 8.0.10 (rolldown rc.17). The repro script then executes each bundle in node and checks the namespaces returned by the two dynamic imports against the modules' declared exports. Vite 6 ✅ runs; vite 8 💥 crashes with exit code 1.

Minimal source:

```ts
// src/form.ts — dynamic-import target #1
import {actionImpl} from './action.ts';
export function formImpl() { return 'form-result'; }
export function callActionFromForm() { return actionImpl(); }

// src/action.ts — dynamic-import target #2; cycles with form.ts
import {formImpl} from './form.ts';
export function actionImpl() { return 'action-result'; }
export function callFormFromAction() { return formImpl(); }

// src/main.ts — entry; both modules are dynamic-import targets
const formNs   = await import('./form.ts');
const actionNs = await import('./action.ts');
// app validates each namespace exports exactly its declared keys
```

### What is expected?

`Object.keys(actionNs)` returns `['actionImpl', 'callFormFromAction']` — the two exports declared by `action.ts`.

This is what vite 6 / rollup produces (and what the ES module dynamic-import contract requires).

Vite 8 / rolldown produces:

```
Object.keys(actionNs) === ['actionImpl', 'callFormFromAction', 'n', 't']
```

`n` and `t` are `formImpl` and `callActionFromForm` from `form.ts`, renamed and leaked into `action`'s namespace.

Inspecting `dist-v8/assets/action-<hash>.js`:

```js
export { actionImpl, callFormFromAction, formImpl as n, callActionFromForm as t };
//                                       ^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^^^^
//                                       form.ts's exports leaked into action's chunk
```

Both `form.ts` and `action.ts` are merged into a single chunk (cycle prevention), and rolldown names that chunk after `action.ts`'s basename (`action-<hash>.js`). The chunk's top-level exports become the union of both members' symbols, so `import('./action')` resolves to a polluted namespace.

`form.ts` happens to be safe because rolldown emits a small re-export facade for it (`form-<hash>.js`) that only re-exports form's own symbols. Which member is "polluted" vs "facaded" is determined by which basename rolldown picked for the merged chunk — both are at risk in general.

App code that iterates the namespace and validates the contract (e.g. `Object.entries(mod).forEach(([k, v]) => { if (!isExpected(v)) throw … })`, a common module-as-registry pattern) crashes:

```
💥 App crashed with uncaught error:
Error: action module contract violated: expected exports [actionImpl, callFormFromAction], got [actionImpl, callFormFromAction, n, t]
    at validate (file:///.../dist-v8/assets/main-<hash>.js:68:107)
    at main   (file:///.../dist-v8/assets/main-<hash>.js:76:2)
```

### System Info

```Shell
System:
    OS: macOS 26.3.1
    CPU: (11) arm64 Apple M3 Pro
    Memory: 1.78 GB / 36.00 GB
    Shell: 5.9 - /bin/zsh
Binaries:
    Node: 22.22.0
    pnpm: 10.17.0
npmPackages:
    vite (vite8 alias): 8.0.10
    vite (vite6 alias): 6.4.2
    rolldown (transitive, via vite 8): 1.0.0-rc.17
```

Related closed issues that are NOT the same bug but live in the same area:
- #9093 — circular-dependency detection blocks too many merges
- #8989 / #9161 — `__exportAll is not a function` runtime errors from circular imports involving entry chunks
- #9269 — excessive automatic code splitting

aminpaks:
## Summary

Fixes #9320. Two dynamic-import targets with a static-import cycle leak each other's named exports into their dynamic-import namespaces — `Object.keys(actionNs)` returned `['actionImpl', 'callFormFromAction', 'n', 't']` instead of `['actionImpl', 'callFormFromAction']`.

Root cause: when both modules form a cycle, `find_merge_target` (in `chunk_optimizer.rs`) finds multiple candidates that satisfy its "every other entry imports me" rule. Picking one is arbitrary and merges the other entry's module into the chosen target's chunk file — which is what `import('./<target>.js')` resolves to at runtime — so the merged-in entry's named exports become observable on the wrong namespace.

- Add a guard before `find_merge_target` that detects the cyclic shape: `>=2` of the candidate dynamic-entry chunks' own entry modules with named exports sit in the pending common-chunk's `info.modules`. When that holds, refuse the merge and let `assign_modules_to_chunk` fall back to a separate common chunk so each dynamic entry stays a facade exposing only its own exports.
- Length-check fast path skips the scan for the common single-entry case (≈1 ns); cycle-detection short-circuits at the second collision; no `HashSet` allocation (`dynamic_chunk_entry_modules` is a small `Vec`, linear `contains` is faster).
- Companion guard inside `find_dynamic_dominator` (covering the disjoint shared-module case) is unchanged.
- Adds regression test at `crates/rolldown/tests/rolldown/issues/9320/` with `_test.mjs` asserting both namespaces' `Object.keys`.

No other snapshots changed. Reverting the fix makes the new test fail with the exact bug output from the issue.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -18,7 +18,8 @@ use crate::{
     IncludeContext, SymbolIncludeReason, include_runtime_symbol, include_symbol,
   },
   types::linking_metadata::{
-    LinkingMetadata, included_info_to_linking_metadata_vec, linking_metadata_vec_to_included_info,
+    LinkingMetadata, LinkingMetadataVec, included_info_to_linking_metadata_vec,
+    linking_metadata_vec_to_included_info,
   },
 };
 
@@ -397,11 +398,10 @@ impl GenerateStage<'_> {
         }
         let chunk_idxs: Vec<_> = bits.index_of_one().map(ChunkIdx::from_raw).collect();
 
-        let merge_target = Self::try_insert_into_existing_chunk(
+        let merge_target = self.try_insert_into_existing_chunk(
           &chunk_idxs,
           &static_entry_chunk_reference,
           chunk_graph,
-          &self.link_output.module_table,
           &dynamic_entry_to_dynamic_importers,
           temp_chunk,
           temp_chunk_graph,
@@ -558,14 +558,16 @@ impl GenerateStage<'_> {
   ///
   /// Returns `Some(ChunkIdx)` if a suitable merge target is found, `None` otherwise.
   fn try_insert_into_existing_chunk(
+    &self,
     chunk_idxs: &[ChunkIdx],
     entry_chunk_reference: &FxHashMap<ChunkIdx, FxHashSet<ChunkIdx>>,
     chunk_graph: &ChunkGraph,
-    module_table: &ModuleTable,
     dynamic_entry_to_dynamic_importers: &FxHashMap<ModuleIdx, FxHashSet<ModuleIdx>>,
     info: &ChunkCandidate,
     temp_chunk_graph: &ChunkOptimizationGraph,
   ) -> Option<ChunkIdx> {
+    let module_table = &self.link_output.module_table;
+    let linking_infos = &self.link_output.metas;
     let mut user_defined_entry = vec![];
     let mut dynamic_entry = vec![];
     for &idx in chunk_idxs {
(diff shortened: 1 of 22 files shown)


TheAlexLichter:
Pushed a change so that the export-pollution guard uses the actual linked export surface instead of the syntactical `named_exports`.

This should fix cyclic dynamic entries that export only via `export * from` as shown in the `9320_star` regression fixture.

IWANABETHATGUY:
Found a test case that is still broken. I will tweak the code a little. The rest looks good to me.

aminpaks:
@IWANABETHATGUY, is there anything I can help with to land this?

IWANABETHATGUY (approved):

PR #9329, opened 2026-05-08
fix: Fix same-target common chunk merge cycle check

schiller-manuel:
# Fix common chunk merge false positive for same-target merges

Fixes a chunk optimizer case where multiple pending common chunks that should all merge into the same entry could block each other via a transient circular-dependency check.

This showed up with React's CJS `react/jsx-runtime` facade:

```js
module.exports = require('./cjs/react-jsx-runtime.production.js')
```

In a graph with one user entry and nested dynamic imports, Rolldown could emit a separate `jsx-runtime` common chunk even though all consumers are reachable through the same entry. The optimizer now treats other valid same-target pending merges as internal to the target when checking for cycles.

created using GPT 5.5

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -192,49 +192,45 @@ impl ChunkOptimizationGraph {
     }
   }
 
-  /// Checks if merging `source_chunk` into `target_chunk` would create a circular dependency.
-  ///
-  /// Returns `true` if the merge would create a cycle, `false` if it's safe to merge.
-  ///
-  /// After merging, the combined chunk has deps = `source.deps ∪ target.deps`, and any chunk
-  /// that previously depended on `source` now depends on `target`. A cycle exists iff `target`
-  /// is reachable from `source.deps ∪ target.deps` in the resulting graph.
-  ///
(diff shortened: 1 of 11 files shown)


IWANABETHATGUY:
Please file an issue first

IWANABETHATGUY:
Thanks for your contribution. 
The issue is already fixed in https://github.com/rolldown/rolldown/pull/9305.I prefer 9305 since it is a more general solution rather than patching the existing code.

schiller-manuel:
@IWANABETHATGUY thanks for working on this.

BTW I read the contribution guideline at https://rolldown.rs/contribution-guide/#submitting-a-pull-request before creating the PR, it wasn't obvious to me that I should have created an issue first.

IWANABETHATGUY:
> @IWANABETHATGUY thanks for working on this.
> 
> BTW I read the contribution guideline at https://rolldown.rs/contribution-guide/#submitting-a-pull-request before creating the PR, it wasn't obvious to me that I should have created an issue first.

Yeah, you are right. Creating an issue and discussing it with a team member can help avoid missing some important context, which may improve the AI-generated code. Sometimes there are multiple ways to solve the same issue, and AI usually chooses the easiest, but it may not be the correct one.

schiller-manuel:
will do in future!

PR #9356, opened 2026-05-11
fix: stabilize chunk content hashes

### LINKED ISSUE #9339: [Bug]: Non-deterministic hash when entry/chunkFileNames are used

Hello, I noticed that when using `codeSplitting` and `entryFileNames`/`chunkFileNames` that content hashes change even though the files are byte-for-byte identical.

I've created a minimal reproduction here: https://github.com/ryanto/rolldown-hash-issue. The reproduction builds two apps, where the 2nd build has an extra isolated entry point. The extra entry point causes all hashes in the build to change, even if their content is unchanged. My understanding is that the hash is based off a files content.

I've checked two builds into this repo so that you can see their output.

The rolldown runtime file gets a different content hash across the two builds, even though its content is identical. From there, every file gets a different hash because of the new runtime. What's interesting about this is that removing `entryFileNames` and `chunkFileNames` causes the chunk hashes to stabilize. 

PS: I'm a big fan of rolldown, thank you for this project :)

### What is expected?

I would expect the hashes to be based on file content. If the file content is the same across builds then I do not expect the `[hash]` identifier to change.

The hash is changing across builds that use `entryFileNames` and `chunkFileNames`.

### System Info

```Shell
System:
    OS: macOS 15.7.3
    CPU: (10) arm64 Apple M1 Max
    Memory: 448.30 MB / 32.00 GB
    Shell: 3.4.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 24.9.0 - /Users/ryan/.nodenv/versions/24.9.0/bin/node
    npm: 11.6.0 - /Users/ryan/.nodenv/versions/24.9.0/bin/npm
    pnpm: 10.33.2 - /Users/ryan/Library/pnpm/pnpm
  Browsers:
    Brave Browser: 146.1.88.127
    Chrome: 144.0.7559.60
    Firefox: 144.0.2
    Safari: 18.6
    Safari Technology Preview: 26.0
  npmPackages:
    rolldown: ^1.0.0 => 1.0.0
```

### Any additional comments?

I think this might be causing the issue since preliminary file name might be what's different across the builds: https://github.com/rolldown/rolldown/blob/main/crates/rolldown/src/utils/chunk/finalize_chunks.rs#L101

DaZuiZui:
## Summary

- stabilize chunk content hashes by removing preliminary filenames from the hash input
- keep transitive chunk content and `augmentChunkHash` contributions in the hash calculation
- add a regression test for adding an isolated entry without changing an existing entry hash
- update affected hash snapshots

Fixes #9339.

## Tests

- `just test`

## AI Usage

This PR was prepared with assistance from OpenAI Codex. I reviewed the issue, implementation, snapshots, and test results before submitting.


hyfdev:
Thanks for your effort. I added you as the co-author of https://github.com/rolldown/rolldown/pull/9444, which fixes https://github.com/rolldown/rolldown/issues/9339.

#9444 especially removed the code related source map debug id, because I think it's not relevant to the real cause. If you think it's valuable, please raise another PR and explain your thougths.

PR #9381, opened 2026-05-13
feat: emit warnings for invalid pure annotations

### LINKED ISSUE #8898: [Feature Request]: Emit warnings for invalid pure annotations

It's good to have warnings in some scenario where pure annotations are not able 
to be interpreted due to the position of the comment.

For example, pure annotation in the example down below is not able 
to be interpreted due to position and it's ignored by the parser.

```js
/* #__PURE__ */ React.createElement;
```

whereas, rollup would emit this warning to help people adjust the placement of pure annotation.

```
Rollup completed with warnings:
main.js (1:0): A comment "/* #__PURE__ */" in "main.js" contains an annotation that Rollup cannot interpret due to the position of the comment. The comment will be removed to avoid issues. [(link)](https://rollupjs.org/configuration-options/#pure)
/main.js (1:0)
1: /* #__PURE__ */ React.createElement;
```

https://rollupjs.org/repl/?version=4.59.0&shareable=eyJleGFtcGxlIjpudWxsLCJtb2R1bGVzIjpbeyJjb2RlIjoiLyogI19fUFVSRV9fICovIFJlYWN0LmNyZWF0ZUVsZW1lbnQ7IiwiaXNFbnRyeSI6dHJ1ZSwibmFtZSI6Im1haW4uanMifV0sIm9wdGlvbnMiOnt9fQ==

Original issue: https://github.com/oxc-project/oxc/issues/20334

No need

Kyujenius:
Surfaces `/* #__PURE__ */` / `/* @__PURE__ */` annotations that the parser cannot apply, using Rollup's `INVALID_ANNOTATION` log code. The parser metadata comes from oxc-project/oxc#20687 (in `oxc@0.130.0`, already a dependency).

`PreProcessEcmaAst::build` iterates `program.comments` after the semantic-warnings block and pushes one diagnostic per `PureNotApplied` comment. No AST mutation; codegen and tree-shaking unchanged.

Five positive fixtures + one negative (a valid annotation that must produce zero warnings, guarding against false positives) cover the three contexts oxc marks plus the `@__PURE__` variant and a mixed valid/invalid case. Two existing snapshots (`remove_unused_pure_comment_calls`, `property_read_side_effects`) gained the warning because their inputs already contained misplaced annotations — their `# Assets` output is unchanged.

One open question: Rollup also _strips_ invalid annotations from output (per its log text "The comment will be removed to avoid issues"). I left that out to avoid a behavior change during the RC phase; happy to follow up in a separate PR if desired.

Fixes #8898.


crates/rolldown/src/utils/pre_process_ecma_ast.rs
```diff
@@ -87,6 +88,30 @@ impl PreProcessEcmaAst {
...
+    // from applying them (expression-level, statement-level, or variable declarator).
+    // Aligns with Rollup's `INVALID_ANNOTATION` log code.
+    let invalid_pure_spans: Vec<oxc::span::Span> = ast.program.with_dependent(|_owner, dep| {
+      dep
+        .program
+        .comments
+        .iter()
+        .filter(|c| c.content == CommentContent::PureNotApplied)
+        .map(|c| c.span)
+        .collect()
```
IWANABETHATGUY: We could move this into the `with_dependent` closure to avoid the intermediate `Vec<Span>` allocation.

crates/rolldown/tests/rolldown/tree_shaking/property_read_side_effects/artifacts.snap
```diff
@@ -1,6 +1,34 @@
...
 ---
+# warnings
+
+## INVALID_ANNOTATION
+
+```text
+[INVALID_ANNOTATION] A comment "/*#__PURE__*/" in "main.js" contains an annotation that Rolldown cannot interpret due to the position of the comment.
+    ╭─[ main.js:12:1 ]
+    │
+ 12 │ /*#__PURE__*/ test().a.b.c; // Should be tree-shaken when propertyReadSideEffects: false
```
IWANABETHATGUY, line 12: https://rollupjs.org/repl/?version=4.60.3&shareable=eyJleGFtcGxlIjpudWxsLCJtb2R1bGVzIjpbeyJjb2RlIjoiLyojX19QVVJFX18qLyB0ZXN0KCkuYS5iLmMiLCJpc0VudHJ5Ijp0cnVlLCJuYW1lIjoibWFpbi5qcyJ9LHsiY29kZSI6ImV4cG9ydCBjb25zdCBxdXggPSAnUVVYJzsiLCJpc0VudHJ5IjpmYWxzZSwibmFtZSI6InF1eC5qcyJ9XSwib3B0aW9ucyI6eyJvdXRwdXQiOnsiZm9ybWF0IjoiZXMifSwidHJlZXNoYWtlIjp7InByb3BlcnR5UmVhZFNpZGVFZmZlY3RzIjp0cnVlfX19

I doubt the warning is correct. If you set `treeshake.propertyReadSideEffect` to `false`, the statement will be eliminated. If you keep `treeshake.propertyReadSideEffect` set to `false` but remove the pure annotation, the statement will be preserved. Rollup considers the pure annotation valid and marks the global `test` call as sideEffects-free.

Kyujenius: > I doubt the warning is correct. If you set `treeshake.propertyReadSideEffect` to `false`, the statement will be eliminated. If you keep `treeshake.propertyReadSideEffect` set to `false` but remove the pure annotation, the statement will be preserved. Rollup considers the pure annotation valid and marks the global `test` call as sideEffects-free.

You're right. I checked the REPL too, and dug into the oxc source. The two annotations are valid under Rollup's semantics; PURE attaches to the innermost call (`test()`), not the outer member-access chain.

I think the mismatch sits on the oxc side: `CommentContent::PureNotApplied` is set when the AST's top-level expression node isn't a `CallExpression`/`NewExpression`, so chain patterns get flagged even though Rollup and esbuild consider them applied. Note the other incidental drift (the `parens_no` lines in the esbuild fixture) is a legitimate NotApplied case — esbuild treats it the same way, so those stay correct.

How about I move this PR to draft and send an upstream PR on oxc first - refining `PureNotApplied` to skip chained-call patterns? then, false-positive drift here disappears and we can resume

IWANABETHATGUY: Let's file an issue upstream first (of course, it would be better if you could fix it upstream), but I don't think that is a blocker to merging this PR? Since this issue is filed by h-a-n-a, I will wait for @h-a-n-a to decide when to merge this pr.

Kyujenius: Thanks. I'll file the oxc issue tomorrow.

crates/rolldown/tests/esbuild/dce/remove_unused_pure_comment_calls/artifacts.snap
```diff
@@ -1,6 +1,34 @@
 ---
 source: crates/rolldown_testing/src/integration_test.rs
 ---
+# warnings
+
+## INVALID_ANNOTATION
+
+```text
+[INVALID_ANNOTATION] A comment "/* @__PURE__ */" in "entry.js" contains an annotation that Rolldown cannot interpret due to the position of the comment.
```
h-a-n-a, line 9: Would you please add a link to the [Rolldown document](https://rolldown.rs/in-depth/dead-code-elimination#pure)? So that we can tell user on how to correctly use it instead of just telling user this is wrong. [For your reference](https://github.com/rolldown/rolldown/blob/4e353804a3704722e8ba94fa0b7ebf8aae4f5a93/crates/rolldown_error/src/build_diagnostic/events/eval.rs#L47).

crates/rolldown/tests/rolldown/tree_shaking/property_read_side_effects/artifacts.snap
h-a-n-a: I second this. This is not a huge blocker that avoids us from merging this PR.

crates/rolldown/tests/esbuild/dce/remove_unused_pure_comment_calls/artifacts.snap
h-a-n-a: For the document itself, I feel like it's not complete, as it only tells people the use case of function calls, but `new Ctor()` is also implicitly considered as one of them. This is not something you need to do in this PR though. Feel free to file a new PR to do so.

Kyujenius: > Would you please add a link to the [Rolldown document](https://rolldown.rs/in-depth/dead-code-elimination#pure)? So that we can tell user on how to correctly use it instead of just telling user this is wrong. [For your reference](https://github.com/rolldown/rolldown/blob/4e353804a3704722e8ba94fa0b7ebf8aae4f5a93/crates/rolldown_error/src/build_diagnostic/events/eval.rs#L47).

Done - added the `add_help` linked, snapshots regenerated.


> For the document itself, I feel like it's not complete, as it only tells people the use case of function calls, but `new Ctor()` is also implicitly considered as one of them. This is not something you need to do in this PR though. Feel free to file a new PR to do so.

Happy to take the doc improvement as a follow-up. I'll send a separate PR for the `new Ctor()` case (and any other implicit forms I spot). Thanks for the pointer!

h-a-n-a:
Can you resolve the conflicts please?

Kyujenius:
> Can you resolve the conflicts please?

Already on it 😄 rebasing onto the latest main and resolving conflicts now. Will push the updated branch shortly.

h-a-n-a (approved):

PR #9388, opened 2026-05-14
feat: allow checks.* to escalate emissions to hard errors

### LINKED ISSUE #9362: [Feature Request]: Allow unresolved imports to be configured as hard errors

Rolldown currently supports `checks.unresolvedImport` as a boolean that controls whether unresolved import warnings are emitted:

```ts
export default {
  checks: {
    unresolvedImport: true,
  },
};
```

This is useful for suppressing or enabling the warning, but it does not appear to provide a built-in way to promote unresolved bare imports from warnings to hard build errors.

For browser-targeted SDK/library bundles, unresolved bare imports are often not acceptable. If a dependency accidentally fails to resolve, the generated browser bundle can still be emitted with an externalized import and fail later at runtime. In these cases, I want the build to fail immediately.

Today this can be implemented with `onLog`:

```js
export default {
  onLog(level, log, defaultHandler) {
    if (log.code === 'UNRESOLVED_IMPORT') {
      defaultHandler('error', log);
      return;
    }

    defaultHandler(level, log);
  },
};
```

That works, but it is more imperative than the existing declarative `checks` configuration and is easy for users to miss. It also requires each project to know the exact log code and implement a custom handler.

The current behavior makes sense as a default because library builds sometimes intentionally leave dependencies external. However, for self-contained browser artifacts it would be useful to express the desired severity directly in config.

One possible API is to allow check severities:

```ts
export default {
  checks: {
    unresolvedImport: 'error',
  },
};
```

For compatibility, the existing boolean behavior could remain:

```ts
export default {
  checks: {
    unresolvedImport: true,  // emit warning, current behavior
    unresolvedImport: false, // suppress warning, current behavior
  },
};
```

Possible accepted values:

```ts
type CheckSeverity = boolean | 'warning' | 'error';

interface ChecksOptions {
  unresolvedImport?: CheckSeverity;
}
```

If broader consistency is preferred, this could apply to all `checks` entries instead of only `unresolvedImport`:

```ts
export default {
  checks: {
    unresolvedImport: 'error',
    circularDependency: 'warning',
    eval: false,
  },
};
```

Alternative API if changing `checks` is not desirable:

```ts
export default {
  failOnWarnings: ['UNRESOLVED_IMPORT'],
};
```

or:

```ts
export default {
  logSeverity: {
    UNRESOLVED_IMPORT: 'error',
  },
};
```

The main goal is to make this common policy declarative without requiring a custom `onLog` hook.

## Context

In the Rolldown source, `checks.unresolvedImport` is documented as controlling whether the warning is emitted. Unresolved path-like imports already become hard errors, while unresolved bare imports are treated as external dependencies and emitted as warnings. This request is specifically about making that latter behavior configurable for projects that require fully resolved/self-contained browser bundles.

IWANABETHATGUY:
Closed #9362
## Summary

Address #9362 by widening every `checks.*` entry to accept `boolean | 'warn' | 'error'`:

- `false` — suppress the check (unchanged).
- `true` or `'warn'` — emit a warning (unchanged default).
- `'error'` — promote the emission to a hard build error that fails the build.

## Implementation

- New `CheckSetting` enum (`Off | Warn | Error`) in `rolldown_common` with custom serde + schemars impls so the public API stays `boolean | 'warn' | 'error'`.
- Generator emits per-field `match` arms that populate two disjoint bitflags on `NormalizedBundlerOptions`: `warn_checks` and `error_checks`. Added `is_check_enabled(kind)` helper for call sites that ask "is this check on at all?".
- `promote_warnings_to_errors` pass runs at the end of `Bundle::generate`/`write` and moves any warning whose kind is in `error_checks` into the error channel, so the escalation works for every check kind — not just `unresolvedImport`.
- `unresolvedImport` keeps a resolve-time fast-path in `resolve_utils.rs` to fail early and skip the synthetic external `ResolvedId`.

## Test plan

- [x] New fixture `crates/rolldown/tests/rolldown/errors/unresolved_import_as_error/` — `unresolvedImport: 'error'` fails the build with `[UNRESOLVED_IMPORT] Module not found.`
- [x] New fixture `crates/rolldown/tests/rolldown/errors/eval_as_error/` — `eval: 'error'` fails the build with `[EVAL] …`, verifying cross-check escalation through the post-processing pass.
- [x] New fixture `crates/rolldown/tests/rolldown/warnings/unresolved_import_explicit_warn/` — `unresolvedImport: 'warn'` produces output identical to the existing boolean-default fixture.
- [x] Existing `unresolved_import_treated_as_external` snapshot unchanged (default behavior preserved).
- [x] `cargo test -p rolldown --test integration` — 1706 passed, 0 failed.
- [x] `cargo clippy` clean across `rolldown`, `rolldown_common`, `rolldown_binding`, `rolldown_testing`, `rolldown_error`.

### DIFF
--- a/packages/rolldown/src/options/generated/checks-options.ts
+++ b/packages/rolldown/src/options/generated/checks-options.ts
@@ -8,33 +8,49 @@ export interface ChecksOptions {
    * Circular dependencies lead to a bigger bundle size and sometimes cause execution order issues and are better to avoid.
    *
    * {@include ../docs/checks-circular-dependency.md}
+   *
+   * - `false` disables the check.
+   * - `'warn'` emits a warning (default when the check is enabled).
+   * - `'error'` promotes the emission to a hard build error.
    * @default false
    * */
-  circularDependency?: boolean;
+  circularDependency?: false | 'warn' | 'error';
 
   /**
    * Whether to emit warnings when detecting uses of direct `eval`s.
    *
    * See [Avoiding Direct `eval` in Troubleshooting page](https://rolldown.rs/guide/troubleshooting#avoiding-direct-eval) for more details.
-   * @default true
+   *
+   * - `false` disables the check.
+   * - `'warn'` emits a warning (default when the check is enabled).
+   * - `'error'` promotes the emission to a hard build error.
+   * @default 'warn'
    * */
-  eval?: boolean;
+  eval?: false | 'warn' | 'error';
 
   /**
    * Whether to emit warnings when the `output.globals` option is missing when needed.
    *
    * See [`output.globals`](https://rolldown.rs/reference/OutputOptions.globals).
-   * @default true
+   *
+   * - `false` disables the check.
+   * - `'warn'` emits a warning (default when the check is enabled).
+   * - `'error'` promotes the emission to a hard build error.
+   * @default 'warn'
    * */
-  missingGlobalName?: boolean;
+  missingGlobalName?: false | 'warn' | 'error';
 
   /**
    * Whether to emit warnings when the `output.name` option is missing when needed.
    *
    * See [`output.name`](https://rolldown.rs/reference/OutputOptions.name).
-   * @default true
+   *
+   * - `false` disables the check.
+   * - `'warn'` emits a warning (default when the check is enabled).
+   * - `'error'` promotes the emission to a hard build error.
+   * @default 'warn'
    * */
-  missingNameOptionForIifeExport?: boolean;
+  missingNameOptionForIifeExport?: false | 'warn' | 'error';
 
   /**
    * Whether to emit warnings when a `#__PURE__` / `@__PURE__` annotation has no effect due to its position.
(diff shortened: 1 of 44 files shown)


IWANABETHATGUY:
* **#9388** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9388?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9388?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9388).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyf0 (commented):
LGTM @sapphi-red cc

hyf0 (approved):

packages/rolldown/src/options/generated/checks-options.ts
```diff
@@ -8,175 +8,263 @@ export interface ChecksOptions {
...
    *
    * {@include ../docs/checks-circular-dependency.md}
+   *
+   * - `false` disables the check.
+   * - `'warn'` emits a warning (default when the check is enabled).
+   * - `'error'` promotes the emission to a hard build error.
    * @default false
    * */
-  circularDependency?: boolean;
+  circularDependency?: false | 'warn' | 'error';
```
sapphi-red, line 17: This breaks the type compat. `true` is disallowed now.

packages/rolldown/src/utils/validator.ts
```diff
@@ -315,109 +315,109 @@ isTypeTrue<IsSchemaSubType<typeof WatcherOptionsSchema, WatcherOptions>>();
 
 const ChecksOptionsSchema = v.strictObject({
   circularDependency: v.pipe(
-    v.optional(v.boolean()),
+    v.optional(v.union([v.literal(false), v.picklist(['warn', 'error'])])),
```
sapphi-red, line 318: This is a breaking change. Passing `true` will error now.

sapphi-red:
I personally find it confusing how `checks` and `onLog` option interacts after this PR.

hyfdev:
> I personally find it confusing how `checks` and `onLog` option interacts after this PR.

Let's talk about this in today's meeting. @IWANABETHATGUY cc

---

Context:

- The original intention of `checks` is to give users the ability to turn off some checks. For this purpose, using `onLog` can only "ignore" some checks.

IWANABETHATGUY:
> I personally find it confusing how `checks` and `onLog` option interacts after this PR.

I think this is like syntactic sugar for previous checks + onLog. It could avoid the user from writing boilerplate. It also improve the perf a little(Although it is trivial)

IWANABETHATGUY:
Also, I will fix the issue @sapphi-red mentioned after the meeting, or we could revert it if we decide it is not necessary.

PR #9409, opened 2026-05-15
fix(bundler): prune stale runtime helpers after flattening entry-level external re-exports

### LINKED ISSUE #9374: [Bug]: rolldown/runtime is false-positively injected

https://repl.rolldown.rs/#eNptkE0OwiAUhK/y8jaoaeqexJXHEGOoUlODD0Op1jTcXcD+2NgVZL6ZCUOHJfIOrdH6Yl6Unw2V1TV3dVQJ+RLJ8BxAdX8Y66CDiyorUvuEwUNpzR3YEGOCBKk2WYNRNjqdY2DVCQJQrVOWpOZwYMOdHbOIKno0jkOyBSM5+z5JDqy/5beaJd/AipEVkUXkBfm1IPQZTqlh34/y3dU/dtMPybeztnFM+I7agYQdMMmm6mibVydlsXpc+ldbxNoi6OhD8zNEtXSqdug/FZGSyQ==

### What is expected?

Output bundle does not contain rollup/runtime code

Output bundle contains rollup/runtime code, which is not actually used!

### System Info

```Shell
Rollup Repl
```

### Any additional comments?

If you comment `export * from 'external'` - rolldown/runtime is disappears from output

dangreen:
Hi! I'm new to Rust and not deeply familiar with this codebase, so I worked on this fix together with Claude. The changes look fairly targeted to me and the test coverage seems solid - hopefully the code quality meets the bar. Happy to iterate on any feedback.

## Summary

Fixes #9374. When an entry's `export * from 'external'` is flattened by `find_entry_level_external_module` (so the re-export is emitted literally at chunk-render time instead of via `__reExport(my_ns, importee_ns)`), the runtime helpers (`__exportAll`, `__reExport`) and their support code stayed in `used_symbol_refs` / `chunk.depended_runtime_helper`, so dead helper declarations leaked into the runtime chunk and got re-exported across chunks (`as n`, `as r`).

## Root cause

`reference_needed_symbols` eagerly references `__reExport`, the namespace object, and the importee's namespace ref on every `export * from './foo'` where the importee has `has_dynamic_exports = true`. Later, `find_entry_level_external_module` flips `has_dynamic_exports` to `false` for the affected modules - the finalizer correctly skips emitting the `__reExport(...)` call (it gates on `has_dynamic_exports`), but the helper symbols stay marked as used and the runtime stmts stay included, producing dead code.

Additionally, the existing importer-walk loop in `find_entry_level_external_module` was a no-op: `if !invalidated_modules.insert(idx) { continue; }` short-circuited every initial item because `q` was seeded from the same set, so transitive importers were never invalidated and their `has_dynamic_exports` was never recomputed.

## Fix

In `find_entry_level_external_module`, after `propagate_has_dynamic_exports` runs:

1. Fix the importer-walk loop with a separate `walk_visited` set so `invalidated_modules` actually grows to cover transitive importers.
2. For each module whose `has_dynamic_exports` was just cleared and whose `module_namespace_included_reason` doesn't contain `Unknown` (so the namespace isn't needed for `import * as ns`):
   - Clear `NAMESPACE_STMT_IDX` from `stmt_info_included` if it was included solely via `ReExportDynamicExports`.
   - Clear `ReExportDynamicExports`-flagged stmts (their side-effect flag and helper push were gated on `has_dynamic_exports = true`).
   - Drop `RuntimeHelper::ReExport` from `meta.depended_runtime_helper`.
3. Re-aggregate `chunk.depended_runtime_helper` from the updated per-module bitsets for affected chunks.

Extracted into `prune_stale_dynamic_reexport_refs` to keep `find_entry_level_external_module` focused.

## Tests

- `crates/rolldown/tests/rolldown/issues/9374` - regression test reproducing the exact REPL output from the issue.
- `crates/rolldown/tests/rolldown/issues/9374_cjs_mix` - edge case where an entry mixes ESM-dynamic re-export and CJS re-export. `has_dynamic_exports` stays `true` (because CJS importees short-circuit propagation to `true`), so the cleanup correctly skips this module and the `RuntimeHelper::ReExport` bit pushed by the CJS path (which doesn't set `ReExportDynamicExports` meta) is preserved.
- Updated snapshots in `issues/6992`, `7115`, `7233`, `7233_chain` - these tests previously froze the same bug in their expected output; the diff shows the dead `__exportAll`/`__reExport` exports and `import "./_virtual/_rolldown/runtime.js"` lines being removed.
- One cosmetic snapshot diff in `issues/5923` - extra `//#region`/`//#endregion` markers around `main.js`, unrelated functionally. Same artifact as the existing `5923` snapshot has on its own (empty runtime chunk markers).

All 1714 integration tests pass; clippy clean.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/code_splitting.rs
@@ -13,7 +13,7 @@ use rolldown_common::{
   Chunk, ChunkIdx, ChunkKind, ChunkMeta, EntryPointKind, ExportsKind, ImportKind, ImportRecordIdx,
   ImportRecordMeta, IndexModules, Module, ModuleIdx, ModuleNamespaceIncludedReason, ModuleTag,
   ModuleTagBitSet, ModuleTagRegistry, PostChunkOptimizationOperation, PreserveEntrySignatures,
-  SymbolRef, WrapKind,
+  RuntimeHelper, StmtInfoMeta, StmtInfos, SymbolRef, WrapKind,
 };
 use rolldown_error::BuildResult;
 use rolldown_utils::{
@@ -654,21 +654,32 @@ impl GenerateStage<'_> {
       vec.sort_by_key(|idx| self.link_output.module_table[*idx].exec_order());
       chunk_graph.chunk_table[chunk_idx].entry_level_external_module_idx = vec;
     }
-    // re propagate `meta.has_dynamic_exports` for affect modules
+    // Extend `invalidated_modules` with every transitive importer of the
+    // directly-affected modules: a module's `has_dynamic_exports` is derived
+    // from its importees' values, so importers must be revisited too.
     let mut q = invalidated_modules.iter().copied().collect::<VecDeque<_>>();
+    let mut walk_visited = FxHashSet::default();
     while let Some(idx) = q.pop_front() {
-      if !invalidated_modules.insert(idx) {
+      if !walk_visited.insert(idx) {
         continue;
       }
       let Module::Normal(module) = &self.link_output.module_table[idx] else {
         continue;
       };
-      q.extend(module.importers_idx.iter());
+      for &importer_idx in &module.importers_idx {
+        if invalidated_modules.insert(importer_idx) {
+          q.push_back(importer_idx);
+        }
+      }
     }
 
     if invalidated_modules.is_empty() {
       return;
     }
+
+    // Snapshot the set before `propagate_has_dynamic_exports` drains it, so
+    // the cleanup pass below knows which modules to revisit.
+    let modules_to_revisit: Vec<ModuleIdx> = invalidated_modules.iter().copied().collect();
 
     let mut visited = FxHashSet::with_capacity(invalidated_modules.len());
     for module_idx in invalidated_modules.clone() {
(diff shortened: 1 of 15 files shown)


dangreen:
@IWANABETHATGUY Hi. How soon can you look at it?

dangreen:
@IWANABETHATGUY ?

IWANABETHATGUY:
> @IWANABETHATGUY ?

I'm sorry, I have a lot on my plate right now. Usually, I like to tackle high-priority tasks first and then move on to the lower priority ones. Thanks for understanding!

dangreen:
@IWANABETHATGUY 

Hi,

Sorry for pinging again, but it has already been about a month since this PR was opened, and it still has not received an initial review.

I understand that the team may have a lot of work, but this PR fixes basic bundling / DCE functionality, and the fix can reduce bundle size for users affected by this class of issues. In my opinion, this kind of correctness issue should be prioritized, because it directly affects Rolldown’s output quality.

To be honest, from the outside it is starting to feel like fixes for basic functionality are not getting much attention. Right now, Rolldown’s DCE quality is still behind Rollup in some cases because of bugs like the ones I have reported, and this PR is one concrete step toward fixing that.

I opened the issue, prepared the PR, added the required coverage, and I’m happy to address feedback. But without even an initial review, there is not much else I can do to move this forward.

Could someone please take a look or let me know whether this kind of fix is currently considered a priority?

Thanks.

dangreen:
@hyf0 @IWANABETHATGUY @shulaoda ping

PR #9420, opened 2026-05-16
perf(scoping): cache semantic Stats on AstScopes to speed up rebuilds

Dunqing:
## Summary

Each `SemanticBuilder::build` walks the AST twice — once via `Stats::count` to pre-allocate the symbol/scope/reference buffers, then again to actually populate them. `Stats::count` is roughly 30% of the build cost. When a prior build's `Stats` is fed back in via `.with_stats(...)`, that first walk is skipped.

`pre_process_ecma_ast.rs` already caches `Stats` on `PreProcessEcmaAst` and reuses it across the multiple in-pipeline rebuilds (transformer → inject → DCE → final). This change extends the same trick to downstream consumers that previously rebuilt scoping from scratch:

- **HMR finalization** (`hmr_stage.rs`, 3 sites — module invalidate / self-accepting / module update). `clone_with_another_arena` preserves semantic IDs, so the link-time scoping is logically valid for the cloned AST, but `oxc::semantic::Scoping` is not `Clone` (it contains a self-referential `ScopingCell`), so a rebuild is still required. Feeding the cached `Stats` through `ModuleRenderInput` skips `Stats::count` on every HMR cycle per affected module.
- **`enhanced_transform`** (the NAPI transform API). Two rebuilds used bare `SemanticBuilder::new()`: one in the take-or-rebuild fallback after the define plugin, and one before inject. The latter also silently dropped `with_enum_eval`, which is set by `semantic_builder_for_transform()` for the TS enum-lowering path. Both now go through `semantic_builder_for_transform().with_stats(...)`.

`AstScopes` gains a `stats` field and `stats()` getter; `AstScopes::new`, `SymbolRefDbForModule::new`, and `AstScanner::new` take `Stats`. `set_scoping` is extended to update stats alongside scoping so `merge_from_build` (used by the incremental-build cache) keeps the two in sync. `EcmaAst` gets `make_scoping_with_stats` / `make_scoping_using_stats` helpers.

`AstScopes::stats()` is documented as a pre-allocation hint, not an authoritative count: link-stage facade symbols created via `create_facade_root_symbol_ref` are not reflected, so the value may under-count by the facade delta. Slight under-allocation triggers a small reallocation during rebuild but is not a correctness issue.

## Not addressed

- The forced rebuilds in `pre_process_ecma_ast.rs` Steps 3–5 (after transformer / inject / DCE). Eliminating those requires the corresponding oxc APIs (`Transformer::build_with_scoping`, `Compressor::dead_code_elimination_with_scoping`, `InjectGlobalVariables::build`) to return `Scoping` instead of consuming it — an upstream oxc change.
- Eliminating the HMR rebuild entirely. Conceptually possible because `clone_with_another_arena` preserves semantic IDs, but blocked by `Scoping` not being `Clone`.

## Verification

- `cargo check --workspace` clean.
- `cargo test --workspace` passes (1713 rolldown tests + the rest of the workspace).
- Not profiled — the Rust HMR finalization path isn't exercised by `cargo test`. JS-side HMR e2e (`packages/rolldown-tests` / `pnpm test`) is the next step to confirm there's no behavioural regression and to quantify the improvement.

### DIFF
--- a/crates/rolldown/src/ast_scanner/cjs_export_analyzer.rs
+++ b/crates/rolldown/src/ast_scanner/cjs_export_analyzer.rs
@@ -261,7 +261,8 @@ mod tests {
     let ret = Parser::new(allocator, source, source_type).parse();
     let program = ret.program;
     let semantic_ret = SemanticBuilder::new().build(&program);
-    (AstScopes::new(semantic_ret.semantic.into_scoping()), program)
+    let stats = semantic_ret.semantic.stats();
+    (AstScopes::new(semantic_ret.semantic.into_scoping(), stats), program)
   }
 
   fn extract_call_expr<'a>(
--- a/crates/rolldown/src/ast_scanner/mod.rs
+++ b/crates/rolldown/src/ast_scanner/mod.rs
@@ -12,7 +12,7 @@ use const_eval::{ConstEvalCtx, try_extract_const_literal};
 use oxc::ast::ast::{BindingPattern, Expression, ImportExpression};
 use oxc::ast::{AstKind, ast};
 use oxc::ast_visit::walk;
-use oxc::semantic::{Reference, ReferenceId, ScopeFlags, Scoping};
+use oxc::semantic::{Reference, ReferenceId, ScopeFlags, Scoping, Stats};
 use oxc::span::SPAN;
 use oxc::{
   ast::{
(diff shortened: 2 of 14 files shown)


Dunqing:
* **#9420** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9420?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9420?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9420).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #9424, opened 2026-05-17
feat: update oxc to 0.131.0

shulaoda:
## Summary

Bumps the oxc toolchain from `0.130.0` to `0.131.0`.

- Rust crates: `oxc`, `oxc_allocator`, `oxc_ecmascript`, `oxc_napi`, `oxc_str`, `oxc_minify_napi`, `oxc_parser_napi`, `oxc_transform_napi`, `oxc_traverse` → `0.131.0`.
- npm catalog: `@oxc-project/runtime`, `@oxc-project/types`, `oxc-minify`, `oxc-parser`, `oxc-transform` → `0.131.0`.
- Regenerate `crates/rolldown_plugin_oxc_runtime/src/generated/embedded_helpers.rs` against the new `@oxc-project/runtime`.
- Refresh the `dce_of_iife` esbuild snapshot to reflect upstream minifier changes (one IIFE is now wrapped with a `/* @__PURE__ */` annotation and the trailing `stuff();` call is dropped).

No source code changes were required, `cargo check` passed cleanly against the new version.

## Notes

- `oxc_resolver` / `oxc_resolver_napi` stay at `11.19.1` (already the latest).
- No public API changes were exposed to rolldown.

PR #9431, opened 2026-05-18
fix(treeshake): propagate pure annotation through compound exprs

### LINKED ISSUE #9425: side-effect detector loses `PureAnnotation` flag when a `/* @__PURE__ */` call is nested inside `ObjectExpression` / `ArrayExpression`
### Summary

`SideEffectDetector` in `crates/rolldown/src/ast_scanner/side_effect_detector/mod.rs` only sets `SideEffectDetail::PureAnnotation` for a `CallExpression` that is itself the **top-level** expression being inspected. When the pure-annotated call is nested inside a compound expression (object literal property value, array element, sequence, etc.) the detector falls through to oxc's `MayHaveSideEffects`, which honors the `pure` flag and reports `false`, and the `PureAnnotation` flag never propagates up to the statement.

Concrete fall-through in `detect_side_effect_of_expr`:

```rust
// crates/rolldown/src/ast_scanner/side_effect_detector/mod.rs
match expr {
    // ... explicit handlers for Member, Identifier, Assignment, Chain, Update, New, Call ...
    _ => expr.may_have_side_effects(self).into(),  // ObjectExpression hits this
}
```

Because the statement-level `side_effect` doesn't intersect `PureAnnotation` / `GlobalVarAccess` / `Unknown`, `EcmaViewMeta::ExecutionOrderSensitive` is not set, and `wrap_modules` (`crates/rolldown/src/stages/link_stage/wrapping.rs`) takes the `avoid_wrapping` branch:

```rust
let avoid_wrapping = on_demand_wrapping
  && !module.meta.contains(EcmaViewMeta::ExecutionOrderSensitive)
  && module.import_records.is_empty()
  && !linking_info.required_by_other_module;
```

The module is given `WrapKind::None` and its top-level `var` initializer is emitted at chunk top level, before any `setup.js`-style sibling module runs.

### Reproduction

Minimal source — already present as two integration tests:

- `crates/rolldown/tests/rolldown/function/experimental/strict_execution_order/concatenate_wrapped_modules`
- `crates/rolldown/tests/rolldown/function/experimental/strict_execution_order/issue_5303`

```js
// foo_inner.js
export default {
  foo: /* #__PURE__ */ (() => globalValue)(),
};

// setup.js — runs before foo_inner.js per strictExecutionOrder
globalThis.globalValue = 'foo';
```

Expected bundled output (current snapshots): `foo_inner.js` is wrapped in `__esmMin(...)` so the assignment happens after `setup.js`.

Actual bundled output after bumping oxc to 0.131:

```js
//#region \0rolldown/runtime.js
//#endregion
//#region foo_inner.js
var foo_inner_default = { foo: /* @__PURE__ */ (() => globalValue)() };
__esmMin((() => {
  //#region setup.js
  globalThis.globalValue = "foo";
  //#endregion
  //#region foo.js
  assert.strictEqual(foo_inner_default.foo, "foo");
  //#endregion
  //#region main.js
  assert.equal(foo_inner_default.foo, "foo");
  //#endregion
}))();
```

`var foo_inner_default = ...` runs at top level, before `globalThis.globalValue` is assigned, producing:

```
ReferenceError: globalValue is not defined
```

### Why this was latent before oxc 0.131

oxc 0.130's DCE pass unconditionally inlined `(/* @__PURE__ */ (() => globalValue)())` down to a bare `globalValue` identifier. After inlining, rolldown's `GlobalVarAccess` detection path (an independent code path from `PureAnnotation`) caught the unresolved global and marked the module `ExecutionOrderSensitive`, so it got wrapped. The wrap-correctness was accidentally relying on oxc inlining the IIFE — the detector's nested-PURE handling has been incomplete this whole time.

oxc 0.131 PR oxc-project/oxc#22349 added `iife_inline_would_lose_pure`: in DCE-only mode the minifier now intentionally **preserves** a pure-annotated IIFE whose body has side effects, so the annotation survives for downstream tree-shakers. That is the correct call on oxc's side — but it removes the inline-into-bare-global behavior that rolldown was implicitly leaning on, surfacing this detector gap.

### Suggested fix

Stop delegating compound expressions to oxc's `MayHaveSideEffects` and recurse through rolldown's own detector instead, so flags like `PureAnnotation` and `GlobalVarAccess` propagate to the statement level. At minimum, handle `ObjectExpression`, `ArrayExpression`, and `SequenceExpression` in `detect_side_effect_of_expr` with explicit arms that fold their children's `SideEffectDetail` together. The existing `detect_side_effect_of_call_expr` already produces the right flags for a leaf pure call; the missing piece is propagation through the surrounding compound expression.

### Related

- oxc PR that surfaced this: oxc-project/oxc#22349
- oxc 0.131 bump where CI caught it: rolldown/rolldown#9424

Dunqing:
### Root cause

`SideEffectDetector::detect_side_effect_of_expr`'s catch-all arm collapses any compound expression (`ObjectExpression`, `ArrayExpression`, `SequenceExpression`, `Conditional`/`Logical`/`Binary`/`Unary` expressions, `TemplateLiteral`) into a plain bool via `From<bool> for SideEffectDetail`:

```rust
impl From<bool> for SideEffectDetail {
  fn from(value: bool) -> Self {
    if value { SideEffectDetail::Unknown } else { SideEffectDetail::empty() }
  }
}
```

That conversion can only encode `Unknown` or `empty`. The `PureAnnotation` and `GlobalVarAccess` metadata bits on a buried call are silently dropped at the catch-all and never reach the statement-level intersection that drives `EcmaViewMeta::ExecutionOrderSensitive`.

Concretely, `export default { foo: /* @__PURE__ */ (() => globalValue)() }` reaches statement level with `SideEffectDetail::empty()`. `wrap_modules` (`crates/rolldown/src/stages/link_stage/wrapping.rs`) takes the `avoid_wrapping` branch, `WrapKind::None` is emitted, and the module's top-level `var foo_inner_default = ...` runs at chunk top level — before sibling modules like `setup.js` that the IIFE depends on:

```js
var foo_inner_default = { foo: /* @__PURE__ */ (() => globalValue)() };  // globalValue is undefined here
__esmMin((() => {
  globalThis.globalValue = "foo";
  // ...
}))();
```

### Why this was latent before oxc 0.131

oxc 0.130's DCE pass unconditionally inlined `/* @__PURE__ */ (() => globalValue)()` down to a bare `globalValue` identifier. After inlining, the detector's `Identifier` arm independently flagged the unresolved global via `GlobalVarAccess` (a code path separate from `PureAnnotation`), the module was marked `ExecutionOrderSensitive`, and wrapping kicked in. Wrap-correctness was accidentally riding on oxc's IIFE inlining.

oxc 0.131 (oxc-project/oxc#22349) intentionally **preserves** pure-annotated IIFEs in DCE-only mode so downstream tree-shakers can see the annotation. That is the right call on oxc's side — but it removes the inline-to-bare-global behavior rolldown was implicitly leaning on, exposing this detector gap.

### Changes

- Add explicit arms in `detect_side_effect_of_expr` for the transparent compound expression types so metadata propagates: `ObjectExpression`, `ArrayExpression`, `SequenceExpression`, `ConditionalExpression`, `LogicalExpression`, `BinaryExpression`, `UnaryExpression`, `TemplateLiteral`.
- Each arm goes through a single helper `fold_compound(expr, children)` that:
  1. Asks oxc's `may_have_side_effects` for the spec-level boolean (handles ToPrimitive coercion, getter triggers in spread, Symbol checks in templates, etc.). If oxc says yes, return `Unknown` and stop.
  2. Otherwise recurses through the children via rolldown's own detector to harvest `PureAnnotation` / `GlobalVarAccess` flags.
  3. Strips `Unknown` from the harvested result since oxc already certified the parent has no side effect — only metadata bits should propagate.
- Peel transparent syntactic wrappers (`(x)`, `x as T`, `x satisfies T`, `x!`, `<T>x`, `x<T>`) at the top of `detect_side_effect_of_expr` via `Expression::get_inner_expression()` — they add no runtime semantics, so a single peel covers all five wrapper variants (and any future ones oxc adds to that method).

This mirrors the existing gate-and-harvest pattern already used by `detect_side_effect_of_call_expr` and the `NewExpression` arm.

### Tests

New unit tests in `side_effect_detector::test`:

- `test_pure_annotation_propagates_through_compound_expr` — covers `PureAnnotation` propagation through `Object`/`Array`/`Sequence`/`Conditional`/`Logical`/`Binary` (`===`) / `Unary` (`typeof`) / computed-key / nested compound, plus `GlobalVarAccess` propagation through compound, plus TS wrapper passthroughs (`as`, `satisfies`, `!`).
- `test_pure_annotation_not_propagated_through_function_body` — verifies a pure annotation inside an arrow/function body does **not** propagate to the statement level (function bodies aren't evaluated at module init).
- `test_compound_expr_side_effectful_operand_still_unknown` — verifies a real side effect anywhere in a compound expression still surfaces as `Unknown`; the new arms don't weaken side-effect detection.

Existing integration suites kept green without snapshot changes:

```sh
cargo test --package rolldown --lib ast_scanner::side_effect_detector::test::
cargo test --package rolldown --test integration -- strict_execution_order  # 24 passed
cargo test --package rolldown --test integration -- tree_shaking             # 72 passed
```

The two pre-existing integration fixtures called out in the issue (`function/experimental/strict_execution_order/concatenate_wrapped_modules` and `.../issue_5303`) pass with their current snapshots — those snapshots already encode the wrapped behavior, so this PR restores conformance to them after the oxc 0.131 bump (rolldown/rolldown#9424) would otherwise break them.

Closes #9425.

### DIFF
--- a/crates/rolldown/src/ast_scanner/side_effect_detector/mod.rs
+++ b/crates/rolldown/src/ast_scanner/side_effect_detector/mod.rs
@@ -266,12 +266,16 @@ impl<'a> SideEffectDetector<'a> {
         expr,
         // `p.computed == true` => `p.key` is always an Expression variant
         // (oxc's parser only emits Static/PrivateIdentifier for `key:` syntax).
-        obj.properties.iter().flat_map(|prop| match prop {
-          ast::ObjectPropertyKind::ObjectProperty(p) => {
-            [p.computed.then(|| p.key.to_expression()), Some(&p.value)]
-          }
-          ast::ObjectPropertyKind::SpreadProperty(s) => [Some(&s.argument), None],
-        }).flatten(),
+        obj
+          .properties
+          .iter()
+          .flat_map(|prop| match prop {
+            ast::ObjectPropertyKind::ObjectProperty(p) => {
+              [p.computed.then(|| p.key.to_expression()), Some(&p.value)]
+            }
+            ast::ObjectPropertyKind::SpreadProperty(s) => [Some(&s.argument), None],
+          })
+          .flatten(),
       ),
       Expression::ArrayExpression(arr) => self.fold_compound(
         expr,
--- a/crates/rolldown/src/ast_scanner/side_effect_detector/mod.rs
+++ b/crates/rolldown/src/ast_scanner/side_effect_detector/mod.rs
@@ -190,6 +190,9 @@ impl<'a> SideEffectDetector<'a> {
   }
 
   fn detect_side_effect_of_expr(&self, expr: &Expression) -> SideEffectDetail {
+    // Peel transparent wrappers (`(x)`, `x as T`, `x satisfies T`, `x!`, `<T>x`, `x<T>`)
+    // — they add no runtime semantics, so we recurse into the inner expression.
+    let expr = expr.get_inner_expression();
     match expr {
       // --- Bundler-specific overrides (metadata or custom logic) ---
       oxc::ast::match_member_expression!(Expression) => {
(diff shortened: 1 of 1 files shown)


Dunqing:
* **#9432** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9432?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9431** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9431?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9431?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9431).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

shulaoda (approved):

PR #9439, opened 2026-05-18
fix(dev/lazy): make `resolve_id` idempotent when the resolved id is already a lazy entry

h-a-n-a:
## Summary

This PR fixed an issue that `resolve_id` is not idempotent, causing lazy compilation to fail.

In `LazyCompilationPlugin::resolve_id`, calling `ctx.resolve` may trigger 
other plugins' resolve_id hooks (in this case, vite alias plugin), and these hooks may also call `ctx.resolve`,
so it may cause `LazyCompilationPlugin::resolve_id` to be called multiple times for the same module 
so that the `original_id` may be marked as `rolldown-lazy=1` already.

Here we changed this behavior and check if the module has already been marked as a lazy entry.
Otherwise, `delete modules[$STABLE_PROXY_MODULE_ID]` may not be aligned with the module ID.

closes https://github.com/vitejs/vite/issues/22454

## Test

Added `lazy-aliased-import` test.

### DIFF
--- a/crates/rolldown_plugin_lazy_compilation/src/lazy_compilation_plugin.rs
+++ b/crates/rolldown_plugin_lazy_compilation/src/lazy_compilation_plugin.rs
@@ -107,7 +107,19 @@ impl Plugin for LazyCompilationPlugin {
         )
         .await??;
 
-      let lazy_id: ArcStr = format!("{}?rolldown-lazy=1", original_id.id).into();
+      // Idempotent: Calling `ctx.resolve` may trigger other plugins' resolve_id hooks,
+      // and these hooks may also call `ctx.resolve`,
+      // so it may cause `LazyCompilationPlugin::resolve_id` to be called multiple times for the same module
+      // so that the `original_id` may be marked as `rolldown-lazy=1` already.
+      //
+      // Here we check if the module has already been marked as a lazy entry.
+      // Otherwise, `delete modules[$STABLE_PROXY_MODULE_ID]` may not be aligned with the module ID,
+      // causing the bundler failing to invalidation downstream.
+      let lazy_id: ArcStr = if original_id.id.as_str().ends_with("?rolldown-lazy=1") {
+        original_id.id.as_str().into()
+      } else {
+        format!("{}?rolldown-lazy=1", original_id.id).into()
+      };
       self.lazy_entries.insert(lazy_id.clone());
 
       return Ok(Some(HookResolveIdOutput {
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -101,5 +101,9 @@
     "packages/test-dev-server/tests/playground/lazy-nested-dynamic-import": {
       "entry": ["dev.config.mjs", "app.js", "outer.js", "inner.js"], // Playwright test with dev config and source files
     },
+    "packages/test-dev-server/tests/playground/lazy-aliased-import": {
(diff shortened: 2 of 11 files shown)


h-a-n-a:
* **#9439** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9439?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9439?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9439).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown_plugin_lazy_compilation/src/lazy_compilation_plugin.rs
sapphi-red, line 104: `skipSelf` won't help because the `specifier` will be different, right?
I wonder why it doesn't end up called recursively though 🤔

h-a-n-a: `skip_self` does not propagate transitively to the nested `resolve_id` calls. So the `viteAliasPlugin` still get to call `LazyCompilationPlugin::resolve_id`.

sapphi-red: Hmm, is it related to this test failure?
https://github.com/rolldown/rolldown/blob/776ec77feb2a9a55bebb6342cb293eb4cd528905/packages/rollup-tests/src/failed-tests.json#L281

h-a-n-a: I think they're related. This test is supposed to test the case where the hook is getting called recursively with the same args and being able to bailout the `this.resolve` resolution with `null`, which we're yet to support. But it still does not fix this case as the `specifier` is still not the same as you mentioned above.

h-a-n-a: ~~I'm not quite sure if the current `skipSelf` on the Rust side works. Let me add a test case.~~
It works as https://github.com/rolldown/rolldown/blob/main/packages/rolldown/tests/fixtures/plugin/context/resolve3/_config.ts passes.

sapphi-red: Ah, I think I got it.

sapphi-red (approved):

h-a-n-a:
### Merge activity

* **May 19, 6:34 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **May 19, 6:34 AM UTC**: `h-a-n-a` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **May 19, 6:39 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9451, opened 2026-05-18
fix: prevent entries_aware subgroup merging from leaking side effects across entries

### LINKED ISSUE #9450: [Bug]: codeSplitting rewrites side-effects across entry points

https://repl.rolldown.rs/#eNqVUtuK2zAQ/ZVBL05C6qR9NGyh9PK0S6DtW12Kao0TgT1jJHmzi/G/r2RFTsJeYPUgRnM5OnNmBlGLYhCGm0bxkfKKqdb73NngJVG8FFmLygd027Fx4B47hAF+ntJ2ndNMFkaoDbeQpfKspJI2qxWsNiV5KOsgAhbPSm9gKAlAU9e7ItoAsoBMk8IHTyBbR9//5PvwafaO0829uyyuWOGvrtHOafIfnrwAe8N9Zwv4kxxwjoVDssUCSlFx2zKV4vRvPEjOaLRfjtL4JGd6fDV8h2aPvw8G7YEbVcDH7b/tdnuRPibzbzRiF+H2kZLwYZJaYS37JgknxrVIiqRhze+rEQ1QM88TyTf24DmpIFkAV1z1rWebS6W+33vjVluHhGaRfdvdfWXyD3fLUqHK1rBYws3nqJIHXSyDEcbJDeZoDPuqeUw+OC49FRRFkCfxnaZ1xTh6Iuf30ZH2kapLUnG1WlZ9g36T5FFqB1GHxVXrE/OYl7/ZybRbsRXfwQyQGjg7Iv/TrOqeqrDOUaUztwv0H7vdhFuSGD3yvS9upEPrxPgEZhoqVA==

### What is expected?

The side effects (event listener registration) should only be performed in each entry point.
Meaning if loading "a" then only `console.error("index.ts");` should be executed and not `console.error("index2.ts");`

Both side effects are combined into the same chunk and thus both entry points will execute both side effects.

### System Info

```Shell
Binaries:
    Node: 24.14.0
    npm: 11.9.0
```

### Any additional comments?

The idea behind the chunking was to reduce the total number of chunks by merging all related code into common chunks. In my actual use case I have 3 entry points but rolldown generates 19 output chunks.

So I hoped to tweak rolldown to generate at most 7 chunks:
- entry1: all code and dependencies only used by entry 1
- entry2: all code and dependencies only used by entry 1
- entry3: all code and dependencies only used by entry 1
- chunk12: code shared with entry1 and 2
- chunk13: code shared with entry1 and 3
- chunk23: code shared with entry2 and 3
- chunk123: code shared with entry1, 2 and 3

zennnnnnn11:
## Summary

Fixes #9450

When `codeSplitting.groups` is configured with `entriesAware: true` and a high `entriesAwareMergeThreshold` without a `test` filter, user-defined entry modules get captured into entries_aware groups. The merge logic (`merge_entries_aware_subgroups`) then combines small subgroups — including subgroups containing entry modules from different entries — into a single `ChunkKind::Common` chunk. This causes one entry's side effects to execute when loading a different entry.

**Before fix:** `entry-a.js` output contains both entries' side effects:
```js
console.log("side-effect-a");
console.log("side-effect-b"); // leaked from entry-b
```

**After fix:** each entry only contains its own side effects:
```
entry-a.js              -> side-effect-a only
entry-b.js              -> side-effect-b only
common~entry-a~entry-b.js -> shared code only
```

## Fix

Remove user-defined entry modules from entries_aware groups before subgroup splitting and merging in `process_entries_aware_groups()`. This follows the same pattern as `extract_runtime_chunk()`: modules with a fixed destination (their entry chunk) are removed from groups to prevent misplacement. Removed entry modules fall through to the normal assignment loop in `split_chunks()`.

- Only affects entries_aware groups (non-entries_aware groups like `basic` are unchanged)
- Uses `user_defined_entry_modules` (dynamic import entries can still be captured)

## Test plan

- [x] `entries_aware_side_effect_leak` -- core bug: 2 entries + high merge threshold + no test filter
- [x] `entries_aware_entry_modules_with_dynamic_import` -- static + dynamic import mix (original issue scenario)
- [x] `entries_aware_three_entries_no_leak` -- 3 entries with multiple shared dependency layers
- [x] `entries_aware_with_test_filter_excludes_entries` -- fix is a no-op when test filter already excludes entries
- [x] All 30 `advanced_chunks` tests pass (0 regressions)
- [x] All 17 `code_splitting` tests pass

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/manual_code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/manual_code_splitting.rs
@@ -247,7 +247,18 @@ impl ManualSplitter<'_> {
     entries_aware_groups: Vec<ModuleGroup>,
     module_groups: &mut IndexVec<ModuleGroupIdx, ModuleGroup>,
   ) {
-    for group in entries_aware_groups {
+    for mut group in entries_aware_groups {
+      if group.modules.is_empty() {
+        continue;
+      }
+
+      // User-defined entry modules have a fixed destination (their entry chunk).
+      // Remove them so subgroup merging cannot combine entries with different
+      // bits into a single Common chunk, which would leak side effects across
+      // entry boundaries. Same pattern as `extract_runtime_chunk`.
+      for &entry_idx in &self.link_output.user_defined_entry_modules {
+        group.remove_module(entry_idx, &self.link_output.module_table);
+      }
       if group.modules.is_empty() {
         continue;
       }
--- a/crates/rolldown/tests/rolldown/function/advanced_chunks/entries_aware_entry_modules_with_dynamic_import/_config.json
+++ b/crates/rolldown/tests/rolldown/function/advanced_chunks/entries_aware_entry_modules_with_dynamic_import/_config.json
@@ -1,0 +1,19 @@
+{
+  "_comment": "Same as entries_aware_side_effect_leak but entry-b uses a dynamic import for the shared module. This matches the original bug report where one entry statically imports and another dynamically imports the shared code.",
+  "config": {
+    "input": [
+      { "name": "entry-a", "import": "./entry-a.js" },
+      { "name": "entry-b", "import": "./entry-b.js" }
+    ],
+    "manualCodeSplitting": {
+      "groups": [
+        {
+          "name": "common",
+          "entriesAware": true,
+          "entriesAwareMergeThreshold": 10000
+        }
+      ]
+    }
+  },
+  "hiddenRuntimeModule": true
+}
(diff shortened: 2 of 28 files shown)


hyfdev:
https://github.com/rolldown/rolldown/issues/9450 is not a bug. Reason is explained in the issue.

susnux:
I think this PR solves the problem, because even with the suggested workaround of `strictExecutionOrder` the result is invalid:

entry1: console.error("1")
entry2: console.error("2")

with the current chunking even with strict execution order set, the result when loading entry2 is:
> 1
> 2

while the expected result is
> 2

From the tests of this PR it seems it properly fixes the problem here.

hyfdev:
I appreciate the efforts, but this fix is not a correct fix. It just happen to work on your case. `strictExecutionOrder` is the solution. Please follow https://github.com/rolldown/rolldown/issues/9463 for your case.

PR #9458, opened 2026-05-19
perf: reuse scoping rebuilds in preprocessing

Dunqing:
## Summary
- Reuses stale `Scoping` storage during Rolldown preprocessing rebuilds through Oxc's conservative `build_scoping` API.
- Adds a `pre_process_ecma_ast` Criterion group so the PR benchmark run can measure preprocessing directly.
- Temporarily patches Oxc crates to `oxc-project/oxc@codex/fast-scoping-rebuilds` so this PR builds against oxc-project/oxc#22562 before the APIs are released; `Cargo.lock` currently pins Oxc commit `51cd80ca9fa177714f6a0c713b7e48d09ed099da`.
- Verified with `cargo check -p rolldown --lib` and `cargo check -p bench --bench bench --features codspeed`. Earlier short local benchmarks showed the node-skipping version was noisy; this version is intentionally lower risk so PR benchmark data can decide whether deeper optimization is worth a follow-up.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1792,8 +1792,7 @@ dependencies = [
 [[package]]
 name = "oxc"
 version = "0.132.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3fecf76aaae6c29dd8dcde82e8a526ba8e7859ecfa98a2b8b5d54fc17965a241"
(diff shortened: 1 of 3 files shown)


Dunqing:
DON'T REVIEW, JUST AN EXPERIMENT. OPEN PR FOR TRIGGERING BENCHMARK CI

Dunqing:
Slop

PR #9460, opened 2026-05-19
perf: reuse scoping rebuilds in preprocessing via Scoping::reset

Dunqing:
## Summary

Threads a `recycled_scoping: Option<Scoping>` through `PreProcessEcmaAst::build` so the stale `Scoping` after each AST-mutating pass (transformer / define / inject / DCE) can be recycled. The next `recreate_scoping` call uses `Scoping::reset` to clear it in place — preserving the bumpalo arena's chunks and the flat-table heap capacity — and feeds it back to `SemanticBuilder::with_scoping`. This avoids the allocator init cost on each rebuild (3-4 times per file).

DCE is switched from `dead_code_elimination_with_scoping` (consumes the scoping) to `dead_code_elimination_with_scoping_returning_scoping` so the post-DCE scoping can be recycled too.

The initial Step 1 build (`with_check_syntax_error(true)`) is unchanged — the syntax checker still uses the default builder.

## Benchmark

`cargo bench -p bench`, this branch vs `origin/main`:

| Bench | main | this PR | Δ |
|---|---|---|---|
| threejs | 23.27 ms | 22.35 ms | **−4.0%** |
| threejs-sourcemap | 27.82 ms | 27.38 ms | **−1.6%** |
| rome_ts | 44.54 ms | 43.31 ms | **−2.8%** |
| multi-duplicated-top-level-symbol | 19.42 ms | 19.94 ms | +2.7% (noisy, criterion interval [19.26, 21.0]) |

Three out of four show clear wins.

## Depends on

Upstream oxc PR adding the APIs: https://github.com/oxc-project/oxc/pull/22571. The `[patch.crates-io]` section in this PR points at that branch; remove the patch once oxc 0.133.0 (or whichever version the APIs land in) is published.

## Test Plan

- [x] `cargo test -p rolldown --lib` — 56 tests pass

AI disclosure: drafted with Claude Code, reviewed manually.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -379,7 +379,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -690,7 +690,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1275,7 +1275,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi",
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1640,7 +1640,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


Dunqing:
Closing this PR.

## Why

This PR threaded a `recycled_scoping: Option<Scoping>` through `PreProcessEcmaAst::build` to use upstream oxc's new `Scoping::reset` + `SemanticBuilder::with_scoping` + `Compressor::dead_code_elimination_with_scoping_returning_scoping` APIs (oxc PR #22580 in its original form).

After investigation, the rolldown-side changes here **don't actually deliver measurable performance improvements**:

- I micro-benched the arena-reuse path directly (build → drop+rebuild vs `reset` + `with_scoping`). On real files the savings range from **+5.2% (small files) to −0.8% (large files)** — net-zero on average. The `reset()` work (draining cell-internal containers, pulling the bumpalo `Allocator` out, rebuilding the cell) costs roughly as much as the chunk allocations it avoids.

- The earlier rolldown-bench wins I reported (~6% on `rome_ts`, ~4% on `threejs`) were **Criterion measurement noise**, not real signal. Math check: arena-reuse saves ~10 µs per `recreate_scoping` call × 3-4 calls per file, totaling ~30-40 µs per file. For a 23 ms `threejs` bundle that's well under 1%. The claimed 4% had to come from somewhere else (system noise + run-to-run variance).

- The wins that DO appear in rolldown's CodSpeed run come entirely from the **upstream `unresolved_references` pre-reserve fix** in oxc PR #22580 (now reduced to just that one commit). That fix benefits every `SemanticBuilder::build` call regardless of whether the caller uses `Scoping::reset` or not. Rolldown picks it up automatically via its oxc dependency once the oxc PR lands — **no source change in rolldown is needed**.

## The right path

Wait for oxc #22580 to merge. When the next oxc release ships, rolldown's `recreate_scoping` will benefit automatically with zero diff in rolldown itself.

If a future iteration of rolldown's preprocessing pipeline benefits from new oxc APIs (e.g., a `SemanticBuilder::with_class_table(false)` opt-out for the transformer/define/inject/DCE rebuilds — possible follow-up), that's a separate, narrower change.

## Honest acknowledgment

I shipped this PR with bench numbers I hadn't sanity-checked against a controlled experiment. The numbers turned out to be noise. Closing this is the right call — the API surface and code complexity added by threading `recycled_scoping` everywhere aren't justified by a real measured win.

PR #9351, opened 2026-05-11
test: add regression for CJS facade chunk merge into entry

IWANABETHATGUY:
The issue is fixed in https://github.com/rolldown/rolldown/pull/9305
Adds a regression test (`crates/rolldown/tests/rolldown/optimization/chunk_merging/cjs_facade_reexport_merges_into_entry`) reproducing the scenario from #9331 — two CJS facade modules (`module.exports = require('./*-impl.js')`) consumed across a chain of nested dynamic imports.

Snapshots the expected 3-chunk output (user entry / dynamic route / dynamic child, with both facades and their impls merged into the user entry — no stray common chunk for the facade) so future changes to the chunk optimizer can't regress this case.

closed #9331

PR #9305, opened 2026-05-07
feat(chunk-optimization): dedupe already-loaded dynamic deps

IWANABETHATGUY:
Adds a pre-materialization chunk optimization pass that removes dynamic-entry dependency bits for modules already guaranteed to be loaded by every importer of that dynamic entry. This lets shared modules stay grouped with the static entry path instead of being split into extra chunks or pulled into dynamic chunks unnecessarily.

### Already-Loaded Algorithm

Consider this module graph:

```mermaid
flowchart LR
  main["main entry"]
  shared["shared.js"]
  route["route.js dynamic entry"]

  main -->|"static import"| shared
  main -.->|"dynamic import()"| route
  route -->|"static import"| shared
```

When `route.js` runs, `main` has already loaded `shared.js`. So `shared.js` does not need to keep the `route` entry bit.

Before reduction:

```txt
shared.js dependent entries = { main, route }
route.js  dependent entries = { route }
```

After reduction:

```txt
shared.js dependent entries = { main }
route.js  dependent entries = { route }
```

That lets normal chunk grouping put `shared.js` with `main`, while `route.js` imports the binding from the already-loaded chunk instead of forcing an extra shared chunk.

The pass works on temporary **atoms**, where each atom is a group of modules with the same dependent-entry bitset:

```txt
atom(shared.js) = {
  modules: [shared.js],
  dependentEntries: { main, route }
}
```

Then it computes the atoms loaded before each dynamic entry executes:

```txt
staticLoaded[entry] = atoms loaded by the entry's static dependency graph
alreadyLoaded[dynamicEntry] = intersection of loaded atoms from every importer entry
```

The key detail is the **intersection**. If `route.js` can be dynamically imported by both `main1` and `main2`, an atom can only drop the `route` bit if it is already loaded by **both** importers.

Pseudo-code:

```txt
atoms = groupModulesByDependentEntryBits()
staticLoaded = computeStaticLoadedAtomsForEachEntry()

for each included dynamic import:
  dynamicImporters[dynamicEntry].add(importerEntry)

queue = all dynamic entries

while queue is not empty:
  dynamicEntry = queue.pop()

  nextAlreadyLoaded = intersection over importerEntry in dynamicImporters[dynamicEntry] of:
    staticLoaded[importerEntry] union alreadyLoaded[importerEntry]

  if nextAlreadyLoaded changed:
    alreadyLoaded[dynamicEntry] = nextAlreadyLoaded

    for childDynamicEntry imported by dynamicEntry:
[...]

PR #9466, opened 2026-05-19
fix(chunk-optimization): keep CJS facade wrappers in shared chunks when dynamic siblings would back-import them

### LINKED ISSUE #8361: [Bug]: __commonJSMin is not a function since rc4

### LINKED ISSUE #9441: [Bug]: CJS subpath import in lazy chunk throws "X is not a function" — chunk-split interop mismatch in rolldown 1.0.0 / 1.0.1
## Reproduction link or steps

**Repo (starting point, see "partial repro" note below):**
<https://github.com/thedoublejay/rolldown-chunk-split-cjs-repro>

**Open in Stackblitz:**
<https://stackblitz.com/github/thedoublejay/rolldown-chunk-split-cjs-repro>

```bash
git clone https://github.com/thedoublejay/rolldown-chunk-split-cjs-repro
cd rolldown-chunk-split-cjs-repro
npm install
npm run build       # vite 8.0.13, rolldown 1.0.1
npm run preview     # browse http://localhost:4173
```

The repo sets up the same conditions where a real-world Vite 8.0.13 build broke:

- 3 × `React.lazy(() => import(...))` boundaries, all importing the same `lodash/<sub>` CJS submodules (forces rolldown to extract them to a shared vendor chunk)
- ESM libs (`@mui/material`, `@tanstack/react-query`, `clsx`, `@emotion/react`) mixed with the CJS modules in the shared chunk, matching the real failing chunk's shape
- `tss-react/mui` `withStyles` in a lazy component (matches the legacy component that surfaced the bug)
- `vite.config.js` mirrors the failing config's distinctive flags: `react({ jsxImportSource: "@emotion/react" })` + `rolldown.moduleTypes: { ".js": "jsx" }`

### **Important — partial repro**

This standalone repo **does not yet trigger the broken emission on its own** — rolldown emits a consistent interop shape here. The same Vite/rolldown versions + the same lodash subpaths + the same lazy boundary structure DO reproduce in a larger codebase (~5000 modules / ~1200-source vendor chunks). The trigger appears to depend on graph shape we couldn't isolate down to a small case.

I'm filing this rather than waiting because the production-side evidence below is concrete (minified output + bisect-narrowed regression range), and a maintainer hint at the chunking heuristic that changed between rolldown `1.0.0-rc.18` and `1.0.0` / `1.0.1` would let me push the repo over the edge. Happy to invest more time on shrinking the case.

## What is expected?

When rolldown splits a CJS module into a separate output chunk, the import-site code and the export-side emission should agree on the interop shape:

- Either both sides treat the export as a `__commonJSMin(...)` lazy-wrapper function, **or**
- both sides treat the export as a static namespace value.

This is what rolldown `1.0.0-rc.18` (shipped in Vite `8.0.11`) does, and what the same source produces correctly there.

In rolldown `1.0.0` / `1.0.1`, when a CJS module ends up in a separate output chunk (not the main bundle), the importer and exporter disagree:

- The **import-site** emits `a(c())` — i.e. calls `c` as a `__commonJSMin()` lazy-wrapper, then passes the result to `_interopRequireDefault`.
- The **export-side** emits the resolved module namespace directly as the export, not wrapped in a callable getter.

Result: `TypeError: <minified> is not a function` thrown synchronously at top-level of the lazy chunk (before any of its `export` bindings evaluate), which surfaces as a `React.lazy` rejection caught by the nearest `ErrorBoundary`.

### Real-world emitted code (anonymized)

```js
// Importing chunk (lazy)
import { Vt as t, fi as n, p as r, q as i, vi as a } from "./main.js";
import { n as s } from "./shared-A.chunk.js";
import { T as c } from "./shared-B.chunk.js";   // ← shared CJS+ESM vendor chunk

var l = a(n()),     // ← works: n() returns the __commonJSMin wrapper from main
    u = a(i()),     // ← works: i() returns the __commonJSMin wrapper from main
    d = a(c())      // ← throws: c is not callable, exporter emitted a value
```

### Browser console

```
TypeError: c is not a function
    at app.<hash>.chunk.js:1:272
Caused by: React ErrorBoundary TypeError: c is not a function
    at Lazy (<anonymous>)
    at Suspense (<anonymous>)
```

Reproducible with both `lodash@4.17.23` and `lodash@4.18.1`, both `@vitejs/plugin-react@6.0.0` and `@vitejs/plugin-react@6.0.2`, so neither of those is the variable.

## System Info

```Shell
System:
  OS: macOS 26.5
  CPU: arm64 Apple Silicon
Binaries:
  Node: 24.14.1
  npm: 11.11.0
  Yarn: 4.13.0
npmPackages (broken configuration):
  vite: 8.0.13
  rolldown: 1.0.1 (transitive via vite)
  @vitejs/plugin-react: 6.0.2
  react / react-dom: 19.2.x
```

### Bisect: which Vite/rolldown versions are affected?

| Vite version | Bundled rolldown | Status |
|---|---|---|
| 8.0.1  | `1.0.0-rc.10` | ✅ works |
| 8.0.11 | `1.0.0-rc.18` (Apr 29, 2026 — last RC) | ✅ works |
| 8.0.12 | `1.0.0` (May 7, 2026 — first stable) | ⚠️ likely affected (same rolldown family) — not directly retested |
| 8.0.13 | `1.0.1` (May 13, 2026) | ❌ reproduces |

Bisect was carried out against the real-world failing app: same source, same install, only the Vite/rolldown version changed.

### Workaround in use

Pin Vite to exact `8.0.11`.

### Related

- [#9407](https://github.com/rolldown/rolldown/issues/9407) — distinct bug, same rolldown 1.0.1 family: `require("react")` left as runtime `__require` even when react is inlined elsewhere in the same build. Different surface (SSR + dangling `__require`) but same theme of CJS-interop emission inconsistency across the bundle graph.
- [#7449](https://github.com/rolldown/rolldown/issues/7449) — incorrect execution order with CJS + advancedChunks (different mechanism, but also CJS chunking-related).

### Triage hint

If a maintainer can hint at what specifically changed in chunk emission between rolldown `1.0.0-rc.18` and `1.0.0` / `1.0.1` for shared CJS modules — particularly when a vendor chunk mixes ESM and `__commonJSMin`-wrapped CJS modules — I should be able to take the linked repo over the edge into a clean standalone repro.

kjanoudi:
Closes #9441

## Problem

Since rolldown 1.0.1 (PR #9305 + the lock-in fixture from #9351), CJS-facade modules shared between a static entry and one or more dynamic-entry chunks get **merged into the static entry**. Sibling dynamic chunks then emit:

```js
// dynamic-child.js
import { t as require_X } from "./entry.js";
var import_X = require_X();   // top-level invocation
```

Under pure ESM, the dynamic edge postpones the child's evaluation until after the entry's body has finished, so the `var require_X = __commonJSMin(...)` slot in the entry has run and the binding is initialized. But under **static-wrapper composition** — Cloudflare Workers SSR + react-router 7 static route manifests, miniflare, similar setups that statically wire all chunks — the child chunk's body evaluates before the entry's body completes, hitting TDZ:

```
TypeError: require_X is not a function
```

Manifests as `Cloudflare 1101` in real apps and `__exportAll is not a function` / `e is not a function` etc. in the issue reports collected on #9441.

## Minimal reproduction

https://github.com/kjanoudi/rolldown-9441-repro — the source is lifted verbatim from this repo's regression-lock fixture `crates/rolldown/tests/rolldown/optimization/chunk_merging/cjs_facade_reexport_merges_into_entry/`. Two CJS facades (`a.js`, `b.js`), each `module.exports = require('./X-impl.js')`; entry statically imports both and dynamically imports a route chunk that also imports both.

- vite 8.0.13 (rolldown 1.0.1) → both facades merged into entry, route + child statically back-import wrappers from entry.
- vite 8.0.12 (rolldown 1.0.0) → `b` facade has its own chunk, no back-import.

## Discriminator

PR #9305 specifically fixed #8361 — a different cycle pattern where a CJS facade chunk had a static back-edge into the main chunk. A blanket "don't dedupe wrapped atoms" gate re-introduces that 8361 cycle. The two cycle patterns differ in whether the dropped dynamic entries are **effective** or **ineffective**:

- **#8361 / #8361_2**: dynamic imports collapse via `INEFFECTIVE_DYNAMIC_IMPORT` because the dynamic-entry's own module body is also statically reachable from the surviving parent. No separate sibling chunk emerges → no back-import → safe to reduce.
- **#9441**: dynamic entries are effective (reached only via `import()`). The sibling chunks emerge separately and statically back-import the wrapper from the parent → unsafe.

The pre-chunk-formation equivalent of `detect_ineffective_dynamic_imports.rs`'s check is: the dynamic-entry's module carries more than just its own entry bit in `index_splitting_info[d_module].bits`. At dedupe time the splitting-info writes are deferred to the post-loop assignment, so this view is still the un-reduced one.

### 1. `dynamic_already_loaded.rs::can_use_reduced_dependent_entries`

Before falling through to the existing checks, if the atom contains any wrapped module (`wrap_kind != WrapKind::None`), refuse the reduction when **any** removed dynamic-entry is effective (its module body has bits beyond its own entry bit).

This keeps the wrapper atom at multi-entry bits, so the chunker doesn't collapse it into the static entry's chunk.

### 2. `chunk_optimizer.rs::try_insert_into_existing_chunk`

The downstream common-chunk-merge pass would still hoist the temp common chunk into the user-defined entry chunk. Refuse this merge when the candidate `info.modules` contain any wrapped module **and** a dynamic-entry chunk depends on the candidate.

Both gates target the same "wrapped + effective dynamic sibling" hazard. Pure-ESM atoms are untouched. The #8361 / #8361_2 cycles remain fixed.

## Result

The lock-in fixture's `artifacts.snap` now emits:

```
a.js   Common Chunk: [Shared-By: entry.js, route.js]
b.js   Common Chunk: [Shared-By: entry.js, route.js, child.js]
entry.js  imports require_a from a.js, require_b from b.js  ← leaf imports
route.js  imports require_a from a.js, require_b from b.js  ← leaf imports
child.js  imports require_b from b.js                       ← leaf import
```

No entry↔leaf back-edges. Cycle eliminated.

Other affected snapshots (5 fixtures: `2038/b`, `2085`, `8053`, `strict_execution_order/{esbuild_issue_2598,issue_4684}`) are the same pattern — a wrapped module that PR #9305 had hoisted into a parent chunk now lives in a dedicated common chunk. All test executions still pass (cycle assertions in `issues/8361`, `issues/8361_2` remain green).

## Test plan

- [x] `pnpm install && cargo test --test integration -p rolldown` → 1715 passing, 0 failing (only `test262::test262_module_code` fails — unrelated, requires the test262 submodule).
- [x] `issues/8361` and `issues/8361_2` cycle assertions remain green (PR #9305's intent preserved).
- [x] Lock-in fixture `cjs_facade_reexport_merges_into_entry` snap updated — no more back-imports.

Marked as draft to allow maintainer review of the predicate's wording before landing.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -356,6 +356,7 @@ impl GenerateStage<'_> {
     bits_to_chunk: &mut FxHashMap<BitSet, ChunkIdx>,
     input_base: &ArcStr,
     temp_chunk_graph: &mut ChunkOptimizationGraph,
+    index_splitting_info: &IndexSplittingInfo,
   ) {
     let static_entry_chunk_reference: FxHashMap<ChunkIdx, FxHashSet<ChunkIdx>> =
       self.construct_static_entry_to_reached_dynamic_entries_map(chunk_graph);
@@ -397,14 +398,14 @@ impl GenerateStage<'_> {
         }
         let chunk_idxs: Vec<_> = bits.index_of_one().map(ChunkIdx::from_raw).collect();
 
-        let merge_target = Self::try_insert_into_existing_chunk(
+        let merge_target = self.try_insert_into_existing_chunk(
           &chunk_idxs,
           &static_entry_chunk_reference,
           chunk_graph,
-          &self.link_output.module_table,
           &dynamic_entry_to_dynamic_importers,
           temp_chunk,
           temp_chunk_graph,
+          index_splitting_info,
         );
 
         Some((bits.clone(), *temp_chunk_idx, chunk_idxs, merge_target))
@@ -557,15 +558,18 @@ impl GenerateStage<'_> {
   /// entry chunk when possible, rather than creating a separate common chunk.
   ///
   /// Returns `Some(ChunkIdx)` if a suitable merge target is found, `None` otherwise.
+  #[expect(clippy::too_many_arguments)]
   fn try_insert_into_existing_chunk(
+    &self,
     chunk_idxs: &[ChunkIdx],
     entry_chunk_reference: &FxHashMap<ChunkIdx, FxHashSet<ChunkIdx>>,
     chunk_graph: &ChunkGraph,
-    module_table: &ModuleTable,
     dynamic_entry_to_dynamic_importers: &FxHashMap<ModuleIdx, FxHashSet<ModuleIdx>>,
     info: &ChunkCandidate,
     temp_chunk_graph: &ChunkOptimizationGraph,
+    index_splitting_info: &IndexSplittingInfo,
   ) -> Option<ChunkIdx> {
+    let module_table = &self.link_output.module_table;
     let mut user_defined_entry = vec![];
     let mut dynamic_entry = vec![];
     for &idx in chunk_idxs {
@@ -582,6 +586,26 @@ impl GenerateStage<'_> {
         }
         ChunkKind::Common => return None,
       }
+    }
+    // rolldown#9441: don't hoist a wrapped atom into the parent entry when an
+    // *effective* dynamic-entry sibling depends on it (mirrors the predicate
+    // in dynamic_already_loaded::can_use_reduced_dependent_entries).
+    if info
+      .modules
+      .iter()
+      .any(|module_idx| !matches!(self.link_output.metas[*module_idx].wrap_kind(), WrapKind::None))
+      && dynamic_entry.iter().any(|dyn_chunk_idx| {
+        let Some(dyn_chunk) = chunk_graph.chunk_table.get(*dyn_chunk_idx) else {
+          return true;
+        };
+        let ChunkKind::EntryPoint { module: dyn_module_idx, .. } = dyn_chunk.kind else {
+          return true;
+        };
+        let d_bits = &index_splitting_info[dyn_module_idx].bits;
+        d_bits.bit_count() == 1 && d_bits.has_bit(dyn_chunk_idx.raw())
+      })
+    {
+      return None;
     }
     let user_defined_entry_modules = Self::collect_entry_modules(&user_defined_entry, chunk_graph)?;
 
--- a/crates/rolldown/src/stages/generate_stage/code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/code_splitting.rs
@@ -953,6 +953,7 @@ impl GenerateStage<'_> {
         bits_to_chunk,
         input_base,
         &mut temp_chunk_graph,
+        index_splitting_info,
       );
 
       self.optimize_facade_entry_chunks(
--- a/crates/rolldown/src/stages/generate_stage/dynamic_already_loaded.rs
+++ b/crates/rolldown/src/stages/generate_stage/dynamic_already_loaded.rs
@@ -3,7 +3,7 @@ use std::collections::VecDeque;
 use oxc_index::{IndexVec, index_vec};
 use rolldown_common::{
   ChunkIdx, ImportKind, ImportRecordIdx, ImportRecordMeta, ModuleIdx, PreserveEntrySignatures,
-  StmtInfoIdx,
+  StmtInfoIdx, WrapKind,
 };
 use rolldown_utils::BitSet;
 use rustc_hash::{FxHashMap, FxHashSet};
@@ -75,6 +75,7 @@ impl GenerateStage<'_> {
           &atoms[atom_idx].dependent_entries,
           chunk_graph,
           &dynamic_entry_modules_by_entry,
+          index_splitting_info,
         )
         && (!self.should_check_reduced_atom_static_cycle(
           &original_dependent_entries,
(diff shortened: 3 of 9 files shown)


jacobg:
@kjanoudi Thanks. Do you mind publishing tp pkg.pr.new so I can test?
https://github.com/rolldown/rolldown/actions/workflows/publish-to-pkg.pr.new.yml

IWANABETHATGUY:
I don't think this is the correct fix.

IWANABETHATGUY:
https://github.com/IWANABETHATGUY-reproduction/rolldown-9441-repro/tree/main, I can't reproduce your runtime error. I put more details in https://github.com/IWANABETHATGUY-reproduction/rolldown-9441-repro/blob/main/FINDINGS.md

kjanoudi:
Thanks, closing the PR given this:
> If a Cloudflare Workers / react-router production failure is real, the proximate cause is host-side leaf pre-evaluation interacting with the cycle — not a Rolldown-unique bug in the bundled output.

IWANABETHATGUY:
> Thanks, closing the PR given this:
> 
> > If a Cloudflare Workers / react-router production failure is real, the proximate cause is host-side leaf pre-evaluation interacting with the cycle — not a Rolldown-unique bug in the bundled output.

If you could provide us with a proper reproduction, I would like to take a look and try to fix it.

IWANABETHATGUY:
> Thanks, closing the PR given this:
> 
> > If a Cloudflare Workers / react-router production failure is real, the proximate cause is host-side leaf pre-evaluation interacting with the cycle — not a Rolldown-unique bug in the bundled output.

Can you tell us how to reproduce the cloudflare workers/ react-router runtime error ?

PR #9477, opened 2026-05-20
docs(lazy-barrel): expand reasoning behind LARGE_BARREL_MODULES advice

shulaoda:
(Not a blocker) I think we should add the reasons a bit more here later on.

_Originally posted by @sapphi-red in https://github.com/rolldown/rolldown/pull/9236#discussion_r3233274448_
            

## Summary

Follow-up to #9236 addressing [sapphi's review comment](https://github.com/rolldown/rolldown/pull/9236#discussion_r3233274448) ("I think we should add the reasons a bit more here later on").

When `lazyBarrel` is enabled and Rolldown emits the `LARGE_BARREL_MODULES` advice, three questions naturally come up:

1. Why is there still overhead if lazy barrel is already on?
2. Why is the recommended fix a plugin instead of built-in behavior?
3. How do I dismiss the message if I have accepted the cost?

This PR adds a dedicated **Large barrel modules** section to `docs/in-depth/lazy-barrel-optimization.md` that answers all three, and updates the `LargeBarrelModules` event-kind doc comment to link to it.


sapphi-red (approved):
Thanks!

PR #9481, opened 2026-05-20
perf(transform): reserve scoping headroom for the transformer

Dunqing:
## Summary

Reserve 50% excess capacity on the `SemanticBuilder` calls that feed `Transformer::build_with_scoping`. The transformer grows the scoping `Vec`s in place — TS enums, decorators, JSX and ES lowering all add symbols/refs/scopes — so pre-reserving headroom avoids realloc mid-traversal.

## Sites

- `crates/rolldown/src/utils/pre_process_ecma_ast.rs` — gated on `is_not_js || should_transform_js || contains_script_closing_tag`. Plain JS modules skip the headroom (`with_excess_capacity(0.0)` is the default no-op).
- `crates/rolldown_plugin_vite_transform/src/lib.rs` — always transforms, no gate.
- `crates/rolldown_common/src/utils/enhanced_transform.rs` — two sites: first build + post-define fallback. The inject-only build is left alone (handful of bindings, not worth 50% headroom).

## Ratio

Picked empirically by instrumenting `pre_process_ecma_ast` on rome_ts (1195 TS modules). 96% see zero growth through the transformer; the remaining 4% grow up to 50% on scopes / 25% on symbols / 20% on refs.

| ratio | modules still reallocating |
|------:|---------------------------:|
| 0.00  | 49 (4.10%) |
| 0.20  | 6  (0.50%) |
| **0.50**  | **1 (0.08%)** |
| 2.00  | 0 — but regresses perf (3× memory cost on every module to save one extra realloc) |


Dunqing:
* **#9481** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9481?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9481?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9481).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyfdev (dismissed):
We need have a way to track this, otherwise the optimization might be break without notice in the future change.

crates/rolldown/src/utils/pre_process_ecma_ast.rs
```diff
@@ -162,13 +173,8 @@ impl PreProcessEcmaAst {
...
     // Step 3: Transform TypeScript and jsx.
     // Note: Currently, oxc_transform supports es syntax up to ES2024 (unicode-sets-regex).
-    let is_not_js = !matches!(parsed_type, OxcParseType::Js);
     let mut preserve_jsx = false;
-    if is_not_js
-      || bundle_options.transform_options.should_transform_js()
-      // Run transformer on JS files containing `</script` to handle tagged template literals.
-      || contains_script_closing_tag(ast.source().as_bytes())
-    {
+    if will_run_transformer {
```
hyfdev, line 177: - `define` changes can make `scoping` become `None` before the transformer.
- Then this path rebuilds via `recreate_scoping`.
- `recreate_scoping` uses `with_stats(self.stats)`, so `with_excess_capacity(0.5)` does not apply.
- Browser builds commonly have the default `process.env.NODE_ENV` define, so this is not a cold path.
- Can we reserve the same 50% headroom for the transformer rebuild path too?

shulaoda:
Discuss with @Dunqing some details, no blockers for me

shulaoda (approved):

crates/rolldown/src/utils/pre_process_ecma_ast.rs
Dunqing: I will take a careful look later.

PR #9486, opened 2026-05-20
fix(sourcemap): preserve names through `collapse_sourcemaps`

eoin:
This change results in `collapse_sourcemaps` more closely mirroring the original rollup implementation, fixing an issue with sourcemap generation. `collapse_sourcemaps` previously kept only `first_map.names`, silently dropping any identifier the minifier (or an intermediate transform) renamed. As a result, Vite 8 bundles emit near-empty `names` arrays, breaking downstream consumers (e.g. Sentry) that rely on the names array to recover original identifier names from minified positions. This seems to be a version 8 regression.

### DIFF
--- a/crates/rolldown_sourcemap/src/lib.rs
+++ b/crates/rolldown_sourcemap/src/lib.rs
@@ -56,40 +56,62 @@ pub fn collapse_sourcemaps(sourcemap_chain: &[&SourceMap]) -> SourceMap {
   let first_map = sourcemap_chain.first().expect("sourcemap_chain should not be empty");
   let chain_without_last = &sourcemap_chain[..sourcemap_chain.len() - 1];
 
-  // Pre-compute lookup tables in reverse order so we avoid reversing on every token lookup.
-  let sourcemap_and_lookup_table: Vec<_> = chain_without_last
+  // Concatenate each map's names into one pool and record the starting index of each map's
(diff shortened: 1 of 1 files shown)


eoin:
It looks like this fixes a regression introduced in https://github.com/vitejs/vite/commit/47a4d5069.

 | Commit | Bundler | `with-define-object-*.js.map` `names` |
  | --- | --- | --- |
  | [`aa4739ab3`](https://github.com/vitejs/vite/commit/aa4739ab3) | rollup | `["main", "mainInner", "define_testDefineObject_default"]` |
  | [`47a4d5069`](https://github.com/vitejs/vite/commit/47a4d5069) | rolldown 1.0.0-beta.7 | `[]` |
  | `main` | | `[]` |
  | This branch | | `["main", "mainInner", "\"hello\""]` |

This change results in `names` getting populated again, triggering a snapshot failure [in `js-sourcemap.spec.ts`](https://github.com/rolldown/rolldown/actions/runs/26172384420/job/76993897217?pr=9486#step:7:415).

I'm not sure why `\"hello\"` replaces `define_testDefineObject_default`, but I suspect that's a separate issue.

IWANABETHATGUY:

1. Reduce allocations.
2. Improve performance.

IWANABETHATGUY (approved):

Issue #4857, opened 2025-06-07
[Feature Request]: import defer


Boshen:
### What problem does this feature solve?

Proposal repo: https://github.com/tc39/proposal-defer-import-eval
TypeScript PR: https://github.com/microsoft/TypeScript/pull/60757
Rollup issue: https://github.com/rollup/rollup/issues/5932
Webpack PR: https://github.com/webpack/webpack/pull/16567

Oxc has this syntax enabled by default already.

### What does the proposed API look like?

Up to discussion.

PR #9503, opened 2026-05-21
feat: drop `defer` keyword

TheAlexLichter:
Until real support (https://github.com/rolldown/rolldown/issues/4857) of `import defer`, this PR should ensure that `defer` is always dropped.


sapphi-red:
I think we can have it as a workaround but I think we should emit a warning as it changes the semantics (the side effects runs immediately rather than when the imported var is used).

crates/rolldown_common/src/utils/enhanced_transform.rs
```diff
@@ -37,6 +39,34 @@ use crate::inner_bundler_options::types::tsconfig_merge::merge_transform_options
...
+    }
+    walk_mut::walk_import_expression(self, it);
+  }
+}
+
+fn drop_import_defer_phase(program: &mut ast::Program<'_>) -> Vec<oxc::span::Span> {
+  let mut dropper = DropImportDeferPhase { spans: vec![] };
+  dropper.visit_program(program);
+  dropper.spans
+}
```
IWANABETHATGUY: I prefer not adding any extra passes for the oxc transform on the rolldown side.

IWANABETHATGUY: This makes the transform in oxc and `transform` in rolldown more consistent.

TheAlexLichter: fair point! I'll double check and send a PR to oxc if needed for this part.

IWANABETHATGUY (approved):

PR #9504, opened 2026-05-21
test: failing test for #9441

TheAlexLichter:
https://github.com/rolldown/rolldown/issues/9441#issuecomment-4511142252


IWANABETHATGUY:
I can't reproduce the runtime error with this test case, also the generated chunk did not have a static circular reference, we probably need to find the minimum reproduction that could reproduce the runtime error described in https://github.com/rolldown/rolldown/issues/9441#issuecomment-4511142252

TheAlexLichter:
(Updated the case with a better test)

crates/rolldown/tests/rolldown/function/chunk_optimization/.cjs_wrapper_must_not_be_hoisted_to_entry_cycle/_config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "snapshot": false
+}
```
IWANABETHATGUY, line 3: I prefix the test case with `.` so we can merge it first

IWANABETHATGUY (approved):

Issue #4788, opened 2025-06-02
[Feature Request]: reduce the number of chunks by merging sideeffect free chunks


sapphi-red:
### What problem does this feature solve?

~~The aim is to improve the output chunking to follow the limitation set by `advancedChunks` option.~~ The aim is to improve the output chunking when the `advancedChunks` option is not specified.

Currently, IIUC Rolldown does not try merging sideeffect free chunks. Without doing this, many small chunks would be generated when there's multiple entries.

For example, for the following modules, Rolldown currently generates 3 entry chunks and 2 shared chunks.

```mermaid
flowchart LR
  page-a --> common1
  page-b --> common1
  page-b --> common2
  page-c --> common2
```

If `common1` and `common2` don't have a side effect, this can be improved to 3 entry chunks and 1 shared chunk. This is what rollup does when `output.experimentalMinChunkSize` is set ([bundler explorer](https://bundler.sxzz.dev/#eNqlUctugzAQ/JWVLyQSJQlHpJ6qHqMeekWqiFlSt2AjY7dREf9ePyCBJESKcsDIs+vZmdmWFCRpSZ3t8SmLvhp74SQZASGh5s6qWkgFLVBRVYJvoINCigqCaNUjpjdIZcrtRwVvRIlRKfaLwFMFy6GIB0dVaE4VExwk8hzlYgmtrYKZcHrcc7u3nT2MHCSJkhq70Ivcnat2wL2qzzrji874tr/dQ/7CYewto/TcqAOuGr1bPn1E/pzqU8SD7BHidffDLKc9/ZKeIR06U+JYj2TWxpTMIbNk8YgsPpFJUZa5+OWR6S3YfkR6pTIhz7HIdGmS9mEIrWqtkuEKkOU/GaeYv3xq/t2MCgAV4+/sDxPYrD/Wa/eFk+pW5LrE2R4XsPn1wGoFRhNKViFXWXkcZXCARklG1esBqbZLfJNmhQnYvRybHJ1fWp+Irq/nMcHvSmMscMu4C2XGn5dDuq77BxOmi/s=)).
[...]

PR #9519, opened 2026-05-22
feat(chunking): add min chunk size common chunk merging

aminpaks:
## Summary

Addresses #4788.

Adds opt-in `experimental.chunkOptimization.minChunkSize` support for coalescing small, side-effect-free, automatically generated common chunks before chunk materialization.

The default is `0`, so existing output is unchanged unless users explicitly opt in.

## Details

Example usage:

```ts
experimental: {
  chunkOptimization: {
    minChunkSize: 10_000,
  },
}
```

The new pass is intentionally conservative:

- only runs when `minChunkSize > 0`
- only considers automatically generated common chunks
- does not run when manual/advanced chunking is configured
- is disabled by `chunkOptimization: false` or `mergeCommonChunks: false`
- uses approximate source bytes as the threshold
- requires the source chunk and its static dependency closure to be side-effect-free
- keeps dynamic imports as loading boundaries
- bails out under the existing TLA safety gate
- rejects merges that would create a static chunk cycle

This implements the safe common-to-common case from #4788 without claiming full Rollup `experimentalMinChunkSize` parity.

## Testing

- `cargo test -p rolldown min_size_common_merge --lib -- --nocapture`
- `cargo test -p rolldown --test integration 4788_min_chunk_size -- --nocapture`
- `cargo test -p rolldown --test integration 4788_min_chunk_size_tla_bailout -- --nocapture`
- `cargo fmt --all --check`
- `vp run --filter rolldown-tests test:main build-api/min-chunk-size.test.ts`
- `vp check packages/rolldown/src/options/input-options.ts packages/rolldown/tests/build-api/min-chunk-size.test.ts crates/rolldown/tests/rolldown/issues/4788_min_chunk_size_tla_bailout/_test.mjs crates/rolldown/tests/rolldown/issues/4788_min_chunk_size_tla_bailout/_config.json`

## AI disclosure

AI assistance was used while developing and reviewing this change. I reviewed the implementation, validated the behavior locally, and am responsible for the submitted PR.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -43,14 +43,23 @@ struct CommonChunkMerge {
 /// A lightweight representation of a chunk used during optimization passes.
 struct ChunkCandidate {
   modules: Vec<ModuleIdx>,
+  bits: BitSet,
+  approx_size: u32,
   /// Whether this chunk needs to be created in the final chunk graph.
   needs_creation: bool,
+  /// Redirect target when this temporary chunk was merged into another
+  /// temporary chunk before materialization.
+  merged_into: Option<ChunkIdx>,
   dependencies: FxHashSet<ChunkIdx>,
   /// Whether any module in this chunk has side effects.
   has_side_effects: bool,
 }
 
 type TempIndexChunks = IndexVec<ChunkIdx, ChunkCandidate>;
+
+fn module_size(module_table: &ModuleTable, module_idx: ModuleIdx) -> u32 {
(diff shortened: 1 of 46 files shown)


tathyajoshi:
@IWANABETHATGUY @hyf0 would appreciate if we could get a review on this soon!

hyfdev:
Will get to this when having bandwidth. A lot of things are ongoing, would appreciate if you could understand the situation.

aminpaks:
> Will get to this when having bandwidth. A lot of things are ongoing, would appreciate if you could understand the situation.

Totally. No rush.

hyfdev:
There's a significant redesign around the whole chunking https://github.com/hyfdev/rolldown-chunking-c4. Will get back to this when the redesign work it's done.

Thanks for your effort.

PR #9522, opened 2026-05-23
fix: apply transform.dropLabels before scanning (#9521)

### LINKED ISSUE #9521: [Bug]: `transform.dropLabels` does not prevent dynamic imports inside dropped blocks from being resolved

```ts
// main.ts
DEBUG: {
    import('./debug.ts').then(({ init }) => init());
}
```

```ts
// debug.ts
import { SomeClass } from 'some-uninstalled-package';
export function init() { ... }
```

```ts
// rolldown.config.ts

const build = await rolldown({
    input: 'main.ts',
    transform: { dropLabels: ['DEBUG'] },
});
await build.write({ file: 'out.js', format: 'esm' });
```

`some-uninstalled-package` is not installed. Expected: build succeeds because the `DEBUG:` block is dropped. Actual: build fails with a resolution error for `some-uninstalled-package`.

### What is expected?

`transform.dropLabels` removes the labeled block from the source before the module is scanned for imports. The dynamic `import('./debug.ts')` inside the `DEBUG:` block is therefore never seen by the module graph traversal, and `debug.ts` (along with its transitive dependencies) is never resolved or bundled.

This matches Rollup's behavior: transform hooks run on a module's source **before** that source is parsed for its imports. If a transform removes an `import()` call, Rollup does not follow it.

Rolldown resolves dynamic imports **before** applying `transform.dropLabels`. The `DEBUG:` block is dropped from the output, but the `import('./debug.ts')` call has already been discovered and added to the module graph. Rolldown then attempts to resolve `debug.ts` and all its transitive dependencies, including `some-uninstalled-package`, causing a build failure.

This also means that even when the package IS resolvable, its code ends up in the module graph and must be explicitly excluded via a virtual stub plugin- defeating the purpose of label-based dead code elimination for dynamic imports.

### System Info

```Shell
rolldown: 1.0.2
```

### Any additional comments?

Discovered while implementing a debug/release build split for a Minecraft scripting addon. The intent was to use `DEBUG:` labels to gate debug-only imports so the release build has zero trace of the debug module chain. Static imports were already known to require special handling, but dynamic imports inside dropped labels were expected to be transparently excluded.

The workaround is a virtual stub plugin that intercepts the problematic package's resolution and returns empty exports:

```ts
const stub = {
    name: 'stub',
    resolveId: (id) => id.includes('some-uninstalled-package') ? '\0stub' : undefined,
    load: (id) => id === '\0stub' ? 'export const SomeClass = undefined;' : undefined,
};
```

This works but is not maintainable- it requires manually tracking every export of every debug-only dependency. The root fix would be to apply `dropLabels` (and other built-in OXC transforms in `InputOptions.transform`) before the module is scanned for imports, consistent with how Rollup processes plugin `transform` hooks.

IWANABETHATGUY:
Move the transform.dropLabels pass from the module finalizer into PreProcessor so labeled statements are stripped before AstScanner::scan runs. Dynamic imports and require() calls nested inside dropped labels no longer enter the module graph, so unresolvable transitive deps stop breaking the build.

Closes #9521.

As a side effect, the redundant finalizer block is removed and spurious UNRESOLVED_IMPORT warnings for dropped-label requires disappear (see updated esbuild/dce/drop_labels snapshot).

### DIFF
--- a/crates/rolldown/src/module_finalizers/impl_visit_mut.rs
+++ b/crates/rolldown/src/module_finalizers/impl_visit_mut.rs
@@ -317,13 +317,6 @@ impl<'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'_, 'ast> {
   fn visit_statement(&mut self, it: &mut ast::Statement<'ast>) {
     _ = self.try_inline_json_module_prop(it);
 
-    if !self.ctx.options.drop_labels.is_empty() {
-      if let ast::Statement::LabeledStatement(stmt) = it {
-        if self.ctx.options.drop_labels.contains(stmt.label.name.as_str()) {
-          *it = self.snippet.builder.statement_empty(stmt.span);
-        }
-      }
-    }
     walk_mut::walk_statement(self, it);
 
     // transform top level `var a = 1, b = 1;` to `a = 1, b = 1`
--- a/crates/rolldown/src/module_loader/runtime_module_task.rs
+++ b/crates/rolldown/src/module_loader/runtime_module_task.rs
@@ -225,7 +225,7 @@ impl<Fs: FileSystem + Clone + 'static> RuntimeModuleTask<Fs> {
     let mut ast = EcmaCompiler::parse(filename, source.clone(), source_type)?;
 
     ast.program.with_mut(|fields| {
-      let mut pre_processor = PreProcessor::new(fields.allocator, false);
+      let mut pre_processor = PreProcessor::new(fields.allocator, false, None);
       pre_processor.visit_program(fields.program);
     });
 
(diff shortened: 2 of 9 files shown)


IWANABETHATGUY:
* **#9522** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9522?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9522?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9522).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

h-a-n-a (approved):

PR #9526, opened 2026-05-23
fix(oxc-runtime): route require() to CJS helper variant (#9263)

### LINKED ISSUE #9263: [Bug]: Class fields in CJS-classified modules bind _defineProperty to namespace via __toCommonJS, throwing at init

IWANABETHATGUY:
The on-disk `@oxc-project/runtime` package ships two variants of every helper: a CJS form at `src/helpers/<name>.js` that sets `module.exports = fn`, and an ESM form at `src/helpers/esm/<name>.js` that does `export default fn`. The plugin previously embedded only the ESM variant and reused it for every caller, which meant `require("@oxc-project/runtime/helpers/X")` returned the ESM namespace `{ default: fn }` rather than the function itself.

This PR embeds both variants and dispatches resolution by `ImportKind`: `require()` gets the CJS variant (callable directly), `import` / dynamic `import()` get the ESM variant. Relative imports between virtual helpers preserve the importer's directory so an ESM helper's `./typeof.js` stays in `esm/`. Fixes #9263, where `_defineProperty` resolved to the ESM namespace inside CJS-classified modules and threw `TypeError: _defineProperty is not a function` at init.


IWANABETHATGUY:
* **#9526** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9526?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9526?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9526).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyfdev:
I'm not sure this is a good direction. I'm worred that this might cause the runtime code have dual package hazard.

What will happen if the runtime code got both `import` and `require`?

> This PR embeds both variants and dispatches resolution by ImportKind: require() gets the CJS variant (callable directly), import / dynamic import() get the ESM variant. Relative imports between virtual helpers preserve the importer's directory so an ESM helper's ./typeof.js stays in esm/. 

From the PR description, I assume the anwser is that rolldown will ship both esm and cjs version code.

---

I think we need to refer to other bundlers and see how does they deal with the situation. @sapphi-red cc

IWANABETHATGUY:
> I'm not sure this is a good direction. I'm worred that this might cause the runtime code have dual package hazard.
> 
> What will happen if the runtime code got both `import` and `require`?
> 
> > This PR embeds both variants and dispatches resolution by ImportKind: require() gets the CJS variant (callable directly), import / dynamic import() get the ESM variant. Relative imports between virtual helpers preserve the importer's directory so an ESM helper's ./typeof.js stays in esm/.
> 
> From the PR description, I assume the anwser is that rolldown will ship both esm and cjs version code.
> 
> I think we need to refer to other bundlers and see how does they deal with the situation. @sapphi-red cc

## 1. Almost all helpers are pure stateless functions

I scanned all 116 ESM helpers. The overwhelming majority are pure functions like:

```js
function _defineProperty(e, r, t) { /* …pure… */ }
function _typeof(o) { /* …pure… */ }
```

For those, "two copies in one bundle" just means two equivalent function objects — no helper is compared by identity, only called, so duplication is at worst a few wasted bytes.

The exceptions worth naming explicitly:

| Helper | Why it isn't strictly pure | Risk if duplicated |
|---|---|---|
| `OverloadYield` | Constructor used as a brand. `wrapAsyncGenerator` does `o instanceof OverloadYield` to detect values produced by `awaitAsyncGenerator` / `asyncGeneratorDelegate`. | The `instanceof` check could miss values if `wrapAsyncGenerator` and `awaitAsyncGenerator` resolve to different variants and an async generator from one classification delegates to one from the other. |
| `AwaitValue` | Constructor (`this.wrapped = t`). No other helper checks `instanceof AwaitValue`, but user-emitted regenerator code might. | Same shape as `OverloadYield`, but only matters if user code does cross-variant identity checks on the wrapped value. |
| `wrapNativeSuper` | Holds a closure-local `Map` cache mapping native ctor → `Wrapper`. | Two variants → two separate caches → wrapping the same native twice yields two `Wrapper`s, so `obj instanceof Wrapper` across variants would fail. Almost never reached in practice — each transformed `extends NativeCtor` site goes through one specific call. |
| `classPrivateFieldLooseKey` | Module-level `var id = 0` counter producing `__private_<id>_<name>` keys. | Each duplicate copy restarts at 0, so the IDs are no longer globally unique. Harmless in practice because keys are scoped to the class instance they're written on. |

Within a single user module, the helper *and* its transitive `./Foo.js` imports go through the same `ImportKind` (the PR's relative-import resolver preserves the importer's directory — `esm/wrapAsyncGenerator.js` resolves `./OverloadYield.js` to `esm/OverloadYield.js`, and the CJS sibling stays in CJS). So the only way to trip the `OverloadYield` brand check is to pass a value produced by helpers from one variant into helpers from the other variant — i.e. async-generator delegation across the CJS/ESM module boundary. Niche, but worth flagging.

## 2. The on-disk package already does exactly this dispatch

Look at `@oxc-project/runtime`'s own `package.json` `exports`:
image
```json
"./helpers/defineProperty": [
  {
    "node":    "./src/helpers/defineProperty.js",       // CJS variant
    "import":  "./src/helpers/esm/defineProperty.js",   // ESM variant
    "default": "./src/helpers/defineProperty.js"
  },
  "./src/helpers/defineProperty.js"
]
```

So if `oxc-resolver` (or Node, or webpack) were resolving `@oxc-project/runtime/helpers/defineProperty` against the real package, `require()` would already give the CJS file and `import` would already give the ESM file. The PR isn't inventing a new dual-shipping policy — it's making the embedded plugin honor the policy the upstream package already encodes via `exports` conditions. Until now we were always returning the ESM variant regardless of caller, which is what causes #9263.

## 3. The CJS variant is interop-safe by construction

The on-disk CJS helper ends with:

```js
module.exports = _defineProperty;
module.exports.__esModule = true;
module.exports["default"] = module.exports;
```

So `require("…")`, `require("…").default`, and `import X from "…"` all yield the function. Two copies of this in one bundle still both behave as `fn`. There's no shape divergence to expose.

## 4. Other bundlers in this niche behave the same way

`@babel/runtime` ships the identical dual-export shape (it's where this package's structure comes from), and webpack/Rollup pick the variant via `exports` conditions on a per-caller basis — same outcome as this PR. `@swc/helpers` does the same. So even if a real app reaches one helper from both `import` and `require()` callers, it'd already end up with two copies under those bundlers.

## 5. The alternatives are worse

The other ways to fix #9263 are:

- **(a) Keep ESM-only and fix the CJS classification adapter** so `var X = (init_X(), __toCommonJS(X_exports))` binds to the function instead of the namespace. This is doable but tangled — the helper is a virtual `\0…` module, the CJS-emitter path needs to special-case it, and we'd be growing logic in the linking/codegen stages rather than where the helper lives.
- **(b) Always emit the CJS variant.** Then every ESM helper import in user code becomes a `__toESM(require_…)` shape, which the rest of tree-shaking and chunking treats as a "CJS classified" module. Helpers would no longer be statically-inlinable function decls. That regresses ESM output noticeably.
- **(c) This PR.** Match what the package's own `exports` map says. Matches Node and matches what every other bundler does when resolving this package off disk.

IWANABETHATGUY:
Another reason I don't prefer the way `Keep ESM-only and fix the CJS classification adapter` is written in https://github.com/IWANABETHATGUY-reproduction/9263-comparison/blob/main/README.md

hyf0 (approved):
Webpack + babel runtime has the same behavior as this PR. LGTM.

h-a-n-a (approved):

IWANABETHATGUY:
### Merge activity

* **May 25, 1:33 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **May 25, 1:36 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **May 25, 1:41 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

IWANABETHATGUY:
I have fixed the issues(refactor) that @shulaoda mentioned, so I merged.

sapphi-red:
I guess we can add `module` condition to the runtime package to reduce the output size. I think it's not urgent though.

PR #9527, opened 2026-05-23
test: add failing TDZ cycle case

TheAlexLichter:
Found by [the fuzzer](https://github.com/TheAlexLichter/rolldown-module-fuzzer).

Native Node and Rollup throw a TDZ `ReferenceError` for this cycle, but Rolldown currently folds the cycle and the generated bundle resolves. The reduced fixture keeps the namespace read through a dynamically imported re-export trigger.

Fuzzer repro:

```bash
vp run diff -- --fuzz --seed 339772623 --cases 1 --max-width 6 --max-depth 8 --paths static,namespace,dynamic,default,named-reexport,star-reexport
```


IWANABETHATGUY:
closed, because https://github.com/rolldown/rolldown/issues/9530

PR #9528, opened 2026-05-23
test: add failing TLA static and dynamic import facade case

TheAlexLichter:
Found by [the fuzzer](https://github.com/TheAlexLichter/rolldown-module-fuzzer).

Native Node and Rollup resolve this test case successfully, but Rolldown currently produces an unsettled top-level await warning at runtime.

Fuzzer repro:

```bash
vp run diff -- --fuzz --seed 339772549 --cases 1 --max-width 6 --max-depth 8 --paths static,namespace,dynamic,default,named-reexport,star-reexport
```


IWANABETHATGUY:
closed, because https://github.com/rolldown/rolldown/issues/9530

PR #9545, opened 2026-05-25
fix: validate return type of `external` option callback

### LINKED ISSUE #9005: The `Value is non of these types` error thrown for invalid return value in function options is difficult to figure the reason

shulaoda:
## Summary

- Add `assertCallbackReturn` utility to validate return values of user-provided option callbacks at the JS layer, before they cross to Rust
- Apply validation to the `external` option callback, producing a clear error like `Invalid return value for option "external": expected boolean, got string.` instead of the cryptic NAPI error `Value is non of these types 'bool', 'UnknownReturnValue'`
- The utility supports multiple expected types (`string | string[]`) for future use with other option callbacks

Closes #9005


shulaoda:
I’ve currently validated only the `external` case. I’ll gradually expand the validation to other function options in subsequent iterations.

sapphi-red:
Wouldn't this have additional overhead because now we have a duplicate check?

shulaoda:
> Wouldn't this have additional overhead because now we have a duplicate check?

To be honest, this should be a very cheap check, so I don’t think we need to worry about it too much.

sapphi-red:
For this specific case, I agree that it's cheap. But for example, the return value of `transform` hook wouldn't be cheap, especially when the `map` is passed as an object.
https://rolldown.rs/reference/Interface.FunctionPluginHooks#transform

shulaoda:
> For this specific case, I agree that it's cheap. But for example, the return value of `transform` hook wouldn't be cheap, especially when the `map` is passed as an object. https://rolldown.rs/reference/Interface.FunctionPluginHooks#transform

My understanding is that we currently only check the top-level type, i.e., simple categories like `boolean | number | object`. Are you suggesting that we should validate these types more strictly?

sapphi-red:
> My understanding is that we currently only check the top-level type

Oh, really? I thought napi-rs checks the nested properties as well. If napi-rs's generated code only checks the top-level type, the only concern with this approach is how to ensure the code is synced with the actual types.

shulaoda:
> > My understanding is that we currently only check the top-level type
> 
> Oh, really? I thought napi-rs checks the nested properties as well. If napi-rs's generated code only checks the top-level type, the only concern with this approach is how to ensure the code is synced with the actual types.

Sorry, what I meant is that our `assertCallbackReturn` in this PR only checks the top-level return type, while napi-rs actually validates full type consistency.

sapphi-red:
But that means that the error message would be the same if the user returns a different nested type. If this is only for the top-level type, I'm not sure if it helps a lot.

shulaoda:
> But that means that the error message would be the same if the user returns a different nested type. If this is only for the top-level type, I'm not sure if it helps a lot.

I still feel there's some mismatch here, since the Rust-side types don't always align with the JavaScript-side types, and there may be internal conversions happening. So do you think we should keep investing in this?

sapphi-red:
If you don't have any ideas, let's hold off for now.

PR #9546, opened 2026-05-25
fix(vite-resolve): preserve slash separators for Sass partial exports

cjc0013:
## Summary
- keep retry-with-prefix specifiers slash-separated for bare deep imports so package exports see keys like `./styles/_mixins` on Windows
- preserve the existing path-based prefix behavior for relative and absolute specifiers
- add a `viteResolvePlugin` fixture that resolves a Sass partial through package `exports`

## Context
This follows the maintainer note in vitejs/vite#22505 that the fix belongs in `viteResolvePlugin` exported from `rolldown/experimental`.

Refs vitejs/vite#22294 and vitejs/vite#22505.

## Validation
- `git diff --check`
- Not run locally: Rust/fixture test suites, because `cargo`, `pnpm`, and `just` are not available in this local environment.


sapphi-red (approved):
Thanks!

PR #9547, opened 2026-05-25
docs: explain the policy for Rust crates

sapphi-red:
close #9536
close #9540

PR #9553, opened 2026-05-26
chore: point published crates' readme to root README.md

Boshen:
## Summary

- Adds `readme = "../../README.md"` to every published crate in `crates/` (every crate without `publish = false`, 33 in total).
- At publish time, Cargo rewrites the path and copies the root `README.md` into the packaged tarball, so each crate's page on crates.io renders the project README instead of being empty.

Verified locally with `cargo package` on `rolldown`: the tarball contains `rolldown-0.1.0/README.md` and the packaged `Cargo.toml` is rewritten to `readme = "README.md"`.


hyfdev:
@sapphi-red is working on https://github.com/rolldown/rolldown/pull/9547. He has some thoughts on this.

Issue #8468, opened 2026-02-26
[Bug]: `paths` aliases in referenced TSConfigs are not respected by Rolldown


KTrain5169:
### Reproduction link or steps

Reproduction repo:

https://github.com/KTrain5169/rolldown-tsconfig-paths-repro

Reproduction steps and system info should be included in the README file.

### What is expected?

No warning, and Rolldown resolves the paths correctly.

### What is actually happening?

```bash
app\index.ts (1:26) [UNRESOLVED_IMPORT] Warning: Could not resolve '1/index' in app/index.ts
   ╭─[ app/index.ts:1:27 ]
   │
 1 │ import { function1 } from "1/index";
   │                           ────┬────
   │                               ╰────── Module not found, treating it as an external dependency
───╯

server\index.ts (1:26) [UNRESOLVED_IMPORT] Warning: Could not resolve '2/index' in server/index.ts
   ╭─[ server/index.ts:1:27 ]
   │
 1 │ import { function2 } from "2/index";
   │                           ────┬────
   │                               ╰────── Module not found, treating it as an external dependency
───╯

<DIR>/app.js  chunk │ size: 0.09 kB
<DIR>/server.js  chunk │ size: 0.09 kB
```

### System Info

```Shell
System:
    OS: Windows 11 10.0.26200
    CPU: (12) x64 13th Gen Intel(R) Core(TM) i5-1334U
    Memory: 2.22 GB / 15.69 GB
  Binaries:
    Node: 22.14.0 - C:\Program Files\nodejs\node.EXE
    npm: 11.7.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 10.30.0 - C:\Users\KT\AppData\Local\pnpm\pnpm.CMD
    bun: 1.3.3 - C:\Users\KT\.bun\bin\bun.EXE
  Browsers:
    Chrome: 145.0.7632.117
    Edge: Chromium (140.0.3485.54)
    Firefox: 148.0 - C:\Program Files\Mozilla Firefox\firefox.exe
  npmPackages:
    rolldown: 1.0.0-rc.5 => 1.0.0-rc.5
```

### Any additional comments?

#7177 has been closed as completed because a PR was merged that seems to have fixed it, which made me think I was doing something wrong, please tell me if that is what ends up being the case.

KTrain5169:
[...]

PR #9571, opened 2026-05-26
feat(plugin-vite-resolve): support custom tsconfig path via TsconfigDiscovery::Manual

rubnogueira:
## Summary

Rust is not my strong suit, so please review the implementation of this PR. I was having some issues regarding my CI usage where I had to use a different tsconfig.json, and I found that there was an implementation in rolldown for that, but that is not exposed to the Vite side.

I did some changes where I widened the `viteResolvePlugin` napi binding so `tsconfigPaths` accepts `boolean | { configFile: string }` instead of only a `bool`, and route the manual path through to
  `oxc_resolver::TsconfigDiscovery::Manual`.

## Motivation

The Rust scaffolding for a manually-specified tsconfig already exists everywhere except the vite-resolve plugin boundary:

  - `oxc_resolver::TsconfigDiscovery::Manual(TsconfigOptions { config_file, references })` — supported.
  - `rolldown_common::TsConfig::{Auto, Manual}` at
  [`crates/rolldown_common/src/inner_bundler_options/types/tsconfig.rs#L12-L15`](crates/rolldown_common/src/inner_bundler_options/types/tsconfig.rs#L12-L15) — used by the transform
  pipeline at [`crates/rolldown_common/src/inner_bundler_options/types/transform_options.rs#L41-L48`](crates/rolldown_common/src/inner_bundler_options/types/transform_options.rs#L41-L48)
  and the standalone resolver at [`crates/rolldown_resolver/src/resolver_config.rs#L100-L108`](crates/rolldown_resolver/src/resolver_config.rs#L100-L108).
  - `ResolverFactory`'s napi binding already exposes `tsconfig?: 'auto' | TsconfigOptions` to JS.

  The only thing missing was the `viteResolvePlugin` field. It was hardcoded to `bool` and `.then_some(TsconfigDiscovery::Auto)`, so the `Manual` arm was unreachable from JS:

  ```rust
  // crates/rolldown_plugin_vite_resolve/src/resolver.rs (before)
  tsconfig: base_options.tsconfig_paths.then_some(TsconfigDiscovery::Auto),
  ```

  ## Changes

  - `crates/rolldown_plugin_vite_resolve/src/resolver.rs` — added `TsconfigPathsOption { Disabled, Auto, Manual(PathBuf) }`; mapped the `Manual` arm to
  `TsconfigDiscovery::Manual(TsconfigOptions { config_file, references: Auto })` in `get_resolve_options`.
  - `crates/rolldown_plugin_vite_resolve/src/vite_resolve_plugin.rs` — same field-type swap on `ViteResolveResolveOptions`.
  - `crates/rolldown_plugin_vite_resolve/src/lib.rs` — re-export `TsconfigPathsOption`.
  - `crates/rolldown_binding/src/options/plugin/config/binding_vite_resolve_plugin_config.rs` — new `BindingTsconfigPathsOptions { config_file: String }` napi struct; changed the field to
  `napi::Either<bool, BindingTsconfigPathsOptions>` and converted into the new enum in the `From` impl.

  Generated TS:

  ```ts
  interface BindingTsconfigPathsOptions {
    configFile: string
  }
  // in BindingViteResolvePluginResolveOptions:
  tsconfigPaths: boolean | BindingTsconfigPathsOptions
  ```

  `true` / `false` semantics are unchanged.

  ## Related

  - vitejs/vite#22112 — feature request for `resolve.tsconfigPaths` to accept a config object so consumers can stop depending on `vite-tsconfig-paths`. Blocks on this PR.
  - vitest-dev/vitest#10176 — downstream pain point: users dynamically switching between `tsconfig.json` and `tsconfig.prod.json` (the old `esbuild.tsconfigRaw` workaround broke when
  vitest moved off esbuild).
  - vitejs/vite#22533 — PR on vite side

  ## Test plan

  - [ ] `cargo check --workspace` passes
  - [ ] `pnpm --filter rolldown build-native:debug` regenerates `binding.d.cts` with the new union type
  - [ ] Existing tests pass; `tsconfigPaths: true`/`false` still behave the same
  - [ ] Manual smoke: passing `{ configFile: '/abs/path/tsconfig.prod.json' }` to `viteResolvePlugin` no longer throws `Failed to convert napi value into rust type 'bool'`

### DIFF
--- a/crates/rolldown_binding/src/options/plugin/config/binding_vite_resolve_plugin_config.rs
+++ b/crates/rolldown_binding/src/options/plugin/config/binding_vite_resolve_plugin_config.rs
@@ -1,10 +1,10 @@
 use derive_more::Debug;
-use std::sync::Arc;
+use std::{path::PathBuf, sync::Arc};
 
 use napi::bindgen_prelude::FnArgs;
 use rolldown_plugin_vite_resolve::{
   FinalizeBareSpecifierCallback, FinalizeOtherSpecifiersCallback, OnLogCallback,
-  ViteResolveOptions, ViteResolveResolveOptions,
+  TsconfigPathsOption, ViteResolveOptions, ViteResolveResolveOptions,
 };
 
 use crate::{
@@ -143,6 +143,12 @@ impl From<BindingViteResolvePluginConfig> for ViteResolveOptions {
 
 #[napi_derive::napi(object, object_to_js = false)]
 #[derive(Debug)]
+pub struct BindingTsconfigPathsOptions {
+  pub config_file: String,
+}
+
+#[napi_derive::napi(object, object_to_js = false)]
+#[derive(Debug)]
 #[expect(clippy::struct_excessive_bools)]
 pub struct BindingViteResolvePluginResolveOptions {
   pub is_build: bool,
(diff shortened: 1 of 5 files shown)


rubnogueira:
@shulaoda 

https://github.com/rolldown/rolldown/issues/8468#issuecomment-4610344118

I just noticed this comment, do you have a chance of reviewing this PR and the related vite PR? Thank you

shulaoda:
> @shulaoda
> 
> [#8468 (comment)](https://github.com/rolldown/rolldown/issues/8468#issuecomment-4610344118)
> 
> I just noticed this comment, do you have a chance of reviewing this PR and the related vite PR? Thank you

Since this part is used by Vite, we should have Sapphi review it first.

rubnogueira:
@sapphi-red any chance that you could possibly review this PR?

I'm writing some workarounds at my company, and this would greatly simplify things. Thank you :)

sapphi-red:
It has to be discussed in the team meeting. But for your usecase, I don't think tsconfig paths is the right way. I think using `resolve.alias` or `imports` field better suits your usecase.

rubnogueira:
We override not only the paths on a specific workflow that we have but also manipulate the references array and the composite flag.

That is the main reason why we would like to provide an alternative tsconfig.json file path.

PR #9563, opened 2026-05-26
chore(deps): upgrade oxc to 0.133.0

Dunqing:
## Summary

- Bump oxc Rust crates and `@oxc-project/*` / `oxc-*` npm packages from `0.132.0` to `0.133.0`. ~~Bump `oxc_resolver` / `oxc_resolver_napi` from `11.19.1` to `11.19.2`.~~
- Regenerate `embedded_helpers.rs` to point at `@oxc-project+runtime@0.133.0` (version string only — no helper-content drift).
- Handle two new oxc fields surfaced by `cargo check`:
  - `oxc::transformer::DecoratorOptions.strict_null_checks` — hardcoded to `true` in the `From` impl (matches oxc's default). Exposing it through rolldown's wrapper would be a separate API change.
  - `oxc_minify_napi::CodegenOptions.legal_comments` — set to `None`.
- Extend `DecoratorOptionSchema` and `CodegenOptionsSchema` in `validator.ts` to mirror the new optional fields.
- Refresh 32 snapshots driven by upstream oxc PR [#22547](https://github.com/oxc-project/oxc/pull/22547) ("fix(minifier): preserve IIFE structure in DCE-only mode"). In DCE-only mode (rolldown's per-module preprocess via `Compressor::dead_code_elimination_with_scoping`), oxc no longer inlines `(() => …)()` IIFEs that carry — or could carry — `/*#__PURE__*/` annotations, so downstream tree-shaking can see them. Empty-IIFE elision still fires. Side effects: `property_read_side_effects` now correctly tree-shakes `/*#__PURE__*/ test().a.b.c;` lines and drops the previously spurious `INVALID_ANNOTATION` warnings; TS class `(() => new Foo())()` patterns and `Symbol((() => Math.random() < 0.5)() ? …)` calls are preserved as the sources actually wrote them; `dce_of_iife/diff.md` divergence vs esbuild shrunk.

## Issues fixed

Pulled in from upstream oxc fixes between 0.132.0 and 0.133.0:

- Fixes #9437 — Pure-annotated IIFE returning an array is inlined (via oxc [#22547](https://github.com/oxc-project/oxc/pull/22547))
- Fixes #9494 — Cross-module enum inlining mis-folds auto-increment member to 0 (via oxc [#22646](https://github.com/oxc-project/oxc/pull/22646))
[...]

PR #9580, opened 2026-05-27
feat(transform): respect decorator strictNullChecks option

kylecannon:
## Summary

Builds on #9563, which upgraded oxc to `0.133.0` and surfaced the `strictNullChecks` decorator-metadata option (from oxc-project/oxc#22266) in the TypeScript types, the validator, and the generated binding.

However, #9563's Rust `From<DecoratorOptions> for oxc::transformer::DecoratorOptions` hardcodes `strict_null_checks: true`, and the napi options normalizer doesn't carry the field, so `transform.decorator.strictNullChecks: false` is currently **silently ignored** through rolldown's transform/bundle pipeline (the option exists in the type surface but does nothing).

This PR wires it through so the option actually takes effect.

## What changed

- `rolldown_common::DecoratorOptions` gains a `strict_null_checks: Option<bool>` field; its `From` impl now uses `options.strict_null_checks.unwrap_or(true)` (mirroring oxc's own default) instead of the hardcoded `true`.
- `normalize_binding_transform_options` forwards `strict_null_checks` from the incoming napi options into `DecoratorOptions` (previously dropped).
- Docs (`transform.md`) + tests.

## Behavior

When `emitDecoratorMetadata` is enabled (legacy decorators):

| Source | `strictNullChecks: true` (default) | `strictNullChecks: false` |
| --- | --- | --- |
| `string \| null` | `Object` (matches `tsc` strict) | `String` (matches `tsc --strictNullChecks false` / `babel-plugin-transform-typescript-metadata`) |
| `number \| undefined` | `Object` | `Number` |

Defaults to `true`, so there is **no behavior change for existing users**; only an explicit `strictNullChecks: false` is newly honored.

> Note: this is **not** inferred from `tsconfig.json` *yet*. `oxc_resolver`'s `CompilerOptions` does not currently parse `strictNullChecks`, so (unlike `experimentalDecorators`/`emitDecoratorMetadata`) it must be set explicitly on `transform.decorator`. Documented in `transform.md`. See Follow-up below.

## Follow-up: inferring from `tsconfig.json`

[oxc-project/oxc-resolver#1166](https://github.com/oxc-project/oxc-resolver/pull/1166) (`feat(tsconfig): parse strict and strictNullChecks compiler options`) is the enabling resolver-side change. It adds `strict`/`strictNullChecks` to `CompilerOptions`, inherits them through `extends`, and exposes `effective_strict_null_checks()` implementing tsc's `strictNullChecks ?? strict` precedence.

Once it lands and is released, a small downstream follow-up here can forward `compiler_options.effective_strict_null_checks()` in `merge_transform_options_with_tsconfig` into `DecoratorOptions.strict_null_checks` (mapping `None` to the transformer default), plus add the field to the inline `BindingTsconfigCompilerOptions` path, making `strictNullChecks` auto-inferred from tsconfig like the other decorator options. This PR is intentionally scoped to the explicit option so it's useful immediately and independent of that release.

## Test plan

- [x] Rebased/merged on top of the latest `main` (which already has oxc `0.133.0`); `cargo check --workspace` passes
- [x] New `transform.test.ts` cases assert `string | null` emits `Object` by default and `String` under `strictNullChecks: false` (verified locally; the full suite runs in CI)

## AI usage disclosure


shulaoda (approved):
Thanks!

PR #9582, opened 2026-05-27
fix(testing): canonicalize temp dir early and use platform-specific separator in test262

shulaoda:
## Summary

- Move `dunce::canonicalize` from after Node.js execution to right after `create_dir_all`, so Node.js receives the resolved path (fixes Windows 8.3 short name mismatch, e.g. `RUNNER~1` vs `RUNNERADMIN`)
- Use platform-specific path separators in `normalize_output` — `/` on Unix, `\` on Windows — so the temp dir replacement actually matches Node.js error output

## Context

#9551 added `source-phase-import` test cases — the first tests whose Node.js error output contains a temp directory path. This exposed two independent bugs in `normalize_output`:

1. **Path content mismatch** (CI): `std::env::temp_dir()` can return an 8.3 short path on Windows, but `dunce::canonicalize` was called *after* Node.js execution, so the normalization pattern used the long path while Node.js had already printed the short path.

2. **Separator mismatch** (local Windows): `normalize_output` only replaced `{temp_dir}/` (forward slash), but Node.js on Windows outputs `{temp_dir}\` (backslash). The replacement never matched.

Neither bug was visible before #9551 because no prior test produced temp dir paths in its error output.

Ref: https://github.com/rolldown/rolldown/actions/runs/26496002954/job/78024279123

### DIFF
--- a/crates/rolldown/tests/integration/test262.rs
+++ b/crates/rolldown/tests/integration/test262.rs
@@ -410,10 +410,12 @@ globalThis.$DONE = function(error) {
   /// - Replaces temp directory paths with `<temp>/`
   /// - Removes stack trace lines (lines starting with "at ")
   fn normalize_output(output: &str, temp_dir: &Path) -> String {
-    let pattern = format!("{}/", temp_dir.to_string_lossy());
-    let output = output.replace(&pattern, "<temp>/");
+    let temp_dir_str = temp_dir.to_string_lossy();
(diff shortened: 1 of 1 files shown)


shulaoda:
I have to verify it on my PC locally, and it passed now 😅

shulaoda:
Merge this directly to fix the CI issue and avoid blocking the release.

PR #9590, opened 2026-05-27
feat(transform): infer decorator strictNullChecks from tsconfig

kylecannon:
## Summary

Follow-up to #9580 (merged), which added the `transform.decorator.strictNullChecks` option and the `DecoratorOptions.strict_null_checks` field. That PR noted the value could not yet be inferred from `tsconfig.json` because `oxc_resolver` did not parse `strictNullChecks`.

`oxc_resolver` now parses `strict` and `strictNullChecks` (oxc-project/oxc-resolver#1166, already on `main` via the resolver bump). This PR forwards the value, so `strictNullChecks` is now inferred from `tsconfig.json` like `experimentalDecorators` and `emitDecoratorMetadata`.

## What changed

- `merge_transform_options_with_tsconfig` reads `strictNullChecks ?? strict` (matching tsc's precedence) from the resolved tsconfig and sets `DecoratorOptions.strict_null_checks` when the option is not set explicitly.
- The explicit `transform.decorator.strictNullChecks` option still takes precedence over tsconfig (with a conflict warning when `warn_on_conflict` is set).
- When tsconfig sets neither flag, the transformer default (`true`) is kept, so existing behavior is unchanged.
- The inline tsconfig type `BindingTsconfigCompilerOptions` gains `strict` and `strictNullChecks`, mapped into `oxc_resolver::TsConfig`, so the non-disk `transform({ tsconfig: { compilerOptions } })` path works too.
- `tsconfig.md` lists the new option; tests cover inference, the `strict` fallback, the default, and explicit-option precedence.

## Behavior

With `emitDecoratorMetadata` enabled, for a `string | null` field:

| tsconfig | emitted `design:type` |
| --- | --- |
| `strictNullChecks: true` (or `strict: true`, or unset) | `Object` |
| `strictNullChecks: false` (or `strict: false`) | `String` |

`strictNullChecks` overrides `strict`, and an explicit `transform.decorator.strictNullChecks` overrides tsconfig.

## Test plan

- [x] `cargo check --workspace --locked`, `cargo clippy -p rolldown_common -p rolldown_binding --all-targets -- --deny warnings`, `cargo fmt --all --check`
- [x] `vp check` (0 type/lint errors)
- [x] Node suite (`test:main`): 752 passed. New `transform.test.ts` cases assert that tsconfig `strictNullChecks: false` (and `strict: false`) emit `String`, the default emits `Object`, and an explicit option overrides tsconfig.

## AI usage disclosure


Cargo.toml
```diff
@@ -265,10 +265,10 @@ oxc_index = { version = "4", features = [
   "rayon",
   "serde",
 ] } # https://github.com/oxc-project/oxc-index-vec
-oxc_resolver = { version = "11.19.1", features = [
+oxc_resolver = { version = "11.20.0", features = [
   "yarn_pnp",
 ] } # https://github.com/oxc-project/oxc-resolver
-oxc_resolver_napi = { version = "11.19.1", default-features = false, features = ["yarn_pnp"] }
```
Dunqing: Please revert here; I think Rolldown has already used the latest OXC resolver version.

crates/rolldown_common/src/inner_bundler_options/types/tsconfig_merge.rs
```diff
@@ -151,6 +151,29 @@ pub fn merge_transform_options_with_tsconfig(
...
+          "transform.decorator",
+          "strictNullChecks",
+          "tsconfig.json",
+          "compilerOptions.strictNullChecks",
+        )
+        .with_severity_warning(),
+      );
+    }
+  }
+
```
Dunqing, line 183: @sapphi-red can you review this PR? I think you may be more familiar with this

Cargo.toml
kylecannon: Done, reverted. Rebased on the latest `main` (now on `oxc_resolver` 11.21.0, which includes the `strict`/`strictNullChecks` parsing), so this PR no longer touches `Cargo.toml`/`Cargo.lock` and only contains the tsconfig-inference wiring.

kylecannon: pushed!

crates/rolldown_common/src/inner_bundler_options/types/tsconfig_merge.rs
```diff
@@ -151,6 +151,29 @@ pub fn merge_transform_options_with_tsconfig(
...
+  // `strictNullChecks` (falling back to `strict`, mirroring tsc's `strictNullChecks ?? strict`)
+  // controls whether `null`/`undefined` are elided from union `design:type` metadata under
+  // `emitDecoratorMetadata`. The explicit `transform.decorator.strictNullChecks` option wins; when
+  // tsconfig sets neither flag, rolldown keeps the transformer default (`true`).
+  let tsconfig_strict_null_checks = compiler_options.strict_null_checks.or(compiler_options.strict);
+  if let Some(strict_null_checks) = tsconfig_strict_null_checks {
+    if transform_options.decorator.as_ref().is_none_or(|d| d.strict_null_checks.is_none()) {
+      let mut decorator = transform_options.decorator.unwrap_or_default();
+      decorator.strict_null_checks = Some(strict_null_checks);
+      transform_options.decorator = Some(decorator);
```
sapphi-red: Do you have a reason to set the `decorator` option when the `decorator` option is not set by the user and `strictNullChecks` (or `strict`) is explicitly set? If not, I think we should only set the option if `decorator` option is set.

crates/rolldown_common/src/inner_bundler_options/types/tsconfig_merge.rs
```diff
@@ -151,6 +151,29 @@ pub fn merge_transform_options_with_tsconfig(
     }
   }
 
+  // `strictNullChecks` (falling back to `strict`, mirroring tsc's `strictNullChecks ?? strict`)
+  // controls whether `null`/`undefined` are elided from union `design:type` metadata under
+  // `emitDecoratorMetadata`. The explicit `transform.decorator.strictNullChecks` option wins; when
+  // tsconfig sets neither flag, rolldown keeps the transformer default (`true`).
+  let tsconfig_strict_null_checks = compiler_options.strict_null_checks.or(compiler_options.strict);
```
sapphi-red: The `strict` option is enabled by default only in TS 6.0+. I think we should align with TS 6.0+ and also add some docs that we align with TS 6.0+.

crates/rolldown_common/src/inner_bundler_options/types/tsconfig_merge.rs
kylecannon: Good catch. You're right that the earlier `experimentalDecorators` block leaves `decorator` as `Some` for almost every project, so the option was being attached even without decorators. Fixed in 2d90e413: I now forward `strictNullChecks` only when `emitDecoratorMetadata` is enabled (the only case it affects), mutating the existing decorator option in place rather than creating one.

crates/rolldown_common/src/inner_bundler_options/types/tsconfig_merge.rs
kylecannon: Confirmed, TS 6.0 made `strict` (and thus `strictNullChecks`) default to `true`. The unset-default here is `true`, which matches TS 6.0+. I added a code comment and a note in tsconfig.md stating we align with TS 6.0+. Let me know if you'd prefer different behavior for pre-6.0 tsconfigs (currently we always default to enabled when neither flag is set).

sapphi-red (approved):
LGTM

PR #9591, opened 2026-05-27
fix(finalizers): hoist top-level classes as `let`, not `var`

kylecannon:
## Summary

When a top-level `class X {}` is rewritten so its name binding can participate in the bundler's symbol table, `ScopeHoistingFinalizer::get_transformed_class_decl` was lowering it to `var X = class X {}`. That silently changes the binding from lexical (TDZ) to function-scoped (hoisted as `undefined`), which breaks any reference that relies on the class being unreachable before its declaration.

This PR emits `let` instead of `var` so the bundled output preserves the TDZ semantics of the original `class` keyword.

## The bug

The most visible breakage is decorator metadata under `emitDecoratorMetadata`. OXC's legacy decorator transform wraps forward-referenced class identifiers in a guard so the metadata expression evaluates safely at decoration time:

```js
_decorateMetadata("design:type", typeof LaterClass === "undefined" ? Object : LaterClass)
```

With a `let` binding, `typeof T` on a TDZ identifier throws `ReferenceError` and the application surfaces the real ordering problem. With `var`, the hoisted binding evaluates to `undefined`, the guard short-circuits to `Object`, and the metadata is silently wrong forever — frameworks like NestJS that consume `design:type` at DI resolution then resolve the wrong dependency.

The same hazard applies to any non-decorator forward reference:

```ts
class A { static b = new B() } // bundled as `var A = class { static b = new B() }`
class B {}                     // bundled as `var B = class {}` — hoisted as `undefined`
```

Under `var`, `new B()` observes `B` as `undefined` and throws `TypeError: B is not a constructor` at runtime instead of the `ReferenceError` the original source guaranteed.

## How tsc and babel handle this (reference)

Given the same source:

```ts
class Source {
  @D() laterRef!: LaterClass;
}
class LaterClass {
  tag = 'later';
}
```

**Babel** (`@babel/plugin-proposal-decorators` legacy + `babel-plugin-transform-typescript-metadata`) emits:

```js
let Source = (_dec = D(), _dec2 = Reflect.metadata("design:type", typeof LaterClass === "undefined" ? Object : LaterClass), _class = class Source { ... }, ...);
class LaterClass { ... }
```

Both bindings stay lexical — `Source` as a `let`, `LaterClass` left as a class declaration. The metadata call runs at module-init time, `LaterClass` is in TDZ at that moment, `typeof` throws `ReferenceError: Cannot access 'LaterClass' before initialization`, and the developer immediately learns about the ordering bug.

**tsc** (`experimentalDecorators` + `emitDecoratorMetadata`) emits the same structure: `class Source { ... }` interleaved with the `__decorate` / `__metadata` calls and `class LaterClass { ... }` after, again all lexical. Same `ReferenceError` surfaces.

**rolldown before this PR** lowered both top-level classes to `var Source = class { ... }` / `var LaterClass = class { ... }`. The `var`-hoisted `LaterClass` binding evaluated to `undefined` at decoration time, the `typeof X === "undefined" ? Object : X` guard short-circuited to `Object`, and the bug went silent.

**rolldown after this PR** lowers to `let Source = class { ... }` / `let LaterClass = class { ... }` — matching babel's semantics. The `typeof` guard now throws `ReferenceError` from TDZ exactly like babel and tsc, instead of silently producing `Object` metadata.

## Fix

```diff
-  /// rewrite toplevel `class ClassName {}` to `var ClassName = class {}`
+  /// rewrite toplevel `class ClassName {}` to `let ClassName = class {}`.
+  /// `let` preserves the lexical-binding / TDZ semantics of the original
+  /// `class` declaration; `var` would hoist the binding as `undefined` and
+  /// silence forward-reference errors.
   fn get_transformed_class_decl(...) -> ... {
     ...
     Some(self.snippet.builder.declaration_variable(
       class.span,
-      VariableDeclarationKind::Var,
+      VariableDeclarationKind::Let,
       self.snippet.builder.vec1(self.snippet.builder.variable_declarator(
         SPAN,
-        VariableDeclarationKind::Var,
+        VariableDeclarationKind::Let,
         ...
```

The inner class expression still carries its own name binding (`class X { ... }`), so self-references like `class Foo { static foo = new Foo() }` continue to work — this is confirmed by `tests/esbuild/default/avoid_tdz` and `keep_names/class_decl_self_ref_deconflict`, both of which still pass with the `var` → `let` swap.

## Test plan

- [x] New regression fixture at `crates/rolldown/tests/rolldown/topics/decorator_metadata_forward_ref/` pins the forward-referenced decorator-metadata case, locking the emit to `let Source = class { ... }` and `let LaterClass = class { ... }`.
- [x] 75 pre-existing snapshots updated mechanically (`var X = class` → `let X = class`); no other content changed in those files.
- [x] Bug reproduction confirmed on this branch's base (`upstream/main`) by temporarily reverting the source change and re-running the new fixture: emit was `var Source = class` / `var LaterClass = class`, matching the broken behaviour described above.
- [x] Local CI parity: `cargo fmt --check`, `cargo clippy --workspace --all-targets -- --deny warnings`, `cargo check --workspace --all-features --all-targets`, `cargo ls-lint`, `cargo run --bin generator` (clean diff), `vp lint`, `vp run --filter rolldown-tests test:types`, `vp run lint-publint`, `vp run lint-knip` all clean.
- [x] `INSTA_UPDATE=no cargo test --workspace --exclude rolldown_binding --exclude bench`: 1725 passed (the 2 local failures are environmental — test262 submodule not cloned, and `external_module_exclusion_package` snapshot expects the `aws-sdk` import to stay unresolved but the local `pnpm install` made it resolvable; both pass on CI).

## AI usage disclosure

### DIFF
--- a/crates/rolldown/src/module_finalizers/mod.rs
+++ b/crates/rolldown/src/module_finalizers/mod.rs
@@ -1284,7 +1284,10 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
     (original_name != canonical_name).then_some((original_name, canonical_name))
   }
 
-  /// rewrite toplevel `class ClassName {}` to `var ClassName = class {}`
+  /// rewrite toplevel `class ClassName {}` to `let ClassName = class {}`.
+  /// `let` preserves the lexical-binding / TDZ semantics of the original
+  /// `class` declaration; `var` would hoist the binding as `undefined` and
+  /// silence forward-reference errors.
   fn get_transformed_class_decl(
     &self,
     class: &mut allocator::Box<'ast, ast::Class<'ast>>,
@@ -1310,10 +1313,10 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
     }
     Some(self.snippet.builder.declaration_variable(
       class.span,
-      VariableDeclarationKind::Var,
+      VariableDeclarationKind::Let,
       self.snippet.builder.vec1(self.snippet.builder.variable_declarator(
         SPAN,
-        VariableDeclarationKind::Var,
+        VariableDeclarationKind::Let,
         ast::BindingPattern::BindingIdentifier(self.snippet.builder.alloc(id)),
         NONE,
         Some(Expression::ClassExpression(ArenaBox::new_in(
--- a/crates/rolldown/tests/esbuild/dce/dce_of_experimental_decorators/artifacts.snap
+++ b/crates/rolldown/tests/esbuild/dce/dce_of_experimental_decorators/artifacts.snap
@@ -16,13 +16,13 @@ const fn = () => {
 //#region keep-these.ts
 let Class = class Class {};
 Class = __decorate([fn], Class);
-var Field = class {};
+let Field = class {};
 __decorate([fn], Field.prototype, "field", void 0);
-var Method = class {
+let Method = class {
 	method() {}
 };
 __decorate([fn], Method.prototype, "method", null);
-var Accessor = class {
+let Accessor = class {
 	#_accessor_accessor_storage;
 	get accessor() {
 		return this.#_accessor_accessor_storage;
@@ -32,13 +32,13 @@ var Accessor = class {
 	}
 };
 __decorate([fn], Accessor.prototype, "accessor", null);
-var Parameter = class {
+let Parameter = class {
 	foo(bar) {}
 };
 __decorate([__decorateParam(0, fn)], Parameter.prototype, "foo", null);
-var StaticField = class {};
+let StaticField = class {};
 __decorate([fn], StaticField, "field", void 0);
-var StaticMethod = class {
+let StaticMethod = class {
 	static method() {}
 };
 __decorate([fn], StaticMethod, "method", null);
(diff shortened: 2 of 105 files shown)


Dunqing:
@shulaoda I see the git blame, and it seems you were the last to touch this code. Can you take a look? I am not familiar with this.

shulaoda:
> @shulaoda I see the git blame, and it seems you were the last to touch this code. Can you take a look? I am not familiar with this.

In fact, I only did a stylistic refactor for this piece of code. Could @hyf0 and @IWANABETHATGUY review it to make sure everything looks correct?

hyfdev:
Thanks for the thorough investigation and the regression fixture — I opened #9680 to discuss this properly.

My current take (details there):

- I initially wanted to align with Rollup (`let`).
- But for modules rolldown wraps to preserve execution order (circular dependencies, etc.), this behavior can't change — they'd keep behaving as today regardless.
- So the change would only be partial (wrapped modules unchanged), and the same forward reference would behave differently in a flat vs a wrapped module.

I'm inclined to keep the current, consistent `var` behavior. Let's converge in #9680 before deciding — thanks again for the deep dive.

hyfdev:
I really appreciate your effort on this one, it leads me to have a deep understanding about js/rolldown behaviors.

However, as said in https://github.com/rolldown/rolldown/issues/9680 the team thinks the current trade off on this behavior is in a practical position, so we're not gonna change the behavior.

Again, thanks for the effort on this subject. if you have more to say, we can continue the discussion at https://github.com/rolldown/rolldown/issues/9680.

PR #9609, opened 2026-05-30
refactor: identify AST nodes by NodeId instead of Span/Address

IWANABETHATGUY:
## Summary

Replaces the two ad-hoc cross-pass AST-node identity keys — oxc `Span` (deduped by `PreProcessor`) and arena `Address` — with oxc's post-semantic `NodeId` for every side table threaded from scan → link → finalize:

- `EcmaView::imports`, `dummy_record_set`, `new_url_references`, `this_expr_replace_map`
- `MemberExprRef::node_id` / `LinkingMetadata::resolved_member_expr_refs`
- `DynamicImportExprInfo::node_id` / `EntryPoint::related_stmt_infos`
- cross-module-optimization state (side-effect-free calls, unreachable dynamic imports)

Cross-module tables are keyed by `(ModuleIdx, NodeId)` because node ids are unique only within a single AST. `Span` is retained purely as source-location metadata (diagnostics, source maps, generated replacement spans); import records now carry an explicit `importer_span` so the TLA import-chain diagnostic no longer needs a reverse lookup over the (formerly span-keyed) `imports` map.

This removes the implicit "spans are stable and unique within a module" contract: finalizer-generated nodes keep `NodeId::DUMMY` and can no longer collide with scan-time records. The `PreProcessor` span-dedup machinery itself is untouched — its one remaining functional job is keeping the synthetic `0..0` span out of scanned ASTs for `span.is_unspanned()` checks; shrinking it down to that job is left as a follow-up.

## Notes

- The incremental-cache (`make_copy`) and HMR paths finalize a *clone* of the scanned AST (`EcmaAst::clone_with_another_arena`). The cache path relies on `clone_in_with_semantic_ids` to preserve node ids, since link/finalize reuse scan-time scoping without re-running semantic (plain `clone_in` would reset them to `NodeId::DUMMY`). The HMR path re-runs `make_semantic` on the clone, which re-stamps every id — lookups still hit because deterministic numbering over the unmutated clone re-derives the scan-time ids. Both mechanisms are documented in `meta/design/ast-mutation.md`.
- `meta/design/ast-mutation.md` is rewritten to document the new `NodeId` contract.

### DIFF
--- a/crates/rolldown/src/ast_scanner/impl_visit.rs
+++ b/crates/rolldown/src/ast_scanner/impl_visit.rs
@@ -1,4 +1,3 @@
-use oxc::allocator::{GetAddress, UnstableAddress};
 use oxc::{
   ast::{
     AstKind,
@@ -13,8 +12,9 @@ use oxc::{
 };
 use rolldown_common::{
   ConstExportMeta, EcmaModuleAstUsage, EcmaViewMeta, ImportKind, ImportRecordMeta, LocalExport,
-  MemberExprObjectReferencedType, OutputFormat, RUNTIME_MODULE_KEY, SideEffectDetail, StmtInfoIdx,
-  StmtInfoMeta, SymbolRefFlags, dynamic_import_usage::DynamicImportExportsUsage,
+  MemberExprObjectReferencedType, MemberExprRef, OutputFormat, RUNTIME_MODULE_KEY,
+  SideEffectDetail, StmtInfoIdx, StmtInfoMeta, SymbolRefFlags,
+  dynamic_import_usage::DynamicImportExportsUsage,
 };
 #[cfg(debug_assertions)]
 use rolldown_ecmascript::ToSourceString;
@@ -192,6 +192,7 @@ impl<'me, 'ast: 'me> Visit<'ast> for AstScanner<'me, 'ast> {
         request.as_str(),
         ImportKind::DynamicImport,
         expr.source.span(),
+        expr.span,
         {
           let mut meta = ImportRecordMeta::empty();
           meta.set(ImportRecordMeta::IsTopLevel, self.is_root_scope());
@@ -199,10 +200,10 @@ impl<'me, 'ast: 'me> Visit<'ast> for AstScanner<'me, 'ast> {
           meta.set(ImportRecordMeta::InTryCatchBlock, self.in_side_try_catch_block());
           meta
         },
-        Some(expr.unstable_address()),
+        Some(expr.node_id()),
       );
       self.init_dynamic_import_binding_usage_info(import_rec_idx);
-      self.result.imports.insert(expr.span, import_rec_idx);
+      self.result.imports.insert(expr.node_id(), import_rec_idx);
     } else if matches!(self.immutable_ctx.options.format, OutputFormat::Cjs)
       && !self.immutable_ctx.options.dynamic_import_in_cjs
     {
@@ -309,7 +310,7 @@ impl<'me, 'ast: 'me> Visit<'ast> for AstScanner<'me, 'ast> {
 
   fn visit_this_expression(&mut self, it: &ast::ThisExpression) {
     if !self.is_this_nested() {
-      self.top_level_this_expr_set.insert(it.span);
+      self.top_level_this_expr_set.insert(it.node_id());
     }
     walk::walk_this_expression(self, it);
   }
@@ -561,7 +562,7 @@ impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
                 self.result.ast_usage.insert(EcmaModuleAstUsage::UnknownExportsRead);
               }
               None => match self.try_extract_parent_static_member_expr_chain(1) {
-                Some((_span, prop)) => {
+                Some((_node_id, _span, prop)) => {
                   self.cjs_named_exports_usage.entry(prop[0].name.clone()).or_default().read += 1;
                 }
                 _ => {
@@ -588,7 +589,7 @@ impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
                 .polyfill_require_for_esm_format_with_node_platform()
             {
               self.current_stmt_info.meta.insert(StmtInfoMeta::HasDummyRecord);
-              self.result.dummy_record_set.insert(ident_ref.span);
+              self.result.dummy_record_set.insert(ident_ref.node_id());
             }
           }
           _ => {}
@@ -615,7 +616,7 @@ impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
               (MemberExprObjectReferencedType::Named, usize::MAX)
             }
           };
-          if let Some((span, props)) =
+          if let Some((node_id, span, props)) =
             self.try_extract_parent_static_member_expr_chain(max_tract_len)
           {
             if !span.is_unspanned() {
@@ -632,14 +633,15 @@ impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
                 let symbol_ref_flags = root_symbol_id.flags_mut(&mut self.result.symbol_ref_db);
                 *symbol_ref_flags |= SymbolRefFlags::HasComputedMemberWrite;
               }
-              self.add_member_expr_reference(
+              self.add_member_expr_reference(MemberExprRef::new(
                 root_symbol_id,
                 props,
+                node_id,
                 span,
                 ty,
                 ident_ref.reference_id.get(),
                 is_member_write,
-              );
+              ));
             }
           } else if is_member_write
             && matches!(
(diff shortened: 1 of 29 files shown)


IWANABETHATGUY:
* **#9609** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9609?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9609?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9609).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown/src/ast_scanner/impl_visit.rs
```diff
@@ -701,7 +703,7 @@ impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
     if let AstKind::CallExpression(call_expr) = parent {
       if ident_ref.name == "eval"
         && !call_expr.optional
-        && call_expr.callee.address() == ident_ref.unstable_address()
+        && call_expr.callee.node_id() == ident_ref.node_id()
```
overlookmotel, line 706: In cases like this, `Address` is cheaper than `NodeId`. `ident_ref.unstable_address()` is essentially `ident_ref as *const _ as usize` (a no-op to LLVM - the value is already in a register), whereas `ident_ref.node_id()` follows the `ident_ref` pointer and reads the `node_id` field of the `IdentifierReference`. If it's not already in L1 cache, that's somewhat costly.

Same for the other side of the comparison `call_expr.callee.node_id()` - an extra memory read there too.

What's our view on this? Could argue either:

1. Priority is perf -> stick to `Address`, or
2. Priority is more understandable and less error-prone code -> use `NodeId`

Alternatively, we could add a new API which uses `Address` under the hood but is more readable e.g. `call_expr.callee.is_same_node_as(ident_ref)`?

hyfdev: I didn't look into the details, but 

> In cases like this, Address is cheaper than NodeId.

Code cleaness and maintainability win over peformance in general.

Span and Address are bug-prone.

overlookmotel: I completely agree about `Span`. Using it as a unique identifier is a footgun - because it's not unique! `Address` can also be error-prone in some cases, and it's not obvious which cases are safe and which are dodgy.

But personally in this case where the question you're asking is "We know X's parent is a `CallExpression`, is X the callee?", I think using `Address` is infallible, and the extra work of consulting the `NodeId`s is fairly pointless. Essentially the question is `ptr::addr_eq(ident_ref, &call_expr.callee)`, and that's what `Address` does, but with more type-level guardrails to prevent mistakes.

Having thought a bit more, personally I would quite favour the new API I suggested `call_expr.callee.is_same_node_as(ident_ref)` - that's both clear and performant. But this isn't my codebase - what do you think?

IWANABETHATGUY: `call_expr.callee.is_same_node_as(ident_ref)` looks good to me

IWANABETHATGUY:
@overlookmotel Thanks for the review. However, I still prefer using NodeId to identify nodes, for the following reasons:

Address is more error-prone. I know it is safe to use in most cases, but it can still potentially introduce bugs when unstable_address is involved.
Address comparison is indeed faster than NodeId comparison, but for Rolldown this should be trivial and probably not even observable, since AST traversal is only a small part of the overall pipeline.
If we keep using Address in some places, we would end up with two ways to identify a node, and three ways to inspect a node if we also count Span, which is used for diagnostics. This would make the code a bit messy.

Overall, I would prefer to use NodeId consistently for all AST node identification.

overlookmotel:
OK, fair enough. I tend towards prioritising perf myself, but obviously simplicity has large advantages - your rationale makes total sense.

My understanding of where we're at with `NodeId` in Oxc is:

1. AST which has come out of `Semantic`: all AST nodes have unique `NodeId`s defined.
2. New AST nodes created in transformer / minifier: have dummy `NodeId` (0).

We need to fix the 2nd, but it wasn't on immediate roadmap. We'll likely need to have 2 x `AstBuilder`s with/without `NodeId` creation. If it's important to Rolldown, we can bump it up the priority list.

This PR: What I haven't figured out yet is whether it's going to hit problems due to the dummy `NodeId`s post-transform. I'd need to explore Rolldown (with an LLM) and get my head around the overall flow. I'm keen to do that - I've felt for a long time that I need to understand Rolldown much better - and this could be a good opportunity.

So... would you like to wait for me to do that and then review this PR (likely next week)? Or would you like to assess yourselves and get this merged earlier?

Sorry, I had hoped to get into it properly this week, but... other stuff has been going on.

IWANABETHATGUY:
> OK, fair enough. I tend towards prioritising perf myself, but obviously simplicity has large advantages - your rationale makes total sense.
> 
> My understanding of where we're at with `NodeId` in Oxc is:
> 
> 1. AST which has come out of `Semantic`: all AST nodes have unique `NodeId`s defined.
> 2. New AST nodes created in transformer / minifier: have dummy `NodeId` (0).
> 
> We need to fix the 2nd, but it wasn't on immediate roadmap. We'll likely need to have 2 x `AstBuilder`s with/without `NodeId` creation. If it's important to Rolldown, we can bump it up the priority list.
> 
> This PR: What I haven't figured out yet is whether it's going to hit problems due to the dummy `NodeId`s post-transform. I'd need to explore Rolldown (with an LLM) and get my head around the overall flow. I'm keen to do that - I've felt for a long time that I need to understand Rolldown much better - and this could be a good opportunity.
> 
> So... would you like to wait for me to do that and then review this PR (likely next week)? Or would you like to assess yourselves and get this merged earlier?
> 
> Sorry, I had hoped to get into it properly this week, but... other stuff has been going on.

No rush. I’ll wait until you finish the review before merging it.

hyfdev:
> So... would you like to wait for me to do that and then review this PR (likely next week)? Or would you like to assess yourselves and get this merged earlier?

@overlookmotel 

There isn't really a hurry for this, but I also get sometimes we want things to get too pefect to be done. If you could finish the review in time that's good, if you can't that's fine too. On the other hands, merging this would be also a way of getting feedbacks for the node id design. __Will plan to merge this after this weeks' release__, so we get a full week time to revert this PR if you find some significant downsides.

I wanna push this forward, because I'm looking at the some internal improvement/refactor. I want the future refactor can build on top of the new basis. It might not be perfect, but we still want to fix bugs on top of the new basis instead of the old one.

hyfdev:
FYI: pushed a merge commit (13a1cc39) that only resolves the merge conflict — no other changes. The conflicted hunk in `module_finalizers/mod.rs` takes main's side, since #9712 replaced its `imports`-map lookup with the precomputed `transitive_esm_init_targets`.

hyfdev:
Pushed c5fef52 (docs/comments only, no behavior change) and edited the PR description, based on findings from an adversarial review pass over the diff:

- **`meta/design/ast-mutation.md`**: the clone-path note attributed id survival in both paths to `clone_in_with_semantic_ids`. That's only true for the incremental-cache path (link/finalize reuse scan-time scoping and never re-run semantic). The HMR renderers re-run `make_semantic` on the clone, which **re-stamps every NodeId** — lookups still hit because `SemanticBuilder` numbers nodes purely by tree shape, so an unmutated clone re-derives exactly the scan-time ids. The doc now describes both mechanisms and the two invariants the HMR path rests on (clone must stay unmutated until `make_semantic`; oxc numbering must stay shape-deterministic — verified for 0.135). Also added: `NodeId::DUMMY == NodeId::ROOT == 0` (the Program node's id — don't ever record a Program-keyed side-table entry), and named the one remaining functional consumer of `PreProcessor`'s span machinery (`span.is_unspanned()` discrimination for synthesized nodes, e.g. the global-`require` rewrite guard) instead of the vague "older code paths".
- **`meta/design/ast-construction.md`**: still said synthetic `SPAN` exists to avoid false-matching "span-keyed side tables" — stale after this PR; false-match protection is `NodeId::DUMMY` now.
- **`hmr_stage.rs`**: a comment at the three `make_semantic` sites stating the unmutated-clone invariant, since it's enforced by convention only and breaking it fails loudly at the `imports[&…]` lookups but *silently* at the `.get()` lookups.
- **PR description**: removed the claim that the `PreProcessor` synthetic-span workaround was removed (it's untouched by this diff; shrinking it is follow-up material), and corrected the HMR mechanism note per the above.

FWIW the same review pass found no correctness issues in the migration itself — every insert/lookup pair is same-node, the `(span, importer_span)` argument pairs are correct at all call sites, and the Address→NodeId equality semantics check out against oxc 0.135's generated delegation impls (including `preserve_parens: true` cases).

hyf0 (approved):

hyfdev:
### Merge activity

* **Jun 13, 1:45 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 13, 1:45 AM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 13, 1:51 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9610, opened 2026-05-30
fix: should build test-dev-server when test-node

situ2001:
As title shows, this error occurs on my machine when I ran `just roll` or `just test` or `just test-node`.

image


shulaoda (approved):
LGTM!

PR #9619, opened 2026-06-01
docs: clarify that `RolldownBuild::close` method should be called in most cases

sapphi-red:


packages/rolldown/src/api/rolldown/rolldown-build.ts
```diff
@@ -72,6 +72,10 @@ export class RolldownBuild {
   /**
    * Close the bundle and free resources.
    *
+   * This method should be called even if the {@linkcode generate} method
+   * or the {@linkcode write} method throwed an error. It should be called
```
sapphi-red:
```suggestion
   * or the {@linkcode write} method threw an error. It should be called
```

IWANABETHATGUY (approved):

PR #9643, opened 2026-06-03
perf: reuse the import-matching tracker stack across named imports

Boshen:
`match_imports_with_exports` allocated a fresh `Vec<ImportTracker>` (`MatchingContext { tracker_stack: Vec::default() }`) for every named import. Each match always starts from an empty stack, and the ambiguous-reexport recursion gets its own cloned stack, so the top-level stack is safe to reuse.

Hoist one `MatchingContext` outside the per-import loop and `clear()` it before each match — reusing the allocation instead of allocating one `Vec` per import. dhat flagged this as the ~1,370-alloc site on three.js.

Behavior-preserving (each match still sees an empty stack). Covered by the existing snapshot suite.

### DIFF
--- a/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
+++ b/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
@@ -822,6 +822,11 @@ impl BindImportsAndExportsContext<'_> {
       return;
     };
     let is_esm = matches!(self.options.format, OutputFormat::Esm);
+    // Reused across all named imports of this module. `match_import_with_export` always starts
+    // from an empty stack (and the ambiguous-reexport recursion gets its own cloned stack), so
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY (approved):

Boshen:
### Merge activity

* **Jun 4, 5:08 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 4, 5:08 AM UTC**: `Boshen` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 4, 5:13 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9646, opened 2026-06-03
perf: reserve capacity for the resolve dedup map

Boshen:
`resolve_dependencies` builds `dedup_map` with `.collect()` and no capacity hint, so it rehashes/reallocates its bucket array as it fills, even though the final size is bounded by `dependencies.len()`.

Reserve `dependencies.len()` up front (the keys borrow `&str`, so no string allocation is involved). Behavior-preserving.


IWANABETHATGUY:
This seems not needed, 
IIRC, the compiler will do the optimization for you if the container has impl `size_hint` https://docs.rs/size_hint/latest/size_hint/struct.SizeHint.html

Issue #9375, opened 2026-05-12
[Bug]: iife variable conflict


lisonge:
### Reproduction link or steps

https://repl.rolldown.rs/#eNptUMFuwjAM/RUvBwoS6gd04jTtOGmH3QiqQuuiTGlcpSljqvrvc1wKaOKS+Pm9F8dvVI0qRmV9jZc89qn2qrjjraoSbDsKEUaoApqIn9ZbAxM0gVrQqktQq1cdtD+bwG3ED6oHh7AT8H5J9h5WK4i/HVID7ZXesZ2O31hFrRJ97XP1Mpe5pxq/2HRnb3NkLo94+NR6I2xFvic2Ozqt9yI7bFKf10FVxDDgtFWBnKvpx+esbuzpYfsnzL8camysxzehlyCyxZZprz3KzkloBif3zbAetQfAS8TgjStgnyBAJh/Ntgkd5LS+G2IB89S8xWhyaQlJQxRWHgPwpsUCsrJMg3ssy/khgIZCa1iXWdvg0jw5OhrX39wwh8kqiXGRTXKnc9pwehOnduYYHMfdRzX9AQAqwF0=

### What is expected?

```js
(function(pinia) {
	//#region index.ts
	var pinia$1 = (0, pinia.createPinia)();
	console.log([pinia$1]);
	//#endregion
})(Pinia);
```

### What is actually happening?

```js
var __expose__ = (function(pinia) {
	//#region \0rolldown/runtime.js
	var __commonJSMin = (cb, mod) => () => (mod || (cb((mod = { exports: {} }).exports, mod), cb = null), mod.exports);
	//#endregion
	return (/* @__PURE__ */ __commonJSMin(((exports, module) => {
		freeExports && typeof module == "object" && module && module.nodeType;
		var pinia = (0, pinia.createPinia)();
		console.log([pinia]);
	})))();
})(Pinia);
```

### System Info

```Shell
chrome148
```

### Any additional comments?

<img width="2477" height="341" alt="Image" src="https://github.com/user-attachments/assets/7059a924-3496-40cf-a631-846354c77c38" />

lisonge:
[...]

Issue #9055, opened 2026-04-10
[Bug]: CJS interop factory shadowed when re-bundling rolldown-emitted CJS output


sapphi-red:
### Reproduction link or steps

[REPL](https://repl.rolldown.rs/#eNptkd1uhCAQhV+F0BtNLN676SM0bbK9JGlcHK0GwQ7DZo3x3QuydvuzN8Y5zJn5Diy85dXCe9PARQwu/hte3eqCq1iOk0ViS4cAtLIW7cgkF6Ua3GMDE4Ruo2YRSskP0kijrHFWg9C2yzZTLk2YBbwi9LCGqf+c++o7Jwni5TSAItFA2xt4RTsB0pzBJaK5gh3n8WS1IHsk7E33VncFW9i51h6qAPtsG69BcrbmB4mJkBjCp+8R3jdG9rTXWQy3aSlTslxXiT/NyZzUrY+HfN/mPdZNSGnK8gGh661hDtV1F7lob71RFA82McvZElVJCOTRhCgfoLVNj/D7suLlh9Y1fsL8oKYV9+B/4gbec0DSNYEjvn4BCVG7kg==)

### What is expected?

The output to not have `const require_greet = require_greet();`.

### What is actually happening?

The output has `const require_greet = require_greet();` which causes TDZ error.

### System Info

```Shell
REPL 1.0.0-rc.15
```

### Any additional comments?

Originally reported at https://github.com/vitejs/vite/issues/22176

hyfdev:
https://github.com/rolldown/rolldown/pull/9392

PR #9648, opened 2026-06-03
fix: cross-chunk CJS wrapper shadowed by author-local binding

### LINKED ISSUE #9630: Automatic code splitting: circular CJS chunk → uninitialized binding ('n is not a function') (es-toolkit via Recharts, Vite 8)
## Summary

In a **production build only** (dev + tests are fine), a route chunk that renders a CommonJS dependency throws `TypeError: n is not a function` at runtime. The error looks like a minified variable collision, but the real cause is a **circular chunk import where a CJS module's binding is read before the chunk that defines it has executed**.

This is the same failure class as #8803 — but that issue states automatic splitting *"produces chunks that don't form cycles."* Here it **does** form a cycle under plain Vite automatic/route-based splitting, with **no `codeSplitting.groups`** configured. So this case appears unreported.

## Environment

- `vite@8.0.14` (bundler: `rolldown@1.0.3`)
- `recharts@3.8.1` → `es-toolkit@1.47.0`
- No `output.codeSplitting.groups` — default Vite automatic splitting.

## Evidence

A sourcemapped build maps the throw to `es-toolkit/dist/compat/math/maxBy.js:2`, a top-level `const require_identity = require("../../function/identity.js")`.

The emitted (minified) code:

```js
un = W((e => {
  var t = ln(), n = n(), r = r();          // <-- n=n() and r=r() are self-referential
  function i(e, i){ /* ... */ r.iteratee(i ?? n.identity) }
  e.maxBy = i
}))
```

`t = ln()` is fine (the module-init function and its result binding got distinct names), but `n = n()` / `r = r()` read a binding that is still uninitialized at that point — the cross-chunk factory and the binding it initializes collapsed to the same name. With `minify: false` it surfaces as a deconfliction failure instead, per #8803.

**Chain:** Recharts does `import maxBy from 'es-toolkit/compat/maxBy'`. es-toolkit's `./compat/*` export is CJS-only (`module.exports = require('../dist/compat/math/maxBy.js').maxBy`), so rolldown bundles + wraps the nested-`require` chain and splits the helper modules (`identity`, `iteratee`) into a different chunk than the consumer — forming the cycle.

## Workarounds (both confirm the cause)

- `output.strictExecutionOrder: true` — fixes it (matches #8803's mitigation), at ~2x bundle-size cost.
- A `manualChunks` group co-locating the consumer + its CJS dep in one chunk — fixes it (removes the split that forms the cycle). This is what we shipped.

## Question

Should automatic (non-grouped) splitting avoid forming these CJS cycles so neither workaround is needed — or, if `strictExecutionOrder` is the intended answer, should the docs note it's required for **automatic** splitting with CJS deps too (not only `codeSplitting.groups`, as #8803 implies)?

I can put together a minimal Vite 8 + Recharts reproduction if that would help.

IWANABETHATGUY:
Fixes #9630.

## Problem

When code-splitting hoists a CommonJS `require_*` wrapper into a different chunk than the module that `require()`s it, an author's local binding of the same name shadowed the imported wrapper. This happens in real code — es-toolkit's compiled output literally writes `var require_isArrayLike = require("./isArrayLike.cjs")`, whose local matches the wrapper name rolldown derives from the filename.

The lazy chunk then emitted the self-referential:

```js
import { n as __commonJSMin, t as require_isArrayLike } from "./isArrayLike.js";
var require_last = __commonJSMin(((exports) => {
  var require_isArrayLike = require_isArrayLike(); // local shadows the import
  ...
```

The hoisted `var` resolves to itself (`undefined`), throwing `TypeError: require_isArrayLike is not a function` at runtime. (A subsequent DCE pass also drops the now-"unused" import, so the import disappears entirely — but the crash is the same either way.)

## Root cause

`deconflict_chunk_symbols` already guards against an author-local inside a CJS `__commonJS((exports, module) => { ... })` closure shadowing a name captured at the chunk's root scope (issues #9055 / #9375). It does this via `chunk_scope_captured_names` — but that set was only seeded from wrappers of modules **in the current chunk**.

A wrapper hoisted into another chunk is imported here as a real root-scope binding (`import { ... as require_foo } from "./other.js"`) and is *equally* captured by every CJS closure in this chunk, yet it was invisible to that loop. So the colliding author-local was not deconflicted.

## Fix

Seed `chunk_scope_captured_names` from cross-chunk CJS wrapper imports too (`chunk.imports_from_other_chunks`), the symmetric counterpart to the existing in-chunk wrapper loop. The colliding local is now renamed, matching the same-chunk output:

```js
import { n as __commonJSMin, t as require_isArrayLike$1 } from "./isArrayLike.js";
var require_last = __commonJSMin(((exports) => {
  var require_isArrayLike = require_isArrayLike$1();
  ...
```

## Test

Adds `tests/rolldown/issues/9630`, a minimal reproduction (eager import of the leaf + dynamic import of the requirer to force the cross-chunk split). The fixture executes the bundled output, so it fails on the runtime `TypeError` without the fix and passes with it.

Verified regression-free: full rolldown fixture, esbuild, and rollup snapshot suites pass with zero existing-snapshot changes.

### DIFF
--- a/crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
+++ b/crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
@@ -121,6 +121,25 @@ pub fn deconflict_chunk_symbols(
         .insert(CompactStr::new(canonical_ref.name(&link_output.symbol_db)));
     }
   }
+  if matches!(format, OutputFormat::Esm) {
+    // A CJS wrapper whose module lives in *another* chunk is hoisted here as a real root-scope
+    // import binding (`import { ... as require_foo } from "./other.js"`) and is likewise captured
+    // by every CJS-wrapped closure in this chunk. The in-chunk loop above can't see it because its
+    // owner module isn't in `chunk.modules`, so we recover it from the cross-chunk import list.
+    // Without this, an author-local of the same name inside a CJS closure shadows the imported
+    // wrapper, emitting the self-referential `var require_foo = require_foo()` (issue #9630).
+    for item in chunk.imports_from_other_chunks.values().flatten() {
+      let canonical_ref = link_output.symbol_db.canonical_ref_for(item.import_ref);
+      let is_cjs_wrapper =
+        link_output.metas[canonical_ref.owner].wrapper_ref.is_some_and(|wrapper_ref| {
+          link_output.symbol_db.canonical_ref_for(wrapper_ref) == canonical_ref
+        });
+      if is_cjs_wrapper {
+        chunk_scope_captured_names
+          .insert(CompactStr::new(canonical_ref.name(&link_output.symbol_db)));
+      }
+    }
+  }
 
   chunk
     .modules
--- a/crates/rolldown/tests/rolldown/issues/9630/_config.json
+++ b/crates/rolldown/tests/rolldown/issues/9630/_config.json
@@ -1,0 +1,10 @@
+{
+  "config": {},
+  "configVariants": [
+    {
+      "_configName": "cjs",
+      "format": "cjs",
+      "treeshake": false
+    }
+  ]
+}
--- a/crates/rolldown/tests/rolldown/issues/9630/_test.mjs
+++ b/crates/rolldown/tests/rolldown/issues/9630/_test.mjs
@@ -1,0 +1,15 @@
+import assert from 'node:assert';
+import { createRequire } from 'node:module';
+
+// The bug (#9630): a CJS wrapper hoisted across a chunk boundary was imported
+// under its un-deconflicted base name (`require_isArrayLike`), which an author's
+// local binding of the same name then shadowed, emitting the self-shadowing
+// `var require_isArrayLike = require_isArrayLike()`. The CJS variant covers the
+// same cross-chunk wrapper reference without treating CJS's generated
+// `const require_<chunk> = require(...)` binding as an ESM import binding.
+const require = createRequire(import.meta.url);
+const mod =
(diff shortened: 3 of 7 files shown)


IWANABETHATGUY:
* **#9648** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9648?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9648?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9648).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyf0 (approved):

shulaoda (approved):

IWANABETHATGUY:
### Merge activity

* **Jun 4, 2:07 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 4, 2:07 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 4, 2:17 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9652, opened 2026-06-05
fix(dev): print build errors on browser refresh after a failed build

h-a-n-a:
## Design Principles

This PR mainly clarifies a few principles that we hold while implementing Rolldown's dev engine and Vite's full-bundle mode dev server. Mostly are written in `meta/design/dev-engine.md`. Just to give you a briefing, here're these things:

1. **Be conservative** — rebuild only when the bundle is stale, and defer it until the full bundle is actually requested (e.g. a page refresh). By default we ship HMR-only output.
2. **Replay errors when needed** — the dev *engine* is stateless across HTTP requests and never caches results from `onHmrUpdates`/`onOutput`; caching and replay are the dev *server*'s job, so errors survive a refresh.
3. **File changes are the only recovery trigger** — after a failed build the engine waits for a file edit (Vite config or source). Not a page refresh, elapsed time, or UI dismissal.
4. **Build errors are always recoverable** — every error via `on_output`/`on_hmr_updates` is treated as a user error (source/plugin), fixable by editing source. Only a panic is unrecoverable.

### Problem

  #9552 made `ensure_latest_bundle_output` return `None` in `FullBuildFailed` /
  `Failed` instead of retrying a build. That fixed a double-`onOutput` and an
  infinite loop — but the retry it removed had been doing double duty: on a
  browser refresh, the dev-server middleware calls `ensureLatestBuildOutput`,
  the retry reran the build, and `onOutput` re-fired the error to the client.

  With the retry gone, a refresh no longer reruns the build, so **a project whose
  build output contains an error stops re-emitting that error on refresh** — the
  error overlay disappears even though the build is still broken.

  The right fix is *not* to rerun the build on access (that brings back the loop
  and the double-emit). It's to make `Failed` / `FullBuildFailed` work as the
  genuine resting states they now are:

  - the error must survive a refresh **without** a rebuild, and
  - recovery must be driven entirely by file changes, re-running the stage that
    actually broke.

### Changes

  **1. Expose errored-and-stale state to the consumer.**
  Snapshot field `last_full_build_failed` → `last_build_errored`, now true for
  *both* `FullBuildFailed` and `Failed { .. }`. This is the predicate the
  consumer pairs with `has_stale_output` to (a) replay its cached build error on
  client reconnect instead of triggering a doomed regen, and (b) avoid the
  access-triggered reload loop against a bundle the engine won't regenerate.

  **2. File-change-driven, stage-aware recovery.**
  Because access no longer retries, the next file change is the only recovery
  trigger — and it must re-run the broken stage:

  - New `ErrorStage` enum (`Hmr` | `Rebuild`); `BundlingTask` tracks
    `hmr_errored` / `rebuild_errored` and collapses them with precedence
    `Rebuild > Hmr` (covers the `Hmr → HmrRebuild` auto-upgrade that then fails
    in rebuild).
  - `CoordinatorState::Failed { last_error_stage }` carries the stage;
    `handle_file_changes` forces `HmrRebuild` on a `Rebuild`-stage failure
    (regardless of `rebuild_strategy`) so the rebuild re-runs, and uses an `Hmr`
    task for an `Hmr`-stage failure.

### Tests

Tests will be added when Rolldown's test-dev-server is fully aligned with Vite's full-bundle mode server.

This PR is meant to be released along with https://github.com/vitejs/vite/pull/22617.

### DIFF
--- a/meta/design/dev-engine.md
+++ b/meta/design/dev-engine.md
@@ -17,6 +17,75 @@ implementation, not the narrative of any particular change.
 
 ---
 
+## Design principles
+
+Four principles govern when the dev engine rebuilds and how its errors
+flow out to the binding consumer. They define rolldown_dev's contract
+with its consumer (typically Vite) and constrain the implementation in
+§7, §13, and §16.
+
+### 1. Conservative rebuilds
+
+Rebuilds happen only when the bundle is **stale** — when input has
+changed since the last build attempt. Page access and browser
+reconnect on their own never trigger a rebuild. In particular: if the
+previous build failed, an access request does not retry — without new
+input the same error would recur.
+
+Realized in: `BundleCoordinator::ensure_latest_bundle_output` returning
+`None` for `Failed` / `FullBuildFailed` (§13b, §13e).
+
+### 2. Errors are emitted on every build
+
+rolldown_dev surfaces build errors to the binding consumer on every
+build via the `on_output` / `on_hmr_updates` callbacks (§16b). It never
+silently retries past an error, never silently swallows one, and never
+caches one across requests — rolldown_dev is stateless across HTTP
+requests. The binding consumer (Vite) is responsible for retaining the
+most recent error and replaying it on each client reconnect, so the
+error overlay appears even after a browser refresh.
+
+Vite-side realization (in `fullBundleEnvironment.ts`): a single
+`lastBuildError: Error | null` field caches the most recent error from
+**either** channel — it is set in both `onOutput` (full-build errors)
+and `onHmrUpdates` (HMR errors), and cleared back to `null` only on a
+successful `onOutput`. It is replayed on the **`vite:client:connect`**
+event for every freshly connected client (including a post-refresh
+reconnect), so the error overlay reappears after a browser refresh.
+The two channels differ only in their _live_
+delivery: an `onOutput` error is additionally logged to the terminal
+(`logger.error`) so a build break is visible without a browser, and is
+broadcast to all clients via `hot.send`; an `onHmrUpdates` error is sent
+to each connected client individually and is not logged to the terminal.
+
+### 3. File changes are the only recovery trigger
+
+After a failed build, the engine waits for a file change before
+rebuilding. Both Vite config edits and user-land source edits are
+valid triggers. Nothing else — not page refresh, not elapsed time, not
+manual UI dismissal — counts as recovery.
+
+Realized in: `handle_file_changes` (§7) is the sole producer of
+post-failure rebuild tasks. `triggerFullBuild` (§13e) is an explicit
+escape hatch for cases the watcher cannot observe (e.g. missing-import
+resolution; see Unresolved Questions).
+
+Corollary: a file change after a failed build must schedule work that
+can undo the failure. In practice this means tracking where the
+failure originated (HMR computation vs incremental rebuild) so the
+next task covers the stage that broke (§7).
+
+### 4. Build errors are recoverable; panics are bugs
+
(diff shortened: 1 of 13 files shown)


h-a-n-a:
* **#9720** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9720?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9684** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9684?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9668** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9668?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9686** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9686?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9652** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9652?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9652?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9652).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyf0 (approved):

meta/design/dev-engine.md
```diff
@@ -17,6 +17,75 @@ implementation, not the narrative of any particular change.
...
+(`logger.error`) so a build break is visible without a browser, and is
+broadcast to all clients via `hot.send`; an `onHmrUpdates` error is sent
+to each connected client individually and is not logged to the terminal.
+
+### 3. File changes are the only recovery trigger
+
+After a failed build, the engine waits for a file change before
+rebuilding. Both Vite config edits and user-land source edits are
+valid triggers. Nothing else — not page refresh, not elapsed time, not
+manual UI dismissal — counts as recovery.
```
sapphi-red, line 66: I think I forgot to mention this. If the HMR errors and then the user triggers a rebuild (by a page reload), Rolldown should run the rebuild instead of returning the cached HMR error. This is to make it easier to recover if the HMR generation is buggy for some reason. Does the current implementation handle that? (It's not an urgent thing so it can be implemented later.)

sapphi-red: I think it would be helpful to add the transitions of calling rebuild in https://github.com/rolldown/rolldown/blob/247bdcdcf7531830440f68370fac88f289cf5662/meta/design/dev-engine.md#transition-map
also maybe convert it in mermaid as the ascii graph is a bit hard to read

sapphi-red (approved):

meta/design/dev-engine.md
h-a-n-a: > If the HMR errors and then the user triggers a rebuild (by a page reload), Rolldown should run the rebuild instead of returning the cached HMR error.

In this PR, this implementation does not handle this. HMR errors are treated as one of the errors listed in the principle. These errors are cached in the bundler and will be restored after a page reload.

I can handle this by adding a flag in `BundleState` that indicates an HMR error and start a rebuild when browser requests come in. I will handle this in this PR stack since it's not merged.

h-a-n-a: We can achieve this by leveraging the newly added `triggerFullBuild` API in https://github.com/rolldown/rolldown/pull/9552

PR #9657, opened 2026-06-06
chore(deps): update docs dependencies

### LINKED ISSUE #8792: doc: backslashes in code blocks rendered as URL-encoded sequences on docs site

bddjr:
Fix #8792

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -19,11 +19,11 @@
     "typedoc-plugin-markdown": "^4.12.0",
     "typedoc-plugin-merge-modules": "^7.0.0",
     "typedoc-vitepress-theme": "^1.1.3",
-    "vitepress": "catalog:",
-    "vitepress-plugin-graphviz": "catalog:",
-    "vitepress-plugin-group-icons": "catalog:",
-    "vitepress-plugin-llms": "catalog:",
-    "vitepress-plugin-og": "catalog:",
+    "vitepress": "^2.0.0-alpha.17",
+    "vitepress-plugin-graphviz": "^0.1.0",
+    "vitepress-plugin-group-icons": "^1.7.5",
+    "vitepress-plugin-llms": "^1.13.1",
+    "vitepress-plugin-og": "^0.0.5",
     "void": "^0.9.2"
   }
 }
--- a/docs/package.json
+++ b/docs/package.json
@@ -13,17 +13,17 @@
     "vue": "catalog:"
   },
   "devDependencies": {
-    "@voidzero-dev/vitepress-theme": "^4.8.2",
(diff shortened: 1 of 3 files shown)


shulaoda (approved):
Thanks!

PR #9666, opened 2026-06-08
chore: remove outdated pnpm configurations

btea:
https://pnpm.io/settings#minimumreleaseage

After a certain period of time has passed, the default `minimumreleaseage` configuration no longer restricts the installation of specific versions of oxc-related dependency packages, so the relevant exclusion configurations can be removed.


shulaoda (approved):
Thanks! Since Oxc releases updates fairly frequently, entries like `minimumReleaseAgeExclude` may continue to be necessary from time to time. I think it makes sense to keep this configuration. 🤔

PR #9669, opened 2026-06-08
fix(finalizer): skip init_*() for tree-shaken wrapped ESM owners

IWANABETHATGUY:
Fixes #9651.

## Problem

Under `cjs` + `inlineDynamicImports` + tree-shaking, the module finalizer panicked with:

```
"init_external" is not in any chunk, which is unexpected
```

This happened when a side-effect-free (`sideEffects: false`) barrel forwards a binding from a wrapped ESM module that is reached **both** statically (`import * as z from './zod'`) **and** through a dynamic `import()` chain. The dynamic import forces the barrel modules to be wrapped as lazy-init ESM, but tree-shaking then drops `external.js` because its namespace is never actually read (only `z.locales` is used, and that resolves straight through to `locales.js`).

When lowering the static `import * as z` from the non-wrapped barrel, `collect_wrapped_esm_init_modules_for_import_record` walked **every** resolved export of the barrel and emitted an `init_*()` call for each wrapped owner — including `external.js`, which tree-shaking had eliminated. Its `init_*` wrapper statement was never included, so the wrapper symbol had no chunk assignment and the finalizer panicked.

## Fix

`add_wrapped_esm_init_module_for_symbol` now only emits `init_*()` for wrapped owners that tree-shaking actually kept, by guarding on `meta.is_included`. This mirrors the existing `is_included` guard in the sibling `generate_transitive_esm_init`. A non-included module produces no `init_*` function in the output, so skipping its init call is required for correctness — there is no case where calling it would be valid.

## Test

Adds a runnable regression fixture under `crates/rolldown/tests/rolldown/issues/9651` that reproduces the panic without the fix and bundles **and executes** successfully with it.

Issue #9651, opened 2026-06-05
[Panic]:  "init_from_json_schema" is not in any chunk, which is unexpected


ibodev1:
### Panic message

```Shell
$ pnpm build
$ node --env-file-if-exists=.env node_modules/rolldown/bin/cli.mjs -c rolldown.config.ts
Rolldown panicked. This is a bug in Rolldown, not your code.

thread '<unnamed>' (25092) panicked at crates\rolldown\src\module_finalizers\mod.rs:512:13:
SymbolRef { owner: 52, symbol: SymbolId(91) } "init_from_json_schema" is not in any chunk, which is unexpected
Rolldown panicked. This is a bug in Rolldown, not your code.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

Please report this issue at: https://github.com/rolldown/rolldown/issues/new?template=panic_report.yml

thread '<unnamed>' (4308) panicked at crates\rolldown\src\module_finalizers\mod.rs:512:13:
SymbolRef { owner: 52, symbol: SymbolId(91) } "init_from_json_schema" is not in any chunk, which is unexpected

Please report this issue at: https://github.com/rolldown/rolldown/issues/new?template=panic_report.yml
Rolldown panicked. This is a bug in Rolldown, not your code.

thread '<unnamed>' (2312) panicked at crates\rolldown\src\module_finalizers\mod.rs:512:13:
SymbolRef { owner: 52, symbol: SymbolId(91) } "init_from_json_schema" is not in any chunk, which is unexpected
Rolldown panicked. This is a bug in Rolldown, not your code.

thread '<unnamed>' (4308) panicked at crates\rolldown\src\module_finalizers\mod.rs:512:13:
SymbolRef { owner: 52, symbol: SymbolId(91) } "init_from_json_schema" is not in any chunk, which is unexpected

Please report this issue at: https://github.com/rolldown/rolldown/issues/new?template=panic_report.yml

Please report this issue at: https://github.com/rolldown/rolldown/issues/new?template=panic_report.yml
Rolldown panicked. This is a bug in Rolldown, not your code.

thread '<unnamed>' (2312) panicked at crates\rolldown\src\module_finalizers\mod.rs:512:13:
SymbolRef { owner: 52, symbol: SymbolId(91) } "init_from_json_schema" is not in any chunk, which is unexpected
Rolldown panicked. This is a bug in Rolldown, not your code.

[...]

PR #9674, opened 2026-06-08
test(9651): drop shimMissingExports from regression fixture

IWANABETHATGUY:
Follow-up to #9669 addressing the Copilot review feedback on the regression fixture: https://github.com/rolldown/rolldown/pull/9669#pullrequestreview-4451531768

## What

The `#9651` regression fixture had `src/dyn/b.js` import a named `z` from `../zod/external.js`, but `external.js` never exports `z`. That import was only valid because the fixture enabled `shimMissingExports`, which is unrelated to the bug under test. The binding is never used — only the `b.js → external.js` import **edge** matters (it pulls the wrapped `external.js` into the dynamic-import chain).

This PR:
- switches `b.js` to a bare side-effect import (`import '../zod/external.js';`), so the fixture is semantically valid without missing-export shimming, and
- drops the now-unnecessary `"shimMissingExports": true` from `_config.json`.

## Why it's safe

- The bundled output is **unchanged** — `artifacts.snap` does not move.
- The fixture still reproduces the original panic on the pre-fix commit: I checked out `5cfff49` (the commit right before #9669 landed), dropped in this updated fixture, and it still panics with `"init_external" is not in any chunk`.

Test-only change; keeps the regression focused on #9651.

PR #9678, opened 2026-06-09
feat: drop no-op init calls for empty wrapped-ESM modules

IWANABETHATGUY:
## Summary

Follow-up to the review nit on #9669: when a wrapped ESM module tree-shakes down to an
empty body, rolldown still emitted a no-op call to its `__esm`/`__esmMin` wrapper, e.g.

```js
function en() { return "en"; }
__esmMin((() => {}))();   // ← dead weight: empty closure, calling it does nothing
```

This PR marks such no-op init calls `@__PURE__` so the default `minify: 'dce-only'`
pass drops them (and, once the wrapper is unreferenced, the wrapper declaration and the
`__esm`/`__esmMin` runtime helper too).

**Before**
```js
function en() { return "en"; }
__esmMin((() => {}))();
```
**After**
```js
function en() { return "en"; }
```

It also collapses the `require(esm)` sequence form `(init_x(), __toCommonJS(x_exports))`
→ `__toCommonJS(x_exports)` via the same mechanism.

## Approach

1. A small generate-stage pre-pass, `compute_init_is_noop`, flags wrapped (`WrapKind::Esm`)
   modules whose `__esm` closure body is provably empty — every top-level statement is a
   hoisted function declaration or a source-less export clause, so nothing lands inside the
   closure. The classification runs in parallel (`par_iter_enumerated`).
2. The finalizer sets `pure = true` on the init `CallExpression` at the three emission sites
   when the callee module is flagged; `dce-only` does the rest.

The classifier is deliberately conservative (a non-qualifying statement only ever *keeps* a
redundant init call — it never drops a needed one), and a `debug_assert!` in the finalizer
cross-checks every flagged module against the *actual* `stmts_inside_closure`, so any
misclassification fails loudly across the fixture suite. Guards exclude shimmed missing
exports (their `x = void 0` lands in the closure) and concatenated wrapper groups.

## Out of scope (kept on purpose)

The empty inlined dynamic import `Promise.resolve().then(() => Object.freeze({ __proto__: null }))`
is **not** removed — Rollup keeps it for `import('./empty.js')` of an empty module, so it's
Rollup-equivalent and stays.

## Why the detection lives in its own pass (and not elsewhere)

The flag is consumed **cross-module** — an importer marks `init_importee()` pure based on the
*importee's* status — which constrains where it can be computed.

### Not in the link stage (`cross_module_optimization.rs`)

That pass runs **before** the two inputs the decision needs even exist:

- **`is_included`** is produced by tree-shaking (`include_statements`), which runs *after*
  `cross_module_optimization`. A module that's tree-shaken away has no init to reason about.
- **`concatenated_wrapped_module_kind`** is assigned during the generate stage's
  chunk/concatenation analysis. We must restrict the optimization to standalone wrappers
  (`ConcatenateWrappedModuleKind::None`): in a concatenated group the shared `init_*` runs the
  *whole group's* closure, so one module's empty body wouldn't prove the call is a no-op. That
  isn't knowable before chunks are formed.

Folding it here would force the decision with incomplete information (pre-tree-shaking,
pre-chunking).

### Not in the finalizer (`module_finalizers/`)

`finalize_modules` runs the per-module finalizers in parallel (`par_iter_mut_enumerated`):

1. **Immutable borrow** — the finalizer holds `&LinkingMetadata` / `&LinkingMetadataVec`
   (`ScopeHoistingFinalizerContext`), read-only, so it *cannot write* the flag mid-pass.
2. **Determinism** — even if it could, an importer finalizing concurrently with its importee
   could read a not-yet-computed (default `false`) value → non-deterministic `@__PURE__`
   placement → flaky snapshots.

So the flag must be settled as a barrier *before* finalization begins.

### The fit

The pre-pass sits in the generate stage **after** tree-shaking and chunk generation (so
`is_included` and `concatenated_wrapped_module_kind` are known) but **immediately before** the
parallel `finalize_modules` (so every importer reads a settled, deterministic value).

## Testing

- Updated 5 fixture snapshots (incl. the #9651 regression); each remains runnable.
- Full `rolldown` + `esbuild` integration suites pass; the `init_is_noop` debug-assert holds
  across all ~1.6k fixtures.
- `just lint-rust` clean (clippy `--deny warnings`, fmt, `cargo check --all-features`).

### DIFF
--- a/crates/rolldown/src/module_finalizers/impl_visit_mut.rs
+++ b/crates/rolldown/src/module_finalizers/impl_visit_mut.rs
@@ -210,6 +210,17 @@ impl<'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'_, 'ast> {
             }
           });
 
+        // Safety net for the `init_is_noop` predictive classifier (see
+        // `generate_stage::compute_init_is_noop`): if a module was flagged as having an empty
+        // `__esm` closure, the statements that actually land inside the closure must be empty.
+        // Otherwise we'd have marked a side-effecting `init_*()` as `@__PURE__` and DCE could
+        // wrongly drop it. Turns any misclassification into a loud failure across the fixtures.
+        debug_assert!(
+          !self.ctx.linking_info.init_is_noop || stmts_inside_closure.is_empty(),
+          "init_is_noop set but the __esm closure is non-empty for {}",
+          self.ctx.module.stable_id
+        );
+
         if is_concatenated_wrapped_module {
           self.rendered_concatenated_wrapped_module_parts.hoisted_functions_or_module_ns_decl =
             declaration_of_module_namespace_object
--- a/crates/rolldown/src/module_finalizers/mod.rs
+++ b/crates/rolldown/src/module_finalizers/mod.rs
@@ -157,12 +157,14 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
           false,
           false,
         );
-        let init_call = self.snippet.builder.expression_call(
+        let init_call = self.snippet.builder.expression_call_with_pure(
           SPAN,
           wrapper_ref_expr,
           NONE,
           self.snippet.builder.vec(),
           false,
+          // A no-op `init_*()` (empty wrapped-ESM closure) is pure; let `dce-only` drop it.
+          importee_linking_info.init_is_noop,
         );
         body.push(self.snippet.builder.statement_expression(SPAN, init_call));
       } else {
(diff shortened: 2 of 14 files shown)


IWANABETHATGUY:
* **#9678** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9678?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9678?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9674** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9674?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9678).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

h-a-n-a (approved):

hyf0 (approved):

IWANABETHATGUY:
### Merge activity

* **Jun 9, 5:34 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 9, 5:39 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 9, 5:44 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9709, opened 2026-06-10
fix(treeshake): drop dead star re-export sources of wrapped barrels

### LINKED ISSUE #9691: Missing import when lazy barrel is enabled
### Discussed in https://github.com/rolldown/rolldown/discussions/9687

<sup>Originally posted by **Desuuuu** June  9, 2026</sup>
When using lazy barrel with version 4.7.7 of `@smithy/node-http-handler` (used by the AWS SDK packages), rolldown seems to be dropping an import from `node:http2`.

Here's a minimal reproduction: https://github.com/Desuuuu/rolldown-lazybarrel-issue

This fails with the following error:

```
> node ./dist/index.js

file://dist/index.js:18823
        ({constants} = node_http2);
        ^

ReferenceError: node_http2 is not defined
```

I'm not sure if the issue comes from rolldown or from `@smithy/node-http-handler` since the issue only started appearing since their latest release (I would guess [this change](https://github.com/smithy-lang/smithy-typescript/pull/2070/changes/bbb5b223b369d651512d65d4f1ffe3b921957e2e) specifically).</div>

IWANABETHATGUY:
### Description

Fixes #9691.

With `codeSplitting: false`, a dynamic `import()` wraps its target, and the wrap propagates through `export *` barrels into every star source. `reference_needed_symbols.rs` then force-marked the barrel's `export *` statements as side-effectful, so a `sideEffects: false` star source whose exports are all unused was force-included anyway:

- the retained dead module bloats the bundle, and
- combined with `lazyBarrel` (which drops the imports the dead body needs, since its exports are never requested), the retained body reads an undefined binding at top level — the `ReferenceError: dep_b is not defined` crash reported with `@aws-sdk/client-s3`.

### Fix

Treat `export *` from an ESM-wrapped importee like a named re-export: let tree-shaking drop the statement, and rely on the finalizer's existing fallback for excluded re-export statements (`generate_transitive_esm_init`), which re-creates the `init_*()` call at the statement's position only when the star source is actually included. Initialization order is preserved, while a fully-unused, side-effect-free star source disappears from the bundle.

The statement is still forced eager when the lazy fallback can't substitute for it:

- the importee has dynamic exports — the eager `__reExport(...)` namespace copy must run;
- the importer is not ESM-wrapped (e.g. an entry module) — the fallback only fires for wrapped importers;
- the importee is TLA-tainted — the fallback emits the call without `await`.

### Tests

- New regression fixture `tests/rolldown/issues/9691` mirrors the `@aws-sdk/client-s3` shape. Without the fix, its node-execution step fails with the original `ReferenceError: dep_b is not defined`; with the fix, the dead handler module is fully dropped in both the `codeSplitting: false` build and the `code-splitting-on` variant, and both execute cleanly.
- Two existing snapshots improve: `export_forms_common_js` drops a no-op `init_a` wrapper, and `strict_execution_order/issue_8777` no longer emits duplicate `init_*()` calls (the force-included `export *` path used to bypass the `generated_init_esm_importee_ids` dedup).

### DIFF
--- a/crates/rolldown/src/stages/link_stage/reference_needed_symbols.rs
+++ b/crates/rolldown/src/stages/link_stage/reference_needed_symbols.rs
@@ -174,8 +174,28 @@ impl LinkStage<'_> {
                       }
                       WrapKind::Esm => {
                         // Turn `import ... from 'bar_esm'` into `init_bar_esm()`
-                        stmt_info.side_effect =
-                          (is_reexport_all || importee.side_effects.has_side_effects()).into();
+                        //
+                        // `export * from 'bar_esm'` is treated like a named re-export and only
+                        // forced to be side-effectful when the eager `init_bar_esm()` call is
+                        // really needed (#9691). If the statement gets tree-shaken while the
+                        // importee is still included, the finalizer re-creates the init call for
+                        // excluded re-export statements of ESM-wrapped importers
+                        // (`generate_transitive_esm_init`), so initialization order is preserved
+                        // while a side-effect-free star source whose exports are all unused can
+                        // be dropped entirely. The eager call is still required when:
+                        // - the importee has dynamic exports: the `__reExport(...)` namespace
+                        //   copy must run eagerly;
+                        // - the importer is not ESM-wrapped (e.g. an entry module): the finalizer
+                        //   fallback only applies to wrapped importers;
+                        // - the importee is TLA-tainted: the fallback emits a plain call without
+                        //   `await`.
+                        let reexport_all_init_is_lazy = is_reexport_all
+                          && !importee_linking_info.has_dynamic_exports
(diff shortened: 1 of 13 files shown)


IWANABETHATGUY:
* **#9709** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9709?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9709?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9709).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Issue #9729, opened 2026-06-12
Manual grouping keeps extra dynamic-import proxy chunks compared to Rollup


aminpaks:
### Reproduction

https://github.com/aminpaks/rolldown-dynamic-imports-extra-proxy-chunks

```sh
git clone https://github.com/aminpaks/rolldown-dynamic-imports-extra-proxy-chunks
cd rolldown-dynamic-imports-extra-proxy-chunks
pnpm install
pnpm run verify
```

### Setup

`src/main.js` dynamically imports three modules:

```js
await Promise.all([
  import('./a.js'),
  import('./b.js'),
  import('./c.js'),
]);
```

Both Rollup and Rolldown are configured with the same grouping intent: put `a.js`, `b.js`, and `c.js` into `shared-abc`.

### Expected behavior

Rolldown matches Rollup's output shape: dynamic imports are rewritten to load the merged chunk directly.

Rollup output:

```text
dist-rollup/
├─ main-*.js
└─ shared-abc-*.js
```

### Actual behavior

Rolldown emits the merged chunk, but keeps one proxy chunk per original dynamic-import target:

```text
dist-rolldown/
├─ main-*.js
├─ a-*.js           # proxy imports shared-abc
├─ b-*.js           # proxy imports shared-abc
├─ c-*.js           # proxy imports shared-abc
└─ shared-abc-*.js
```

Each proxy is effectively:

```js
import {r as A} from './shared-abc-*.js';
export {A};
```

So `import('./a.js')` loads `a-*.js`, then `a-*.js` loads `shared-abc-*.js`.

### Verify output

```text
bundler   version  chunks  shared-abc  a/b/c proxies
--------  -------  ------  ----------  -------------
Rolldown  1.1.0    5       yes         3
Rollup    4.60.4   2       yes         0
```

### Question

Is this proxy-preserving output intentional? If so, is there an option for users who want Rollup-like behavior where dynamic imports point directly at the merged chunk?

hyfdev:
https://github.com/aminpaks/rolldown-dynamic-imports-extra-proxy-chunks shows 404

hyfdev:
Currently the behavior is intentional, but we noticed the this behavior and are thinking how to improve it. @IWANABETHATGUY cc

aminpaks:
> https://github.com/aminpaks/rolldown-dynamic-imports-extra-proxy-chunks shows 404

My bad, it should be public now.

aminpaks:
[...]

PR #9732, opened 2026-06-12
Draft: collapse dynamic manual group facades

aminpaks:
## What

Draft prototype for issue #9729 that removes dynamic-entry proxy chunks when manual code splitting has already moved the entry module into a manual group.

## Why

Issue #9729 shows Rolldown preserving import/re-export proxy chunks for manually grouped dynamic imports, while Rollup rewrites the dynamic import to the shared manual chunk with namespace extraction. This draft demonstrates the implementation shape before asking maintainers to confirm the desired semantics.

Flat manual groups still load the full group. When entries should avoid loading modules they do not reach, use `entriesAware: true`; the draft includes a multi-entry fixture showing `main2` dynamically imports `a/b` without loading the `c` subgroup.

Relates to https://github.com/rolldown/rolldown/issues/9729

## Tophat

Run the fixtures and inspect that dynamic-entry proxy chunks are removed while namespace values remain correct:

```sh
cargo run -q -p rolldown_testing --bin run-fixture crates/rolldown/tests/rolldown/optimization/chunk_merging/dynamic_manual_group_facade/_config.json
cargo run -q -p rolldown_testing --bin run-fixture crates/rolldown/tests/rolldown/optimization/chunk_merging/dynamic_manual_group_multi_entry/_config.json
```

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -952,10 +952,22 @@ impl GenerateStage<'_> {
     input_base: &ArcStr,
     module_is_assigned: &mut IndexBitSet<ModuleIdx>,
     temp_chunk_opt_graph: &ChunkOptimizationGraph,
+    allow_common_chunk_merges: bool,
   ) {
-    // Find empty dynamic entry chunks that should be merged with their target common chunks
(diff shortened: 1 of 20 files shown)


IWANABETHATGUY:
Closed, can't give any suggestion due to this pr is in the wrong direction,

PR #9733, opened 2026-06-13
perf: defer link-stage-output drop to rayon workers after output is produced

Brooooooklyn:
## Summary

Once the bundle output exists, the hot thread still spends ~15ms (on a 20k-module build) inside `free()` tearing down the link stage output (`module_table`, `metas`, `stmt_infos`, ...). The rayon workers are idle at that point, so this PR ships that drop to one of them (`utils/defer_drop.rs`):

- `symbol_db` is `mem::take`'n out for the cache merge **exactly as main consumes it** (inline, unchanged on both incremental and non-incremental paths).
- The remaining `link_stage_output` is `spawn_drop`'d right after `generate()` returns instead of being freed at `bundle_up` scope end.

Only this one object is deferred, governed by one rule (documented in the module): **only values main itself kept alive through the overlapped region are eligible.** main held `link_stage_output` through the `render_error`/`generateBundle` hooks and the write tail anyway — here it is freed concurrently *during* them, so no memory window ever extends (peak RSS measured flat).

Deferred drops cannot pile up — this is enforced, not assumed: every pending drop is counted, and `defer_drop::drain()` blocks until the count is zero at every entry point that starts rayon work (`BundleFactory::build_bundle` plus the three HMR partial-scan entries). In steady state it is a single uncontended lock check.

## Numbers

apps/10000 (20,014 modules), full build incl. minify + sourcemap, binary A/B vs main, order-balanced ABAB pairs, first run per binary discarded, 1.3x-median scan filter:

| pair | base build_ms | this PR | delta |
|---|---|---|---|
| 1 | 203.1 | 194.7 | −8.4 |
| 2 | 199.4 | 195.6 | −3.8 |
| 3 | 209.0 | 194.0 | −15.1 |
| 4 | 193.4 | 183.4 | −9.9 |
| 5 | 200.3 | 184.4 | −15.9 |
| 6 | 205.2 | 190.7 | −14.5 |
| 7 | 193.6 | 185.1 | −8.5 |
| 8 | 215.7 | 188.5 | −27.2 |

**Median pair delta −12.2 ms (~−6% of build), 8/8 pairs negative.** Output is byte-identical (`main.js` / `main.js.map` sha256 match the unmodified baseline). Peak RSS flat (`/usr/bin/time -l`, interleaved runs).

## What was deliberately NOT deferred

Adversarial review (4 rounds) shrank this change to what's provably safe:

- **`ast_table` (per-module AST arenas):** main frees these *before* chunk instantiation/minify allocate; deferring them would overlap the full arena graph with allocation-heavy phases and risk a peak-RSS spike. Kept inline.
- **non-incremental `symbol_db`:** main frees it inline *before* the output hooks; deferring it would extend its lifetime across unbounded plugin hooks. Kept inline (this costs ~5ms of the potential win and is the right trade).

## Notes

- The free still happens — on otherwise-idle workers, overlapping the hook/write tail. Real work is moved off the critical path, not eliminated; in-process memory is released a few ms later within the same build span.
- The pending counter is process-global on purpose: the dropped value is exclusively owned, so the worst cross-instance effect is a bounded (~15ms) wait or background CPU, never correctness.

---

> [!NOTE]
> **Medium Risk**
> Touches build lifecycle and the shared rayon pool with global synchronization; correctness relies on documented invariants (one deferred object per build, drain at entry), though output is intended to stay byte-identical.
> 
> **Overview**
> Moves teardown of heavy **link stage output** (`module_table`, metas, stmt infos, etc.) off the main thread after `GenerateStage::generate()` returns, so ~15ms of `free()` can overlap plugin hooks and the write tail on idle rayon workers.
> 
> Adds **`utils/defer_drop`**: `spawn_drop` enqueues exclusive drops on `rayon::spawn`, with a process-global pending counter and **`drain()`** so deferred work cannot pile up across builds. **`symbol_db`** is still taken with `mem::take` and merged into the cache inline before deferral (unchanged lifecycle for cache integrity).
> 
> **`defer_drop::drain()`** runs at every rayon-work entry: **`BundleFactory::build_bundle`** and the three HMR/lazy-compile paths in **`impl_bundler_hmr.rs`** that skip `build_bundle`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9f95690a8459952aca2809620338f25036502a0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/rolldown/src/bundle/bundle.rs
+++ b/crates/rolldown/src/bundle/bundle.rs
@@ -277,7 +277,13 @@ impl<Fs: FileSystem + Clone + 'static> Bundle<Fs> {
     // performance; reinstate it here, before the fallible steps below, so the
     // cache stays whole on their `Err` paths.
     // See meta/design/bundler-data-lifecycle.md ("Cache integrity on a failed build").
-    self.merge_immutable_fields_for_cache(link_stage_output.symbol_db);
+    self.merge_immutable_fields_for_cache(std::mem::take(&mut link_stage_output.symbol_db));
+
+    // `link_stage_output` is dead from here on (its `symbol_db` was just taken
+    // for the cache merge); ship the remaining heavy fields (module_table,
+    // metas, stmt_infos, ...) to a rayon worker so their free() happens off
+    // the critical path.
+    crate::utils::defer_drop::spawn_drop(link_stage_output);
 
     if let Err(errors) = &bundle_output {
       debug_assert!(errors.iter().all(|e| e.severity() == Severity::Error));
--- a/crates/rolldown/src/bundle/bundle_factory.rs
+++ b/crates/rolldown/src/bundle/bundle_factory.rs
@@ -152,6 +152,12 @@ impl BundleFactory {
     resolver: SharedResolver<Fs>,
     cache: ScanStageCache,
   ) -> Bundle<Fs> {
+    // Every build passes through here exactly once before any scan/link work
+    // starts. Wait for the previous build's deferred drops to retire so they
+    // can never overlap this build's rayon work; a no-op in steady state.
+    // See `utils::defer_drop` for the full invariant.
+    crate::utils::defer_drop::drain();
+
     let bundle_span = self.generate_unique_bundle_span();
     let module_infos = Arc::clone(&self.module_infos_for_incremental_build);
     let transform_dependencies = Arc::clone(&self.transform_dependencies_for_incremental_build);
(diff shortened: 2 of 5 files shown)


hyf0 (approved):

hyfdev:
I think this optimization is useful. My concern is more about how we keep it solid going forward:

- It feels a bit hard-coded to this specific lifecycle point.
- How should future code know when deferred drop is allowed?
- Do we need a dedicated benchmark/guardrail for the workload where this helps?

IWANABETHATGUY (approved):

PR #9735, opened 2026-06-13
test(dev): debug instrumentation for node-dev-server-test CI flake

h-a-n-a:
## What

Temporary, greppable instrumentation to diagnose the intermittent `node-dev-server-test` timeout (the job exceeds the 15-min ceiling and is canceled). Seen on **node 20** while node 22/24 pass the same commit in <2 min, so it's a nondeterministic hang. **Can't reproduce locally**, hence pushing instrumentation to CI.

## Working theory

The browser playground specs mutate a **shared on-disk fixture** (`playground-temp/<name>/`) via anchored string replacement, with **no per-test/per-retry reset** (the playground is copied once in global setup). A flaky first-attempt failure leaves the fixture off-baseline; after that, retries and sibling specs can never recover because `editFile` **silently no-ops** when its anchor string is gone — so the expected rebuild/HMR never fires and the test burns its full `testTimeout × 3 retries`. With specs running sequentially and sharing the dir, one flaky failure snowballs into the 15-min cancellation.

Evidence from the failing run: `rebuild-error.spec.ts` is the first to fail, then `hmr-full-bundle-mode.spec.ts` fails all 5 tests (~11 min), with repeated `[editFile] No changes detected …` warnings throughout.

## Instrumentation (grep prefixes)

- **`[envdbg]`** — node version, cpus/mem/loadavg, resolved vitest config, spec files, copied playgrounds (compare node 20 vs 22/24).
- **`[fixtdbg] ── BEFORE`** — full fixture-tree snapshot at the start of *every* attempt, with `attempt#`/`retryCount`. Confirms whether retries/sibling specs inherit a corrupted fixture.
- **`[fixtdbg] ── FAILED`** (`onTestFailed`) — error + fixture snapshot + server/browser log tails, printed **inline** so it survives the 15-min cancellation (Vitest's batched summary doesn't).
- **`[editFile]`** — logs `APPLIED before -> after`; the `NOOP` case now prints the current file content.
- **`hanging-process`** reporter on both suites — surfaces the `close timed out after 10000ms` / orphaned-process symptom.
- **`RD_TEST_RETRY` / `RD_TEST_TIMEOUT`** env overrides — set `RD_TEST_RETRY=0` to fail fast and let the whole suite finish inside the job budget on a repro run.

## Note

This is **diagnostic-only** and meant to be reverted (or trimmed to whatever turns out to be load-bearing) once the root cause is confirmed. No production code touched.

### DIFF
--- a/packages/test-dev-server/tests/playground/hmr-full-bundle-mode/__tests__/rebuild-error.spec.ts
+++ b/packages/test-dev-server/tests/playground/hmr-full-bundle-mode/__tests__/rebuild-error.spec.ts
@@ -1,5 +1,5 @@
 import { describe, expect, test } from 'vitest';
-import { editFile, page, serverLogs, waitForBuildStable } from '~utils';
+import { editFile, getBuildSeq, page, readFile, serverLogs, waitForBuildStable } from '~utils';
 
 // Covers the design principles in meta/design/dev-engine.md for a rebuild
 // failure:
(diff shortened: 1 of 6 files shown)


h-a-n-a:
Failed: https://github.com/rolldown/rolldown/actions/runs/27463457201/job/81182612946#step:9:604

Successful: https://github.com/rolldown/rolldown/actions/runs/27463457201/job/81182294575#step:9:607

Something in the watcher seems not to be working. File edits sometimes do not reflect in watcher updates even in poll mode. Will investigate. cc @sapphi-red @hyf0

PR #9738, opened 2026-06-13
perf(tree-shaking): hoist already-included guard to call sites in inclusion DFS

Brooooooklyn:
## Summary

In the tree-shaking inclusion DFS, the "already included?" dedup is a `set_bit` guard checked **inside** `include_module`, *after* the call frame + the `ctx.modules[idx]` match + `has_side_effects()` have already been paid. On apps/10000 (20,014 modules), `include_module` is called **169,219×** but only **~19,013 are unique — 88.8% are no-op re-visits** that do setup work only to be rejected by the guard. `include_statement` shows the same pattern (~3.5× revisit).

This hoists a read-only `has_bit()` pre-check to the two hot recursive call sites:
- the per-dependency `for_each` loop in `include_module`, and
- the owner-module re-inclusion in `include_symbol` (called per-symbol, where the owner is very often already included),

so already-included modules are skipped **before** the call + match + side-effect setup. The `set_bit` guard inside `include_module` remains the authoritative first-visit dedup — this is a pure work-skip.

## Correctness

Behavior-preserving and **byte-identical**, verified two ways on apps/10000:
- **Inclusion-set equality:** a debug dump of the final included module-set + per-module statement-set is byte-for-byte identical between baseline and this change (sha256-identical, `diff` empty, ~19,013 modules) — the tree-shaking decision is provably unchanged.
- **Output:** `main.js` and `main.js.map` sha256 match the unmodified baseline exactly.

## Performance

The work removed is deterministic: ~150k redundant `include_module` call frames (+ their match/side-effect setup) eliminated per build. On my local M5 the wall delta sits below the machine's ambient-load noise floor, so I'm relying on **CodSpeed** here to measure the instruction-count reduction cleanly — this PR is up specifically so CI can adjudicate the win.

`+10 / -1`, one file. fmt/clippy clean.

---

> [!NOTE]
> **Low Risk**
> Pure work-skip ahead of existing dedup logic in link-stage tree-shaking; no change to inclusion semantics when implemented as described.
> 
> **Overview**
> **Tree-shaking inclusion DFS** now skips dependency modules that are already marked included **before** loading the module from `ctx.modules`, checking side effects, or calling `include_module`.
> 
> In `include_module`’s per-dependency loop, a read-only `is_module_included_vec.has_bit(dependency_idx)` early return avoids redundant work on repeat graph visits (many `include_module` calls are no-ops that only hit the existing `set_bit` guard inside the function). **Inclusion decisions are unchanged** — `set_bit` in `include_module` remains the authoritative dedup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 715de3a524cae56e6b4e383dcc8e0003d52f1f0b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/rolldown/src/stages/link_stage/tree_shaking/include_statements.rs
+++ b/crates/rolldown/src/stages/link_stage/tree_shaking/include_statements.rs
@@ -706,6 +706,12 @@ pub fn include_module(ctx: &mut IncludeContext, module: &NormalModule) {
 
   // Include imported modules for its side effects
   module_meta.dependencies.iter().copied().for_each(|dependency_idx| {
+    // Guard-hoist: skip already-included dependencies before paying the
+    // `ctx.modules[idx]` match + `has_side_effects()` check. The authoritative
+    // dedup is still `set_bit` inside `include_module`; this is a pure work-skip.
+    if ctx.is_module_included_vec.has_bit(dependency_idx) {
+      return;
+    }
     match &ctx.modules[dependency_idx] {
       Module::Normal(importee) => {
         if !ctx.tree_shaking || importee.side_effects.has_side_effects() {


crates/rolldown/src/stages/link_stage/tree_shaking/include_statements.rs
```diff
@@ -834,7 +840,10 @@ pub fn include_symbol(
       .for_each(|stmt_info_id| {
         include_statement(ctx, module, stmt_info_id);
       });
-    if !is_simulated_facade_chunk {
+    if !is_simulated_facade_chunk && !ctx.is_module_included_vec.has_bit(module.idx) {
```
IWANABETHATGUY: This extra condition check seems redundant, since the first line of `include_module` is 
https://github.com/rolldown/rolldown/blob/edb097a1dcb940ba7f7e31c40aa3141f91171401/crates/rolldown/src/stages/link_stage/tree_shaking/include_statements.rs#L655-L658 module inclusion check

IWANABETHATGUY (approved):

PR #9749, opened 2026-06-15
fix: avoid O(N^2) rendering of high-volume diagnostics (#9748)

### LINKED ISSUE #9748: [Bug]: Build deadlocks (promise never resolves) when emitting a very high volume of warnings

Observed via a Vite 8 + Storybook build (~32k modules) that emits ~44,000 warnings, almost all `MODULE_LEVEL_DIRECTIVE` from MUI (`"use client"`). The triggering condition is a build that surfaces tens of thousands of warnings through the napi callback. No isolated standalone reproduction at this time.

### What is expected?

The build completes regardless of warning count. Warning emission/replay should never be able to deadlock the napi bridge.

The build hangs indefinitely — the build promise never resolves, CPU drops to ~0%, and memory plateaus (deadlock signature, not a slow build). Setting `build.rolldownOptions.logLevel: "silent"` makes the identical build complete in ~120s every time, which points at warning handling. An `onwarn` handler that mutes warnings (returns early without forwarding) does **not** prevent the hang, so the warnings appear to cross the Rust→JS napi boundary before the JS handler runs. Same hang / CPU-0% signature as vitejs/vite#21957 (different trigger); same class as the fixed sync-NAPI deadlock audit in #7311.

### System Info

```
System:
  OS: macOS 26.3.1
  CPU: (10) arm64 Apple M1 Max
  Memory: 20.73 GB / 64.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 24.13.0
  npm: 11.6.2
  pnpm: 11.5.1
Browsers:
  Chrome: 149.0.7827.103
  Firefox: 142.0
  Safari: 26.3.1
npmPackages:
  rolldown: 1.0.3 (transitive, via vite@8.0.16)
  vite: 8.0.16
  storybook: 10.3.3
  @storybook/react-vite: 10.3.3
```

### Any additional comments?

Surfaced through `storybook@10.3.3` / `@storybook/react-vite@10.3.3` on `vite@8.0.16` (resolves `rolldown@1.0.3`). Workaround in place: `build.rolldownOptions.logLevel: "silent"`.

IWANABETHATGUY:
Fixes #9748.

## Problem

Emitting a high volume of warnings in one large module made the build appear to hang. Rendering was `O(N^2)` for two independent reasons, both paid once **per diagnostic**:

1. The ariadne `Source` (its rope + full line index) was rebuilt for the whole file on every diagnostic.
2. `is_context_too_long` built a whole-file `ropey::Rope` on every call, even though only the ~300 chars around the span matter.

With tens of thousands of warnings pointing into the same file, this quadratic work is what actually stalled the build.

## Fix

- Add `Diagnostic::render_batch`, which builds the per-source ariadne `Source` cache and a per-source `LineTable` once and shares them across all diagnostics. Line lookup is `O(log lines)` per diagnostic; the UTF-16 column is read by scanning only the matched line.
- Replace the per-call `ropey::Rope` in `is_context_too_long` with direct string slicing of the prefix/postfix around the span, removing the `ropey` dependency entirely.
- Dispatch warning callbacks **sequentially** in both the binding (`handle_warnings`) and the watcher (`emit_warnings`), awaiting each before the next. A handler may `throw` from `onwarn`/`onLog` to abort the build, so calls must stop at the first failure without racing later in-flight callbacks. The expensive part of #9748 was the per-warning re-render, not the JS round-trip, so this does not reintroduce the hang.

## Residual cost (documented, not a regression)

ariadne still prints the entire labelled line for every diagnostic, so N diagnostics all landing on one very long line (e.g. a minified bundle) cost `O(N * line_len)` inside ariadne itself. That is bounded by the same per-line cost and is unrelated to the cross-diagnostic `O(N^2)` blow-up this PR removes.

## Tests

- `delivers every warning when a build emits a high volume of them` — 20k EVAL warnings complete quickly and every warning is delivered exactly once.
- `throwing from a warning handler aborts before invoking later handlers` — guards the sequential-dispatch / throw-to-abort contract.
- A `LineTable` unit test validates line/column/UTF-16 offset math against an independent linear reference across multi-byte, astral, and trailing-newline edge cases.

> [!note]
> Will consolidate the existing byte locator in https://github.com/rolldown/rolldown/pull/9762

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3224,7 +3224,6 @@ dependencies = [
  "oxc",
  "oxc_resolver",
  "rolldown-ariadne",
- "ropey",
  "rustc-hash",
  "sugar_path",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -199,7 +199,6 @@ regex = "1.12.4"
 regress = "0.11.0"
 rolldown-notify = "10.2.0"
 rolldown-notify-debouncer-full = "0.7.5"
-ropey = "1.6.1"
 rustc-hash = { version = "2.1.1" }
 schemars = "1.0.0"
 self_cell = "1.2.0"
--- a/crates/rolldown_binding/src/utils/mod.rs
+++ b/crates/rolldown_binding/src/utils/mod.rs
@@ -8,6 +8,7 @@ pub mod normalize_binding_options;
 
 use std::any::Any;
 
+use futures::stream::{StreamExt, TryStreamExt};
 use napi_derive::napi;
 use rolldown::{LogLevel, NormalizedBundlerOptions};
 use rolldown_error::{BuildDiagnostic, DiagnosticOptions, filter_out_disabled_diagnostics};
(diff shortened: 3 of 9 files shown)


IWANABETHATGUY:
* **#9749** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9749?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/9749?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=9749).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

IWANABETHATGUY:
This has been verified to solve the issue https://github.com/rolldown/rolldown/issues/9748#issuecomment-4712714527

IWANABETHATGUY:
@codspeedbot explain why this is faster

shulaoda (approved):

PR #9755, opened 2026-06-15
fix: Make external_import_binding_merger deterministic

### LINKED ISSUE #9754: [Bug]: Nondeterministic build output for externalized default imports with different local names

https://github.com/naruaway-sandbox/20260615-rolldown-nondeterministic-build-repro

### What is expected?

When the build input is fixed, the build output should be fixed (deterministic)

Rolldown randomly choose a symbol name to be used in the build output in this case, which makes the build non deterministic

### System Info

```Shell
System:
    OS: macOS 26.4.1
    CPU: (16) arm64 Apple M4 Max
    Memory: 15.47 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 26.0.0 - /opt/homebrew/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 11.12.1 - /opt/homebrew/bin/npm
    pnpm: 11.5.2 - /opt/homebrew/bin/pnpm
    bun: 1.3.14 - /opt/homebrew/bin/bun
    Deno: 2.6.6 - /opt/homebrew/bin/deno
    Watchman: 2026.01.12.00 - /opt/homebrew/bin/watchman
  Browsers:
    Safari: 26.4
```

### Any additional comments?

I just put "System Info" as instructed but the repro https://github.com/naruaway-sandbox/20260615-rolldown-nondeterministic-build-repro uses Pnpm with a fixed Pnpm/Node.js version (latest) for better reproducibility of the issue

naruaway:
Fixes #9754.

## Root cause

When naming an external **default** import, the name was derived from `IndexSet::first()` on the merged symbol set in `external_import_binding_merger`. `first()` reflects raw insertion (module load) order, so the chosen name — and therefore the generated output — could vary between builds, producing nondeterministic output.

## Fix

- Pick the default import name deterministically via `min_by_key((stable_id, name))` instead of `first()`.
- Iterate the binding-merger entries in sorted key order.

Adds the `issues/9754` regression fixture.

---

I verified that this patch stabilizes the output for the build in the repro: https://github.com/naruaway-sandbox/20260615-rolldown-nondeterministic-build-repro

However, I am not sure this fix is the right fix so feel free to close this PR and fix the issue in a different way if it's preferred

### DIFF
--- a/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
+++ b/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
@@ -187,10 +187,20 @@ impl LinkStage<'_> {
     self.external_import_namespace_merger = binding_ctx.external_import_namespace_merger;
 
     for (module_idx, map) in &binding_ctx.external_import_binding_merger {
-      for (key, symbol_set) in map {
+      let mut imported_binding_merger = map.iter().collect::<Vec<_>>();
+      imported_binding_merger.sort_by(|left, right| left.0.as_str().cmp(right.0.as_str()));
+
+      for (key, symbol_set) in imported_binding_merger {
         let name = if key.as_str() == "default" {
           let key = symbol_set
-            .first()
+            .iter()
+            .min_by_key(|symbol_ref| {
+              let symbol_ref = **symbol_ref;
+              (
+                self.module_table.modules[symbol_ref.owner].stable_id().as_str(),
+                symbol_ref.name(&self.symbols),
+              )
+            })
             .map_or_else(|| key.clone(), |sym_ref| sym_ref.name(&self.symbols).into());
           Cow::Owned(key)
         } else if is_validate_identifier_name(key.as_str()) {
--- a/crates/rolldown/tests/rolldown/issues/9754/_config.json
+++ b/crates/rolldown/tests/rolldown/issues/9754/_config.json
@@ -1,0 +1,11 @@
+{
+  "expectExecuted": false,
+  "config": {
+    "platform": "node",
+    "input": [
+      { "name": "entry0", "import": "entry0.js" },
+      { "name": "entry1", "import": "entry1.js" }
+    ],
+    "external": ["node:process"]
+  }
+}
--- a/crates/rolldown/tests/rolldown/issues/9754/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/issues/9754/artifacts.snap
@@ -1,0 +1,26 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+---
+# Assets
+
+## entry0.js
+
+```js
+import process from "node:process";
+//#region mod0.js
+console.log(0, process.pid);
+//#endregion
+export {};
+
+```
+
+## entry1.js
+
+```js
+import process from "node:process";
+//#region mod1.js
+console.log(1, process.pid);
+//#endregion
+export {};
+
+```
--- a/crates/rolldown/tests/rolldown/issues/9754/entry0.js
+++ b/crates/rolldown/tests/rolldown/issues/9754/entry0.js
@@ -1,0 +1,1 @@
+import './mod0.js';
--- a/crates/rolldown/tests/rolldown/issues/9754/entry1.js
+++ b/crates/rolldown/tests/rolldown/issues/9754/entry1.js
@@ -1,0 +1,1 @@
+import './mod1.js';
--- a/crates/rolldown/tests/rolldown/issues/9754/mod0.js
+++ b/crates/rolldown/tests/rolldown/issues/9754/mod0.js
@@ -1,0 +1,3 @@
+import process from 'node:process';
+const process2 = 0;
+console.log(process2, process.pid);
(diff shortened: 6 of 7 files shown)


IWANABETHATGUY:
Nice fix — I verified locally that it stabilizes the output for the repro in #9754.

Could you add a regression test so this stays locked in? A fixture under `crates/rolldown/tests/rolldown/issues/9754/` would do it:

```js
// _config.json
{
  "expectExecuted": false,
  "config": {
    "platform": "node",
    "input": [
      { "name": "entry0", "import": "entry0.js" },
      { "name": "entry1", "import": "entry1.js" }
    ],
    "external": ["node:process"]
  }
}

// entry0.js
import "./mod0.js";

// entry1.js
import "./mod1.js";

// mod0.js
import process from "node:process";
const process2 = 0;
console.log(process2, process.pid);

// mod1.js
import process2 from "node:process";
const process = 1;
console.log(process, process2.pid);
```

The snapshot pins the chosen name (`process`, since `mod0.js` sorts before `mod1.js`), so any future regression that brings the flip back will show up as a diff.

naruaway:
@IWANABETHATGUY Thanks for the review!
I added the test and confirmed that the test is actually effective by temporarily changing the implementation. 
I replaced `min_by_key` to `max_by_key` and confirmed that this changes the chosen symbol as expected in the snapshot like this 👍 :

```
❯ just t-run crates/rolldown/tests/rolldown/issues/9754/_config.json
NEEDS_EXTENDED=false cargo run-fixture crates/rolldown/tests/rolldown/issues/9754/_config.json
   Compiling rolldown v1.1.1 (/Users/naru/repos/github.com/rolldown/rolldown/r/crates/rolldown)
   Compiling rolldown_dev v1.1.1 (/Users/naru/repos/github.com/rolldown/rolldown/r/crates/rolldown_dev)
   Compiling rolldown_testing v0.1.0 (/Users/naru/repos/github.com/rolldown/rolldown/r/crates/rolldown_testing)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.10s
     Running `target/debug/run-fixture crates/rolldown/tests/rolldown/issues/9754/_config.json`
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot file: crates/rolldown/tests/rolldown/issues/9754/artifacts.snap
Snapshot: artifacts
Source: crates/rolldown_testing/src/integration_test.rs:598
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
    2     2 │
    3     3 │ ## entry0.js
    4     4 │
    5     5 │ ```js
    6       │-import process from "node:process";
          6 │+import process2 from "node:process";
    7     7 │ //#region mod0.js
    8       │-console.log(0, process.pid);
          8 │+console.log(0, process2.pid);
    9     9 │ //#endregion
   10    10 │ export {};
   11    11 │
   12    12 │ ```
   13    13 │
   14    14 │ ## entry1.js
   15    15 │
   16    16 │ ```js
   17       │-import process from "node:process";
         17 │+import process2 from "node:process";
   18    18 │ //#region mod1.js
   19       │-console.log(1, process.pid);
         19 │+console.log(1, process2.pid);
   20    20 │ //#endregion
   21    21 │ export {};
   22    22 │
   23    23 │ ```
────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
updated snapshot /Users/naru/repos/github.com/rolldown/rolldown/r/crates/rolldown/tests/rolldown/issues/9754/artifacts.snap
```

crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
```diff
@@ -187,10 +187,20 @@ impl LinkStage<'_> {
     self.external_import_namespace_merger = binding_ctx.external_import_namespace_merger;
 
     for (module_idx, map) in &binding_ctx.external_import_binding_merger {
-      for (key, symbol_set) in map {
+      let mut imported_binding_merger = map.iter().collect::<Vec<_>>();
+      imported_binding_merger.sort_by(|left, right| left.0.as_str().cmp(right.0.as_str()));
```
naruaway, line 191: Note that this sort does not affect the case I reported but assuming there is no perf impact, I would keep it for sure

naruaway:
BTW I think the added test cannot reliably detect the issue since the issue itself is not deterministically reproducible
So I think the test here can just cause flakiness when the issue comes back

I guess the ideal solution is to rely on https://github.com/rolldown/zarara/pull/32 (but I guess we need to run the same build many times to reliably detect this kind of issues)

IWANABETHATGUY (approved):

PR #7322, opened 2025-12-03
chore: enable tracing feature for napi

sapphi-red:
close #7287
refs https://github.com/rolldown/rolldown/pull/7317, https://github.com/napi-rs/napi-rs/pull/3041

With this change, I get the tracing output:

```
$env:RD_LOG="napi=trace"; pnpm build

> @example/typescript@ build D:\documents\GitHub\rolldown\examples\basic-typescript
> rolldown --config ./rolldown.config.js

  2025-12-03T03:05:29.770480Z DEBUG napi: BindingBundler::constructor
    at crates\rolldown_binding\src\binding_bundler.rs:26

  2025-12-03T03:05:29.772016Z DEBUG napi: BindingBundler::write
    at crates\rolldown_binding\src\binding_bundler.rs:26

  2025-12-03T03:05:29.786956Z DEBUG napi: BindingRenderedChunkMeta::chunks
    at crates\rolldown_binding\src\options\plugin\types\binding_render_chunk_meta_chunks.rs:15

  2025-12-03T03:05:29.788489Z DEBUG napi: BindingBundler::close
    at crates\rolldown_binding\src\binding_bundler.rs:26

  2025-12-03T03:05:29.788696Z DEBUG napi: shutdownAsyncRuntime
    at crates\rolldown_binding\src\lib.rs:79

  2025-12-03T03:05:29.788905Z DEBUG napi: BindingOutputChunk::getFileName
    at crates\rolldown_binding\src\types\binding_output_chunk.rs:17

  2025-12-03T03:05:29.789008Z DEBUG napi: BindingOutputChunk::getCode
    at crates\rolldown_binding\src\types\binding_output_chunk.rs:17

<DIR>/entry.js  chunk │ size: 0.23 kB

✔ rolldown v1.0.0-beta.52 Finished in 21.24 ms
```
(note: I had to comment out `debug` option, otherwise I got `failed to set global default subscriber: SetGlobalDefaultError("a global default trace dispatcher has already been set")` error)

Issue #7287, opened 2025-12-01
Track every `#[napi]` call using tracing crate


hyf0:
Refs:

- #7182

sapphi-red:
https://github.com/napi-rs/napi-rs/pull/3041

sapphi-red:
#7322

PR #9766, opened 2026-06-15
docs: document why the napi tracing feature is enabled

Boshen:
Keep the napi `tracing` feature enabled and document the reason in `Cargo.toml`.

With the feature on, `#[napi]` logs every binding call, which helps debug deadlock issues like #7289 by showing which napi function is pending.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -231,6 +231,8 @@ quote = "1"
 syn = { version = "2", default-features = false }
 
 # napi
+# Keep the `tracing` feature on: `#[napi]` logs every binding call, which helps debug deadlocks
+# (e.g. #7289) by showing which napi function is pending.
 napi = { version = "3.0.0", features = ["async", "anyhow", "tracing", "object_indexmap"] }
 napi-build = { version = "2.2.2" }
 napi-derive = { version = "3.0.0", default-features = false, features = ["type-def", "tracing"] }


hyfdev:
I think we previously enable this for a reason. Would you mind wait for feedbacks?

Brooooooklyn:
@sapphi-red does rolldown still need this feature?

hyfdev:
Retitled from `perf:` → `chore:` — no runtime path is touched here, this only shrinks the binding binary size, so it isn't a perf change.

On the change itself: this `tracing` feature was added intentionally in #7322 (closing #7287, *"Track every `#[napi]` call"*). It auto-instruments every `#[napi]` export so `RD_LOG="napi=trace"` prints each binding call — the tooling we relied on to debug NAPI deadlocks/hangs (seeing which binding call never returns). I'd rather not delete it outright.

Better approach: put it behind conditional compilation (a cargo feature, off by default), so the size cost is gone from normal builds but we can still flip it on when we need to chase a deadlock — instead of having to re-add the feature and rebuild mid-incident.

If you'd still rather just remove it, I won't block it.

sapphi-red:
I have the same thoughts with the previous one: https://github.com/rolldown/rolldown/pull/8702

Boshen:
@claude[agent] we can't remove this feature, need to document along the feature usage in Cargo.toml https://github.com/rolldown/rolldown/pull/8702#issuecomment-4062657082

Boshen:
### Merge activity

* **Jun 17, 7:58 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 17, 8:05 AM UTC**: `Boshen` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 17, 8:12 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

IWANABETHATGUY (approved):

PR #9777, opened 2026-06-15
refactor: remove unnecessary `Arc` around sourcemap sender

camc314:
This PR replaces `Arc<std::sync::mpsc::Sender<SourceMapGenMsg>>` with a cloned `Sender<SourceMapGenMsg>`.

`std::sync::mpsc::Sender is already a clonable multi-producer handle. Cloning a sender creates another handle to the same channel, so wrapping it in an `Arc` ads another unnecessary layer of shared ownership witout chaning semantics.

Before, callsers shared

```
  Arc<Sender<SourceMapGenMsg>>
```

After, callers pass:

```
  Sender<SourceMapGenMsg>
```

Each cloned Sender still sends to the same receiver owned by the sourcemap worker thread. The receiver, shutdown behavior, and message propagation remain the same; this change only removes the extra Arc allocation/refcount indirection around a type that already provides the sharing model needed here.


camc314:
@codex review this PR

IWANABETHATGUY (approved):

shulaoda (approved):

PR #9783, opened 2026-06-16
fix: wrap node require helper in pure IIFE

### LINKED ISSUE #9782: [Bug]: import.meta.url is preserved in output even when createRequire(import.meta.url) helper is unused

Reproduction Link - [REPL](https://repl.rolldown.rs/#eNptUMtuxCAM/BXLl22lKHvn2t/gQsBs6BKDwGm3ivLvJUR9SZWQjMczI3s29Kg2DOzoMUo9/ozqpx/QtrYVQiVlpX3AkmJ06Z1Hm9iH2y/VP5NTH5acisAGjnxgeulj2MGXtMDlS3bRrJkendqIZo29fgueNs0AgfMqCk7LcSExY4eGY5ijEZ/KojRycqSxo1KI6mzupLyJleB6BZkJfIhUoVIkK+Rg+ujotIokhvbI2BnETIdFk6QsIXHVvD9rxhZENvZubjS+1tSuPyP4g53H9601OsrUUmUbqGpUsO3N6TBqTm+N1janKrh/Aoofik8=) link

Hi Team,
As you can see in the REPL link, `import.meta.url`, which is used in the `createRequire` function, is left in the output even though `createRequire` itself is not used and gets removed by Rolldown / Oxc.

This seems unexpected because the input file is empty, but the output still contains a standalone `import.meta.url`.

### What is expected?

The expected output should be like the below one 
```
import "node:module";
//#endregion
export {};
```

I think this is happening because, as mentioned in the [Oxc docs](https://oxc.rs/docs/guide/usage/minifier/dead-code-elimination.html#pure-pure), only the function call is treated as pure, not its arguments.

So in a helper like this:

`export var __require = /* @__PURE__ */ createRequire(import.meta.url);`

createRequire(...) can be removed, but the argument import.meta.url may still be preserved.

Using a pure IIFE seems to solve the issue, as mentioned in the Oxc docs:

`export var __require = /* #__PURE__ */ (() => createRequire(import.meta.url))();`

This way, the whole expression, including import.meta.url, can be removed when the helper is unused.

### System Info

```Shell
System:
    OS: Windows 11 10.0.26200
    CPU: (16) x64 13th Gen Intel(R) Core(TM) i5-13500H
    Memory: 1.54 GB / 15.75 GB
  Binaries:
    Node: 24.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - C:\Users\96110\AppData\Roaming\npm\yarn.CMD
    npm: 11.13.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 11.4.0 - C:\Users\96110\AppData\Local\pnpm\bin\pnpm.CMD
  Browsers:
    Chrome: 149.0.7827.114
    Edge: Chromium (149.0.4022.69)
    Internet Explorer: 11.0.26100.8115
  npmPackages:
    rolldown: workspace:* => 1.1.1
```

### Any additional comments?

_No response_

kb019:
fixes #9782
Hi Team,
This PR fixes an issue where `import.meta.url` was preserved in the output even when the `createRequire(import.meta.url)` helper was unused and removed.

The issue happens because the pure annotation on `createRequire(import.meta.url)` only marks the function call as pure, but not the argument. Because of that, `import.meta.url` can still be left in the final output as a standalone expression.

This PR wraps the helper initialization in a pure IIFE, so the whole expression can be removed when the helper is unused.

### DIFF
--- a/crates/rolldown/src/runtime/runtime-tail-node.js
+++ b/crates/rolldown/src/runtime/runtime-tail-node.js
@@ -1,4 +1,4 @@
 // This is created by the rolldown. It's used to polyfill the global `require` function when you bundling code
 // with target `format: 'es'` and `platform: 'node'`. It's used to make sure the global `require` could
 // works in Node.js esm environment.
-export var __require = /* @__PURE__ */ createRequire(import.meta.url);
+export var __require = /* #__PURE__ */ (() => createRequire(import.meta.url))();
--- a/crates/rolldown/tests/rolldown/function/resolve/alias_to_node_builtin_module/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/function/resolve/alias_to_node_builtin_module/artifacts.snap
@@ -9,7 +9,7 @@ source: crates/rolldown_testing/src/integration_test.rs
 import { createRequire } from "node:module";
 //#endregion
 //#region main.js
-(/* @__PURE__ */ createRequire(import.meta.url))("fs");
+(/* @__PURE__ */ (() => createRequire(import.meta.url))())("fs");
 //#endregion
 export {};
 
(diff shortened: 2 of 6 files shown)


sapphi-red (approved):

kb019:
Thanks @sapphi-red

PR #9789, opened 2026-06-16
fix: avoid panic on `output.file` without a file name

shulaoda:
## Problem

A bundle with an `output.file` that has no file-name component crashes the whole build instead of reporting a config error.

`Chunk::generate_preliminary_filename` derives the basename from the raw, unvalidated user option:

```rust
let basename = PathBuf::from(file)
  .file_name()
  .expect("The file should have basename")
  .to_string_lossy()
  .to_string();
```

`Path::file_name()` returns `None` for paths with no final file-name component, e.g. `""`, `"/"`, `"."`, `".."`, or any path ending in `..` (such as `"dist/.."`). Any of those as `output.file` hits the `.expect(...)` and panics:

```js
export default { output: { file: '.' } } // or '/', '', 'dist/..'
```

The panic message (`The file should have basename`) gives the user no actionable information. Rollup rejects such values gracefully.

## Fix

Validate `output.file` upfront in `verify_raw_options`, where the other output options are already validated, so a value without a file name is rejected as a recoverable `INVALID_OPTION` config error before bundling starts:

```text
[INVALID_OPTION] Invalid value "." for option "output.file" - it does not contain a file name. Please provide a path that ends with a file name.
```

This keeps input validation in a single, discoverable place. The downstream `expect` in `generate_preliminary_filename` is left as-is: it now holds as an invariant guaranteed by the upfront check (`NormalizedBundlerOptions` is only ever built in `prepare_build_context`, which runs `verify_raw_options` first), so it is no longer reachable for invalid input.

Added an `errors/invalid_option/output-file-without-name` snapshot test covering the new diagnostic.


h-a-n-a (dismissed):

crates/rolldown_common/src/chunk/mod.rs
```diff
@@ -191,11 +191,13 @@ impl Chunk {
     used_name_counts: &FxDashMap<ArcStr, u32>,
   ) -> anyhow::Result<PreliminaryFilename> {
     if let Some(file) = &options.file {
-      let basename = PathBuf::from(file)
+      let basename = Path::new(file)
         .file_name()
-        .expect("The file should have basename")
+        .ok_or_else(|| {
+          anyhow::anyhow!("Invalid `output.file` {file:?}: it does not contain a file name.")
```
h-a-n-a: Maybe we should put this in the validation function in `rolldown/crates/rolldown/src/utils/prepare_build_context.rs`? so that we can see clearly on how inputs are get validated.

h-a-n-a (approved):

IWANABETHATGUY (approved):

h-a-n-a:
### Merge activity

* **Jun 16, 10:20 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 16, 10:20 AM UTC**: `h-a-n-a` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 16, 10:23 AM UTC**: The [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) couldn't merge this PR because **it failed for an unknown reason** (Fast-forward merges are not supported for forked repositories. Please create a branch in the target repository in order to merge).

PR #9790, opened 2026-06-16
perf: remove the `url` crate (-226 KB)

Boshen:
## Do we really need the `url` crate?

This is a draft to ask exactly that. `url` is one of our heavier dependencies relative to what it actually does for us.

### What it costs

`url` is pulled in **only** by rolldown crates, and it drags the entire IDNA/Unicode stack along:

```
url -> idna -> idna_adapter -> ICU4X
                               (icu_normalizer / icu_properties + *_data + zerovec / yoke / tinystr / ... — ~21 crates)
```

That is **~226 KB** of the shipped `librolldown_binding.node` (`__const` Unicode tables + `url`/`idna` code), all for internationalized-domain-name handling (`münchen.de` -> `xn--mnchen-3ya.de`) that rolldown never needs. `url` mandates `idna` for special-scheme hosts (incl. `file:`) and has **no feature** to turn it off.

### What we actually use `url` for (3 sites)

| site | usage |
|---|---|
| `rolldown_plugin_vite_resolve/src/file_url.rs` | parse `file://` URLs -> path (Vite's `fileURLToPath`) |
| `rolldown_binding/.../normalize_binding_options.rs` | validate `sourcemapBaseUrl` + ensure trailing `/` |
| `rolldown/.../process_code_and_sourcemap.rs` | join the map filename onto `sourcemapBaseUrl` |

All three only ever see ASCII (file paths / sourcemap URLs), never IDN hosts.

### This PR

Replaces those three sites with targeted parsing and drops `url`:
- a hand `file://` splitter (host / path / query / fragment), keeping the exact posix/windows `fileURLToPath` rules and the `%2F` / hostname rejections
- `normalize_sourcemap_base_url()`: validate `scheme://…` and append a trailing `/`
- the base is normalized to end in `/`, so the join becomes a plain append

11 new unit tests, all passing; matches the existing `output/sourcemap-base-url` fixture's expected `//# sourceMappingURL=…` output.

### Impact (aarch64, LTO release, same worktree A/B)

| build | size |
|---|---|
| baseline (`url` + `idna` + ICU4X) | 23,675,392 |
| **this PR (`url` removed)** | **23,443,904 — −226 KB** |

`url` / `idna` / `idna_adapter` / `icu_*` are gone from the binary (`nm` confirms 0 symbols; they remain in `Cargo.lock` only behind the `jsonschema` **test** dependency).

### Open questions (why this is a draft)

- **Correctness trade-off**: we'd now own URL parsing / joining / validation at these 3 sites instead of `url`'s WHATWG-compliant implementation. Unit tests cover the common cases, but `url`'s edge cases (path `.`/`..` normalization, exotic percent-encoding) are no longer guaranteed. Acceptable for `sourcemapBaseUrl`, a public option that mirrors Rollup?
- **Not yet validated here**: the JS integration fixtures (need a full napi build + vitest) and the Windows `file://` path.
- **Less invasive alternative**: keep `url` and only swap `idna`'s ICU backend for a tiny ASCII-only `idna_adapter` via `[patch.crates-io]` — saves **~129 KB** (57% of this) with **zero** change to `url`'s behavior. Easy to switch to if we'd rather not bring URL handling in-house.

Opening as a draft to get a call on the size-vs-correctness trade before polishing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3009,7 +3009,6 @@ dependencies = [
  "testing_macros",
  "tokio",
  "tracing",
- "url",
  "walkdir",
  "xxhash-rust",
 ]
@@ -3136,7 +3135,6 @@ dependencies = [
  "string_wizard",
  "sugar_path",
  "tracing",
- "url",
 ]
 
 [[package]]
@@ -3752,7 +3750,6 @@ dependencies = [
  "serde",
  "serde_json",
  "sugar_path",
- "url",
  "urlencoding",
 ]
 
@@ -4636,18 +4633,6 @@ name = "unsafe-libyaml"
 version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "673aac59facbab8a9007c7f6108d11f63b603f7cabff99fabf650fea5c32b861"
-
-[[package]]
-name = "url"
-version = "2.5.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ff67a8a4397373c3ef660812acab3268222035010ab8680ec4215f38ba3d0eed"
-dependencies = [
- "form_urlencoded",
- "idna",
- "percent-encoding",
- "serde",
-]
 
 [[package]]
 name = "urlencoding"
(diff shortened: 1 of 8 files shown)


Boshen:
Superseded by #9811, which takes the least-invasive route both this PR and #9794 identified: keep `url` untouched and just disable `idna`'s ICU backend by pinning `idna_adapter` to its no-ICU 1.0.0 release (−129 KB, no hand-rolled URL parser to own).

PR #9793, opened 2026-06-16
perf(rolldown): skip redundant canonical_ref lookup in chunk symbol deconfliction

hyfdev:
Skip a redundant `canonical_ref` union-find lookup per declared symbol in `deconflict_chunk_symbols`: the caller already resolved the canonical ref, so pass it to a new `add_canonical_symbol_in_root_scope` instead of re-resolving it inside `add_symbol_in_root_scope`. Behavior-preserving.

Prepared with AI assistance.

### DIFF
--- a/crates/rolldown/src/utils/renamer.rs
+++ b/crates/rolldown/src/utils/renamer.rs
@@ -166,6 +166,23 @@ impl<'name> Renamer<'name> {
   /// other top-level names and nested scope names that could cause capture.
   pub fn add_symbol_in_root_scope(&mut self, symbol_ref: SymbolRef, needs_deconflict: bool) {
     let canonical_ref = symbol_ref.canonical_ref(self.symbol_db);
+    self.add_canonical_symbol_in_root_scope(canonical_ref, needs_deconflict);
+  }
+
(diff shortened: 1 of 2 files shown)


crates/rolldown/src/utils/renamer.rs
```diff
@@ -166,6 +166,23 @@ impl<'name> Renamer<'name> {
...
   pub fn add_symbol_in_root_scope(&mut self, symbol_ref: SymbolRef, needs_deconflict: bool) {
     let canonical_ref = symbol_ref.canonical_ref(self.symbol_db);
+    self.add_canonical_symbol_in_root_scope(canonical_ref, needs_deconflict);
+  }
+
+  /// Same as [`Self::add_symbol_in_root_scope`] but takes an already-canonicalized
+  /// `canonical_ref`. Hot callers (e.g. `deconflict_chunk_symbols`) that have already
+  /// resolved the canonical ref use this to avoid a redundant `canonical_ref` union-find
+  /// walk. `canonical_ref` MUST be canonical (i.e. `symbol_ref.canonical_ref(db)`).
+  pub fn add_canonical_symbol_in_root_scope(
```
hyfdev, line 176: I think this is a bug-prone change, this require human-invloved judement to correctly call some functions.

PR #9794, opened 2026-06-16
perf: replace the `url` crate with `iri-string` (-210 KB)

Boshen:
## Do we need the `url` crate? No — a 0-dep RFC 3986 parser covers it.

`url` is pulled in **only** by rolldown crates, and it drags the entire IDNA/Unicode stack along:

```
url -> idna -> idna_adapter -> ICU4X (icu_normalizer / icu_properties + *_data + zerovec / yoke / ... — ~21 crates)
```

That's ~210 KB of the shipped `librolldown_binding.node` for internationalized-domain-name handling (`münchen.de` -> punycode) that rolldown never needs — and `url` has no feature to disable `idna`.

[`iri-string`](https://crates.io/crates/iri-string) is a **zero-dependency**, RFC 3986/3987 URI parser. It treats the host as opaque text, so it has **no IDNA/ICU** at all, while still being a real, spec-compliant parser (validation + relative resolution) rather than hand-rolled string slicing.

### What we use `url` for, and the replacement

| site | usage | with `iri-string` |
|---|---|---|
| `file_url.rs` | parse `file://` -> path | `UriStr` components (authority/path/query/fragment) |
| `normalize_binding_options.rs` | validate `sourcemapBaseUrl` + trailing `/` | `UriStr::new` validation + append `/` |
| `process_code_and_sourcemap.rs` | join map filename onto base | base is normalized to end in `/`, so plain append |

10 new unit tests, all passing; matches the `output/sourcemap-base-url` fixture.

### Impact (aarch64, LTO release, same-worktree A/B)

| approach | size | saved | parser |
|---|---|---|---|
| baseline (`url` + `idna` + ICU4X) | 23,675,392 | — | WHATWG |
| `idna_adapter` ASCII stub (keeps `url`) | 23,543,232 | −129 KB | WHATWG, needs a vendored crate + `[patch]` |
| **`iri-string` (this PR)** | **23,460,448** | **−210 KB** | **real RFC 3986, no ICU, 0 deps** |
| full hand-roll (#9790) | 23,443,904 | −226 KB | hand-written |

`iri-string` captures **93%** of removing `url` entirely (−210 of −226 KB), only 16 KB behind hand-rolling, but the tricky parts (`file://` parsing, RFC-3986 resolution) are done by a real parser. It removes `url` + `idna` + the ICU4X stack from the binary (`idna`/ICU remain only behind the `jsonschema` **test** dep).

### Trade-off

`iri-string` is RFC 3986 (stricter than `url`'s WHATWG) — it rejects a few malformed inputs `url` would auto-normalize (e.g. raw unencoded spaces). For the well-formed `file://`/sourcemap URLs rolldown actually sees they agree, and `sourcemapBaseUrl` validation just gets marginally stricter.

### Relationship to #9790

This is the **recommended alternative** to #9790 (which hand-rolls the parsing): near-identical savings, but no hand-written URL parser to own. Draft to pick between the two approaches.

Not yet validated here: the JS integration fixtures (need a full napi build + vitest) and the Windows `file://` path.


Boshen:
Superseded by #9811. Rather than swapping the parser, #9811 keeps `url` as-is and disables `idna`'s ICU backend via the `idna_adapter = "=1.0.0"` pin — the −129 KB "keep `url`" option from this PR's table, but using the official no-ICU `idna_adapter` release instead of a vendored stub + `[patch]`.

PR #9795, opened 2026-06-16
perf: remove num-format dependency from vite reporter

Boshen:
## What

`num-format` was used at exactly one place — `crates/rolldown_plugin_vite_reporter/src/utils.rs`'s `display_size`, to insert thousands separators (`quotient.to_formatted_string(&Locale::en)`) into the build size report, e.g. `1,234.56 kB`.

This PR replaces that single call with a small `group_thousands` helper that inserts a `,` every three digits from the right, then drops `num-format` from the crate manifest and the root `[workspace.dependencies]`.

## Why (size mechanism)

`num-format` bundles a static data table for 500+ locales (separators, grouping rules, digit shapes, etc.) so it can format numbers for any locale. We only ever needed `Locale::en` — plain 3-digit grouping with `,`. By porting the ~10-line helper we drop the entire crate and its embedded locale table from the dependency graph, which shrinks the shipped `rolldown_binding` cdylib (`.node`). The helper reuses the `itoa` dependency the crate already has.

## Correctness

The output is byte-identical to `num_format`'s `Locale::en` (standard 3-digit grouping, `,` separator). Added a unit test covering edge cases:

- `0 -> "0"`, `5 -> "5"`, `999 -> "999"`
- `1000 -> "1,000"`, `12345 -> "12,345"`, `1234567 -> "1,234,567"`, `1000000 -> "1,000,000"`

`cargo test -p rolldown_plugin_vite_reporter` passes, clippy (pedantic) is clean, and `cargo check -p rolldown_binding` compiles. Reporter output / snapshots are unchanged.

**Binary size impact:** **−32 bytes (negligible)** on the stripped release cdylib (fat-LTO, `strip="symbols"`, baseline `697a7c4e58`). The 500+-locale data table was already removed by the linker's `dead_strip` (only `Locale::en` was ever referenced), so dropping the crate yields no meaningful `.node` size change. The value of this PR is one fewer dependency + faster compiles, **not** binary size.


shulaoda (approved):

shulaoda:
### Merge activity

* **Jun 16, 7:00 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 16, 7:00 PM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 16, 7:06 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9802, opened 2026-06-16
perf(oxc-runtime): store embedded helpers deflate-compressed

Boshen:
## Summary

The `OxcRuntimePlugin` embeds every `@oxc-project/runtime` helper — both the ESM and CJS variants, 232 entries total — directly in the binary as uncompressed JS string literals (`arcstr::literal!`). That's ~187 KiB of `.rodata` that genuinely ships in the `.node` cdylib (the plugin is unconditionally registered, so the data is not dead-stripped).

This PR stores each helper body **raw-DEFLATE-compressed** (no zlib/gzip header, max level) and inflates it **lazily on first use**, caching the inflated `ArcStr` so each helper is inflated at most once per process. Helpers are already loaded at most once per helper per build behind an existing cache, so the runtime cost is negligible.

## Why this is ~free at runtime

The `flate2` crate (with the `zlib-rs` backend) is **already linked** into the binary via `oxc_compat -> oxc-browserslist -> flate2` (single version `1.1.9` across the whole graph). Decompressing the helpers reuses that already-present inflate engine, so the win is real `.rodata` savings rather than just trading data for code.

Measured section breakdown (release cdylib, macOS aarch64):

| section | base | branch | delta |
| --- | --- | --- | --- |
| `__TEXT,__const` (`.rodata`) | 1,453,816 | 1,334,008 | **−117 KiB** |
| `__TEXT,__text` (code) | 17,461,544 | 17,467,576 | +6 KiB |

The +6 KiB of code is just the thin inflate/cache wrapper — confirming the inflate engine itself was already linked.

## Implementation

- The generator (`tasks/generator/src/generators/oxc_runtime_helper.rs`) now emits each helper body as raw-DEFLATE bytes (`&'static [u8]`) in the `phf` maps, keyed identically. The map value is a small `Helper { compressed, len }` struct.
- The accessor inflates with `flate2::read::DeflateDecoder` and memoizes per-helper in a process-wide cache (keyed by the entry's `'static` address, which is unique per entry across both maps).
- Regenerated `embedded_helpers.rs` via `cargo run --bin generator`.

## Output is byte-identical

The emitted helper JS does not change. Verified across **all 232 helpers** that the inflated output is byte-identical to the upstream `@oxc-project/runtime` sources, plus unit tests covering non-empty inflation of every entry, exact spot-checks of known helpers, and cache stability. No snapshots changed.

**Binary size impact:** −96.2 KiB shipped `.node`/dylib (23,476,336 → 23,377,776 bytes), of which the embedded-helper `.rodata` reduction is −117 KiB (partly offset by +6 KiB inflate/cache code and macOS 16 KiB segment page padding).


IWANABETHATGUY (approved):

IWANABETHATGUY:
IIRC, you designed the oxc runtime embedd solution. Can you also have a look?

Boshen (commented):
The code is bad, I'll come back to this.

Boshen:
Closing in favor of tracking issue #10043 (assigned to @shulaoda).

PR #9826, opened 2026-06-17
docs: relocate meta/design to internal-docs, split design from implementation

h-a-n-a:
## Motivation

Per the maintainer discussion: `meta/` is an opaque prefix that's hard to discover. Move the docs to a self-describing top-level `internal-docs/`, and separate the **design/principles** (_why_) from **implementation** (_how_) so principles become first-class rather than buried inside implementation-focused docs.

## What changed

- **`meta/design/` → top-level `internal-docs/`**, one folder per feature. Every feature folder has an `implementation.md` (the _how_); features with substantial, separable design content also get a `design.md` (the _why_).
- **Split into `design.md` + `implementation.md`:** `dev-engine`, `watch-mode`, `devtools`, `cache`, `lazy-compilation`. `template/` is now a `design.md` / `implementation.md` template pair.
- **Left whole (implementation-only):** docs whose "why" is a thin, embedded aside or that are short and coherent (e.g. `module-tags`, `plugin-asset-module`) — splitting them would fragment a tight narrative. A `design.md` can be added later, on demand.
- **All references rewired:** code comments, inter-doc links, `AGENTS.md`, docs, and the generated `_config.schema.json` / `binding.d.cts`. Design-vs-implementation references are routed by *meaning* (e.g. error-replay refs point at `design.md` principles, not implementation sections), not by blind string replace.
- **`AGENTS.md`** documents the new one-folder-per-feature convention.

## Verification

- Every inter-doc link resolves; every external code/doc reference resolves to a real file.
- No `meta/design` references remain outside `CHANGELOG.md` (historical entries, intentionally left).
- The only unresolved links are the `template/` placeholder examples and a pre-existing dead `app-scenario-chunking` link in `manual-code-splitting`.

## Notes for reviewers

- The generated `_config.schema.json` and `binding.d.cts` had the path swap applied to match their (also-updated) Rust sources; a normal `just build-rolldown` reproduces them identically.
- Docs renames are git-tracked as renames; `dev-engine` etc. show as rename + content split.


IWANABETHATGUY:
All minor — none blocking.

- Pre-existing dead link, now touched. internal-docs/manual-code-splitting/implementation.md:99 → ../app-scenario-chunking/implementation.md (no such file). The PR correctly flags it as pre-existing, but since this PR is already rewiring that exact line, it'd be cheap to either drop the link or convert it to plain text rather than carry a known-broken link into the new structure.
- Template placeholder links will trip a link-checker. template/design.md:30 and template/implementation.md:25 link to ../other-feature/implementation.md as illustrative examples. If link-check CI is ever added (this reorg makes that more likely), these fail. Consider rendering them as inline code (`../other-feature/implementation.md`) instead of real []() links so they read as examples, not targets.

hyfdev (approved):

PR #9828, opened 2026-06-17
perf: compile out debug/trace tracing in release builds

Boshen:
## What

Adds a `release_max_level_info` cargo feature on `rolldown_binding` that turns on `tracing/release_max_level_info`, statically compiling out `debug!`/`trace!` events and `#[instrument(level = "debug" | "trace")]` spans in the shipped binary. The feature is enabled by `build-binding:release` and `build-binding:wasi:release`.

## Why

The pipeline carries 79 `#[tracing::instrument]` attributes (57 of them `debug`/`trace`), and `tracing` had no `release_max_level` set — so all of that span machinery (the `Instrumented` future wrapper, field recording, and enter/exit guards woven into every `poll`) was compiled into the release binary.

Measured on `librolldown_binding.dylib` (darwin-arm64, fat-LTO + stripped):

| build | size |
| --- | --- |
| baseline | 23,327,440 B |
| this PR | 22,892,768 B |
| **delta** | **−434,672 B (−1.9%, ~0.42 MiB)** |

## Notes

- Opt-in, not default: `build-binding:profile` (chrome tracing) is intentionally left uncapped so profiling keeps its `debug`/`trace` spans. The `tracing` cascade picks the *most restrictive* enabled `release_max_level_*`, so this cannot be a default that the profile build overrides — hence the feature is enabled only on the release scripts.
- No effect on debug builds or tests (`release_max_level_*` only applies when `debug_assertions` is off).
- `info`/`warn`/`error` logging in the shipped binary is unchanged; only `debug`/`trace` are removed.


IWANABETHATGUY:
### Merge activity

* **Jun 18, 3:17 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.

IWANABETHATGUY:
Closed since Codex’s findings are correct; merging this PR would break `devtools` in vite/rolldown.

PR #9831, opened 2026-06-17
refactor: single-source the chunk $N symbol-naming algorithm

Dunqing:
`renamer.rs` (chunk-internal deconfliction, producing `canonical_names`) and `compute_cross_chunk_links.rs` (cross-chunk export aliases, producing `exports_to_other_chunks`) each carried their own copy of the same `$N` conflict-suffix naming and `Reverse(exec_order)` ordering — kept in sync only by a hand-written pinned-SHA comment, with no test guarding the contract. This consolidates both into one source of truth.

### What changed

- A shared `ConflictResolver` owns the `$N` naming kernel; both sites call it instead of re-implementing the loop.
- A single `deconflict_order_key` owns the ordering. It returns a borrowed `&str` (so the export sort no longer allocates a `CompactStr` per comparison) and takes `(&ModuleTable, &SymbolRefDb)` rather than the whole `&LinkStageOutput`.
- The pinned-SHA sync comments are gone, replaced by a `debug_assert` that enforces the `chunk.modules` exec-order invariant in code.
- Unit tests now pin the kernel and the ordering — the contract that previously had none.

### Behavior

Unchanged. Chunk-internal naming is byte-identical and cross-chunk export naming is output-neutral — verified by the full Rust test suite (zero snapshot drift) and clippy.

### Hot path

The deconfliction path stays as tight as the original inline loops: `resolve` takes its base by value, so the name the caller already owns moves straight through to the result rather than being re-allocated, and it is `#[inline]`d so the loop compiles down like the hand-written code. The cross-chunk resolver is pre-sized and `"default"` is built with `CompactStr::new_const`.

Intentionally untouched: `NestedScopeRenamer`, the namespace/wrapper logic, the minified base54 export generator, and all read-only consumers of `canonical_names`.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
+++ b/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
@@ -1,8 +1,6 @@
-use std::borrow::Cow;
-use std::cmp::Reverse;
-
 use super::GenerateStage;
 use crate::chunk_graph::ChunkGraph;
+use crate::utils::chunk::conflict_resolver::{ConflictResolver, deconflict_order_key};
 use crate::utils::chunk::normalize_preserve_entry_signature;
 use crate::utils::external_import_interop::external_import_needs_interop;
 use itertools::{Itertools, multizip};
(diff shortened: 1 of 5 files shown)


IWANABETHATGUY:
@Dunqing, It seems the vite CI is broken on the main branch. Can you hold off on this for a few and wait for the CI to be fixed?

Dunqing:
> @Dunqing, It seems the vite CI is broken on the main branch. Can you hold off on this for a few and wait for the CI to be fixed?

Sure, I opened it for triggering the benchmark.

shulaoda (approved):

IWANABETHATGUY (approved):

IWANABETHATGUY:
### Merge activity

* **Jun 18, 10:22 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 18, 10:22 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 18, 10:28 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

camc314 (commented):
nice @Dunqing 🚀this is what I had in mind from our mtg!!

PR #9834, opened 2026-06-17
test(bench): add CodSpeed micro-benchmarks for rolldown_sourcemap

hyfdev:
## What

Isolated CodSpeed micro-benchmarks for **`rolldown_sourcemap`** — the per-chunk sourcemap machinery on the codegen hot path — added to the existing `bench` binary, so it gets its own deltas instead of being diluted inside the end-to-end bundle numbers.

## Cases (3)

Each case is a distinct code path, **sized to a large bundle chunk** so the measured op dominates the instruction count and a size-dependent regression shows. Inputs are **real oxc-codegen sourcemaps** (with `with_pre_compute_sourcemap_data(true)` to mirror production), built **outside** the timed loop so the measured region stays inside `rolldown_sourcemap`:

| Case | Exercises |
|---|---|
| `sourcemap/join_with_sourcemap` | `SourceJoiner::join` over a large chunk of mapped modules + prepended banner + footer → `ConcatSourceMapBuilder` (token buffer, source dedup, line-offset across plain↔map) |
| `sourcemap/join_no_sourcemap` | the `enable_sourcemap == false` fast path (plain modules, builder skipped) — the most common production build |
| `sourcemap/collapse_codegen_chain` | `collapse_sourcemaps` over a fine, per-token oxc chain (pretty→minify→pretty, 3 links, big program → tens of thousands of tokens) |

`join` rebuilds the builder + output per call, so building the joiner once outside `iter` still measures a full merge.

## Coverage

The surface is `SourceJoiner::join` (both branches) + `collapse_sourcemaps`. **Deliberately not separate cases:** a *coarse/sparse-map* collapse — at scale it's dominated by the same O(token-count) loop as the fine chain, so it'd largely duplicate `collapse_codegen_chain` (its only distinct bit, the lookup-miss branch, is a correctness concern with a unit test, not a perf path); name remap (oxc emits no names → dead here); `adjust_sourcemap_dst_lines` (rebuild shape already covered by collapse; shebang-only cold path); `append_source_dyn` (HMR).

## Notes

- **No CI change, no new link-graph crate, no link-time bloat** — synchronous, in the existing `bench` binary (already built/run by CodSpeed; does not reintroduce the separate bench binary removed in #8694); `rolldown_sourcemap` + `oxc` are already linked transitively via `rolldown`.
- Adversarially reviewed; `cargo clippy -p bench --benches` clean; cases run + are deterministic.


IWANABETHATGUY (approved):

h-a-n-a (approved):

shulaoda (approved):

hyfdev:
### Merge activity

* **Jun 18, 9:52 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 18, 9:52 AM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 18, 9:58 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #4232, opened 2025-04-21
feat: rolldown_filter_analyzer

IWANABETHATGUY:
### Description
1. The new crate is used to detect if a plugin hook has a potential opportunity to use `filter` to improve performance, the basic idea is pretty straightforward: Whether there exists a path that reaches a statement that returns undefined without encountering any sideeffects statement, if exists we assume the plugin could be filtered with `filter` attribute 
2. Will add js part in next pr.

PR #9838, opened 2026-06-17
feat: remove `rolldown_filter_analyzer`

camc314:
I may be missing context here, but it appears this crate doesn't do anything - it was added in #4232, but the JS side was never merged.

This PR removes this dead code, as well as removing the unused `cfg` feature from oxc.

As a result, this reduces binary size by ~30k bytes, and potentially improves performance as the cfg checks in side semantic builder will now be skipped


Boshen:
Landing https://github.com/rolldown/rolldown/pull/9839 for the actual change we need: remove cfg.

camc314:
> Landing #9839 for the actual change we need: remove cfg.

Worth mentioning that `rolldown_filter_analyzer` is dead code, so might make sense to delete

Issue #9843, opened 2026-06-18
[Bug]: PluginError wrapping loses id, plugin, and frame fields in error objects exposed to JS


franckchen:
## Reproduction steps

When using Vite 8 (`vite@8.0.x`) with rolldown as the bundler, trigger a build failure in a `.ts`/`.tsx` file (any syntax error processed by `builtin:vite-transform`), then inspect the error thrown by `build()`:

```js
build().catch(err => {
  console.log(err.errors[0].id);      // undefined  ← expected: file path
  console.log(err.errors[0].plugin);  // undefined  ← expected: 'builtin:vite-transform'
  console.log(err.errors[0].frame);   // undefined  ← expected: code snippet
  console.log(err.errors[0].loc);     // { line: N, column: N }  ✅ works
  console.log(err.errors[0].code);    // 'PLUGIN_ERROR'          ✅ works
});
```

## What is expected?

`err.errors[0]` should expose the same diagnostic fields that `RollupError` provided in rollup@4:

| Field    | Expected value |
|----------|----------------|
| `id`     | Absolute path of the file containing the error, e.g. `/project/src/main.tsx` |
| `plugin` | Name of the plugin that threw the error, e.g. `builtin:vite-transform` |
| `frame`  | Source code snippet around the error location |
| `loc`    | `{ line, column }` — already works ✅ |
| `code`   | Error code string — already works ✅ |

## What is actually happening?

`id`, `plugin`, and `frame` are all `undefined` in `errors[0]`. The information exists internally but is dropped in the error-wrapping pipeline.

## Root cause (traced through source, verified on latest main `01e4eee`)

**1.** `ViteTransformPlugin::transform` (`crates/rolldown_plugin_vite_transform/src/lib.rs`) throws a `BatchedBuildDiagnostic` wrapping a `BuildDiagnostic` of inner type `OxcError`. `OxcError` stores the file path in `self.id` and correctly implements `BuildEvent::id()` returning `Some(path)`.

**2.** In `BatchedBuildDiagnostic::from<anyhow::Error>` (`crates/rolldown_error/src/build_diagnostic/mod.rs:187`), each inner `BuildDiagnostic` is re-wrapped in a `PluginError`:
```rust
.map(|diag| BuildDiagnostic::plugin_error(plugin.clone(), diag.into()))
```

[...]

PR #9844, opened 2026-06-18
fix(plugin_error): preserve inner diagnostic's file id through plugin-error wrapping

hyfdev:
`PluginError` re-wraps each inner diagnostic but did not override `id()`, so the offending file path was dropped from the error objects exposed to JS (`err.errors[0].id` was `undefined` — a regression from rollup@4). Delegate `id()` to the inner diagnostic so `NativeError.id` / `loc.file` are populated. Addresses the `id` field of #9843 (`plugin`/`frame` need new binding fields — separate follow-up).

Prepared with AI assistance.


hyfdev:
Does not take a holistic approach on solving https://github.com/rolldown/rolldown/issues/9843

PR #9841, opened 2026-06-18
chore(deps): update oxc_resolver and oxc_resolver_napi to 11.21.3

Boshen:
Bumps `oxc_resolver` and `oxc_resolver_napi` from `11.21.0` to `11.21.3` (latest).

### What changed upstream

- **11.21.1** — replace the `papaya` caching library with `dashmap`; tsconfig handling fixes for non-TypeScript / `allowJs`-disabled `.js` importers.
- **11.21.2** — replace `fancy-regex` with `regress`.
- **11.21.3** — auto-discovered tsconfig now only attaches to files it actually owns; `normalize_with` refactored to reduce binary size.

In the lockfile this removes `papaya` (and its `seize` dependency) and bumps `pnp` `0.12.9` → `0.12.10`. `dashmap` and `regress` were already present in the dependency tree, so there are no net-new dependencies.

### Binary size impact

Measured on the `darwin-arm64` release binding (`profile.release`: `lto = "fat"`, `codegen-units = 1`, `strip = "symbols"`):

| binding | `main` (11.21.0) | this PR (11.21.3) | Δ |
| --- | --- | --- | --- |
| `rolldown-binding.darwin-arm64.node` | 23,327,472 B (22.24 MiB) | 23,045,856 B (21.97 MiB) | **−281,616 B (−275 KiB, −1.2%)** |

PR #9845, opened 2026-06-18
docs(tsconfig): align auto-discovery docs with oxc-resolver behavior

shulaoda:
Related to #9841

## What

Updates the `tsconfig` auto-discovery documentation (`packages/rolldown/src/options/docs/tsconfig.md`, injected into the `tsconfig` option's JSDoc) to reflect two oxc-resolver fixes that already ship in the version Rolldown uses (`11.21.3`).

## Why

The docs still described oxc-resolver's previous behavior, which no longer matches what the resolver does.

### 1. Fallback for unowned files (oxc-project/oxc-resolver#1220)

Previously, when no `tsconfig.json` owned a file, the resolver fell back to the outermost (topmost) ancestor `tsconfig.json`, leaking its `paths`/`baseUrl` to unrelated files. It now applies no config in that case (an inferred project with no path mapping), matching TypeScript and typescript-go.

Doc change:

> If no `tsconfig.json` owns the file, no config is applied (no `paths`/`baseUrl`), matching TypeScript.

### 2. Explicit non-TS extensions in `include` (oxc-project/oxc-resolver#1213)

The extension filter used to reject any non-TS/JS file up front, so an explicitly listed extension such as `src/**/*.vue` never matched. The filter now applies only to the default `**/*`-style globs: a bare/omitted `include` still matches only TS/JS, while an entry naming an explicit extension matches verbatim. This lets a `.vue` (or other non-TS) file pick up the project's `paths`/`baseUrl`.

This file-matching rule is general (it applies to both the root tsconfig and referenced projects), so it is now documented as a standalone paragraph instead of being buried in the `references` section:

> Whether a `files`/`include` entry matches a file depends on its extension: by default only TypeScript files (`.ts`/`.tsx`/`.mts`/`.cts`) match, plus `.js`/`.jsx`/`.mjs`/`.cjs` when `allowJs` is enabled. An entry that names an explicit extension (for example `src/**/*.vue`) matches that extension verbatim, so a non-TS file can be owned by the project and pick up its `paths`/`baseUrl`.


packages/rolldown/src/options/docs/tsconfig.md
```diff
@@ -2,9 +2,11 @@
...
+When set to `true`, Rolldown enables auto-discovery mode. For each module, both the resolver and transformer search **upward** from the module's directory, starting at the nearest `tsconfig.json`. If it has `references`, Rolldown checks each referenced project's `files`/`include`/`exclude` and uses the first one that matches the file. If no reference matches, it checks the `tsconfig.json`'s own `files`/`include`/`exclude`. If the file matches neither, Rolldown continues upward to the next `tsconfig.json` and repeats. If no `tsconfig.json` owns the file, no config is applied (no `paths`/`baseUrl`), matching TypeScript.
```
sapphi-red: I think "own" is not clear here.

hyfdev (approved):

h-a-n-a (approved):

sapphi-red (approved):

PR #9847, opened 2026-06-18
chore(skills): translate _config.json when encoding rolldown REPL links

IWANABETHATGUY:
## What

Improves the bundled `rolldown-repl` Claude skill so the **encoder** stops silently dropping a fixture's build config.

The REPL share format has no "options" field — but the REPL stores the bundle config as an ordinary `rolldown.config.ts` **file** in its file map (the one that does `input: import.meta.input`). The encoder now leverages that: when a fixture has a sibling `_config.json`, it translates the config into a generated `rolldown.config.ts` and embeds it in the payload, so the shared link actually reproduces the fixture.

## Changes

`scripts/encode.mjs`:
- Reads `_config.json`'s `config` (a flattened Rust `BundlerOptions`) and emits `defineConfig({...})` with the correct JS-API shape:
  - output options nested under `output` (`codeSplitting`, `format`, `entryFileNames`, …)
  - `define`/`dropLabels` → `transform`, `profilerNames` → `output.generatedCode`
  - `Filename` → `FileNames` renames; input options stay top-level
- **Auto-marks entries** from `config.input` (named entries preserved as `input: { name: 'file' }`, unnamed → `import.meta.input`), so `--entry` is usually unnecessary
- New flags: `--variant <name>` (merge a `configVariants` entry) and `--no-config` (files-only)
- Untranslatable bits (plugins, unknown options, skipped variants) are reported on **stderr**, keeping the URL the only thing on stdout

`SKILL.md`:
- Documents the `_config.json` → `rolldown.config.ts` translation and the new flags
- Adds a **temp-file hygiene rule**: round-trip/decode output must go to the OS temp dir (not the repo), and any temp files created must be deleted afterward

## Notes

The translation table covers the options fixtures commonly use; if rolldown's option surface changes, the tables in `encode.mjs` (`OUTPUT_FIELDS` / `OUTPUT_RENAMES` / `NESTED_FIELDS`) may need updating. Verified by round-tripping the `advanced_chunks/basic_cjs` fixture (incl. its `no-profiler-names` variant) and a synthetic fixture exercising `import.meta.input`, the filename renames, and `transform.define`/`dropLabels`.

### DIFF
--- a/.claude/skills/rolldown-repl/SKILL.md
+++ b/.claude/skills/rolldown-repl/SKILL.md
@@ -26,6 +26,8 @@ If the URL has no `#…` payload, there's nothing to decode — tell the user an
 
 Run the bundled scripts. They use plain Node (`node:zlib`, `node:buffer`, `node:fs`) — no `npm install` needed.
 
+> **Temp files: write outside the repo, and clean up after.** Never write decoded files, round-trip checks, or captured URLs into the project directory — they show up as untracked clutter in the user's working tree. Always target a path under the OS temp dir (POSIX: `/tmp/...`; Windows PowerShell: `$env:TEMP\...`), and **delete any temp file or directory you created once you're done** (e.g. `Remove-Item -Recurse -Force "$env:TEMP\repl-roundtrip"`). Do not pipe the URL to a `.txt` file inside the repo to "save" it — print it to the chat instead.
+
 **Decode** prints a JSON summary to stdout, and with `--write` drops each file onto disk:
 
 ```bash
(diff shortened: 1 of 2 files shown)


h-a-n-a (approved):

IWANABETHATGUY:
### Merge activity

* **Jun 18, 5:42 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 18, 5:42 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 18, 5:45 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9849, opened 2026-06-18
refactor: drop commondir crate in favor of in-house helper

Boshen:
## What

The external [`commondir`](https://crates.io/crates/commondir) crate was used in a single place (`external_module_task.rs`) and only for `CommonDir::try_new(...).common_root()` — duplicating the existing in-house `rolldown_utils::commondir` module of the same name.

This ports that one function into the internal module as `common_dir` and drops the dependency.

## Details

- The port preserves the original semantics exactly: the common root is the **deepest common ancestor bounded by the first path's parent**. This matters for the single-entry case (returns the entry's parent dir, not the entry itself), so a naive longest-common-prefix fold would have silently changed external-module render paths.
- Added unit tests for the multi-file, partially-shared, single-file, and empty-input cases.
- Updating the only caller and removing the dep also drops `thiserror` v1 (+ `thiserror-impl` v1) from the tree — `commondir` was its last remaining user, so everything now dedupes onto `thiserror` v2.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -323,7 +323,7 @@ version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0fa961b519f0b462e3a3b4a34b64d119eeaca1d59af726fe450bbba07a9fc0a1"
 dependencies = [
- "thiserror 2.0.18",
+ "thiserror",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


IWANABETHATGUY (approved):

IWANABETHATGUY:
### Merge activity

* **Jun 18, 6:46 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 18, 6:50 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 18, 6:57 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9851, opened 2026-06-18
perf: drop urlencoding crate in favor of percent-encoding

Boshen:
## What

Remove the `urlencoding` dependency and migrate its 3 consumers to `percent-encoding`, which is already in the dependency tree (it's a dependency of `url`).

- `rolldown_plugin_vite_resolve` (`file_url.rs`) — 2 decode sites
- `rolldown_plugin_data_url` (`lib.rs`) — 1 decode site
- `rolldown_testing` (`integration_test.rs`) — 2 encode sites

`urlencoding` is dropped from the workspace `Cargo.toml` and disappears from `Cargo.lock`.

## Why

It overlapped with `percent-encoding` (the canonical crate from the `url`/rust-url family), which is compiled regardless. This removes one crate from the dependency graph. `rolldown_plugin_utils` already uses `percent-encoding` (`src/uri.rs`), so this also unifies the codebase on a single percent-encoding library.

## Correctness

- **Decode** is semantically identical: `urlencoding::decode(s)?` → `percent_encoding::percent_decode_str(s).decode_utf8()?`. Both decode only `%XX`, leave `+` literal (these are percent-encoded paths / data URLs, not form bodies), and error on invalid UTF-8. All sites are in `anyhow::Result` fns, and both error types satisfy `?`.
- **Encode** uses an `AsciiSet` matching `urlencoding::encode`'s RFC 3986 unreserved set:
  ```rust
  const DATA_URL_ENCODE_SET: &AsciiSet =
      &NON_ALPHANUMERIC.remove(b'-').remove(b'.').remove(b'_').remove(b'~');
  ```
  i.e. encode everything except `A-Za-z0-9` and `-._~` — so the generated `data:text/javascript,…` URLs round-trip identically through the decode path.

## Verification

- `cargo check` / `cargo test` pass for all three crates (`rolldown_plugin_vite_resolve` unit tests green; `rolldown_plugin_data_url` has `test = false`).
- `urlencoding` removed from `Cargo.lock`; no version churn elsewhere.

## Note on impact

This is dependency-surface hygiene rather than a perf/size win — `urlencoding` is a tiny, zero-dependency crate and its replacement was already present. Net effect is one fewer crate in the graph.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3190,11 +3190,11 @@ version = "1.1.1"
 dependencies = [
  "arcstr",
  "base64-simd",
+ "percent-encoding",
  "rolldown_common",
  "rolldown_plugin",
  "rolldown_utils",
  "simdutf8",
- "urlencoding",
 ]
 
 [[package]]
@@ -3548,6 +3548,7 @@ dependencies = [
  "owo-colors",
  "oxc_resolver",
  "parking_lot",
+ "percent-encoding",
  "pnp",
  "rolldown_common",
  "rolldown_plugin",
@@ -3559,7 +3560,6 @@ dependencies = [
  "serde_json",
  "sugar_path",
  "url",
- "urlencoding",
 ]
 
 [[package]]
@@ -3630,6 +3630,7 @@ dependencies = [
  "json-strip-comments",
  "jsonschema",
  "oxc",
+ "percent-encoding",
  "regex",
  "rolldown",
  "rolldown_common",
@@ -3643,7 +3644,6 @@ dependencies = [
  "serde_json",
  "sugar_path",
  "tokio",
- "urlencoding",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


hyfdev (approved):

Boshen:
### Merge activity

* **Jun 18, 6:32 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 18, 6:45 AM UTC**: `Boshen` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 18, 6:50 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9880, opened 2026-06-18
fix: propagate errors from output.globals function

shulaoda:
When the user-supplied `output.globals` function returned an `Err`, `GlobalsOutputOption::call` collapsed it to `None` via `.ok()`. The iife/umd renderers treat `None` as "no global name configured", so a throwing globals callback silently fell back to the external's identifier name (plus a misleading `missing_global_name` warning) instead of surfacing the real error.

`call` now returns `anyhow::Result<Option<String>>` and the iife/umd call sites `?`-propagate it. A failing globals function now aborts the build with its actual error, matching Rollup. The `None` warning + fallback path now only fires for a genuine `output.globals` map miss, and a function returning `""` still maps to `Some("")` as before.


IWANABETHATGUY:
Test coverage — main gap

  - No test accompanies the fix. There's an existing fixture
  **tests/fixtures/output/format/iife/globals-with-function/** that's a
  near-perfect template — adding a sibling fixture whose `globals` function
  throws and asserting the build rejects (e.g. `await
  expect(build).rejects.toThrow(...)`) would lock in the new behavior and guard against a future regression to .ok().

IWANABETHATGUY (approved):

Issue #6099, opened 2025-09-09
sourcemapping did not generate correctly


IWANABETHATGUY:
> I think we have to pass the original span here to preserve the mapping. I expect `import_react` to be mapped from `react` and `version` to be mapped from `version`.
> 
> <img width="1638" height="344" alt="image" src="https://github.com/user-attachments/assets/b8648ab0-359b-42b7-ab8c-a6fad370c6f9" />
> 
[...]

PR #365, opened 2023-11-22
fix: deconflict for duplicate canonical names

hyf0:
### Description

Fixes #364

### Test Plan

---

PR #9881, opened 2026-06-18
perf(sourcemap): owned merge in SourceJoiner::join (4005→5 allocs/chunk)

hyfdev:
## What

`SourceJoiner::join` used to borrow every input sourcemap into `ConcatSourceMapBuilder`, then detach the combined map by deep-copying every name, source, and `sourcesContent` string. On a 2,000-module chunk that meant 4,005 allocations in `join` alone.

This change moves owned maps into the concat builder instead:

- `SourceJoiner::join` consumes the joiner and uses `add_sourcemap_owned` for maps it can take. A mixed-path fallback keeps borrowed `Source` implementations correct and copies only their borrowed entries.
- Normal chunk rendering now detaches each module's map before its source text is shared with the plugin-facing `RenderedModule` view. The text stays shared through `Arc`; a lightweight chunk source owns the map and moves it into the builder. This is required for the optimization to cover the real ESM/CJS/IIFE/UMD production path, not only synthetic owned inputs and HMR.
- `SourceMapSource` stores its map as an `Option` and exposes `take_sourcemap`, leaving the plugin-facing view mapless while preserving `RenderedModule.code`.
- The no-sourcemap fast path no longer scans every source for line offsets that only the sourcemap builder needs.
- `oxc_sourcemap` is bumped to `8.0.2`, which contains the owned concat API from oxc-project/oxc-sourcemap#368 (the released successor to #365).

## Measured impact

| metric | before | after |
|---|---:|---:|
| deterministic allocations, `sourcemap/join_with_sourcemap` (2,000 modules) | 4,005 | **5** (-99.9%) |
| CodSpeed simulation, `join_with_sourcemap` | 10.4 ms | **9.5 ms** (+9.31% efficiency) |
| CodSpeed simulation, `join_no_sourcemap` | 1.6 ms | **1.4 ms** (+10.96% efficiency) |
| per-`join` wall-clock in the original prototype (release, mimalloc) | ~417 µs | **~314 µs** (-25%) |

CodSpeed's aggregate verdict is a **10.13% performance improvement**. It used `main@4b9c35e` because a newer successful baseline was unavailable and flags a runtime-environment mismatch, so the exact percentages are directional. The allocation snapshot is committed and reproduced identically across two runs. `join` runs once per output chunk and remains a small part of a full build, so this is an allocation-pressure and per-operation improvement, not a large end-to-end claim.

## Correctness coverage

- Mixed owned + borrowed maps retain both mappings, source IDs, source contents, and line offsets.
- The plugin-facing `RenderedModule` still exposes identical code after its map is detached, while chunk rendering receives that map.
- Existing sourcemap fixtures pass, including CJS live bindings, source exclusion, debug IDs, inline URLs, issue #6099, post-banner/shebang mapping, and the esbuild source-map suite.

## Verification

- Full GitHub CI, including CodSpeed, passes at `a078bad82`.
- `cargo clippy -p rolldown_sourcemap -p rolldown -p bench --all-targets -- --deny warnings`
- `cargo test -p rolldown_sourcemap`
- `cargo test -p rolldown --lib` (70 passed)
- `cargo test -p rolldown --test integration -- --skip test262::test262_module_code` (26 passed; test262 is not initialized in the worktree)
- `cargo allocs` twice with an identical snapshot

## Upstream

- oxc-project/oxc-sourcemap#368 — merged and released as `oxc_sourcemap@8.0.2`

### DIFF
--- a/crates/rolldown_sourcemap/src/source_joiner.rs
+++ b/crates/rolldown_sourcemap/src/source_joiner.rs
@@ -48,8 +48,6 @@ impl<'source> SourceJoiner<'source> {
       + sources_len;
     let mut ret_source = String::with_capacity(size_hint_of_ret_source);
 
-    let mut line_offset = 0;
-
     let mut sourcemap_builder = self.enable_sourcemap.then(|| {
       ConcatSourceMapBuilder::with_capacity(
         self.names_len,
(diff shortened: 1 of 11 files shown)


crates/rolldown_sourcemap/src/source_joiner.rs
```diff
@@ -36,17 +36,18 @@ impl<'source> SourceJoiner<'source> {
     self.prepend_source.push(Box::new(source));
   }
 
-  pub fn join(&self) -> (String, Option<SourceMap>) {
+  pub fn join(mut self) -> (String, Option<SourceMap>) {
```
IWANABETHATGUY, line 39: https://crates.io/crates/rolldown_sourcemap

- File: `crates/rolldown_sourcemap/src/source_joiner.rs`
- Line: 39

Because `rolldown_sourcemap` is a published crate at version 1.1.2, changing `join` from `&self` to `self` is a breaking public API. Probably, we should mark this as a breaking change?

IWANABETHATGUY (approved):

Boshen:
Wait for upgrade from oxc-sourcemap.

crates/rolldown_sourcemap/src/source_joiner.rs
shulaoda: We currently make no guarantees regarding breaking changes in crates.

IWANABETHATGUY: > We currently make no guarantees regarding breaking changes in crates.

Good to know, but the agent still doesn't know unless you tell it explicitly, so I drafted a pr to fix it:  https://github.com/rolldown/rolldown/pull/9905

hyfdev:
Solved by https://github.com/rolldown/rolldown/pull/10250

PR #9886, opened 2026-06-19
perf(sourcemap): move owned maps into the concat builder when joining

Boshen:
## Summary

`SourceJoiner::join` borrowed each input `SourceMap` into `ConcatSourceMapBuilder` and then `into_owned_sourcemap` deep-copied every string — including the full `sourcesContent` (the original module text) — to produce the owned result. The joiner's inputs are owned (`SourceMapSource`) and discarded after the join, so we can **move** them in instead of borrowing and recopying.

- `Source::take_sourcemap(&mut self)` — new trait method (default `None`); `SourceMapSource` yields its map via `mem::take`.
- `SourceJoiner::join` now **consumes `self`** and feeds owned maps to the new `ConcatSourceMapBuilder::add_sourcemap_owned`, returning via `into_sourcemap` (no `into_owned`). All `join` call sites are terminal, so no caller changes were needed.

No name/source/`sourcesContent` string is copied during the join; the cost no longer scales with total source text size.

## Performance

From the upstream `concat_owned` bench (200 owned module maps, sweeping `sourcesContent` size) — median time per join:

| content/module | borrow + `into_owned_sourcemap` | move-in |
|---|---|---|
| 256 B | 37 µs | 32 µs |
| 2 KB | 47 µs | 32 µs |
| 8 KB | 70 µs | 33 µs |
| 32 KB | 542 µs | 33 µs |

Move-in is flat (it moves heap buffers, not bytes); the old path grows linearly with `sourcesContent`. Output `mappings` are byte-identical.

## Blocked on

Depends on the unreleased owned move-in API in oxc-project/oxc-sourcemap#368. Pinned via a temporary `[patch.crates-io]` git dependency on that branch. **Before merge:** that PR must land and be released, then drop the patch and bump the `oxc_sourcemap` version. Draft until then.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1997,8 +1997,7 @@ dependencies = [
 [[package]]
 name = "oxc_sourcemap"
 version = "8.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "174d2498c2390ccfab0a9caaf47ef5accda1d509aaa34a5be8f45e00e5299f04"
+source = "git+https://github.com/oxc-project/oxc-sourcemap.git?branch=perf%2Fconcat-owned-move-in#dbd6ef9f13e2721279bcbc7f8bbd65523447c323"
 dependencies = [
  "base64-simd",
  "json-escape-simd",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -310,3 +310,9 @@ opt-level = "s"
 # it will strip these information for non-debug files
 debug = "full"
 strip = "none"
+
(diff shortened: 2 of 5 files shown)


hyfdev:
I continued the original adoption in #9881 now that `oxc_sourcemap 8.0.2` is released.

Validation found two gaps: shared `RenderedModule` sources could silently lose their sourcemaps, and the consuming `join()` change left the main bench uncompilable.

#9881 now fixes both, covers the real ESM/CJS/IIFE/UMD path, updates the allocation snapshot from 4005 to 5, and has fully green CI.

To avoid maintaining two copies, I suggest keeping #9881 and closing this draft.

PR #9898, opened 2026-06-20
fix(treeshake): track separately exported pure functions

mturac:
### Summary

- preserve side-effect-free function metadata when a top-level function is declared first and exported later via `export { foo }`
- let cross-module optimization mark calls to Vue-style `defineComponent` exports as pure in dynamically imported chunks
- add a regression fixture with a Vue-shaped runtime and a dynamic icon chunk

### Background

Vue declares `defineComponent` with `@__NO_SIDE_EFFECTS__` and then exports it from an export list. The scanner already handled `export function foo()` and exported function-valued variables, but it missed this separated declaration/export shape. In the Vite reproduction from vitejs/vite#22725, that meant unused icon component factories in a dynamic chunk were kept as side-effectful bare calls.

Addresses vitejs/vite#22725.

### Tests

- `PATH="$PWD/node_modules/.bin:$PATH" just build-rolldown`
- `PATH="$PWD/node_modules/.bin:$PATH" pnpm exec vp run --filter rolldown-tests test:main fixtures-concurrent.test.ts -t dynamic-import-side-effects-barrel`
- direct Rolldown smoke with the Vite-style repro: dynamic icon chunk is 2.2 kB and no longer contains `Ri24HoursFill`
- `git diff --check`

### DIFF
--- a/crates/rolldown/src/ast_scanner/impl_visit.rs
+++ b/crates/rolldown/src/ast_scanner/impl_visit.rs
@@ -17,7 +17,7 @@ use rolldown_common::{
 };
 #[cfg(debug_assertions)]
 use rolldown_ecmascript::ToSourceString;
-use rolldown_ecmascript_utils::{ExpressionExt, is_top_level};
+use rolldown_ecmascript_utils::{ExpressionExt, FunctionExt, is_top_level};
 use rolldown_error::BuildDiagnostic;
 use rolldown_std_utils::OptionExt;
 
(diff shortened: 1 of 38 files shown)


mturac:
Closing this to keep my open PR queue focused on changes with a cleaner risk profile. This branch is broad and CodSpeed reported a potential performance regression, so I do not want to leave it open while it needs deeper performance work.

PR #9899, opened 2026-06-20
perf(rolldown): probe the name map once in ConflictResolver::resolve

Boshen:
## What

`ConflictResolver::resolve` is the per-symbol deconfliction primitive (shared by `Renamer` and the cross-chunk export pass) — called once for every top-level symbol in every chunk.

Its fast path (the common case where the bare name is free) hashed and walked the `used` map **twice** for the same key:

```rust
if !self.used.contains_key(&base) && accept(&base, true) {
  self.used.insert(base.clone(), 0);   // second hash + table walk of `base`
  return base;
}
```

This replaces the `contains_key` + `insert` pair with a single `entry()` probe. Semantics are identical:

- `Vacant` + accepted → the old fast path (write the slot we already located)
- `Vacant` + vetoed → leave `base` unrecorded, start the suffix search at 1
- `Occupied` → jump past it using its stored counter

The clone count is unchanged (the map must own the key either way); the win is **one fewer hash + table walk per symbol** on the hot path.

## Evidence (cargo-asm, release / fat-LTO)

Measured on the inlined caller `Renamer::add_symbol_in_root_scope`:

| | asm lines | `contains_key` | `insert` |
|---|---|---|---|
| before | 927 | 1 | 4 |
| after | 783 | 0 | 3 |

## Verification

- `ConflictResolver` unit tests (incl. the counter-jump / veto edge cases) pass
- `topics/deconflict` integration snapshots pass **byte-identically**
- clippy clean

Companion of #9900, which applies the same single-probe pattern to the `canonical_names` dedup in `add_symbol_in_root_scope`. The two are independent (disjoint files) and can land in either order.

### DIFF
--- a/crates/rolldown/src/utils/chunk/conflict_resolver.rs
+++ b/crates/rolldown/src/utils/chunk/conflict_resolver.rs
@@ -1,4 +1,5 @@
 use core::cmp::Reverse;
+use std::collections::hash_map::Entry;
 
 use oxc_str::CompactStr;
 use rolldown_common::{ModuleTable, SymbolRef, SymbolRefDb};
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY (approved):

Boshen:
### Merge activity

* **Jun 21, 1:37 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 21, 1:37 PM UTC**: `Boshen` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 21, 1:42 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9901, opened 2026-06-20
perf: cut two heap allocations from wrapped ESM init finalize

Boshen:
## What

`wrapped_esm_init_stmt_for_import_record` (scope-hoisting finalize path) performed two heap allocations per wrapped-ESM import record. This removes both, with no behavior change.

### 1. Per-record `FxIndexSet` → `Vec`

`collect_wrapped_esm_init_modules_for_import_record` returned an owned `FxIndexSet<ModuleIdx>` only to dedup owners within a single record. Its sole caller already filters every owner through the persistent `generated_init_esm_importee_ids` (`FxHashSet`) **before** building the init call, so the local set's dedup is fully subsumed by the global one — a duplicate is dropped at the global check and never re-emits. Switching to `Vec<ModuleIdx>` replaces the IndexMap insert/rehash with a plain push and drops one hashmap allocation per record.

### 2. Drop the `.collect::<Vec<_>>()` of init expressions

The expressions were collected into a `std::Vec` only to end the `&mut self` iterator borrow before reaching for `self.ast_factory`. `AstFactory` is `Copy`, so copying it out lets the iterator stay borrowed while nodes are still built through the copy. Driving the iterator by hand means the common 0- and 1-statement cases allocate nothing, and only the rare `(a(), b())` sequence case allocates — straight into the arena. Every branch consumes the iterator to exhaustion, preserving the global-dedup side effects exactly.

## Evidence

`cargo asm -p rolldown --lib wrapped_esm_init_stmt` (aarch64, release `lto = "fat"`):

| | before | after |
|---|---|---|
| instructions | 808 | **731** |
| calls (`bl`) | 40 | **27** |

Removed calls include `indexmap::IndexMap::insert_full`, `drop_in_place<IndexSet<ModuleIdx>>`, and the std `Vec`'s `__rust_alloc` / `RawVecInner::reserve` / `__rust_dealloc`. The remaining `hashbrown` calls are the global `generated_init_esm_importee_ids` dedup, which is intentionally kept.

## Verification

`cargo test -p rolldown` — 1756 passed, 0 failed; all snapshots unchanged (byte-identical output).

### DIFF
--- a/crates/rolldown/src/module_finalizers/mod.rs
+++ b/crates/rolldown/src/module_finalizers/mod.rs
@@ -161,10 +161,16 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
   fn collect_wrapped_esm_init_modules_for_import_record(
     &self,
     rec_idx: ImportRecordIdx,
-  ) -> FxIndexSet<ModuleIdx> {
+  ) -> Vec<ModuleIdx> {
     // See internal-docs/linking/reference-needed-symbols/implementation.md for why this follows
     // canonical owners through non-wrapped barrel modules.
-    let mut init_modules = FxIndexSet::default();
+    //
+    // Duplicate owners are tolerated here rather than deduped via a set: the sole caller filters
+    // each owner through `generated_init_esm_importee_ids` (a persistent `FxHashSet`) *before*
+    // building the init call, so a repeat owner is dropped at that check and never re-emits. A
+    // `Vec` push avoids the per-record hashing/allocation of an `FxIndexSet` that the global set
+    // already subsumes.
+    let mut init_modules = Vec::new();
     let rec = &self.ctx.module.import_records[rec_idx];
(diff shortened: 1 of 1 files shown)


hyfdev (approved):
Been thinking a to have infra to track every stage's mem/cpu opts

Boshen:
### Merge activity

* **Jun 20, 7:18 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 20, 7:18 AM UTC**: `Boshen` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 20, 7:24 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #5931, opened 2025-08-27
fix: handle errors thrown in `onLog` and `onwarn` options

sapphi-red:
Throwing an error in `onLog` or `onwarn` caused the process to exit.
See https://github.com/napi-rs/website/pull/352/files for the reason why this happens.

PR #9907, opened 2026-06-21
fix(plugin): propagate onLog hook errors to fail the build

jaehafe:
## What this PR solves

`PluginContext::{info,warn,debug}` are synchronous, but the user's `onLog` callback is async. To bridge that, each call spawns a detached task and discards the result with `let _ = on_log.call(...).await` — the FIXME left in #5931:

```rust
rolldown_utils::futures::spawn(async move {
    // FIXME: should collect error happened here and cause the build to fail later
    let _ = on_log.call(level, log).await;
});
```

As a result, an error thrown by `onLog`/`onwarn` was silently dropped when the log originated from a native (Rust) plugin via `ctx.info`/`warn`/`debug`, even though the same throw already fails the build on the core path (`module_task.rs` propagates it with `?`). The existing on-log/throw and onwarn/throw fixtures already assert that a throwing handler should surface as a build error, so this path was inconsistent.

## Fix

- Collect the spawned task handles in a shared `Vec<JoinHandle<_>>` on `PluginDriver` (`SharedLogHookTasks`), cloned into each plugin context.
- Drain and await them at the end of `Bundle::{write,generate,scan}` via `merge_log_hook_errors`, folding any error into the `BuildResult`.

## Alternatives explored

Making `info`/`warn`/`debug` async to await the error inline (fail at the exact log site). Rejected as API-breaking; the barrier keeps the sync signature.

## Notes for reviewers

- Errors surface at the end-of-build barrier, not the log call site (a consequence of keeping the methods sync).
- Builds without `onLog` are unaffected (the collector stays empty, the await is a no-op).
- **Open question:** failing applies to all levels incl. `info`/`debug`, matching the throw fixtures — confirm intended scope?

## Tests

- Added `on_log_error_fails_build`; fails on `main` (build succeeds), passes here.
- Existing `info_warn_debug` still passes and is now deterministic.


IWANABETHATGUY:
I prefer using a JS-side e2e test for this issue. I have added a test in https://github.com/rolldown/rolldown/issues/9922. Why? A JS-side test case could test the whole JS glue -> binding -> core pipeline, while a Rust-side test only tests core.

IWANABETHATGUY:
Following up on the JS-side e2e test I mentioned — here's the concrete fixture (from #9922), so this PR can assert the fix through the full JS glue → binding → core path rather than only at the Rust layer.

Drop it at `packages/rolldown/tests/fixtures/on-log/native-plugin-warn-throw/`. It mirrors the existing sibling `on-log/throw` fixture, but routes the warning through a **native** plugin (`esmExternalRequirePlugin`'s `ctx.warn` in `buildStart`) instead of core.

<details>
<summary><code>main.js</code></summary>

```js
const ext = require('ext');
console.log(ext);
```

</details>

<details>
<summary><code>_config.ts</code></summary>

```ts
import { defineTest } from 'rolldown-tests';
import { esmExternalRequirePlugin } from 'rolldown/plugins';
import { expect, vi } from 'vitest';

const fn = vi.fn();

export default defineTest({
  config: {
    external: ['ext'],
    plugins: [esmExternalRequirePlugin({ external: ['ext'] })],
    onLog(_level, log) {
      fn();
      expect(log.message).toContain('duplicate external');
      throw new Error('convert log to error');
    },
  },
  afterTest: () => {
    // The native plugin warning must actually have reached `onLog`...
    expect(fn).toHaveBeenCalled();
    // ...and the build should then have aborted. Reaching here means the error
    // thrown from `onLog` was swallowed (native_plugin_context.rs).
    throw new Error(
      'Expected the build to fail because the onLog handler threw, but it succeeded.',
    );
  },
  catchError(err: any) {
    expect(fn).toHaveBeenCalled();
    expect(err).toBeInstanceOf(Error);
    expect(err.message).toContain('convert log to error');
  },
});
```

</details>

Why this is red on `main` and green with this PR, given the harness flow in `fixture.test.ts`:

- On `main` the warning is swallowed → the build **succeeds** → `output` is truthy → `afterTest` runs and throws → ❌ test fails.
- With this PR the thrown `onLog` aborts the build → `compileFixture` rejects → `catchError` validates the error and returns → `output` is `undefined` so `afterTest` is skipped → ✅ test passes.

The `'duplicate external'` substring matches the actual warning emitted by `rolldown_plugin_esm_external_require` (`crates/rolldown_plugin_esm_external_require/src/lib.rs`), and `'ext'` being in both the top-level `external` and the plugin's `external` is what triggers it. Happy to keep the Rust test too — the JS one is just to lock in the end-to-end behavior.

jaehafe:
> Following up on the JS-side e2e test I mentioned — here's the concrete fixture (from #9922), so this PR can assert the fix through the full JS glue → binding → core path rather than only at the Rust layer.
> 
> Drop it at `packages/rolldown/tests/fixtures/on-log/native-plugin-warn-throw/`. It mirrors the existing sibling `on-log/throw` fixture, but routes the warning through a **native** plugin (`esmExternalRequirePlugin`'s `ctx.warn` in `buildStart`) instead of core.
> 
> `main.js`
> ```js
> const ext = require('ext');
> console.log(ext);
> ```
> 
> `_config.ts`
> ```ts
> import { defineTest } from 'rolldown-tests';
> import { esmExternalRequirePlugin } from 'rolldown/plugins';
> import { expect, vi } from 'vitest';
> 
> const fn = vi.fn();
> 
> export default defineTest({
>   config: {
>     external: ['ext'],
>     plugins: [esmExternalRequirePlugin({ external: ['ext'] })],
>     onLog(_level, log) {
>       fn();
>       expect(log.message).toContain('duplicate external');
>       throw new Error('convert log to error');
>     },
>   },
>   afterTest: () => {
>     // The native plugin warning must actually have reached `onLog`...
>     expect(fn).toHaveBeenCalled();
>     // ...and the build should then have aborted. Reaching here means the error
>     // thrown from `onLog` was swallowed (native_plugin_context.rs).
>     throw new Error(
>       'Expected the build to fail because the onLog handler threw, but it succeeded.',
>     );
>   },
>   catchError(err: any) {
>     expect(fn).toHaveBeenCalled();
>     expect(err).toBeInstanceOf(Error);
>     expect(err.message).toContain('convert log to error');
>   },
> });
> ```
> 
> Why this is red on `main` and green with this PR, given the harness flow in `fixture.test.ts`:
> 
> * On `main` the warning is swallowed → the build **succeeds** → `output` is truthy → `afterTest` runs and throws → ❌ test fails.
> * With this PR the thrown `onLog` aborts the build → `compileFixture` rejects → `catchError` validates the error and returns → `output` is `undefined` so `afterTest` is skipped → ✅ test passes.
> 
> The `'duplicate external'` substring matches the actual warning emitted by `rolldown_plugin_esm_external_require` (`crates/rolldown_plugin_esm_external_require/src/lib.rs`), and `'ext'` being in both the top-level `external` and the plugin's `external` is what triggers it. Happy to keep the Rust test too — the JS one is just to lock in the end-to-end behavior.

 Added the JS fixture, kept the Rust test too. Both green locally, thanks!

PR #9908, opened 2026-06-21
fix(cache): require() should return same module when called mutiple times

### LINKED ISSUE #9890: [Bug]: `require()` of a bundled ESM module returns a different object on each call

https://repl.rolldown.rs/#eNp1T8sSgjAM/JVML+iMA/c6/IIHz15KCYITW+wDZRj+3RQUT96yye5mdxKNkJNAE9yY33yajZA/fBCaobbGB1BQgsNH7BzusrxAf2dGtj9ezHqv/t4vpijgZGvkBWilW/Qbcy+TcVmy3Lc2Ug0VQnAR80V1tkS1fRrQ0TmORSP0rjPBQ6PIJ1J6bglzstdd9rGS2eHrmv5zDRQymc48Lbm2pitae+Krty7AWmdQFJErTXNymFk5MIVUQB/E/AYZ32Vb

### What is expected?

When bundled code `require()`s an ES module more than once, rolldown returns a different object on every call, violating Node.js CommonJS semantics where `require()` caches the module and returns the same object each time.

This is the rolldown equivalent of https://github.com/evanw/esbuild/issues/4440.

```
a === b: true
```
Node.js caches by resolved module and returns the same namespace-backed object every time.

```
a === b: false
```

Rolldown inlines `__toCommonJS(...)` at each call site, and `__toCommonJS` allocates a fresh wrapper every time:

```js
const a = (init_esm(), __toCommonJS(esm_exports));   // object #1
const b = (init_esm(), __toCommonJS(esm_exports));   // object #2  → a !== b
```

### System Info

```Shell
REPL 1.1.2
```

### Any additional comments?

I've talked with LLM if there's an easy way to solve this, but didn't seem to:

<details>
<summary>Approaches considered</summary>

#### What a fix has to satisfy

1. **Identity** — repeated `require()` of the same module returns the same object.
2. **Cross-chunk** — that identity holds even when caller and module land in different chunks.
3. **Cycle-safe** — a `require()` that fires *during the module's own initialization* (self-require or a require cycle) still gets a usable object, not `undefined`. Covered by `bundler_esm_cjs_tests/16-19`.
4. **Clean namespace** — the fix must not add observable properties to the module namespace object (the thing you get from `import * as`).
5. **`module.exports` export** — for `export { x as 'module.exports' }`, `__toCommonJS` returns that value *by reference at call time*, not via a live getter, so it can't be snapshotted before the body assigns it.
6. **No WeakMap** — `WeakMap` isn't guaranteed to be available in all target environments (old runtimes / constrained engines), so relying on it means either dropping those targets or carrying a `typeof WeakMap` fallback path. Avoiding it entirely keeps the output portable with no fallback branch.

#### The candidates

Starting from esbuild's fix, each WeakMap-free idea fails a different constraint:

- **esbuild's `__toCommonJSCached` ([#4441](https://github.com/evanw/esbuild/pull/4441))** — a `WeakMap` keyed on the namespace object; builds the wrapper on first call and caches it. Ships a `typeof WeakMap` fallback precisely because `WeakMap` may be unavailable.
  → Correct on everything **except (6)**: the WeakMap dependency (and its fallback path) is the bar to clear without one.

- **A — cache in a variable at the call site:** `(init(), x_cjs ??= __toCommonJS(x_exports))`.
  → Fails **(2)**: with code splitting each chunk declares its own `x_cjs`, so two chunks requiring the same module get two different objects.

- **A′ — promote that cache var to a real per-module binding, exported from the owning chunk.**
  → Correct, but the cache slot has to be *writable across chunks*, and rolldown's cross-chunk exports are getter-based (read-only). Needs new export-mechanism plumbing — the heaviest option.

- **B — stash the cached object on the namespace object** under a `Symbol`.
  → Fails **(4)**: the slot shows up in `Object.getOwnPropertySymbols(ns)` / `console.log(ns)`, since `ns` is the same object exposed via `import * as`.

- **C1 — a separate memoized thunk:** `var x_cjs = __esmMin(() => (init(), __toCommonJS(x_exports)))`.
  → Fails **(3)**: the memoized value isn't assigned until the thunk *returns*, so a re-entrant (cyclic) call gets `undefined`. Also needs a new cross-chunk symbol.

- **C2 — fold the conversion into the init closure's return:** `init = __esmMin(() => { …body…; return __toCommonJS(ns) })`, and lower each `require` to a bare `init()`.
  → Gets (1), (2), (4), but still fails **(3)** for the same reason as C1 (the return value doesn't exist mid-body) — prototyping this regressed `bundler_esm_cjs_tests/16-19` — and it entangles the change with TLA / concatenated closures.

#### A WeakMap-free candidate that clears all six: `__toCommonJSInit` eager-init wrapper

Wrap the init closure with a new runtime helper that computes and memoizes `__toCommonJS(exports)` **before** running the body:

```js
__toCommonJSInit = (exports, fn, res, err) => () => {
  if (err) throw err[0];
  try { return fn && (res = __toCommonJS(exports), fn(fn = 0)), res; }
  catch (e) { throw err = [e], e; }
};
// init_x = __toCommonJSInit(x_exports, () => { …body… });
// require site → init_x()
```

- **(1) Identity, (3) cycle-safe:** `res` is set before the body runs, so a re-entrant `require` during init already sees the finished object; every later call returns the same `res`.
- **(2) Cross-chunk:** reuses the existing `init_x` symbol — no new cross-chunk binding. Consumer chunks import *fewer* symbols than before (just `init_x`, no longer `x_exports` + `__toCommonJS`).
- **(4) Clean namespace:** the cache lives in the closure, never on `ns`.
- **(6):** no WeakMap, so no fallback branch and no target restrictions.

</details>

kb019:
fixes #9890

This PR fixes a caching issue where calling `require()` on the same bundled ESM module more than once returned different objects.
Most of the files are changes to the snapshot. Below is the list of files i changed

1. crates/rolldown/src/module_finalizers/impl_visit_mut.rs
2. crates/rolldown/src/module_finalizers/mod.rs
3. crates/rolldown/src/runtime/runtime-base.js
4. crates/rolldown/src/stages/link_stage/wrapping.rs
5. crates/rolldown_ecmascript_utils/src/ast_factory.rs
6. crates/rolldown_common/src/generated/runtime_helper.rs (generated by running the task generator)

Test files

All the tests for this change are present in the below directory 
**crates/rolldown/tests/rolldown/issues/cache_issue**

AI Usage:

I used ai only to undestand some of the code in runtime-base.js and to create the function __toCommonJSInit and __toCommonJSInitMin. The functions were refracted along the way when some tests were failing. **No ai** was used for rest of the files


hyfdev:
Thanks for your time but this is not a general solution to the problem.

We don't encourage PR first behavior. Raising is a fix is simple. The hard part is to ensure the fix is a long-term friendly solution.

If you are interested in solving this for all Rolldown users, make sure to reach out through the Discord or the linked issue for further discussions

https://github.com/rolldown/rolldown/issues/9890

kb019:
> Thanks for your time but this is not a general solution to the problem.
> 
> We don't encourage PR first behavior. Raising is a fix is simple. The hard part is to ensure the fix is a long-term friendly solution.
>
> If you are interested in solving this for all Rolldown users, make sure to reach out through the Discord or the linked issue for further discussions
> 
> #9890

Totally agree with you @hyf0 , Thanks for the feedback. I will start the discussion on the issue page. Thank you.

PR #9915, opened 2026-06-22
fix: git for windows treats symlink files as regular files

AliceLanniste:
# Description
when run `just test-node-rolldown` on windows,`import-glob/follow-symlinks` will throw an error

the error result:
屏幕截图 2026-06-22 094720

The reason for the bug is git for Windows defaults `core.symlinks` to `false`, which means symlinks tracked in the repository are checked out as plain text. so `import-glob/follow-symlinks`   cannot link `components/` directory.

# fix
I add `beforeTest` in  `import-glob/follow-symlinks/_config.ts`, it will  bind the soft-link file when it meets git for windows


IWANABETHATGUY:
Thanks for the PR! I’ll make a few small adjustments directly on this branch to better align it with the project style and reduce some back-and-forth.

IWANABETHATGUY (approved):

PR #9919, opened 2026-06-22
perf(common): pack TaggedSymbolRef into 8 bytes

IWANABETHATGUY:
### What

Replace the `TaggedSymbolRef` enum

```rust
enum { LinkOnly(SymbolRef), Normal(SymbolRef) } // 12 bytes
```

with a packed struct that stores the link-only tag in the **high bit of the `SymbolId`**, bringing it down to **8 bytes** (same size as `SymbolRef`).

### Why 12 bytes?

`SymbolRef` is `{ owner: ModuleIdx (u32), symbol: SymbolId (NonMaxU32) }` = 8 bytes. The `NonMaxU32` niche could host a discriminant for a *single* payload variant, but two payload-carrying variants force Rust to add a separate 4-byte-aligned discriminant → 12 bytes.

### Why the bit-steal is sound

A single module can never reach 2³¹ symbols: oxc caps a source file at <4GB (u32 spans), and the densest distinct-symbol binding (`a=>a=>…`, 3 bytes each) tops out well under 2³¹. `pack()` asserts this invariant, so a pathological input panics loudly instead of silently corrupting a symbol id.

### Impact

Shrinks every `StmtInfo::declared_symbols` element from 12 → 8 bytes, with **no growth of `StmtInfo` itself**. A `const` assertion pins `size_of::<TaggedSymbolRef>() == size_of::<SymbolRef>()`.

### Verification

- Unit tests: normal / link-only roundtrip, largest taggable `SymbolId`, panic-on-overflow, size invariant.
- `cargo clippy` + `cargo fmt`: clean.
- Snapshot tests: identical (behavior unchanged).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9920, opened 2026-06-22
perf(common): inline declared_symbols with SmallVec

IWANABETHATGUY:
> [!NOTE]
> Base PR #9919 (pack `TaggedSymbolRef` to 8 bytes) is now merged into `main`; this PR contains only the `Vec` → `SmallVec` change on top of it.

### What

Switch `StmtInfo::declared_symbols` storage from `Vec<TaggedSymbolRef>` to `SmallVec<[TaggedSymbolRef; 2]>`, keeping up to two declared symbols inline and only spilling to the heap beyond that.

### Why

The overwhelming majority of statements declare zero, one, or two top-level symbols, but `Vec` rounds its first allocation up to capacity 4 (an 8-byte element falls in the `MIN_NON_ZERO_CAP = 4` bucket) — a ~3.7× over-allocation that heap-allocates for nearly every statement.

Because `TaggedSymbolRef` is packed to 8 bytes (the base PR), `SmallVec<[_; 2]>` is **24 bytes — identical to `Vec`** — so `StmtInfo` does not grow. A `const` assertion pins `size_of::<DeclaredSymbols>() == size_of::<Vec<TaggedSymbolRef>>()`.

### Impact (threejs10x)

Measured two independent ways — both agree.

**1. Per-site — exact final heap state of every container.** Instrumented `LinkStage::link()` to walk every `StmtInfo.declared_symbols` after a real threejs10x link and read each one's `len()` / `spilled()` / `capacity()`.

| storage | heap allocations | live heap bytes |
|---|---|---|
| `Vec` | 24061 | 793,952 B (~775 KB) |
| `SmallVec<[_; 2]>` | 180 | 29,760 B (~29 KB) |

**~99% of these allocations elided.** Length histogram (`len: count`): `{0: 7920, 1: 23531, 2: 350, 3: 40, 4: 10, 5: 10, 6: 10, 9: 30, 10: 30, 14: 10, 17: 10, 20: 10, 27: 10, 68: 10}` — non-empty = 24061, `len ≥ 3` (what still spills with inline cap 2) = 180. (Large buckets are multiples of 10: threejs10x = 10× three.js.)

**2. Whole-program — real counting global allocator, A/B.** Wrapped the system allocator to count every `alloc`/`realloc` op, toggling *only* the inline capacity: `[_; 2]` (this PR) vs `[_; 0]` (no inline buffer → allocates like `Vec` for every non-empty container; the walk confirms `spilled = 24061` in that build).

| build | total alloc/realloc ops | total bytes requested |
|---|---|---|
| `SmallVec<[_; 2]>` (this PR) | ~1,003,900 | ~1316 MB |
| `SmallVec<[_; 0]>` (≈ `Vec`) | ~1,029,000 | ~1316 MB |
| **delta** | **≈ −24,600 … −25,000 ops** | **~0** |

Run-to-run noise was ±~150–270 ops, so the ~24k delta is ~100× the noise floor.

> **Measured vs derived.** The `24061` allocation count is empirical — both the per-site walk and the inline-cap-0 A/B build report it directly (`spilled = 24061`). The `~775 KB` is the one figure *derived* rather than read live: it applies `Vec`'s real `MIN_NON_ZERO_CAP = 4` rounding, whereas the `[_; 0]` A/B build spills exact-fit (~218 KB), so that build confirms the count but not the bytes.
>
> **What this does and doesn't buy.** The win is **allocation *count*** — ~24k near-per-statement `malloc`s on a hot path turned into inline writes — not memory footprint. Whole-program allocation *volume* is essentially unchanged (~1316 MB either way), and even this struct's live bytes (~746 KB saved) are tiny against that. So the absolute whole-program memory win is modest by design; the value is allocator/CPU pressure, bought with **zero growth of `StmtInfo`** (the `size_of` assertion).

### Verification

- `cargo clippy` (both crates, all targets) + `cargo fmt`: clean.
- Release `StmtInfo == 80` assertion: holds.
- Snapshot tests (`cargo test -p rolldown`, excluding hmr/test262): 1723 passed, 0 failed — zero behavior change.


hyfdev (approved):
Nice!

hyfdev:
### Merge activity

* **Jun 28, 9:24 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 28, 9:25 AM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 28, 9:52 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9923, opened 2026-06-22
perf: trim allocations & type size in sourcemap collapse, chain elements, and identifier names

Boshen:
Three small, self-contained allocation / type-size wins found while profiling a three.js bundle. Each is measured **deterministically** (allocation / byte / size counts, not wall-time), keeps behavior identical, and degrades gracefully.

### 1. `perf(sourcemap)`: reserve the token buffer in `collapse_sourcemaps`

The output `Box<[Token]>` was built with `last_map.get_source_view_tokens().filter_map(..).collect()`. The source iterator is slice-backed (exact `size_hint`), but `filter_map` resets the lower size-hint to `0`, so `collect` reserved nothing and grew the buffer through ~log2(n) reallocations. Nearly every token remaps successfully, so reserving the upper bound is safe.

Measured on `collapse_codegen_chain` (52,004 tokens, counting allocator):

- **15 → 0 reallocs**, 4.84 MB → 1.70 MB allocator traffic.

### 2. `perf(common)`: box the large `SourcemapChainElement` variants

The enum was **224 B** — `Transform`/`Load` inline a ~200 B `SourceMap`, while `Omitted`/`Null` are 12 B. A `Vec<SourcemapChainElement>` is stored per module, so every element paid for the largest variant, and being >128 B it `memcpy`s on every move / `Vec` reallocation. Boxing the `SourceMap` payloads (keeping the tuple shape so existing `Transform((plugin_idx, _))` patterns still match):

- **224 → 24 B (−89%)**, verified with `-Zprint-type-sizes`; guarded by a compile-time `size_of` assertion.

### 3. `perf(utils)`: avoid a wasted allocation in `legitimize_identifier_name`

The function allocated `String::with_capacity(name.len())` at the top, but the common case — an already-valid identifier — returns `Cow::Borrowed(name)` early and never touches it. Moved the allocation past the early return.

Measured with a dhat heap profile of a three.js bundle:

- `legitimize_identifier_name`: **1,334 → 256 allocations (−81%)**; whole-bundle 84,652 → 83,566.

---

Opened as a draft for review — happy to split into three separate PRs if you'd prefer to land them independently.

### DIFF
--- a/crates/rolldown/src/stages/scan_stage.rs
+++ b/crates/rolldown/src/stages/scan_stage.rs
@@ -223,7 +223,7 @@ impl<Fs: FileSystem + Clone + 'static> ScanStage<Fs> {
               map
                 .entry(module_idx)
                 .or_default()
-                .push(SourcemapChainElement::Transform((plugin_idx, generated_sourcemap)));
+                .push(SourcemapChainElement::Transform((plugin_idx, Box::new(generated_sourcemap))));
             }
             SourceMapGenMsg::Terminate => {
               break;
--- a/crates/rolldown/src/stages/scan_stage.rs
+++ b/crates/rolldown/src/stages/scan_stage.rs
@@ -220,10 +220,10 @@ impl<Fs: FileSystem + Clone + 'static> ScanStage<Fs> {
                 source: id.as_str().into(),
                 ..Default::default()
               });
-              map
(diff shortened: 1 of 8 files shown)


Boshen:
Split into three independent PRs: #9924, #9925, #9926.

PR #9924, opened 2026-06-22
perf(sourcemap): reserve token buffer capacity in collapse_sourcemaps

Boshen:
`collapse_sourcemaps` built its output `Box<[Token]>` via `last_map.get_source_view_tokens().filter_map(..).collect()`. The source iterator is slice-backed (exact `size_hint`), but `filter_map` resets the lower size-hint to `0`, so `collect` reserved nothing and grew the buffer through ~log2(n) reallocations. Nearly every token remaps successfully, so reserving the upper bound up front is safe.

Measured on `collapse_codegen_chain` (52,004 tokens, counting allocator):

- **15 → 0 reallocs**, 4.84 MB → 1.70 MB allocator traffic.

Degrades gracefully: if the iterator ever loses its exact `size_hint`, this reverts to the original grow-from-zero behaviour. All `rolldown_sourcemap` tests pass and the output token count is unchanged.

### DIFF
--- a/crates/rolldown_sourcemap/src/lib.rs
+++ b/crates/rolldown_sourcemap/src/lib.rs
@@ -72,29 +72,26 @@ pub fn collapse_sourcemaps(sourcemap_chain: &[&SourceMap]) -> SourceMap {
     .map(|sourcemap| (*sourcemap, sourcemap.generate_lookup_table()))
     .collect();
 
-  let tokens: Box<[Token]> = last_map
-    .get_source_view_tokens()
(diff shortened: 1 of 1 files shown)


Boshen:
Closing in favor of the upstream fix oxc-project/oxc-sourcemap#377 (return `impl ExactSizeIterator` from the slice-backed accessors). Once that lands in an oxc_sourcemap release and the dep is bumped, this can return as the clean `Vec::with_capacity(svt.len())` form instead of `size_hint().0`.

PR #9926, opened 2026-06-22
perf(utils): avoid wasted allocation in legitimize_identifier_name

Boshen:
`legitimize_identifier_name` allocated `String::with_capacity(name.len())` at the top of the function, but the common case — a name that is already a valid identifier — returns `Cow::Borrowed(name)` early and never touches that String. So every already-valid identifier allocated and immediately freed a String. Move the allocation past the early return so it only happens when the name actually needs legitimizing.

Measured with a dhat heap profile of a three.js bundle:

- `legitimize_identifier_name`: **1,334 → 256 allocations (−81%)**; whole-bundle 84,652 → 83,566.

No behaviour change (all 57 `rolldown_utils` tests pass); identical output, the String is just allocated later.

### DIFF
--- a/crates/rolldown_utils/src/ecmascript.rs
+++ b/crates/rolldown_utils/src/ecmascript.rs
@@ -54,7 +54,6 @@ pub fn legitimize_json_local_binding_name(
 }
 
 pub fn legitimize_identifier_name(name: &str) -> Cow<'_, str> {
-  let mut legitimized = String::with_capacity(name.len());
   let mut chars_indices = name.char_indices();
 
   let mut first_invalid_char_index = None;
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY (approved):

IWANABETHATGUY:
### Merge activity

* **Jun 22, 2:14 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 22, 2:14 PM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 22, 2:22 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #8284, opened 2026-02-12
feat: add INEFFECTIVE_DYNAMIC_IMPORT warning in core

shulaoda:
closes #7541, #7600, #7971

---

Thanks to @luke358 for the effort in PR #7600.
Thanks to @AliceLanniste for the effort in PR #7971.

Co-authored-by: luke358 <qq1494135711@gmail.com>
Co-authored-by: AliceLanniste <1399789151@qq.com>

PR #9927, opened 2026-06-22
refactor(rolldown_common): model ModuleId as a classified Path/Virtual/Bare enum

Boshen:
## Summary

`ModuleId` was a `struct { inner: ArcStr }` that every call site re-interpreted as a filesystem path through `SugarPath::as_path()` — treating virtual (`\0…`), bare (`react`), and URL ids as paths, and round-tripping bytes we already know are UTF-8 through `to_string_lossy`.

This remodels it as an enum classified **once** at construction:

```rust
pub struct ModuleId { repr: Repr }   // private enum Repr { Path | Virtual | Bare }
```

- **Gated path ops** — `as_path() -> Option<&Path>` (`Some` only for `Path`), plus `is_path()`, `is_virtual()`, `kind()`, `is_in_node_modules()`, `representative_name()`. Non-path ids are no longer silently path-parsed.
- **`ArcStr` kept** as the backing store — cheap clone, durable cross-build identity, `literal!` const for the runtime module. `ModuleIdx` (the `u32` per-build handle) stays the hot graph key, so `ModuleId` optimizes for being a faithful string/path identity rather than for size/hash speed.
- **Lossless** — every `to_string_lossy` on a `ModuleId` / `StableModuleId`-derived path becomes `to_str()` (module ids are always valid UTF-8). This is done in the shared `path_ext` helpers, so it also covers `ast_scanner` and preserve-modules naming.
- **`Hash` / `Eq` / `Ord` / `Borrow<str>`** are hand-implemented over `as_str()` (byte-exact, ignoring the discriminant) so `&str` lookups into `module_id_to_idx` / `user_defined_entry` keep working.
- **`StableModuleId::new`** branches on the cached `kind()` instead of re-running `is_absolute()` / `\0` detection.

## Why

A module id is sometimes a path and sometimes not. Treating it uniformly as a path was wasteful (redundant UTF-8 scans, `to_string_lossy` on known-UTF-8 data) and a latent correctness hazard. Classifying once makes the path / not-path distinction explicit and confines path semantics to where they belong.

## Validation

- `cargo test` — 1756 integration/snapshot tests pass with **zero snapshot/fixture churn**; behavior is identical.
- `clippy` (pedantic) and `fmt` clean.
- **Perf-neutral.** Benchmarked main vs branch (criterion, `bundle` group); a main-vs-main control showed the apparent ~3% delta was a first-run warm-up artifact correlated with case order, not the code — no regression, no real speedup, as expected for a modeling change whose per-module path logic is cost-equivalent.

### DIFF
--- a/crates/rolldown/src/ecmascript/ecma_module_view_factory.rs
+++ b/crates/rolldown/src/ecmascript/ecma_module_view_factory.rs
@@ -6,9 +6,7 @@ use rolldown_common::{
   side_effects::{DeterminedSideEffects, HookSideEffects},
 };
 use rolldown_error::BuildResult;
-use rolldown_std_utils::PathExt;
 use rolldown_utils::{ecmascript::legitimize_identifier_name, indexmap::FxIndexSet};
-use sugar_path::SugarPath;
 
 use crate::{
   ast_scanner::{AstScanner, ScanResult},
@@ -44,7 +42,7 @@ pub async fn create_ecma_view(
 
   let module_id = ctx.resolved_id.id.clone();
 
-  let repr_name = module_id.as_path().representative_file_name();
+  let repr_name = module_id.representative_name();
   let repr_name = legitimize_identifier_name(&repr_name);
 
   let scan_result = ast.program.with_mut(|fields| {
@@ -226,9 +224,13 @@ pub fn lazy_check_side_effects(
   if check_package_json_side_effects
     && let Some(side_effects) = resolved_id.package_json.as_ref().and_then(|p| {
       // the glob expr is based on parent path of package.json, which is package path
-      // so we should use the relative path of the module to package path
-      let module_path_relative_to_package =
-        resolved_id.id.as_path().relative(p.realpath().parent()?);
+      // so we should use the relative path of the module to package path.
+      // A plugin `resolveId` hook can return a non-absolute id (e.g. virtual `\0dep`)
+      // together with a `packageJsonPath`, so `package_json` is populated even when the
+      // id is not a `Path` kind. Treat the raw id as a path here (matching the historical
+      // behavior) so `sideEffects` is still honored for such plugin-resolved modules,
+      // rather than gating on `as_path()` and silently dropping the package metadata.
+      let module_path_relative_to_package = resolved_id.id.relative_path(p.realpath().parent()?);
       p.check_side_effects_for(&module_path_relative_to_package.to_string_lossy())
         .map(DeterminedSideEffects::UserDefined)
     })
--- a/crates/rolldown/src/module_loader/external_module_task.rs
+++ b/crates/rolldown/src/module_loader/external_module_task.rs
@@ -67,8 +67,8 @@ impl<Fs: FileSystem> ExternalModuleTask<Fs> {
       }),
     );
 
-    let need_renormalize_render_path = !matches!(resolved_id.external, ResolvedExternal::Absolute)
-      && Path::new(resolved_id.id.as_str()).is_absolute();
+    let need_renormalize_render_path =
+      !matches!(resolved_id.external, ResolvedExternal::Absolute) && resolved_id.id.is_path();
 
     let file_name: ArcStr = if need_renormalize_render_path {
       let entries_common_dir = commondir::common_dir(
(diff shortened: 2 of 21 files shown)


IWANABETHATGUY:
One behavioral subtlety worth confirming around the new `is_in_node_modules()`:

`crates/rolldown/src/stages/generate_stage/detect_ineffective_dynamic_imports.rs:29`
```rust
!importer_id.is_in_node_modules() && pre_rendered_chunk.module_ids.contains(importer_id)
```

`crates/rolldown_common/src/types/module_id.rs:135`
```rust
pub fn is_in_node_modules(&self) -> bool {
  self.as_path().is_some_and(rolldown_std_utils::PathExt::is_in_node_modules)
}
```

For a `Virtual`/`Bare` id, `as_path()` is `None`, so the string is never scanned for a `node_modules` component — whereas the old code path-scanned the raw string unconditionally. The behavior therefore diverges only when a non-path importer id literally contains a `node_modules` component (e.g. a relative `foo/node_modules/bar`): old → `true`, new → `false`, which flips the `!`.

Since `dynamic_importers` hold resolved module ids (absolute paths → `Path` kind, scanned identically; or virtual → no `node_modules`), a relative id sitting in `node_modules` doesn't look reachable here, so this seems benign. Just flagging it so the short-circuit is intentional.

Boshen:
@IWANABETHATGUY confirmed intentional — and good catch that the branch was untested.

`dynamic_importers` only ever hold resolved normal-module ids (`importer_path = module.id()`, set in `module_loader.rs`), which are either absolute filesystem paths (`Path` kind — scanned exactly as before) or virtual `\0…` ids (which never contain a `node_modules` path component). A relative id like `foo/node_modules/bar` isn't a resolved module id, so the diverging branch is unreachable in a real build. The path-gated check is also arguably more correct: it now means "a real path under `node_modules`", not "any string that happens to contain that substring".

On the coverage gap: the suppression branch has had no test since it was introduced in #8284 — the existing fixtures only exercise the positive-warning path with root-level importers, so nothing guarded the short-circuit either way. I've added `tests/rolldown/warnings/dynamic_import_from_node_modules_no_warning`: a module dynamically imported from within `node_modules` while also statically imported, with `expectWarning: false`. It passes with the guard and, with the guard removed, emits exactly the `INEFFECTIVE_DYNAMIC_IMPORT` warning — so the short-circuit is now locked in.

Boshen:
Addressed both Codex review threads in `bedba1345`:

**1. Stale module-id internal doc** (`internal-docs/module-id/implementation.md`) — updated the `### ModuleId` section to reflect the new classified `repr: Path | Virtual | Bare (ArcStr)` model, and clarified that equality/ordering/hashing remain raw string comparison over `as_str()` (so `&str` map lookups still work) while the classification only gates path logic.

**2. `sideEffects` for non-absolute plugin-resolved ids** (`ecma_module_view_factory.rs`) — good catch, this was a real behavioral regression. A plugin `resolveId` hook can return a virtual/bare id together with a `packageJsonPath`, so `resolved_id.package_json` is populated even when the id isn't a `Path` kind. The new `as_path()?` short-circuited the package `sideEffects` check for those ids. Restored the historical behavior by computing the relative path from the raw id string (`resolved_id.id.as_str().relative(...)` — exactly what the old unconditional `as_path()` did), so `sideEffects: false` is honored again for such plugin-resolved modules. The `package_json` (121) and `side_effect` (58) integration tests pass.

Boshen:
@codex review

IWANABETHATGUY (approved):

hyfdev (approved):

IWANABETHATGUY:
### Merge activity

* **Jun 24, 6:50 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 24, 6:50 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 24, 6:57 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9928, opened 2026-06-22
perf(utils): avoid allocation in default_sanitize_file_name for clean names

Boshen:
## Summary

`default_sanitize_file_name` (called once per output chunk/asset filename) unconditionally allocated `String::with_capacity(str.len())` and rebuilt the name char-by-char — even though the common case is a filename that contains **no** invalid characters (`index.js`, `react.production.min.js`, any path without shell/NTFS-unsafe chars) and needs no rewriting at all. This is the same wasted-allocation pattern recently fixed in `legitimize_identifier_name` (#9926).

The function now returns `Cow<str>`:
- **Clean path** (common): scan for the first invalid char; if none, return `Cow::Borrowed(str)` — zero allocation, zero copy.
- **Dirty path**: allocate only when a replacement is actually needed, and bulk-copy the valid prefix (drive letter included) with `push_str` instead of one `char` at a time.

The scan is done over **bytes, not chars**. Every invalid character is ASCII (≤ 0x7F), and UTF-8 guarantees that no byte of a multi-byte character is < 0x80, so a byte scan finds exactly the same positions as a char scan without per-char UTF-8 decoding, and every match lands on a char boundary. The dirty-path rewrite uses `u8::try_from(char)` rather than `char as u8` so a non-ASCII char (e.g. `😀`, whose low byte is `0x00`) is never truncated into a false match.

Both call sites in `rolldown_common` already do `.into()` into `ArcStr` (which implements `From<Cow<str>>`), so they are unchanged.

## Measured impact

Measured locally with a Criterion microbench (not committed), original `String` version vs this change:

| input | before | after | change |
|---|---|---|---|
| `clean_short` (`index.js`) | 21.4 ns | 6.7 ns | **−69%** |
| `clean_long` (78-char ASCII path) | 122 ns | 46 ns | **−63%** |
| `clean_unicode` (30-char Cyrillic path) | 85 ns | 32 ns | **−63%** |
| `dirty` (needs rewriting) | 60 ns | 49 ns | **−18%** |

All changes statistically significant (p < 0.05). The byte scan is what unlocks the gains on longer and non-ASCII paths (no per-char decoding).

## Correctness

Output is byte-for-byte identical to the previous implementation, including Windows drive-letter semantics (`C:/foo.js` preserved, later `:` still replaced: `C:/a:b.js` → `C:/a_b.js`). All `rolldown_utils` tests pass.

- `test_sanitize_file_name` — the borrowed/owned split, empty string, and the Windows-drive paths.
- `test_sanitize_unicode` — clean multi-byte names (2-, 3-, and 4-byte sequences: `café.js`, `日本語.js`, `компоненты/Кнопка.js`, `emoji_😀.js`) returned borrowed, and multi-byte chars surviving the rewrite path verbatim (`a?é` → `a_é`, `a?😀` → `a_😀`, `日本?語` → `日本_語`, `café:dir.js` → `café_dir.js`).

### DIFF
--- a/crates/rolldown_utils/src/sanitize_filename.rs
+++ b/crates/rolldown_utils/src/sanitize_filename.rs
@@ -1,54 +1,69 @@
-macro_rules! matches_invalid_chars {
-  ($chars:ident) => {
-    matches!($chars,
-      '\u{0000}'
-        ..='\u{001f}'
-          | '"'
-          | '#'
-          | '$'
-          | '%'
-          | '&'
-          | '*'
-          | '+'
-          | ','
-          | ':'
-          | ';'
-          | '<'
-          | '='
-          | '>'
-          | '?'
-          | '['
-          | ']'
-          | '^'
-          | '`'
-          | '{'
-          | '|'
-          | '}'
-          | '\u{007f}'
-    )
-  };
-}
+use std::borrow::Cow;
 
 // Follow from https://github.com/rollup/rollup/blob/master/src/utils/sanitizeFileName.ts
-pub fn default_sanitize_file_name(str: &str) -> String {
-  let mut sanitized = String::with_capacity(str.len());
-  let mut chars = str.chars();
+//
+// Every character that must be replaced is ASCII (<= 0x7F). UTF-8 guarantees that no byte of a
+// multi-byte character is < 0x80, so scanning bytes finds exactly the same positions as scanning
+// chars — without decoding — and every match lands on a char boundary.
+const fn is_invalid_byte(byte: u8) -> bool {
+  matches!(
+    byte,
+    0x00
+      ..=0x1f // C0 control characters
+      | b'"'
+      | b'#'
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY (approved):

Boshen:
### Merge activity

* **Jun 23, 8:51 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 23, 8:51 AM UTC**: `Boshen` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 23, 8:56 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9939, opened 2026-06-23
feat: add self-sufficient, scoped pure top-level call treeshaking

fubhy:
## Summary

Adds `treeshake.pureTopLevelCalls`, an opt-in DCE escape hatch for trusted modules whose top-level call/new initializers are intentionally side-effect-free.

No linked issue.

## Details

- Supports boolean, string/RegExp patterns, ordered rules, and callbacks.
- Resolves the option per module before scanning, then stores the result in `FlatOptions`.
- Documents the option in the treeshake reference and DCE guide.
- Keeps the behavior scoped because it can remove real side effects.

## Alternatives

Considered emitting source-level PURE annotations, but implementing this directly in side-effect analysis avoids rewriting source and keeps the behavior explicit in Rolldown config.

## Review Notes

The main review focus should be the exact side-effect semantics and the rule/callback option shape.

## Validation

- `cargo +1.95.0 fmt --check`
- `cargo test -p rolldown_common pure_top_level_calls --lib`
- `cargo test -p rolldown ast_scanner::side_effect_detector::test::test_pure_top_level_calls`
- `cargo check -p rolldown_common -p rolldown -p rolldown_binding`
- `pnpm exec tsc --noEmit`

## Checklist

- Read the contribution guide.
- Searched open PRs; #9898 is related but not a duplicate.
- Updated docs.
- Included focused tests.
- AI assistance was used; changes were reviewed and validated locally.

### DIFF
--- a/crates/rolldown/src/ast_scanner/side_effect_detector/mod.rs
+++ b/crates/rolldown/src/ast_scanner/side_effect_detector/mod.rs
@@ -25,6 +25,26 @@ pub struct SideEffectDetector<'a> {
   namespace_object_symbol_ids: Option<&'a FxHashSet<SymbolId>>,
 }
 
+#[derive(Clone, Copy, Default)]
+struct ExprSideEffectContext {
+  assignment_context: bool,
+  in_callee: bool,
+}
+
+impl ExprSideEffectContext {
+  fn assignment_context(self) -> Self {
+    Self { assignment_context: true, ..self }
+  }
+
+  fn callee(self) -> Self {
+    Self { in_callee: true, ..self }
+  }
+
+  fn argument(self) -> Self {
+    Self { in_callee: false, ..self }
+  }
+}
+
 impl<'a> SideEffectDetector<'a> {
   pub fn new(
     scope: &'a AstScopes,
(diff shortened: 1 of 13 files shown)


fubhy:
This is mainly aimed at leaf/application bundling.

For libraries, publishing /`*#__PURE__*/` annotations makes sense because the purity information travels with the package and benefits every downstream bundler.

For applications, that can be unnecessary indirection. The app build is the final consumer, so Rolldown can apply the heuristic directly during side-effect analysis without another JS-side parse/transform pass.

```ts
treeshake: {
  pureTopLevelCalls: [{ test: /\/src\//, pure: true }]
}
```

It also lets app authors retroactively apply this to dependencies that did not publish sufficient PURE annotations, without patching or transforming those packages first:

```ts
treeshake: {
  pureTopLevelCalls: [{ test: /node_modules\/some-lib\//, pure: true }]
}
```

So the annotation/plugin approach remains useful for library publishing, while this option gives application builds a faster, more ergonomic, and retroactive way to express the same assumption.

IWANABETHATGUY:
Did you start a discussion before opening a PR? I can't find it.

fubhy:
> Did you start a discussion before opening a PR? I can't find it.

No. This was preceded by a ~200 comment long thread in the Effect Discord ;-)

IWANABETHATGUY:
Ok, let's file an issue and discuss with the rolldown member before the implementation.

IWANABETHATGUY:
I am going to close this PR because, in the AI era, code is the easy part, but the decision is not. Let's reopen it if we decide to adopt the feature.

fubhy:
Ok. Let's also update the contribution guidelines then. Because I just spent multiple hours reasonign through this and discussing with the Effect community (admittedly, not your community) to come up with this PR. I also read the contribution guideliens myself and they do not state anywhere that a discussion is required before opening a PR (the opposite actually).

Issue #9946, opened 2026-06-23
[Bug]: bundledDev emits a cyclic ESM import used at module-init as unbound (`X is not defined`)


ShMcK:
### Reproduction link or steps

Minimal, self-contained (pure ESM, no CJS/barrel/dynamic-import). 4 files:

**vite.config.js**
```js
export default { experimental: { bundledDev: true } };
```
**index.html**
```html
<!doctype html><html><body><div id="app"></div>
<script type="module" src="/src/main.js"></script></body></html>
```
**src/main.js**
```js
import { defaults } from './a.js';
document.getElementById('app').textContent = 'button=' + defaults.button.name;
```
**src/a.js**
```js
import { B } from './b.js';
// top-level (module-init) use of the cyclic import — this is what breaks
export const defaults = { button: B };
```
**src/b.js**
```js
import { defaults } from './a.js'; // cycle back to a.js
export class B {}
export const usesA = () => defaults;  // used only inside a fn -> bound fine
```

Run `vite` (dev), open the page.

### What is expected?

Page renders `button=B`. `B` resolves to the class exported by `b.js`, as it does
in classic Vite dev (rollup+esbuild) and in `vite build`.

### What is actually happening?

`ReferenceError: B is not defined`.

The served entry chunk `/assets/index.js` emits:

```js
var defaults = { button: B }   // B referenced...
```

…but `class B` is **not declared anywhere in that chunk** (it lives in `b.js`'s
separate lazy chunk). So the value-position reference to the cyclic import is
emitted **unbound**.

Key detail: only imports **used during module initialization (top-level)** break —
`export const x = { button: B }`, `class X extends B`, `const y = makeThing(B)`, etc.
Sibling imports from the *same* module that are referenced only **inside function
bodies** are bound correctly. The unbound one happens to participate in an import
**cycle**.

This is `experimental.bundledDev` only; `experimental.devMode.lazy: false` does not
change it. Classic dev avoids it (esbuild optimizeDeps flattens dep cycles + native
per-module ESM live bindings).

### Real-world impact

Surfaced trying to enable `bundledDev` on a large app. The same unbound-cyclic-import
pattern broke, in cascade, `@sentry/core` (`class ServerRuntimeClient extends BaseClient`
→ `BaseClient is not defined`; also `defineIntegration`), `@datadog/browser-core`
(`ONE_SECOND`), and first-party code (`export const defaultProps = { buttonClass: SelectButton }`
[...]

Issue #9485, opened 2026-05-20
[Bug]: Circular import generated in vite dev mode


rob-clarke:
### Reproduction link or steps

https://github.com/rob-clarke/rolldown-issue

- Run `npm run dev` (start `vite` dev server)
- Open site in browser
- Observe failure to import `roslib` due to circular import in generated module

### What is expected?

Running `vite dev` results in a functional JS module capable of being loaded by the browser.

### What is actually happening?

The bundled module has a circular import that results in a failure to load the module. Reported error is:
`undefined is not a function` when attempting to call `__CommonJSMin`

### System Info

```Shell
Running in WSL. Browser is Firefox 150.0.3
  System:
    OS: Linux 6.6 Ubuntu 24.04.4 LTS 24.04.4 LTS (Noble Numbat)
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11850H @ 2.50GHz
    Memory: 1.51 GB / 5.79 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 24.15.0 - /home/rc13011/.nvm/versions/node/v24.15.0/bin/node
    npm: 11.12.1 - /home/rc13011/.nvm/versions/node/v24.15.0/bin/npm
```

### Any additional comments?

Bug encountered when attempting to use `roslib` with latest `vite`. `roslib` has a dependency on `eventemitter3` which has a ESM interface that exports an imports from a CJS file. This appears to be a trigger of this issue.

Using `"overrides": {"rolldown": "1.0.0"}` does not show the same behaviour.

Appears related to #9401 but as I understand it this repro does not use dynamic import or require (unless `vite dev` injects a dynamic import).

IWANABETHATGUY:
it should be fixed by https://github.com/rolldown/rolldown/pull/9419

IWANABETHATGUY:
I can confirm this is already fixed in the latest rolldown

PR #9419, opened 2026-05-16
refactor: always split runtime module first

IWANABETHATGUY:
## Summary

Reworks runtime-module chunk placement (see `meta/design/code-splitting.md` →
"Runtime Module Placement"): the runtime module is always extracted into its own
chunk first, then merged back into a proven-safe host chunk (the sole runtime
consumer / a chunk with the same reachability bitset / a consumer-set dominator),
guarded by TLA- and static-cycle checks. The initial layout is cycle-safe by
construction instead of place-by-bitset-then-peel-on-cycle.

A follow-up commit fixes a regression this surfaced: single-file `output.file`
(es/cjs) builds that pull in a runtime helper (e.g. a reified namespace) were
wrongly rejected with `[INVALID_OPTION] InvalidOutputFile`. The multi-chunk guard
in `generate()` counted the merged-away runtime chunk's tombstone (still in
`chunk_table`, but skipped at render time). It now counts only live chunks,
mirroring the render-time filter.

## Why some Rollup alignment tests are skipped

Three `chunking-form` samples are added to
`packages/rollup-tests/src/failed-tests.json`. **These are not workarounds for a
bug** — Rolldown's output is behaviorally correct; it just emits one fewer chunk
than Rollup.

The `chunking-form` suite only asserts that Rolldown emits the **same number of
chunks** as Rollup (`assert.strictEqual(actualChunkCount, expectedChunkCount)`);
it does not compare chunk contents. With the runtime now folded into a
same-bitset host chunk, these multi-entry topologies no longer need a standalone
runtime chunk, so Rolldown emits one fewer chunk than Rollup: Rollup hoists shared
content into a `generated-*.js` chunk, while Rolldown folds that content (plus the
runtime) into an existing entry chunk.

Each was verified by **executing** both Rolldown's `_actual/` and Rollup's
`_expected/` output and comparing observable behavior — entry exports, console
output, and computed values all match. The only differences are the chunk count
and cosmetic representation (Rolldown's namespace objects carry the
spec-compliant `Symbol.toStringTag: "Module"`, which Rollup's frozen-object
namespaces omit).

Affected samples (both `generates es` and `generates cjs`):

- `entry-point-without-own-code` — Handles entry points that contain no own code except imports and exports
- `namespace-object-import` — namespace object import
- `namespace-reexports` — namespace rendering with reexports

To re-check one, run the sample and diff its `_actual/` against `_expected/`, e.g.
[...]

PR #9947, opened 2026-06-23
test(dev): full-bundle-mode regression test for cyclic import binding (#9946)

ShMcK:
### What

Adds a minimal full-bundle-mode (`experimental.devMode`) browser playground that
exercises a cyclic import whose binding is used at **module init**.

Files: `a.js -> b.js -> a.js`. `b.js` exports `class B`; `a.js` uses it at the top
level:

```js
// a.js
import { B } from './b.js';
export const defaults = { button: B };   // module-init use of a cyclic import
```

`main.js` renders `button=B`. The spec asserts `.app` reads `button=B`.

### Why

Regression test for #9946. In full-bundle-mode the served chunk references `B`
but never declares/imports it, so the page throws `ReferenceError: B is not
defined` at load and `.app` stays empty. Sibling imports used only inside
function bodies bind correctly — only the module-init reference over a cycle
breaks. The same pattern broke `@sentry/core` (`class ServerRuntimeClient
extends BaseClient`), `@datadog/browser-core` (`ONE_SECOND`), and first-party
app code when enabling `bundledDev`.

Reproduces on the published rolldown **1.1.2**. CI here will confirm against
`main` — if it already passes, the test stands as a regression guard; if it
fails, it's a live repro for #9946.

Related: #9691 (lazy mode drops an import), #9485 (the CJS-interop circular
variant, fixed by #9419 — this pure-ESM module-init case is distinct and still
reproduces on 1.1.2).

### DIFF
--- a/packages/test-dev-server/tests/playground/circular-import-binding/__tests__/circular-import-binding.spec.ts
+++ b/packages/test-dev-server/tests/playground/circular-import-binding/__tests__/circular-import-binding.spec.ts
@@ -1,0 +1,19 @@
+import { describe, expect, test } from 'vitest';
+import { page } from '~utils';
+
+// Regression test for https://github.com/rolldown/rolldown/issues/9946.
+//
+// In full-bundle-mode (`experimental.devMode`), an imported binding used at
+// MODULE INIT (top-level) inside an import cycle is emitted unbound. The served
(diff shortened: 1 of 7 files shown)


h-a-n-a:
I'm drafting this as this is not fixed and ready to be reviewed yet.

h-a-n-a (approved):

PR #9948, opened 2026-06-23
fix: deterministically keep the shortest name for deduplicated assets

### LINKED ISSUE #9940: [Bug]: Non-deterministic output: deduplicated assets take their filename from emission order

- StackBlitz (`node build.mjs` in the terminal there will run a build 10 times and will report in case the output differs): https://stackblitz.com/github/mitio/rolldown-asset-dedup-nondeterminism-repro
- GitHub repo: https://github.com/mitio/rolldown-asset-dedup-nondeterminism-repro

Turns out rolldown's build is non-deterministic when you have identical content under different names in your bundle tree. The minimal reproduction is just 2 identical files, each imported by its own module and then one final entry file:

```
src/assets/alpha.png   # identical content with beta.png
src/assets/beta.png    # identical content with alpha.png
src/mod_a.js           # import u from './assets/alpha.png'; export default u
src/mod_b.js           # import u from './assets/beta.png';  export default u
src/entry.js           # import './mod_a.js'; import './mod_b.js'; console.log(...)
```

### What is expected?

The build process should produce the same output every time when ran with the same input content.

When deduplicating assets, the file name chosen to be used should be deterministically-determined.

Here's how I believe it happens:

When two emitted assets have identical content but different names (e.g. the same logo under two filenames, imported as asset modules), rolldown deduplicates them by content hash and the surviving asset's output filename is taken from whichever copy emits **first**. Assets emit concurrently from parallel module processing and which one is emitted first is thus non-deterministic, especially on systems with multiple processors. The one name of the deduplicated asset chosen by this process then lands in `[name]` of `assetFileNames`, so it changes the bytes — and therefore the `[hash]` — of every chunk that references that duplicate asset. That's how byte-identical input produces different asset and chunk hashes per build.

### System Info

`rolldown 1.1.2`

This happens on macOS (Apple Silicon) and Ubuntu Linux (arm64, multi-core) as well. Don't think this would be relevant but let me know if it is and I'll add more info.

The system info below is from StackBlitz:

```Shell
  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 22.22.3 - /usr/local/bin/node
  npmPackages:
    rolldown: 1.1.2 => 1.1.2
```

### Any additional comments?

We were hit quite bad by this at Dext (other factors were at play as well, of course, but this non-determinism was one of the required conditions...) It would be great if we manage to resolve it.

Full disclosure: the reproduction was created with the help of Claude Code. I did try to simplify it as much as possible and massaged its output. Looks absolutely legit to me but I might have missed something.

x1024:
When two emitted assets have identical content but different names, rolldown deduplicates them by content hash. The surviving asset's [name] was taken from whichever copy emits first, which is non-deterministic.

Fix by always choosing the alphabetically-first name.

Fixes rolldown/rolldown#9940

Claude was used, the work was reviewed by me. I mostly have Rust experience from advent of code.

---

### The problem

When several assets are emitted with identical content but different names, rolldown deduplicates them by content hash into one output asset. The surviving asset took its `[name]` from whichever copy was emitted first, and because assets are emitted concurrently during module processing that order is not stable. The chosen name flows into `assetFileNames` and changes the `[hash]` of every chunk that references the asset, so byte-identical input could produce different file names from one build to the next. Fixes #9940.

### Deterministic survivor name

The survivor is now chosen deterministically: the shortest name wins, with ties broken lexicographically. This is the rule Rollup uses in `finalizeAssetsWithSameSource`, and the same order rolldown already applies when it sorts an asset's `names`, so the chosen `filename` stays consistent with `names[0]`. An earlier version of this PR kept the alphabetically-first name, which ignores length and disagrees with Rollup; it has been changed to shortest-then-lexicographic.

### Fixing a metadata-loss race

The first time a hash is seen, the asset is now inserted into `self.files` while the `source_hash_to_reference_id` shard lock is still held, before the reference id becomes visible. Previously the lock was dropped before that insert, so a concurrent duplicate could look up the reference id, find nothing in `self.files` yet, and silently lose its `name` and `originalFileName`.

### Behavior change (Rollup parity)

This restores Rollup's behavior. For example, two identical CSS assets named `style` and `style2` now deduplicate to `style-[hash].css`, the shorter name, exactly as Rollup does. Vite's `css-codesplit-consistent` test was flipped to expect `style2` when vite started building with rolldown, precisely because rolldown diverged here; with this fix it can go back to its original Rollup assertion.

### Tests

Tests live in `crates/rolldown/tests/rolldown/function/asset_dedup_filename` and cover the shortest name winning regardless of emission order, the shortest name beating the lexicographically-first one, every duplicate's `name` and `originalFileName` being collected and sorted onto the survivor, explicit `fileName` assets being left untouched, and concurrent emission never dropping a name. Each test fails on the current code and passes with the fix.

### DIFF
--- a/crates/rolldown/tests/rolldown/function/asset_dedup_filename/entry.js
+++ b/crates/rolldown/tests/rolldown/function/asset_dedup_filename/entry.js
@@ -1,0 +1,1 @@
+console.log('entry');
--- a/crates/rolldown/tests/rolldown/function/asset_dedup_filename/mod.rs
+++ b/crates/rolldown/tests/rolldown/function/asset_dedup_filename/mod.rs
@@ -1,0 +1,114 @@
+use std::{borrow::Cow, sync::Arc};
+
+use rolldown::{AssetFilenamesOutputOption, Bundler, BundlerOptions, InputItem};
+use rolldown_common::{EmittedAsset, Output};
+use rolldown_plugin::{HookUsage, Plugin, PluginContext};
+
+/// All emitted assets share this exact `source`, so the file emitter
+/// deduplicates them into a single output asset whose surviving file name must
+/// depend only on the names, not the (parallel, non-deterministic) emission order.
+const SHARED_SOURCE: &str = "shared-asset-bytes-used-for-deduplication";
+
+/// Emits one asset per `name` (in order), all with [`SHARED_SOURCE`] and no
+/// explicit `file_name`.
+#[derive(Debug)]
+struct EmitDuplicatesPlugin {
+  names: Vec<&'static str>,
+}
+
+impl Plugin for EmitDuplicatesPlugin {
+  fn name(&self) -> Cow<'static, str> {
+    "emit-duplicates".into()
+  }
+
+  fn register_hook_usage(&self) -> HookUsage {
+    HookUsage::BuildStart
+  }
+
+  async fn build_start(
+    &self,
+    ctx: &PluginContext,
+    _args: &rolldown_plugin::HookBuildStartArgs<'_>,
+  ) -> rolldown_plugin::HookNoopReturn {
+    for name in &self.names {
+      ctx.emit_file(
+        EmittedAsset {
+          file_name: None,
+          original_file_name: None,
+          name: Some((*name).to_string()),
+          source: SHARED_SOURCE.to_string().into(),
+        },
+        None,
+        None,
+      )?;
+    }
+    Ok(())
+  }
+}
+
+/// Bundles a dummy entry while emitting identical-content assets named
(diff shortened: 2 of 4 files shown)


x1024:
The test is failing because vite is testing for consistent code-splitting (how? This must be non-deterministic), but asserting that only the file with the alphabetically-last name gets kept. The vite tests needs to be updated to assert the other way around and we'll be OK.

shulaoda (approved):
Thanks! LGTM!

PR #9960, opened 2026-06-24
fix(treeshake): apply @__NO_SIDE_EFFECTS__ to cross-chunk namespace calls

### LINKED ISSUE #9956: [Bug]: NO_SIDE_EFFECTS annotation does not work for namespace import calls across chunk boundaries

`@__NO_SIDE_EFFECTS__` works for namespace imports when Rolldown can inline/collapse the namespace in a single chunk, but fails when the same namespace-member call crosses a chunk boundary.

### What is expected?

`@__NO_SIDE_EFFECTS__` should apply consistently to calls of the annotated exported binding, even when the call is made through a namespace import and the imported module is emitted as a shared chunk.

In split/shared output, the calls remain.

https://repl.rolldown.rs/#eNqlU9tugkAQ/ZUJL6uGQnp5omli4iXpS/ugjyQEYdBtcNYsi7Ux/HsHUC5VmyZ94LLnzM6eOTN7tBLLO1papWmsPsmJFCVy7ZisRMnyrjG2FTEhtzulDRwhxkQSTioaCki02oI4bxM++YSHKpQDwzytvs2GwdEnAEm73HhQ/QMIJKO/7kLhgXDc04IPFnaPX3X51SUfdfmo4Qt+F0OfrMK2sk2oMe5U2wJ1ke4IxkHw9h4sXqezYDafzybLRRDAyG2qSnKKjFQECU0xSgf7MM1xWNfCnmUqRSdV64GoechkjIBJgpERNtTRz2WwRpNrqhEGio5zZR4+iWaHnYYXuK2qVfObjirNn3Xww26g5RmdI5vWNuTsWgfpzcYIwgzmSp1mwnFP9n5kgpP6VJcVckUc5Jz8uy9V1MyqYUrFg4eSKdvWtryvoEL+r+DxpoKnvoJyqPoKKuTH/aiz2uf2nW/IhRbXheUGMwRGAVO5lRQajLsir1vUt4fV7fn4tTTjbSgZ+AbM20eC

### System Info

```Shell
System:
    OS: macOS 26.5.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 156.33 MB / 32.00 GB
    Shell: 4.6.0 - /etc/profiles/per-user/fubhy/bin/fish
  Binaries:
    Node: 24.15.0 - /nix/store/3mvbmkd7lmxi3xlka1avra24flwbgjv9-nodejs-24.15.0/bin/node
    npm: 11.12.1 - /nix/store/3mvbmkd7lmxi3xlka1avra24flwbgjv9-nodejs-24.15.0/bin/npm
    pnpm: 11.4.0 - /nix/store/l3awryym5fh66w3ilk9f1gcxkq06fi2r-corepack-0.35.0/bin/pnpm
    bun: 1.3.13 - /nix/store/jgl1hjr0cf6s32jcnqmkp44av4flamap-bun-1.3.13/bin/bun
    Deno: 2.8.3 - /nix/store/a9avpz7i9y2yjhmdimq719ih6wbb4s0p-deno-2.8.3/bin/deno
  npmPackages:
    rolldown: workspace:* => 1.1.2
```

### Any additional comments?

`cross_module_optimization.rs` only checks identifier callees for `SideEffectsFreeFunction`, so calls that remain as namespace-member imports across chunks are missed.

IWANABETHATGUY:
## Summary

`@__NO_SIDE_EFFECTS__`-annotated functions called through a namespace import (`ns.fn()`) were **not** tree-shaken when the annotated module landed in a separate/shared chunk, even though the same calls via named imports (`fn()`) and single-chunk namespace calls were eliminated correctly.

Fixes #9956.

## Root cause

There are two independent DCE mechanisms for `@__NO_SIDE_EFFECTS__`:

1. **Link-stage `cross_module_optimization`** — runs before chunking, resolves symbols cross-module, marks side-effect-free call statements for tree-shaking. But it only inspected `it.callee.as_identifier()`, so namespace-member callees (`ns.fn`, a `StaticMemberExpression`) were never recognized.
2. **Per-chunk `dce-only` minifier** — reads the annotation directly, but only works when the annotated declaration and the call land in the *same* chunk.

So a namespace call across a chunk boundary fell through both: mechanism (1) skipped the member-expression callee, and mechanism (2) couldn't see the annotation living in the other chunk.

## Fix

Two coordinated changes in `cross_module_optimization.rs`:

1. New `resolve_callee_canonical_symbol()` unifies identifier callees with member-expression callees. For the latter it resolves `ns.fn` via the module's `resolved_member_expr_refs`, guarded by `prop_and_related_span_list.is_empty()` so `ns.fn.bar()` isn't misattributed to `fn`.
2. The `SideEffectDetector` re-run previously passed `None` for `namespace_object_symbol_ids`. Even after marking the call pure, the detector recurses into the `ns.fn` callee *read*, which looks impure unless it knows `ns` is a namespace object. The set is reconstructed per-module from `named_imports` (the `Specifier::Star` entries — mirroring the scanner's `add_star_import`, which isn't persisted).

Both halves are required: marking the call pure alone leaves it kept because the callee read still reports a side effect.

## Tests

- `no_side_effects_namespace_cross_chunk` — minimal regression: two entries force `shared.js` into a shared chunk; both `ns.fn()` calls (and the chunk) are now eliminated.
- `no_side_effects_namespace_cross_chunk_cases` — boundary matrix: computed access (`ns['fn']()`) and the function-expression form are eliminated; results that are used and `ns.fn.call(...)` (trailing property access) are preserved.

Both fixtures were verified to reproduce the bug on `main` (calls remain) and pass with the fix. Full Rust suite green: `cargo test -p rolldown` (1760 integration + 69 unit), clippy `--deny warnings`, `fmt --check`, and `cargo check --all-features` all clean.

### DIFF
--- a/crates/rolldown/src/stages/link_stage/cross_module_optimization.rs
+++ b/crates/rolldown/src/stages/link_stage/cross_module_optimization.rs
@@ -3,16 +3,16 @@ use oxc::{
     AstBuilder, AstKind,
     ast::{
       BindingIdentifier, BindingPattern, Declaration, ExportDefaultDeclaration,
-      ExportDefaultDeclarationKind, ExportNamedDeclaration,
+      ExportDefaultDeclarationKind, ExportNamedDeclaration, Expression,
     },
   },
   ast_visit::{Visit, walk},
-  semantic::NodeId,
+  semantic::{NodeId, SymbolId},
 };
 use rolldown_common::{
-  AstScopes, ConstExportMeta, EcmaViewMeta, FlatOptions, GetLocalDb, IndexModules, ModuleIdx,
-  SharedNormalizedBundlerOptions, SideEffectDetail, StmtInfoIdx, SymbolRef, SymbolRefDb,
-  SymbolRefFlags,
+  AstScopes, ConstExportMeta, EcmaViewMeta, FlatOptions, GetLocalDb, IndexModules,
+  MemberExprRefResolutionMap, ModuleIdx, SharedNormalizedBundlerOptions, SideEffectDetail,
+  Specifier, StmtInfoIdx, SymbolRef, SymbolRefDb, SymbolRefFlags,
 };
 use rolldown_ecmascript_utils::ExpressionExt;
 use rolldown_utils::rayon::{IntoParallelRefIterator, ParallelIterator};
@@ -193,6 +193,18 @@ impl LinkStage<'_> {
             module_idx_and_stmt_idx_to_dynamic_import_expr_node_id_map
               .get(&module_idx)
               .unwrap_or(&default_stmt_idx_to_dynamic_import_expr_node_ids);
+          // Local symbols of `import * as ns` bindings. Property reads on ES module namespace
+          // objects are side-effect-free, so the `SideEffectDetector` re-run below must know
+          // about them to drop `ns.fn()` calls to `@__NO_SIDE_EFFECTS__` functions. The scanner
+          // builds the same set (see `add_star_import`), but it isn't persisted, so reconstruct
+          // it here from the persisted named imports.
+          let namespace_object_symbol_ids: FxHashSet<SymbolId> = module
+            .named_imports
+            .iter()
+            .filter_map(|(local_ref, named_import)| {
+              matches!(named_import.imported, Specifier::Star).then_some(local_ref.symbol)
+            })
+            .collect();
           let mut ctx = CrossModuleOptimizationRunnerContext {
             local_constant_symbol_map: FxHashMap::default(),
             side_effect_detail_mutations: FxHashMap::default(),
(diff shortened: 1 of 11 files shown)


hyfdev (approved):

IWANABETHATGUY:
### Merge activity

* **Jun 24, 3:50 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 24, 3:50 PM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 24, 3:59 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9974, opened 2026-06-26
fix: preserve export keyword on multi-declarator exports under keepNames

### LINKED ISSUE #9973: [Bug]: `keepNames` drops the `export` keyword on multi-declarator exports (`export const a = 1, b = 2`)
## Reproduction link or steps

**REPL (copy the whole line):**

```
https://repl.rolldown.rs/#eNptUEFOxDAM/Irly4JURQJuQZy484JcQuquwrZO1LiwUpW/46ZaoBKXJJ4Zz9hZcUC7YuSerkbK9ma0v3WHQUu65jQLhMRFwMMLPHTwrtfjs+MDFxR8UlD7CK3MC9UO5zSOffpio5ohnv/E/MPsgXFqpiv0NESm10ZDhWFOE5xubSfHP/kq9Msoh4a71TFA5LyIhd3STCTeNKjbyLRIY5sS4EKU3/xExcI2fJNUPeu97qSrZB8u/kzmoyS+LXHA9vGbm8OeMulHcohUHGpIVafNSJ0+VTZ6oSJYvwEvkoOo
```

The REPL opens with `keepNames` already enabled and the entry below preloaded.

**Or reproduce manually:**

1. Open https://repl.rolldown.rs/
2. Set `index.ts` to:
   ```js
   export const a = 1, b = 2;
   export const c = 3;
   ```
3. In `rolldown.config.ts`, enable `keepNames`:
   ```js

   export default defineConfig({
     input: import.meta.input,
     output: {
       keepNames: true,
     },
   })
   ```
4. Look at the generated output.

---

## What is expected?

All three names should be exported, exactly as with `keepNames: false`:

```js
const a = 1;
const b = 2;
const c = 3;
export { a, b, c };
```

---

With `output.keepNames: true`, the multi-declarator export (`export const a = 1, b = 2;`) loses its `export` keyword. `a` and `b` are silently dropped from the bundle; only the single-declarator export `c` survives:

```js
const c = 3;
export { c };
```

`a` and `b` are gone.

This is not only silent loss. If another module imports those names, the build **fails** instead. Given:

```js
// dep.js
export const a = 1, b = 2;

// main.js (entry)
import { a, b } from './dep.js';
console.log(a, b);
```

`keepNames: false` builds fine, but `keepNames: true` aborts with:

```
[MISSING_EXPORT] "a" is not exported by "dep.js".
[MISSING_EXPORT] "b" is not exported by "dep.js".
```

So depending on how the multi-declarator export is consumed, `keepNames: true` either produces a wrong bundle (names missing) or fails the build outright. Single-declarator exports (`export const x = 1;`) are not affected.

shulaoda:
### What this solves

fixes #9973

With `keepNames: true`, a multi-declarator export loses its `export` keyword. The exported names are silently dropped from the bundle, or the build fails with `MISSING_EXPORT` when another module imports them by name. Single-declarator exports are unaffected.

```js
// input
export const a = 1, b = 2;
export const c = 3;

// before (keepNames: true)  ->  a, b are gone
const c = 3;
export { c };

// after (keepNames: true)
const a = 1;
const b = 2;
const c = 3;
export { a, b, c };
```

### Root cause

`visit_declaration` is shared by both a standalone declaration statement and an export's inner declaration (oxc calls it from `walk_statement` and from `walk_export_named_declaration`). Under `keepNames` it splits a multi-declarator `VariableDeclaration` into separate non-export statements and empties the original declarators. For an export, this runs first and clears the list, so the export-aware split in `visit_export_named_declaration` (`declarations.len() > 1`) is skipped and the `export` is lost.

### The fix

`visit_export_named_declaration` sets a flag before walking its inner declaration. `visit_declaration` consumes the flag (`std::mem::take`) and skips the non-export split for that direct declaration, so the export-aware split runs instead. Consuming the flag keeps it scoped to the direct child only, so nested multi-declarator declarations inside an export's initializer still split.

### Alternatives considered

- Doing the export-aware split before the walk: would move declarators out early and skip nested transforms (dynamic import / `require` transpose / `defer`).
- Detecting the export parent by AST address comparison: relies on oxc `GetAddress` aliasing internals, fragile.

### Tests

Added `tests/rolldown/issues/9973`: a module imports the multi-declarator export by name and asserts the values, which fails to build (`MISSING_EXPORT`) without this fix. Full `rolldown` integration suite passes with no other snapshot changes.

### DIFF
--- a/crates/rolldown/src/utils/tweak_ast_for_scanning.rs
+++ b/crates/rolldown/src/utils/tweak_ast_for_scanning.rs
@@ -20,6 +20,11 @@ pub struct PreProcessor<'ast, 'a> {
   drop_labels: Option<&'a FxHashSet<String>>,
   statement_stack: Vec<Address>,
   statement_replace_map: FxHashMap<Address, Vec<Statement<'ast>>>,
+  /// Set before walking an `ExportNamedDeclaration`'s inner declaration and taken by
+  /// `visit_declaration`, so the export's direct declaration skips the non-export
+  /// multi-declarator split (which would drop the `export` keyword) while nested
+  /// declarations still split. Reset after the walk for exports without a declaration.
+  next_declaration_is_exported: bool,
   /// Spans of `import defer ...` statements / expressions whose `defer` phase
   /// was lowered to a regular import. Read after `visit_program` to emit the
   /// `UNSUPPORTED_FEATURE` warning.
@@ -39,6 +44,7 @@ impl<'ast, 'a> PreProcessor<'ast, 'a> {
       drop_labels: drop_labels.filter(|set| !set.is_empty()),
       statement_stack: vec![],
       statement_replace_map: FxHashMap::default(),
+      next_declaration_is_exported: false,
       defer_spans: vec![],
     }
   }
@@ -188,9 +194,10 @@ impl<'ast> VisitMut<'ast> for PreProcessor<'ast, '_> {
   }
 
   fn visit_declaration(&mut self, it: &mut Declaration<'ast>) {
+    let is_exported = std::mem::take(&mut self.next_declaration_is_exported);
     match it {
       Declaration::VariableDeclaration(decl) => {
-        if decl.declarations.len() > 1 && self.keep_names {
+        if !is_exported && decl.declarations.len() > 1 && self.keep_names {
           let stmt_addr = self.statement_stack.last().copied().unwrap();
           let new_stmts = self.split_var_declaration(decl, None);
           self.statement_replace_map.insert(stmt_addr, new_stmts);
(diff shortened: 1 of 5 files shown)


crates/rolldown/src/utils/tweak_ast_for_scanning.rs
```diff
@@ -188,9 +194,10 @@ impl<'ast> VisitMut<'ast> for PreProcessor<'ast, '_> {
   }
 
   fn visit_declaration(&mut self, it: &mut Declaration<'ast>) {
+    let is_exported = std::mem::take(&mut self.next_declaration_is_exported);
```
IWANABETHATGUY, line 197: nit: 
`let is_exported = self.next_declaration_is_exported`

IWANABETHATGUY (approved):

crates/rolldown/src/utils/tweak_ast_for_scanning.rs
shulaoda: This is intentional. Otherwise we'd have to do:

```rust
let is_exported = self.next_declaration_is_exported;
self.next_declaration_is_exported = false;
```

Issue #9464, opened 2026-05-19
infra: attempt to remove tokio


Boshen:
This issue will be used as a context sync for our attempts to remove Tokio entirely, purely for fun and learning purposes.

Please have your clanker add a comment after any attempt.

Boshen:
Attempt: https://github.com/rolldown/rolldown/pull/9469

Boshen:
The team has agreed that the above PR is legitimate:

* Our usage of tokio and rayon results in two thread pools and real thread contention, removing tokio can improve performance (which the PR showed).
* It simplifies the code where we don't need to decide whether to use tokio or rayon

Brooooooklyn:
## Pluggable async runtime refactor

Follows the accepted direction from #9469, but keeps Rolldown's Rust and JavaScript async APIs intact: instead of converting the stack to synchronous APIs, the napi-rs async backend becomes replaceable, and Rolldown schedules async polling, CPU work, and bounded blocking I/O in one shared pool.

This comment is the single progress tracker for the refactor and is edited in place. Last updated: **2026-08-09**.

### Landed upstream

| PR | What | State |
|---|---|---|
| napi-rs/napi-rs#3352 | Pluggable `AsyncRuntime` SPI (+1,056/−115, additive — no semver break, byte-identical tokio fallback) | merged 2026-07-15 |
| napi-rs/napi-rs#3353 | First-class non-threaded `wasm32-wasip1` target + loaders in `@napi-rs/cli` (`wasm32-wasi` stays the threaded flavor's artifact name) | merged 2026-07-21 |
| napi-rs/napi-rs#3393 | The scheduler extracted from Rolldown as first-party packages: `napi-async-runtime` (crates.io) + `@napi-rs/async-runtime` (npm), wire-identical to the in-tree implementation | merged 2026-07-21 |
| napi-rs/napi-rs#3420 | Lock-free scheduler internals (injector run queue, lock-free spawn fast path — closes the SPI-adapter gap to napi-tokio) + drain-linger surface | merged 2026-07-27 |
| toyobayashi/emnapi#220, toyobayashi/emnapi#221 | Memory-growth view refresh; `setImmediate` bound to the global receiver (workerd) | merged, in `@emnapi/*` 2.0.0-alpha.3 |

Published: `napi-async-runtime` **0.2.0** (crates.io), `@napi-rs/async-runtime` **0.1.0** (npm).

### Where the work is now

[...]

PR #9977, opened 2026-06-26
feat: add shared async runtime

Brooooooklyn:
## Status

- Draft PR, rebased onto current `main` at `f09947ab017d6df74299f691853dcfc4f4f0f86e`.
- Current head: `0e072788718c2d771b0e85bd200f1065cfc9a0f1`.
- Uses napi-rs#3352 at `c1f70451d6ba957f09166fdd4fd345e31589e2f8`.
- Tracking issue: #9464.
- Local native and WASI validation is green. GitHub scheduler, Rust, native lifecycle, Node, WASI, and repository checks are green; only Vite Test Ubuntu failed in the independently reproduced upstream `scan jsx-runtime` pre-bundle timing regression (`ERR_OUTDATED_OPTIMIZED_DEP`).

## Summary

Add an optional Rolldown-owned async runtime that shares one Rayon-backed worker pool across async future polling, existing Rayon CPU-parallel stages, and bounded blocking filesystem/NAPI work.

The existing Tokio runtime remains the default behind `tokio-runtime`. The shared implementation is opt-in behind `async-runtime`.

## Architecture

`rolldown_utils::async_runtime` provides two flavors:

- `MultiThread`: async-task runnables and blocking closures use one fixed Rayon pool. Wakeups are batched into bounded queue-drain jobs.
- `CurrentThread`: a reentrancy-safe cooperative FIFO executor driven by a host task/timer registration. Blocking work runs inline.

The multi-thread scheduler has no separate async or blocking thread pool. `maxBlockingTasks` limits how many workers from the fixed pool may block at once. The NAPI `AsyncRuntime::spawn_blocking` hook routes NAPI and transitive callers through that bounded lane.

Scheduler admission, exact-owner blocking handoff, park-deadlock verdicts, timer publication/fire, and shutdown stop publication are serialized so work cannot cross a final stop verdict or be stranded by a missed wake. Host task and timer callbacks are snapshotted with panic-contained ownership and cancelled deterministically when their environment exits.

Classic bundler operations now hold explicit lifecycle guards. Failed build cleanup is cancellation-safe, retains each failure's originating cwd, preserves the first terminal error identity, and serializes retry/replay instead of launching abandoned fire-and-forget cleanup.

## Public API and configuration

```ts
import {
  configureAsyncRuntime,
  getAsyncRuntimeMetrics,
  getRuntimeCapabilities,
} from 'rolldown/experimental'

configureAsyncRuntime({
  flavor: 'MultiThread',
  workerThreads: 12,
  maxBlockingTasks: 11,
})
```

Environment equivalents:

```text
ROLLDOWN_RUNTIME=single|current-thread|multi|multi-thread
ROLLDOWN_WORKER_THREADS
ROLLDOWN_MAX_BLOCKING_THREADS
ROLLDOWN_PARK_DEADLINE_MS
```

Native environment-derived worker counts are capped at 256; native Tokio blocking threads are capped at 512. Explicit shared-runtime values above 256 reject with a validation error. WebAssembly shared-runtime builds normalize to one `CurrentThread` lane. Threadless `wasm32-wasip1` requires `async-runtime`; the unusable Tokio-only combination fails at compile time.

`getRuntimeCapabilities()` is the supported way to discover backend, target, timer, watch, and threading behavior. Generated NAPI declarations and the TypeScript wrapper expose the same limits and terminology.

## Dependencies

Uses the exact draft commit from napi-rs#3352 through `[patch.crates-io]`. The patch should be replaced with a released version before merge.

The WASI dependency set uses emnapi 1.11.2. emnapi#220 remains applied as a pnpm patch until released upstream. The non-threaded browser/WASI package switch remains in stacked PR #9978.

## Verification on `0e0727887`

Passed on Apple Silicon macOS with Rust 1.96.0, Node 26.4.0, pnpm 11.9.0, and WASI SDK 33:

- 319 `rolldown_utils` scheduler/unit tests;
- 33 binding async-runtime, 15 runtime-lease, 8 classic-bundler, and 6 core bundle-handle tests;
- `just test-async-runtime`: both shared-runtime flavors, main Node suite, and watch/dev-watch suites;
- workspace all-target Clippy and async-runtime-specific all-target Clippy with `-D warnings`;
- `just build-rolldown`, `just build-rolldown-async-runtime`, and `just build-rolldown-wasi`;
- locked `wasm32-wasip1` shared-runtime checks with and without default features;
- compile-time rejection of Tokio-only `wasm32-wasip1`;
- locked `wasm32-wasip1-threads` shared-runtime check;
- all 11 threaded WASI runtime lifecycle scenarios;
- 9/9 threaded WASI stability output comparisons identical;
- Node, repository, Rust formatting, generated declaration, and diff validation.

A final independent architecture review found one generated declaration mismatch; regeneration fixed it, and the review found no remaining scheduler, NAPI public API, or lifecycle issue.

## Performance note

Earlier benchmark numbers in this PR were measured on older scheduler heads and are historical, not acceptance data for `0e0727887`. The committed benchmark harness and results under `internal-docs/async-runtime/` remain the reproducible source for a fresh performance run.

## Current limitations

Native watch mode works on both shared-runtime flavors through the runtime timer facility. `BindingDevEngine` remains unsupported on native `CurrentThread`. WASI watch remains skipped because it stalls during the initial build before debounce timers are involved.


Brooooooklyn:
Superseded by #10268, which collapses this stack (#9977 + #9978 + #10243) into a single PR: it patches oxc/napi/emnapi to the tokio-free forks, makes the bespoke async runtime the **default**, ports Rolldown's own async off tokio, keeps the Cloudflare workerd tests, and drops tokio from the shipped graph (`cargo tree -p rolldown_binding -i tokio` finds none). Closing in favor of #10268.

PR #9978, opened 2026-06-26
feat: add non-threaded WASI build

Brooooooklyn:
## Status

- Draft PR, restacked on #9977 at `0e072788718c2d771b0e85bd200f1065cfc9a0f1`.
- Current head: `1d6bb6b60b36e1f9a4b924d39a8223f8624c3cf3` (13 commits on #9977).
- Uses napi-rs#3353 at `ae5196b5c1be97363834e29582ca4aadd4a3ab0f`.
- Tracking issue: #9464.
- Conflict-free; local validation and independent adversarial review are green. GitHub CI and the release-preview gate are running on the exact current head.

## Summary

Add a genuine non-threaded `wasm32-wasip1` Rolldown build and switch
`@rolldown/browser` to that artifact.

This PR is stacked on #9977 as 13 commits.

## Runtime and loader

The new binding is built with:

```text
wasm32-wasip1
--no-default-features
--features async-runtime
```

Its generated Node/browser loaders:

- use unshared `WebAssembly.Memory`;
- set `asyncWorkPoolSize: 0`;
- do not import or construct Web Workers;
- do not import `node:worker_threads`;
- do not require `Atomics.wait`.

The existing threaded WASI build and scripts remain available. Threaded and
single-thread flavors now use distinct artifact and npm-package names, so both
loader sets can be built, tested, and published side by side without collision
guards or build-order coupling.

## Build and CI

- Add debug/release `wasm32-wasip1` build scripts and a direct
  `just build-rolldown-wasi-single` recipe.
- Switch browser-package builds to the single-thread binding.
- Build and validate the complete threaded and single-thread artifact sets.
- Assert the single-thread loaders contain no Worker/shared-memory paths.
- Run threaded stability coverage immediately after the threaded build.
- Exercise real worker-environment teardown for both `CurrentThread` and
  `MultiThread` on macOS and Windows.
- Run the full Rolldown Node `test:main` profile against the single-thread WASI
  binding.
- Keep the synchronous devtools writer compiled only for tests and threadless
  WASI, preserving warning-free native all-target builds.
- Use package-filtered commands, native `node` for nested generation, and
  recipe-scoped `TARGET` so native runtime and single-thread WASI recipes work
  under both POSIX shells and PowerShell.
- Bootstrap missing WASI npm package skeletons from a clean checkout while preserving release-downloaded package-local Wasm artifacts.\n- Serialize package replacement with crash-recoverable filesystem transactions and a stale-lock bakery guard that never mutates successor-owned paths.\n- Run 32 rollback, contention, ABA, symlink, and crash-boundary regressions on Linux plus the native macOS/Windows lifecycle matrix.\n- Pack browser consumers through `vp pm pack`, and run real-Wasm loader tests
  only in jobs that built the WASI artifact while retaining unit/codegen coverage
  in native-only Node jobs.

The WASI Node profile is green. Unsupported behavior is represented by explicit
runtime/target-aware skips rather than treating the suite as generally
non-passing: wasm error-object round trips, directory-symlink traversal, dev
mode on `CurrentThread`, and WASI watch tests.

## Temporary upstream backports

Until the upstream drafts are released:

- Cargo pins napi-rs#3353 at
  `ae5196b5c1be97363834e29582ca4aadd4a3ab0f` for the target, linker, loader,
  deferred workerd loader, and per-flavor artifact implementation.
- pnpm patches `@napi-rs/cli@3.7.2` with the regenerated ESM/CJS/standalone CLI
  output from that branch.
- emnapi, `@emnapi/core`, and `@emnapi/runtime` resolve to 1.11.2. This consumes
  the released bound-`setImmediate` fix from emnapi#221.
- pnpm still patches `@emnapi/core@1.11.2` with emnapi#220 to refresh detached
  views after unshared `memory.grow`.

These pins and patches should be removed before merge or in a follow-up once
the upstream releases are available.

## Verification

- frozen pnpm install with both patched dependencies represented in the lock
- locked native `async-runtime` binding check using napi-rs#3353's `napi-build`
- focused napi `spawn_blocking` routing test
- threaded WASI stability comparisons
- full Node `test:main` WASI profile with explicit capability-based skips
- supported WASI example builds
- 2,512-module fixture with an async JavaScript transform callback on every
  module
- 19,014-module production fixture with minification and sourcemaps
- generated-loader checks for artifact completeness and threadlessness\n- clean-checkout package bootstrap and release-style staging with source Wasm absent and the registry unreachable\n- 32/32 focused package transaction and recovery tests, including two-process equal-ticket ordering and every reclaim publication crash boundary

### Local WASI run

Verified on Apple Silicon with Rust 1.96.0, both rustup WASI targets, and WASI
SDK 33:

- threaded artifact set complete; module exports `_initialize`
- threaded stability: all 9 repeated-output comparisons identical
- single-thread artifact set complete; module exports `_initialize`
- threadlessness guard passed: no Worker, shared memory, or async worker pool
- full Node profile: 31 files passed / 3 skipped; 768 tests passed / 35 skipped
- all supported example builds passed

### Exact-head automation\n\n- CI: https://github.com/rolldown/rolldown/actions/runs/29003691159\n- Release preview: https://github.com/rolldown/rolldown/actions/runs/29003694346\n\n## Dependencies

- #9977
- https://github.com/napi-rs/napi-rs/pull/3353
- https://github.com/toyobayashi/emnapi/pull/220
- Tracking: #9464


Brooooooklyn:
Superseded by #10268, which collapses this stack (#9977 + #9978 + #10243) into a single PR: it patches oxc/napi/emnapi to the tokio-free forks, makes the bespoke async runtime the **default**, ports Rolldown's own async off tokio, keeps the Cloudflare workerd tests, and drops tokio from the shipped graph (`cargo tree -p rolldown_binding -i tokio` finds none). Closing in favor of #10268.

PR #9980, opened 2026-06-26
fix: reconstruct nested rest patterns in into_expression

IWANABETHATGUY:
Stacked on #9976.

### What this solves

After #9976, `BindingPattern::into_expression` rebuilds rest elements as spreads, but it still assumes the rest *argument* is a `BindingIdentifier` and panics via `unreachable!` otherwise:

```rust
let BindingPattern::BindingIdentifier(ref id) = rest.argument else {
  unreachable!("The rest element should be `BindingIdentifier`")
};
```

That assumption holds for **object** rest — the grammar forces `{ a, ...rest }` to bind an identifier — but not for **array** rest. Both of these are legal JS where the rest argument is a nested pattern:

```js
const [a, ...[b, c]] = x;   // rest argument is an ArrayPattern
const [a, ...{ length }] = x; // rest argument is an ObjectPattern
```

So `into_expression` could panic on a perfectly valid pattern. (As noted in #9976, the array path isn't reached today by the only caller — the Vite `__vitePreload` rewrite — because a namespace object isn't iterable, so this is a latent panic rather than an active miscompile.)

### The fix

Recurse through `into_expression` on the rest argument instead of asserting it is an identifier, mirroring the sibling `into_assignment_target`, which already does exactly this for both object and array rest with no guard. The rebuilt spread now round-trips any nested rest pattern. The object path is unchanged behaviorally (its argument is always an identifier) but now shares the same uniform code path.

Adds a regression test (`const [a, ...[b, c]] = x;`) that would hit the old `unreachable!`.

### DIFF
--- a/crates/rolldown_ecmascript_utils/src/extensions/ast_ext/binding_pattern_ext.rs
+++ b/crates/rolldown_ecmascript_utils/src/extensions/ast_ext/binding_pattern_ext.rs
@@ -96,12 +96,9 @@ impl<'ast> BindingPatternExt<'ast> for BindingPattern<'ast> {
           )));
         });
         if let Some(rest) = obj_pat.rest.take() {
-          let BindingPattern::BindingIdentifier(ref id) = rest.argument else {
-            unreachable!("The rest element should be `BindingIdentifier`")
-          };
           properties.push(ObjectPropertyKind::SpreadProperty(
             ast_factory
-              .alloc_spread_element(SPAN, ast_factory.expression_identifier(SPAN, id.name)),
+              .alloc_spread_element(SPAN, rest.unbox().argument.into_expression(ast_factory)),
           ));
         }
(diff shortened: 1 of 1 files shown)


hyfdev (approved):

shulaoda (approved):

IWANABETHATGUY:
### Merge activity

* **Jun 26, 12:40 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 26, 12:41 PM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 26, 1:03 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9986, opened 2026-06-26
fix(treeshake): keep computed-key side effects on namespace member access

shulaoda:
### What this solves

When the object of a member expression is an ES module namespace import (`import * as ns`), the side-effect detector treated the whole access as side-effect-free. That is correct for the property read itself (namespace objects are frozen), but a computed key is still evaluated and can have its own side effects. So a statement like `ns[foo()]` was judged pure and tree-shaken away, silently dropping `foo()`.

```js
import * as ns from './dep.js'
ns[foo()] // foo() was silently dropped
```

### The fix

For a computed member access on a namespace object, return the side effect of the key expression instead of unconditionally reporting no side effect; static and private accesses stay side-effect-free. This mirrors how `collect_write_target_metadata` already handles computed assignment targets.

### Tests

Added `tests/rolldown/tree_shaking/namespace_computed_key_side_effect`: a computed namespace key whose side effect throws at runtime if it is dropped. It fails without the fix and passes with it, and the full integration suite passes with no other snapshot changes.


IWANABETHATGUY (approved):

shulaoda:
### Merge activity

* **Jun 26, 3:59 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 26, 4:00 PM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 26, 4:06 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9987, opened 2026-06-26
fix: don't classify await in a strict-mode function as top-level await

shulaoda:
### What this solves

`is_valid_tla_scope` accepted any scope whose flags intersected `Top | StrictMode`. oxc marks a class scope as exactly `StrictMode` (so a class computed key like `class T { [await x]() {} }` counts as top-level), but it also adds `StrictMode` to a function or arrow that carries its own `"use strict"` directive (`Function | StrictMode`). So `await` inside such a function was misclassified as top-level await.

For a non-ESM output format this produced a spurious build error:

```js
// bundled to cjs
export async function f() {
  'use strict'
  await Promise.resolve() // "Top-level await is currently not supported with the 'cjs' output format"
}
```

even though the `await` is inside an async function and is valid for any output format.

### The fix

Use `flag.is_top() || flag.is_block()`. `is_block()` is `is_empty() || self == StrictMode`, so it still covers block scopes and class scopes, while a strict function (`Function | StrictMode`) is neither top nor block and is correctly rejected.

### Tests

Added `tests/rolldown/misc/use_strict/await_in_strict_function_is_not_tla`: a cjs build of an async function with a `"use strict"` directive and an inner `await`. It fails with the spurious TLA error without the fix and passes with it, and the full integration suite passes with no other snapshot changes.

### DIFF
--- a/crates/rolldown/src/ast_scanner/mod.rs
+++ b/crates/rolldown/src/ast_scanner/mod.rs
@@ -285,11 +285,7 @@ impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
   /// }
   /// ```
   pub fn is_valid_tla_scope(&self) -> bool {
-    self
-      .scope_stack
-      .iter()
-      .rev()
-      .all(|flag| flag.intersects(ScopeFlags::Top | ScopeFlags::StrictMode) || flag.is_block())
+    self.scope_stack.iter().rev().all(|flag| flag.is_top() || flag.is_block())
   }
 
   pub fn is_root_scope(&self) -> bool {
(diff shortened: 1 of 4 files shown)


IWANABETHATGUY (approved):

shulaoda:
### Merge activity

* **Jun 28, 5:50 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 28, 5:51 AM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 28, 5:56 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9988, opened 2026-06-26
fix: deconflict chunk symbols by exec_order, not chunk render order

shulaoda:
### What this solves

With `experimental.chunkModulesOrder: "moduleId"`, chunk-symbol deconfliction used the wrong priority order. When a chunk holds two or more side-effect-free leaf modules (e.g. modules that only `export function foo() {}`, so they are not constant-folded away) whose id order differs from their exec_order, a debug build panics:

```
chunk.modules must be in ascending exec_order for deconfliction
```

In a release build (the `debug_assert` is compiled out) it does not panic but picks the wrong deconfliction winner: when two modules declare the same top-level name, the module that keeps the clean name depends on the `moduleId` render order instead of `exec_order`, so it disagrees with the `exec_order` priority used everywhere else (including cross-chunk export naming).

### Root cause

Deconfliction must process modules in descending `exec_order` (entry first) so the entry's symbols keep their clean names; this is the canonical order encoded by `deconflict_order_key` = `(Reverse(exec_order), name)`. `deconflict_chunk_symbols` achieved it with `chunk.modules.iter().rev()`, which only yields descending exec_order because the default `ExecOrder` mode happens to sort `chunk.modules` ascending by exec_order. `chunkModulesOrder: "moduleId"` intentionally orders leaf modules by id for the output, so `chunk.modules.rev()` is no longer exec_order-descending. PR #9831 added a `debug_assert` to document this invariant, but the `moduleId` branch violates it.

### The fix

Drive the deconfliction order from `exec_order` directly — `sorted_by_key(Reverse(exec_order))` — instead of relying on `chunk.modules`'s order, and drop the now-redundant `debug_assert`. This decouples the deconfliction priority (always exec_order) from the chunk's render order, so it is correct for every `chunkModulesOrder`, matching `deconflict_order_key` and the cross-chunk naming. For the default `ExecOrder` mode the order is unchanged (exec_order is a unique total order, so the explicit sort equals the previous `.rev()`), confirmed by zero snapshot drift across the suite.

### Tests

Added `tests/rolldown/topics/chunk_modules_order/side_effect_free_leaf_modules`: two side-effect-free leaf modules that both declare `dup`, with id order opposite to exec_order, under `chunkModulesOrder: "moduleId"`. It panicked without the fix; the snapshot now pins the exec_order winner (the higher-exec_order module keeps `dup`), which matches the default `ExecOrder` output. The full integration suite passes with no other snapshot changes.

### DIFF
--- a/crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
+++ b/crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
@@ -1,3 +1,6 @@
+use std::cmp::Reverse;
+
+use itertools::Itertools;
 use oxc_str::CompactStr;
 
 use crate::{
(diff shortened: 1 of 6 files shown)


crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
```diff
@@ -141,25 +144,15 @@ pub fn deconflict_chunk_symbols(
...
+  // priority as `deconflict_order_key` used for cross-chunk export naming. This is
+  // driven by `exec_order` explicitly rather than by `chunk.modules`'s order, which
+  // under `chunkModulesOrder: "moduleId"` is not exec-order-sorted.
   chunk
     .modules
     .iter()
     .copied()
-    // Starts with entry module
-    .rev()
+    .sorted_by_key(|idx| Reverse(link_output.module_table[*idx].exec_order()))
```
IWANABETHATGUY, line 155: I think you could still preserve the previous implementation for the `chunkModulesOrder: 'exec-order'` branch, since most users still rely on that order, allowing us to skip sorting in that case.

hyfdev:
Rough thought:

> deconflict chunk symbols by exec_order,

I think execution order is less stable than order sorted base on filenames. Will this make cache invlidate more frequently?

PR #9989, opened 2026-06-26
fix: avoid spurious leading newline in addon hooks (banner/footer/intro/outro)

shulaoda:
### What this solves

The addon driver joins each plugin's `banner` / `footer` / `intro` / `outro` result with `\n`, but it pushes the `\n` separator before every result, including the first one. When there is no corresponding output option the accumulator starts as an empty string (`injection.unwrap_or_default()` in `ecma_generator.rs`), so the first plugin result becomes `"\n" + result`.

This adds a spurious blank first line, and it breaks the common pattern of adding a shebang via the `banner` hook:

```js
// plugin
{ name: 'cli', banner: () => '#!/usr/bin/env node' }
```

```js
// before (no output.banner): shebang is no longer on line 1
\n#!/usr/bin/env node
...
```

A shebang is only valid on the first line, so it is no longer recognized; the downstream `banner.starts_with("#!")` shebang detection in `ecma_generator.rs` also evaluates to false. Rollup joins addons with `\n` and skips empty parts, so it does not emit the leading newline.

### The fix

Only push the `\n` separator when the accumulator is non-empty, so it sits between results rather than before the first one. Applied identically to `banner`, `footer`, `intro`, and `outro`.

### DIFF
--- a/crates/rolldown_plugin/src/plugin_driver/output_hooks.rs
+++ b/crates/rolldown_plugin/src/plugin_driver/output_hooks.rs
@@ -45,7 +45,9 @@ impl PluginDriver {
       let result = plugin.call_banner(ctx, &args).await;
       self.record_timing(plugin_idx, start);
       if let Some(r) = result.with_context(|| CausedPlugin::new(plugin.call_name()))? {
-        banner.push('\n');
+        if !banner.is_empty() {
+          banner.push('\n');
+        }
         banner.push_str(r.as_str());
       }
     }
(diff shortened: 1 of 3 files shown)


hyfdev (approved):

hyfdev:
### Merge activity

* **Jun 27, 4:02 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 27, 4:02 PM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 27, 4:07 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

Issue #8197, opened 2026-02-04
[Feature Request]: improve treeshaking for `require('./foo.json')`


sapphi-red:
### What problem does this feature solve?

Improve treeshaking for `const foo = require('./foo.json')` like it is supported for `import foo from './foo.json'`.

[REPL](https://repl.rolldown.rs/#eNpFTc0KwjAMfpWQyyaUih4Lvkkvc2YyKQ22nQil726i1B0C+f4rLugqrvFGb/vI+kd0OzY4C5w55gILM1wg0XNbE42DPQohHo7DwScf9dTIgWzg+6iq3FeTHkJX0kbNYI/1sT/+jVUNAHjlPTo4mU5cp6TEWXHDJlUv8YepUC7YPm2zQaA=)
([REPL for import (working)](https://repl.rolldown.rs/#eNpFTcsKAjEM/JWQyyqUih77Lb2saytKt5G0ilD67yZIdw+BeWRmGkZ0DR/5Fr72WRRndDs3uChdX8QVIhFEphUmexIsPuXJs896C+VCKdhE94OackfVpSKgq/wO3eBIjZ2N/3eaBgC86h4dnM0QrjOrcFHesUvVR/7TXEOp2H9LkUA9))

### What does the proposed API look like?

Not an API change

hyfdev:
I saw https://github.com/rolldown/rolldown/pull/9990 attemps to fix this by adding support inside of the rolldown core.

I prefer not to add code to rolldown core to support this, it increases the internal complexity. It's doable but I tend to avoid it if possible. I saw treeshaking, strictExecutionOrder and barrelImports are producing bugs affecting each other. I tend avoid adding more complexity if there's another way to do it.

I wonder how rollup support this, any reason we couldn't do the same following rollup?

If needed, a builtin json plugin would be better than builtin features for maintaince and stablity.

---

I changed the priority to P2 as it's part of the vite8 migration but we need to reach at a conclusion before we implement it.

hyfdev:
**How rollup achieves `require('./foo.json')` treeshaking — for reference**

Rollup has no core/builtin feature for this; it falls out of two plugins + rollup's general optimizer:

1. **`@rollup/plugin-json`** lowers `foo.json` into a *real ESM module* whose default export is a plain object literal built from per-key bindings:
   ```js
   export var foo = 111;
   export var bar = 999;
   export default { foo, bar };
   ```
[...]

Issue #2995, opened 2024-12-01
[Feature Request]: Support object tree shaking 


lishaduck:
### What problem does this feature solve?

I switched from using `require()` to load a `package.json` to a static `import`. This inlined the JSON (fine by me; I would've prefered externalizing it but putting `./package.json` as external didn't seem to work), which enlarged the file significantly. I don't want to it bundle all of the file, just what's used (the `version` field). Using a namespace import worked, but Node.js stabilized JSON modules without named exports, just a default export, so I'd prefer to use that.

### What does the proposed API look like?

Rollup recently landed object tree shaking (though it was reverted), which seems like a more correct solution than special-casing JSON default import that are only used statically.

hyf0:
This should be supported. Could you create a minimal reproduction for your case in https://stackblitz.com/github/rolldown/rolldown-starter-stackblitz?  @IWANABETHATGUY cc

IWANABETHATGUY:
I think this is related to https://github.com/rolldown/rolldown/pull/3003, seems the wrapkind detector is wrong, we think `test.json` `module.exportsKind = Esm` even `test.json` is `require` by entry.

IWANABETHATGUY:
for esbuild, it is correct determined, https://hyrious.me/esbuild-repl/?version=0.23.0&b=e%00entry.js%00const+a+%3D+require%28%27.%2Fpackage.json%27%29%0Aconsole.log%28a.a%29&b=%00test.json%00&b=%00package.json%00%7B%0A++%22type%22%3A+%22module%22%0A%7D&o=%7B%0A++treeShaking%3A+true%2C%0A+%0A%22bundle%22%3A+true%2C%0Aformat%3A+%22esm%22%2C%0Aminify%3A+false%2C%0Adefine%3A+%7B%0A++%22a.b%22%3A+%22a.c%22%0A%7D%0A%7D.

IWANABETHATGUY:
If we remove `type: module` in `rolldown/examples/basic/package.json` the output should same as esbuild.

IWANABETHATGUY:
## Output when have `type: module` in package.json
<summary>
 <details>

```js

//#region rolldown:runtime
var __defProp = Object.defineProperty;
var __getOwnPropDesc = Object.getOwnPropertyDescriptor;
var __getOwnPropNames = Object.getOwnPropertyNames;
var __hasOwnProp = Object.prototype.hasOwnProperty;
var __esm = (fn, res) => function() {
[...]

PR #9990, opened 2026-06-26
perf: tree-shake JSON require binding properties

TheAlexLichter:
This PR tree-shakes unused properties from JSON files accessed through simple top-level `require()` bindings. It conservatively falls back when behavior could be affected through:
* object identity, 
* mutation,
* evaluation timing,
* external exposure, 
* or unsupported access patterns

This should solve https://github.com/rolldown/rolldown/issues/8197 and is meant to iterate further on so we can fully solve https://github.com/rolldown/rolldown/issues/2995 **eventually**.


IWANABETHATGUY:
I have no idea how to implement `require JSON tree shaking` yet, and I also don't have the bandwidth to do so. Given that this is just a P3 task, I prefer to leave it as is.

TheAlexLichter:
the main problem is that this is a Vite 8 migration regression. Not necessarily a blocker but p3 seems a bit low for it.

hyfdev:
I think this PR needs to prove it's doing something that's architecturally right. If it can't and the team isn't given bandwidth to this feature, we will have to delay handling this PR.

If Vite8 migration requires this feature and it's indeed important, we should raise it and allocate bandwidth.

Adding features to bundler increases the complexity in a non-linear way, reviewing if it's correct takes more time than implementing it.

I suggest we start a discussion first.

---

Also I think rollup doesn't support this as a builtin feature, any reason we couldn't do the same following rollup?

If needed, a builtin `json` plugin would be better than builtin features for maintaince and stablity.

TheAlexLichter:
I see, this is a bit bigger than I expect. Let me draft up an impl plan in https://github.com/rolldown/rolldown/issues/2995 instead.

PR #9991, opened 2026-06-26
fix(plugin): make lazy hook metadata enumerable

TheAlexLichter:
This is more my alley :D 

These properties should be `enumerable` (oversight that they aren't). Otherwise they don't behave like normal JS objects and they are not listed when spreading or using `Object.keys` (and others).

Related: https://github.com/vitejs/vite/pull/22765


hyfdev (approved):
👍

PR #9972, opened 2026-06-25
fix: handle JSON default mutation bailouts

TheAlexLichter:
This PR hardens the existing JSON default-import tree-shaking optimization. It prevents Rolldown from replacing `json.a` with the static split JSON export when the shared JSON default object may have been mutated elsewhere. 

It also catches the less obvious case where that mutation happens through a named re-export.

## Example

[REPL](https://repl.rolldown.rs/#eNpdj81ugzAQhF9ltRdaidI7Vd/EFxcvaVpYI/8kkRDvnt0QB5Kbd2Y033jGHtsZj+zo0vxFfTO2211jp+c4+ZCgaj7HnGzyQazqy/Bd7zz3xwP0wY+aWU+JeNaQYRGiH6gZ/OHtbo7e0buYAiBsU8i01LiVlx075WnJDM4mC8uDeQ52mqjsMqz2DQLfUDk6qYqC2HIFsVNWBF0KgnqbhwQ2vtJefqjFO6k076W1ejYMYFB3GWzlNQXvDNarnDlHklONzjL8EKRA9BF/7T+xQcOLsgR2kq7BJooJlyvh0pW+)

Right now, Rolldown treats config.mode like a static JSON named export and emit something equivalent to:
```js
var config_default = { mode };
config_default.mode = "dev";
console.log(mode); // "prod" - wrong
```

After:

```js
var mode = "prod";
var config_default = { mode };

config_default.mode = "dev";

console.log(config_default.mode); // "dev" - correct
```

PR #9996, opened 2026-06-28
fix(treeshake): also bail JSON default split when the object escapes

IWANABETHATGUY:
## What this solves

[#9972](https://github.com/rolldown/rolldown/pull/9972) keeps a JSON default object materialized (instead of splitting `data.key` reads into per-key exports) once it observes a member **write** — `data.foo = ...`, including through a named re-export. But the split optimization is equally unsound when the default object **escapes** without a write rolldown can see, and `data.key` reads then resolve to the stale split export instead of the live object:

- aliasing — `const o = data; o.foo = ...`
- call argument — `f(data)` where `f` mutates it
- dynamic index — `data[k] = ...`

Minimal reproduction (alias case) — **[live REPL](https://repl.rolldown.rs/#eNptk8Fu2zAMhl/lHy9OUE+9u8hhHYYBK7Acht10qGzTiTqFCiS5WRYY2EPsCfckg+QmTYFdBJDC/4n8SZ3omRpyJnFMVNNAzYl6k4x6il5yINRcJWrqqKETND0bN7KmBppaHnxgTZi00FTTzlhRT/GsPoez1u72PiRkJIbgd6jU7YVf3WnRcnuLx5x6xCJtGV++rb+i58GMLsG3T9ylJTh2Zs8R7REtW9nAOGsi9zXSliUfc6bQdmPKDZZs3JrA/QsHVrB3pmOlpfMS0yzCqtR3p6WEqvSKFSozJA6XIj/kqEDLA9ZLPRc+Cx6xG2NCy2edwlpgBGYYuEvcox2t6zPgteezNLDp0bLzBxxMRMz8zjh3ROBDsCnlJn15fM/h/Q8+QlPcO5s0FRr/LD4vhuB/scAkVPOcqmWN3j770HE/T+DSwasvf3//Qcx4OzuYtsEfIqzElAvzA5zfbLLv6welxQ5YvFaPd6uLV0uctGDWQ/iATyH4sKjuv39ucCVZocJNGbWKKVjZ2OF4hVziBlWdm5qd01TwmqrlnZYpDyTPzztWzm8W1frhLb2qr8IsoZqYmhRGnmoK3rneH0R1Xga7Uemyu/+5ebPGp7yYVvhjucb0stJnWZULe5nEeYOvBYtijiYr+zHlzzRj1Y6TUXNSpmX+VdP0D4saQXo=)**:

```js
// data.json — { "value": "before" }
import data from './data.json';
const alias = data;            // `data` escapes
alias.value = 'after';         // mutate the shared object
data.value;                    // main: "before"  ❌   (should be "after")
```

On `main` the read is compiled to a bare `value` split const (`"before"`), divorced from the mutated object.

## The fix

Flip the JSON-default scan from a **blacklist** ("optimize unless we spot a write") to a **whitelist** ("only split when every reference to the default is a plain `data.key` read"). The enabling signal: a static `data.key` access is recorded as a `MemberExpr`, so **any bare-symbol reference** to the default binding is an escape we can't reason about — mark it non-splittable, in addition to the member writes #9972 already handles. The three ad-hoc helpers are folded into one classifier (`collect_non_splittable_json_defaults` / `json_default_made_non_splittable_by` / `json_default_canonical_ref`), replacing the write-only `collect_json_default_imports_with_member_write`.

This is free in practice: if the object escapes, the whole thing is already materialized, so no per-key tree-shaking was possible to begin with — bailing only changes *which* form is emitted, never what gets dropped.

## Alternatives explored

Continuing to extend the blacklist with more mutation patterns (the #9972 direction). Rejected — it is unsound by construction: every escape pattern not yet enumerated is a latent miscompile. The whitelist is sound by construction and strictly supersedes the write-only checks.

## Tests

`tree_shaking/json_default_escape_bailout` — an **executed** fixture (real runtime asserts) covering the two escape forms `main` still mis-optimizes: **aliasing** (`const o = data`) and **passing as a call argument** (`f(data)`). Both fail on `main` and pass here.

The other non-splittable categories are intentionally **not** duplicated — each already has dedicated coverage that passes on `main`:

- computed-write → `issues/9484_computed_write`
- named-re-export / cross-module write → `tree_shaking/json_default_import` (+ `issues/9484`)
- read-only split + tree-shake (no over-bail) → `tree_shaking/json_default_import` (`foo.json`)

The `json_default_import` snapshot also drops a now-redundant split for the whole-object-use `bailout.json` case.

Full `rolldown` suite: **1771 passed, 0 failed**; `clippy --deny warnings` clean.

## Reviewer attention

- The soundness of the *"bare `Symbol` = escape"* signal — i.e. that the scanner records static `x.key` accesses as `MemberExpr` and everything else (alias, arg, dynamic index, return) falls through to a bare `Symbol`.
- No over-bail: the read-only coverage above + full green suite confirm read-only JSON defaults still split and tree-shake unused keys.


crates/rolldown/tests/rolldown/tree_shaking/json_default_import/artifacts.snap
```diff
@@ -20,15 +20,14 @@ reexport_mutation_default.a = "mutated";
...
 var bailout_default = {
-	a,
+	a: "bailout_a",
 	b: "bailout_b"
 };
 //#endregion
 //#region main.js
 assert.strictEqual(a$1, "used");
-assert.strictEqual(a, "bailout_a");
+assert.strictEqual(bailout_default.a, "bailout_a");
```
IWANABETHATGUY, line 30: **Why this snapshot changed** — only the `bailout.json` region differs from `main`.

`bailout.json`'s default is used as a *whole object* (`assert.deepEqual(mod2, …)` in `main.js`), which this PR now classifies as an **escape**, so it stays materialized as `var bailout_default = { a, b }`.

Before (the merged write-only bailout), the `mod2.a` read was still rewritten to a per-key split const, so the output emitted **both** `var a = "bailout_a"` **and** the full `bailout_default` referencing it via `{ a }`. That split was pure redundancy: an escaping object keeps every key, so nothing was tree-shakeable. Dropping it is behavior-preserving — `bailout_default` is identical either way and the executed fixture's asserts still pass — it just removes the dead const and reads through `bailout_default.a`.

The other regions are intentionally unchanged: `foo.json` is read-only (`mod.a`) so it still splits and tree-shakes the unused `b` (`var a$1 = "used"`), and `reexport-mutation.json` is mutated through the re-export so it stays materialized.

hyfdev (approved):

hyfdev:
### Merge activity

* **Jun 28, 12:11 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 28, 12:13 PM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 28, 12:19 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

Issue #5726, opened 2025-08-14
[Bug]: Rolldown creates unnecessary facade chunks


wmertens:
### Reproduction link or steps

rolldown produces unnecessary facade chunks, which causes extra round trips and therefore severely impacts startup time.

https://stackblitz.com/edit/vitejs-rolldown-vite-5gozxfed?file=vite.config.ts

in the repro, run `vite build`. It creates 2 files with `counter` in the name, and one of them is a facade chunk.

(relevant code:
```js
      {
        name: 'signature',
        enforce: 'pre',
        resolveId: {
          order: 'pre',
          async handler(id, imp) {
            if (id === './counter.js') {
              console.log('turning counter into a chunk');
              const result = await this.resolve(id, imp, { skipSelf: true });
              if (result) {
                this.emitFile({
                  id: result.id,
                  type: 'chunk',
                  preserveSignature: 'allow-extension',
                });
              }
              return result;
            }
          },
        },
      }
 ```
)

### What is expected?

Facades are only created when needed

### What is actually happening?

Any entry gets a facade

### System Info

```Shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 20.19.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.8.2 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    rolldown-vite:  7.1.2
```

### Any additional comments?

See https://github.com/rollup/rollup/pull/5891 for how this was solved in rollup.

TheAlexLichter:
As a note, this is also an issue in https://github.com/nuxt/nuxt/ with rolldown-vite.

IWANABETHATGUY:
We will try to add this optimization in September.

hakimio:
@IWANABETHATGUY any progress?

TheAlexLichter:
Note: While there were more improvements to the default chunking behavior, this is still an issue with beta.58 / vite 8.0.0-beta.5

<img width="736" height="380" alt="Image" src="https://github.com/user-attachments/assets/319c6586-d95a-476b-933f-b06786088a14" />

IWANABETHATGUY:
[...]

PR #9997, opened 2026-06-28
fix(rolldown): keep entry facade when a shared chunk holds another entry's module

### LINKED ISSUE #9463: `chunkOptimization` causes the output has wrong execution order with `strictExecutionOrder: true`
Repl: https://repl.rolldown.rs/#eNqVUk2L2zAQ/SuDLk5C6qQ9CrZQ2u1pl0DbW12Kao0TUXtk9LHJ1vi/V7JixyHLwvpgRvPx5s286VjFeMeMrmupj5SXmiq1z52NXmL8pcialSGgmlYbB+65Rejg2zlt1zqlyUIPldENZGN5VlBBm9UKVpuCApR1kAD5TekddAUBKGq948kGEBwyRRJPgUC2Tr4/o+/dh8nbD388tWhUg+REPUGUB09/Y5NG/ROxFQdnPM7KtHfzntYZVbr7E5Y+pu+MRDOrCYha4ve2Vs4pCoOcywD2RvvWcvg5OuASix+JBjkUrNRNo6lgZ7j0BdJGof10FCYkzbrdhh/R7PHHwaA96FpyeL/9vd1uZ+n9aP5KRhoz/kOkoLCmKKHESvh6FIT1azZuejyC6X0lfQeV1pPS+cYeAicZpYjgUpc+KpALKe+fgvGgrENCs8i+7B4/awoP96CFRJmtYbGEu49pSwF0sYxGPBNdY47G6FA1yR+C/TJQQcbjeka+wxVcMU6exPltdIR9pnJOKp1so6WvMVyoOArlIO1hcTX6wDzl5a9OMtzs7SgT0jjJxZEGOYtWeSrjVaZ1XUjO2nzd7YYGBbE+ID+F4lo4tI71/wG7t0xM

Expected:

Executing from entry `b.js` should not trigger side effect of entry `a.js`

Current:

Executing from entry `b.js` triggers side effect of entry `a.js`

 _Originally posted by @hyf0 in [#9450](https://github.com/rolldown/rolldown/issues/9450#issuecomment-4487757944)_

hyfdev:
## Summary

Fixes the `strictExecutionOrder` execution-order corruption reported in #9463: with `experimental.chunkOptimization` and a `codeSplitting` group that captures multiple **user-defined entry** modules, executing one entry eagerly ran another entry's top-level side effects.

Reproduction (the issue's REPL, as `tests/rolldown/issues/9463/`): entry `a` statically imports `shared`, entry `b` dynamically imports it, with an `entriesAware` group + large `entriesAwareMergeThreshold` + `strictExecutionOrder`.

Before — loading `b` runs `a`'s side effects:
```js
// a.js  → init_a/init_b/init_shared live here; chunk top-level unconditionally calls:
init_a();
export { init_b as t };
// b.js
import { t as init_b } from "./a.js";   // importing a.js runs init_a() first
init_b();
```
After — each entry runs only its own `init_*`:
```js
// a.js → import { n as init_a } from "./common~a~b~shared.js"; init_a();
// b.js → import { t as init_b } from "./common~a~b~shared.js"; init_b();
```

## Root cause

Manual code splitting can place several user-defined entry modules into one shared `Common` chunk (the `entriesAware` `entriesAwareMergeThreshold` merges subgroups by bitset similarity and unions their bits; a plain group unions bits directly). `chunkOptimization`'s facade elimination (`find_facade_chunk_merge_ops`) then folds that shared chunk **into** one entry chunk, which makes that entry chunk eagerly run its own `init_*`. The other entry, whose module also lives in the shared chunk, is then forced to `import` the first entry chunk to reach its own module — and running that chunk eagerly executes the first entry's side effects.

The existing safety check only ran for `preserveEntrySignatures: 'strict'`. Entries with no exports resolve to `AllowExtension` and took the `_ => true` arm, so nothing blocked the fold.

## The fix

In the user-defined-entry branch of `find_facade_chunk_merge_ops`, only fold the common chunk into the entry chunk when that chunk does **not** also hold another user-defined entry's module. Otherwise the thin facade is kept, so each entry imports the (wrapped) shared chunk and runs only its own `init_*`.

A common chunk that holds just this entry's module (plus non-entry deps a sibling genuinely depends on) is still folded — preserving the #5726 facade elimination.

## Tests

Three fixtures with **behavioral** `_test.mjs` that execute the output and assert the actual semantics (not just chunk snapshots) — each module records its top-level side effect in a global; loading entry `b` must run only `b`'s side effects and must not contain another entry's:
- `tests/rolldown/issues/9463/` — the reported case (entriesAware group, dynamic shared dep).
- `tests/rolldown/issues/9463_plain_group/` — plain (non-`entriesAware`) group, to prove the guard is not entriesAware-specific.
- `tests/rolldown/issues/9463_three_entries/` — three entries, to prove it scales past two.

All three were confirmed to **fail on the pre-fix code** (e.g. `["a","shared-foo","b","shared-foo"]`) and pass with the fix.

I also audited the rest of the chunk-optimizer merge/fold/wrap sites for the same "eager cross-entry execution" family and verified (by running probes) that the **strict** path is otherwise closed: under `strictExecutionOrder` every side-effectful module is wrapped (including under `onDemandWrapping`, since an observable side effect makes a module execution-order-sensitive), so the only eager leak was the entry-chunk fold fixed here. Dynamic→group / dynamic→common / dynamic→user-entry / static-via-intermediate / plain-group shapes were all checked and are isolated under strict. Full `rolldown` integration suite: no snapshot regressions; `#5726` and `strict_execution_order/on_demand_wrapping` unchanged.

## Out of scope / follow-up

This addresses the `strictExecutionOrder` case from the issue. The same logical leak still exists **without** `strictExecutionOrder` when a manual code-splitting group co-locates modules of different entry reachability: there the shared chunk's modules are eager (unwrapped), so loading one entry runs the co-located code of another. For non-entry modules this is the pre-existing manual-group over-inclusion tradeoff (see `advanced_chunks/entries_aware_merge_shared_deps`); for entry modules it is a genuine isolation gap. The structural cause is manual code splitting producing chunks whose `bits` is a union across modules of different reachability. Tracked in #9998; a complete fix would make manual splitting respect reachability boundaries (or keep cross-entry modules wrapped in non-strict mode too).

Closes #9463

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -813,12 +813,33 @@ impl GenerateStage<'_> {
 
       if meta.intersects(ChunkMeta::UserDefinedEntry) {
         if matches!(target_chunk.kind, ChunkKind::Common) {
-          let can_merge = match chunk.preserve_entry_signature {
-            Some(PreserveEntrySignatures::Strict) => {
-              self.can_merge_without_changing_entry_signature(chunk, &target_chunk.modules)
-            }
-            _ => true,
-          };
+          // Execution-isolation guard (issue #9463).
+          //
+          // Folding the common chunk *into* this user-defined entry chunk makes the
+          // entry chunk eagerly run this entry's top-level (its `init_*` call)
+          // whenever the chunk is loaded. If the common chunk also holds *another*
+          // user-defined entry's module, that other entry would be forced to import
+          // this entry chunk just to reach its own module — and would then run this
+          // entry's side effects. e.g. loading entry `b` would trigger entry `a`'s
+          // side effects. This happens when manual code splitting (a `codeSplitting`
+          // group, possibly via `entriesAware` subgroup merging) lumps several
+          // entries' modules into one shared chunk.
+          //
+          // Keep the thin facade in that case, so each entry imports the (wrapped)
+          // shared chunk and runs only its own `init_*`. A shared chunk that holds
+          // just this entry's module (plus non-entry deps a sibling genuinely
+          // depends on) is still folded in, preserving the #5726 facade-elimination.
+          let holds_other_user_entry = target_chunk.modules.iter().any(|&module_idx| {
+            module_idx != module
+              && self.link_output.user_defined_entry_modules.contains(&module_idx)
+          });
+          let can_merge = !holds_other_user_entry
+            && match chunk.preserve_entry_signature {
+              Some(PreserveEntrySignatures::Strict) => {
+                self.can_merge_without_changing_entry_signature(chunk, &target_chunk.modules)
+              }
+              _ => true,
+            };
           if can_merge {
             // merge all common chunk modules into entry chunk
             // swap original from_chunk_idx and target_chunk_idx
--- a/crates/rolldown/tests/rolldown/issues/9463/_config.json
+++ b/crates/rolldown/tests/rolldown/issues/9463/_config.json
@@ -1,0 +1,22 @@
+{
+  "_comment": "https://github.com/rolldown/rolldown/issues/9463 - With `strictExecutionOrder`, a `codeSplitting` group (here `entriesAware` + a large `entriesAwareMergeThreshold`) lumps both entry modules and the shared module into one chunk. `chunkOptimization` then folded that chunk into entry `a`, so executing entry `b` (`import \"./a.js\"; init_b()`) eagerly ran `init_a()` and triggered entry `a`'s side effect. Each entry must run only its own `init_*`.",
+  "config": {
+    "input": [
+      { "name": "a", "import": "./a.js" },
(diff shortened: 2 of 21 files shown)


IWANABETHATGUY (approved):

hyfdev:
### Merge activity

* **Jun 28, 1:18 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 28, 1:18 PM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 28, 1:23 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10009, opened 2026-06-29
fix: preserve included external import declarations

chuganzy:
This PR aims to fix a case where the finalizer removed an included external import declaration.

The finalizer handles both `import` declarations and `export ... from` declarations. 
Previously, when an import record did not resolve to an internal module, the finalizer removed the statement. That is not safe for an included external import declaration: the imported bindings may still be referenced by code that survived tree-shaking.

This showed up with `strictExecutionOrder` because module bodies can be moved into `__esmMin` wrappers. The wrapper body could still contain code like:

```js
value = first(second);
```

while the corresponding top-level import had been removed:

```js
import { first, second } from "dep";
```

This keeps included unresolved/external `ImportDeclaration`s in the output, while leaving unresolved `export ... from` handling unchanged.

### DIFF
--- a/crates/rolldown/src/module_finalizers/mod.rs
+++ b/crates/rolldown/src/module_finalizers/mod.rs
@@ -319,7 +319,9 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
     rec_idx: ImportRecordIdx,
   ) -> bool {
     let rec = &self.ctx.module.import_records[rec_idx];
-    let Some(resolved_module_idx) = rec.resolved_module else { return true };
+    let Some(resolved_module_idx) = rec.resolved_module else {
+      return !stmt.is_import_declaration();
+    };
     let Module::Normal(importee) = &self.ctx.modules[resolved_module_idx] else {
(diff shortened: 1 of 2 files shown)


IWANABETHATGUY:
Could you please file an issue with a minimal reproduction first? I’m not sure what you’re trying to fix. We’ve updated our CONTRIBUTING.md, https://github.com/rolldown/rolldown/blob/main/docs/contribution-guide/index.md, and in the future we may close such PRs.

chuganzy:
@IWANABETHATGUY Apologies for not creating the issue first and for not providing enough context in the PR description.

I filed a minimal issue here: https://github.com/rolldown/rolldown/issues/10013

The original failure was found while bundling Storybook. I used Codex to help reduce that failure into a minimal reproduction, but I reviewed and adjusted the test case and fix by hand.

The reduced case shows that with `strictExecutionOrder: true`, Rolldown can keep a lazy `__esmMin` wrapper that still references external bindings while dropping the corresponding `ImportDeclaration`, leaving free identifiers in the generated chunk.

PR #10021, opened 2026-06-30
feat(experimental): backport minChunkSize small-leaf chunk merging onto rc.18

hyfdev:
## What

Backports `experimental.minChunkSize` (the small side-effect-free leaf chunk
merging from #10010 / #9949) onto the **`v1.0.0-rc.18`** release commit
(`2c4a957`), so it can ship as a preview build for a downstream consumer pinned
to rc.18 (vite 8.0.11 pins `rolldown@1.0.0-rc.18`). The 1.1.x line that carries
the original PR produces invalid output (`i0 is not defined` dangling-namespace
bug) for that consumer, so this PR delivers the feature on the known-good rc.18
base instead.

Base of this PR is a branch pinned at `v1.0.0-rc.18`, so the diff is exactly the
backport (not the 438 commits between rc.18 and main).

## How

- New pass `crates/rolldown/src/stages/generate_stage/min_chunk_size.rs`
  (duplicate small ESM leaf chunks into importers; ESM-only; shadow-global guard).
- `chunk_graph`: `duplicated_leaf_modules` + `duplicated_leaf_pinned_names`.
- `compute_cross_chunk_links`: treat duplicated leaves as chunk-local.
- `deconflict_chunk_symbols`: pin duplicated-leaf names before deconfliction.
- `renamer`: `pin_name`.
- Option plumbing (rust binding + TS input-options / validator / bindingify).

### rc.18 adaptations vs the original PR
- `renamer.pin_name` reserves via `self.reserve` (rc.18 has no `resolver` field).
- `collect_declared_symbols` reads `stmt_infos` from the module (rc.18 keeps them
  on the module, not on `LinkStageOutput`) and matches `TaggedSymbolRef::Normal`
  inline (rc.18 has no `is_normal()`).
- The original PR's `module_finalizers` edit is **omitted**: it targets a
  wrapped-esm-init fallback that does not exist in rc.18, and rc.18's ESM symbol
  resolution already emits local canonical names (the cross-chunk path is
  CJS-only), so a duplicated leaf resolves locally without it.

## Validation
- `cargo check` clean; release napi binding built.
- Feature: 2 entries sharing a tiny leaf → off = 3 chunks (standalone shared
  leaf), on (`minChunkSize: 100000`) = 2 chunks, leaf duplicated into both, 0
  cross-chunk imports; emitted output imports and executes correctly.

### DIFF
--- a/crates/rolldown/src/chunk_graph.rs
+++ b/crates/rolldown/src/chunk_graph.rs
@@ -1,6 +1,7 @@
 use arcstr::ArcStr;
 use itertools::Itertools;
 use oxc_index::{IndexVec, index_vec};
+use oxc_str::CompactStr;
 use rolldown_common::{
   Chunk, ChunkIdx, ChunkModulesOrderBy, ChunkTable, EcmaViewMeta, ModuleIdx,
   PostChunkOptimizationOperation, RuntimeHelper, SymbolRef,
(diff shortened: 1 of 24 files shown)


hyfdev:
Closing in favor of #10187. The diff is intentionally the same, but the fresh PR gives downstream consumers a new pkg.pr.new URL so pnpm does not reuse the old mutable @10021 tarball cache.

PR #10023, opened 2026-06-30
fix: keep fragments after the newline fragment in MagicString::last_line

shulaoda:
### What this solves

`MagicString::last_line` (exposed to JS as `lastLine()`) returns the content after the final newline of the generated string. Each `intro`/`outro` is a sequence of fragments, and the function scanned a deque from the back and, on the first fragment (from the end) that contained a newline, returned that fragment's tail. It never included the fragments it had already visited in that reverse scan, which come *after* the newline-bearing fragment in output order, so they were silently dropped. The same mistake was in the global-outro, per-chunk-outro and per-chunk-intro scans.

For example, appending `"a\nb"` then `"c"` produces `"a\nbc"`, but `lastLine()` returned `"b"` instead of `"bc"`.

### The fix

Scan fragments from the back, pushing each fragment's relevant slice onto a buffer and stopping at the first newline that completes the last line. Sections (outro, chunks last-to-first, then intro) are visited only until that newline is found, so earlier sections are never touched, and the result is built once with an exact-capacity allocation. This matches the reference `magic-string`, which searches the concatenated outro/intro as a whole.

### Tests

The regression tests live on the JS side next to the other hand-written magic-string cases (`packages/rolldown/tests/magic-string/rolldown-magic-string.test.ts`), since `last_line` is only reached through the `lastLine()` binding. Two `lastLine` guards cover a newline in a non-final outro fragment and a non-final intro fragment; both fail on the base (`bc` / `qrx` are dropped) and pass with the fix.


IWANABETHATGUY:
Fix looks good — the reverse-scan + `pieces` buffer correctly keeps the fragments after the newline-bearing one, and it's a bit leaner than the old per-chunk `String` rebuilds. Two things on the tests, though:

**1. `joins_content_tail_with_trailing_outro_fragments` doesn't actually guard the regression — it passes on the base too.**

In that case the `\n` lives in the chunk *content*, not in a non-final intro/outro fragment:

```rust
let mut s = MagicString::new("a\nb");
s.append("c");
s.append("d");
assert_eq!(s.last_line(), "bcd");
```

The old code already handled this — its content branch did `return content[after_nl..] + line_str`, and `line_str` had accumulated `"cd"` from the trailing outro fragments, so it returned `"bcd"` on `main`. I confirmed by running the equivalent through `lastLine()` on `main`: it passes unfixed. So the description's "they fail on the old code" isn't true for this one — it's a redundant test. The other two (`keeps_outro_fragments_after_a_newline_fragment` / `keeps_intro_fragments_after_a_newline_fragment`) are the real guards. Suggest dropping it or swapping it for a case that genuinely exercises the bug.

**2. Could we put these regression tests on the JS side instead?**

`MagicString::last_line` is only ever reached through the `lastLine()` napi binding, and we already keep the hand-written magic-string regression tests in `packages/rolldown/tests/magic-string/rolldown-magic-string.test.ts`. Keeping these next to the rest of the custom cases there (rather than in the crate) matches where the API is actually consumed and where reviewers will look for them. Something like:

```ts
describe('lastLine - keeps fragments after the newline fragment (#10023)', () => {
  it('keeps outro fragments appended after a newline fragment', () => {
    const s = new MagicString('x');
    s.append('pre');
    s.append('a\nb');
    s.append('c');
    assert.strictEqual(s.toString(), 'xprea\nbc');
    assert.strictEqual(s.lastLine(), 'bc');
  });

  it('keeps intro fragments prepended after a newline fragment', () => {
    const s = new MagicString('x');
    s.prepend('r');
    s.prepend('p\nq');
    s.prepend('pre');
    assert.strictEqual(s.toString(), 'prep\nqrx');
    assert.strictEqual(s.lastLine(), 'qrx');
  });
});
```

IWANABETHATGUY (approved):

PR #10025, opened 2026-06-30
fix: reset the word-boundary state on newline in Hires::Boundary sourcemaps

shulaoda:
### What this solves

In `Hires::Boundary` mode the sourcemap builder tracks `char_in_hires_boundary` so that only the first character of each word run emits a mapping token. A `'\n'` is handled in its own match arm that bumps the line but never clears that flag. So when a source line ends with a word character (identifier, number, keyword), the flag stays set across the newline, the first word character of the next line takes the "already inside a word" branch and is skipped, and that line gets no start-of-line mapping. For example `const a = 1\nconst b = 2` left the second line's leading word unmapped.

`Hires::Boundary` is used widely (plugin transform sourcemaps, `generateMap({ hires: "boundary" })`, and the asset-module / copy-module / vite-dynamic-import-vars / vite-import-glob builtins), so this degrades sourcemaps for common no-semicolon and identifier-terminated lines. The reference magic-string runs `'\n'` through the same boundary logic, which resets the flag, so it does not lose the mapping.

### The fix

Reset `char_in_hires_boundary` to `false` in the `'\n'` arm so a newline ends the current word run and the first word character of the next line still emits its boundary token.

### Tests

Added `hires_boundary_maps_word_at_start_of_next_line` in `crates/string_wizard/tests/magic_string_source_map.rs`: for a source whose first line ends with a word char and whose second line starts with one, it asserts a token maps the start of the second generated line back to the second source line. It fails on the old code and passes with the fix. The existing `Hires::Boundary` snapshot is unchanged because its lines end with non-word characters.

### DIFF
--- a/crates/rolldown/tests/rolldown/function/module_types/copy/basic/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/function/module_types/copy/basic/artifacts.snap
@@ -29,6 +29,7 @@ console.log(data, data2);
 
 ```
 - ../main.js
+(1:0) "const " --> (3:0) "const "
 (1:6) "data2 = require(" --> (3:6) "data2 = ("
 (1:6) "data2 = require(" --> (3:15) "/* @__PURE__ */ ((x) => typeof require !== \"undefined\" ? require : typeof Proxy !== \"undefined\" ? new Proxy(x, { get: (a, b) => (typeof require !== \"undefined\" ? require : a)[b] }) : x)(function(x) {\n"
 (1:6) "data2 = require(" --> (6:1) "))"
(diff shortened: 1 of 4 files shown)


hyfdev (approved):

hyfdev:
### Merge activity

* **Jul 2, 3:47 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 2, 3:47 AM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 2, 3:51 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10029, opened 2026-06-30
fix: trim an emptied chunk's outro/intro instead of skipping past it

shulaoda:
### What this solves

`trim_start` / `trim_end` walk the chunks and stop at the first non-whitespace. Each chunk's output order is intro, content, outro. When a chunk's content is already empty (for example after `remove()`), the start trimmer `continue`d to the next chunk without trimming or inspecting that chunk's outro (and the end trimmer symmetrically skipped the intro). The outro comes right after the empty content in output order, so if it holds non-whitespace the trim should abort there. Because it didn't, trimming proceeded into the following chunk and deleted whitespace that is not leading (resp. trailing). For example, `"A B"` with the first character removed and `"X"` appended into the emptied chunk's outro renders as `"X B"`, so `trim_start` should be a no-op, but it returned `"XB"`. The reference magic-string's `Chunk.trimStart` trims the outro and aborts when it is non-empty even if the content is empty.

### The fix

In the empty-content branch of `trim_start_aborted` (and symmetrically in `trim_end_aborted`), trim the chunk's outro (resp. intro) and abort if non-whitespace remains, mirroring the path already taken when the content is trimmed down to empty.

### Tests

Added a `trim` test module with two cases: an emptied chunk whose outro precedes following whitespace (`trim_start`) and whose intro follows preceding whitespace (`trim_end`); both assert the trim is a no-op. They fail on the old code (`"XB"` / `"AX"`) and pass with the fix, and the expected `"X B"` / `"A X"` outputs match magic-string.

### DIFF
--- a/crates/string_wizard/src/magic_string/trim.rs
+++ b/crates/string_wizard/src/magic_string/trim.rs
@@ -60,7 +60,11 @@ impl<'text> MagicString<'text> {
       let trimmed_content = trim_start_pattern(&content, pattern);
 
       if content.is_empty() {
-        // Content is empty (e.g., from a remove), continue to next chunk
+        // Content is empty (e.g. from a remove); the outro is next in output order, so trim
+        // it and abort if non-whitespace remains.
+        if trim_deque_start(&mut chunk.outro, pattern) {
+          return true;
+        }
         chunk_idx = next_idx;
         continue;
       }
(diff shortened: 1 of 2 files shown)


IWANABETHATGUY (approved):

shulaoda:
### Merge activity

* **Jul 2, 2:31 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 2, 2:31 AM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 2, 2:35 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10032, opened 2026-06-30
fix: check finalize_other_specifiers in its own Debug attribute

shulaoda:
### What this solves

The `#[debug(...)]` attribute on `BindingViteResolvePluginConfig::finalize_other_specifiers` tested `finalize_bare_specifier.is_some()` (the previous field, copy-pasted) instead of `finalize_other_specifiers.is_some()`. So when the config is `Debug`-formatted (logging/tracing), the `finalize_other_specifiers` field reflects whether the *bare-specifier* callback is set, not its own: it prints `Some(<finalize_other_specifiers>)` while actually being `None`, and `None` while actually being set.

### The fix

Reference `finalize_other_specifiers` in its own debug condition, matching the sibling `finalize_bare_specifier` field.


shulaoda:
Small change, so merge it directly

PR #10033, opened 2026-06-30
chore: remove the unused ScheduledBuild napi struct

shulaoda:
### What this solves

`ScheduledBuild` (a `#[napi]` struct with `wait` / `already_scheduled`) is never constructed: no Rust code builds or returns it, and it has no `#[napi(constructor)]`, so JS cannot instantiate it either. napi-rs does not even emit it into the generated bindings because it never appears in the public API surface, so it is unreachable dead code.

### The change

Remove the struct and its impl, and drop the `BundlingFuture` import that it was the only user of. Rebuilding leaves the generated bindings unchanged and `tsc` still passes, confirming there was no JS-side surface to break.

### DIFF
--- a/crates/rolldown_binding/src/binding_dev_engine.rs
+++ b/crates/rolldown_binding/src/binding_dev_engine.rs
@@ -1,6 +1,6 @@
 use napi_derive::napi;
 use rolldown_dev::{
-  BundleState, BundlingFuture, OnAdditionalAssetsCallback, OnHmrUpdatesCallback, OnOutputCallback,
+  BundleState, OnAdditionalAssetsCallback, OnHmrUpdatesCallback, OnOutputCallback,
 };
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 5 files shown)


hyfdev (approved):

hyfdev:
### Merge activity

* **Jun 30, 2:37 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jun 30, 2:38 PM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jun 30, 2:49 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10034, opened 2026-06-30
chore: remove the unused BindingGenerateHmrPatchReturn napi type

shulaoda:
### What this solves

`BindingGenerateHmrPatchReturn` (a `#[napi]` enum with `Ok` / `Error` variants and a `from_errors` constructor) is never used: no Rust code constructs or returns it, and napi-rs does not emit it into the generated bindings because it never appears in the API surface. It is unreachable dead code.

### The change

Delete the file and its `pub mod` declaration in `types/mod.rs` (there are no other references or re-exports). Rebuilding leaves the generated bindings unchanged and the crate still builds clean under `clippy --deny warnings`.


shulaoda:

PR #10036, opened 2026-06-30
chore: remove the dead output_assets vector in render_chunk_to_assets

shulaoda:
### What this solves

In `render_chunk_to_assets`, `output_assets` is declared, sorted, and extended into `output`, but nothing ever pushes to it: the loop pushes every chunk and asset into `output` directly. So it is always empty, and its sort and `output.extend(output_assets)` are both no-ops. Asset ordering is already handled by the subsequent `output.sort_unstable_by` (file type, then filename).

### The change

Remove the `output_assets` declaration, its no-op sort (with the stale comment), and the no-op extend. Output is byte-identical: the full integration suite passes with zero snapshot drift.


shulaoda:

PR #10038, opened 2026-06-30
chore: drop the redundant visited.insert in manual code splitting

shulaoda:
### What this solves

In `add_module_and_dependencies_to_group_recursively`, `module_idx` is already inserted into `visited` at the top (`let is_visited = !visited.insert(module_idx)`), and the early returns in between never remove it. The second `visited.insert(module_idx)` later in the function therefore always re-inserts an element that is already present, a no-op.

### The change

Remove the redundant insert. Behavior is unchanged: the crate builds clean under `clippy --deny warnings` and the full integration suite passes with zero snapshot drift.


shulaoda:

PR #10049, opened 2026-06-30
refactor: drop redundant program scope enter/leave in finalizer

shulaoda:
### What

In `ScopeHoistingFinalizer::visit_program`, removes the manual `enter_scope` / `leave_scope` pair wrapped around `walk_mut::walk_program`, plus the now-unused `semantic::ScopeFlags` import.

### Why

`walk_mut::walk_program` already calls `enter_scope` and `leave_scope` internally, using the exact same scope-flag computation (`ScopeFlags::Top`, with `StrictMode` added when the program is strict or has a `"use strict"` directive) and the same `scope_id`. The manual wrapper duplicated that, so `enter_scope` / `leave_scope` ran twice per program.

The duplication was harmless because the finalizer's `enter_scope` / `leave_scope` only derive two boolean states (`TopLevel`, `IsRootLevel`) via `scope_stack.iter().rev().all(...)` predicates, and pushing the same `Top` flag twice does not change an `.all()` result. Nothing reads the stack depth or length. Removing the redundant wrapper keeps behavior identical while dropping dead nesting.

### DIFF
--- a/crates/rolldown/src/module_finalizers/impl_visit_mut.rs
+++ b/crates/rolldown/src/module_finalizers/impl_visit_mut.rs
@@ -9,7 +9,6 @@ use oxc::{
     match_member_expression,
   },
   ast_visit::{VisitMut, walk_mut},
-  semantic::ScopeFlags,
   span::{SPAN, Span},
 };
 use oxc_str::CompactStr;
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY (approved):

shulaoda:
### Merge activity

* **Jul 1, 2:51 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 1, 2:51 AM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 1, 2:59 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10051, opened 2026-06-30
chore: remove commented-out re_export_all_names field

shulaoda:
### What

Removes the commented-out `// pub re_export_all_names: FxHashSet<CompactStr>,` field from `LinkingMetadata`.

### Why

The field is fully commented out and has no other references anywhere in the workspace, so it is dead. Removing the stale line keeps the struct definition clean. `FxHashSet` is still imported and used by another field, so the import is untouched.


shulaoda:

PR #10052, opened 2026-06-30
chore: remove unused ExpressionExt methods

shulaoda:
### What

Removes four methods from the `ExpressionExt` trait (declaration plus impl) in `rolldown_ecmascript_utils`:

- `as_call_expression_mut`
- `as_unary_expression`
- `as_binary_expression`
- `as_static_member_expr_mut`

### Why

None of these four methods have any callers anywhere in the workspace. `ExpressionExt` is an internal utility trait, so the unused entries are dead code. The remaining methods (`as_call_expression`, `as_identifier`, `as_string_literal`, `as_static_module_request`, the `is_import_meta*` checks, and so on) are still used, so the trait and its imports stay. The crate and its main consumer both compile cleanly after the removal.


shulaoda:

PR #10053, opened 2026-06-30
chore: remove unused StatementExt methods

shulaoda:
### What

Removes four methods from the `StatementExt` trait (declaration plus impl) in `rolldown_ecmascript_utils`:

- `is_import_declaration`
- `as_export_named_declaration`
- `as_function_declaration`
- `is_function_declaration`

### Why

None of these four methods have any callers in the workspace. `as_function_declaration` was only called by `is_function_declaration`, which itself is unused, so both go together. `StatementExt` is an internal utility trait, so the unused entries are dead code. The remaining methods (`as_import_declaration`, `as_export_default_declaration_mut`, `as_export_all_declaration`, `as_export_named_declaration_mut`, `is_module_declaration_with_source`) are still used, so the trait and its import stay. The crate and its main consumer both compile cleanly after the removal.


shulaoda:

PR #10054, opened 2026-06-30
chore: remove dead Diagnostic::with_kind method

shulaoda:
### What

Removes the `Diagnostic::with_kind` builder method from `rolldown_error`.

### Why

The method has no callers anywhere in the workspace. `rolldown_error` is an internal crate, so this dead `pub fn` is just unused surface. Removing it keeps the `Diagnostic` API to what is actually used.


shulaoda:

PR #10055, opened 2026-06-30
chore: remove unused From impl for RolldownLabelSpan

shulaoda:
### What

Removes the `From<(DiagnosticFileId, Range<usize>)>` impl for `RolldownLabelSpan` in `rolldown_error`.

### Why

The impl has no callers. Both construction sites for `RolldownLabelSpan` use the tuple-struct syntax `RolldownLabelSpan(id, range)` directly, and nothing in the workspace converts a `(DiagnosticFileId, Range<usize>)` tuple via `.into()` or `RolldownLabelSpan::from(..)`. The `Range` import is still used by the struct definition, so it stays.


shulaoda:

Issue #10020, opened 2026-06-30
preserveModules emits undeclared named exports for default-imported JSON modules (1.1.3 regression)


fengmk2:
Under `preserveModules`, a default-imported JSON module is emitted with a named export for each top-level key, but those bindings are never declared, so loading the output fails at ESM link time.

Minimal repro: https://github.com/why-reproductions-are-required/rolldown-preserve-modules-json-repro

```js
// src/index.js
import metrics from "./data.json" with { type: "json" }; // data.json: { "aBeeZee": [...], "Abel": [...] }
export function get(name) { return metrics[name]; }
```

Build with `preserveModules: true`, then `import("./dist/index.js")`:

```
SyntaxError: Export 'Abel' is not defined in module
```

The emitted `dist/data.js` differs from 1.1.2 only in the export line:

```diff
-export { data_default as default };                 // 1.1.2 (works)
+export { Abel, aBeeZee, data_default as default };  // 1.1.3 (Abel/aBeeZee never declared)
```

`Abel`/`aBeeZee` exist only as properties of `data_default`, not as bindings, and aren't used by the importer (it reads the default export dynamically), yet they're emitted and break linking.

**Versions:** rolldown 1.1.3 (1.1.2 works) · Node v24.18.0

**Likely cause:** #9934 ("preserve used re-exports under preserveModules"), which emits each preserved module's full declared interface; for a JSON module the per-key "interface" has no backing bindings.

**Real-world impact:** breaks building [cloudflare/vinext](https://github.com/cloudflare/vinext) ([`fallback-metrics.ts`](https://github.com/cloudflare/vinext/blob/main/packages/vinext/src/build/google-fonts/fallback-metrics.ts#L8) default-imports `fallback-metrics-data.json`); consumers' `vite.config.ts` then crash with `Export 'aBeeZee' is not defined in module`.

PR #9934, opened 2026-06-23
fix: preserve used re-exports under preserveModules (#9122)

IWANABETHATGUY:
## What

Fixes #9122. Under `preserveModules`, a module that re-exports an imported binding (`export { X } from './y'`) was dropping `X` from its own emitted file's export list when `X`'s canonical home was another module. Canonical resolution makes consumers bind the canonical symbol directly and bypass the facade, so the facade was tree-shaken out of this module's interface even though it was still consumed *through* it.

## How

- **`compute_cross_chunk_links`**: emit each preserved user module's full declared interface; the `used_symbol_refs` filter keeps only the ones actually used.
- **`include_statements`**: add `preserve_reexported_interfaces`, a post-pass run after the inclusion fixpoint. It re-marks a re-export facade as used (and includes its re-export statement so the cross-chunk import is generated) only when its canonical symbol survived tree-shaking. The pass is made **chain-granular** via `normal_symbol_exports_chain_map`: a facade is preserved only when it lies on the export chain of a *used* import — i.e. some module imports it *through* this module. This avoids resurrecting a side-effect-only `wrapper` that re-exports `foo` from `a` when a consumer imports `foo` straight from `a`.
- The synthetic runtime module is excluded so its helpers stay demand-driven.

Genuinely-unused exports stay tree-shaken, matching Rollup.

## Tests

- `issues/9122` fixture: entry imports `StateCode`/`getX` through `wrapper` (`export { … } from './wrapper.js'`), so `wrapper` correctly keeps them.
- `preserve_modules/shake_unrelated_reexport` fixture: covers the dropped case — an unrelated re-export facade is not resurrected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10056, opened 2026-07-01
fix(generate): keep full JSON interface under preserveModules namespa…

IWANABETHATGUY:
### What

Builds on the parent PR (#10020). A JSON module imported as a namespace (`import * as ns from './x.json'`) keeps every top-level key materialized — the finalizer only inlines JSON keys when the module namespace object is **not** included (`need_inline_json_prop`, see `finalizer_context.rs`).

> **This is pre-existing behavior, not a regression.** On `main` a namespace-imported JSON chunk under `preserveModules` already exports its full interface correctly — `import('./dist/data.js')` exposes every top-level key. Unlike #10020 (which *is* a #9934 regression), there is no namespace bug on `main`. I verified the guard test passes on `main`.

The reason this PR exists is purely to keep the stack consistent: the parent PR's #10020 narrowing (skip keys absent from `json_module_none_self_reference_included_symbol`) is too coarse for the namespace case — those keys are absent from that set yet still materialized, so the **parent PR alone** would drop them:

```js
// parent PR (#10020) alone: only the namespace object survives
export { data_exports };
// with this PR: pre-existing full interface preserved
export { answer, data_exports, data_default as default, nested };
```

### Fix

Mirror the finalizer's **full** inlining condition rather than the key-set alone: only drop a key when JSON property inlining is actually active — JSON, ESM exports, and the namespace object **not** included. The stack's net behavior then matches `main` for the namespace case.

### Test

`crates/rolldown/tests/rolldown/misc/preserve_modules/json_namespace_preserves_exports`: `import * as` a JSON module under preserveModules and assert (via `_test.mjs`) that `import('./dist/data.js')` re-exposes every top-level key plus `default`. Verified: passes on `main`, fails on the parent PR alone, passes here.

---

> Stacked on the #10020 fix.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
+++ b/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
@@ -438,18 +438,25 @@ impl GenerateStage<'_> {
               .map(|(name, export)| (name, export.symbol_ref))
             {
               // `preserveModules` emits a module's complete declared interface (#9934). A JSON
-              // module synthesizes a named export per top-level key, but a key that is reached
-              // only through its own default-export object has its `var` binding folded into that
-              // object by the finalizer's `try_inline_json_module_prop`, leaving no standalone
-              // declaration. Listing such a key here produces an `export { key }` with no binding
-              // -> `SyntaxError: Export 'x' is not defined in module` (#10020).
+              // module synthesizes a named export per top-level key, but the finalizer
+              // (`try_inline_json_module_prop`) may fold a key's `var` binding into the
+              // self-contained default-export object, leaving no standalone declaration. Listing
+              // such a key here produces an `export { key }` with no binding ->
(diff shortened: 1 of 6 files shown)


shulaoda (approved):

IWANABETHATGUY:
### Merge activity

* **Jul 1, 2:49 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 1, 2:49 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 1, 2:54 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10060, opened 2026-07-01
fix(dev): disable lazy barrel in dev mode

shulaoda:
## What is this PR solving?

Dev mode (`experimental.devMode`) requires treeshaking to be disabled. We already force `treeshake` off when `experimental.dev_mode.is_some()` in `prepare_build_context`.

The `lazy_barrel` optimization is a form of treeshaking for barrel (re-export) files: it lazily tracks and prunes re-exports based on what is actually used. It is enabled by default (`is_lazy_barrel_enabled()` returns `true` when the option is `None`). This means that in dev mode, treeshaking is off but lazy barrel is still on, which is inconsistent and can lead to incorrect behavior.

This PR forces `lazy_barrel` off whenever dev mode is enabled. While doing so, the two separate `experimental.dev_mode.is_some()` checks (one forcing `incremental_build` on, the other forcing `treeshake` off) are merged into a single block, so all dev-mode overrides live in one place.

## Changes

`crates/rolldown/src/utils/prepare_build_context.rs`:

```rust
let mut raw_treeshake = raw_options.treeshake;
let mut experimental = raw_options.experimental.unwrap_or_default();
if experimental.dev_mode.is_some() {
  experimental.incremental_build = Some(true);
  // Dev mode requires treeshaking to be disabled, and lazy barrel relies on
  // treeshaking, so it must be disabled as well.
  raw_treeshake = TreeshakeOptions::Boolean(false);
  experimental.lazy_barrel = Some(false);
}
```

Because `is_lazy_barrel_enabled()` reads `self.lazy_barrel.unwrap_or(true)`, setting it to `Some(false)` forces the feature off, overriding the default-on behavior. All downstream call sites (`module_task`, `module_loader`, `flat_options`) go through `is_lazy_barrel_enabled()`, so this single override turns the feature off everywhere for dev mode.

## Why do it here?

This block already normalizes dev-mode-specific overrides, and it runs before `flat_options` and the module loader read the flag, so the override takes effect for the whole build. Keeping the `treeshake`, `incremental_build`, and `lazy_barrel` decisions in one obvious block makes the shared reason explicit. `raw_treeshake` is declared just above the block (it only depends on `raw_options.treeshake`), so pulling it up to merge the two checks is safe.

### DIFF
--- a/crates/rolldown/src/utils/prepare_build_context.rs
+++ b/crates/rolldown/src/utils/prepare_build_context.rs
@@ -272,9 +272,14 @@ pub fn prepare_build_context(
       .unwrap_or_default(),
   );
 
+  let mut raw_treeshake = raw_options.treeshake;
   let mut experimental = raw_options.experimental.unwrap_or_default();
   if experimental.dev_mode.is_some() {
     experimental.incremental_build = Some(true);
+    // Dev mode requires treeshaking to be disabled, and lazy barrel relies on
+    // treeshaking, so it must be disabled as well.
+    raw_treeshake = TreeshakeOptions::Boolean(false);
+    experimental.lazy_barrel = Some(false);
   }
 
   if experimental.attach_debug_info.is_none() {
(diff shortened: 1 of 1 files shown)


h-a-n-a (approved):

h-a-n-a:
### Merge activity

* **Jul 1, 6:41 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 1, 6:41 AM UTC**: `h-a-n-a` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 1, 6:47 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

h-a-n-a:
Context: 

This is caused by inconsistent tree shaking in FBM. In the future, we can investigate whether the ROI for enabling barrel optimization is high enough for us to reenable it again. If so, then we can start thinking of whether it's implemented based on [one-pipeline module](https://gist.github.com/h-a-n-a/f0f70070f0d2ef2f1ea158ffa5073eb4#file-rolldown-dev-hmr-asset-resolution-gap-md) (making HMR and normal build in the same pipeline) or implemented based on the current standalone HMR architecture.

Issue #7597, opened 2025-12-19
[Bug]: panic in v1.0.0-beta.55 with react-bootstrap and --minify


dumbmatter:
### Reproduction link or steps

Minimal example: https://github.com/dumbmatter/rolldown-react-bootstrap-minify-bug

It's this code:

```
import { Modal } from "react-bootstrap";
window.Modal = Modal;
```

And this command:

```
$ rolldown index.jsx --minify --file bundle.js
```

This works if you downgrade rolldown to v1.0.0-beta.54, or if you remove the --minify option. But in  v1.0.0-beta.55 it produces an error.

### What is expected?

It should run without error.

### What is actually happening?

```
$ pnpm build

> rolldown-react-bootstrap-bug@1.0.0 build /home/user/rolldown-react-bootstrap-bug
> rolldown index.jsx --minify --file bundle.js


thread '<unnamed>' (1139236) panicked at crates/rolldown/src/module_finalizers/mod.rs:104:7:
canonical name not found for SymbolRef { owner: 34, symbol: SymbolId(4294967292) }, original_name: "import_prop_types" in module "node_modules/.pnpm/react-bootstrap@2.10.10_@types+react@19.2.7_react-dom@19.2.3_react@19.2.3__react@19.2.3/node_modules/react-bootstrap/esm/CloseButton.js" when finalizing module "node_modules/.pnpm/react-bootstrap@2.10.10_@types+react@19.2.7_react-dom@19.2.3_react@19.2.3__react@19.2.3/node_modules/react-bootstrap/esm/CloseButton.js" in chunk ""
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

 ERROR  Panic in async function                                                                                                                                                             

  

 ELIFECYCLE  Command failed with exit code 1.
```

### System Info

```Shell
System:
    OS: Linux 6.8 Ubuntu 24.04.3 LTS 24.04.3 LTS (Noble Numbat)
    CPU: (8) x64 Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
    Memory: 15.05 GB / 31.13 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 24.12.0 - /usr/bin/node
    npm: 11.6.2 - /usr/bin/npm
    pnpm: 10.25.0 - /home/user/.local/share/pnpm/pnpm
  Browsers:
    Chrome: 143.0.7499.146
    Firefox: 146.0
[...]

PR #10080, opened 2026-07-02
fix: gate sideEffects:false modules' side effects on body demand

### LINKED ISSUE #9691: Missing import when lazy barrel is enabled

### LINKED ISSUE #9806: [Bug]: lazyBarrel emits undefined bindings for unused barrel re-export initializers

Create this minimal package:

```json
{
  "type": "module",
  "scripts": {
    "build": "rolldown -c",
    "start": "node dist/out.mjs"
  },
  "dependencies": {
    "mnemonist": "0.40.4",
    "rolldown": "1.1.0"
  }
}
```

`entry.mjs`:

```js
import { LRUCacheWithDelete } from 'mnemonist';

const cache = new LRUCacheWithDelete(2);
cache.set('a', 1);

console.log(cache.get('a'));
```

`rolldown.config.mjs`:

```js
export default {
  input: 'entry.mjs',
  output: {
    file: 'dist/out.mjs',
    format: 'esm'
  }
};
```

Run:

```sh
npm install
npm run build
npm run start
```

### What is expected?

The bundled output should execute successfully and print:

```txt
1
```

The generated bundle throws at module evaluation time:

```txt
file:///private/tmp/rolldown-mnemonist-repro/dist/out.mjs:845
FibonacciHeap.MinFibonacciHeap;
^

ReferenceError: FibonacciHeap is not defined
    at file:///private/tmp/rolldown-mnemonist-repro/dist/out.mjs:845:1
```

The emitted bundle contains property reads for unused barrel re-export-derived bindings, but no bindings for the imported names:

```js
module.exports = LRUCacheWithDelete;
})))(), 1);
FibonacciHeap.MinFibonacciHeap;
FibonacciHeap.MaxFibonacciHeap;
Heap.MinHeap;
Heap.MaxHeap;
SuffixArray.GeneralizedSuffixArray;
Vector.Uint8Vector;
Vector.Uint8ClampedVector;
Vector.Int8Vector;
Vector.Uint16Vector;
Vector.Int16Vector;
Vector.Uint32Vector;
Vector.Int32Vector;
Vector.Float32Vector;
Vector.Float64Vector;
Vector.PointerVector;
//#endregion
//#region entry.mjs
const cache = new import_lru_cache_with_delete.default(2);
cache.set("a", 1);
console.log(cache.get("a"));
//#endregion
```

A quick scan of the generated file shows no declaration/import for `FibonacciHeap`:

```txt
FibonacciHeap.MinFibonacciHeap 26168
FibonacciHeap = -1
var FibonacciHeap -1
const FibonacciHeap -1
import_fibonacci -1
```

### System Info

```Shell
System:
  OS: macOS 26.5
  CPU: (14) arm64 Apple M4 Pro
  Memory: 636.27 MB / 48.00 GB
  Shell: 4.7.1 - /opt/homebrew/bin/fish
Binaries:
  Node: 26.0.0 - /opt/homebrew/bin/node
  Yarn: 1.22.22 - /opt/homebrew/bin/yarn
  npm: 11.12.1 - /opt/homebrew/bin/npm
  pnpm: 10.33.0 - /opt/homebrew/bin/pnpm
  bun: 1.3.11 - /Users/abhi/.local/share/mise/installs/bun/latest/bin/bun
npmPackages:
  rolldown: 1.1.0 => 1.1.0
```

### Any additional comments?

`mnemonist/index.mjs` is a side-effect-free ESM barrel. It imports defaults from several modules and derives named re-export values, for example:

```js
import {default as FibonacciHeap} from './fibonacci-heap.js';
const MinFibonacciHeap = FibonacciHeap.MinFibonacciHeap;
const MaxFibonacciHeap = FibonacciHeap.MaxFibonacciHeap;

export {FibonacciHeap, MinFibonacciHeap, MaxFibonacciHeap};
export {default as LRUCacheWithDelete} from './lru-cache-with-delete.js';
```

The entry only imports `LRUCacheWithDelete`, but Rolldown keeps bare property reads for the other derived re-export values while dropping the imports that define the objects being read.

Adding this config makes the same repro execute successfully and print `1`:

```js
export default {
  input: 'entry.mjs',
  output: {
    file: 'dist/out.mjs',
    format: 'esm'
  },
  experimental: {
    lazyBarrel: false
  }
};
```

### LINKED ISSUE #9961: [Bug]: Referenced cross-module binding dropped under `strictExecutionOrder` + `sideEffects: false` (regression in 1.1.3)

Minimal reproduction: https://github.com/KurtGokhan/rolldown-checkglobals-repro

```bash
git clone https://github.com/KurtGokhan/rolldown-checkglobals-repro
cd rolldown-checkglobals-repro
npm install
npm run build

grep -c 'checkGlobals()'        dist/assets/*.js   # => 1  (call kept)
grep -c 'function checkGlobals'  dist/assets/*.js   # => 0  (definition gone)

npm run preview   # blank page; console: "ReferenceError: checkGlobals is not defined"
```

The bug surfaces in the wild via [`msw`](https://github.com/mswjs/msw): its `core/index.mjs` calls `checkGlobals()` at module top level, while its `package.json` declares `"sideEffects": false`.

```js
// node_modules/msw/lib/core/index.mjs
import { checkGlobals } from './utils/internal/checkGlobals.mjs';
// ...
checkGlobals();           // top-level side effect
export { http, HttpResponse, /* ... */ };
```

Relevant config (`vite.config.js`):

```js
export default defineConfig({
  build: {
    minify: false,
    rolldownOptions: { output: { strictExecutionOrder: true } },
  },
});
```

**Trigger conditions (all required):**
- `output.strictExecutionOrder: true` — without it, both the call *and* the definition are removed together (no dangling reference)
- the dependency declares `"sideEffects": false`
- that dependency calls an imported cross-module function at its **module top level**
- other exports of the dependency are used, so its module is retained

### What is expected?

A referenced binding must be retained whenever the reference (the `checkGlobals()` call) is retained, regardless of `"sideEffects": false`. Either keep both the call and the definition, or remove both — never keep the call while dropping the definition.

This is what rolldown `1.0.3` (vite `8.0.16`) does: the definition is present and the app works.

rolldown keeps the top-level `checkGlobals()` **call** but tree-shakes the `checkGlobals` function **definition** out of every chunk, leaving a free identifier → `ReferenceError: checkGlobals is not defined` at module-eval time → blank app.

This is a regression between rolldown `1.0.3` and `1.1.3`:

| vite    | rolldown | result                                  |
|---------|----------|-----------------------------------------|
| 8.1.0   | 1.1.3    | ❌ definition dropped, `ReferenceError` |
| 8.0.16  | 1.0.3    | ✅ definition retained, works           |

### System Info

```Shell
  System:
    OS: macOS 26.5.1
    CPU: (12) arm64 Apple M3 Pro
    Memory: 845.58 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.18.0
    npm: 11.16.0
    pnpm: 10.28.1
    bun: 1.3.13
    Deno: 2.8.3+
  Browsers:
    Chrome: 149.0.7827.156
    Firefox Developer Edition: 151.0
    Safari: 26.5
  npmPackages:
    rolldown: 1.1.3 (transitive, via vite 8.1.0)
    vite: 8.1.0
    msw: 2.14.6
```

### Any additional comments?

`msw` declaring `"sideEffects": false` while running `checkGlobals()` at import time is arguably inaccurate on msw's side, but a bundler must still not drop a *referenced* binding — so this looks like a rolldown correctness bug in scope-hoisting / side-effect handling. The concerning part is that the same mechanism could silently drop other referenced cross-module bindings that don't happen to throw loudly at load.

Originally hit while bumping vite `8.0.x` → `8.1.0` in a real app, where it produced a fully blank page (`ReferenceError: checkGlobals is not defined`).

### LINKED ISSUE #9964: experimental.lazyBarrel drops a still-referenced import from a sideEffects:false `export *` barrel under a CJS-wrapped ESM consumer → ReferenceError (repros on 1.1.3)
### Summary

With `experimental.lazyBarrel` (on by default since 1.1.0), rolldown prunes an import that is still referenced by retained code. A `sideEffects:false` module's `import` is dropped, but statement-level DCE keeps the now-dead derived-initializer statements that use it inside an `__esm`-wrapped consumer, so the emitted bundle references a binding that is never declared/imported → `ReferenceError: eg is not defined` at module evaluation.

`experimental.lazyBarrel: false` fixes it.

### Versions

- **Broken:** rolldown **1.1.2** and **1.1.3** (latest).
- **Works:** rolldown **1.0.3** (lazyBarrel was off by default before 1.1.0), and any version with `experimental.lazyBarrel: false`.

### Reproduction

Repro repo: https://github.com/hyf0/rolldown-lazybarrel-reexport-repro

Pure JS, only dependency is `rolldown`. Five source modules (inlined below).

`package.json`
```json
{
  "type": "module",
  "private": true,
  "sideEffects": false,
  "dependencies": { "rolldown": "1.1.3" }
}
```

`src/lib/eg.js` — defines `eg` (the import that gets pruned)
```js
const primitives = { string: { tag: 's' }, number: { tag: 'n' } };
const higher = { object: (shape) => ({ tag: 'o', shape, parse: () => true }) };
export const eg = { ...primitives, ...higher };
```

`src/lib/account.js` — `sideEffects:false`; imports `eg`, defines derived initializers
```js
import { eg } from './eg.js';
export const AccountSettings = eg.object({ a: eg.string, n: eg.number });
export const AccountMeta = eg.object({ b: eg.string });
```

`src/lib/index.js` — re-exports the schema module via `export *`, plus an unrelated util
```js
export * from './account.js';
export const objectKeys = (o) => Object.keys(o);
```

`src/story.js` — ESM consumer; imports **only** the util, never the schemas
```js
import { objectKeys } from './lib/index.js';
export const run = () => (objectKeys ? 'ok' : 'no');
```

`src/entry.cjs` — CommonJS entry that `require`s the ESM consumer → wraps it in `__esm`
```js
const { run } = require('./story.js');
console.log('result:', run(), 'REPRO_OK');
```

`build.mjs`
```js
const disable = process.env.LAZY_BARREL === 'off';
  input: { entry: './src/entry.cjs' },
  ...(disable ? { experimental: { lazyBarrel: false } } : {}),
});
await bundle.write({ dir: disable ? './dist-fixed' : './dist', format: 'es' });
await bundle.close();
```

```
npm install
node build.mjs && node dist/entry.js                       # ReferenceError: eg is not defined
LAZY_BARREL=off node build.mjs && node dist-fixed/entry.js  # result: ok REPRO_OK
```

### What the emitted code looks like

In `dist/entry.js` (lazyBarrel on) the whole `eg.js` defining region (`primitives`, `higher`, `eg`, the `init_eg` initializer) is **absent**, yet the `__esm`-wrapped module still contains the retained `eg.object(...)` initializers referencing `eg`. Diffing against the `lazyBarrel:false` bundle, the **only** difference is the missing `eg.js` region and its `init_eg()` calls — i.e. lazyBarrel pruned exactly the still-needed defining code.

### Minimal trigger (each verified necessary by deletion)

1. an `export *` re-export of a **separate** `sideEffects:false` module that imports `eg`, distinct from the module providing the export the consumer actually uses (collapsing the schema into the consumed module, or using a named re-export instead of `export *`, hides it);
2. a CommonJS → ESM `require` that wraps the consumer in the `__esm` helper (a pure-ESM import chain does not reproduce — DCE then drops the dead initializers too);
3. `experimental.lazyBarrel` enabled (the default).

Not required: nested barrels, code-splitting, minification, `legacy.inconsistentCjsInterop`, Vite/Storybook.

### Relation to prior issues

Same family as #9806 (flat barrel, no wrap) and #9691 / #9709 (`codeSplitting:false` dynamic-import wrap). The closest prior fix is **#9757** ("load locally-used imports on a re-exported record", closing #9713), which is included in 1.1.2/1.1.3 — but this variant still reproduces on **1.1.3**, so it is not covered by that fix.

### Workaround

`experimental: { lazyBarrel: false }`.

### LINKED ISSUE #10013: [Bug]: strictExecutionOrder emits external bindings without their import
### Problem

I found this while bundling Storybook. The reproduction below is reduced, but the original failure was a generated Storybook bundle where a lazy `__esmMin` wrapper still referenced external bindings after the corresponding external import had been removed.

https://github.com/storybookjs/storybook/blob/d0ef37ec3f69e9ceff6328e9772cc506dd58f995/code/builders/builder-vite/src/utils/vite-features.ts#L27

When `strictExecutionOrder` is enabled, Rolldown can generate a shared chunk that still references bindings from an external module, but the corresponding external `import` statement is missing from the output.

The generated code can look like this:

```js
var init_leaf = __esmMin(() => {
  value = first(second);
});
```

In that chunk, `first` and `second` came from an external import:

```js
import { first, second } from "dep";
```

but the import declaration is removed, so the emitted chunk contains free identifiers and fails at runtime.

### Expected behavior

If generated code still references bindings from an external import, the `ImportDeclaration` that defines those bindings should remain in the same output chunk.

### Why this seems tied to `strictExecutionOrder`

With `strictExecutionOrder`, ESM modules may be wrapped in lazy `__esmMin` initializers to preserve evaluation order across entries. That wrapper can keep top-level module code alive even after tree-shaking. In this case, the wrapper body still uses the external bindings, but finalization removes the external import statement because the import record does not resolve to an internal module.

### Minimal reproduction

```js

const modules = {
  'entry-a': `import { value } from 'first'; console.log(value); export { read } from 'first';`,
  'entry-b': `import { value } from 'second'; console.log(value); export { read } from 'second';`,
  first: `export { value, read } from 'shared';`,
  second: `export { value, read } from 'shared';`,
  shared: `export { value, read } from 'leaf';`,
  leaf: `import { first, second } from 'dep'; export const value = first(second); export function read() { return value; }`,
};

const plugin = {
  name: 'virtual',
  resolveId(id) {
    if (id in modules) return id;
  },
  load(id) {
    return modules[id];
  },
};

  input: {
    a: 'entry-a',
    b: 'entry-b',
  },
  external: ['dep'],
  platform: 'browser',
  plugins: [plugin],
  treeshake: {
    moduleSideEffects: false,
  },
});

const { output } = await bundle.generate({
  format: 'es',
  strictExecutionOrder: true,
  minify: false,
});

for (const chunk of output) {
  if (chunk.type === 'chunk' && chunk.code.includes('first(second)')) {
    console.log(chunk.code);
  }
}
```

Related PR: #10009

### LINKED ISSUE #10048: Rolldown emits `__commonJS` helper call sites without the definition (vite 8.1.x Storybook build → `__commonJS is not defined`)
### Summary

Building a Storybook (bundled via `@storybook/react-vite`) with **Vite 8.1.x** produces an iframe chunk that *calls* Rolldown's full `__commonJS` CJS-interop helper but never emits its *definition*. The build succeeds; the bundle throws `__commonJS is not defined` as soon as it loads, so the Storybook UI never initializes. **Vite 8.0.16 builds the same project cleanly** — there the helper is defined where it is used. The regression tracks the Rolldown bump that shipped in Vite 8.1.0 (Rolldown 1.1.2) and persists through Vite 8.1.2 (Rolldown 1.1.3).

The tell is that the *sibling* helper `__commonJSMin` is emitted and imported correctly in the very same 8.1.x bundle, and so are Rolldown's numbered `__commonJS$1` / `__commonJS$2` / … instances. Only the bare `__commonJS` binding used by one group of call sites is referenced without ever being assigned, declared, or imported into any loaded chunk — which reads as a Rolldown codegen / chunking bug.

### Environment

| | |
|---|---|
| vite (broken) | 8.1.0 and 8.1.2 |
| vite (working) | 8.0.16 |
| rolldown bindings | 1.1.2 (via vite 8.1.0), 1.1.3 (via vite 8.1.2) |
| storybook / @storybook/react-vite | 10.4.6 |
| node | 24.18.0 |
| pnpm | 11.9.0 |
| OS | macOS (darwin arm64) |

A self-contained minimal reproduction (bare `@storybook/react-vite` + one trivial story) is linked in the comments, along with a headless-browser check that observes the runtime error.

### Steps to reproduce

1. A project using `@storybook/react-vite` with `vite@8.1.2` installed.
2. `storybook build` — the build **succeeds**.
3. Serve / open the built `storybook-static` (or let Chromatic publish + verify it).
4. The iframe bundle throws on load:

   ```
   ReferenceError: __commonJS is not defined
   ```

### Evidence

From the built iframe chunk (`build.minify = false`, so the helper keeps its readable name). The bare `__commonJS` is called four times — wrapping a CommonJS chain pulled in by `react-i18next` (`react-i18next → hoist-non-react-statics → react-is`):

```bash
$ grep -oE '__commonJS\(\{ "[^"]+"' storybook-static/assets/iframe-*.js
__commonJS({ "…/react-is/cjs/react-is.production.min.js"
__commonJS({ "…/react-is/cjs/react-is.development.js"
__commonJS({ "…/react-is/index.js"
__commonJS({ "…/hoist-non-react-statics/dist/hoist-non-react-statics.cjs.js"
```

…but its definition — as an assignment, a declaration, or an import — is emitted into no chunk in the bundle:

```bash
# bare __commonJS: assignment `__commonJS =`, declaration, or `import { … as __commonJS }`
$ grep -rnE '(^|[^0-9$A-Za-z_])__commonJS[[:space:]]*=|as __commonJS[[:space:],}]|(var|let|const|function)[[:space:]]+__commonJS\b' storybook-static/assets
(no match)

# meanwhile the sibling helper IS defined and imported in the same build:
$ grep -rl '__commonJSMin[[:space:]]*=' storybook-static/assets   # defined
$ grep -rl 'as __commonJSMin' storybook-static/assets             # imported
```

> Note on grepping: matching only `__commonJS =` is misleading, because Rolldown also emits the helper via bare assignment and via numbered variants (`__commonJS$1 =`, `__commonJS2 =`, …). Those numbered instances *are* defined here; it is specifically the bare `__commonJS` used by the four call sites above that is missing. Under default minification the same holds — the orphaned helper is just renamed (e.g. `K`) and still never defined. The reproduction includes a minify-agnostic runtime check that loads the iframe in headless Chromium and catches the actual `ReferenceError`.

### Expected

The `__commonJS` interop helper is defined in (or imported into) every chunk that calls it, as under Vite 8.0.16.

### Actual

The bare `__commonJS` call sites are emitted without the definition → runtime `ReferenceError: __commonJS is not defined`.

### What appears to trigger it

It is a graph-size threshold rather than a single switch. In the reproduction, none of these alone trips it — they have to be combined:

- the `react-i18next → hoist-non-react-statics → react-is` CommonJS chain (a dual production/development CJS package),
- the Storybook addon set (`addon-docs`, `addon-links`, `storybook-react-i18next`, `@vueless/storybook-dark-mode`),
- `lodash` forced in-bundle via `ssr.noExternal`.

Below the threshold Rolldown inlines a working `__commonJS` per chunk; at/above it, Rolldown appears to hoist the helper toward a shared chunk and then emits neither its definition nor an import for it. Plain Vite (no Storybook) and a stripped-down Storybook both build correctly, so the fault surfaces through `@storybook/builder-vite`'s single-virtual-entry iframe build once the CommonJS graph is large enough.

### Workaround

Pin `vite` to `8.0.16`.

### Notes

- The downstream app build (a React Router SSR build) is unaffected; only the Storybook build trips the fault.

### Related

- Downstream Storybook report: storybookjs/storybook#35332

IWANABETHATGUY:
### What this PR solves

Any wrap trigger — `strictExecutionOrder`, `require()` of ESM, or an inlined dynamic import under `codeSplitting: false` — force-includes a `sideEffects: false` ESM module through its wrapper ref, retaining binding-reading side-effect statements that plain (unwrapped) tree-shaking provably drops. With `experimental.lazyBarrel`, those retained statements reference import records the loader correctly never loaded (it defers records by requested exports), so the emitted chunk contains free identifiers and throws `ReferenceError` at load. The tree-shaking inconsistency between `strictExecutionOrder: true` and `false` is the shared root cause of the family:

fixes #9691, fixes #9806, fixes #9961, fixes #9964, fixes #10013, fixes #10048

### How

A binding-reading side-effect statement of a user-declared side-effect-free ESM module is now included **iff the module's body is demanded**: one of its own (non-re-export) exports or its namespace object becomes used. This mirrors the lazy-barrel loader's `local`/`All` classification exactly — the loader loads *every* plain import record precisely in those cases — so a kept statement can never reference an unloaded record, and a deferred record is only ever referenced by dropped statements. The loader itself needs no changes.

Behavior deliberately kept as-is (pinned by existing tests):

- statements with no module-symbol references (a bare `console.log()`) and import/re-export statements (they drive wrapper `init_*` chains and side-effect imports) keep the unconditional sweep — `package_json_side_effects_false_keep_*` esbuild fixtures;
- bodies of modules whose own export is demanded (#7597's top-level asserts) or whose namespace is observed (`import * as ns`, `require(esm)`) are retained unchanged;
- user-defined entries, `eval` modules, and CommonJS modules are exempt.

Dynamic entries are **not** exempt: a live dynamic entry's body joins through namespace/own-export demand, while a dead pure dynamic import must not resurrect a body its removal already models as an empty namespace (covered by the new `tree_shaking/dead_dynamic_entry_side_effect_free` fixture).

### Alternatives explored

- Loader-side fix (force-load records referenced by retained statements): rejected — the deferral strategy is correct; the retention is the bug.
- Blanket statement-level stripping for `sideEffects: false` modules: rejected — breaks the esbuild-parity `keep_*` fixtures (namespace/require demand must keep the body).
- Statement-granular demand edges (keep a statement once any symbol it references is used): rejected — still dropped #7597's asserts.

### Notes for reviewers

- Inclusion changes only for `UserDefined(false)` ESM modules reached exclusively through re-exports or wrapper refs; every pre-existing snapshot is byte-identical (full integration suite: 1785 passed, 0 failed).
- Each `issues/*` fixture executes its output with node; `issues/10013` and the dynamic-entry fixture additionally carry a local `node_modules/dep` that throws on evaluation, so re-emitting the dropped external import fails execution, not just the snapshot.
- Residual and deferred (benign): wrapped modules whose body was pruned still emit empty `init_x` shells; a follow-up no-op-init cleanup can drop them.


IWANABETHATGUY:
* **#10086** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10086?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10080** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10080?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10080?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10080).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyfdev (approved):
I found some issues but I'm not sure if it's correct to fix them in this direction. The treeshking part becomes too complicated, it's very hard to  manually reason if the behavior is correct.

However, tests are passed, it's possitive to prevent future regressions after merging them.

hyfdev:
### Merge activity

* **Jul 2, 10:57 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 2, 10:58 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 2, 10:58 AM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 2, 11:03 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

IWANABETHATGUY:
> I found some issues but I'm not sure if it's correct to fix them in this direction. The treeshking part becomes too complicated, it's very hard to manually reason if the behavior is correct.
> 
> However, tests are passed, it's possitive to prevent future regressions after merging them.

Do you mean this issue https://github.com/rolldown/rolldown/issues/10085?

hyfdev:
> > I found some issues but I'm not sure if it's correct to fix them in this direction. The treeshking part becomes too complicated, it's very hard to manually reason if the behavior is correct.
> > However, tests are passed, it's possitive to prevent future regressions after merging them.
> 
> Do you mean this issue #10085?

No.

PR #10093, opened 2026-07-02
perf: avoid redundant sourcemap string copies in collapse and minify paths

Boshen:
### Description

`collapse_sourcemaps` only copies strings from the *first* map of the chain, yet its `&[&SourceMap<'static>]` signature forced callers to `into_owned()` freshly generated codegen maps before collapsing — deep-copying strings (including whole-chunk `sourcesContent`) that were immediately discarded.

- `rolldown_sourcemap::collapse_sourcemaps` now accepts `&[&SourceMap<'_>]`; existing callers compile unchanged via covariance, and the output stays owned since it only copies from the first map.
- `EcmaCompiler::dce_or_minify` returns the codegen map borrowing its input instead of detaching it to `'static`.
- `minify_chunks` collapses the minify map before swapping in the minified content, removing a full copy of the pre-minify chunk text + all mangled names per minified chunk.
- `NormalModule::render` (mutations path) and `enhanced_transform` drop the same immediately-discarded `into_owned` copies (one module-source copy per mutated module / per transform call with an input map).
- `adjust_sourcemap_dst_lines` moves the map through `into_parts()` instead of re-cloning every name/source/`sourcesContent` string, and shifts tokens in place when none fall inside the stripped prefix (always the case for rolldown-generated maps, since prepended text carries no tokens). Side effect: `file`/`sourceRoot`/`debugId`/ignore-list are now preserved instead of silently dropped.

Outputs are byte-identical on `three10x` with `minify: true` + `sourcemap: 'file'` (chunk and 19.6 MB map hash equal before/after); user CPU is unchanged and peak RSS drops ~15–40 MB on that fixture since the removed copies were transient.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/minify_chunks.rs
+++ b/crates/rolldown/src/stages/generate_stage/minify_chunks.rs
@@ -44,25 +44,35 @@ impl GenerateStage<'_> {
           };
 
           let allocator_guard = allocator_pool.get();
-          // TODO: Do we need to ensure `chunk.preliminary_filename` to be absolute path?
-          let (minified_content, new_map) = EcmaCompiler::dce_or_minify(
-            &allocator_guard,
-            chunk.content.try_as_inner_str()?,
-            options.format.source_type().with_jsx(true),
-            chunk.map.is_some(),
-            chunk.preliminary_filename.as_str(),
-            compress,
-            minify_option.clone(),
-            codegen_options,
-          );
+          // The minify map borrows the pre-minify `chunk.content` (as `sourcesContent`,
+          // which the collapse discards), so collapse before swapping in the minified
+          // content instead of paying an `into_owned` copy of the whole chunk text.
+          let (minified_content, collapsed_map) = {
+            // TODO: Do we need to ensure `chunk.preliminary_filename` to be absolute path?
+            let (minified_content, new_map) = EcmaCompiler::dce_or_minify(
+              &allocator_guard,
+              chunk.content.try_as_inner_str()?,
+              options.format.source_type().with_jsx(true),
+              chunk.map.is_some(),
+              chunk.preliminary_filename.as_str(),
+              compress,
+              minify_option.clone(),
+              codegen_options,
+            );
+            let collapsed_map = match (&chunk.map, &new_map) {
+              (Some(origin_map), Some(new_map)) => {
+                Some(collapse_sourcemaps(&[origin_map, new_map]))
+              }
+              _ => {
+                // TODO: Map is dirty. Should we reset the `chunk.map` to `None`?
+                None
+              }
+            };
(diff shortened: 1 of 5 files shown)


IWANABETHATGUY (approved):

crates/rolldown/src/stages/generate_stage/minify_chunks.rs
```diff
@@ -44,25 +44,35 @@ impl GenerateStage<'_> {
...
-            chunk.preliminary_filename.as_str(),
-            compress,
-            minify_option.clone(),
-            codegen_options,
-          );
+          // The minify map borrows the pre-minify `chunk.content` (as `sourcesContent`,
+          // which the collapse discards), so collapse before swapping in the minified
+          // content instead of paying an `into_owned` copy of the whole chunk text.
+          let (minified_content, collapsed_map) = {
+            // TODO: Do we need to ensure `chunk.preliminary_filename` to be absolute path?
```
TheAlexLichter, line 51: @IWANABETHATGUY can you check this TODO?

crates/rolldown/src/stages/generate_stage/minify_chunks.rs
```diff
@@ -44,25 +44,35 @@ impl GenerateStage<'_> {
...
+              compress,
+              minify_option.clone(),
+              codegen_options,
+            );
+            let collapsed_map = match (&chunk.map, &new_map) {
+              (Some(origin_map), Some(new_map)) => {
+                Some(collapse_sourcemaps(&[origin_map, new_map]))
+              }
+              _ => {
+                // TODO: Map is dirty. Should we reset the `chunk.map` to `None`?
```
TheAlexLichter, line 67: (Same here)

crates/rolldown/src/stages/generate_stage/minify_chunks.rs
Boshen: These TODOs were already in the codebase.

crates/rolldown/src/stages/generate_stage/minify_chunks.rs
Boshen: These TODOs were already in the codebase.

Boshen:
### Merge activity

* **Jul 13, 1:49 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 13, 1:49 AM UTC**: `Boshen` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 13, 1:54 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10096, opened 2026-07-02
refactor(treeshake): replace inclusion mutual recursion with a worklist engine

IWANABETHATGUY:
### Description

Part 2 of the tree-shaking inclusion refactor stack (stacked on #10095).

Replaces the `include_module` ↔ `include_symbol` ↔ `include_statement` mutual recursion with an explicit worklist engine. Edge producers push typed work items — `WorkItem::Module(idx)` (structural inclusion: sweep side-effect statements, evaluate side-effectful deps), `WorkItem::Symbol(ref, reason)` (a symbol became used: retain its declaration, wrapper, owner module), `WorkItem::Statement(idx, stmt)` (include one statement and follow its references) — onto a queue that the public `include_*` entry points drain to empty before returning. Handlers and internal helpers only push; only the public entry points drain, so drains never nest.

Why:

- no stack-depth limit on deep module graphs (the recursion could overflow)
- inclusion order becomes a visible queue discipline instead of an implicit call stack — the class of traversal-ordering bugs becomes inspectable and testable
- every inclusion flows through one loop, giving a single choke point for future reason/provenance tracking

Correctness argument: the final inclusion sets are a monotone closure over the pushed edges (all handler dedup is idempotent bit-setting; reason flags are ORed), so drain order affects only traversal order, never results. LIFO drain mirrors the depth-first shape of the recursion it replaces. Confirmed empirically: byte-identical output across the integration suite, zero snapshot diffs.

Public API is unchanged — `passes.rs`, `dynamic_entries.rs`, and `chunk_optimizer.rs` keep calling the same `include_*` functions, apart from the one new context field.

#### Review follow-up: queue-invariant hardening

The drain-before-return invariant made helper boundaries subtle — especially `include_declaring_statements`, which was enqueue-only yet callable from outside the engine (driver entry loop, dynamic entries), legitimately pre-seeding the queue across a public entry. Addressed:

- split it: `enqueue_declaring_statements` (private, engine-internal, used by handlers) vs `include_declaring_statements` (`pub(super)`, now drains like every other public entry) — so "queue empty between public entries" is a true invariant, not a mostly-true one
- `debug_assert!` at every public entry point that the queue is empty on entry and after draining; the full fixture suite runs with these live
- `WorkItem` and the `pending` field narrowed to `pub(in crate::stages)` (the minimum: `chunk_optimizer` constructs `IncludeContext` with a struct literal), documented as not-API

Stack: #10095 → #10096 → #10097 → #10098

PR #10095, opened 2026-07-02
refactor(treeshake): split include_statements.rs into focused modules

IWANABETHATGUY:
### Description

Part 1 of the tree-shaking inclusion refactor stack.

`include_statements.rs` had grown to ~1200 lines interleaving the inclusion fixpoint driver, the three mutually recursive inclusion functions, dynamic-entry topological ordering, the body-demand gating model, and several standalone passes — with 10+ orthogonal policies (CJS bailouts, constant/enum inlining bypasses, JSON self-reference tracking, `eval` handling, HMR namespace registration, property-write retention) inlined directly into the recursion. Reviewing any change meant re-deriving which clause belonged to which concern; every bug fix added another inline branch.

Pure code motion, in three commits:

- **Split into focused modules**
  - `dynamic_entries.rs` — dynamic-entry topological ordering (tarjan), aliveness determination, per-iteration processing (~200 lines unrelated to inclusion mechanics)
  - `on_demand.rs` — the body-demand gating model (`side_effects_included_on_demand`, `compute_on_demand_side_effect_stmts`, `is_gated_side_effect_stmt`)
  - `passes.rs` — standalone passes around the fixpoint (CJS bailout export inclusion, runtime-helper collection/inclusion, `preserveModules` re-export interface preservation)
- **Extract named policies** — each inline block of `include_module` / `include_symbol` / `include_statement` becomes a named function with a doc contract (`sweep_side_effect_statements`, `include_side_effectful_dependencies`, `is_bypassed_inlined_constant`, `drain_body_demand_stmts`, `follow_cjs_namespace_alias`, `note_namespace_inclusion_reason`, `demand_esm_init_wrapper`, `note_json_self_reference`, `include_property_write_referencing_stmts`, `scan_import_records_for_cjs_bailout`, `is_inlined_enum_member_access`), so the three skeletons read as a sequence of policies. Also dedupes `include_declaring_statements`, which replaces five hand-rolled copies of the "a used binding keeps its declaring statements" loop.
- **wasm build fix** — the move dropped an incidental `itertools::Itertools` import that wasm targets need for `zip_eq` (`rolldown_utils::rayon` shims `par_iter()` to a plain `Iterator` there); caught by the browser/WASI CI jobs.

No behavior change: byte-identical output across the integration suite (zero snapshot diffs), clippy clean, `wasm32-wasip1-threads` checked.

Stack: #10095 → #10096 → #10097 → #10098

PR #10097, opened 2026-07-02
refactor(treeshake): make body demand a second module bit instead of a stmt multimap

IWANABETHATGUY:
### Description

Part 3 of the tree-shaking inclusion refactor stack (stacked on #10096).

#### Before

The body-demand machinery stored, for **every** body-demand key of a module, a full copy of that module's gated side-effect statement list (`FxHashMap<SymbolRef, Vec<(ModuleIdx, StmtInfoIdx)>>`), drained by map removal. That one mutable map entangled three different things: the **demand edge** (key → module), the **gate predicate** (which statements are gated), and the **progress state** (encoded as entry presence/absence). A module with N own exports carried N copies of the same statement list, and "has this module's body already joined?" was emergent from which keys happened to have been removed.

#### After

The rule being implemented — *a `sideEffects: false` module's gated statements join once its body is demanded* — is a boolean event on a **module**, and the state now says exactly that. Module inclusion is two explicit bits:

- `is_module_included_vec` — structural inclusion (wrapper/interop resolvable, side-effectful deps evaluated); unchanged
- `body_demand_swept` — the module's gated side-effect statements have joined because one of its own exports / its namespace object was actually used

and each of the three entangled concerns lives where it belongs:

- **Demand edge** — `body_demand_keys: FxHashMap<SymbolRef, ModuleIdx>` ("whose body does using this symbol demand?"), immutable and shared across both `IncludeContext` constructions (link stage and chunk-optimizer replay); no more `&mut` threading. Every key's canonical is owned by the module itself (own exports and namespace refs never link across modules). Before, `entry(key).or_default().extend(..)` would have silently merged two modules' statements if that invariant ever broke; now it's an explicit `debug_assert!` exercised by the whole suite.
- **Gate predicate** — the gated statements are enumerated at demand time from the immutable `stmt_infos`, via the same `is_gated_side_effect_stmt` used when building the key map. One source of truth; no O(keys × gated stmts) materialized copies.
- **Progress** — `body_demand_swept: FxHashSet<ModuleIdx>`, owned per run. `insert()` returning `false` short-circuits, so the sweep happens exactly once per **module**. Before, `remove(&key)` only guaranteed once per *key*: demanding a second export of the same module re-enqueued all its statements (harmless only because `include_statement` is idempotent — you had to know that to trust it). The generate-stage replay gets shared edges plus a fresh swept set: graph-scoped facts vs run-scoped progress.

#### Why `FxHashMap`/`FxHashSet` rather than `IndexBitSet`

`is_module_included_vec` is already an `IndexBitSet` — dense domain, touched for essentially every module on the hot path. The two body-demand structures sit on the other side of that line: they are sparse, usually **empty**. A module contributes keys only if it is user-declared side-effect-free, ESM, non-entry, without `eval`, **and** has a top-level side-effecting statement that reads module-level bindings — the rare #9806-family pattern.

- `body_demand_keys` can't be dense-indexed at all: `SymbolRef` is a two-dimensional key (`owner` × `SymbolId`), so a dense encoding would cost O(total symbols in the graph). The map *is* probed once per symbol in `include_symbol`, but a `get` on an empty `FxHashMap` is a few instructions — the common case stays effectively free.
- `body_demand_swept` is constructed fresh per `IncludeContext` (twice in the link stage, again in the chunk-optimizer replay). `IndexBitSet::new(modules.len())` would allocate and zero the full module range each time to record what is almost always nothing; `FxHashSet::default()` allocates nothing until the first insert, and it is only touched after a `body_demand_keys` hit. Its cardinality is bounded by the number of gated modules, so both structures are zero-cost together in the common case.
- Neither structure is ever iterated (membership test / probe only), so hash iteration order cannot leak into output.

#### Explored and rejected

Deriving `is_module_included` from the statement bits. The `SymbolIncludeReason::SimulatedFacadeChunk` path (chunk optimizer) deliberately includes statements **without** structurally including the module, so a derived bit is a behavior change, not a refactor.

No behavior change: byte-identical output across the integration suite, zero snapshot diffs.

Stack: #10095 → #10096 → #10097 → #10098

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -15,8 +15,8 @@ use rustc_hash::{FxHashMap, FxHashSet};
 use crate::{
   chunk_graph::ChunkGraph,
   stages::link_stage::{
-    IncludeContext, SymbolIncludeReason, compute_on_demand_side_effect_stmts,
-    include_runtime_symbol, include_symbol,
+    IncludeContext, SymbolIncludeReason, compute_body_demand_keys, include_runtime_symbol,
+    include_symbol,
   },
   types::linking_metadata::{
     LinkingMetadata, LinkingMetadataVec, included_info_to_linking_metadata_vec,
@@ -1026,7 +1026,7 @@ impl GenerateStage<'_> {
     // Replay the link-stage inclusion semantics: side-effectful statements of
     // user-declared side-effect-free modules join only through body demand.
     // Already-included statements make the replayed edges no-ops.
-    let mut on_demand_side_effect_stmts = compute_on_demand_side_effect_stmts(
+    let body_demand_keys = compute_body_demand_keys(
       &self.link_output.module_table.modules,
       &self.link_output.stmt_infos,
       &self.link_output.symbol_db,
@@ -1055,7 +1055,8 @@ impl GenerateStage<'_> {
       inline_const_smart: self.options.optimization.is_inline_const_smart_mode(),
       json_module_none_self_reference_included_symbol: FxHashMap::default(),
       entry_module_idxs: &self.link_output.user_defined_entry_modules,
-      on_demand_side_effect_stmts: &mut on_demand_side_effect_stmts,
+      body_demand_keys: &body_demand_keys,
+      body_demand_swept: FxHashSet::default(),
       pending: Vec::new(),
     };
 
--- a/crates/rolldown/src/stages/link_stage/mod.rs
+++ b/crates/rolldown/src/stages/link_stage/mod.rs
@@ -41,7 +41,7 @@ mod tree_shaking;
 
 pub use tree_shaking::{
   IncludeContext, ModuleInclusionVec, ModuleNamespaceReasonVec, StmtInclusionVec,
-  SymbolIncludeReason, compute_on_demand_side_effect_stmts, include_runtime_symbol, include_symbol,
+  SymbolIncludeReason, compute_body_demand_keys, include_runtime_symbol, include_symbol,
 };
 mod wrapping;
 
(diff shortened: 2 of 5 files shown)


IWANABETHATGUY:
* **#10098** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10098?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10097** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10097?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10097?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10096** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10096?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10095** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10095?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10097).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyfdev (approved):

IWANABETHATGUY:
### Merge activity

* **Jul 3, 6:00 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 3, 6:00 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 3, 6:05 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

Issue #9224, opened 2026-04-24
Circular runtime helper import between entry chunk and code-split CJS chunk causes "e is not a function" at load (rc.17, regression from rc.15)


Stanzilla:
### Reproduction link or steps

I don't have a minimal REPL yet (the trigger is a real-world dependency graph), but the behavior is fully deterministic and I can produce one if needed. The key shape is:

- App entry imports a CommonJS module `@faceit/protos-commonjs` (a generated protobuf package), which `require()`s `google-protobuf`.
- `google-protobuf` is its own large CommonJS module (Closure-compiler output).
- Rolldown chunks `google-protobuf` separately from the entry, lifts the `__commonJSMin` runtime helper into the entry chunk, and re-imports it from `google-protobuf` — creating a cycle:

  - `google-protobuf-*.js`:
    ```js
    import { n as e } from "./index-*.js";
    var t = e((e => { /* google-protobuf factory */ }));
    ```
  - `index-*.js`:
    ```js
    import { t as require_google_protobuf } from "./google-protobuf-*.js";
    …
    __commonJSMin = (e, t) => () => (t || (e((t={exports:{}}).exports, t), e = null), t.exports);
    …
    export { __commonJSMin as n, require_timestamp_pb as t };
    ```

When the browser runs the cycle, `google-protobuf` executes first; at that point `n` (the live binding to `__commonJSMin`) is still uninitialized, so the top-level `var t = e(…)` calls `undefined` as a function.

### What is expected?

The `__commonJSMin`/`__commonJS` runtime helper either gets inlined into the chunk that needs it (the way rolldown 1.0.0-rc.15 did), or some other arrangement that does not produce a circular ESM import between an entry chunk and a code-split CJS chunk. No runtime error.

### What is actually happening?

Production build of two of our apps (Vite 8.0.10, rolldown 1.0.0-rc.17) crashes immediately on load with a blank page and:

```
google-protobuf-DvwQlcDQ.js:1 Uncaught TypeError: e is not a function
    at google-protobuf-DvwQlcDQ.js:1:462
```

Because `var t = e(…)` runs at top level of the `google-protobuf` chunk before `index` has had a chance to assign `__commonJSMin`, `e` is `undefined`.

### Comparison with rolldown 1.0.0-rc.15 (Vite 8.0.8 — works)

Same source, same vite config, just a different bundled rolldown. The chunk is self-contained, no cross-chunk runtime import:

```js
var e=Object.create, … ,
[...]

PR #10101, opened 2026-07-02
fix(chunk-optimizer): follow entry facade edges in runtime placement cycle check

### LINKED ISSUE #9993: [Bug]: `__commonJSMin is not a function` — runtime helper lifted into entry chunk again in 1.1.2 (regression of #9224)

This is a **regression in rolldown 1.1.2 of the bug fixed for #9224** (which was itself a rc.17 regression from rc.15). The runtime CJS-interop helper (`__commonJSMin`) is once again lifted into the **entry chunk** and re-imported by a code-split CJS chunk that is evaluated *before* the entry body runs, so the helper is still uninitialized at the call site → `TypeError: __commonJSMin is not a function`.

I don't have a minimal REPL (same as #9224 — the trigger is a real dependency graph), but the before/after is fully deterministic. It surfaces via **vite**, whose only dependency delta between the working and broken versions is the bundled rolldown:

| vite | bundled rolldown | result |
| --- | --- | --- |
| 8.0.16 | `1.0.3` | ✅ works |
| 8.1.0 | `~1.1.2` | ❌ `__commonJSMin is not a function` |

Stack (TanStack Start SSR prerender of a Vite app; Node ESM evaluating the server function):

```
    at .../functions/__server.func/_libs/@tanstack/react-router+[...].mjs:3917:71
    at .../functions/__server.func/recharts+[...].mjs:15:71
```

### The before/after that pins it

**rolldown 1.0.3 (works)** — helper lives in a dedicated *leaf* runtime module `_runtime.mjs` (no imports → always initialized before any consumer):

```js
// _libs/recharts+[...].mjs
import { n as __exportAll } from "../_runtime.mjs";   // leaf helper module, no cycle
```

**rolldown 1.1.2 (broken)** — helper is inlined into the **server entry** `_ssr/ssr.mjs` (hoisted `var`, exported `as s`) and re-imported by sibling chunks:

```js
// recharts+[...].mjs  (and _libs/@tanstack/react-router+[...].mjs, etc.)
import { c as __exportAll, l as __require, s as __commonJSMin, u as __toESM } from "./_ssr/ssr.mjs";
//                                          ^^^^^^^^^^^^^^^^^ imported from the ENTRY chunk
var require_use_sync_external_store_shim_production = __commonJSMin((exports) => { ... }); // top-level call

// _ssr/ssr.mjs  (the entry)
var __commonJSMin = (cb, mod) => () => (mod || (cb((mod = { exports: {} }).exports, mod), cb = null), mod.exports);  // line 14
...
export { ..., __commonJSMin as s, ... };
```

Because `_ssr/ssr.mjs` is the entry, it imports the route/lib chunks at its top; ESM evaluates those imports depth-first **before** the entry body assigns the `var __commonJSMin`, so the live binding is still `undefined` when the sibling chunk runs its top-level `__commonJSMin(...)`.

This is exactly the cycle shape described in #9224. It reproduces with **default chunking** — removing a manual `output.codeSplitting.groups` entry for one library just moves the failure to the next CJS-wrapped sibling chunk (e.g. `@tanstack/react-router`), confirming it is the general runtime-helper-in-entry placement, not a specific group config (so it is distinct from the `advancedChunks`-specific #9887, though likely the same underlying cause).

### What is expected?

The `__commonJSMin`/`__commonJS` runtime helper should be emitted so it is initialized before any consumer — e.g. inlined into the chunk that needs it, or kept in a dependency-free leaf module like rolldown 1.0.3 emitted (`_runtime.mjs`) — never lifted into an entry chunk that imports its own consumers, which creates the circular ESM init.

In rolldown 1.1.2 the helper is placed in the entry chunk and imported by code-split CJS chunks that execute first, throwing `__commonJSMin is not a function` at module-init. Pinning rolldown back to 1.0.3 (via vite 8.0.16) avoids it.

### Versions

- rolldown 1.1.2 (broken) / 1.0.3 (works), as bundled by vite 8.1.0 / 8.0.16
- @tanstack/react-start SSR prerender, nitro 3.0.260610-beta, Node 24/26

hyfdev:
### Description

This is also more like a workaround fix, I'm working on how to solve these issues in a row. It requires some structure changes, meanwhile let's fix it first since it's p1

---

Fixes #9993 (regression of #9224, introduced in 1.1.1 by #9419).

When a `codeSplitting` group captures an **entry module** into another chunk, the entry chunk gets a render-time static import of that module's wrapper/namespace symbol from the capturing chunk (via `collect_depended_symbols`). `chunk_reaches_via_static_import` only walked module **import records**, so the runtime-merge decision in `try_merge_runtime_chunk` could not see that back-edge:

- `runtime_target_would_create_static_cycle` reported "no cycle" for a target that does statically import a helper consumer back, and
- `find_consumer_dominator` treated that entry chunk as a downstream sink.

The runtime module was then merged into the entry chunk of `entry-2`, while the group chunk both called `__commonJSMin` at top level and was statically imported back by that entry chunk — so ESM evaluated the group chunk before the helper initialized: `TypeError: __commonJSMin is not a function`.

This PR makes the reachability BFS also follow the entry-module facade edge (EntryPoint chunk → chunk physically containing its entry module). With it, the merge is rejected for such graphs and the runtime stays in a dependency-free leaf chunk — the same layout rolldown 1.0.3 emitted. A spurious edge (e.g. for an unwrapped entry with no re-exported symbols) only makes the merge decision more conservative; worst case is one extra `rolldown-runtime` chunk, never a broken output. No existing test snapshot changes.

Verified against the reporter's standalone repro (https://github.com/thevuong/rolldown-9993-repro): `npm run repro` fails on 1.1.3 and current main, passes with this fix.

### Out of scope (pre-existing, not part of the 1.1.1 regression window)

The fixture's chunk graph still contains a benign-looking static cycle (`v.js` ↔ `entry-2.js`) where `entry-2`'s facade calls the group chunk's top-level `require_node3` binding. Loading `entry-2.js` first is fine, but loading `entry-1.js` as the evaluation root still throws `require_node3 is not a function` — and rolldown **1.0.3 emits the same entry-1-first crash** for this input, so that cross-chunk evaluation-order defect predates this regression (it matches the reporter's follow-up observation that patching only the helper's emit form is insufficient in general). The new test documents this and pins only the regression.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
+++ b/crates/rolldown/src/stages/generate_stage/chunk_optimizer.rs
@@ -1549,6 +1549,10 @@ impl GenerateStage<'_> {
   ///   post-optimization graph.
   /// - Self-edges (`target_chunk == current`) are skipped — an intra-chunk
   ///   import can't form an inter-chunk cycle.
+  /// - Besides module import records, an entry-point chunk whose entry module
+  ///   was captured into another chunk (e.g. by a `codeSplitting` group) gets a
+  ///   render-time static import of that module's wrapper/namespace from the
+  ///   capturing chunk, so that facade edge is followed too (#9993).
   fn chunk_reaches_via_static_import(
     from: ChunkIdx,
     to: ChunkIdx,
@@ -1563,6 +1567,14 @@ impl GenerateStage<'_> {
       }
       if current == to {
         return true;
+      }
+      if let ChunkKind::EntryPoint { module: entry_module_idx, .. } =
+        chunk_graph.chunk_table[current].kind
+        && let Some(entry_module_chunk) = chunk_graph.module_to_chunk[entry_module_idx]
+        && entry_module_chunk != current
+        && !chunk_graph.post_chunk_optimization_operations.contains_key(&entry_module_chunk)
+      {
+        queue.push_back(entry_module_chunk);
       }
       for &module_idx in &chunk_graph.chunk_table[current].modules {
         let Some(module) = module_table[module_idx].as_normal() else {
--- a/crates/rolldown/tests/rolldown/issues/9993/_config.json
+++ b/crates/rolldown/tests/rolldown/issues/9993/_config.json
@@ -1,0 +1,13 @@
+{
+  "config": {
+    "input": [
+      { "name": "entry-1", "import": "./node4.cjs" },
+      { "name": "entry-2", "import": "./node3.cjs" }
+    ],
+    "preserveEntrySignatures": "allow-extension",
+    "codeSplitting": {
+      "includeDependenciesRecursively": false,
+      "groups": [{ "name": "v", "test": "node[0-4]" }]
+    }
+  }
+}
--- a/crates/rolldown/tests/rolldown/issues/9993/_test.mjs
+++ b/crates/rolldown/tests/rolldown/issues/9993/_test.mjs
@@ -1,0 +1,24 @@
+import assert from 'node:assert';
+
+// Regression test for #9993 (a regression of #9224): the runtime module must
+// not be merged into a chunk that is part of a static import cycle with a
+// chunk calling runtime helpers (e.g. `__commonJSMin`) at top level. Before
+// the fix, the entry chunk of `entry-2` hosted the runtime while the group
+// chunk `v.js` both consumed `__commonJSMin` at top level and was statically
+// imported back by `entry-2.js` (entry-2's entry module was captured into the
+// group chunk), so evaluating `entry-2.js` threw
+// `TypeError: __commonJSMin is not a function`.
+//
+// Note: entry-2 must be imported first here. Loading `entry-1.js` as the
+// evaluation root still throws `require_node3 is not a function` because
+// entry-2's facade calls the group chunk's top-level `require_node3` binding
+// before that chunk's body runs. That cross-chunk evaluation-order defect is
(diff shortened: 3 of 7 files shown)


IWANABETHATGUY (approved):

hyfdev:
### Merge activity

* **Jul 4, 5:20 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 4, 5:20 AM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 4, 5:25 AM UTC**: The [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) couldn't merge this PR because **it was not satisfying all requirements** (Failed CI: 'node-dev-server-test-ubuntu (24) / Node Dev Server Test').
* **Jul 5, 4:02 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 5, 4:43 PM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

Issue #9998, opened 2026-06-28
Manual codeSplitting groups leak cross-entry execution without strictExecutionOrder (#9463 follow-up)


hyfdev:
Follow-up to #9463 / #9997. #9997 closed the `strictExecutionOrder` case (and the rest of the strict family is isolated, because under `strictExecutionOrder` every side-effectful module is wrapped, so the only eager cross-entry path was the entry-chunk fold fixed there). The **same logical leak still exists without `strictExecutionOrder`**.

### Reproduction (verified)

```ts
// rolldown.config.ts — two entries, NO strictExecutionOrder
export default {
  input: { a: './a.js', b: './b.js' },
  experimental: { chunkOptimization: true },
  output: {
    codeSplitting: {
      groups: [{ name: 'common', entriesAware: true, entriesAwareMergeThreshold: 10_000 }],
    },
  },
}
```

```js
// a.js
import { foo } from './shared.js';
(globalThis.se ??= []).push('a');
foo();

// b.js
(globalThis.se ??= []).push('b');
import('./shared.js').then(({ foo }) => foo());

// shared.js
export function foo() {
  (globalThis.se ??= []).push('shared-foo');
}
```

Loading **only** the built `b.js` yields `se === ["a", "shared-foo", "b", "shared-foo"]` — entry `a`'s top-level side effect ran. Expected: `["b", "shared-foo"]`.

(The same config **with** `strictExecutionOrder: true` is correct — that is what #9997 covers.)

### Cause

A manual `codeSplitting` group co-locates modules of different entry-reachability into one chunk (its `bits` becomes a union — via `entriesAwareMergeThreshold` subgroup merging, or directly for a plain group). Under `strictExecutionOrder` the co-located modules are wrapped in `init_*` functions, so each entry only runs what it calls. Without it, the co-located modules are emitted eagerly, so loading one entry runs another entry's top-level code.

Two facets:

- **Entry modules** (the repro above): a genuine entry-isolation bug — executing entry `b` must never run entry `a`.
- **Non-entry modules**: the pre-existing manual-group over-inclusion tradeoff — e.g. `crates/rolldown/tests/rolldown/function/advanced_chunks/entries_aware_merge_shared_deps` already has `entry-a` load a chunk containing `lib-b`.

### Possible directions

- Make manual code splitting respect reachability boundaries (don't union `bits` across modules reachable from disjoint entry sets when it would co-locate eagerly-executed code).
[...]

Issue #9597, opened 2026-05-29
[Bug]: circular import generated when using dynamic import and require


sapphi-red:
### Reproduction link or steps

[REPL](https://repl.rolldown.rs/#eNqNkk1v2zAMhv8KodMGxI4/hrTzsGMw7LTDdpsHV7FpV50qpZLsJg3y36sPx3WKFAhgGCZFvnz4ygfSkuJAhGwwiesH7QNSzBILUtu443JD+Z97puOqovW+5qyuZG+2vana/uWlSuA7JN9KMUjWgMKnnin8VJJ46ZTy+EGX5LM9xt1WKqNjl5167AwkhVE9Hhd+cmrr5yQ+vhIktaLpCYQ9unETh98ogAxUgYdw9YejzYQoHijvcdQItHA4VboyMjJm7+0KiSspMzsh+8iuGeaZX6emE4KzdU7g4ysBcquVf2CTs3t06cKp3/OdiU7tzcR8MtFNODcxn0xUkvNGPou4lqJl3WyXCydhq+USfqBARQ02sNnDXU1VJ0H1AqItXFgTomjDBHRjU9WynekV2rR9NFqV2+wmS5Ms+7L6mmar9CZf3d5NzA22tOeWvRQATFjVInwDlASFUfsoKUlhg7OLcxXHhXvjzqASlBfw1zaMgev554+NQtT39D8W0FKu0Se3CjWqAddO/zfrBHXI2k2hnMvnyOkIzaQoiW8I+05k2ihWm/UO697Yol+qQTXXB3hkgrX7nyLgrMMvNi9x8OGa7D0N1nhuzdOGHF8BDZ9o+Q==)

### What is expected?

No error happens when executing `dist/entry-0.js`.

### What is actually happening?

The following error happens when executing `dist/entry-0.js`:
```
TypeError: __exportAll is not a function
```
In other words, the cyclic import "entry-0 -> node0 -> entry-0" is not output.

### System Info

```Shell
REPL 1.0.3
```

### Any additional comments?

This seemed to be working with 1.0.0.
Found at https://github.com/sapphi-red/zarara/issues/25#issuecomment-4560201943

PR #10104, opened 2026-07-03
feat(code-splitting): wrap strict execution order modules on demand

### LINKED ISSUE #9887: [Bug]: `advancedChunks` + CJS + `includeDependenciesRecursively: false` → cross-chunk wrapped-ESM init cycle (`init_X is not a function`); `strictExecutionOrder` does not help

A REPL link can't carry the `advancedChunks` config, and this bug *requires* it, so here is a self-contained copy-paste reproduction.

`package.json`
```json
{ "name": "repro", "private": true, "type": "module", "devDependencies": { "rolldown": "1.1.2" } }
```

`shared.js`
```js
export const extend = Object.assign
```

`dep.js`
```js
import { extend } from "./hub.js" // imports back from the hub chunk -> cross-chunk cycle
export const useDep = () => extend({}, {})
```

`interop.cjs`
```js
require("./shared.js")
require("./dep.js")
```

`hub.js`
```js
import "./interop.cjs"
export * from "./shared.js"
export * from "./dep.js"
```

`main.js`
```js
import { useDep, extend } from "./hub.js"
export const go = () => [useDep, extend]
```

`build.mjs`
```js
import { rolldown } from "rolldown"
import { pathToFileURL } from "node:url"
import { rmSync } from "node:fs"

const groups = [
  { name: "dep", test: "dep\\.js$" },
  { name: "hub", test: "(hub|interop|shared)\\." },
]

for (const [label, opts] of [
  ["includeDependenciesRecursively:false", { incdep: false }],
  ["includeDependenciesRecursively:false + strictExecutionOrder:true", { incdep: false, seo: true }],
  ["includeDependenciesRecursively:true  (control)", { incdep: true }],
]) {
  const dir = "dist-" + label.replace(/[^a-z]+/gi, "_")
  rmSync(dir, { recursive: true, force: true })
  const b = await rolldown({ input: { main: "./main.js", hub: "./hub.js" }, platform: "browser", logLevel: "silent" })
  await b.write({
    format: "esm", dir, entryFileNames: "[name].js", chunkFileNames: "[name].js",
    strictExecutionOrder: opts.seo === true,
    advancedChunks: { includeDependenciesRecursively: opts.incdep, groups },
  })
  await b.close()
  try { await import(pathToFileURL(`${process.cwd()}/${dir}/main.js`)); console.log("  OK   ", label) }
  catch (e) { console.log("  THROW", label, "->", e.message) }
}
```

Run:
```sh
npm install
node build.mjs
```

Output (rolldown 1.1.2):
```
  THROW includeDependenciesRecursively:false                                 -> init_dep is not a function
  THROW includeDependenciesRecursively:false + strictExecutionOrder:true     -> init_dep is not a function
  OK    includeDependenciesRecursively:true  (control)
```

### What is expected?

Importing the entry runs without error (as it does with `includeDependenciesRecursively: true`), or rolldown errors at build time instead of emitting code that throws at runtime.

`TypeError: init_dep is not a function` at load.

The `hub` chunk imports `init_dep` from the `dep` chunk and **calls it eagerly at module-eval time**, but `dep` imports `extend` back from the `hub` chunk, so the two chunks form an **import cycle** and `hub` runs before `dep` has assigned `var init_dep = __esmMin(...)`. The import binding is present — the wrapper is simply called before it is initialized:

```js
// hub chunk (generated)
import { n as init_dep, r as useDep } from "./dep.js";
var require_interop = /* @__PURE__ */ __commonJSMin((() => {
	init_shared();
	init_dep();          // <-- init_dep is still undefined here
}));
...
require_interop();
init_dep();
```

Notes:

- **`output.strictExecutionOrder: true` does not fix this variant.** It changes the emitted output but the cycle and the premature call remain. (As a control, `strictExecutionOrder: true` *does* fix the simpler CJS + `advancedChunks` order bug in #7449.)
- **`includeDependenciesRecursively: true` (the default) fixes it** — the dependency is pulled into the dependent's chunk, so no cross-chunk cycle forms.
- **Not a regression.** Reproduces on rolldown 1.0.2, 1.0.3, 1.1.0, 1.1.1, and 1.1.2 (latest).

Minimal trigger (each ingredient is necessary — removing any one makes the output correct):

1. a CJS module that `require()`s the ESM modules (forces `__esmMin` wrapping + an eager init call);
2. `advancedChunks` groups that place a dependency in a different chunk than a module that depends back into the first chunk (a cross-chunk cycle);
3. `includeDependenciesRecursively: false`;
4. an eager init at module-eval time (here, the CJS `require_*()`).

The namespace-import shape is not required (named imports reproduce it too).

### System Info

```
rolldown 1.1.2 (also 1.0.2 / 1.0.3 / 1.1.0 / 1.1.1)
node 24.x, macOS arm64
```

### Any additional comments?

This is the same failure class as #7449 (*incorrect execution order with CJS + advancedChunks*) and #8803 / #9630, but specifically the case the documented `strictExecutionOrder` workaround (#8803) does **not** cover, because `includeDependenciesRecursively: false` forces the cross-chunk cycle. `includeDependenciesRecursively: false` is a meaningful real-world setting (it improves vendor-chunk content/hash stability), so a manual-chunking config with any CJS dependency in a cycle can hit silent, runtime-only breakage with no working escape hatch.

Minimum bar would be to error at build time rather than emit code that throws at load. Filing separately from #7449 since the workaround gap is specific to `includeDependenciesRecursively: false`; happy to fold it into #7449 if preferred.

### LINKED ISSUE #10236: [Bug]: CJS output + on-demand strict execution order emits an undefined entry init call

This requires CJS output, so here is a self-contained reproduction.

`package.json`

```json
{
  "private": true,
  "type": "module",
  "scripts": {
    "repro": "node build.mjs"
  },
  "devDependencies": {
    "rolldown": "1.1.5"
  }
}
```

`module-0000.mjs`

```js
export const vdef = { v: 0 };
```

`module-0001.mjs`

```js
export { vdef as vdef2 } from "./module-0000.mjs";
```

`module-0002.mjs`

```js
import { vdef as direct } from "./module-0000.mjs";
import { vdef2 as forwarded } from "./module-0001.mjs";

globalThis.__orderEvent({
  module: "a",
  value: 1352912 + (direct === forwarded ? 0 : 987654321),
});
```

`module-0003.mjs`

```js
import { vdef } from "./module-0000.mjs";

globalThis.__orderEvent({
  module: "b",
  value: 2765533 + (vdef === vdef ? 0 : 987654321),
});
```

`build.mjs`

```js
import { rmSync } from "node:fs";
import { createRequire } from "node:module";
import { dirname, join } from "node:path";
import { fileURLToPath } from "node:url";
import { rolldown } from "rolldown";

const root = dirname(fileURLToPath(import.meta.url));
const dist = join(root, "dist");

rmSync(dist, { recursive: true, force: true });

const build = await rolldown({
  input: {
    __entry_0000: join(root, "module-0002.mjs"),
    __entry_0001: join(root, "module-0003.mjs"),
  },
  preserveEntrySignatures: "allow-extension",
  experimental: {
    onDemandWrapping: true,
  },
});

await build.write({
  dir: dist,
  format: "cjs",
  codeSplitting: true,
  strictExecutionOrder: true,
  entryFileNames: "entries/[name].cjs",
  chunkFileNames: "chunks/[name].cjs",
});

await build.close();

globalThis.__orderEvent = (event) => console.log(event);
require(join(dist, "entries/__entry_0000.cjs"));
```

Run:

```sh
npm install
npm run repro
```

Rolldown 1.1.5 throws:

```text
ReferenceError: init_module_0002 is not defined
```

### What is expected?

Loading the generated CommonJS entry should complete without error and execute the entry once. Rolldown must not emit a call to an `init_*` function that is absent from the entry chunk.

With `format: "cjs"` and `experimental.onDemandWrapping: true`, the generated entry chunk calls `init_module_0002()`, but that function is neither defined nor imported in the chunk. Loading the entry therefore throws `ReferenceError: init_module_0002 is not defined`.

### System Info

```shell
System:
  OS: macOS arm64

Binaries:
  Node: 24.18.0

npmPackages:
  rolldown: 1.1.5
```

### Any additional comments?

The failure is specific to the CJS-output/on-demand cell:

- CJS output + on-demand wrapping on `rolldown@1.1.5`: fails
- CJS output + wrap-all: passes
- ESM output + on-demand wrapping: passes
- CJS output + on-demand wrapping on the frozen #10104 snapshot at `42628c18b`: passes

The directed fuzzer cell reproduced the same failure for 8 seeds. The seeds only vary numeric payloads; the module graph is identical.

Evidence: https://github.com/hyf0/rolldown-order-fuzzer/blob/853fb775d1e949ced567deb0c163d044a031ba4f/.agents/evidence/cjs-output.json

### Fix status

The implementation fix is already in #10104 and was verified on its frozen snapshot at `42628c18b`. This issue documents a bug in the released `rolldown@1.1.5`; it does not request a separate fix on `main`.

After the rebased PR head is re-verified against this reproduction, #10104 should close this issue with `Fixes #10236`.

hyfdev:
## Summary

This PR adds an opt-in on-demand planner for `output.strictExecutionOrder` while preserving both the flag-off path and strict mode's existing wrap-all default.

| Configuration | Execution-order behavior |
| --- | --- |
| `strictExecutionOrder: false` | No execution-order wrapping |
| `strictExecutionOrder: true` | Wrap every eligible module (wrap-all, the conservative default) |
| `strictExecutionOrder: true` + `experimental.onDemandWrapping: true` | Derive a conservative wrapping plan from the generated chunk graph |

The on-demand plan is deliberately conservative rather than minimal. It begins with predicted source-order versus chunk-order hazards, then closes over importers, trigger hosts, retained re-export paths, and cyclic chunks where eager evaluation cannot be proven safe.

This PR does not change the default behavior of `strictExecutionOrder` and does not enable strict execution order globally.

Fixes #10236.

> [!IMPORTANT]
> Strict-order output may change the entry-chunk shape. When an entry implementation must remain inert while another chunk is evaluated, Rolldown emits a small entry facade that calls the relevant `init_*()` function and moves the implementation into a separate chunk. Reviewers and downstream tooling should expect this facade/implementation split.

## Why the analysis runs after chunking

Interop wrapping answers how a module is represented and remains a link-stage decision. Execution-order wrapping answers when an already-included module body may run, which depends on generated chunk placement, manual groups, facade elimination, shared entries, and chunk cycles.

This PR keeps link-owned `WrapKind` and user-code liveness immutable, computes an `OrderWrapPlan` after provisional chunking, and lowers that plan into a separate `OrderWrapState`. Lowering may add synthetic wrappers, init calls, runtime symbols, entry facades, and topology state; it cannot make an excluded user statement or binding live.

The on-demand path then projects imports added by lowering back into the chunk graph. If those imports create a new chunk cycle, every eligible module in the cyclic chunks is added to the plan and projection repeats until stable. The process is monotone and bounded by the module count.

## Correctness hardening

- Preserve init obligations through re-export barrels without reviving tree-shaken imports or unrelated exports.
- Share init-obligation enumeration between emission, cross-chunk registration, and fixpoint projection so their record gates cannot drift.
- Preserve the same link-stage tree-shaking result in wrap-all and on-demand, including consumer-local named and namespace re-export paths.
- Keep cross-chunk `init_*` bindings live and deconflicted for both ESM and CJS output.
- Restore user, dynamic, and emitted entry facades after strict lowering changes chunk topology.
- Recompute topology-derived metadata and final output validation after lowering.

Tree-shaking parity here is specifically between wrap-all and on-demand. It does not claim byte identity with flag-off output and does not include the separate `wrappedModuleTreeshaking` work.

## Review guide

1. Start with `internal-docs/code-splitting/design.md` for the invariants, mode split, conservative decisions, and projected/omitted edge inventory.
2. Read `order_analysis.rs` for source-order versus chunk-order prediction and the emergent-cycle fixpoint.
3. Read `order_wrap_state.rs` and `order_wrapping.rs` for the late synthetic state and lowering boundary.
4. Read `esm_init_obligations.rs` together with the module finalizers and cross-chunk-link pass for the shared emit/register/project contract.
5. Review #10287 for config/variant normalization and the current-main output baseline.
6. Review this PR's test diff for removal of every temporary expected-failure marker and the same-position snapshots of the final facade/runtime/chunk layouts.

## Test-first sequence

The earlier test-first layers are already merged into `main`:

1. #10253 adds the fail-closed runtime-failure harness and core integration regressions.
2. #10252 adds the later hardening fixtures and paired-mode coverage.

The remaining Graphite stack is:

1. #10287 establishes the final config/variant review layout and current-`main` snapshots while preserving all expected-failure markers.
2. #10104 (this PR) adds the implementation, removes all temporary failure markers, and records the implementation output.

The final tree contains 96 strict-enabled fixture configs and 200 named strict executions. Between #10287 and #10104, 21 config files only delete 44 expected-failure markers, and all 93 changed snapshots retain identical variant-heading order.

## Validation

Local validation ran on rebased head `79efe909bb3f0bcd2abcc76dcd33e58069c3474c`, based on `main` at `c4e25974b8ff7598bf507ae4a2d7602cbcbc86fd`.

- The 86 changed fixtures passed twice on #10287 with the old implementation; the second pass was snapshot-stable.
- All 96 strict-enabled fixture configs passed twice on this implementation, with a snapshot-stable final pass.
- Full Rust workspace tests passed with `just test-rust`.
- Rolldown Node tests passed: 793 main tests and 38 watcher tests.
- The Rollup compatibility suite passed 1,213 tests with 0 failures.
- `cargo check --workspace --all-features --all-targets --locked` and `cargo fmt --all --check` passed.
- Clippy passed with `--deny warnings` for every changed Rust crate: `rolldown`, `rolldown_common`, and `rolldown_ecmascript_utils`.
- `just lint-node` and `just lint-repo` passed.
- Full-workspace `just lint-rust` is locally blocked on Rust 1.97 by `clippy::ignore-without-reason` in the unchanged `rolldown_plugin_isolated_declaration/tests/mod.rs`; the failing path is byte-identical to `main`.

## Frozen closure evidence

The broader correctness campaign was frozen on one build of `42628c18b`:

- Differential fuzzer full matrix: 3,000/3,000 passed (ESM/CJS/mixed inputs × wrap-all/on-demand × 500 cases).
- Directed barrel/definer family: 600/600 passed; the same generated shapes failed about 23.7% of the time before the fix.
- Directed eager-barrel family: 20/20 passed in wrap-all and 20/20 in on-demand.
- #9887 strict-path bracket passed on the branch and failed 20/20 on released `rolldown@1.1.5`. This PR restores strict mode as an escape hatch; the non-strict path remains a separate issue, so this PR does not close #9887.
- #10236 CJS-output/on-demand bracket passed on the branch and failed all 8 directed seeds on released `rolldown@1.1.5`.
- ComfyUI, Camunda, shadcn-admin, and vue-vben-admin passed the flag-off/on-demand/wrap-all browser-probed matrix.

Fuzzer source and machine-readable evidence: https://github.com/hyf0/rolldown-order-fuzzer/tree/853fb775d1e949ced567deb0c163d044a031ba4f/.agents/evidence

## Boundaries and known residuals

- Top-level-await output remains legal, but this PR does not promise stronger TLA scheduling semantics than the default build.
- External-module evaluation order remains outside the strict-order guarantee; generated output must still be legal.
- #9998 is a separate bug in the non-strict/default code-splitting path and is not fixed here. The strict path covered by this PR passes its directed controls.
- Camunda still emits two `init_*` functions with no call site (`OverflowMenu$1` and `RadioButton$1`). They were harmless on all probed routes and have no synthetic reproduction.
- The partial-forwarder per-obligation refinement is present in the shared contract, but no constructible corpus shape currently exercises that narrower branch directly.
- On vue-vben-admin, the emergent-cycle fixpoint converged in two rounds and added 141 wrappers over the one-shot plan. `ROLLDOWN_ORDER_DEBUG=1` reproduces the per-round trace.


IWANABETHATGUY:
I recommend splitting all newly added tests into a separate PR to make the PR review more manageable.

hyfdev:
* **#10104** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10104?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#10298](https://github.com/rolldown/rolldown/pull/10298) <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10298?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#10307](https://github.com/rolldown/rolldown/pull/10307) <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10307?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10104?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10287** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10287?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10104).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyfdev:
> I recommend splitting all newly added tests into a separate PR to make the PR review more manageable.

Done

hyfdev:
## Multi-angle review (Claude Fable 5)

> This review was produced by Claude (Fable 5) at hyf0's request: six specialized review passes (architecture/layering, order analysis, lowering & chunk topology, init-obligation contract & emission, API/perf, tests) plus a manual read of the core files, against head `408c5b3` / diff base `bb18ff9`. Everything below was verified against the actual code, not the PR description.

**Verdict: no blocker found.** One cheap pre-merge hardening is recommended, plus two doc/comms items. The architecture migration is real and the PR's headline claims all reproduce independently.

### Claims verified

- `override_wrap_kind()` / `hoist_esm_wrapper` are gone repo-wide; the link-stage strict-mode `WrapKind` override loop is fully removed. Link-stage file changes are genuine link facts (`execution_dependencies`, star re-export topology), not order logic leaking in.
- Test pairing claim reproduces exactly: 94 strict fixtures / 200 strict execution units / 100 exact wrap-all↔on-demand pairs / 0 unmatched, each pair differing only in the mode flag. All 44 `expectExecutionFailure` markers and 8 `#[should_panic]`s are removed and flipped to assert the fixed behavior.
- Flag-off output is byte-identical across every changed snapshot (the one suspicious case, `issues/9374`, is a snapshot-sectioning false positive).
- Fixpoint termination holds as documented: `all_at_risk` is monotone, hard-capped at module count, fail-closed `assert!` on non-convergence. Output is deterministic — every render-order-affecting consumer re-sorts by unique `exec_order`.
- The emit/register/project contract is genuinely one shared enumeration (`record_is_init_obligation` + `for_each_init_obligation_record` + shared collectors), and the CJS live-getter + hoisted `function init_*()` shape directly fixes the #10236 class.

### Recommended before merge

**Tighten `eager_forwarder_discharges_own_hops` (`esm_init_obligations.rs:503`) to match the Emit gate.** The Emit/Register gate is `stmt included && !nested`, but this full-delegation fast path checks inclusion bits only. The in-code OPEN QUESTION comment argues a nested + included + directly-consumed hop is unconstructible, but the supporting premise is weaker than stated: `module_owns_reexport_init` is false for *any* eager unplanned module, so eager forwarders' re-export records do get marked nested on retained star paths — the only remaining defense is the unproven "never simultaneously included and directly consumed". If it ever fires, the failure is a silent never-called `init_*` (the exact #10236 drift direction). The fix is near-zero cost: requiring records to also be non-nested makes a nested hop fall into the partial-delegation path (`forwarder_discharged_targets`), which already handles nested correctly; the conservative direction only ever adds a redundant memoized init call. This converts a documented assumption into a structural guarantee.

### With the merge

- **Docs point at a hidden option:** the public `strictExecutionOrder` JSDoc and both user-facing markdown docs reference `experimental.onDemandWrapping`, which is still `@hidden not ready for public usage yet` (`input-options.ts:651`). Either drop the reference or unhide.
- **Changelog note for existing strict users:** wrap-all output shape changes (entry facade split + separate runtime chunk; measured 1→3 and 2→5 chunks on existing fixtures). The "does not change default behavior" claim only covers flag-off; current `strictExecutionOrder` users see a visible shape change.

### Non-blocking follow-ups

- The projection's nested-record skip is guarded only by a `debug_assert!` (`order_analysis.rs:403`); release builds silently drop the edge, and under-projection is the crash direction. Consider a hard assert (or emit the edge) — the comment itself flags the dormant concatenated-inner corner.
- Strict coverage gaps: 0 fixtures for sourcemap, `preserveModules`, `inlineDynamicImports`, iife/umd, watch/incremental. Incremental + strict (generate-stage facade symbols entering the cache) is the least-certain combination.
- 7 strict fixtures are snapshot-only (`expectExecuted: false`), including `on_demand_wrapping` itself, whose embedded `strictEqual(a, 2)` never actually runs.
- No strict-mode benchmark exists; the double link pass + fixpoint run entirely outside Codspeed coverage.
- Smaller items: possible facade-split miss for an entry co-located into another entry's chunk by a manual group (low confidence — worth a directed fixture); `restore_order_wrap_entry_facades` resurrects all matching facades but wires only the first (`debug_assert!(len <= 1)` would lock the invariant); `chunk_is_live` vs the renderers disagree on `RemovedWithPreservedExports` (benign only because such chunks are always empty — undocumented); the actual-order traversal is recursive where the expected-order walk was deliberately made iterative; `onDemandWrapping` without `strictExecutionOrder` is a silent no-op with no warning; one dead exported `init_forwarder` in `emergent_cycle_partial_forwarder` (the disclosed Camunda-residual class).

### Architecture note

"Immutable link state" is currently enforced by convention, not the compiler: `set_wrap_kind` is still `pub`, inclusion/wrapper fields are `pub`, and generate-stage code holds `&mut metas`. No generate-stage caller exists today (grep-verified) and `finalize_chunk_plan`'s debug asserts cover planned modules, but a future generate-stage mutation would compile cleanly. Sealing these to `pub(in link_stage)`-level visibility would turn the strongest design claim into a compile error.

crates/rolldown/tests/rolldown/issues/9463/artifacts.snap
```diff
@@ -88,29 +100,49 @@ init_b();
...
+export { __exportAll as n, __esmMin as t };
+
+```
+
+### shared.js
+
+```js
+import { i as init_shared, r as foo } from "./common~a~b~shared.js";
+init_shared();
+export { foo };
```
IWANABETHATGUY, line 146: regression: generate two more chunks than before.

crates/rolldown/tests/rolldown/issues/9463/artifacts.snap
```diff
@@ -31,31 +31,43 @@ init_b();
...
 //#endregion
 export { init_a as n, init_b as t };
 
 ```
 
+## rolldown-runtime.js
+
+```js
+// HIDDEN [\0rolldown/runtime.js]
+export { __exportAll as n, __esmMin as t };
```
IWANABETHATGUY, line 67: regression: generate one more chunk than before.

crates/rolldown/tests/rolldown/function/experimental/strict_execution_order/concatenate_basic/artifacts.snap
```diff
@@ -51,16 +54,39 @@ source: crates/rolldown_testing/src/integration_test.rs
...
+
+### main2.js
+
+```js
+import { t as __esmMin } from "./rolldown-runtime.js";
 //#region setup.js
-var init_setup = __esmMin((() => {
-	globalThis.globalValue = "foo";
-}));
+function init_setup() {
```
hyfdev, line 67: Is this necessary? we don't have this before?

crates/rolldown/tests/rolldown/issues/9463/artifacts.snap
hyfdev: The extra chunk is the runtime eviction, and it's a recoverable regression rather than a required shape. On `main` the runtime starts standalone and `try_merge_runtime_chunk()` folds it into a host chunk (here `common~a~b~shared`) once that's proven safe. Order lowering runs after that proof and can add helper consumers (`__esmMin` for every `init_*` wrapper) and new chunks, so the proof goes stale; `ensure_runtime_module_for_order_wraps` currently re-evicts unconditionally instead of re-proving. Eviction is always safe (a runtime-only chunk executes no user code and can't join a cycle — the #9597 class), but in this fixture the only helper consumer is the common chunk itself, so folding back is provably safe. The fix is to re-run the merge proof after lowering with the post-lowering consumer set — order state already knows each wrapper's chunk and helper demand. Tracked in #10294 (item 1).

crates/rolldown/tests/rolldown/issues/9463/artifacts.snap
hyfdev: Two separate deltas here, both wrap-all-only — the on-demand cell keeps `shared.js` unwrapped and collapsed:

1. `rolldown-runtime.js` — same runtime eviction as the other thread; recoverable by re-running the fold proof after lowering.
2. `shared.js` — wrap-all wraps `shared.js` exactly as it did before this PR (the old output also had the empty `init_shared`); what changed is trigger placement, which is now a single rule: an order-wrapped entry's activation lives in an entry facade, unconditionally (`create_order_wrap_entry_facades`). That forces the dynamic entry into its own chunk, which in turn defeats the same-chunk `import()` collapse, so the old `Promise.resolve().then(() => (init_shared(), shared_exports))` shape (correct for this graph, zero extra chunks) is no longer produced. Restoring it means either teaching the dynamic-import collapse to emit the init call through the shared `esm_init_target` view when the target is order-wrapped, or more narrowly skipping the facade when the entry's init is a noop with no transitive init targets (true for `shared.js` here). Tracked in #10294 (item 2).

crates/rolldown/tests/rolldown/function/experimental/strict_execution_order/concatenate_basic/artifacts.snap
hyfdev: Answering my own question after digging in (cc @IWANABETHATGUY, related to the chunk-count threads):

Yes — the hoisted-function form is load-bearing for on-demand, not cosmetic. Before this PR every `init_*` caller was itself deferred (another wrapper body or an entry tail), so `var init_x = __esmMin(...)` was always assigned before any call. On-demand breaks that premise: eager (unwrapped) modules call `init_*()` at their record positions while chunk bodies are still evaluating. The `var` form then crashes in two real shapes:

- a module-level import cycle placed in one chunk makes one direction a textual forward reference — an eager caller hits the not-yet-assigned `var` (`init_x is not a function`);
- inside a chunk cycle, a sibling chunk calls an `init_*` before the defining chunk reached its assignment (the vue-vben-admin `qe is not a function` class); CJS output pairs this with live-getter exports.

A hoisted `function init_x() { return (init_x = __esmMin(cb))(); }` is callable from scope instantiation, which makes correctness position-independent — that's also exactly the WebKit out-of-order-evaluation failure in #10280, which `var`-form helpers cannot survive. Interop wrappers keep the `var` form since their callers are still all deferred. Making the form conditional on predicted topology would save ~20–30 bytes per wrapper post-minify at the cost of making the output shape depend on prediction again — the exact failure mode this PR removes.

hyfdev: One clarification on the #10280 citation above, since it can read as if Safari were the motivation. It isn't — #10280 is only the acyclic, WebKit-specific member of the family (a spec-conforming host can't hit that shape; WebKit's concurrent-`import()` reordering does). The form is required by engine-independent, spec-deterministic failures:

- #10265: released rolldown 1.1.5, `onDemandWrapping`, **single chunk** — no code splitting, no chunk cycle — `TypeError: init_a is not a function` on any engine, because the generated `init_a()` call site runs before the later `var init_a = __esmMin(...)` assignment in the same scope. Plain `var` hoisting, deterministic REPL repro.
- At the review-baselines commit ([`fb81314`](https://github.com/rolldown/rolldown/commit/fb8131409ab67b947dc0c6227c9fbdb56e2a9294)), `wrapped_entry_chunk_cycle` (4 cells, `outputContains: ["is not a function", ...]`) and `m4_dynamic_facade_dependency` (`TypeError: init_m29 is not a function`) were marked expected execution failures — those fixtures execute under **Node** in CI and crashed there with the `var` form; this PR's hoisted form (plus the facades) is what flipped them to passing. No browser involved: within a chunk cycle some chunk always evaluates before its sibling per spec, and a `var`-form init referenced at that moment is still unassigned.

So Safari is the bonus, not the reason: reverting to `var` would resurrect Node-reproducible startup crashes that are in the executed test suite today, and #10280 would additionally stay broken on WebKit.

hyfdev: Follow-up recorded in #10294 (item 5): narrow the hoisted form to the hazardous patterns (same-chunk forward references, cyclic chunk graphs) and keep the `var` rendering elsewhere once proven safe.

hyfdev:
## Deep review (Claude Fable 5, max effort — "Fable Max")

> Produced by Claude Fable 5 running at maximum reasoning effort at hyf0's request, against head `220783a` (merge-base `03e1e34`). Method: direct reads of `order_analysis.rs`, the generate-stage orchestration, the prediction path, the scanner change, and representative snapshots, plus five parallel deep-read passes (lowering/state, init-obligation contract, cross-chunk links/rendering, scanner/options/docs, test audit); every major finding below was re-verified by hand in the code. This complements the earlier Fable 5 multi-angle review — overlapping items are referenced, not repeated.

**Verdict: the architecture is the right one and I found no blocker. Two majors are worth handling before merge (an analyzer soundness gap, and a set of ungated flag-off deviations), plus two cheap fixes (cross-target determinism, a quadratic strict-mode hot loop) and one structural assert I'd add. Everything else is follow-up material, much of it already tracked in #10294.**

### Claims re-verified at this head

- The earlier review's pre-merge recommendation is applied: `eager_forwarder_discharges_own_hops` now also requires records to be non-nested (`esm_init_obligations.rs:508-524`).
- 44 expected-failure markers across 21 configs are pure deletions; the 8 `#[should_panic]`s in `strict_execution_order_invariants` flipped to enforced-passing; all 93 changed snapshots keep byte-identical variant-heading order; a mechanical sweep of 686 strict/TLA/invariants asset blocks found 0 `init_*` calls that are neither defined nor imported in their chunk.
- `override_wrap_kind`/`hoist_esm_wrapper` are gone; every `set_wrap_kind` caller is link-stage; lowering's mutable surface is only `symbol_db` + `OrderWrapState`.
- Determinism inside the order files holds — every hash-set-driven effect is either sorted on unique `exec_order` first or lands in an order-insensitive sink — with one escape (item 3 below).
- The fuzzer evidence at `hyf0/rolldown-order-fuzzer@853fb77` exists and pins tested binary SHAs. The 3,000-case matrix is frozen on `42628c18b`, and this head differs from the twice-validated `79efe909` only by rebase carry-over on 6 files (core order files byte-identical) — one fuzzer re-run on the final head would close the loop.

### Major

**1. Order sensitivity misses definition-time global reads in class positions (silent-reorder direction).**
`is_order_sensitive` collects `GlobalVarAccess` reasons only in the expression arms the analyzer handles itself; the oxc-delegate arms return a bare boolean: class declarations (`stmt_eval_analyzer/mod.rs:462-467`), `export default class` (`:524-526`), and the expression catch-all (`:399-404`). So `var x = Object.freeze({})` is order-sensitive while `class C { static x = Object.freeze({}) }` is not — inconsistent by AST position. A module whose only eval-time work is a whitelisted-global read in a class position is invisible to the expected/actual comparison and can be silently displaced across a module that patches that global (polyfill-then-use). This is pre-existing analyzer behavior, but the PR widens its blast radius from "import-free leaf modules" (old on-demand) to "every planless module"; wrap-all is immune. Suggest: collect reasons through class definition-time positions (or conservatively mark classes with static initializers / computed keys / heritage as order-sensitive), add a `get_stmt_order_sensitivity` unit test for `class C { static x = Math.max(1, 2) }`, and add a fuzzer family that patches a whitelisted global before a class-static-field reader.

**2. Four flag-off deviations sit outside the byte-identical claim's coverage** (every changed snapshot is a strict config, so none of these are pinned by tests):

- The **entry-level** `add_transitive_esm_init_depended_symbols` call (`compute_cross_chunk_links.rs:471-476`) has no strict gate, and legacy interop `transitive_esm_init_targets` exist flag-off. For an interop-wrapped entry behind a facade chunk, with an excluded re-export whose wrapped targets share the entry's *host* chunk, the facade gains a dead cross-chunk `init_*` import that base never emitted. Cheapest fix: gate this call on `is_strict_execution_order_enabled()` (the per-module call is provably inert flag-off); otherwise a fixture proving the shape unreachable.
- The shared collector now serves the flag-off interop path (`module_finalizers/mod.rs:216`) with two behavior changes vs. the base collector: star records iterate **sorted non-ambiguous** resolved exports (ambiguous-export owners no longer receive init calls; emission order changed from hash order to name order), and named/namespace specifiers are **liveness-gated per importer** (an `ns.x`-only consumer inits only `x`'s owner where base inited every wrapped exporter). Plausibly intended tree-shaking parity — but whether a dropped per-site init can ever drop a side effect deserves a directed flag-off fixture, or an explicit "known behavior fix" note in the PR body.
- `await_if_tla` flipped `false → true` for excluded-statement transitive init calls (`module_finalizers/mod.rs:1644`). Reads as a genuine latent-bug fix, but it changes flag-off output with no flag-off test naming it.
- Tombstone chunks no longer consume `[name]` dedup slots once the sweep triggers `rebuild_sorted_chunk_idx_vec(true)` (`finalize_chunk_plan.rs:47-50`) — byte-visible only under a `[name]` collision with a tombstone. Arguably an improvement; worth one sentence of acknowledgment.

### Cheap pre-merge fixes

**3. Cross-target output divergence (native vs wasm32/WASI).** `add_transitive_esm_init_depended_symbols` iterates `FxHashMap::values()` (`compute_cross_chunk_links.rs:623`) into the chunk's `depended_symbols`, whose insertion order feeds imported-symbol rename order (`$1/$2` suffixes) in `deconflict_chunk_symbols`. FxHash is unseeded but hashes differently on 32-bit vs 64-bit targets, so native and WASI builds can resolve rename collisions differently. Three independent review passes converged on this single site; every other hash-ordered effect is sorted before use. Fix: iterate keys sorted by `StmtInfoIdx` (or store an `FxIndexMap`).

**4. Quadratic strict-mode hot loop.** `module_is_in_live_chunk` is a linear scan of `chunk.modules` (`chunk_graph.rs:68-73`), called from `add_depended_symbol_with_wrapped_esm_init` (`compute_cross_chunk_links.rs:557-561`) for every canonical referenced symbol of every included statement under strict. A 10k-module chunk times ~100k symbol references goes quadratic. `module_to_chunk[m] == Some(c) && chunk_is_live(c)` plus an O(1) removed-modules set is equivalent and constant-time.

### One structural assert I'd add

The emergent-cycle projector's soundness rests on mirroring the three registration paths, and one consumption disjunct has no projection counterpart: the `referenced_symbols_by_entry_point_chunk` arm of `collect_frozen_reexport_usage` (`order_wrapping.rs:693-698`). No concrete break was constructed, but under-projection is the crash direction this machinery exists to prevent. Proposal: after the **final** cross-chunk link pass, run Tarjan over the final `imports_from_other_chunks` graph and assert every order-wrap-eligible module in a nontrivial SCC is in the plan. Debug-only is fine and it's cheap; it converts any future projector hole from a fuzzer-only runtime crash into a deterministic build failure. Same genre as the facade zero-static-indegree assert this PR already added (`compute_cross_chunk_links.rs:193-214`) — a structural invariant, not the internal semantic verifier `design.md` rightly rejects.

### Minor / follow-up (listing for the record; several belong next to #10294)

- `ConcatenateWrappedModuleKind` + `Chunk::module_groups` now have zero writers but ~12 live-looking read sites, including an unreachable render path (`format/esm.rs:225`). Sweep the machinery or centralize the "dormant on this branch" note. Related dead arm: the disabled-splitting `DynamicImport` overlay (`order_wrap_state.rs:536-539`) is unreachable — splitting-disabled dynamic importees are always interop-wrapped.
- Real lowering filters plan members by `wrap_kind() != None` (`order_wrapping.rs:558`) while the fixpoint probe uses the full `is_order_wrap_eligible` predicate (`order_analysis.rs:553-557`). Both filters are dead today, but a plan-invariant regression would mint wrappers the projector never modeled — silently, in release. Same predicate (or a hard assert) on both sides closes the drift channel.
- `restore_order_wrap_entry_facades` un-removes every matching facade chunk but wires only the first (`order_wrapping.rs:348-391`); `debug_assert!(candidates.len() <= 1)` would pin the invariant the code relies on.
- `execute_actual_module` (`order_analysis.rs:752-785`) recurses per interop importee; the sibling expected-order walk was deliberately converted to an explicit stack. Deep CJS require chains are exactly where strict mode gets deployed.
- Fixpoint-cap violation is a release `assert!` (`order_analysis.rs:227`). Consider degrading to wrap-all plus an internal-error diagnostic instead of a build crash — conservative output beats a panic if the monotonicity invariant ever regresses.
- `experimental.onDemandWrapping` without `strictExecutionOrder` is a silent no-op; a generate-stage warning would prevent inert configs. Also `input-options.ts:652` still marks the option `@hidden not ready for public usage yet` while two public docs recommend it by name — pick one.
- The scanner now reads an output option (`is_strict_execution_order_enabled`, `ast_scanner/impl_visit.rs:95`). Safe today because every `write()` rebuilds from merged options, but worth a comment so future scan caching across outputs doesn't silently break `TopLevelImportRead` completeness.

### Production-readiness beyond the code

- **The feature's value proposition has no numbers.** No wrap-all vs on-demand vs flag-off size deltas, no build-time cost of the doubled link pass + fixpoint on a large app; Codspeed exercises flag-off only. `ROLLDOWN_ORDER_DEBUG` already counts the plan — publishing wrapped/eligible ratios plus size/time on the four real apps in the validation matrix would justify on-demand, calibrate whether the conservative closures over-wrap in practice, and double as the graduation gate for un-hiding the flag.
- **Combinational coverage:** iife/umd and preserveModules under strict are neither exercised nor rejected (silent unknowns — validate or fixture); sourcemaps, multi-output builds, and an executed HMR cycle under strict have zero coverage (dev+strict is snapshot-only via `issue_4636`). TLA, minify, CJS output, and externals are well covered.
- **Migration note for existing strict users:** wrap-all output shape changes (unconditional entry facades + runtime eviction; 1→3 and 3→5 chunk counts on existing fixtures, tracked in #10294). "Does not change default behavior" covers flag-off only — this deserves an explicit changelog entry.

### Design assessment

Asked "how would you solve this from scratch," I land on the same family: selective deferral is the only lever that preserves user chunk layout (membership changes break the manual-chunks contract, universal deferral pays for hazards that don't exist); the hazard only exists after provisional chunking, so the analysis belongs exactly where it is; and a side table with shared read views is the correct second-system response to the `override_wrap_kind` bridge's drift class (#10236, #10265). The conservative monotone closure is what makes the emergent-cycle fixpoint converge — a minimal wrap set would trade that away for an NP-hard objective and no soundness gain. The two audit decisions (double link-pass prediction, in-order interop representation) both read as scars from implemented-and-reverted alternatives, and the hoisted self-rebinding `function init_*` form is load-bearing rather than cosmetic. Every disagreement I have is at the margins listed above; none is architectural.

IWANABETHATGUY:
Took another pass over this after the two Fable reviews landed. The architecture holds up and I won't repeat what those already covered — everything below is stuff neither of them mentioned, from reading the diff at `220783a`. Three of these I'd really like to see addressed before merge; the rest can go to #10294 or follow-ups.

## The ones I'd fix before merge

**1. The facade-split gate only sees pre-lowering import edges** — `order_wrapping.rs:253`

`create_order_wrap_entry_facades` builds `imported_chunks` from `analysis.import_edges` (line 214), which is the pre-lowering baseline. But lowering itself can create the first cross-chunk edge into an interop entry's chunk — e.g. an `init_*` forward through an excluded re-export hop (the Hole-2 shape; `emergent_cycle_excluded_forwarder_import` shows the linker minting exactly these baseline-invisible edges). When that emergent edge is acyclic, the fixpoint adds no at-risk modules and converges without ever revisiting the facade decision, so the entry keeps its inline trigger. Evaluating the importing chunk then runs the whole entry program early — which is the exact bug class this feature exists to prevent.

The interesting part is that `analyze_execution_order` already computes post-lowering edges each round and throws them away after cycle detection (`order_analysis.rs:187-218`). Returning the final round's edges instead of the baseline at `order_analysis.rs:242` looks like it would close this. Worth noting the Tarjan assert proposed in the deep review wouldn't catch this one, since the missed edge is acyclic.

**2. Deconflict doesn't see force-included runtime statements** — `deconflict_chunk_symbols.rs:137`

The other two consumers of statement inclusion both OR in the overlay (`compute_cross_chunk_links.rs:390`, `module_finalizers/mod.rs:1622`), but deconflict filters on raw `stmt_info_included` only. So a runtime helper that tree-shaking didn't include but order-wrapping forces (say `__esmMin`) gets rendered without its symbol ever being registered with the renamer. If a user module in the same chunk has a top-level `__esmMin` — single-chunk builds put the runtime right next to user code via `order_wrapping.rs:451-456` — you get two same-named root declarations: a SyntaxError if the user binding is `let`/`const`, or the init wrappers calling the user's value if it's a `var`.

I couldn't find any other path that registers these symbols (synthetic statements only register the wrapper refs, and `order_live_symbols` only feeds the CJS captured-names set), so I think this one is real. It's also a nice illustration of why the overlay pattern makes me nervous — see the maintainability note below.

**3. Moving the runtime sweep after exec-order assignment changes flag-off chunk order** — `finalize_chunk_plan.rs:47`

On main the sweep ran before exec-order assignment — there was even a comment saying it had to. Now exec orders are assigned first, and the Common-chunk comparator keys on `modules[0]`'s exec order (`code_splitting.rs:187-216`). The runtime module is always `modules[0]` with exec order 0, so when the sweep later strips it out of a still-live *shared* chunk, that chunk keeps an exec_order derived from a module that's no longer there. `rebuild_sorted_chunk_idx_vec(true)` just re-sorts the stale values, and `renumber_live_chunks` is strict-only.

Concretely: strict off, sweep fires (the #9374 walk-back shape), runtime shares chunk C with some user module. Main keyed C on that module's exec order; this PR keys it on 0, so C can sort ahead of chunks it used to follow. That feeds `sorted_chunk_idx_vec` *and* the cross-chunk import sort (`compute_cross_chunk_links.rs:110/124`), so entry chunks can emit `import "./C.js"` before a chunk that should run first — side effects execute in a different order than main, on the default path. The deep review noticed the benign name-slot half of this move (which I agree is a wash — tombstones sorted last anyway); this half is the one that bites.

## Worth fixing, less urgent

**4. `dynamicImports` metadata points at the implementation chunk, not what the code actually imports** — `compute_cross_chunk_links.rs:358`

The set records `module_to_chunk[importee]`, but the finalizer rewrites the specifier via `entry_module_to_entry_chunk` — after a facade split those differ. Your own `manual_group_phantom_dynamic` snapshot shows it: `group-l.js` emits `import("./m6.js")`, but its metadata will list `group-l.js` and omit `m6.js`. Hashing is fine (placeholders come from rendered content), but modulepreload-style consumers and the devtools graph get the wrong direct edge. Recording `entry_module_to_entry_chunk` with a `module_to_chunk` fallback should do it.

**5. `restore_order_wrap_entry_facades` revives every matching tombstone but wires only the first** — `order_wrapping.rs:389`

Both earlier reviews suggested pinning this with `debug_assert!(len <= 1)`, but I don't think the invariant actually holds: emit the same module twice via `this.emitFile` with different names (`FileEmitter::emit_chunk` doesn't dedup) and you get two EmittedChunk facades, both of which the optimizer tombstones (same module means no export conflict, `chunk_optimizer.rs:945-959`). If that module lands in the plan, restore un-tombstones both and the second one renders as a stray near-empty entry asset. So rather than an assert, I'd only un-tombstone the facade we actually re-point to.

**6. Dead importers can force runtime helpers into the output** — `order_wrap_state.rs:33`

`populate_order_import_overlays` walks every module with no `is_included` check, and `required_runtime_helpers()` folds all overlays — while its sibling `live_symbols()` carefully gates on `overlay_is_live(key.importer)`. Net effect: a fully tree-shaken export-star barrel of a planned target mints a ReExport overlay, and `forces_runtime_stmt` ships a `__reExport` definition nobody calls. Bloat-only as far as I can tell, and gating the fold the same way `live_symbols` does looks safe — the overlays themselves need to stay, they feed the hop walk.

## Maintainability

**7. The liveness overlay pattern already claimed its first victim.** Order-lowering's demand is OR-ed into consumers at five call sites instead of being folded into `stmt_info_included`/`used_symbol_refs` — and item 2 above is exactly the site that forgot. Since lowering runs while the builder is still mutable (seal is at `mod.rs:150-152`, after `finalize_chunk_plan`, and `LinkStageOutput` is per-bundle), folding the demand into the canonical sets before seal — or at least routing every read through one accessor — would make the next consumer safe by construction.

**8. A few duplications that will drift:**
- The export-star `WrapKind::Esm` arm in `module_finalizers/mod.rs:1753` is now token-for-token `wrapped_esm_init_call_expr(importee.idx, SPAN, false, true)` — this PR removed the last difference and the comment that used to justify the fork. Might as well call the helper before the two grow apart again (the helper already gained the `esm_init_target` indirection this arm mirrors by hand).
- `ensure_runtime_module_for_order_wraps` mints the runtime chunk twice inline (`order_wrapping.rs:419-441`, `465-484`), so the `"rolldown-runtime"` literal now lives in three places — and the only copy with the "Vite/plugin-rsc match on this name" warning is the one in `code_splitting.rs:1071`. One shared constructor would keep the contract in one place.
- The 16-line `order_live_symbols` block is verbatim in `generate_stage/mod.rs:180-195` and `compute_cross_chunk_links.rs:261-276`. Extract a helper — but don't reuse the first *result*: `merge_cjs_namespace` links canonical refs between the two computations, so the second recompute is load-bearing.

## Perf quickies

- `add_wrapped_esm_init_targets_for_static_member_reads` (`esm_init_obligations.rs:359`) rescans every included statement's refs per not-directly-used binding. Under wrap-all that's per wrapped module per link pass; and since the Emit call site isn't strict-gated, `ns.x`-only bindings of cross-chunk importees pay it on flag-off builds too, where main did O(1) pushes. A per-module member-read index keyed by `object_ref` would kill both.
- The `TopLevelImportReadDetector` walk (`ast_scanner/impl_visit.rs:98`) is gated on strict only, but the bit it computes is read solely by the on-demand analysis (`order_analysis.rs:1005`) — wrap-all pays a per-module AST walk for nothing. Gate it on `onDemandWrapping` too.
- `create_order_wrap_entry_facades` builds `dynamic_target_modules_by_chunk` (a full modules × records sweep, `order_wrapping.rs:222-247`) unconditionally, but only reads it behind `!on_demand || ...` — wrap-all computes it and throws it away.

## Nits

- `detect_ineffective_dynamic_imports.rs:18` and `order_wrapping.rs:272` inline `matches!(..., Some(Removed))`, which is `!chunk_is_live()` — the helper this same PR adds (and `order_wrapping.rs` uses a few lines below).
- `ReexportRetentionEvidence` is only ever consumed via `.any()`; a bool argument would say the same thing.
- `OrderSyntheticStmt.chunk`/`.owner` are written but only read by debug_asserts — placement truth already lives in `synthetic_statements_by_chunk`.
- The `// See internal-docs/devtools/implementation.md ...` comment above `trace_action_chunks_infos` got dropped in the `generate_stage/mod.rs` reshuffle; the doc still pins that call's position in the action lifecycle, and AGENTS.md asks for the in-code reference.

---

One meta-note: I also chased a few things that turned out fine and might save the next reader time — the dead-specifier/ambiguous-star flag-off deviations the deep review asked fixtures for can't actually drop a side effect (side-effectful targets always force an earlier init through the barrel; the residual is byte-parity only, so a strict-gate on `binding_is_opaque` or one parity snapshot would settle it), and the recursive actual-order walk can't overflow on any input that `wrap_module_recursively` doesn't already blow up on first.

Nice work overall — the emit/register/project contract centralization in particular paid off; most of what I found lives in the seams around it, not in it.

hyfdev:
Verified all three pre-merge items against the code — all real. Fixed as three stacked draft PRs continuing the follow-up stack, each with a directed fixture:

- **Item 3 → #10334** (`re-derive chunk exec order after the runtime sweep`). Exactly the mechanism you described: the sweep's `chunk.modules.retain(!= runtime)` ran after the Common-chunk comparator had already keyed the shared chunk on the runtime's exec order 0. The exec-order assignment is now a reusable method and re-runs when the sweep removes the runtime from a live chunk. The fixture (`topics/runtime/sweep_shared_chunk_exec_order`) was snapshot-verified against main at the merge-base: the fixed branch reproduces main's chunk order.
- **Item 1 → #10335** (`gate entry facades on post-lowering import edges`). The gate now consumes the converged fixpoint round's edges instead of the pre-lowering baseline. One nuance from building the fixture (`facade_gate_emergent_edge`): the facade-decision gap is real and pinned (snapshot differentiator verified against the unfixed base), but an *executable* early-run does not appear constructible — a Hole-2 edge is projection-only (the real lowering never renders that forward; it exists for cycle detection), and a Hole-1 edge implies a baseline edge already exists, so the gate already split. Details and the attempt inventory are in the PR body; if a rendered baseline-invisible path is ever found, the fixture converts straight into an executed assertion.
- **Item 2 → #10336** (`deconflict force-included runtime helper statements`). Deconflict now ORs in `forces_runtime_stmt`, mirroring the other two consumers. The fixture reproduces the failure on the unfixed base — two root `var __esmMin`, with `init_late` throwing `TypeError: __esmMin is not a function` — and executes correctly on the fix. One precision on the frame: a strict-order build relocates a co-hosted runtime to a standalone chunk (`ensure_runtime_module_for_order_wraps`) unless code splitting is disabled, so the constructible collision is `codeSplitting: false`; the fix itself is mode-independent. Your item-7 point (route every read through one accessor) is the right structural cure — kept this diff minimal and filed the accessor refactor with the rest.

Items 4–8, the perf quickies, and the nits are being recorded on #10294 so the follow-up ledger stays in one place. Two notes on the meta-note: the byte-parity conclusion on the collector unification matches what #10324's executing fixtures found independently (both leaves' effects still run via the barrel's own init), and agreed on the recursion-depth argument — dropping that item.

Full follow-up stack on this PR, bottom-up: #10320 #10321 #10322 #10323 #10324 #10334 #10335 #10336.

IWANABETHATGUY (approved):

Issue #7416, opened 2025-12-10
Properly update internal states for incremental builds


sapphi-red:
We should refactor how things are handled.

https://github.com/rolldown/rolldown/pull/7348#issue-3698877793

- [ ] add tests that ensures that the states are updated properly for the following changes:
    - [ ] when a import is removed
    - [ ] when a file is deleted
- [ ] the state does not hold the removed modules (https://github.com/rolldown/rolldown/pull/5932#issuecomment-3632744609)

PR #7348, opened 2025-12-05
fix(dev): remove `imports` when an import is removed

sapphi-red:
I'm not sure if this is the supposed way to implement this fix. It feels we should do a refactor.
I also think should have a test that ensures the states are updated properly (a test that ensures the incremental build & full build generates the same states).

PR #10107, opened 2026-07-03
fix(dev): update importer relationships of cached modules in incremental build

shulaoda:
### Description

Part of #7416. This PR fixes the importer relationship item. The remaining items are listed at the end.

#### Problem

Importer relationships live in two layers:

1. The `importers` edge list on `ScanStageCache`. It is the authoritative reverse dependency graph, and #7348 already keeps it up to date during partial scans.
2. The materialized `importers`, `importers_idx` and `dynamic_importers` sets on each `NormalModule`, which are derived from that edge list.

A partial scan rebuilds the second layer only for the modules it rescans. When editing module A adds or removes an import of module B, or changes it between static and dynamic, the change lands on the sets of B. But B is usually served from cache and never rescanned, so its sets stay stale indefinitely.

The stale sets are consumed by:

- HMR boundary propagation (`importers_idx`). A patch can miss a newly added importer, whose module instance on the client then silently keeps the old dependency. A patch can also run modules that no longer import the edited file.
- `import.meta.hot.invalidate()`, which reads `caller.importers_idx` directly without any rescan.
- The link and generate stages on incremental rebuilds (`code_splitting`, `chunk_optimizer`, `on_demand_wrapping`, `create_exports_for_ecma_modules`).
- `this.getModuleInfo(id).importers` and `.dynamicImporters`.

#### Fix

The edge list is already correct, so this PR rebuilds the second layer from it for exactly the affected modules:

- `ScanStageCache` gains `modules_with_changed_importers`. `ModuleLoader::mark_module_importers_changed` records the target module next to every mutation of the `importers` edge list during a partial scan. The marks are written straight to the cache so they survive a failed scan.
- `ScanStageCache::merge` drains the set and rebuilds the sets of each listed module from the edge list through the new `EcmaView::rebuild_importer_sets`, the same helper the loader uses for rescanned modules.
- `merge` also refreshes the `ModuleInfo` exposed to plugins for those modules, because the scan refreshes it only for modules it produced. This keeps `ctx.getModuleInfo(id)` in `advancedChunks` name callbacks and plugin queries consistent with the merged graph. `merge` now takes `&PluginDriver` for this.
- The `user_defined_entry_modules` recompute in `merge` now also keeps the modules of `EmittedUserDefined` entry rows. Partial scans do not re-run `buildStart`, so entries emitted via `emitFile(type: 'chunk')` are absent from the scan output's set; without this, the refreshed `ModuleInfo` of such a module would report `isEntry: false` after any partial merge that touches its importers.

#### Behavior change in an existing snapshot

`delete_file_not_used_anymore` step 2 changes from a patch to a full reload. That step edits an orphaned module whose only importer dropped the import in step 0. The old patch came from walking the stale phantom edge. It ran a module that no longer imports the edited file and shipped dead code to the client. A full reload is the honest answer while orphaned modules remain in the graph state. The ideal result is a noop, because a fresh build would not contain the module at all, but that requires the orphan cleanup tracked in #7416. The test harness forbids executing output in fixtures that hit a full reload, which is why the fixture now sets `"expectExecuted": false`.

#### Tests

- `hmr/import_added_to_cached_module`: a module adds an import of a cached module. A later edit of the cached module must reach the new importer. Without the fix the patch misses that boundary.
- `hmr/import_removed_from_cached_module`: the mirror case. A later edit of the cached module must not run the old importer again. Without the fix the patch still contains it.

#### Docs

`internal-docs/cache/implementation.md` and `internal-docs/bundler-data-lifecycle/implementation.md` are updated for the new field and the new merge steps.

#### Remaining items of #7416

- The state still holds removed modules. Editing an orphaned module therefore triggers a full reload instead of a noop, and its module and AST stay in memory.
- There is no general check yet that an incremental build and a full build produce the same states. The new fixtures cover the importer sets only.
- There is no recovery story for a failed partial scan. A failed scan leaves `module_id_to_idx` marked as seen for modules whose results never reached the snapshot, so the edge list and the snapshot can drift apart until the affected files change again.
- Dynamic import entry points are never removed from the cached entry list when the dynamic import goes away.

### DIFF
--- a/crates/rolldown/src/bundle/bundle.rs
+++ b/crates/rolldown/src/bundle/bundle.rs
@@ -253,7 +253,7 @@ impl<Fs: FileSystem + Clone + 'static> Bundle<Fs> {
       return Ok(output);
     }
 
-    self.cache.merge(output)?;
+    self.cache.merge(output, &self.plugin_driver)?;
     self.cache.update_defer_sync_data(&self.options).await?;
     Ok(self.cache.create_output())
   }
--- a/crates/rolldown/src/hmr/hmr_stage.rs
+++ b/crates/rolldown/src/hmr/hmr_stage.rs
@@ -281,7 +281,8 @@ impl<'a, Fs: FileSystem + Clone + 'static> HmrStage<'a, Fs> {
           .collect::<Vec<_>>(),
       );
 
-      self.cache.merge(module_loader_output.into())?;
+      let plugin_driver = Arc::clone(&self.plugin_driver);
+      self.cache.merge(module_loader_output.into(), &plugin_driver)?;
 
       let options = Arc::clone(&self.options);
       self.cache.update_defer_sync_data(&options).await?;
@@ -374,7 +375,11 @@ impl<'a, Fs: FileSystem + Clone + 'static> HmrStage<'a, Fs> {
     let module_loader_output = module_loader.fetch_modules(fetch_mode).await?;
     drop(module_loader);
 
-    self.cache.merge(module_loader_output.into()).map_err(|e| vec![anyhow::anyhow!(e).into()])?;
+    let plugin_driver = Arc::clone(&self.plugin_driver);
+    self
+      .cache
+      .merge(module_loader_output.into(), &plugin_driver)
+      .map_err(|e| vec![anyhow::anyhow!(e).into()])?;
 
     let options = Arc::clone(&self.options);
     self.cache.update_defer_sync_data(&options).await?;
--- a/crates/rolldown/src/module_loader/module_loader.rs
+++ b/crates/rolldown/src/module_loader/module_loader.rs
@@ -217,6 +217,16 @@ impl<'a, Fs: FileSystem + Clone + 'static> ModuleLoader<'a, Fs> {
     })
   }
 
+  /// Marks `idx` for importer-set re-derivation in `ScanStageCache::merge`;
+  /// call it next to every mutation of `intermediate_normal_modules.importers`.
+  /// Writes straight to the cache so marks survive a failed scan. No-op in
+  /// full scan mode, which rebuilds every module's sets anyway.
+  fn mark_module_importers_changed(&mut self, idx: ModuleIdx) {
+    if !self.is_full_scan {
+      self.cache.modules_with_changed_importers.insert(idx);
+    }
+  }
+
   /// Lazy barrel must not skip any of an entry's re-exports — an entry has to
   /// preserve all of its exports. Request `All` from it: if it has already been
   /// loaded as a barrel, load its remaining (deferred) records now; otherwise
@@ -425,11 +435,15 @@ impl<'a, Fs: FileSystem + Clone + 'static> ModuleLoader<'a, Fs> {
             && let Some(previous_module) =
               self.cache.get_snapshot().module_table.modules.get(module_idx)
           {
-            let resolved_deps =
-              previous_module.import_records().into_iter().filter_map(|r| r.resolved_module);
+            let resolved_deps = previous_module
+              .import_records()
+              .into_iter()
+              .filter_map(|r| r.resolved_module)
+              .collect::<Vec<_>>();
             for dep_idx in resolved_deps {
               self.intermediate_normal_modules.importers[dep_idx]
                 .retain(|v| v.importer_idx != module_idx);
+              self.mark_module_importers_changed(dep_idx);
             }
           }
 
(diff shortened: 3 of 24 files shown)


shulaoda:
* **#10117** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10117?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10110** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10110?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10107** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10107?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10107?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10107).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

h-a-n-a (commented):
LGTM, with some reviews:

**1. `internal-docs/cache/implementation.md` — stale caller references left behind**
The writers table (line 371) was updated to the two post-PR `merge` call sites, but the Callers section at line 304 still lists `hmr_stage.rs:286, :379, :621`, and the `update_defer_sync_data` row (line 373) still cites `:289/382/623`. There are only two HMR call paths now (`merge` at `hmr_stage.rs:285/381`, `update_defer_sync_data` at `:288/385`) — the third site near `:621` doesn't exist anymore. Worth syncing all three spots while this PR touches the table.

**2. `delete_file_not_used_anymore` — `"expectExecuted": false` disables execution for the whole fixture**
Only step 2 becomes a full reload, but the flag also turns off the runtime assertions for the initial build and steps 0/1/3, making the fixture snapshot-only. Could the orphaned-module edit be split into its own fixture so the original steps keep executing?

**3. (non-blocking) failed-scan marks can materialize edge-list/snapshot drift into the HMR sets**
Marks surviving a failed scan interacts with the known "no recovery story for a failed partial scan" item: if a scan removes edge A→B, marks B, then fails before merge, the next successful merge (from any unrelated edit) rebuilds B's `importers_idx` without A — while the snapshot and the client still run an old A that imports B. Editing B then skips boundary A and the client silently keeps the stale B. Pre-PR, that drift stayed confined to the edge list. Might be worth folding "reconcile or drop marks from a failed scan" into the #7416 recovery item.

**4. (non-blocking, perf) rescans mark all old and new deps unconditionally**
An edit whose import set didn't change still marks every dep at the retain site and again at the re-add sites, so every keystroke rebuild re-derives N importer sets and re-allocates N `ModuleInfo`s (clone + sort of each dep's full importer list). Noticeable when the edited file imports a hub/barrel module with many importers; modules already in the scan output also get rebuilt twice (loader + merge). A net-change guard, or subtracting the scan output's idxs from the drained set, would trim the common case — fine as a follow-up.

h-a-n-a (approved):

shulaoda:
### Merge activity

* **Jul 6, 3:47 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 6, 4:10 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 6, 4:11 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 6, 4:14 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 6, 4:31 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 6, 7:09 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 6, 7:14 AM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 6, 7:25 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10109, opened 2026-07-03
test: add regression test for #10099 (lazyBarrel drops default-import binding but keeps its property reads)

### LINKED ISSUE #10099: Tree-shaking removes default import binding but keeps property accesses, causing ReferenceError
### Describe the bug

When building a project that imports `element-plus`, Rolldown's tree-shaking incorrectly removes the `import defaults_default from "./defaults.mjs"` statement but **keeps** the code that references `defaults_default`, causing a `ReferenceError: defaults_default is not defined` at runtime.

### Reproduction

**Minimal source pattern that triggers the bug:**

Module `defaults.mjs`:
```js
var defaults_default = makeInstaller([...components]);
export { defaults_default as default };
```

Module `index.mjs` (element-plus/es/index.mjs):
```js
import defaults_default from "./defaults.mjs";
const install = defaults_default.install;
const version = defaults_default.version;
var element_plus_default = defaults_default;
export { /* ... many named exports ... */ install, version, element_plus_default as default };
```

When a consumer imports only a named export (e.g. `import { ElNotification } from 'element-plus'`), Rolldown tree-shakes the module and produces output like:

```js
// defaults_default is NOT defined/imported anywhere in the output
defaults_default.install, defaults_default.version;  // ← ReferenceError!
```

The `import` binding and the variable assignment (`var element_plus_default = defaults_default`) are removed, but the property access expressions (`defaults_default.install`, `defaults_default.version`) are retained as dead side-effects.

### System Info

```
Rolldown: 1.1.3 (via Vite 8.0.16)
Node: v22.21.0
OS: Windows 11
```

### Used Package Manager

pnpm

IWANABETHATGUY:
## What

Adds a Rust integration-test fixture at `crates/rolldown/tests/rolldown/issues/10099/` closes #10099.

## Why

#10099: importing only `dayjs` from `element-plus` emitted a bundle that threw `ReferenceError: defaults_default is not defined`. The `sideEffects:false` barrel (`index.mjs`) re-exports an **external** module's default (`dayjs`) as a passthrough while reading `defaults_default.install`/`.version` at the top level. Importing only that passthrough force-includes the barrel; under `experimental.lazyBarrel` the tree-shaker dropped the `defaults_default` import **and** its definition but kept the property reads → dangling reference.

Fixed by #10080 (gate `sideEffects:false` modules' side effects on body demand). This PR adds the regression test for the specific element-plus shape, which none of #10080's fixtures cover: here the inclusion driver is an **external-default passthrough** (the sibling fixtures use local passthroughs or dynamic-import wrapping).

## The fixture

- `_config.json`: `external:["extdep"]`, `experimental.lazyBarrel:true`, `treeshake.moduleSideEffects:false`
- barrel `index.js` default-imports `defaults.js`, reads `.install`/`.version`, and re-exports the external default as `extdep`
- entry `main.js` imports only `{ extdep }` and asserts it resolves
- `node_modules/extdep/` stub so the external passthrough resolves at runtime

`experimental.lazyBarrel: true` is load-bearing: the bug is lazyBarrel-gated and lazyBarrel is default-off since #10071, so without it the test would pass on both the base and the fix.

## Verification

- **Green** on current `main` (fix present): the harness executes the compiled entry (`expectExecuted` default) and the snapshot is clean — no `defaults_default`.
- **Red** on the pre-fix base: the exact fixture source + config through published `rolldown@1.1.3` throws `ReferenceError: defaults_default is not defined`.

Test-only; no source changes.

### DIFF
--- a/crates/rolldown/tests/rolldown/issues/10099/_config.json
+++ b/crates/rolldown/tests/rolldown/issues/10099/_config.json
@@ -1,0 +1,12 @@
+{
+  "_comment": "Regression test for https://github.com/rolldown/rolldown/issues/10099 - the barrel `index.js` default-imports `defaults.js` and reads `defaults_default.install`/`.version` at the top level, while re-exporting an EXTERNAL module's default (`extdep`, real-world `dayjs` in element-plus). The entry imports ONLY that external passthrough, which force-includes the barrel (an external default re-export is not resolvable from sub-modules). Under `experimental.lazyBarrel` + `moduleSideEffects: false`, tree-shaking dropped the `defaults_default` import AND its definition in `defaults.js`, yet kept the `defaults_default.install/.version` property-read statements -> `ReferenceError: defaults_default is not defined` when the compiled entry runs. Fixed by #10080 (gate a declared-pure module's side-effect statements on body demand); executing the entry must not throw.",
+  "config": {
+    "external": ["extdep"],
+    "experimental": {
+      "lazyBarrel": true
+    },
+    "treeshake": {
+      "moduleSideEffects": false
+    }
+  }
+}
--- a/crates/rolldown/tests/rolldown/issues/10099/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/issues/10099/artifacts.snap
@@ -1,0 +1,15 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+---
+# Assets
+
+## main.js
+
+```js
+import external_default from "extdep";
+//#region main.js
+if (!external_default || external_default.name !== "extdep-default") throw new Error("external passthrough did not resolve");
+console.log("ok:", external_default.name);
+//#endregion
+
+```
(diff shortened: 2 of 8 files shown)


hyfdev (approved):

shulaoda (approved):

IWANABETHATGUY:
### Merge activity

* **Jul 3, 8:06 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 3, 8:06 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 3, 8:11 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10110, opened 2026-07-03
fix(dev): revert cache mutations when a partial scan fails

shulaoda:
### Description

Part of #7416. Stacked on #10107. This PR makes a failed scan harmless to the incremental cache. The remaining items of #7416 are listed at the end.

#### Problem

A partial scan applies task results as they arrive. When the scan aborts (the typical case is a syntax error in an edited file), the cache is left half updated while the snapshot receives nothing, because `merge` never runs on the error path:

1. Re-scanned modules were flipped to `Seen` in `module_id_to_idx`, so they are treated as fresh even though their new content never landed. The next build silently serves their old code and the error disappears.
2. Completed tasks already rewrote parts of the `importers` edge list, which then disagrees with the snapshot.
3. Newly discovered modules got an index and a `module_id_to_idx` entry, but no slot in the snapshot table. Since new indexes are allocated from the map length and `merge` appends at the table length, a leftover entry shifts every later module by one slot and corrupts all index based lookups.

Concrete reproduction (the new fixture): break `b.js`, then edit `a.js`. The rebuild reports success, ships the pre-break `b.js`, and the error overlay is gone.

#### Fix

A failed partial scan now reverts its cache mutations, so the cache holds a single invariant: between builds it is either empty (no snapshot, only a full scan is possible) or a valid graph any partial scan can build on. There is no third "poisoned" state and no consumer has to reason about failed builds.

`ModuleLoader::revert_partial_scan` runs at the scan error exit. The snapshot is never touched by a scan, so it acts as the clean master copy everything else is recovered from:

- `importers` is re-derived from the snapshot (`ScanStageCache::derive_importers_from_snapshot`), since every edge is a resolved import record of some module. Slot order is not preserved, and consumers treat slots as sets. The rebuild cost is paid only on failure, the happy path pays nothing.
- Existing `module_id_to_idx` entries end the scan at their pre-scan values (`Seen` to `Invalidate` back to `Seen` with the same index), so only the keys inserted for newly discovered modules need removal. The loader records them in `new_module_ids`.
- `modules_with_changed_importers` is cleared, matching the reverted edge list.
- `barrel_state` is restored from an upfront clone, taken only when lazy barrel is enabled. When disabled the state is empty and never mutated.
- The scanned files go into `ScanStageCache::pending_rescans`, a work queue the next partial scan drains. Reverting alone would lose the knowledge that these files changed, and the next build would silently serve their old content. The queue keeps their errors surfacing until the files are fixed. Only files the graph still needs are queued: configured or emitted entries, and files something still imports on the freshest edge state. A broken file whose last import was removed by the very scan that failed is dropped; otherwise its retry would fail every later partial build that a fresh build of the same tree would pass, wedging the session until a restart (new fixture `hmr/error_recovery/remove_import_of_broken_file`).
- `transform_dependencies` entries of the newly allocated (and now freed) module indices are dropped, so a later module reusing an index does not inherit them.

`Bundler::incremental_bundle` now picks the scan mode from `has_snapshot()` alone. A failed full scan needs no revert, its cache reset already leaves the empty state that forces a full retry.

One adjacent panic became reachable and is now handled: a client calling `import.meta.hot.invalidate()` for a module the current graph does not know (for example after a failed full build reset the cache) gets a full reload instead of a server panic. The same applies to `compile_lazy_entry`: a lazy chunk request arriving after a failed full build now returns an error instead of panicking on the missing snapshot.

Two consumers of the retry queue needed matching fixes:

- `HmrStage::compute_hmr_update` folds the queued files into its stale and changed sets before computing client boundaries. Without this, an edit rolled back by a failed scan reached the server graph on the successful retry but never a client patch, leaving connected clients silently running the old module. The new fixture `hmr/error_recovery/fix_broken_file_delivers_other_edits` asserts the recovery patch contains both the fixed file and the edit made while broken.
- The import chain enrichment of unresolved import errors now runs before the revert, while the failed scan's modules still exist in the lookup structures the trace walks. Running it after erased the chain from dev mode diagnostics.

Plugin side state (`module_infos`, JS side caches) is not reverted. Hooks that ran during the aborted scan already updated them. This residue is informational, keyed by module id, and overwritten by the next successful scan. The exception is `transform_dependencies`, which is keyed by module index: the entries of indices freed by the revert are dropped, as described above.

#### Behavior

- The existing error recovery fixtures are unchanged. Breaking a file and fixing the same file still recovers with a plain HMR patch.
- New fixture `hmr/error_recovery/remove_import_of_broken_file`: break `b.js`, then remove the `import './b.js'` from `a.js`. That build still surfaces the queued error once, then drops `b.js` from the queue since nothing imports it anymore, and the following edit builds normally again.
- New fixture `hmr/error_recovery/edit_other_file_while_broken`: break `b.js`, edit `a.js` (the error surfaces again instead of a fake success with stale code), fix `b.js` (builds succeed again, the queued edit of `a.js` lands in the same scan), edit `a.js` once more (a normal patch, proving the cache stayed incremental through the whole episode).

#### Docs

The "Cache integrity on a failed build" section of `internal-docs/bundler-data-lifecycle/implementation.md` now documents the revert and the two state invariant. `internal-docs/cache/implementation.md` is updated for the new field and methods.

#### Remaining items of #7416

- The state still holds removed modules. Editing an orphaned module triggers a full reload instead of a noop, and its module and AST stay in memory.
- There is no general check yet that an incremental build and a full build produce the same states. A dedicated comparison in the test harness would also guard the revert introduced here.
- Dynamic import entry points are never removed from the cached entry list when the dynamic import goes away.

### DIFF
--- a/crates/rolldown/src/bundler/impl_bundler_incremental_build.rs
+++ b/crates/rolldown/src/bundler/impl_bundler_incremental_build.rs
@@ -57,6 +57,10 @@ impl Bundler {
     is_write: bool,
     scan_mode: ScanMode<ArcStr>,
   ) -> BuildResult<BundleOutput> {
+    // No snapshot, no graph to build on: scan fully. Deciding before the
+    // `BundleMode` choice applies the `IncrementalFullBuild` reset rules to
+    // the plugin driver state.
+    let scan_mode = if self.cache.has_snapshot() { scan_mode } else { ScanMode::Full };
     let bundle_mode = match scan_mode {
       ScanMode::Full => BundleMode::IncrementalFullBuild,
       ScanMode::Partial(_) => BundleMode::IncrementalBuild,
--- a/crates/rolldown/src/hmr/hmr_stage.rs
+++ b/crates/rolldown/src/hmr/hmr_stage.rs
@@ -92,12 +92,13 @@ impl<'a, Fs: FileSystem + Clone + 'static> HmrStage<'a, Fs> {
       first_invalidated_by,
     );
     // Look up by stable_id (matches what client sends from createModuleHotContext)
-    let module_idx = self
-      .cache
-      .module_idx_by_stable_id
-      .get(invalidate_caller.as_str())
-      .copied()
-      .unwrap_or_else(|| panic!("Not found modules for file: {invalidate_caller}"));
+    let Some(module_idx) =
+      self.cache.module_idx_by_stable_id.get(invalidate_caller.as_str()).copied()
+    else {
+      // The client references a module the current graph does not know,
+      // e.g. its bundle predates a failed full build that reset the cache.
+      return Ok(HmrUpdate::FullReload { reason: format!("unknown module `{invalidate_caller}`") });
+    };
 
     let caller = self.module_table().modules[module_idx].as_normal().unwrap();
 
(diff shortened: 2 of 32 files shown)


shulaoda:
* **#10117** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10117?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10110** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10110?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10110?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10107** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10107?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10110).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

shulaoda:
@codex[agent] review

h-a-n-a (approved):

h-a-n-a:
### Merge activity

* **Jul 6, 8:10 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 6, 8:11 AM UTC**: `h-a-n-a` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 6, 8:16 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9155, opened 2026-04-19
feat: skip barrel files when all imports in the barrel files are marked as side effect free

IWANABETHATGUY:
Closed #8920

PR #10111, opened 2026-07-03
feat(linking): skip side-effect-free modules in per-entry reachability

### LINKED ISSUE #8920: [Feature Request]: skip barrel files when all imports in the barrel files are marked as side effect free

```mermaid
graph LR                                                                                                                                              
    entry-light["entry-light.js"] -->|"import { light }"| barrel["barrel.js"]
    entry-heavy["entry-heavy.js"] -->|"import { createService }"| barrel                                                                                
                                                                                                                                                            barrel -->|"import { HeavyService }"| heavy["heavy.js"]                                                                                             
    barrel -->|"export { light }"| light["light.js"]                                                                                                    

    light:::leaf
    heavy:::leaf

    classDef leaf fill:#e8f5e9,stroke:#388e3c
```
https://github.com/sapphi-red-repros/barrel-split-export-repro

For example, if `heavy.js` is marked as sideeffect free, we can ignore that import when bundling `light.js`. So we can have with 2 chunks like Rollup does. But Rolldown now generates 3 chunks: one for entry-light, one for entry-heavy, one for the shared barrel file.


Not an API change

Although I'm not sure how the other bundlers define "sideEffect-free" for modules. We need to change this definition to do this optimization.

> While you can mark individual expressions or functions, you can also mark entire modules as side-effect-free. If you mark a module as side-effect-free, Rolldown will treat every statement in that module as side-effect-free _when none of its exports are used_.
> https://rolldown.rs/in-depth/dead-code-elimination#marking-entire-modules-as-side-effect-free:~:text=While%20you%20can%20mark%20individual%20expressions%20or%20functions%2C%20you%20can%20also%20mark%20entire%20modules%20as%20side%2Deffect%2Dfree.%20If%20you%20mark%20a%20module%20as%20side%2Deffect%2Dfree%2C%20Rolldown%20will%20treat%20every%20statement%20in%20that%20module%20as%20side%2Deffect%2Dfree%20when%20none%20of%20its%20exports%20are%20used.

---

Originally reported at https://github.com/vitejs/vite/issues/21966

IWANABETHATGUY:
### Description

Closes #8920. Minimal alternative to #9155 — no per-entry inclusion tracking; only the code-splitting reachability edges change.

#### Root cause

Rolldown already agrees in two places on what forces a module to be **loaded**:

- tree-shaking (`include_side_effectful_dependencies`) follows a static import-record edge only when the importee `side_effects().has_side_effects()`; used bindings pull in the module that canonically owns them (re-exports resolve *through* a barrel to the origin module), and
- `compute_cross_chunk_links` emits a bare `import "chunk.js"` for a record edge only when the importee has side effects; all other cross-chunk imports are symbol-driven.

The code-splitting BFS (`determine_reachable_modules_for_entry`) was the one place without that filter: it walks `meta.dependencies`, the union of **all** static record targets plus the symbol-owner dependencies merged in by `patch_module_dependencies`. So `entry-light → barrel` (a pure record edge whose imported binding canonically lives in `light.js`) still marked the barrel — and transitively `heavy.js` — as reachable from the light entry, producing a shared chunk that neither the inclusion semantics nor the emitted imports actually load:

```
# before: 3 chunks, entry-light loads HeavyService through the shared chunk
barrel.js   (heavy.js + light.js + barrel.js, bits {light,heavy})
entry-heavy.js
entry-light.js

# after: 2 chunks, same as Rollup
entry-heavy.js (heavy.js + barrel.js + entry-heavy.js)
entry-light.js (light.js + entry-light.js)
```

#### Fix

`patch_module_dependencies` now additionally records `meta.load_dependencies` — the edges that actually force a module to be loaded at runtime:

- modules owning the canonical symbols referenced by this module's included statements (incl. CJS namespace-alias owners, runtime-when-helpers, entry-chunk export owners), plus
- import-record targets whose evaluation has side effects (the exact `include_side_effectful_dependencies` rule; with tree-shaking disabled the set equals `dependencies`, so nothing changes there).

`determine_reachable_modules_for_entry` walks `load_dependencies`. Every other `meta.dependencies` consumer (chunk optimizer, manual code splitting, on-demand wrapping, dynamic-already-loaded) keeps the broad set, so their behavior is untouched.

One deliberate exception: record edges **into entry modules** are never pruned. An entry's chunk exists regardless, and letting static importers participate in its bit pattern keeps shared code co-located with the entry chunk — Rollup reaches the same topology by collapsing code-less entry facades onto the chunk holding their exports. Without this, a statically imported re-export barrel that is also a dynamic entry pushes its re-export targets into a separate shared chunk and leaves the dynamic chunk an empty facade (caught by rollup's `chunking-form/entry-without-code-dynamic` in CI: 4 chunks instead of 3; mirrored here as the `code_splitting/static_import_of_code_less_dynamic_entry` fixture).

Every included module stays reachable: a module only becomes included via a side-effect-full record edge or via a used symbol, and both edge kinds are present in `load_dependencies` (the existing `debug_assert` at the chunk-assignment site checks exactly this invariant across the whole suite). The suite surfaced one previously masked asymmetry — an entry export resolving to a facade binding with a CJS `namespace_alias` (`esbuild/importstar/export_other_nested_common_js`) contributed no dependency edge from the entry-chunk loop, unlike the statement walk; the entry-chunk loop now mirrors the same namespace-alias handling.

#### Snapshot changes (all audited)

Beyond the new `issues/8920` and `code_splitting/static_import_of_code_less_dynamic_entry` fixtures, 5 existing snapshots change, all in the same direction — a module that an entry never actually loads is no longer assigned to that entry's chunk group:

- `issues/rolldown_vite_593`, `chunk_merging/transitive_dynamic_dep_should_not_merge_into_side_effectful_entry`, `tree_shaking/no_side_effects_namespace_cross_chunk_cases`: single-consumer modules move out of shared chunks into their only consumer; shared chunks keep only genuinely shared code.
- `strict_execution_order/strip_plain_chunk_imports`: the `common.js` chunk (reachable only through plain record edges — what this fixture is about) colocates with its one real consumer; the executed `node:assert` checks still pass.
- `esbuild/splitting/splitting_cross_chunk_assignment_dependencies_recursive`: the pure `set*` helper chain colocates into the `c.js` entry, letting the default `dce-only` pass eliminate the provably dead chain entirely (its `a.js`/`b.js` outputs were already empty on main).

Fixtures involving record edges into dynamic entry modules (`issue_8809`, `dynamic_entry_shared_module_not_merged`, `9320_star`) keep today's output byte-for-byte thanks to the entry-edge exception above; slimming those dynamic-entry namespaces/facades is the existing proxy-chunk follow-up topic, not this PR.

Fixtures that execute their output all still pass; full integration suite is green (1756 passed, hmr/watcher skipped locally on Windows).

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/code_splitting.rs
@@ -1091,7 +1091,12 @@ impl GenerateStage<'_> {
       if is_user_defined_entry {
         index_splitting_info[module_idx].tags_bit_set.set_bit(ModuleTag::INITIAL_BIT);
       }
-      meta.dependencies.iter().copied().for_each(|dep_idx| {
+      // Walk only the dependencies this module actually loads at runtime (referenced-symbol
+      // owners and side-effect-full import targets). A side-effect-free module whose bindings
+      // resolve elsewhere — e.g. a pure barrel that only re-exports — must not be treated as
+      // reachable, otherwise it (and its subtree) is shared into this entry's chunk group even
+      // though the emitted code never imports it (#8920).
+      meta.load_dependencies.iter().copied().for_each(|dep_idx| {
         q.push_back(dep_idx);
       });
     }
--- a/crates/rolldown/src/stages/link_stage/patch_module_dependencies.rs
+++ b/crates/rolldown/src/stages/link_stage/patch_module_dependencies.rs
@@ -1,5 +1,5 @@
 use rayon::iter::ParallelIterator;
-use rolldown_common::{Module, RuntimeHelper};
+use rolldown_common::{Module, ModuleIdx, RuntimeHelper};
 use rolldown_utils::{index_vec_ext::IndexVecRefExt, indexmap::FxIndexSet};
 
 use super::LinkStage;
@@ -20,6 +20,15 @@ impl LinkStage<'_> {
           |(symbol_ref, _came_from_cjs)| {
             let canonical_ref = self.symbols.canonical_ref_for(*symbol_ref);
             extended_dependencies.insert(canonical_ref.owner);
+            // An entry export may resolve to a facade binding whose value lives on a CJS
+            // module's namespace (e.g. re-exporting a name that a CJS module only provides
+            // dynamically). Inclusion follows that alias (`follow_cjs_namespace_alias`), so the
+            // aliased module is a real dependency of the entry, same as in the statement walk
+            // below.
+            let symbol = self.symbols.get(canonical_ref);
+            if let Some(ns) = &symbol.namespace_alias {
+              extended_dependencies.insert(ns.namespace_ref.owner);
+            }
           },
         );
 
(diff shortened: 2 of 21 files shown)


hyfdev (approved):

IWANABETHATGUY:
### Merge activity

* **Jul 7, 3:41 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 7, 3:41 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 7, 3:46 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10115, opened 2026-07-03
test(dev): assert incremental scan state matches a fresh full build after each HMR step

shulaoda:
### Description

Part of #7416. Stacked on #10112. This adds the test #7416 asks for: one that ensures an incremental build and a full build produce the same states.

#### What it does

After each HMR step of a dev fixture, the test harness now builds a fresh bundler with the same options and plugins against the current file state and runs a full build. The fresh bundler has no history, so its scan state is by definition the correct answer for the files on disk. The harness then asserts the incremental state matches it:

- `importers` and `dynamic_importers` of every module, compared as sets of module ids. Importer entries pointing at modules the fresh build does not know are ignored: those are the outgoing records of orphaned modules, which stay in the state by design (issue #7416). A stale importer entry from a module the fresh build *does* have is still reported.
- `importers_idx`, mapped back to module ids since the two sides use different index spaces, with the same orphan filter.
- Every import record's kind and resolved target.
- Entry points, where the fresh set must be a subset of the incremental set. Extras on the incremental side are allowed for now, because removed modules and stale dynamic import entries are separate open items of #7416.

Any difference fails the test with a per module, per field report.

#### Why per step instead of once at the end

An end of run comparison has a blind spot: a divergence on a cached module is silently repaired as soon as a later step happens to rescan that module, so only the intermediate state is wrong, which is exactly when HMR reads it. This was verified by temporarily reintroducing the bug fixed in #10107 (a cached module keeping a stale importer set). The end of run check passed, the per step check caught it immediately and reported the exact missing importer.

#### When it is skipped

- Steps whose file state does not build, e.g. the error recovery fixtures while their syntax error is in place. A state that cannot be built has no full build to compare against.
- Steps whose incremental build failed. A failed scan is reverted, so the state intentionally stays at the last good build and cannot mirror the current files until a later scan retries them.
- Fixtures with lazy barrel enabled. It loads import records on demand, so the loaded subset depends on request history and legitimately differs between a session and a fresh build.

#### Opting out

`dev.checkStateParity` defaults to `true`, so every existing and future dev fixture gets the check for free (all 38 current HMR fixtures pass with it enabled). A fixture that intentionally documents a known divergence can set it to `false`, and the flag in its config then serves as a visible marker of the known gap until the underlying fix lands.

#### Implementation notes

- The comparison lives inside the `rolldown` crate behind the existing `testing` feature (`Bundler::assert_scan_state_parity_with`), so it can read the private cache directly and no new public API surface is added.
[...]

PR #10112, opened 2026-07-03
test(dev): restore runtime assertions of delete_file_not_used_anymore

shulaoda:
### Description

Addresses the review feedback on #10107. Stacked on #10110. Two of the four comments need changes and they live here. The third comment (a failed scan leaves marks that later materialize edge list drift into the HMR importer sets) is fixed by #10110, which reverts all cache mutations of an aborted partial scan. The fourth comment (rescans mark every dependency even when the import set did not change) stays a follow-up, with one note for whoever picks it up: skipping scan output indexes in `merge` is not a pure win, because the loader refreshes `ModuleInfo` only for modules with non empty importers and the merge loop currently covers the empty case.

#### Stale doc references

`internal-docs/cache/implementation.md` still pointed at three `merge` call sites in `hmr_stage.rs`. The site near `:621` no longer exists. The `Callers` section and the writers table rows for `merge` and `update_defer_sync_data` now cite the two real HMR call sites (the update path and the lazy compile path).

#### Restore runtime assertions of `delete_file_not_used_anymore`

`"expectExecuted": false` is a fixture wide switch. Only the orphan recreation step needs it, but it also turned off the runtime assertions of the initial build and of every other step, making the whole fixture snapshot only.

The fixture is now split:

- `delete_file_not_used_anymore` keeps the executing steps (drop the import, then delete the file) and runs its runtime assertions again.
- The new `hmr/recreate_orphan_module` fixture carries the full four step sequence (drop the import, delete the file, recreate it, re-import it) as snapshot only. It documents the current limitation that editing or recreating an orphaned module triggers a full reload until the orphan cleanup tracked in #7416 lands.

The recreation step cannot stay in the executing fixture, because the re-import step needs the recreated file on disk, so the sequence only makes sense as a whole.

PR #10117, opened 2026-07-03
fix(dev): prune dynamic import entry points when their imports are removed

shulaoda:
### Description

Part of #7416. Stacked on #10115. This makes `ScanStageCache::merge` clean up dynamic import entry points when the dynamic imports behind them go away. Until now the entry list only ever grew.

#### Problem

Two related gaps in the entry point handling of `merge`:

1. Stale `related_stmt_infos` rows. The old code removed a re-scanned importer's rows only when the scan output contained a matching entry point, keyed by the incoming rows. When a re-scanned module removed its dynamic import entirely, the scan output had no entry for the target at all, so the importer's old rows stayed behind and kept pointing at statement and node ids of a module version that no longer exists.
2. Zombie entries. When the last dynamic importer of a module disappeared, the `DynamicImport` entry point stayed in the cache forever. Every incremental rebuild kept emitting a chunk for a module nothing imports dynamically anymore, which a fresh full build of the same files would not produce.

#### Fix

- Before merging the incoming entry points, drop the rows of every re-scanned module from the cached dynamic import entries. The incoming entries then re-add exactly the rows that still exist. This subsumes the old incoming-keyed `extract_if`, which is removed.
- After merging, prune every `DynamicImport` entry whose module no longer has any dynamic edge in the `importers` edge list. The edge list is reliable at merge time since #10107 and #10110. Other entry kinds are untouched.
- The prune is intentionally not transitive on the cache: edges of orphaned importers survive in the edge list, so an entry only they point at stays cached, and re-adding an import of the orphan revives the entry without a re-scan. `ScanStageCache::create_output` compensates by filtering `DynamicImport` entries against module reachability (`compute_live_modules`, a BFS from the non dynamic entries), so only entries a live module dynamically imports produce chunks, matching a fresh build. This covers the transitive case: `main` imports `hub`, `hub` dynamically imports `lazy`, and dropping the `hub` import from `main` must also stop emitting the `lazy` chunk even though `hub` is never re-scanned.
- Partial scans no longer create a `DynamicImport` entry for a module that is a configured user entry. The scan local `user_defined_entry_ids` set is empty in partial mode, so the existing full scan filter never matched there, and a dynamic import of the entry module produced a second entry row (and a second chunk) for it.

#### Tests

- New fixture `hmr/remove_transitive_dynamic_import_entry`: `main` statically imports `hub`, `hub` dynamically imports `lazy`. Dropping the `hub` import from `main` removes the `lazy` chunk from the build output without re-scanning `hub`; re-adding it revives the chunk.
- New fixture `hmr/remove_dynamic_import_entry`: two modules dynamically import `lazy.js` and drop it one by one. After the first drop the entry survives through the remaining importer. After the second drop the `lazy.js` chunk disappears from the build output. Re-adding the import brings the entry and the chunk back.
- The state parity check from #10115 is tightened for this: `DynamicImport` entries now must match a fresh full build exactly while a live module imports them; an entry kept in the cache only by orphans' dynamic import records is tolerated, since `create_output` filters it from the output and it exists precisely to make revival cheap (other kinds still allow extras on the incremental side until orphan cleanup lands). Reintroducing the bug locally makes the parity check fail on the new fixture with "dynamic import entry only exists in the incremental state", so the whole HMR suite now guards this behavior.

#### Docs

The entry point step of the `merge` algorithm in `internal-docs/cache/implementation.md` is updated.

### DIFF
--- a/crates/rolldown/src/bundler/impl_bundler_testing.rs
+++ b/crates/rolldown/src/bundler/impl_bundler_testing.rs
@@ -1,6 +1,6 @@
 use std::collections::BTreeSet;
 
-use rolldown_common::{Module, ModuleIdx, ModuleTable};
+use rolldown_common::{EntryPointKind, Module, ModuleIdx, ModuleTable};
 
 use super::Bundler;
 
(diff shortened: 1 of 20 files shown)


shulaoda:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10117?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#10117** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10117?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10117?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10110** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10110?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10107** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10107?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10117).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #10127, opened 2026-07-03
chore: remove dead BuildDiagnostic::downcast_ref method

shulaoda:
### What

Removes the `BuildDiagnostic::downcast_ref` method from `rolldown_error`.

### Why

The method has no callers anywhere in the workspace. The `downcast_ref` calls that do exist (in `napi_error.rs` and `plugin_error.rs`) are `anyhow::Error::downcast_ref`, downcasting an `anyhow::Error` into `napi::Error` or `BuildDiagnostic`, not this method. The sibling `downcast_mut` is still used (by the module loader) and stays.


shulaoda:

PR #10128, opened 2026-07-03
chore: remove dead LazyCompilationContext::is_lazy_module method

shulaoda:
### What

Removes the `LazyCompilationContext::is_lazy_module` method from `rolldown_plugin_lazy_compilation`.

### Why

The method has no callers anywhere in the workspace. It is a dead `pub fn` on the context shared between the plugin and the dev engine. The sibling `mark_as_fetched` on the same impl is still used and stays.


shulaoda:

PR #10129, opened 2026-07-03
chore: remove dead CopyModulePlugin::is_active method

shulaoda:
### What

Removes the `CopyModulePlugin::is_active` method from `rolldown_plugin_copy_module`.

### Why

The method has no callers anywhere in the workspace. The plugin is registered unconditionally in `apply_inner_plugins`, and the `resolve_id` hook does its own inline `self.copy_extensions.is_empty()` check rather than going through `is_active`, so the method is dead.


shulaoda:

PR #10132, opened 2026-07-04
fix(hmr): fall back to full reload when a changed module is not registered as executed

### LINKED ISSUE #10149: Infra: Flaky node dev server test workflow
https://github.com/rolldown/rolldown/actions/runs/28724034674/job/85178167635?pr=10139

@h-a-n-a @shulaoda cc

shulaoda:
closes #10149

## Problem

Editing a file right after a page loads can be silently lost in full bundle mode. The dev engine decides whether a client needs an HMR update by checking the client's `executed_modules` set, but that set is reported asynchronously by the browser runtime (batched `hmr:module-registered` messages over the socket) and can lag what the client actually ran. When the edit is processed before the reports arrive, `compute_out_hmr_prerequisites` skipped the changed module for that client and produced an empty update. Vite's bundled dev ignores updates that carry no patches, and the HMR module cache had already advanced past the edit, so nothing ever re-delivers it. The page stays stale forever with no error.

This is what turned main red after #10084. The vite-plus 0.2.2 bump added CPU contention on the 4-core CI runners, which stretched the report lag past the watcher debounce window, so the first edit in `playground/hmr-full-bundle-mode` started losing the race in roughly one of five jobs ([example run](https://github.com/rolldown/rolldown/actions/runs/28688065031/job/85085418310)). The three test failures in that suite are a cascade from the single lost edit.

## Fix

`compute_out_hmr_prerequisites` now receives the existing `changed_modules` set. When a changed module is not registered as executed by a client, the update for that client falls back to a full reload with an explicit reason instead of being skipped. The default `RebuildStrategy::Auto` upgrades the task to a rebuild, and Vite defers the reload until `onOutput`, so the reload always lands on fresh output.

The invalidate flow passes an empty `changed_modules` set and keeps the old skip. Nothing changed on disk there, so an unexecuted importer has nothing to re-run. The importer pruning inside `propagate_update` is also untouched, since that behavior is pinned by the `client-module-execution-status` fixture.

Known cost: with lazy compilation, a client that never fetched the changed lazy module now reloads unnecessarily. We cannot tell it apart from a client whose loaded bundle bakes in the now stale initializer without a bundle epoch in the protocol.

## Tests

A new snapshot fixture `topics/hmr/changed_module_not_executed_by_client` drives an HMR step against a client with an empty `executed_modules` set, enabled by a new `dev.unregisteredClient` test config flag. On the old code the same fixture produces an empty patch (`applyUpdates([])`) instead of the full reload, so the snapshot pins the fix. All existing HMR snapshots, the full integration suite, and the node dev server fixtures pass unchanged.

## Follow ups

Two narrower windows of the same class remain and are left for separate issues. A changed module that is already registered while its accept boundary importer is not can still yield an empty patch. And when no client is registered at all, the update list is empty and Vite still ignores it, which needs a check on the Vite side when a client first registers.


h-a-n-a:
I would merge this PR as a workaround. The new design of HMR is going to fix this, which removes the lag and `executed_modules` itself.

PR #10135, opened 2026-07-04
refactor: type construct_vite_preload_call to take an ObjectPattern

shulaoda:
### What

Changes `BuildImportAnalysisVisitor::construct_vite_preload_call` to take `Box<ObjectPattern>` instead of a generic `BindingPattern`, and updates its three callers to pass the `ObjectPattern` directly.

Inside the function, the value is wrapped back into `BindingPattern::ObjectPattern(..)` at the two spots that need a `BindingPattern` (the `const { .. } = await import()` declarator and the `return { .. }` expression).

### Why

`construct_vite_preload_call` had a `BindingPattern::BindingIdentifier` branch that was never reached. All three callers only ever pass an object pattern:

- `rewrite_member_expr` builds an `ObjectPattern` inline
- `rewrite_call_expr` only proceeds when the `.then()` callback's first parameter is an `ObjectPattern`
- `visit_variable_declaration` only proceeds when the declarator id is an `ObjectPattern`

Since `ast_utils` and `ast_visit` are private modules, the method is not reachable from outside the crate, so those three are the complete set of callers and the identifier branch was dead.

Rather than deleting the dead branch and keeping the loose `BindingPattern` parameter, encoding the `ObjectPattern` invariant in the signature removes the branch by construction and lets the callers drop their now-redundant `matches!` guards and whole-pattern clones.

### Verification

- `cargo check` / `cargo clippy --deny warnings` / `cargo fmt --all --check`
- `just build-rolldown` rebuilds the `.node` binding; `tsc` passes
- The `build-import-analysis` fixtures (destructuring, member access, nested import, complex syntax, and so on) all pass with zero snapshot drift, confirming the generated transform output is unchanged

### DIFF
--- a/crates/rolldown_plugin_vite_build_import_analysis/src/ast_utils.rs
+++ b/crates/rolldown_plugin_vite_build_import_analysis/src/ast_utils.rs
@@ -61,7 +61,7 @@ impl<'a> BuildImportAnalysisVisitor<'a> {
       *await_expr = Expression::AwaitExpression(AwaitExpression::boxed(
         SPAN,
         self.construct_vite_preload_call(
-          BindingPattern::ObjectPattern(ObjectPattern::boxed(
+          ObjectPattern::boxed(
             SPAN,
             oxc::allocator::Vec::from_value_in(
               BindingProperty::new(
@@ -76,7 +76,7 @@ impl<'a> BuildImportAnalysisVisitor<'a> {
             ),
             NONE,
             &self.ast_factory,
-          )),
+          ),
           await_expr.take_in(&self.ast_factory.allocator()),
         ),
         &self.ast_factory,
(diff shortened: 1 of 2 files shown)


sapphi-red (approved):

shulaoda:
### Merge activity

* **Jul 6, 7:12 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 6, 7:12 AM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 6, 7:20 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

Issue #6995, opened 2025-11-13
[Feature Request]: Filesystem Build Cache


Boshen:
### What problem does this feature solve?

Build cache is needed when slowness is unavoidable:
 
* Slower js plugins. e.g. react, vue, angular compiler.
* Way too many modules. e.g. 10k+ modules.

This unlocks:

* Rolldown build produces the cache -> vitest picks them up.

Resources:

* https://webpack.js.org/configuration/cache

### What does the proposed API look like?

To be announced.

IWANABETHATGUY:
It seems this issue is duplicated with https://github.com/rolldown/rolldown/issues/4090

shalldie:
This is a crucial basis for me to promote the team to upgrade to Vite.

PR #10137, opened 2026-07-04
feat: add experimental persistent transform cache (experimental.transformCache)

christianvuerings:
### Description

This PR is a concrete, minimal implementation proposal for the filesystem build cache tracked in #6995, following the direction discussed in #802. Since new features should be agreed on first, this is opened as a **draft** to give that discussion something tangible to evaluate; happy to iterate on any part of it or split it up.

It adds `experimental.transformCache`: an opt-in filesystem cache that persists per-module plugin `transform` pipeline results (code, module type, side-effect override, transform sourcemap chain) across builds and processes. On a cache hit the entire transform pipeline is skipped for that module; `load` hooks, parsing and all later stages run unchanged.

```js
export default {
  experimental: {
    // or `transformCache: true` for the defaults
    transformCache: {
      dir: 'node_modules/.cache/rolldown', // default
      key: myConfigAndLockfileHash,        // extra invalidation key
    },
  },
};
```

### Motivation

We are migrating a very large monorepo (100k+ modules, CI fleets building overlapping module graphs all day) at our large company from a legacy bundler to Rolldown. Rolldown itself is fast, but as #6995 notes, two things still dominate cold builds: slow JavaScript `transform` plugins (every call crosses the NAPI boundary and runs single-threaded JS) and sheer module count. A persistent transform cache removes the dominant cost for unchanged modules, which is almost all of them on a typical CI build.

The on-disk layout is deliberately simple and content-addressed (`<dir>/transform-v1/<hash[0..2]>/<hash>`, one self-contained file per entry, atomic rename writes, machine-portable keys built from cwd-relative module ids) so that **remote caching** can be layered on top externally: CI writers populate an object store, developers and CI readers warm the local directory from an HTTP mirror. That is the same architecture we run today on our legacy bundler (memory, then disk, then HTTP/S3), and this PR is the building block that makes it possible with Rolldown without Rolldown itself needing any network code.

### Design summary

- Cache seam is exactly one place: `ModuleTask::load_source` routes string sources through `transform_source_with_cache` instead of `transform_source` when enabled.
- Entry key: `xxh3(salt, stable_id, module_type, xxh3(post_load_source))`, with `salt = xxh3(format_version, rolldown_version, options.key, ordered plugin names)`.
- Entry format: magic + version byte + JSON metadata header + raw code bytes (keeps the biggest blob unescaped; no new dependencies, `serde_json` and `xxhash` were already in the tree).
- Correctness details:
  - Side effects are stored as a delta: only a transform-hook override is replayed, so hits never clobber side effects derived from resolution data (`package.json` `sideEffects`) that sits outside the cache key.
  - The transform sourcemap chain (including `Omitted`/`Null` markers used for `SOURCEMAP_BROKEN` warnings) is persisted and replayed; warm builds produce byte-identical chunks and sourcemaps.
  - The cache silently disables itself when `experimental.nativeMagicString` + sourcemaps is active, because that path generates transform maps outside the module task.
  - Every cache failure (corrupt entry, unreadable dir, write error) degrades to a miss or a dropped write; the cache can never fail a build.
- Cache reads/writes go through `spawn_blocking`, mirroring how `load_source` does file IO.

`internal-docs/transform-cache/{design,implementation}.md` in this PR cover the rationale, rejected alternatives and data flow in more detail.

### Alternatives explored

- Persisting scan results (AST, symbols, scopes), as originally benchmarked in #802. The oxc AST is arena-allocated and self-referential, so this requires serializable mirrors of large internal types. The transform pipeline is where the expensive JS work lives, so caching at that seam captures most of the win at a small fraction of the surface area and risk.
- Automatic invalidation on plugin config changes. Rolldown cannot observe plugin configuration or implementations, so this follows the established contract of webpack's `cache.version` and Metro's `getCacheKey`: callers fold a config/lockfile hash into `key`. Ordered plugin names are part of the salt, so adding/removing/reordering plugins invalidates automatically.
- A pluggable get/set store trait for remote backends inside Rolldown. Deliberately deferred; the dumb directory layout supports external sync tooling today and a trait can be added later without changing the entry format.

### Known limitations (documented in the option's JSDoc and internal docs)

- Transform-hook side channels (`this.emitFile`, `this.addWatchFile`, custom module meta written during `transform`) are not replayed on hits.
- No eviction policy; entries are content-addressed and callers manage the directory lifecycle.

### Tests

- `crates/rolldown/tests/rolldown/topics/transform_cache`: warm build runs zero transform hooks and produces identical output; changing `key` invalidates all entries.
- Unit tests for the entry codec (roundtrip, version/garbage rejection).
- Verified through the JS API with the built binding: a JS transform plugin is called once cold and zero times warm across processes, `map` mappings are byte-identical cold vs warm, and a changed `key` re-runs transforms.
- `cargo clippy --workspace --all-targets` clean; full `--test integration` suite passes (the only failure locally is `test262_module_code` from an uninitialized `test262` submodule).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2434,6 +2434,7 @@ dependencies = [
  "oxc_allocator",
  "oxc_ecmascript",
  "oxc_index",
+ "oxc_sourcemap",
  "oxc_str",
  "oxc_traverse",
  "petgraph",
--- a/crates/rolldown/Cargo.toml
+++ b/crates/rolldown/Cargo.toml
@@ -34,6 +34,7 @@ oxc = { workspace = true }
 oxc_allocator = { workspace = true }
 oxc_ecmascript = { workspace = true }
 oxc_index = { workspace = true }
+oxc_sourcemap = { workspace = true }
 oxc_str = { workspace = true }
 oxc_traverse = { workspace = true }
 petgraph = { workspace = true }
--- a/crates/rolldown/src/module_loader/module_loader.rs
+++ b/crates/rolldown/src/module_loader/module_loader.rs
@@ -34,6 +34,7 @@ use tracing::Instrument;
 use crate::module_loader::module_task::ModuleTaskOwner;
 use crate::types::scan_stage_cache::ScanStageCache;
 use crate::utils::load_entry_module::load_entry_module;
+use crate::utils::transform_cache::TransformCache;
 use crate::{SharedOptions, SharedResolver};
 
 use super::external_module_task::ExternalModuleTask;
@@ -190,13 +191,19 @@ impl<'a, Fs: FileSystem + Clone + 'static> ModuleLoader<'a, Fs> {
       },
     };
 
+    // The native magic-string path generates transform sourcemaps outside the
+    // module task, so cache entries could not capture them; bypass the cache.
+    let transform_cache =
+      if magic_string_tx.is_none() { TransformCache::new(&options, &plugin_driver) } else { None };
+
     // Unbounded as defense in depth. If any sender ends up driven from a sync
     // napi binding through `block_on`, a bounded channel could deadlock: the
     // send future would wait on capacity that only the consumer can free, but
     // the consumer may need the JS thread — pinned by `block_on` — to run
     // plugin hooks first. Keeping the channel unbounded removes that edge.
     let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
-    let shared_context = Arc::new(TaskContext { options, tx, resolver, fs, plugin_driver, meta });
+    let shared_context =
+      Arc::new(TaskContext { options, tx, resolver, fs, plugin_driver, meta, transform_cache });
 
     let importers = std::mem::take(&mut cache.importers);
     let intermediate_normal_modules = IntermediateNormalModules::new(is_full_scan, importers);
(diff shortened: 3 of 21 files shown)


christianvuerings:
CI status note: the only remaining failure, node-dev-server-test-ubuntu (24), is the known hmr-full-bundle-mode flake that currently also fails on main itself (for example https://github.com/rolldown/rolldown/actions/runs/28701829934 on the commit this branch is based on, where it failed on Node 20, and https://github.com/rolldown/rolldown/actions/runs/28688715267 where it failed on Node 22). The failure mode (first HMR patch never applied, then cascading editFile mismatches in the shared playground) looks like what #10132 addresses. Happy to rebase once that lands. The earlier Rust Validation failure was real (an unfulfilled clippy expect) and is fixed in f0edf82.

christianvuerings:
Update: CI is fully green as of 336e700 (the hmr-full-bundle-mode flake did not reproduce on this run).

h-a-n-a:
Thanks for your contribution! 

Currently we haven't figure out the technical details for persistent cache support in Rolldown. But one thing to add is that we're not adding cache for only `transform` stage and we're targeting to the whole bundler pipeline. That said, this PR may only be part of the roadmap that we're exploring and there's something missing. However, the team's bandwidth is very low at the moment. We may not able to proceed this PR and design at the moment. 

According to Rolldown [contribution guide](https://rolldown.rs/contribution-guide/#contribution-guide), I'm closing this for now. Thanks for your understanding. Nevertheless, we'd also like to hear from your thoughts in the discussion of persistent cache support.

christianvuerings:
@h-a-n-a Thanks, understood! Quick context: this is the blocker for our Metro to Rolldown migration, so we're keen to help push it forward. Given the team's low bandwidth, could we take on the design and implementation ourselves? We can propose a whole-pipeline design in #6995, build it out, and test it rigorously at scale on our monorepo. What next steps would be most useful for the team?

PR #10138, opened 2026-07-04
chore(rolldown_utils): remove unused time module

shulaoda:
### What

Deletes `crates/rolldown_utils/src/time.rs` (its only item, `current_utc_timestamp_ms`) and drops `pub mod time;` from the crate root.

### Why

`rolldown_utils::time::current_utc_timestamp_ms` has no callers anywhere in the workspace. The only place that uses a function by that name is `rolldown_devtools`, which defines its own local `current_utc_timestamp_ms` (returning `u128`) and calls that, not this `String`-returning one. There is no qualified use of `rolldown_utils::time` and no re-export, so the module is dead.


shulaoda:

PR #10139, opened 2026-07-05
fix(watcher): stop the coordinator when a watcher is dropped without close

shulaoda:
### What

Adds `impl Drop for Watcher` that signals the coordinator to shut down, and extracts the synchronous part of `close` into a shared `signal_close` helper that both `close` and `drop` call. Also adds a regression test.

### Why

The spawned coordinator owns the tasks, and each task's fs watcher holds a `tx.clone()` of the `WatcherMsg` channel. So the coordinator transitively keeps senders alive, and its `rx.recv()` never observes a channel close. The `None => break` arms in the event loop are therefore unreachable, and a `Watcher` dropped without an explicit `close()` leaves the coordinator task (and all its tasks, bundlers, and fs watchers) running forever.

`close` cannot be reused directly from `drop`, since it is `async` (it awaits the coordinator finishing) and `drop` is synchronous. `drop` runs only the synchronous signalling part, which is enough for the coordinator to observe the `Close` message and stop; the spawned task then finishes on its own.

Calling `close` and then dropping stays safe: the second signal is idempotent, and the final `tx.send` just fails on the already-closed channel and is ignored.

### Verification

- `cargo clippy -p rolldown_watcher --deny warnings` / `cargo fmt --all --check`
- `cargo test -p rolldown_watcher` passes
- The added `dropping_a_running_watcher_shuts_down_the_coordinator` test runs a real (in-memory, `skip_write`) build, waits for it to reach `Idle`, drops the watcher without `close`, and asserts `on_close` fires. It was confirmed to fail (30s timeout) when the `Drop` impl is removed, so it genuinely guards the leak.


hyfdev:
Thanks for digging into this. The core diagnosis in the Why section makes sense for a directly-owned `rolldown_watcher::Watcher`: the coordinator owns the tasks, each task's fs watcher keeps a `tx.clone()` alive, so `rx.recv()` will not observe channel close, and `Drop` needs a synchronous close signal.

I think this still does not cover the public JS/NAPI path, though. `BindingWatcher` stores `Arc<rolldown_watcher::Watcher>`, and the JS watcher starts a non-awaited `inner.waitForClose()` as the event-loop keepalive. The binding implementation clones that `Arc` into the pending `wait_for_close` future before awaiting the coordinator handle. That means if the JS watcher is abandoned without an explicit `close()`, the pending keepalive future can still be the strong owner of `Watcher`, so `Watcher::drop()` never runs and this new `Drop` impl never gets a chance to send `WatcherMsg::Close`.

I verified the PR's new Rust test passes, and `cargo test -p rolldown_watcher`, `cargo clippy -p rolldown_watcher --all-targets -- --deny warnings`, and `cargo fmt --all --check` pass locally. But I also tried a local regression-style test that mimics `Arc<Watcher> + pending wait_for_close()`, and dropping the outer handle did not trigger `on_close`; only an explicit `close()` ended the watcher.

So I think this is a valid low-risk improvement for the direct Rust API, but it should either narrow the claim to that scope or also adjust the NAPI/keepalive ownership so the JS leak path is actually fixed.

PR #10154, opened 2026-07-06
feat(binding): record plugin hook result kind in tracing spans

IWANABETHATGUY:
## What this PR does

Adds a `result_kind` field to the tracing spans of the per-module JS plugin hooks — `resolveId`, `resolveDynamicImport`, `load`, `transform`, and `renderChunk`. The field is recorded as:

- `"value"` when the hook returns a value, and
- `"null"` when it returns nothing (an early return).

`Err` results are intentionally left unrecorded — a thrown hook is not an early return. A small `record_hook_result` helper does the recording after each hook resolves, and each span now declares `result_kind = tracing::field::Empty` up front so the value can be attached to the span's END event.

## Why

In a `chrome-json` trace the recorded value lands on the span's END event, so counting `result_kind == "null"` per plugin gives an accurate tally of how often each hook early-returns. That's the primary signal for "this hook should declare a `filter`", and it's now available straight from a trace with **no JS-side instrumentation**.

## Notes for reviewers

- No user-facing/runtime behavior change — this only adds a tracing field. When tracing is disabled the `record`/`instrument` calls are effectively no-ops.
- The hook bodies were restructured from a trailing method chain into a `let span = …; let result = …; record_hook_result(&span, &result); result` shape so the span can be recorded onto once the result is known. `resolve_dynamic_import` additionally swaps its `match` for a `let … else` early return.

### DIFF
--- a/crates/rolldown_binding/src/options/plugin/js_plugin.rs
+++ b/crates/rolldown_binding/src/options/plugin/js_plugin.rs
@@ -38,6 +38,20 @@ pub struct JsPlugin {
   pub(crate) filter_expr_cache: FilterExprCache,
 }
 
+/// Records whether a per-module hook (resolveId / load / transform) produced a
+/// value or returned nothing, onto the hook's tracing span's `result_kind`
+/// field. In a `chrome-json` trace the value lands on the span's END event, so
+/// counting `result_kind == "null"` per plugin gives an accurate early-return
+/// tally (the primary "this hook should have a `filter`" signal) without any
+/// JS-side instrumentation. `Err` results are left unrecorded — a thrown hook
+/// is not an early return. The span must declare `result_kind = field::Empty`.
+#[inline]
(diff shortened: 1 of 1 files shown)


shulaoda (approved):

IWANABETHATGUY:
### Merge activity

* **Jul 7, 7:28 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 7, 7:28 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 7, 7:33 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10159, opened 2026-07-06
feat: add experimental persistent build cache (experimental.buildCache)

christianvuerings:
### Description

This is a follow-up to #10137 (persistent cache for the `transform` stage only), which was closed with the feedback that rolldown is [targeting caching for the whole bundler pipeline, not just `transform`](https://github.com/rolldown/rolldown/pull/10137#issuecomment-4889927322). This PR extends that work into a concrete, minimal implementation of a whole-module-pipeline persistent cache, and is opened as a **draft** to give the persistent cache design work (#6995, #802) something tangible to evaluate and benchmark against. If the team prefers to keep this in the design phase for now, closing it is completely fine; the design notes and benchmark data below may still be useful input.

It adds `experimental.buildCache`: an opt-in filesystem cache that persists, per module, everything a module task produces before the native parse: the plugin `load` pipeline output, the plugin `transform` pipeline output (code, module type, side-effect override, sourcemap chain) and the resolved dependencies (`resolveId` hooks plus the native resolver, including `package.json` side-effects data). On a cache hit **no plugin hook and no resolver call runs for that module**; parsing, scanning and all later stages run unchanged on the cached post-transform source.

```js
export default {
  experimental: {
    // or `buildCache: true` for the defaults
    buildCache: {
      dir: 'node_modules/.cache/rolldown', // default
      key: myConfigAndLockfileHash,        // extra invalidation key
    },
  },
};
```

### Benchmarks (10,000 modules, before/after)

Full methodology, raw runs and reproduction scripts: https://gist.github.com/christianvuerings/e43cdf8b2597cdf8565f99ca7e4b57a4. Every run is a fresh process, so warm numbers are cross-process. Release builds, Apple M3 Max.

| Scenario (10,001 modules) | no cache | warm cache | speedup |
| --- | ---: | ---: | ---: |
| babel in `transform` hook | 34,270 ms | 422 ms | 81x |
| babel in `transform` + no-op `resolveId`/`load` hooks | 33,594 ms | 428 ms | 78x |
| babel in `load` hook (SFC-compiler shape) | 34,395 ms | 433 ms | 79x |
| no JS plugins (native only) | 213 ms | 218 ms (cache self-disables) | ~none |

Two rows worth calling out:

- **babel in `load`**: a transform-only cache (#10137, measured on the same fixture) gives **zero** benefit here (34,838 ms warm) because its key is the post-`load` source, so `load` hooks must run on every build. Keying on raw file content lets this design skip the `load` pipeline too, which is where Vue SFC / Svelte-style compilers do their work.
- **no JS plugins**: with nothing expensive to skip, the cache would only add fs reads and entry decoding (measured 1.8x slower warm, 5.7x slower cold before this guard existed), so `BuildCache::new` returns `None` when no plugin beyond rolldown's cheap inner plugins registers a `resolveId`/`load`/`transform` hook; such builds never touch the cache directory.

Cold builds (cache being written) cost 1-2% on plugin-heavy builds. Warm output is byte-identical to an uncached build (`diff -rq` over the whole 10k-module output, both scenarios, and byte-identical sourcemap `mappings` in the JS tests).

### Design summary

- Cache seam is exactly one place: `ModuleTask::run_inner` tries the cache before `load_source` and stores after `resolve_dependencies`.
- Entry key: `xxh3(salt, stable_id, asserted_module_type, xxh3(raw file bytes))`, with `salt = xxh3(format_version, rolldown_version, options.key, platform, sorted moduleTypes, ordered plugin names)`. Keys and entry contents are machine-portable (cwd-relative paths), so external tooling can layer **remote caching** on top by syncing the directory (CI writers populate an object store, readers warm the local dir), without rolldown needing any network code.
- Resolved dependencies are stored positionally: re-scanning the cached code deterministically reproduces the import records they align with, so import records themselves never need serialization (and the arena-allocated oxc AST stays out of the cache entirely; re-parsing is the cheap, safe part).
- Correctness details:
  - On a hit every cached non-external dependency is stat-ed; a deleted or moved file turns the entry into a miss so the importer re-resolves and fails or externalizes exactly like a cold build.
  - The full `ResolvedId` is serialized including `package.json` `sideEffects` data, so tree shaking is identical cold vs warm.
  - Side effects are stored as a delta: only a load/transform-hook override is replayed, never side effects derived from resolution data outside the cache key.
  - Modules whose resolution produced warnings (e.g. missing import treated as external) are never stored, so warnings replay on every build.
  - The whole sourcemap chain (`Load`, `Transform`, `Omitted`, `Null` elements) is persisted and replayed.
  - Virtual modules, data URLs and `rolldown:` ids bypass the cache; the runtime module task is untouched. The cache silently disables itself when `experimental.nativeMagicString` + sourcemaps is active.
  - Every cache failure (corrupt entry, unreadable dir, write error) degrades to a miss or a dropped write; the cache can never fail a build.
  - The cache disables itself when no plugin registers a cacheable hook, so enabling it on a plugin-free config costs nothing.
- Entry format: magic + version byte + JSON metadata header + raw code bytes, one self-contained file per entry at `<dir>/build-v1/<hash[0..2]>/<hash>`, atomic rename writes. No new dependencies (`serde_json`, `xxhash` and `oxc_sourcemap` were already in the tree).
- `internal-docs/build-cache/{design,implementation}.md` cover the rationale, rejected alternatives and data flow in more detail.

### Known limitations (documented in the option's JSDoc and internal docs)

- `load`/`transform` hooks are assumed to be deterministic functions of the module's id and content; hook side channels (`this.emitFile`, `this.addWatchFile`, custom module meta) are not replayed on hits.
- Resolution inputs outside module files (lockfile changes, `tsconfig.json` paths, newly added files that shadow an old resolution) and plugin config changes are invisible to the key; callers fold a config/lockfile hash into `key`, the same contract as webpack's `cache.version` and Metro's `getCacheKey`.
- No eviction policy; entries are content-addressed and callers manage the directory lifecycle.

### Tests

- `crates/rolldown/tests/rolldown/topics/build_cache`: warm build runs zero `resolveId`/`load`/`transform` hooks with identical output; `key` invalidation; editing one module re-runs only that module's pipeline; deleting a cached dependency falls back to fresh resolution instead of replaying stale state; a build with no cacheable plugin hooks never creates the cache directory.
- Unit tests for the entry codec (roundtrip including resolved deps and `package.json` side-effects globs, version/garbage rejection).
- `packages/rolldown/tests/behaviors/build-cache.test.ts`: the same assertions through the JS plugin path, plus byte-identical sourcemap `mappings` cold vs warm.
- `cargo clippy --workspace --all-targets` and `cargo fmt --check` clean; full `--test integration` suite passes (the only failure locally is `test262_module_code` from an uninitialized `test262` submodule); `vp check`, knip and the node type tests pass.

### AI usage disclosure

Per the [contribution guide's AI policy](https://rolldown.rs/contribution-guide/#ai-usage-policy): this PR was implemented with AI assistance (Claude Code). The design, code, tests and docs were reviewed, and all benchmark numbers come from real runs (methodology and scripts in the linked gist).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2443,6 +2443,7 @@ dependencies = [
  "oxc_allocator",
  "oxc_ecmascript",
  "oxc_index",
+ "oxc_sourcemap",
  "oxc_str",
  "oxc_traverse",
  "petgraph",
--- a/crates/rolldown/Cargo.toml
+++ b/crates/rolldown/Cargo.toml
@@ -34,6 +34,7 @@ oxc = { workspace = true }
 oxc_allocator = { workspace = true }
 oxc_ecmascript = { workspace = true }
 oxc_index = { workspace = true }
+oxc_sourcemap = { workspace = true }
 oxc_str = { workspace = true }
 oxc_traverse = { workspace = true }
 petgraph = { workspace = true }
--- a/crates/rolldown/src/module_loader/module_loader.rs
+++ b/crates/rolldown/src/module_loader/module_loader.rs
@@ -33,6 +33,7 @@ use tracing::Instrument;
 
 use crate::module_loader::module_task::ModuleTaskOwner;
 use crate::types::scan_stage_cache::ScanStageCache;
+use crate::utils::build_cache::BuildCache;
 use crate::utils::load_entry_module::load_entry_module;
 use crate::{SharedOptions, SharedResolver};
 
@@ -193,13 +194,19 @@ impl<'a, Fs: FileSystem + Clone + 'static> ModuleLoader<'a, Fs> {
       },
     };
 
+    // The native magic-string path generates transform sourcemaps outside the
+    // module task, so cache entries could not capture them; bypass the cache.
+    let build_cache =
+      if magic_string_tx.is_none() { BuildCache::new(&options, &plugin_driver) } else { None };
+
     // Unbounded as defense in depth. If any sender ends up driven from a sync
     // napi binding through `block_on`, a bounded channel could deadlock: the
     // send future would wait on capacity that only the consumer can free, but
     // the consumer may need the JS thread — pinned by `block_on` — to run
     // plugin hooks first. Keeping the channel unbounded removes that edge.
     let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
-    let shared_context = Arc::new(TaskContext { options, tx, resolver, fs, plugin_driver, meta });
+    let shared_context =
+      Arc::new(TaskContext { options, tx, resolver, fs, plugin_driver, meta, build_cache });
 
     let importers = std::mem::take(&mut cache.importers);
     let intermediate_normal_modules = IntermediateNormalModules::new(is_full_scan, importers);
--- a/crates/rolldown/src/module_loader/module_task.rs
+++ b/crates/rolldown/src/module_loader/module_task.rs
@@ -1,3 +1,4 @@
+use std::path::Path;
 use std::sync::Arc;
 
 use arcstr::ArcStr;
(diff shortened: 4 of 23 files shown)


christianvuerings:
@IWANABETHATGUY could you take a look at this PR? It would unblock a big monorepo migration to Rolldown

christianvuerings:
@h-a-n-a could you take a look at this PR? I addressed your feedback from #10137

h-a-n-a:
> @h-a-n-a could you take a look at this PR? I addressed your feedback from #10137

Again, the team's bandwidth is very low at the moment since we put most of our time on moving bundled-dev in Vite forward. We may not able to proceed this PR and design at the moment.

PR #10161, opened 2026-07-07
chore(deps): update oxc to 0.139.0

shulaoda:
Updates oxc from `0.138.0` to `0.139.0`.

### Breaking change: `MangleOptions` gained a `reserved` field

oxc 0.139.0 adds `mangle.reserved` (names that bindings must not be renamed to, and that bindings already carrying them keep). Since the generated TS type now includes it, the valibot schema must cover it. To avoid the option being accepted by validation but silently dropped, it is wired through end to end:

- `RawMangleOptions` gains an optional `reserved` field, and `into_mangle_options` passes it to oxc (defaults to empty).
- `normalize_binding_options.rs` reads `reserved` from the napi mangle options.
- `minify_options_conversion.rs` propagates it in the reverse conversion.
- `MangleOptionsSchema` in `validator.ts` accepts `reserved?: string[]`.

So `minify.mangle.reserved` now works from the JS API down to the mangler.

### Breaking change: oxc moved to num-bigint 0.5

Bumped the workspace `num-bigint` from `0.4.6` to `0.5.1` so `ConstantValue::BigInt` uses the same `BigInt` type as `oxc_ecmascript`. The `num-bigint 0.4.8` still in the lockfile is a test-only transitive dependency of `jsonschema`.

### Snapshot changes

From minifier improvements in 0.139.0: pure IIFEs like `(() => 123)()` are now removed by DCE, and a single-use const is inlined into its `console.log` call site. Embedded `@oxc-project/runtime` helpers regenerated for 0.139.0.

`cargo check`, `just roll` (Rust and Node tests, clippy, fmt, lints, esbuild diff) all pass. Rollup compat tests: 1216 passed, 0 failed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -454,9 +454,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b82ac4a3c2ca9c3460964f020e1402edd5753411d7737aa39c3714ad1b5420e"
+checksum = "2d6914041f254d6e9176c01941b21115dcfb7089e55135a35411081bd106ef3f"
 dependencies = [
  "crossbeam-utils",
 ]
@@ -670,7 +670,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "843fba2746e448b37e26a819579957415c8cef339bf08564fe8b7ddbd959573c"
 dependencies = [
  "crc32fast",
- "miniz_oxide",
+ "miniz_oxide 0.8.9",
  "zlib-rs",
 ]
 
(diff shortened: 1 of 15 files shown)


sapphi-red (approved):

hyfdev (approved):

shulaoda:
### Merge activity

* **Jul 7, 5:44 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 7, 5:44 AM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 7, 5:49 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #7625, opened 2025-12-22
feat: add import path for unresolved import diagnostics

sapphi-red:
Added a help for UNRESOLVED_IMPORT error to output the import path.
```
[UNRESOLVED_IMPORT] Error: Could not resolve '../pkg' in node_modules/.pnpm/lightningcss@1.30.2/node_modules/lightningcss/node/index.js
    ╭─[ node_modules/.pnpm/lightningcss@1.30.2/node_modules/lightningcss/node/index.js:17:28 ]
    │
 17 │   module.exports = require(`../pkg`);
    │                            ────┬───  
    │                                ╰───── Module not found.
    │ 
    │ Help: 'node_modules/.pnpm/lightningcss@1.30.2/node_modules/lightningcss/node/index.js' is imported by the following path:
    │         - node_modules/.pnpm/lightningcss@1.30.2/node_modules/lightningcss/node/index.js
    │         - node_modules/.pnpm/lightningcss@1.30.2/node_modules/lightningcss/node/index.mjs
    │         - node_modules/.pnpm/vite@8.0.0-beta.4_@types+node@24.10.4/node_modules/vite/dist/node/chunks/node.js
    │         - node_modules/.pnpm/vite@8.0.0-beta.4_@types+node@24.10.4/node_modules/vite/dist/node/index.js
    │         - src/main.tsx
    │         - index.html
────╯
```

PR #10162, opened 2026-07-07
fix(dev): don't panic when an HMR rebuild hits an unresolved import

### LINKED ISSUE #10136: [Panic]: thread 'rolldown-worker' (11046093) panicked at crates/rolldown_common/src/types/hybrid_index_vec.rs:76:28
### Panic message

```Shell
Rolldown panicked. This is a bug in Rolldown, not your code.

thread 'rolldown-worker' (11046093) panicked at crates/rolldown_common/src/types/hybrid_index_vec.rs:76:28:
HybridIndexVec::Map missing idx 2 (len=1)

Please report this issue at: https://github.com/rolldown/rolldown/issues/new?template=panic_report.yml
```

### Reproduction

https://github.com/btea/rolldown-panic

1. Running `pnpm dev` works perfectly.
2. The following changes were made to line 2 of the `src/App.vue` file, and after saving, an error occurred.
https://github.com/btea/rolldown-panic/blob/3e7a512b0697d3985749e4a7228531f7d7d18bc6/src/App.vue#L2

```diff
-import HelloWorld from './components/HelloWorld.vue'
+import HelloWorld from './src/components/HelloWorld.vue'
```

### System Info

```Shell
System:
    OS: macOS 14.4.1
    CPU: (8) arm64 Apple M3
    Memory: 143.34 MB / 16.00 GB
    Shell: 3.2.57 - /bin/sh
  Binaries:
    Node: 24.18.0 - /Users/xxx/.vite-plus/js_runtime/node/24.18.0/bin/node
    npm: 11.16.0 - /Users/xxx/.vite-plus/js_runtime/node/24.18.0/bin/npm
    pnpm: 11.0.3 - /Users/xxx/.vite-plus/bin/pnpm
  Browsers:
    Chrome: 149.0.7827.201
    Chrome Canary: 152.0.7929.0
    Firefox: 136.0.4
    Safari: 17.4.1
```

### Additional context

_No response_

shulaoda:
### Description

Fixes #10136.

In a dev (FBM) session, editing a file so that one of its imports no longer resolves crashed the whole worker thread instead of reporting the error:

```
thread 'rolldown-worker' panicked at crates/rolldown_common/src/types/hybrid_index_vec.rs:76:28:
HybridIndexVec::Map missing idx 2 (len=1)
```

#### Root cause

When a rebuild fails with an unresolved import, `trace_import_chain_from_modules` enriches the error with the import chain (added in #7625). The walk follows `importers` edges, which cover the whole module graph. But in a partial scan, `intermediate_normal_modules.modules` is the sparse `HybridIndexVec::Map` and only holds the modules that were re-scanned. The first step from the changed module to its unchanged importer looks up an index that is not in the map, and `HybridIndexVec::get` panics on missing keys.

The existing error-recovery tests did not catch this because they use syntax errors. Only a `DiagnosableResolveError` triggers the chain tracing.

#### Fix

- Add a non-panicking `HybridIndexVec::try_get`. The panicking `get` is kept as is, so real indexing bugs still surface loudly.
- Add a non-panicking `ScanStageCache::snapshot()` accessor.
- In the chain walk, read modules that are missing from the sparse table out of the last successful build's snapshot. A slot that is present but empty (the module whose scan just failed) stays skipped, so the chain output matches full builds.
- Also stop the walk at entries recorded in the snapshot. The per-scan entry set is empty in a partial scan, so the old check never fired there.

With the fix, the rebuild reports a normal `UNRESOLVED_IMPORT` diagnostic with the full import chain, and the next edit recovers via a regular HMR patch.

#### Test

New fixture `topics/hmr/error_recovery/from_rebuild_unresolved_import`. Step 0 introduces an unresolvable import (this reproduced the panic before the fix), step 1 reverts it and verifies HMR recovers.

### DIFF
--- a/crates/rolldown/src/module_loader/module_loader.rs
+++ b/crates/rolldown/src/module_loader/module_loader.rs
@@ -986,14 +986,23 @@ impl<'a, Fs: FileSystem + Clone + 'static> ModuleLoader<'a, Fs> {
       }
       visited.insert(idx);
 
-      let module_opt = self.intermediate_normal_modules.modules.get(idx);
-      if let Some(module) = module_opt {
-        if let Some(normal) = module.as_normal() {
-          chain.push(normal.id.to_string());
-        }
+      // A partial scan only holds re-scanned modules, so read unchanged ones
+      // from the last successful build's snapshot. An empty slot stays skipped
+      // to keep the failed module out of its own chain.
+      let module_opt = match self.intermediate_normal_modules.modules.try_get(idx) {
+        Some(slot) => slot.as_ref(),
+        None => self.cache.snapshot().and_then(|snapshot| snapshot.module_table.modules.get(idx)),
+      };
+      if let Some(normal) = module_opt.and_then(Module::as_normal) {
+        chain.push(normal.id.to_string());
       }
 
-      if user_defined_entry_ids.contains(&idx) {
+      if user_defined_entry_ids.contains(&idx)
+        || self
+          .cache
+          .snapshot()
+          .is_some_and(|snapshot| snapshot.user_defined_entry_modules.contains(&idx))
+      {
         break;
       }
 
--- a/crates/rolldown/src/types/scan_stage_cache.rs
+++ b/crates/rolldown/src/types/scan_stage_cache.rs
@@ -79,6 +79,11 @@ impl ScanStageCache {
     self.snapshot.as_ref().unwrap()
   }
 
+  /// Non-panicking variant of [`Self::get_snapshot`].
+  pub fn snapshot(&self) -> Option<&NormalizedScanStageOutput> {
+    self.snapshot.as_ref()
+  }
+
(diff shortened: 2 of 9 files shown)


h-a-n-a (approved):

shulaoda:
### Merge activity

* **Jul 7, 2:20 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 7, 2:20 AM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 7, 2:26 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10164, opened 2026-07-07
feat(dev): client-side HMR

h-a-n-a:
## Summary

This stack of PRs focus on moving bundled-dev (full-bundle mode) HMR to client-side.

This PR adds the two client-side data structures (a **module factory map** next to the existing module cache, and a **module graph** shipped as a payload with `__rolldown_runtime__.registerGraph` in the bundle and patched by deltas), so boundary computation, bubbling, and disposal move from the server to the client runtime.

## Shape changes

- Initial/rebuild bundle: Added `__rolldown_runtime__.registerGraph` for storing module graph payload in runtime.
- HMR patch: 
    - Replaced `createEsmInitializer` / `createCjsInitializer` with `__rolldown_runtime__.registerFactory` that can be reused for HMR invalidation
    - Moved `applyUpdates` to the place where web-socket messages are received. It applies the change as soon as the new HMR patch has been executed. (Note: This happens in Vite's dev server  `FbmHMRClient`)

### Pros

- **Each browser tab maintains its own state.** (This works today too, but with a delay: the browser currently tells the server over WebSocket which modules have executed, and whether a module has executed affects HMR boundary computation.)
- **Smaller HMR payloads** — send only changed, not-yet-sent factories. Today every update sends the browser all modules on the path from the changed modules to the boundary; we want to send only the changed modules.
- **Remove the lazy-compilation dedup code**, and at the same time fix the oversized lazy chunk problem. Today a lazy chunk contains all modules, so multiple lazy modules ship duplicate modules.
### Building blocks of client-side HMR

- **Module Factory map** (register / update) — when a module updates, the module code along the whole update chain must re-execute.
- **Module Cache** (register / clear) — tells whether a module is registered, which affects HMR boundary computation.
- **Module Graph** — drives client-side bubbling.
- **Client-side HMR boundary computation, client bubble, module disposal** (clearing the module cache).

### Breaking down Webpack

- The Module Factory map and Module Cache exist natively.
- Module graph: Webpack gets it by rewriting inside `require`.
- The key part of HMR is knowing whether a module has executed, since that affects boundary computation. Dynamic import supports this natively — just check at runtime whether the module cache entry exists. Module execution errors also do not matter, because registering exports is the first step of a Webpack module's execution.

### Rolldown principles

- Keep the semantics of the Rolldown initial bundle and the full-rebuild bundle; use scope hoisting.
- The server does not compute the HMR boundary.
- Smaller patches, smaller lazy chunks.

### Rolldown design

- **Module Cache** already exists; we only need to add a **Module Factory map**.
- **Module graph:** does not exist today. Because the initial bundle is scope-hoisted, we need another way to deliver the module graph. We can insert a module graph payload into the initial bundle (or a rebuild bundle) that records the edges; during HMR the client then computes importers from it.
- **"Has this module executed" check:** once HMR moves to the client, this is natively supported, and there is no delay that makes the HMR boundary computation inaccurate.
- **Factory optimizations:**
  - **Incremental Module Factory** — the initial bundle carries the module graph and registers it directly as a payload. Later HMR updates send only module graph deltas, applied incrementally.
  - **The server is stateful** — it keeps a per-tab record (the ship map) of each browser tab's factory registration state and registered versions, and uses it to decide whether a factory must be re-sent to the browser.
  - **After a browser refresh**, the factories along the update chain must be re-sent during HMR.

## Testing

Vite commit in `packages/test-dev-server` is checked out to a dedicated commit solely for this refactor. `packages/vite-tests` is now based on the same commit as the one in `test-dev-server` in order to let `hmr` tests pass.


h-a-n-a:
* **#10293** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10293?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10271** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10271?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10181** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10181?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10223** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10223?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10207** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10207?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10208** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10208?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10164** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10164?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10164?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10164).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

h-a-n-a:
@codex review this

hyfdev:
## Architecture review: four blocking issues

Reviewed `49a06d5a8`. The client-side direction makes sense, but several core invariants do not hold.

### 1. The server-computed factory “superset” is not a superset

Minimal case:

```js
// child.js
export const value = 1;

// middle.js
import { value } from './child.js';

if (false) {
  import.meta.hot.accept('./child.js', () => {});
}

export { value };

// main.js
import { value } from './middle.js';
export { value };

import.meta.hot.accept((next) => {
  console.log(next.value);
});
```

The relevant graph and acceptance state are:

```text
main ──imports──> middle ──imports──> child
 │
 └─ live self-accept

middle:
  static scanner: accepts child
  browser runtime: does not accept child
```

On a `child.js` edit:

```text
Server static walk:
child -> sees middle's syntactic accept -> stop
Factories sent: { child }

Browser live walk:
child -> middle -> main
Update set: { child, middle, main }
```

The browser then fails because `middle` and `main` have no factories.

I reproduced this with a temporary Rust HMR fixture:

```text
Error: [test-hmr] full reload: no factory for middle.js
```

The problem is in `collect_client_update_superset`: `propagate_update` stops at statically scanned accept sites, while the browser walks accept callbacks that actually executed. A newly-added accept has the same problem because the server walks the new scan result while the browser must use the old generation's live state.

This invalidates the central invariant behind the one-push protocol.

### 2. A cold scope-hoisted module can execute stale code after its file changed

Minimal case:

```js
// child.js
module.exports = { value: 1 };

// main.js
function loadChild() {
  return require('./child.js');
}

import.meta.hot.accept();

process.on('beforeExit', () => {
  assert.strictEqual(loadChild().value, 2);
});
```

Change `child.js` to:

```js
module.exports = { value: 2 };
```

The generated state becomes:

```text
Initial bundle:
loadChild -> lexical require_child closure containing value: 1

HMR patch:
factories["child.js"] = factory containing value: 2

Later:
loadChild -> lexical require_child -> value: 1
```

The patch factory does not replace the initial scope-hoisted/CJS wrapper.

I reproduced this locally. Even after importing the new patch, the assertion fails:

```text
AssertionError: 1 !== 2
```

Therefore this assumption is incorrect:

```text
not executed at update time => safe to noop
```

The old initial wrapper can still execute later. Deferred execution paths must dispatch through mutable runtime state, or edits to cold wrapped modules must force a rebuild/reload.

### 3. Factory registration has no cross-channel version ordering

`registerFactory` is unconditional last-write-wins:

```js
registerFactory(id, kind, fn) {
  this.factories.set(id, { kind, fn });
}
```

Lazy chunks and HMR patches are loaded through independent HTTP imports, so this ordering is possible:

```text
t1  lazy payload L1 is generated with foo@v1, response is delayed
t2  foo is edited
t3  HMR payload H2 arrives and registers foo@v2
t4  delayed L1 arrives and overwrites the map with foo@v1
t5  next eviction executes foo@v1
```

A direct runtime probe confirms the behavior:

```text
register foo@v2
register foo@v1
init foo
=> v1
```

The follow-up server ship map does not solve this: it versions delivery bookkeeping, but `registerFactory` and `registerGraph` still carry no generation that lets the browser reject an older payload.

Factory and graph registrations need monotonic versions enforced by the runtime.

### 4. There are three different HMR transition models

Current ordering:

```text
Vite FBM:
compute boundaries on old state
-> import patch
-> evict/apply

Rust test runtime:
import patch, mutating graph/factories
-> compute boundaries
-> evict/apply

Default runtime:
import patch
-> stop
```

Relevant implementations:

- `packages/test-dev-server/vite/packages/vite/src/client/fbmHmrClient.ts`
- `crates/rolldown_testing/src/integration_test.rs`
- `crates/rolldown_plugin_hmr/src/runtime/runtime-extra-dev-default.js`

The tests therefore do not exercise the same state transition as Vite, and the default runtime has no apply path at all. This is already visible in the existing standalone-runtime review comments.

The walk/apply state machine should have one shared implementation. Vite should provide the hot-context/event adapter, rather than owning a separate HMR engine.

## Suggested protocol shape

Either:

```text
One push:
server ignores accept metadata
-> sends a versioned reverse-reachable factory closure
-> browser alone decides boundaries
```

or, preferably:

```text
1. Server announces changedIds + graph generation
2. Browser computes the exact update set from old live state
3. Browser requests missing factory ids + versions
4. Server returns versioned graph/factory data
5. Browser awaits dispose, evicts, re-executes, and applies callbacks
```

The current design mixes server-side static boundary truncation with client-side runtime boundary decisions. Those two sources of truth are what cause findings 1 and 2.

h-a-n-a:
1. The server-computed factory “superset” is not a superset
Nice finding. This is actually a case where we're not able to cover for now. Just the same as Vite.
2. A cold scope-hoisted module can execute stale code after its file changed
This should be regarded as a side-effect, which should be handled by `dispose()` API.
3. Factory registration has no cross-channel version ordering
Yes, this is not expected. I haven't though of a better way to solve this without bloating size. Given, it's pretty edge. I will create a new PR for this in the future.

packages/vite-tests/run.ts
```diff
@@ -4,6 +4,10 @@ import { styleText } from 'node:util';
 import { x } from 'tinyexec';
 
 const REPO_PATH = path.resolve(import.meta.dirname, './repo');
+const ROLLDOWN_REPO_ROOT = path.resolve(import.meta.dirname, '../..');
+// Single source of truth for the vite commit used by tests in this repo:
+// the pin of the `packages/test-dev-server/vite` submodule.
+const VITE_SUBMODULE_PATH = 'packages/test-dev-server/vite';
```
sapphi-red, line 10: We should add a CI to make sure the submodule is updated to the latest commit like https://github.com/rolldown/rolldown/blob/main/.github/workflows/update-test-dependencies.yml

packages/test-dev-server/tests/playground/hmr-whole-chain-dispose/foo.js
```diff
@@ -0,0 +1,24 @@
+import { bar } from './bar.js';
+
+document.querySelector('.chain').textContent = bar;
+
+// Record what this generation sees in its `hot.data` bag AT EXECUTION time.
+(window.__hotStates ??= []).push({
```
sapphi-red, line 6: If possible, I think it's easier to debug if it's written to the document rather than a global variable.

packages/test-dev-server/tests/playground/hmr-runtime-acceptance/__tests__/hmr-runtime-acceptance.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    editFile('dead.js', (code) => code.replace("'dead-v1'", "'dead-v2'"));
+
+    // The client's walk finds no executed boundary -> clean full page reload.
+    // The reloaded page runs a fresh bundle with the NEW content — never a
+    // silently stale page. (The reload may bounce once while the server
+    // regenerates the stale bundle, hence the generous poll.)
+    await expect
+      .poll(() => page.textContent('.dead'), { timeout: 15_000 })
+      .toBe('dead-v2');
+    await expect.poll(readMarker, { timeout: 15_000 }).toBe(null);
```
sapphi-red: I guess we can add `document.querySelctor(".something").textContent = 'rendered'`at the end of the browser script and wait for that content written before checking `expect(readMarker).toBe(null)`. That way we don't rely on the fixed timeout.

packages/test-dev-server/tests/playground/hmr-dynamic-import-self-accept/dev.config.mjs
```diff
@@ -0,0 +1,17 @@
...
+  platform: 'browser',
+  build: {
+    input: {
+      main: 'main.js',
+    },
+    platform: 'browser',
+    treeshake: false,
+    experimental: {
+      // Non-lazy: dynamic import() targets compile into the initial build (no lazy
+      // proxy). Lazy dynamic-import HMR is a separate follow-up.
```
sapphi-red, line 13: Does this mean the lazy bundling won't work after we merge this PR?

packages/test-dev-server/tests/playground/hmr-accept-syntax/dep-plain.js
```diff
@@ -0,0 +1,7 @@
+// Plain accept-dep: `import.meta.hot.accept('./dep', cb)`.
+import { value } from './dep-plain-target.js';
+document.querySelector('.dep-plain').textContent = value;
+
+import.meta.hot.accept('./dep-plain-target.js', (mod) => {
+  document.querySelector('.dep-plain').textContent = mod.value;
+});
```
sapphi-red, line 7: Not needed in this PR, but we should have a test case with

```javascript
import { value } from './dep-plain-target.js';
document.querySelector('.dep-plain').textContent = value;
​
import.meta.hot.accept('./dep-plain-target', (mod) => {
  document.querySelector('.dep-plain').textContent = mod.value;
});
```

(the `.js` is omitted for the `import.meta.hot.accept`)

packages/test-dev-server/tests/fixtures/load-dependencies/package.json
sapphi-red, line 1: It seems this one and `transform-dependencies` and `multiple-edits` aren't covered by other tests.

crates/rolldown/tests/rolldown/topics/hmr/dynamic_import_self_accept/artifacts.snap
sapphi-red, line 1: Probably nice to reduce the file size a bit later so that the parse/execution overhead on the browser side is less. For example, removing empty `finally` or dropping empty `edges` / `dynamicEdges`.

h-a-n-a: Yes, definitely! Will definitely optimize the module graph payload

packages/test-dev-server/tests/playground/hmr-dynamic-import-self-accept/dev.config.mjs
h-a-n-a: No. Lazy bundling still works after this PR. This is about hot-update bubbling across a lazy dynamic-import edge. This is never supposed even before this PR. I added here so that it can be used as a test case in the future.

packages/test-dev-server/tests/playground/hmr-runtime-acceptance/__tests__/hmr-runtime-acceptance.spec.ts
h-a-n-a: It does not seem necessary anymore. I removed the timeout

packages/test-dev-server/tests/fixtures/load-dependencies/package.json
h-a-n-a: Thanks! Restored. This was deleted by mistake.

packages/test-dev-server/tests/playground/hmr-runtime-acceptance/__tests__/hmr-runtime-acceptance.spec.ts
sapphi-red: Would you check the other places as well?

hyfdev:
## Second-round review at `f9ee9377a`

> This pass was done by Claude Fable 5; findings were verified against the code before posting.

Follow-up to my first-round review. Status there, as I read the current head: **#1** stands as an accepted limitation (should be spelled out in #10271 — and focus item 1 below is a second hole in the same invariant); **#2**/**#3** acknowledged but have no owning PR in the stack; **#4** unanswered and still visible in the tree (the test runtime walks the post-import graph at `crates/rolldown_testing/src/integration_test.rs:889`, the FBM client walks the pre-import graph, the default runtime has no walk at all).

---

## Review focus — needs a decision or a fix in this PR

### 1. A re-run factory can hit a newly-added static dep that is in the graph, never executed, and never shipped

The shipped superset covers the upward importer walk plus `new_added_modules` (graph-new modules only). A re-run executes *new* code, and new code can add *downward* static edges to modules that fail all three coverage conditions:

```js
// routes/heavy.js — in the graph, but only reachable through a never-triggered import()
// m.js — the edit adds:
import { heavy } from './routes/heavy.js';
import.meta.hot.accept();
```

Server walk from `m`: self-accepting → superset = `{m}`. `heavy` is not graph-new and not on the up-walk → factory not shipped. Client: the pre-check only covers the update set (`hmr-runtime.js:91-95`), so `m` is evicted, its factory re-runs and hits `initModule("routes/heavy.js")` (`impl_traverse_for_hmr_ast_finalizer.rs:45-58`) → not in the module cache, no factory → `MissingFactoryError` mid-apply, after `m` was already re-registered with empty exports.

With lazy compilation on, the single-tab case is mostly covered (the real module tends to be graph-new at that point), but the multi-tab variant survives: tab A triggers the lazy compile (the module enters the table), tab B never loads that chunk; editing `m` ships only `m` to both tabs, and tab B throws mid-apply.

Same class as first-round #1, different mechanism (down-edges of re-run code vs. acceptance divergence). If the verdict is "accepted limitation, falls back to reload", it belongs in #10271's invariant statement: *superset = upward static chain ∪ graph-new modules; runtime-acceptance divergence and new down-edges of re-run code are not covered; any escape falls back to reload.* A real fix could ship the changed modules' newly-added static deps that no payload has carried yet.

### 2. Dev mode doesn't validate the output format — non-ESM output silently loses the module graph

The graph prelude is rendered only in `format/esm.rs` (`render_chunk_content`); `cjs.rs`/`iife.rs`/`umd.rs` have no equivalent. Nothing in dev normalization (`prepare_build_context.rs:277-283`) or the JS dev API forces or validates `format: 'esm'`. A dev build with `format: 'cjs'` produces output with no `registerGraph` at all, so the client walk has no edges. A one-line hard validation error for non-ESM formats in dev mode closes it.

### 3. NoBoundary/circular now broadcasts FullReload to every tab — the per-client Noop is gone

`collect_client_update_superset` returns one verdict fanned out to all clients (`hmr_stage.rs:216-224`). Previously a client that had never executed the changed module got a Noop; now a tab that never loaded a lazy route reloads because someone edited an orphan module under it. The Codex P2 comment raised this and it's unanswered. I assume #10207 ("move full-reload to client side") owns this — please confirm it restores per-tab decisions (send changedIds, let the client decide reload vs. ignore). If it doesn't, this is a real behavior regression with no owner.

### 4. `connectClient` doesn't await `registerClient` — updates computed in the connect window are lost

`void this.#devEngine.registerClient(clientId)` (`full-bundle-dev-environment.ts:116`) is fire-and-forget; the Rust session appears a tick later. A change computed in that window snapshots clients without the new one (`bundling_task.rs:209-214`), so that tab gets neither a patch nor a reload, and its initial bundle may predate the edit. There's no WS reconnect in the default runtime to recover. Also raised by Codex on 07-14. Awaiting registration before acking the connection closes it.

### 5. The "Restored" fixtures push hasn't landed

The "Restored. This was deleted by mistake." reply — `load-dependencies`/`transform-dependencies`/`multiple-edits` are still absent at `f9ee9377a`. Either land that push, or decide the playground ports (`hmr-watch-file-load`/`hmr-watch-file-transform`/`hmr-multiple-edits`) supersede the node fixtures and close the thread that way.

---

## Follow-up material — fine after this PR, listed so it doesn't get lost

### 6. The default runtime keeps a dead accept/invalidate surface

On the non-Vite path, `hmr:update` only imports the patch (no walk/evict/re-run — `runtime-extra-dev-default.js:100-107`), accept callbacks are pushed but never read, and `invalidate()` still sends `hmr:invalidate` that the server deliberately drops (`dev-server.ts:202-204`; the Rust handler is deleted). `#handleHmrUpdates`'s `fromInvalidate` parameter has no caller passing `true` (`full-bundle-dev-environment.ts:166`). If "the FBM client is the only real consumer" is the decision, remove the dead surface (or at least the dead `invalidate()` send + `fromInvalidate`) and say so; otherwise the shared walk from first-round #4 is the fix.

### 7. `notifyPayloadDelivered(filename)` cannot attribute patches to clients as shaped — check at #10208

The stub's key is filename-only (`dev_engine.rs:283`), but patches are rendered once and broadcast to all clients under one filename (`hmr_stage.rs:229-241`), while lazy chunks are per-request. For #10208's ship map to work, either payload filenames become per-(client, payload) unique or the signal needs a client id. Relatedly, the node env writes its own `${seed}.js` per client (N identical files per batch, never cleaned up) and ignores the Rust-side `patch.filename` entirely (`full-bundle-dev-environment.ts:210`), which makes filename-keyed tracking unusable on that environment.

### 8. Two test assertions didn't survive anywhere in-repo — candidates for #10181

- **`watchChange` firing during a dev-server HMR edit** — the deleted `edit` fixture asserted it via an injected transform value; the only remaining `watchChange` coverage is standalone watch mode, a different code path from the dev engine.
- **`invalidate` multi-level propagation order** — the deleted `invalidation` fixture asserted exact ordering across two levels via a records array; `hmr-client-invalidate` covers the three basic outcomes only. `invalidate` also has zero crate-level coverage.

### 9. Nit: comment/code mismatch in `module_graph_delta.rs`

Lines 54-55 claim "HotAccept, new URL, css url … are skipped", but the filter only excludes `HotAccept`; `NewUrl` records to asset modules pass `Module::Normal` and genuinely become edges (parity with the server's `importers_idx`, so behavior is fine — the comment isn't).

### 10. The red Vite CI is stable pin drift, not flakiness

`tsconfig-json-load-error` fails because the pinned Vite commit's error-message regex predates main's tsconfig error changes; it will stay red until the pin moves. +1 to the submodule-freshness CI ask; the pin being an unmerged working commit also means merging here needs the Vite side coordinated.

### 11. Still unowned from the first round

Protocol versioning (first-round #3, acknowledged as "a new PR in the future") and the single shared walk implementation (first-round #4) have no owning PR in the stack — suggest tracking issues so they don't evaporate.

---

## Checked and not issues

For completeness, things I chased and ruled out: ghost importer rows after module deletion are symmetric with the server (`rebuild_importer_sets` only rebuilds a module's own records, on both sides), so both walks reload in lockstep — pre-existing, no divergence; static-dep init completeness holds for *existing* deps (execution order guarantees residency; focus item 1 is specifically about *added* deps); the graph-prelude placement in `esm.rs` is correct in all four paths (verified against snapshots, including the `sources: None` and grouped cases); multi-chunk row ownership is sound (`registerGraph` writes only `i < localCount`); seq stamping has no ordering bug (builds are serialized, and reload → new clientId → fresh session makes a reset unnecessary).

sapphi-red (approved):
no more comments from my side. I'll leave the implementation part to @hyfdev.

hyfdev (approved):
Well done.

packages/test-dev-server/tests/playground/hmr-whole-chain-dispose/foo.js
h-a-n-a: Will optimize in future PRs.

shulaoda (approved):
We don't need to solve everything in a single PR. Some of the smaller issues can be addressed in follow-up PRs on top of this stack.

PR #10166, opened 2026-07-07
fix: tree-shake unused circular declarators exported via export list

### LINKED ISSUE #10165: [Bug]: Tree-shaking doesn't work with variable declarations having multiple declarators with circular dependent functions exported with an export list

[REPL that demonstrates the difference in behavior between using an export list and a named export declaration](https://repl.rolldown.rs/#eNptU8lu6zAM/BVClySA4dxd9AHNA3rsqUdfZItO1MqiYdFtA8P/Xkpemi5AFmtIDTnkeFSNKkZlvcGP/CXEZ6+Kr3OmajniR0c9wwiPGTQwQdNTC6XKjy2ZwaEkluqu7Esfv8cjPBEj8EUzaNDeQAW6R7C+doNBIw8SRKCBu4EB3zCeaThfInxNuZ4YbBuropEmUBXcDzhlaqu4tvoFzL1K+eeLDXBGDvCQyp+gpr7Hmt0V0NnWei20eZ7HdhdtDhke4R72B7j/B+OUxRgIwQqd9ocFO23Yw/5wK1sYKxHEsbwhDH7HMRKpm9/UeoOqjbraMH1DvY2/yUBnkjTdgZSrsNZDkDk2ohJWHTakoktLaRTvNDgjv/0rMFGxhCBpvm0sFdwiv/tLoZ9NKllKT84Zevd5Tb6x55y37fwRmdc0L1c0GWysx/8pvFprt17biZBVviTqwaX/7cJ+jFOzXnxULH7JW2SdJyiLU4pOa6zDAAGdeED8V10TKrti8iAf1LV4T1eRTK5Qx5Z8KP10KH3U1+n6VZ+jyUhEzcq+YbOm1EypDHYo74+vLcp7UchghSkSCdObpDlxX2A1fQIY4x3s)

### What is expected?

It is expected that circular dependent functions declared in the same variable declaration are correctly eliminated if they are unused, regardless of whether they were exported using an export list or a usual named export declaration.

Rolldown does not eliminate functions exported with an export list.

### System Info

```Shell
System:
    OS: Windows 10 10.0.19045
    CPU: (16) x64 AMD Ryzen 7 5800X 8-Core Processor             
    Memory: 41.89 GB / 63.91 GB
  Binaries:
    Node: 24.5.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 11.5.1 - C:\Program Files\nodejs\npm.CMD
    pnpm: 11.5.3 - C:\Program Files\nodejs\pnpm.CMD
    Deno: 2.5.4 - C:\Users\user\.deno\bin\deno.EXE
  Browsers:
    Chrome: 149.0.7827.201
    Edge: Chromium (150.0.4078.48)
    Firefox: 152.0 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.19041.5794
  npmPackages:
    rolldown: ^1.1.4 => 1.1.4
```

### Any additional comments?

This is an edge case that was not covered in a PR that introduced multi-declarator tree-shaking (#1564).

IWANABETHATGUY:
### Description

Fixes #10165.

`let f = () => {}, a = () => b(), b = () => a();` followed by `export { f, a, b }` kept the unused `a`/`b` pair in the output, while the equivalent inline form `export let F = ..., A = ..., B = ...` tree-shook them.

Tree-shaking includes whole top-level statements. Before scanning, only exported multi-declarator declarations (and, under `keepNames`, all of them) were split into one statement per declarator, so demanding one binding of a bare declaration kept every declarator in it. The default dce-only minify pass normally cleans up such leftovers, but it counts references and cannot delete two functions that reference each other in a cycle — so the circular pair survived.

The fix splits bare top-level multi-declarator declarations in the pre-scan the same way exported ones already are (`split_multi_declarator` in `tweak_ast_for_scanning.rs`), giving inclusion analysis per-declarator granularity. This matches esbuild, which creates one part per declarator.

Two details:

- The split (both the existing export path and the new bare path) is now limited to declarations where **every** declarator binds a plain identifier, because destructuring performs binding-time work (iterator calls, property reads). Previously the export path split when *any* declarator was a plain identifier. The regression fixture covers this with an iterator side effect that must survive when only a sibling plain binding is demanded.
- The first replacement statement keeps the original statement span, so comments attached to the statement's start position stay attached.

Snapshot changes are cosmetic: joined declarations now print as separate statements (all their declarators are used), and two fixtures with mixed identifier/destructuring declarations print grouped again.

### Alternatives explored

- Gating the split on "a binding appears in an export list": covers the issue as filed but misses the same bug when the demand comes from inside the module (nothing exported at all), so the split is unconditional at the top level.
- Relying on the minifier: reference-counting DCE fundamentally cannot remove closed reference cycles; only demand-driven inclusion can, and it needs per-declarator statements to do so.

### Test plan

- New fixture `crates/rolldown/tests/rolldown/issues/10165` mirrors the issue (fails without the fix: `a`/`b` stay in the output) and asserts at runtime that a destructuring side effect next to an exported plain binding is preserved.
- `cargo test --workspace`: green (1797 integration tests), 9 cosmetic snapshot updates.
- `just test-node-rolldown`: green.


shulaoda (approved):

hyfdev (approved):

hyfdev:
Interesting. Rollup failures. I guess it's not about correctness but more of a behavior change.

IWANABETHATGUY:
> Interesting. Rollup failures. I guess it's not about correctness but more of a behavior change.

Yeah, so I ignore the snapshot diff.

Changing `var a, b, c` to `var a; var b; var c` does not affect correctness, and the minifier will collapse them back.

IWANABETHATGUY:
### Merge activity

* **Jul 7, 7:54 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 7, 7:54 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 7, 7:59 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10168, opened 2026-07-07
refactor: separate tree-shaking side effects from execution order sensitivity

hyfdev:
### What

This PR splits the statement eval analyzer result into two separate answers:

- tree-shaking side effects, stored in the public `StmtEvalFlags`
- execution-order sensitivity, represented by analyzer-local order-sensitive reasons

`StmtInfo::side_effects()` continues to carry only the tree-shaking flags. `EcmaViewMeta::ExecutionOrderSensitive` now comes from `StmtEvalFacts::is_order_sensitive()`, which combines unknown runtime side effects with order-sensitive reasons such as global reads and pure annotations.

The cross-module optimization pass still re-runs the analyzer, but it consumes only `tree_shaking_flags()` because that pass mutates removal-related metadata, not module execution-order metadata.

### Why

Rolldown needs to answer two different questions:

- Can this statement or module be removed without observable behavior?
- Does this statement or module need to keep its relative execution timing?

Keeping both answers in the same public bitset makes later order-sensitive facts look like tree-shaking side effects. That can affect chunk dependency edges or removal decisions for the wrong reason.

This PR is intended as a refactor that keeps existing behavior while making those two answers explicit.

### Tests

- `cargo test -p rolldown stmt_eval_analyzer --lib`
- `cargo check -p rolldown`

PR #10169, opened 2026-07-07
fix: mark import binding reads as execution-order sensitive

hyfdev:
Stacked on #10168.

### What

This PR adds an analyzer-local `ImportBindingRead` order-sensitive reason.

It marks top-level reads of imported values as execution-order sensitive without adding them to the public tree-shaking `StmtEvalFlags`:

- direct imported binding reads, such as `foo`
- member chains rooted at an imported binding, such as `obj.value`
- imported bindings used in computed member keys, such as `local[key]`
- namespace import member reads, such as `ns.foo`

It intentionally does not mark bare namespace object reads, and it does not walk into function bodies for this top-level order analysis. Computed member keys only propagate `ImportBindingRead`; this does not bring back general global or pure-annotation propagation for member expressions.

### Why

An imported binding read is not a tree-shaking side effect by itself. Removing or retaining the statement should still be decided by the tree-shaking side-effect analysis.

It is execution-order sensitive, though, because the observed value can depend on when the importing module runs relative to the exporting module. The generated chunk order needs that information.

### Tests

- `cargo test -p rolldown stmt_eval_analyzer --lib`
- `cargo check -p rolldown`

### DIFF
--- a/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs
+++ b/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs
@@ -6,7 +6,7 @@ use oxc::ast::ast::{
   IdentifierReference, UnaryOperator, VariableDeclarationKind,
 };
 use oxc::ast::match_member_expression;
-use oxc::semantic::{NodeId, SymbolId};
+use oxc::semantic::{NodeId, SymbolFlags, SymbolId};
 use oxc_ecmascript::GlobalContext;
 use oxc_ecmascript::side_effects::{
   MayHaveSideEffects, MayHaveSideEffectsContext, PropertyReadSideEffects,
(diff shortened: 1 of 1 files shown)


hyfdev:
* **#10169** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10169?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10169?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10168** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10168?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#10180](https://github.com/rolldown/rolldown/pull/10180) <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10180?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10169).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyfdev:
wrong direction, it's not sound.

PR #10185, opened 2026-07-08
fix: improve invalid annotation warnings

### LINKED ISSUE #10183: Warnings for invalid pure annotations in node_modules are too noisy and look like errors
### Reproduction

Building [Codiff](https://github.com/nkzw-tech/codiff) with `vp build` emits warnings like:

```text
[INVALID_ANNOTATION] A comment "/*#__PURE__*/" in "node_modules/@lexical/react/dist/LexicalErrorBoundary.prod.mjs" contains an annotation that Rolldown cannot interpret due to the position of the comment.
 - comment ignored due to position in node_modules/@lexical/react/dist/LexicalErrorBoundary.prod.mjs at 581..594
  │
  │ Help: For more information on how to use pure annotations correctly, check the documentation: https://rolldown.rs/in-depth/dead-code-elimination#pure
```

I hit the same warning for multiple `@lexical/react` files under `node_modules`.

### Problems

1. `INVALID_ANNOTATION` is rendered in red even though this is only a warning, which makes the build look broken.
2. The warning points at a transitive dependency in `node_modules`, so the app author usually can't do anything about it directly.
3. Because the warning shows up on every build, it creates noise for humans and coding agents that keep re-flagging it as if it were a real failure.

### Expected

A few possible improvements:

- render this warning in a warning color instead of an error color
- reduce or suppress this class of warning for `node_modules` / transitive dependencies
- or otherwise make it clearer that Rolldown is safely ignoring the misplaced annotation and the build is still fine

### Notes

The current behavior is technically correct, but the UX is pretty rough for downstream consumers when the source is third-party code.

hyfdev:
## Summary

- render `INVALID_ANNOTATION` diagnostics with warning severity so they use warning colors
- warn by default only for real local project files inside `cwd` and outside `node_modules`
- keep dependencies, outside-`cwd` files, and virtual IDs quiet by default so builds do not report warnings users cannot act on
- keep `checks.invalidAnnotation` as a normal boolean: omitted/`true` use the local project scope, while `false` disables the check
- normalize parent-directory components and match `node_modules` case-insensitively to avoid false-positive local classification
- show concise help for correct annotation placement and disabling the check
- document the conservative default and add regression coverage

## Test plan

- `cargo test -p rolldown --lib`
- `cargo test -p rolldown_error`
- `NEEDS_EXTENDED=false cargo test -p rolldown --test integration invalid_pure_annotation -- --ignored --nocapture`
- `cargo test -p rolldown --test integration 10183 -- --nocapture`
- `just t-run crates/rolldown/tests/esbuild/dce/remove_unused_pure_comment_calls/_config.json`
- `cargo clippy -p rolldown -p rolldown_error -p generator --all-targets -- --deny warnings`
- `just build-rolldown`
- `just lint-node`
- `just lint-repo`
- end-to-end Node API check for the conservative default, boolean enable/disable behavior, yellow output, and concise help

Fixes #10183

### DIFF
--- a/crates/rolldown/src/utils/parse_to_ecma_ast.rs
+++ b/crates/rolldown/src/utils/parse_to_ecma_ast.rs
@@ -4,14 +4,15 @@ use json_escape_simd::escape;
 use oxc::{semantic::Scoping, span::SourceType as OxcSourceType};
 use oxc_str::CompactStr;
 use rolldown_common::{
-  ConstExportMeta, ModuleDefFormat, ModuleType, NormalizedBundlerOptions, RUNTIME_MODULE_KEY,
-  StrOrBytes, json_value_to_ecma_ast,
+  ConstExportMeta, ModuleDefFormat, ModuleId, ModuleType, NormalizedBundlerOptions,
+  RUNTIME_MODULE_KEY, StrOrBytes, json_value_to_ecma_ast,
 };
 use rolldown_ecmascript::{EcmaAst, EcmaCompiler};
-use rolldown_error::{BuildDiagnostic, BuildResult};
+use rolldown_error::{BuildDiagnostic, BuildResult, EventKindSwitcher};
 use rolldown_plugin::HookTransformAstArgs;
 use rolldown_utils::mime::guess_mime;
 use rustc_hash::FxHashMap;
+use sugar_path::SugarPath as _;
 
 use super::pre_process_ecma_ast::PreProcessEcmaAst;
 
@@ -113,15 +114,28 @@ pub async fn parse_to_ecma_ast(
     })
     .await?;
 
+  let is_local_project_file = is_local_project_file(&resolved_id.id, &options.normalized_cwd);
+  let should_warn_on_invalid_annotation =
+    options.checks.contains(EventKindSwitcher::InvalidAnnotation) && is_local_project_file;
+
   PreProcessEcmaAst::default().build(
     ecma_ast,
     stable_id,
     resolved_id.id.as_str(),
+    should_warn_on_invalid_annotation,
     &parsed_type,
     replace_global_define_config.as_ref(),
     options,
     has_lazy_export,
   )
+}
+
+fn is_local_project_file(id: &ModuleId, normalized_cwd: &Path) -> bool {
+  if id.is_in_node_modules() {
+    return false;
+  }
+
+  id.as_path().is_some_and(|path| path.normalize().starts_with(normalized_cwd))
 }
 
 fn pre_process_source(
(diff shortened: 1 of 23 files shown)


TheAlexLichter:
> treat explicit checks.invalidAnnotation: true as opting into all modules, while false disables the check

This is the only thing that feels a bit too implicit to me. If we have 3 modes (all, only "project-related", none) then we should consider expressing it that way (while keeping true/false as legacy options).

hyfdev:
- Removed the implicit third mode instead of expanding the API for now.
- Unset / `true`: local project files only.
- `false`: disabled.
- An explicit all-modules mode can be considered separately if needed later.

hyfdev:
### Merge activity

* **Jul 8, 7:13 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 8, 8:32 AM UTC**: `hyf0` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 8, 8:37 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

crates/rolldown/src/utils/parse_to_ecma_ast.rs
```diff
@@ -113,17 +114,33 @@ pub async fn parse_to_ecma_ast(
...
   )
 }
 
+fn is_local_project_file(id: &ModuleId, cwd: &Path) -> bool {
+  id.as_path().is_some_and(|path| {
+    let path = path.normalize();
+    let cwd = cwd.normalize();
+    path.starts_with(cwd.as_ref())
+      && !path.components().any(|component| {
+        component.as_os_str().to_string_lossy().eq_ignore_ascii_case("node_modules")
```
IWANABETHATGUY: We already have `PathExt::is_in_node_modules` (via `ModuleId::is_in_node_modules`) that every other node_modules check in the codebase goes through — but that one matches case-sensitively (`comp == "node_modules"`), while this new check is case-insensitive. So "is this path in node_modules?" can now answer differently depending on which one you call, which is the kind of split that's easy to trip over later.

Could we reuse the existing helper here (`!path.is_in_node_modules()`), or — if the case-insensitive match is intentional for macOS/Windows — push it down into `PathExt::is_in_node_modules` so every call site stays consistent?

crates/rolldown/src/utils/parse_to_ecma_ast.rs
```diff
@@ -113,17 +114,33 @@ pub async fn parse_to_ecma_ast(
...
     replace_global_define_config.as_ref(),
     options,
     has_lazy_export,
   )
 }
 
+fn is_local_project_file(id: &ModuleId, cwd: &Path) -> bool {
+  id.as_path().is_some_and(|path| {
+    let path = path.normalize();
+    let cwd = cwd.normalize();
```
IWANABETHATGUY: `cwd` is constant for the whole build, but we re-normalize it here on every module we parse, so this ends up doing a throwaway `PathBuf` allocation per module for a value that never changes. Could we normalize `cwd` once in the caller (or cache it) and pass the normalized path in? Then it's O(1) per build instead of O(modules).

IWANABETHATGUY (approved):

crates/rolldown/src/utils/parse_to_ecma_ast.rs
hyfdev: Fixed in 689d58a63: `is_local_project_file` now reuses `ModuleId::is_in_node_modules()`, so this path follows the same `node_modules` semantics as the rest of the codebase instead of keeping a local case-insensitive check.

crates/rolldown/src/utils/parse_to_ecma_ast.rs
hyfdev: Fixed in 689d58a63: `prepare_build_context` now computes `normalized_cwd` once on `NormalizedBundlerOptions`, and `parse_to_ecma_ast` passes that cached path into `is_local_project_file`, so we no longer allocate a normalized cwd per parsed module.

PR #10191, opened 2026-07-08
fix: keep deduplicated asset filenames stable once they can be observed

shulaoda:
### Problem

When two emitted assets have identical content but different names, rolldown deduplicates them into a single file. How that file's name gets chosen caused two bugs:

- **vitejs/vite#22856**: rolldown could rename the surviving asset to a shorter name *after* Vite had already read and cached the old name via `getFileName`. Vite's `manifest.json` (and ssr-manifest, preloads, injected `<link>`) then pointed at a CSS file that was never written.
- **rolldown#9940**: when the survivor name depended on which duplicate emitted first, the output was non-deterministic (assets emit concurrently), so identical input produced different filenames between builds.

### Fix

Choose the survivor name based on the phase, matching Rollup:

- **Build phase** (emitted while loading modules, e.g. `import x from './a.png'`): keep the shortest name. This is order-independent, so builds stay deterministic (#9940).
- **Output phase** (emitted from `renderChunk` or `generateBundle`, e.g. Vite's CSS): keep the first emitted name and never rename it, so a name already returned by `getFileName` cannot go stale (#22856).

An asset already written to an earlier incremental build is also never renamed.

### Related

- Fixes vitejs/vite#22856
- Keeps rolldown#9940 fixed

### DIFF
--- a/crates/rolldown/src/bundle/bundle.rs
+++ b/crates/rolldown/src/bundle/bundle.rs
@@ -92,6 +92,19 @@ impl<Fs: FileSystem + Clone + 'static> Bundle<Fs> {
     &mut self,
     scan_mode: ScanMode<ArcStr>,
   ) -> BuildResult<NormalizedScanStageOutput> {
+    // The whole scan (buildStart through buildEnd) is the build phase. Resetting on every exit
+    // means a failed scan cannot leave the emitter stuck in the build phase, where a later emit
+    // (e.g. an HMR patch after a failed rebuild) could still change an already-emitted filename.
+    self.plugin_driver.file_emitter.enter_build_phase();
+    let result = self.scan_modules_impl(scan_mode).await;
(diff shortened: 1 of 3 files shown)


sapphi-red (approved):

sapphi-red:
### Merge activity

* **Jul 8, 8:12 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 8, 8:12 AM UTC**: `sapphi-red` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 8, 8:17 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

Issue #9598, opened 2026-05-29
[Feature Request]: Rebuild when tsconfig.json changes


sapphi-red:
### What problem does this feature solve?

Changing `tsconfig.json` does not trigger the rebuild even if the tsconfig is affecting the build output. I'd be better if the rebuild is triggered in that case.

Reproduction: https://stackblitz.com/edit/rolldown-rolldown-starter-stackblitz-sevqyycn?file=tsconfig.json,src%2Fentry.ts,rolldown.config.mjs (change `"useDefineForClassFields": false` to `"useDefineForClassFields": true`)

related: https://github.com/oxc-project/oxc-resolver/issues/1011

### What does the proposed API look like?

Not an API change

Boshen:
@h-a-n-a please evaluate whether we need to rearchitect with the `resolve_with_context` API.

sapphi-red:
Reopening as I guess the `extends` and `references` is not handled yet

shulaoda:
> Reopening as I guess the `extends` and `references` is not handled yet

I'm currently working on the `extends` issue. At first I thought I'd already solved the `references` part, but it looks like I also need to handle cases where changes to the `include` patterns in a previously referenced `tsconfig` cause files to transition from **matched → missed** or **missed → matched**.

shulaoda:
This currently requires support from `oxc-resolver`, so I'm going to pause this work for now and focus on higher-priority tasks. Once the necessary support lands in `oxc-resolver`, the rest of the implementation should be fairly straightforward.

PR #10205, opened 2026-07-09
refactor: share the main resolver's cache with the transformer's tsconfig lookups

shulaoda:
Groundwork for #9598 (rebuild when tsconfig changes), on top of #10200.

`RawTransformOptions` used to build its own private oxc `Resolver` just for per-file tsconfig discovery. That means two resolvers with two caches: the same tsconfig file is read and parsed twice, and clearing the main resolver's cache (as the watcher already does before every rebuild) never touches the transformer's copy, so the two sides can end up seeing different tsconfig contents. For #9598 we need one clear to invalidate both.

`RawTransformOptions` now receives an `Arc<ResolverGeneric<OsFileSystem>>` derived from the main resolver via the new `Resolver::derive_cache_sharing_resolver()`. oxc's `clone_with_options` shares the cache `Arc` when the `yarn_pnp` option matches, and since we clone with the resolver's own options it always matches. Both sides now read the same tsconfig/fs/package.json cache, and one `clear_cache` on the main resolver invalidates both. The `TsConfig` and `yarn_pnp` parameters of `RawTransformOptions::new` are gone because the tsconfig discovery config comes with the shared resolver, which was built from the same `TsConfig` value.

Notes for reviewers: `rolldown_common` gains a `rolldown_fs` dependency, which is cycle-free since `rolldown_fs` only depends on `oxc_resolver` and `vfs`. The two Raw-mode arms in `prepare_build_context` became identical after this change and were merged into one. The derived resolver inherits the main resolver's full options instead of oxc defaults, but `find_tsconfig` only depends on the `tsconfig` option, which was already configured identically on both sides, so per-file discovery results are unchanged. The transformer also now reads files through the same `OsFileSystem` instance as module resolution, so the yarn pnp setting is applied consistently.

Part of #9598.

### DIFF
--- a/crates/rolldown/src/utils/prepare_build_context.rs
+++ b/crates/rolldown/src/utils/prepare_build_context.rs
@@ -358,32 +358,22 @@ pub fn prepare_build_context(
     }
 
     // Create TransformOptions based on tsconfig mode:
-    // - Auto(true)/Manual: Raw mode, resolves tsconfig per file so rebuilds
-    //   can re-read edited tsconfig files
+    // - Manual/Auto(true): Raw mode, resolves the tsconfig per file
     // - Auto(false): Normal mode without tsconfig
     match tsconfig {
-      TsConfig::Manual(path) => Box::new(TransformOptions::new_raw(
-        RawTransformOptions::new(raw_transform_options, TsConfig::Manual(path), yarn_pnp),
+      TsConfig::Manual(_) | TsConfig::Auto(true) => Box::new(TransformOptions::new_raw(
+        RawTransformOptions::new(
+          raw_transform_options,
+          Arc::new(resolver.clone_default_resolver()),
+        ),
         target,
         jsx_preset,
       )),
-      TsConfig::Auto(is_auto) => {
-        Box::new(if is_auto {
-          // Auto mode: Create Raw mode TransformOptions
-          // Each file will find its nearest tsconfig during compilation
-          TransformOptions::new_raw(
-            RawTransformOptions::new(raw_transform_options, TsConfig::Auto(true), yarn_pnp),
-            target,
-            jsx_preset,
-          )
-        } else {
-          TransformOptions::new(
-            merge_transform_options_with_tsconfig(raw_transform_options, None, &mut warnings)?,
-            target,
-            jsx_preset,
-          )
-        })
-      }
+      TsConfig::Auto(false) => Box::new(TransformOptions::new(
+        merge_transform_options_with_tsconfig(raw_transform_options, None, &mut warnings)?,
+        target,
+        jsx_preset,
+      )),
     }
   };
 
--- a/crates/rolldown_resolver/src/resolver.rs
+++ b/crates/rolldown_resolver/src/resolver.rs
@@ -66,6 +66,12 @@ impl<Fs: FileSystem + Clone + 'static> Resolver<Fs> {
       package_json_cache: DashMap::default(),
     }
   }
+
+  /// The clone shares this resolver's cache, so tsconfig lookups and cache
+  /// clearing stay consistent with module resolution.
+  pub fn clone_default_resolver(&self) -> ResolverGeneric<Fs> {
+    self.default_resolver.clone_with_options(self.default_resolver.options().clone())
+  }
 }
 
 #[derive(Debug)]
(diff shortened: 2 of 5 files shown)


shulaoda:
* **#10262** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10262?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10261** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10261?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10258** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10258?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10257** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10257?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10205** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10205?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10205?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10200** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10200?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10205).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

h-a-n-a:
@codex review

crates/rolldown_resolver/src/resolver.rs
```diff
@@ -66,6 +66,12 @@ impl<Fs: FileSystem + Clone + 'static> Resolver<Fs> {
       package_json_cache: DashMap::default(),
     }
   }
+
+  /// Derive a resolver that shares this resolver's cache, so tsconfig
+  /// lookups and cache clearing stay consistent with module resolution.
+  pub fn derive_cache_sharing_resolver(&self) -> ResolverGeneric<Fs> {
```
hyfdev: Did AI come up with this function name?  Just call it `clone_default_resolver`.

crates/rolldown/src/utils/prepare_build_context.rs
```diff
@@ -358,32 +358,22 @@ pub fn prepare_build_context(
...
-    //   can re-read edited tsconfig files
+    // - Manual/Auto(true): Raw mode, resolves the tsconfig per file
     // - Auto(false): Normal mode without tsconfig
     match tsconfig {
-      TsConfig::Manual(path) => Box::new(TransformOptions::new_raw(
-        RawTransformOptions::new(raw_transform_options, TsConfig::Manual(path), yarn_pnp),
+      TsConfig::Manual(_) | TsConfig::Auto(true) => Box::new(TransformOptions::new_raw(
+        RawTransformOptions::new(
+          raw_transform_options,
+          Arc::new(resolver.derive_cache_sharing_resolver()),
```
hyfdev: Would this cause unexpected situation? How we expect users to configurate resolver that affects the tsconfig lookup?

What's the real-world use case for this PR's change?

shulaoda: > Would this cause unexpected situation? How we expect users to configurate resolver that affects the tsconfig lookup?

No, this is expected. Our original plan was to have the general resolver and the tsconfig resolver share the same global resolver, but we couldn't achieve that due to some implementation constraints.

> What's the real-world use case for this PR's change?

First, they already share the same tsconfig cache. In addition, when a tsconfig changes, we can invalidate the resolver's tsconfig cache at the same time to ensure the subsequent rebuild uses the updated configuration correctly.

h-a-n-a (approved):

shulaoda:
### Merge activity

* **Jul 14, 6:10 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 14, 6:10 AM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 14, 6:15 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10207, opened 2026-07-09
refactor(dev): move full-reload to client side

h-a-n-a:
## Summary

For full design and principles, please refer to https://github.com/rolldown/rolldown/pull/10164#issue-4824909361.

This PR removes the full-reload decision from the server's HMR update type, so the client decides to reload from its own boundary result.

PR #10181, opened 2026-07-07
test(dev): add hot API test cases

h-a-n-a:
## Summary

For full design and principles, please refer to https://github.com/rolldown/rolldown/pull/10164#issue-4824909361.

## The playgrounds

Each playground is a small app (`index.html` + JS modules + `dev.config.mjs` + `package.json`) with a `__tests__/*.spec.ts` that edits a file and asserts on the running page. Several port behavior expectations from Vite's HMR tests (the `hmr-hot-off` fixture says so explicitly).

| Group | Playgrounds | What they exercise |
|---|---|---|
| Accept | `hmr-accept-exports`, `hmr-nested-dep-accept` | `hot.accept` receiving new exports; accepting a dep deeper in the graph |
| Circular graphs | `hmr-circular-accept-outside`, `hmr-circular-self-accept`, `hmr-invalidate-circular` | accept / self-accept / `hot.invalidate` when the edited module sits in an import cycle |
| Lifecycle / state | `hmr-dispose-data`, `hmr-prune` | `hot.dispose` passing state through `hot.data`; `hot.prune` when a module is no longer imported |
| Events | `hmr-hot-events`, `hmr-hot-off` | `hot.on` for built-in `vite:beforeUpdate` / `vite:afterUpdate`; `hot.off` removing one listener while another stays |
| Dynamic import + lazy compilation | `hmr-lazy-dynamic-import-accept-dep`, `hmr-lazy-dynamic-import-self-accept`, `hmr-not-loaded-dynamic-import` | updates to dynamically imported modules, including under lazy compilation and when the module was never loaded |
| Full reload | `hmr-before-full-reload` | behavior around an update that must fall back to a full page reload |

PR #10208, opened 2026-07-09
feat(dev): per-client ship map for HMR patch sizing

h-a-n-a:
## Summary

For full design and principles, please refer to https://github.com/rolldown/rolldown/pull/10164#issue-4824909361.

This PR adds the server-side **ship map**: a per-client record of which factory versions each browser tab already has, so patches and lazy chunks carry only changed, not-yet-shipped factories.

## Example

The stateful server's ship map would incrementally ship factories that are only outdated or never shipped.

### The module graph

```mermaid
flowchart LR
    app["app.js"] --> foo["foo.js<br/>(import.meta.hot.accept — boundary)"]
    foo --> bar["bar.js"]
    bar --> baz["baz.js<br/>(edited twice)"]
```

### Server & client interaction

```mermaid
sequenceDiagram
    participant S as Server (keeps shipped[C])
    participant C as Browser tab C

    Note over C: initial bundle is scope-hoisted — no factories registered on the client yet
    Note over S: shipped[C] = { }

    Note over S,C: ① first edit of baz.js
    S->>C: patch 1 — factories foo@v0 + bar@v0 + baz@v1 (whole re-run chain, nothing shipped yet)
    C->>C: re-run baz → bar → foo, foo accepts
    Note over S: shipped[C] = { foo@v0, bar@v0, baz@v1 }

    Note over S,C: ② second edit of baz.js
    S->>C: patch 2 — factory baz@v2 only (foo@v0 and bar@v0 are still current in shipped[C])
    C->>C: re-run baz → bar → foo, foo accepts
    Note over S: shipped[C] = { foo@v0, bar@v0, baz@v2 }
```

The first edit ships the whole re-run chain because the initial bundle is scope-hoisted, so the client holds no factories yet. The second edit ships only `baz` — the ship map says `foo` and `bar` were already delivered and are not stale.

Rendered, the flow reads:

app.js ──> foo.js ──> bar.js ──> baz.js
           (accept)              (edited)

edit ①  patch = [foo@v0, bar@v0, baz@v1]   shipped[C]: {} -> {foo@v0, bar@v0, baz@v1}
edit ②  patch = [baz@v2]                   shipped[C]: baz v1 -> v2, rest untouched

### DIFF
--- a/crates/rolldown/src/hmr/hmr_stage.rs
+++ b/crates/rolldown/src/hmr/hmr_stage.rs
@@ -9,8 +9,8 @@ use std::{
 use arcstr::ArcStr;
 use oxc_traverse::traverse_mut;
 use rolldown_common::{
-  ClientHmrInput, ClientHmrUpdate, HmrLazyChunkOutput, HmrPatch, HmrUpdate, ImportKind, Module,
-  ModuleIdx, ModuleTable, ScanMode, WatcherChangeKind,
+  ClientHmrInput, ClientHmrUpdate, HmrLazyChunkOutput, HmrPatch, HmrStampTable, HmrUpdate,
+  ImportKind, Module, ModuleIdx, ModuleTable, ScanMode, WatcherChangeKind,
 };
 use rolldown_ecmascript::{EcmaAst, EcmaCompiler, PrintCommentsOptions, PrintOptions};
 use rolldown_ecmascript_utils::AstFactory;
@@ -80,6 +80,7 @@ impl<'a, Fs: FileSystem + Clone + 'static> HmrStage<'a, Fs> {
     &mut self,
     changed_file_paths: &FxIndexMap<String, WatcherChangeKind>,
     clients: &[ClientHmrInput<'_>],
+    stamp_table: &mut HmrStampTable,
   ) -> BuildResult<Vec<ClientHmrUpdate>> {
     tracing::trace!(
       "[HmrStage] starts computing HMR updates\n - changed_file_paths: {:#?}\n - clients: {:#?}",
@@ -198,11 +199,20 @@ impl<'a, Fs: FileSystem + Clone + 'static> HmrStage<'a, Fs> {
       new_added_modules
     };
 
-    // 2. Collect the factories to ship. The client's walk may remove modules from its cache and re-run
+    // 2. Stamp the rebuild: `latest[m] = rebuild_seq` for every changed or newly added
+    // module — the versioned ship map's staleness source.
+    let rebuild_seq = stamp_table.begin_rebuild();
+    for module_idx in changed_modules.iter().chain(new_added_modules.iter()) {
+      if let Module::Normal(module) = &self.module_table().modules[*module_idx] {
+        stamp_table.stamp(module.stable_id.as_arc_str(), rebuild_seq);
+      }
+    }
+
+    // 3. Collect the factories to ship. The client's walk may remove modules from its cache and re-run
     // anything up the changed ids' importer chains, and a re-run without a resident
     // factory forces a reload — so the server must ship a SUPERSET of any client's
     // possible update set, over-approximated on static truth (it never sees runtime
-    // acceptance).
+    // acceptance). The ship map below subtracts what each tab already holds.
     let changed_ids = changed_modules
       .iter()
       .filter_map(|module_idx| {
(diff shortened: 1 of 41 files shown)


h-a-n-a:
* **#10293** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10293?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10271** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10271?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10181** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10181?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10223** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10223?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10207** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10207?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10208** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10208?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10208?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10164** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10164?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10208).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

sapphi-red (approved):
(I only checked the behavior explanation comment in the test cases)

shulaoda (approved):

crates/rolldown/src/hmr/hmr_stage.rs
```diff
@@ -802,6 +873,19 @@ impl<Fs: FileSystem + Clone + 'static> HmrStage<'_, Fs> {
...
+          // visited test spares the ship-map string hashing for all but the first.
+          if result.contains(&dep_idx) {
+            continue;
+          }
+          if let Module::Normal(normal_dep) = &modules[dep_idx] {
+            // Skip deps whose current copy this client already holds per the ship map.
+            let stable_id = normal_dep.stable_id.as_str();
+            if shipped.get(stable_id).is_some_and(|stamp| !stamp_table.is_stale(stable_id, *stamp))
+            {
+              continue;
```
hyfdev, line 886: Important: please keep traversing through current shipped modules. The Vite client does skip fetching a patch when the entire update is irrelevant to that tab, but a fetched patch can still co-carry an unexecuted importer from the server's static superset. I reproduced this with two tabs: a lazy module imports an executed shared module and another never-delivered dependency; editing the shared module makes the second tab fetch a patch that carries the lazy module but not that other dependency, so delivery records the lazy module as current. When that tab later loads the route, this `continue` skips the missing dependency and the parked factory throws `MissingFactoryError` from `initModule`. Please separate "visited" from "carried": omit the already-current module itself, but still walk through its static dependencies.

crates/rolldown_dev/src/bundling_task.rs
```diff
@@ -241,11 +256,29 @@ impl BundlingTask {
...
+            self
+              .dev_context
+              .insert_pending_payload(
+                patch.filename.clone(),
+                PendingPayload {
+                  client_id: update.client_id.clone(),
+                  modules: std::mem::take(&mut patch.carried),
+                },
+              )
+              .await;
```
hyfdev, line 279: Important: please drop updates whose client disconnected during computation. If a client disappears after the snapshot, the sequence-assignment loop leaves its patch at the placeholder `seq = 0`, but this unconditional insertion recreates a pending entry after `remove_client` already cleaned that session, and the invalid update is still exposed through `on_hmr_updates`. Vite ignores it because its socket lookup fails, so the pending entry can never be consumed; repeated races with fresh client IDs grow `pending_payloads` without a global bound. Retain only updates whose session was still present when the sequence was assigned.

crates/rolldown/src/hmr/hmr_stage.rs
```diff
@@ -226,19 +236,50 @@ impl<'a, Fs: FileSystem + Clone + 'static> HmrStage<'a, Fs> {
...
+                self.cache.module_idx_by_stable_id.get(stable_id.as_str()).copied()
+              {
+                if self.module_table().modules[module_idx].is_normal() {
+                  carried.insert(module_idx);
+                }
+              }
+            }
           }
+
+          let update = self.render_hmr_patch(carried, changed_ids.clone(), stamp_table).await?;
```
hyfdev, line 281: Important: please bound the lifetime of these per-client patch artifacts. This loop now creates a distinct filename, and usually a sourcemap, for every client on every edit. The pinned Vite consumer stores them in `BundledDev.memoryFiles`, whose only cleanup is `clear()` when the server closes; importing a patch does not remove it. A long-running dev server therefore permanently retains roughly clients × edits complete patch artifacts. Please add delivery-time deletion plus bounded eviction for files never requested, or share one rendered artifact when clients have the same carried set.

hyfdev (commented):
I found three important issues; details inline.

hyfdev (approved):
Approved with the important follow-ups noted inline.

PR #10210, opened 2026-07-09
docs: add pass-based pipeline design doc

hyfdev:
### Description

Adds `internal-docs/pass-based-pipeline/design.md` — a guiding methodology for structuring bundler-internal pipelines as passes with a compile-time ownership contract.

The contract: every pass is a zero-sized name token (`const`-asserted in `run_pass` at monomorphization) implementing one trait with four typed slots — `InputRead<'a>: Copy` (shared borrows; `Copy` makes `&mut` unrepresentable), `InputOwned` (to modify = to own and hand back), `OutputRead` (minted results — the `run_pass` wrapper seals them unconditionally in `Sealed<T>`, a no-`DerefMut`/no-unwrap container), and `OutputOwned` (the still-mutable data handed to a later pass). A single write-only `PassCtx` carries diagnostics. The driver is a typed `let`-chain: wrong pass order and mutation-after-seal fail to compile, and parallelism candidates are readable from signatures with the compiler checking the join.

Ground rules sit at the top on purpose: this is a methodology, not a migration mandate — no proactive refactors of existing pipeline code. It applies only when a maintainer explicitly asks for a pass-based refactor of a flow, or when designing a new flow.

Doc only, no code changes. No `implementation.md` yet — the first flow that adopts the contract adds it.

Note on the review threads: the first thread's resolution ("defer the trait") was later reversed once the `self`-state hole was closed mechanically (`Pass: Copy + 'static`, then the zero-sized assert in `run_pass`, `e88183a`/`71828f8`) and sealing moved into the `run_pass` wrapper itself (`14762bc`) — this description is the up-to-date record.


internal-docs/pass-based-pipeline/design.md
```diff
@@ -0,0 +1,182 @@
...
+  - what it seals forever
+  - what it hands onward, still mutable
+- The borrow checker enforces the declaration — not comments, not runtime panics.
+- The pipeline is a plain function (the driver); its `let`-chain is the pipeline diagram.
+- Wrong order does not compile: a step cannot name inputs that do not exist yet.
+- Sealed cannot be mutated: the `&mut` cannot be written.
+
+## What it looks like
+
+The whole mechanism is one trait, one marker trait, and one wrapper function:
```
hyfdev: I don't think the common `Pass` abstraction earns its cost yet. The useful properties described here already come from ordinary function signatures, the driver's `let`-chain, and distinct mutable/frozen artifact types. `run(self, cx, read, owned) -> (...)` adds GATs, four artificial slots, and `()`/tuple ceremony, but it does not itself encode semantic pass order, freeze outputs, or make dependencies exhaustive (`self` can still carry pipeline state unless separately restricted).

Could this methodology start with plain typed functions and explicit artifact lifecycles, and defer a shared `Pass` trait until multiple adopted flows show repetition that ordinary functions cannot express?

internal-docs/pass-based-pipeline/design.md
```diff
@@ -0,0 +1,182 @@
...
+  type InputRead<'a>: Copy;         // shared borrows only; Copy makes `&mut` unrepresentable here
+  type InputOwned;                  // data taken over (to modify = to own and hand back); `()` if none
+  type OutputRead: SealedArtifact;  // minted here, frozen from here on
+  type OutputOwned;                 // still-mutable data handed to a later pass
+
+  async fn run(self, cx: &mut PassCtx, read: Self::InputRead<'_>, owned: Self::InputOwned)
+    -> BuildResult<(Self::OutputRead, Self::OutputOwned)>;
+}
+
+/// Marker for frozen types: no `&mut` accessors, no interior mutability.
```
hyfdev: This marker does not enforce the invariant in the comment. Rust only checks that an implementation exists; it does not verify the absence of public mutable fields, mutable methods, or interior mutability, nor prevent a later pass from naming the same type as `InputOwned`. The actual compile-time boundary in the example comes from distinct `DraftChunkGraph` / `ChunkGraph` types and `ChunkGraph`'s restricted API.

Could we describe `SealedArtifact` as a reviewed assertion/inventory mechanism, and attribute immutability to the artifact type itself?

internal-docs/pass-based-pipeline/design.md
```diff
@@ -0,0 +1,182 @@
...
+pub trait SealedArtifact {}
+```
+
+- `self` carries the pass's configuration; `run` consumes it — a pass is single-shot.
+- `PassCtx` is the **single sanctioned `&mut`**: write-only sinks (diagnostics now, devtools trace later). It never contains pipeline data; passes may write it but never read it.
+- Slot conventions: empty slot = `()`; a single artifact = the bare type; multiple = a named per-pass struct (`XxxPassInputRead { .. }`, derives `Copy`).
+- Driver rules: every `OutputOwned` is consumed by exactly one later `InputOwned` (or explicitly dropped). **Seal order follows reference direction**: seal an artifact only when everything its keys/indices point into is already sealed.
+
+```text
+()  ──Split──►  Draft  ──Optimize──►  Draft  ──Seal──►  ChunkGraph (frozen)
```
hyfdev: This freeze point is misleading for the current `GenerateStage`: after chunk optimization, the graph is still mutated by cross-chunk link derivation, initialization-order and wrapping work, export computation, naming, and finalization.

A concrete existing freeze boundary is `UsedSymbolRefsBuilder::seal()`: source liveness becomes read-only while chunk layout remains mutable. Could the diagram use that proven boundary instead, and leave final `ChunkGraph` / `ChunkProgram` sealing to the first adopted flow?

internal-docs/pass-based-pipeline/design.md
```diff
@@ -0,0 +1,182 @@
...
+
+Sealed means sealed:
+
+```rust
+graph.add_chunk(chunk);
+// error[E0599]: no method named `add_chunk` found for struct `ChunkGraph`
+// (mutators exist only on DraftChunkGraph)
+```
+
+### Parallelism is provable from signatures alone
```
hyfdev: “Provable from signatures alone” is too strong unless the contract also requires every effect to be represented in the slots. Disjoint `InputOwned` values establish an important aliasing condition, but `self`, interior-mutable `InputRead` values, plugin or I/O calls, global state, and `PassCtx` effects can still make two bodies order-dependent.

Rust can prove data-race freedom through borrowing and `Send` / `Sync`; deterministic semantic independence additionally needs an explicit no-hidden-effects rule. Could we state those assumptions and narrow this to: under those rules, signatures expose candidates for concurrency and Rust checks the join?

internal-docs/pass-based-pipeline/design.md
hyfdev: Agreed — restructured in 08b7fc281. The contract is now stated as plain typed functions + explicit artifact lifecycle (draft/frozen type pairs); every property in Why holds without the trait. The trait moved to a "Deferred: a shared Pass trait" section that records what packaging it would buy (one `run_pass` home for tracing/diagnostics provenance, pinned signature shape via the `Copy` bound) and what it cannot do (encode order, freeze outputs, restrict `self`), with your adoption trigger: repetition across several adopted flows that ordinary functions cannot express.

internal-docs/pass-based-pipeline/design.md
hyfdev: Right — the marker checks nothing beyond impl existence. Reworded in 08b7fc281: immutability is now attributed to the artifact type's API ("the frozen type exposes no mutators and holds no interior mutability — nothing else grants it"), and `SealedArtifact` appears only in the deferred-trait section, described as a reviewed inventory of frozen types rather than an enforcement mechanism.

internal-docs/pass-based-pipeline/design.md
hyfdev: Fixed in 08b7fc281. The section now anchors on the proven in-tree boundary — `UsedSymbolRefsBuilder::seal()`, with the note that freeze lines are per-artifact, not global (liveness frozen while chunk layout stays mutable). The chunk-graph diagram is explicitly labeled a hypothetical target shape that today's `GenerateStage` does not have (its graph keeps being mutated by link derivation, wrapping, naming, and finalization past chunk optimization), and sealing the graph is left to the first adopting flow.

internal-docs/pass-based-pipeline/design.md
hyfdev: Agreed, narrowed in 08b7fc281. The section is retitled "signatures expose the candidates, the compiler checks the join" and now separates the three layers: signatures expose candidates; the borrow checker plus Send/Sync proves the join is data-race free; semantic independence is a stated discipline (no interior mutability in pipeline data, no globals, no order-dependent external calls, per-branch sinks merged in fixed order) — and a compiling join is deterministic only under those rules.

internal-docs/pass-based-pipeline/design.md
hyfdev: Follow-up for the record: later commits reversed this deferral — `e88183a` reinstated the trait once the `self`-state hole was closed mechanically (`Pass: Copy + 'static`, then the `size_of::<P>() == 0` const assert in `run_pass`), and `14762bc` moved sealing into the wrapper itself (`run_pass` returns `Sealed<P::OutputRead>`), which answers the "the trait does not itself freeze outputs" point. The "Why a trait at all" paragraph in the final text records the cost/benefit. Leaving this note so the thread's conclusion doesn't contradict the doc.

internal-docs/pass-based-pipeline/design.md
```diff
@@ -0,0 +1,253 @@
+# Pass-Based Pipeline — Design & Principles
+
+## Summary
+
+A guiding methodology for structuring bundler-internal pipelines (stage-level dataflow) as passes with a compile-time ownership contract: every pass implements one small trait declaring what it reads, owns, seals, and hands onward. Compiler gates pin the shape; what they cannot pin is a short, explicit review list (see [Enforcement](#enforcement)). This doc defines the contract; there is no implementation.md yet — the first flow that adopts it adds one.
```
hyfdev, line 5: **Delivery risk: full harness before a call site.**

The summary already freezes a complete shape (trait + four slots + `run_pass` + `Sealed`) while admitting there is no `implementation.md` and no adopting flow yet.

Suggestion: either (a) keep this as thin *principles* until the first real flow lands and reverse-write the harness from that PR, or (b) name the first adoption target in Related / Future and treat every compile-error claim as unpaid debt until trybuild exists. Without a call site, this doc mostly freezes maintainer taste, not pipeline behavior.

internal-docs/pass-based-pipeline/design.md
```diff
@@ -0,0 +1,253 @@
...
+
+## Ground rules (read this first)
+
+- This is a methodology, **not a migration mandate**. Do **not** proactively refactor existing pipeline code into passes.
```
hyfdev, line 9: **Ground rules vs cultural gravity.**

The non-migration rule is right. Once this sits in `internal-docs/` as a complete contract, review pressure still tends to become “why isn’t this a Pass?” even when nobody asked for a pass refactor.

Suggestion: add an explicit *acceptance test for calling something a Pass* (nameable artifact, top-level driver step, maintainer-requested or new flow only) and state that failing that test is a reason to reject Pass-ification in review — not only to avoid proactive refactors by authors.

internal-docs/pass-based-pipeline/design.md
```diff
@@ -0,0 +1,253 @@
...
+/// A pass type is a **name**, not a value: `run_pass` compile-time-asserts it is
+/// zero-sized, and the bounds keep it lifetime-free. All runtime data — including
+/// configuration — enters through the declared slots.
+pub trait Pass: Copy + 'static {
+  type InputRead<'a>: Copy;         // shared borrows only; Copy makes `&mut` unrepresentable here
+  type InputOwned;                  // data taken over (to modify = to own and hand back); `()` if none
+  type OutputRead;                  // minted here; `run_pass` wraps it in `Sealed<_>` — frozen unconditionally
+  type OutputOwned;                 // still-mutable data handed to a later pass
+
+  async fn run(self, cx: &mut PassCtx, read: Self::InputRead<'_>, owned: Self::InputOwned)
```
hyfdev: **`async fn` on every pass looks like shape tax.**

Most stage work is sync CPU. Putting `async` on the trait forces `async_fn_in_trait` (already noted) and `.await?` noise on every call site for a capability few passes need.

Suggestion: default the contract to sync `fn run(...)`; add async only for the rare I/O boundary, or keep a thin async driver wrapper outside the trait. Don’t make the common case pay for the uncommon one.

internal-docs/pass-based-pipeline/design.md
```diff
@@ -0,0 +1,253 @@
...
+  let (minted, owned_out) = pass.run(cx, read, owned).await?;
+  Ok((Sealed::new(minted), owned_out)) // the harness seals — there is no unsealed exit
+}
+
+/// The hard freeze: only `&T` ever comes out — no `DerefMut`, no `into_inner`,
+/// private field. No mutation path exists, even for an owner.
+pub struct Sealed<T>(T);
+
+impl<T> Sealed<T> {
+  pub fn new(value: T) -> Self {
```
hyfdev: **`Sealed::new` is `pub` — the “harness seals” claim is social, not constructional.**

If any code can `Sealed::new(x)`, then “nothing exits `OutputRead` unfrozen by construction” is only true for the `run_pass` return path, not for the type system. Minting sealed values outside a seal pass is still quiet.

Suggestion: `Sealed::new` should be `pub(crate)` and only reachable from the harness module (same privacy story as the private field). If examples need construction in tests, put a test-only constructor behind `#[cfg(test)]` or a dedicated test helper in the harness crate.

internal-docs/pass-based-pipeline/design.md
```diff
@@ -0,0 +1,253 @@
...
+- Slot types: empty = `()`; a single artifact = the bare type; multiple = a named per-pass struct (`XxxPassInputRead { .. }`, derives `Copy`), which doubles as the pass's greppable dependency manifest.
+- Impl signatures copy the trait verbatim: write `read: Self::InputRead<'_>` even when the slot is `()` — spelling the parameter as a concrete type drops the method's lifetime binder and fails E0195.
+- Naming: a pass type ends in `Pass` (`OptimizeChunksPass`); its input struct is `<PassName>InputRead`. Since passes are zero-sized name tokens, the suffix keeps call sites unambiguous — and `rg 'struct \w+Pass;'` is the complete pass inventory (it matches unit structs only; a fielded `struct XPass(u64);` evades the grep, and it is the build-time assert that actually rejects it).
+- **The hard guarantee is `Sealed<T>`, and the harness applies it**: `run_pass` wraps every read-side output, so nothing can exit `OutputRead` unfrozen — by construction, not convention. `Sealed<T>` has no mutation path even for an owner, so frozenness survives re-ownership with no ledger discipline needed.
```
hyfdev: **Hard/soft layering looks inverted for correctness.**

`Sealed<T>` blocks `&mut self` through the wrapper — useful anti-forget. It does **not** make `T` a good frozen artifact. `Sealed<LinkingMetadata>` is still a fact bag; indices, field meaning, and “should this concept still be live?” are untouched.

The in-tree precedent you cite below (`UsedSymbolRefsBuilder::seal()` → `UsedSymbolRefs` with a tiny API) is a **domain draft/final type pair**. That is the correctness freeze. Representation changes (`Vec` → `Box<[T]>`) and `Sealed` are hygiene / anti-forget around it.

Suggestion: restate hard guarantees as: (1) draft vs final *types* with mutators only on the draft side, (2) by-value ownership for mutation, (3) driver `let`-chain order. Keep harness `Sealed<T>` as a secondary “cannot forget to wrap `OutputRead`” layer — not the primary correctness story.

internal-docs/pass-based-pipeline/design.md
```diff
@@ -0,0 +1,253 @@
...
+let (_, graph) = run_pass(OptimizeChunksPass, &mut cx, optimize_reads, graph).await?;
+let (graph, ()) = run_pass(SealChunkGraphPass, &mut cx, (), graph).await?; // graph: Sealed<ChunkGraph>
+let (names, ()) = run_pass(AssignNamesPass, &mut cx, names_reads, ()).await?;
+```
+
+`OutputRead = ()` still comes back as `Sealed<()>` — discard it with `_`; a literal `()` pattern does not match it (E0308). The owned side is never wrapped, so `()` patterns are fine there.
+
+Needing to own more than you reshape is the signal that an artifact should be split out — not a reason to widen the slot.
+
+The same rule applies to reads. Entity tables (`ModuleTable`, `SymbolRefDb`) are legitimate inputs; metadata god-structs are not — declaring `&IndexVec<ModuleIdx, LinkingMetadata>` in an `InputRead` would launder today's grab-bag through the contract, and the manifest is only as informative as the granularity of the types it names. A pass declares the specific facts it consumes (`&ModuleSideEffects`, `&WrapKinds`); breaking blobs like `LinkingMetadata` into per-pass artifacts is much of why this contract exists.
```
hyfdev, line 171: **Read-side anti-laundering is strong; write/mint side is still open.**

Banning god-structs from `InputRead` is the right rule and is much of why a contract would pay off. The same hole exists on mint: `OutputRead = SomeBigBag` then `run_pass` seals it still launders a grab-bag — only now it is immutable.

Suggestion: state a symmetric mint rule: `OutputRead` should be a *named, purpose-specific* artifact (or a small struct of those), not a stage dump. If you cannot name what fact the sealed type answers, it is not ready to be an `OutputRead`. That is review-held, but it belongs next to the read rule so the contract’s purpose (“break `LinkingMetadata`”) cannot be satisfied by sealing the bag whole.

internal-docs/pass-based-pipeline/design.md
```diff
@@ -0,0 +1,253 @@
...
+Sealed means sealed:
+
+```rust
+graph.add_chunk(chunk);
+// error[E0599]: no method named `add_chunk` found for struct `Sealed<ChunkGraph>`
+// (method lookup walks through Deref and still finds nothing — mutators exist
+//  only on DraftChunkGraph, and what you hold after sealing is Sealed<ChunkGraph>)
+```
+
+### Parallelism: signatures expose the candidates, the compiler checks the join
```
hyfdev: **Parallelism is a weak primary “Why”.**

The three-layer split (candidates / data-race-free join / semantic discipline) is honest. The realism note then shows the usual win for Rolldown is **rayon inside a pass**, which does not need this trait.

Suggestion: demote this section under “Other benefits” or Future. Lead Why with: wrong order fails to compile, seal-after-mutate fails to compile, cross-pass state has no quiet home, and read/mint manifests force fact granularity. Don’t sell Pass primarily as a concurrency framework.

internal-docs/pass-based-pipeline/design.md
```diff
@@ -0,0 +1,253 @@
...
+- no escape through raw pointers — `#![forbid(unsafe_code)]` on the passes module
+- pass order — `let`-chain scoping; sealing — applied by `run_pass` itself (`Sealed<T>`: no `DerefMut`, no unwrap); ownership transfer — moves
+
+What stays review-held (the honest list):
+
+- frozen representations inside artifacts (`Box<[T]>` over `Vec<T>`) — hygiene, not correctness; the hard freeze is `Sealed<T>`, applied by the harness
+- no interior mutability in pipeline data types; no global statics holding pipeline data (the one remaining way to smuggle state past the bounds)
+- artifact granularity (own exactly what you reshape); new `PassCtx` methods staying write-only (drain methods take `self` by value, unreachable through `&mut`)
+
+Why a trait at all: uniform pass style — every pass declares the same four slots and is invoked through one wrapper — plus the shape gates above, which plain functions cannot carry (nothing stops an extra `&mut` parameter on a plain function except review). The cost — GATs and `()`/tuple ceremony — is accepted deliberately.
```
hyfdev: **“Why a trait at all” still under-proves the cost.**

Agreed that plain functions don’t pin a four-slot shape or a single `run_pass` home for tracing. But:

- review already blocks a random extra `&mut` on a carefully typed driver;
- the trait does **not** block wrong-granularity `InputRead` / bag-shaped `OutputRead` (the laundering you care about);
- ZST + monomorphization assert does not fire under `cargo check` / rust-analyzer (already caveated), so the daily loop’s “impossible to write quietly” bar is softer than this paragraph implies.

The earlier self-review deferral (plain functions first; trait when repetition appears across flows) still looks like the better default. Reinstate the trait when the first two adoptions actually share ceremony that hurts — not because the `self`-state hole is closable on paper.

internal-docs/pass-based-pipeline/design.md
```diff
@@ -0,0 +1,253 @@
...
+
+## Future directions
+
+- Driver-level joins of parallel-candidate passes, once profiling shows a win worth taking (see the realism note in Why — plain `try_join!` only interleaves; multicore needs a scoped-task driver or stays inside passes via rayon).
+- When the first flow adopts the contract, pin the compile-error claims (E0308 / E0425 / E0599 / E0277) as `trybuild` compile-fail tests — the doc promises compiler behavior, and tests should hold it.
+- Incremental cache friendliness: explicit inputs are natural dependency keys, sealed artifacts are natural snapshot/hash units, and a pass is a natural recompute unit. To be honest: the contract was **not** designed with incremental builds as a premise, and nothing in it depends on that — it simply does not stand in the way.
+
+## Related
+
+- `implementation.md` — none yet; added by the first flow that adopts the contract.
```
hyfdev: **Related is empty of a first flow.**

“`implementation.md` — none yet” is honest but leaves the methodology without a success criterion.

Suggestion: name one concrete first adoption (even if small), e.g. a single generate-stage baton already close to seal (`UsedSymbolRefs` is precedent, not a second program), or the next artifact you would extract from a hot path — and define done as: driver order/seal mistakes fail to compile + trybuild pins the errors in this doc. Until that lands, prefer not to grow Future directions (scoped tasks, incremental keys) past what the first flow needs.

hyfdev (commented):
## Adversarial review (methodology / design)

Overall: the **problem** is real (implicit order, shared mutability, fact-bag reads). Ground rules, the honest Enforcement split, the `LinkingMetadata` anti-example, and anchoring freeze on the in-tree `UsedSymbolRefs` seal are the strong parts.

Main pushback: this still looks like **control-flow packaging ahead of evidence**. The main lever for pipeline complexity is artifact / fact granularity (draft–final types and not laundering bags), not a shared `Pass` trait. Several “hard” claims are softer than the prose (`pub Sealed::new`, monomorphization-only ZST assert, `async` on all passes, parallelism as a Why).

Recommendation shape: thin principles + one named first flow + trybuild when adopted; defer full GAT harness until two flows share real ceremony. Inline notes below are concrete edit asks, not a merge block on the existence of a design note — but I would not treat the current harness as law until a call site exists.

(No coupling assumed to parked paper programs such as full sealed-chunk-graph layering; only the already-merged `UsedSymbolRefs` seal is treated as in-tree precedent.)

internal-docs/pass-based-pipeline/design.md
hyfdev: Adopted the (b) shape in 4b03490e0: Related now states the success criterion — the first adoption defines done as driver order/seal mistakes failing to compile plus trybuild pinning the compile-error claims, and until then the claims are labeled principles awaiting their call site. Naming a concrete first flow is deliberately left out: per Ground rules a maintainer picks it, and this doc must not start a program by itself.

internal-docs/pass-based-pipeline/design.md
hyfdev: Adopted in 4b03490e0: the granularity test is now an explicit acceptance test (nameable artifact/baton, top-level driver step, maintainer-requested or new flow — all three or it is not a pass), stated as grounds to reject Pass-ification in review, cutting against "why isn't this a Pass?" pressure as much as against proactive authors.

internal-docs/pass-based-pipeline/design.md
hyfdev: Adopted in 4b03490e0 — the contract is sync by default. Await-needing work (plugin hooks, user callbacks) is hoisted to the driver between passes, which is already this repo's pattern (async `paths` pre-resolution before sync rendering); an `AsyncPass` variant waits until a flow actually needs one. Bonus cascade: the `async_fn_in_trait` note is gone, the E0195 impl-signature rule turned out to be an async-desugaring artifact (the sync form compiles — verified) and is deleted, and the parallelism example upgraded from `try_join!` interleaving to `rayon::join`, which is real multicore.

internal-docs/pass-based-pipeline/design.md
hyfdev: Adopted in 4b03490e0: `Sealed::new` is harness-private — `run_pass`, in the same leaf module, is the sole minting point, and a sibling pass calling `Sealed::new` fails E0624 (compile-verified in the same probe layout as the E0616 field check). "The harness seals" is now constructional on both the wrap and the mint side; test fixtures get a `#[cfg(test)]` helper in the harness crate when needed.

internal-docs/pass-based-pipeline/design.md
hyfdev: Partially adopted in 4b03490e0. The layering is restated as two freezes plus hygiene: mutation-freeze (`Sealed<T>`, applied by `run_pass` — the anti-forget gate) and domain-freeze (draft/final type pairs with mutators only on the draft side — the correctness story, `used_symbol_refs` as precedent), with representation as hygiene. The doc now says explicitly that `Sealed<T>` does not make `T` a good artifact — `Sealed<LinkingMetadata>` is still a fact bag — and the new mint rule is what blocks sealing bags whole. Kept `Sealed` as a hard gate rather than demoting it: it is the one layer that is unconditional and survives re-ownership; the two freezes answer different failure modes rather than ranking on one axis.

internal-docs/pass-based-pipeline/design.md
hyfdev: Adopted in 4b03490e0, placed directly beside the read rule as suggested: `OutputRead` must be a named, purpose-specific artifact — if you cannot say which fact the sealed type answers, it is not ready to be an `OutputRead`; sealing a stage dump makes the bag immutable, not decomposed. Also echoed in the Enforcement review-held list.

internal-docs/pass-based-pipeline/design.md
hyfdev: Adopted in 4b03490e0: the section moved under Other benefits, and Why now leads with the compile-error properties. The sync switch from the other thread strengthened what remains: the example is now a driver-level `rayon::join` (real multicore, compile-verified with shared `&graph` and per-branch sinks), with intra-pass `par_iter_mut` named as the everyday form and the honest discipline list kept.

internal-docs/pass-based-pipeline/design.md
hyfdev: Keeping the trait, with the counter-argument now recorded in the doc: after `Sealed::new` went harness-private, `run_pass` is the sole sealing/minting point, and the `OutputRead` slot is precisely what lets one generic wrapper seal every pass's result — plain functions have no slot for a wrapper to seal, so unconditional sealing would degrade to per-function memory, which is the anti-forget property this round just hardened. Your three sub-points are each true and each recorded (review blocks stray `&mut`; granularity is review-held via the read/mint rules; the ZST assert is build-time-only, caveated in place). The remaining cost also shrank this round: `run` is sync, so the ceremony is four associated types and a tuple. If the first two adoptions show the ceremony hurting anyway, re-deferring stays on the table — that evidence beats re-litigating now.

internal-docs/pass-based-pipeline/design.md
hyfdev: Adopted the criterion, declined the naming, in 4b03490e0: Related now defines done for the first adoption (order/seal mistakes fail to compile in that flow; trybuild pins the error claims in this doc) and labels everything before that as principles awaiting their call site. Naming a specific first target is deliberately left to Ground rules — the previously obvious candidate (sealing the chunk graph) was parked as a program on 07-12, and this doc should not un-park it. Future directions stayed at three lines, one of which is the trybuild debt itself.

internal-docs/pass-based-pipeline/design.md
```diff
@@ -0,0 +1,250 @@
...
+- `self` carries **nothing**: every pass is a zero-sized name token (`const`-asserted in `run_pass`), kept only so call sites read as `run_pass(OptimizeChunksPass, ..)`. Configuration is runtime data and enters through `InputRead` like everything else. Caveat: the assert is evaluated at monomorphization — `cargo build` rejects a stateful pass, but `cargo check` and rust-analyzer do not run it.
+- Slot types: empty = `()`; a single artifact = the bare type; multiple = a named per-pass struct (`XxxPassInputRead { .. }`, derives `Copy`), which doubles as the pass's greppable dependency manifest.
+- Naming: a pass type ends in `Pass` (`OptimizeChunksPass`); its input struct is `<PassName>InputRead`. Since passes are zero-sized name tokens, the suffix keeps call sites unambiguous — and `rg 'struct \w+Pass;'` is the complete pass inventory (it matches unit structs only; a fielded `struct XPass(u64);` evades the grep, and it is the build-time assert that actually rejects it).
```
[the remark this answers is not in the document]
hyfdev: Updated in 457de1060. The caveat now distinguishes code-generating builds from check-only and IDE workflows and says not to rely on their silence. Pass tokens are now explicitly review-held as non-generic unit structs, with any needed visibility, so generic and other zero-sized declaration shapes are disallowed and the inventory grep remains complete.

PR #10219, opened 2026-07-10
docs: module conversion proposal

sapphi-red:
The current non-JS moduleType behavior doesn't work well with `this.load` and has multiple inconsistent / unintuitive behavior. This PR adds a internal-docs to fix these issues as a proposal.

The final goal is to make the latter phase of https://github.com/vitejs/vite/issues/22709 happen.

The background context, the existing problem, the solution, the design decisions, the future extension was thought by me. The implementation investigation, memory section, writing this doc down in a organized way was done by AI.

### DIFF
--- a/internal-docs/module-conversion/design.md
+++ b/internal-docs/module-conversion/design.md
@@ -1,0 +1,714 @@
+# Module Conversion — Design & Principles
+
+## Summary
+
+**Module conversion** is the step that turns a module's native content into
+JavaScript, according to its `ModuleType`. For `js` it is a no-op; for
+`ts`/`tsx`/`jsx` it is the oxc transformer; for `json`/`text`/`base64`/
+`dataurl`/`binary`/`empty` it is a source rewrite; for `asset`/`copy` it emits
+a file and yields a reference to it. These are all the same kind of operation
+— `ModuleType` → JS — and today only some of them are treated that way.
+
+This document defines module conversion as a single named pipeline stage with a
+single observable contract:
+
+> **`ModuleInfo.content` is the module's content as it exists after the
+> `transform` hook and before module conversion.**
+
+TypeScript already satisfies this: `this.load()` on a `.ts` module hands back
+TypeScript source, not the lowered JavaScript, because lowering happens after
+`transform`. Every other module type should behave the same way. That is the
+whole of the proposal; everything below is what it costs to get there.
+
+Naming: the docs currently call this stage `internalTransform`
+(`docs/apis/plugin-api.md:130,141-142,151,248`). That name describes the
+TypeScript/JSX half and hides the rest. It should be renamed to
+**module conversion**.
+
+All file/line references are against the working tree at the time of writing
+and will drift; treat them as starting points.
+
+## Target behavior
+
+What `this.load({ id })` should return, per module type:
+
+| `moduleType` | `content`                                      |
+| ------------ | ---------------------------------------------- |
+| `js`         | `export default "foo"`                         |
+| `jsx`        | `export default <Foo />`                       |
+| `ts`         | `export default "foo" satisfies string`        |
+| `tsx`        | `export default <Foo /> satisfies JSX.Element` |
+| `json`       | `{ "foo": "bar" }`                             |
+| `text`       | `foo`                                          |
+| `base64`     | binary content (not base64-encoded)            |
+| `dataurl`    | binary content (not encoded as a data URL)     |
+| `binary`     | binary content                                 |
+| `empty`      | binary content                                 |
+| `css`        | `.foo { color: red }`                          |
+| `asset`      | binary content                                 |
+| `copy`       | binary content                                 |
+
+Read down the column: it is always "what is in the file, after plugins have had
+their say." No placeholders, no encodings, no generated wrappers.
+
+## What is wrong today
+
+Conversion currently lives in three places, at three different points in the
+pipeline, and `this.load` returns whatever each place happened to leave behind.
+
+**1. `pre_process_source`** — `crates/rolldown/src/utils/parse_to_ecma_ast.rs:141-216`,
+plus the `Json` arm of its caller at `:83-104`. This is the correct home. It
+handles `text`/`base64`/`dataurl`/`binary`/`empty`, and it runs after
+`transform`, in the same stage as TypeScript/JSX lowering.
+
+**2. `AssetModulePlugin::load`** — `crates/rolldown_plugin_asset_module/src/lib.rs:126-189`.
+Runs at `load`, i.e. _before_ `transform`. It reads the file, moves the bytes
+into `EmittedAsset`, and returns `module.exports = "__ROLLDOWN_ASSET__#<ref>"`
+with `module_type: Some(ModuleType::Js)`. By the time anything downstream looks
+at the module, the bytes are gone and it no longer knows it was ever an asset.
+
+**3. `CopyModulePlugin::resolve_id`** — `crates/rolldown_plugin_copy_module/src/lib.rs:51-131`.
+Runs even earlier. It reads the file, emits it, and returns a prefixed id marked
+`external: true`. The module never enters the graph at all.
+
+The observable result:
+
+| `moduleType`          | `this.load` returns today                                                            |     |
+| --------------------- | ------------------------------------------------------------------------------------ | --- |
+| `js`/`jsx`/`ts`/`tsx` | the source                                                                           | ✅  |
+| `json`                | `""`                                                                                 | ❌  |
+| `text`                | `"foo"` (JSON-escaped, quoted)                                                       | ❌  |
+| `base64`              | `"aGVsbG8="`                                                                         | ❌  |
+| `dataurl`             | `"data:text/plain;base64,aGVsbG8="`                                                  | ❌  |
+| `binary`              | `import {__toBinary} from 'rolldown:runtime'; export default __toBinary('aGVsbG8=')` | ❌  |
+| `empty`               | `""`                                                                                 | ❌  |
+| `css`                 | hard error before parse (`module_task.rs:117-124`)                                   | ❌  |
+| `asset`               | `module.exports = "__ROLLDOWN_ASSET__#ref-0"`                                        | ❌  |
+| `copy`                | hard error: "Encountered a module with type `copy`, but no plugin handled it"        | ❌  |
+
+Three of these deserve comment.
+
+**`json` returns the empty string.** `json_value_to_ecma_ast` constructs its AST
+with `ArcStr::from("")` as the source
+(`crates/rolldown_common/src/ecmascript/json_to_program.rs:30`), and
+`ecma_view.source` is `ast.source()`. So every JSON module in the graph reports
+`code: ""`. This is a latent bug that the invariant fixes as a side effect.
+
+**`copy` errors rather than returning a placeholder.** `this.load` bypasses
+`resolveId` entirely — it constructs a `ResolvedId` directly from the specifier
+(`crates/rolldown_plugin/src/plugin_context/native_plugin_context.rs:64-71`) —
+so `CopyModulePlugin::resolve_id` never runs, and the core loader reaches the
+`ModuleType::Copy` arm of `load_source` and bails
+(`crates/rolldown/src/utils/load_source.rs:97-101`).
+
+**`asset` modules currently run through the `transform` hook**, on the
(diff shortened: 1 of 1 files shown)


sapphi-red:
* **#10219** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10219?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10219?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10219).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

shulaoda (approved):
LGTM. If we do this, it also seems to resolve some of the current asset-related issues under FBM. My only concern is whether this would have a significant impact on existing Vite and Rolldown plugins. 🤔

h-a-n-a (approved):
Overall it looks great. There's some details that I need to focus more on but it should not block us from merging it.

internal-docs/module-conversion/design.md
```diff
@@ -0,0 +1,644 @@
...
+`options.platform`. Rather than thread all of that into a core `match`, module
+conversion becomes a Rust-only plugin hook, following the precedent of
+`transform_ast` (`crates/rolldown_plugin/src/plugin.rs:284-292`), which is
+registered in `HookUsage` but has no JavaScript counterpart. `AssetModulePlugin`
+and `CopyModulePlugin` keep their identity as plugins — see
+[plugin-asset-module](../plugin-asset-module/implementation.md) for why that
+matters — and simply move their work from `load`/`resolve_id` to the hook that
+runs at the right time.
+
+**5. Content is shared, never copied.**
```
hyfdev, line 263: Internal sharing and the JavaScript-facing contract need to be separated here. TypeScript `readonly` only prevents replacing the property; a `Uint8Array`'s elements remain writable. Mutating `info.content[0]` must not change the module content or the source retained by `FileEmitter`.

A mutation-isolated snapshot at the JavaScript boundary would keep `content` observation-only while allowing Rust-side storage to remain shared. Content changes can continue to go through `load`/`transform` return values. This does not block this approval.

internal-docs/module-conversion/design.md
```diff
@@ -0,0 +1,644 @@
...
+   Whether `asset` should migrate to `ROLLUP_FILE_URL` is a separate decision
+   with its own compatibility story.
+4. **`Custom(_)` module types have no conversion.** They error today
+   (`parse_to_ecma_ast.rs:208-212`). `convert` should presumably reject them,
+   but an internal-only conversion hook means a JavaScript plugin still cannot
+   _register_ one.
+
+## Unresolved Questions
+
+- **The copy proxy's three properties** — the copy design is settled (proxy
```
hyfdev: A plain JavaScript re-export proxy does not preserve the current `require()` behavior. For a conditional `require('./hello.txt')`, the current copy plugin keeps a runtime `require()` of the emitted path inside the branch. The proxy instead creates a top-level static import and returns `__toCommonJS(proxyNamespace)`, changing both load timing and the returned value.

Static ESM and side-effect imports work, but preserving mixed ESM/`require()` use needs a linker-level transparent external alias that retains each incoming edge's `ImportKind`, or another representation with the same property. Keeping `copy` in phase 2 makes sense; the proxy representation remains unresolved rather than settled. This does not block this approval.

hyfdev (approved):
The `module conversion` model covers this stage well. Exposing post-transform, pre-conversion `content` gives the different module types one predictable observation point without making the internal conversion hook public.

I left two non-blocking notes about mutation isolation at the JavaScript boundary and the phase-2 `copy` representation.

— ~~AI review by GPT-5~~ Review my me manually, AI takes my credit

internal-docs/module-conversion/design.md
```diff
@@ -0,0 +1,644 @@
...
+`asset` inlines a path _string_, `copy` keeps a real import of a file that was
+copied verbatim.
+
+For `this.load` to return its bytes, `copy` must enter the graph. The copy
+module becomes a real module of type `Copy`; `load` reads its bytes; conversion
+emits the file and returns a **proxy** — a re-export from an external
+placeholder:
+
+```js
+export * from '\0copy:ref-0';
```
IWANABETHATGUY, line 490: `copy` is also reachable via `import('./file.wasm')`. Today it's external at `resolveId`, so the output stays a direct dynamic import of the copied file. As an in-graph proxy that target becomes an internal dynamic entry — Rolldown emits a JS proxy chunk, the `import()` loads that chunk, and the chunk statically imports the copied file, so you get an extra chunk, an extra request hop, and different load timing. Same root cause as the `require()` case hyfdev noted below (a plain re-export proxy loses the per-edge `ImportKind`), and whatever preserves that needs to hold for `import()` too — worth a dynamic-import regression test next to the three proxy properties.

internal-docs/module-conversion/design.md
```diff
@@ -0,0 +1,644 @@
...
+be unsound (a module's identity is its id; `try_spawn_new_task` returns early on
+`VisitState::Seen`, `module_loader.rs:320-326`, so the second caller's override
+would be silently dropped), and as an assertion it would only do what the caller
+can do at the call site.
+
+One consequence to know: `this.load` builds a `ResolvedId` with
+`..Default::default()` (`native_plugin_context.rs:64-71`), so it never sees the
+`asserted_module_type` an import edge carries — `with { type: … }`, or
+`new URL()` ⇒ `Asset` (`new_url.rs:69`). A `.png` reached only by `new URL()`,
+in a project with no `moduleTypes` entry for it, therefore comes back as `Js`
```
IWANABETHATGUY: This escape hatch can't actually run for the case it's about. A `.png` with no `moduleTypes` entry, reached by `this.load` first, hits `load_source`'s `(None, None)` arm → `read_to_string`, and a real PNG isn't valid UTF-8, so it errors *during* load. There's no completed `ModuleInfo` left to read `moduleType` off of afterward. This one needs a pre-load type/assertion channel or another raw-content path, not a post-load check.

internal-docs/module-conversion/design.md
```diff
@@ -0,0 +1,644 @@
...
+`module.exports` assignment from leaking into ESM output.
+
+## `ModuleInfo` surface
+
+```ts
+interface ModuleInfo {
+  /** The module's content, after `transform` and before module conversion. */
+  content: string | Uint8Array | null;
+  /** The module type, needed to interpret `content`. */
+  moduleType: ModuleType;
```
IWANABETHATGUY: A required `moduleType: ModuleType` can't describe every `ModuleInfo` state. `module_task.rs` installs a placeholder before `load` even runs, and `load`/`transform` may still pick the final type after that; external modules never get a conversion type at all. Defaulting those to `Js` would surface wrong metadata through `getModuleInfo` mid-load. Suggest `ModuleType | null` (or an explicit unknown state), required only once content exists.

internal-docs/module-conversion/design.md
```diff
@@ -0,0 +1,644 @@
...
+| `js`         | `export default "foo"`                         |
+| `jsx`        | `export default <Foo />`                       |
+| `ts`         | `export default "foo" satisfies string`        |
+| `tsx`        | `export default <Foo /> satisfies JSX.Element` |
+| `json`       | `{ "foo": "bar" }`                             |
+| `text`       | `foo`                                          |
+| `base64`     | binary content (not base64-encoded)            |
+| `dataurl`    | binary content (not encoded as a data URL)     |
+| `binary`     | binary content                                 |
+| `empty`      | empty text                                     |
```
IWANABETHATGUY: This row contradicts the invariant. For a non-empty file mapped to `empty` (or a `load` returning text as `moduleType: 'empty'`), `load_source` keeps the real text as `Str` and `transform` runs on it — `pre_process_source` only discards it afterward, during conversion. So at the promised observation point (after transform, before conversion) the content is the transformed text, not empty text. And stubbing a real file to `empty` is the normal use, not a corner. Either show the raw text in this row, or move `Empty` handling before `transform` and own the behavior change.

internal-docs/module-conversion/design.md
```diff
@@ -0,0 +1,644 @@
...
+
+### Whether `transform` runs at all is decided by who loaded the module
+
+`module_task.rs:284-302` runs `transform` for `StrOrBytes::Str` and skips it for
+`StrOrBytes::Bytes`. Which arm a module lands in is not a property of its module
+type; it is a property of _who produced its content_.
+
+Configure `moduleTypes: { '.png': 'base64' }` and the core reads the file as
+bytes (`load_source.rs:79-96`), so `transform` never fires. Add a plugin whose
+`load` returns a string for that same `.png` and `load_source` takes the `Str`
```
IWANABETHATGUY: Minor: this doesn't behave as written. A bare string with no `moduleType` gets inferred as `Base64` and converted to `StrOrBytes::Bytes`, so `transform` stays skipped. It only fires if the hook *also* returns an explicit `moduleType` (the `(Some, Some)` arm keeps it `Str`). The larger point — dispatch is accidental today — still holds; the example just needs that condition.

internal-docs/module-conversion/design.md
```diff
@@ -0,0 +1,644 @@
...
+interface ModuleInfo {
+  /** The module's content, after `transform` and before module conversion. */
+  content: string | Uint8Array | null;
+  /** The module type, needed to interpret `content`. */
+  moduleType: ModuleType;
+  /**
+   * @deprecated Use `content`. A string-only view: `null` when the module's
+   * content is binary.
+   */
+  code: string | null;
```
IWANABETHATGUY, line 331: `code === null` is documented today as external-or-not-yet-available (`module-info.ts:12`). This also makes it mean loaded-but-binary. The internal `this.load` dedupe guard is handled, but third-party plugins using the same `code === null` check will read a loaded binary as not-loaded, and quietly. Worth a line in the breaking-change note pointing authors at `content`/`moduleType` for that check.

internal-docs/module-conversion/design.md
```diff
@@ -0,0 +1,644 @@
...
+given module type uses is an implementation detail; that it happens _in this
+stage and nowhere earlier_ is the contract.
+
+**2. `this.load` observes the input to conversion, not its output.** This is
+the invariant. It is what makes `this.load` predictable: the content you get
+back is the content the module type names.
+
+**3. `ecma_view.source` stays post-conversion.** It cannot carry pre-conversion
+content, because every AST span indexes into it and the codegen sourcemap builds
+`sourcesContent` from it (`crates/rolldown/src/utils/render_ecma_module.rs:104-113`).
```
IWANABETHATGUY: Small citation fix: `render_ecma_module.rs:104-113` doesn't build `sourcesContent` from `ecma_view.source` — that block is the `map: null` fallback and actually swaps in the *pre-transform* content; the file never touches `ecma_view.source`. The real path is `NormalModule::render` → `EcmaCompiler::print_with`, where the codegen map is produced. The AST-span reason (every span indexes into it) is the real one anyway, so the clause can just be dropped or re-pointed.

IWANABETHATGUY (commented):
Doc-level notes, all non-blocking — the invariant is the right call and I'd approve too. A few spots where the current loader doesn't quite satisfy it yet, plus the `ModuleInfo` shape. I've left out the two hyfdev already raised (shared-buffer mutation isolation, and `require()` under the copy proxy) so this doesn't repeat the thread.

IWANABETHATGUY (approved):

internal-docs/module-conversion/design.md
sapphi-red: I think this is not considered a breaking change because binary `content` only occurs for non-JavaScript module types, which are experimental. I'll clarify this in the doc.

internal-docs/module-conversion/design.md
sapphi-red: I think the current `new URL` handling didn't fit in with the new concept. I've updated this part to explain how to fix that.

internal-docs/module-conversion/design.md
sapphi-red: Updated to use a CJS proxy for `require`s.

internal-docs/module-conversion/design.md
sapphi-red: https://github.com/rolldown/rolldown/pull/10219#discussion_r3637223644

PR #10220, opened 2026-07-10
feat: add `FILE_NOT_FOUND` error

sapphi-red:
Add `FILE_NOT_FOUND` error that exists in Rollup.


sapphi-red:
* **#10222** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10222?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10221** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10221?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10220** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10220?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10220?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10220).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown/src/module_finalizers/mod.rs
```diff
@@ -97,6 +99,20 @@ pub struct ScopeHoistingFinalizer<'me, 'ast: 'me> {
   pub transferred_import_record: FxIndexMap<ImportRecordIdx, String>,
   pub rendered_concatenated_wrapped_module_parts: RenderedConcatenatedModuleParts,
   pub json_module_inlined_prop: Option<Box<FxHashMap<SymbolId, ast::Expression<'ast>>>>,
+  /// Reference ids of `import.meta.ROLLUP_FILE_URL_*` accesses that no emitted file matches.
+  ///
+  /// Deduplicated by reference id, because `try_rewrite_member_expr` runs *twice* on every member
+  /// expression it fails to rewrite: `visit_expression` calls it, and on `None` the arm falls
+  /// through to `walk_expression`, which re-dispatches the very same node into
+  /// `visit_member_expression`, where the identical lookup is attempted again. Pushing
+  /// `BuildDiagnostic`s straight into a `Vec` here would report each unknown reference id twice.
```
sapphi-red, line 107: I think this "twice traverse" issue should be fixed instead. But I didn't have time to properly understand the behavior there. So I left this for now.

crates/rolldown/src/module_finalizers/mod.rs
```diff
@@ -1060,16 +1076,26 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
...
+  ) -> Option<Expression<'ast>> {
     // rewrite `import.meta.ROLLUP_FILE_URL_<referenceId>`
     if let Some(reference_id) = property_name.strip_prefix("ROLLUP_FILE_URL_") {
       // compute relative path from chunk to asset
       let Ok(asset_file_name) = self.ctx.file_emitter.get_file_name(reference_id) else {
+        // Keep the span of the first access, so the diagnostic can point at the source.
+        self
+          .missing_file_reference_ids
+          .borrow_mut()
+          .entry(CompactStr::new(reference_id))
```
hyfdev, line 1090: This treats valid `import.meta.ROLLUP_FILE_URL_OBJ_<referenceId>` accesses as missing files. The existing `strip_prefix("ROLLUP_FILE_URL_")` parses them as `OBJ_<referenceId>`, so this branch reports `FILE_NOT_FOUND` for a valid emitted-file id.

I verified the regression by unignoring the two `resolve-file-url-obj` Rollup tests: both ES and CJS pass on the base commit and fail on this head with `unknown file "OBJ_..."`. Rollup checks `ROLLUP_FILE_URL_OBJ_` before `ROLLUP_FILE_URL_`.

Please handle the object prefix first, ideally returning the `URL` object without `.href`, or at minimum keep it out of this error path instead of reclassifying the passing tests as unsupported.

crates/rolldown/src/module_finalizers/mod.rs
```diff
@@ -97,6 +99,20 @@ pub struct ScopeHoistingFinalizer<'me, 'ast: 'me> {
...
+  /// through to `walk_expression`, which re-dispatches the very same node into
+  /// `visit_member_expression`, where the identical lookup is attempted again. Pushing
+  /// `BuildDiagnostic`s straight into a `Vec` here would report each unknown reference id twice.
+  ///
+  /// Deduplicating also makes a reference id that is accessed several times report once, matching
+  /// Rollup, which throws on the first access it renders.
+  ///
+  /// `RefCell` because the rewrite is reached through `&self` visitor paths that are already
+  /// borrowing the node being replaced.
+  pub missing_file_reference_ids: RefCell<FxIndexMap<CompactStr, Span>>,
```
hyfdev: Could we avoid `RefCell` here? All call sites originate from `VisitMut`, so the receiver chain can be mutable. I verified that changing `try_rewrite_import_meta_prop_expr`, `rewrite_rollup_file_url`, and `try_rewrite_member_expr` to `&mut self`, and storing an `FxIndexMap` directly, passes both `cargo check -p rolldown` and `just build-rolldown`.

The duplicate dispatch still warrants deduplication, but it does not require interior mutability.

crates/rolldown/src/module_finalizers/mod.rs
sapphi-red: I don't think this is a regression. `ROLLUP_FILE_URL_OBJ_<refId>` was never supported in Rolldown. I think it's more of a fix here as previously it was silently ignoring the problem.

crates/rolldown/src/module_finalizers/mod.rs
sapphi-red: Good catch. I've made that refactor.

hyfdev (approved):

PR #10222, opened 2026-07-10
fix: add EMPTY_IMPORT_META warning for non-node CJS output

sapphi-red:
The `EMPTY_IMPORT_META` warning was no emitted for non-node CJS output. This PR fixes that.

### DIFF
--- a/crates/rolldown/src/module_finalizers/finalizer_context.rs
+++ b/crates/rolldown/src/module_finalizers/finalizer_context.rs
@@ -8,7 +8,7 @@ use rolldown_common::{
 pub type FinalizerMutableFields = (
   FxIndexMap<ImportRecordIdx, String>, // transferred_import_record
   RenderedConcatenatedModuleParts,     // rendered_concatenated_wrapped_module_parts
-  Vec<BuildDiagnostic>,                // errors
+  Vec<BuildDiagnostic>,                // diagnostics
 );
 
 use oxc::ast_visit::VisitMut as _;
@@ -147,26 +147,40 @@ impl<'me> ScopeHoistingFinalizerContext<'me> {
         json_module_inlined_prop: need_inline_json_prop.then(|| Box::new(FxHashMap::default())),
         missing_file_reference_ids: FxIndexMap::default(),
         resolve_file_url_errors: Vec::new(),
+        surviving_import_meta_spans: FxIndexMap::default(),
       };
       finalizer.visit_program(oxc_program);
 
-      let missing_file_reference_ids = finalizer.missing_file_reference_ids;
-      let mut errors: Vec<BuildDiagnostic> = if missing_file_reference_ids.is_empty() {
-        vec![]
-      } else {
+      let mut diagnostics = {
         let module = finalizer.ctx.module;
-        missing_file_reference_ids
-          .into_iter()
-          .map(|(reference_id, span)| {
-            BuildDiagnostic::file_not_found(
-              reference_id.as_str(),
+        finalizer
+          .surviving_import_meta_spans
+          .iter()
+          .map(|(span, kind)| {
+            BuildDiagnostic::empty_import_meta(
               module.id.to_string(),
               module.ecma_view.source.clone(),
-              span,
+              *span,
+              finalizer.ctx.options.format.as_str().into(),
+              *kind,
             )
+            .with_severity_warning()
           })
-          .collect()
+          .collect::<Vec<_>>()
       };
+
+      let missing_file_reference_ids = finalizer.missing_file_reference_ids;
+      if !missing_file_reference_ids.is_empty() {
+        let module = finalizer.ctx.module;
+        diagnostics.extend(missing_file_reference_ids.into_iter().map(|(reference_id, span)| {
+          BuildDiagnostic::file_not_found(
+            reference_id.as_str(),
+            module.id.to_string(),
+            module.ecma_view.source.clone(),
+            span,
+          )
+        }));
+      }
 
       let mut resolve_file_url_errors = finalizer.resolve_file_url_errors;
       if !resolve_file_url_errors.is_empty() {
@@ -175,7 +189,7 @@ impl<'me> ScopeHoistingFinalizerContext<'me> {
         resolve_file_url_errors.sort_unstable();
         resolve_file_url_errors.dedup();
         // Attribute each failure to its plugin, so the user sees `[plugin foo] ...`
-        errors.extend(resolve_file_url_errors.into_iter().map(|(plugin_name, message)| {
+        diagnostics.extend(resolve_file_url_errors.into_iter().map(|(plugin_name, message)| {
           BuildDiagnostic::plugin_error(CausedPlugin::new(plugin_name), anyhow::anyhow!(message))
         }));
       }
@@ -183,7 +197,7 @@ impl<'me> ScopeHoistingFinalizerContext<'me> {
       (
         finalizer.transferred_import_record,
         finalizer.rendered_concatenated_wrapped_module_parts,
-        errors,
+        diagnostics,
       )
     })
   }
--- a/crates/rolldown/src/module_finalizers/impl_visit_mut.rs
+++ b/crates/rolldown/src/module_finalizers/impl_visit_mut.rs
@@ -18,6 +18,7 @@ use rolldown_ecmascript_utils::{
   EsmWrapperBodyKind, EsmWrapperCallKind, EsmWrapperDeclKind, EsmWrapperStmtOptions, ExpressionExt,
   JsxExt, JsxMemberExpressionObjectExt,
 };
+use rolldown_error::EmptyImportMetaKind;
 
 use crate::module_finalizers::{KeepNameId, ModuleWrapperMode, TraverseState};
 
@@ -495,6 +496,7 @@ impl<'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'_, 'ast> {
           && meta.meta.name == "import"
           && meta.property.name == "meta"
         {
+          self.record_surviving_import_meta(meta.span, EmptyImportMetaKind::Plain);
           *expr = ast::Expression::new_object_expression(
             SPAN,
             oxc::allocator::Vec::new_in(&self.ast_factory),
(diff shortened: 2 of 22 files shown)


sapphi-red:
* **#10222** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10222?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10222?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10251** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10251?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10222).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown/src/module_finalizers/impl_visit_mut.rs
```diff
@@ -487,6 +487,7 @@ impl<'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'_, 'ast> {
           && meta.meta.name == "import"
           && meta.property.name == "meta"
         {
+          self.record_surviving_import_meta(meta.span, false);
```
hyfdev: This loses the URL-specific diagnostic context for computed static accesses. `import.meta["url"]`, `import.meta[\`url\`]`, and `import.meta?.["url"]` reach this bare `MetaProperty` branch and are recorded with `is_import_meta_url = false`, so the warning no longer includes the polyfill guidance.

I reproduced this with IIFE output: the parent PR's scanner-based warning includes the `import.meta.url` help for all three forms, while this PR still emits one warning but drops the help text. Could we preserve `static_property_name() == "url"` before walking into the bare meta property, and add a computed-access regression test?

crates/rolldown/src/stages/generate_stage/finalize_modules.rs
```diff
@@ -62,35 +62,39 @@ impl GenerateStage<'_> {
...
       })
       .collect::<Vec<_>>();
 
-    if !errors.is_empty() {
+    self.link_output.diagnostics.extend(diagnostics);
+
+    if self.link_output.diagnostics.has_errors() {
+      let errors = self.link_output.diagnostics.extract_errors();
       Err(errors)?;
     }
```
sapphi-red: I'm not sure if this is the way supposed to be written.

crates/rolldown/src/module_finalizers/impl_visit_mut.rs
sapphi-red: This is now fixed 👍

hyfdev (approved):
Verified the finalizer-based warning flow and the computed/optional `import.meta.url` regression cases. LGTM.

crates/rolldown/src/stages/generate_stage/finalize_modules.rs
hyfdev: @IWANABETHATGUY cc

sapphi-red:
### Merge activity

* **Jul 17, 10:33 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 17, 1:39 PM UTC**: `sapphi-red` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 17, 1:45 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

crates/rolldown/src/module_finalizers/finalizer_context.rs
```diff
@@ -147,43 +147,61 @@ impl<'me> ScopeHoistingFinalizerContext<'me> {
...
-              span,
+              *span,
+              finalizer.ctx.options.format.as_str().into(),
+              if *is_import_meta_url {
+                EmptyImportMetaKind::Url
+              } else {
+                EmptyImportMetaKind::Plain
+              },
             )
+            .with_severity_warning()
```
hyfdev, line 167: Default `import.meta.ROLLUP_FILE_URL_*` rewrites on `iife`/`umd` already get an `EMPTY_IMPORT_META` (`EmptyImportMetaKind::RolldownFileUrl`) from `resolve_file_urls` before finalization. This path then emits a second `EMPTY_IMPORT_META` for the same access, because the generated `import.meta.url` keeps `original_expr_span` and is recorded as surviving when it collapses to `{}`.

Node CI fails on that double count on this head:

- `plugin/resolve-file-url/iife-fallback-warns` — expected 1, got 2
- `plugin/resolve-file-url/umd-fallback-warns` — expected 1, got 2
- `plugin/resolve-file-url/computed-key-iife-warns` — expected 1, got 2
- `misc/rollup-file-url-non-esm-format` — expected 2, got 4

Each FILE_URL access that falls back to `new URL(..., import.meta.url).href` needs a single warning. One ownership model is to drop the pre-warn in `resolve_file_urls` and emit `EmptyImportMetaKind::RolldownFileUrl` from the finalizer when that generated meta survives; keeping `resolve_file_urls` as the sole owner for `iife`/`umd` while skipping the finalizer record for that generated meta is also fine if non-node CJS FILE_URL still warns.

This blocks merge.

crates/rolldown/src/module_finalizers/impl_visit_mut.rs
```diff
@@ -495,6 +495,7 @@ impl<'ast> VisitMut<'ast> for ScopeHoistingFinalizer<'_, 'ast> {
...
```
hyfdev: When `resolveFileUrl` injects code such as `import.meta.url`, `parse_injected_expression` zeros every span, so this records `0..0` and the warning has no real code frame. Default FILE_URL rewrites deliberately carry `original_expr_span`; the inject path does not.

This does not block this review. Carrying the original FILE_URL span onto the injected expression (or recording that span when the inject contains unpolyfillable `import.meta`) would make the diagnostic point at the user's reference.

hyfdev (changes requested):
Moving `EMPTY_IMPORT_META` into the finalizer is the right fix for non-node CJS: warn when rewrite actually replaces `import.meta` with `{}`, rather than guessing from format alone in the scanner. The new non-node and ESM fixtures look good, and the focused Rust fixtures pass on this head.

The remaining problem is ownership of the warning for default `ROLLUP_FILE_URL_*` rewrites on `iife`/`umd`: `resolve_file_urls` still emits `RolldownFileUrl`, and the finalizer emits again for the collapsed generated `import.meta.url`. That doubles the log count and fails the Node fixtures listed in the inline comment.

— AI review by grok-4.5

crates/rolldown/src/stages/generate_stage/finalize_modules.rs
hyfdev: I think this part is fine.

crates/rolldown/src/module_finalizers/finalizer_context.rs
sapphi-red: This is now fixed.

crates/rolldown/src/module_finalizers/impl_visit_mut.rs
sapphi-red: I fixed the span so that the error code block shows the `import.meta.ROLLDOWN_FILE_URL_*` part. I didn't add a specific warning message as I didn't find a robust way to do that and since this is an edge case.

crates/rolldown/src/module_finalizers/mod.rs
```diff
@@ -1073,6 +1100,16 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
...
       let relative_asset_path = &self.ctx.chunk.relative_path_for(&absolute_asset_file_name);
 
+      let is_node_cjs = matches!(
+        (self.ctx.options.platform, &self.ctx.options.format),
+        (Platform::Node, OutputFormat::Cjs)
+      );
+      if !self.ctx.options.format.keep_esm_import_export_syntax() && !is_node_cjs {
+        // Record the origin before walking the generated `import.meta.url`. The generic URL
+        // handler reaches the same span later, and first-insert-wins preserves this richer kind.
+        self.record_surviving_import_meta(original_expr_span, EmptyImportMetaKind::RolldownFileUrl);
```
hyfdev, line 1112: This branch is the FILE_URL path the PR mainly targets for non-node CJS, but the fixtures only pin node CJS (no warning) and iife/umd (one warning). A `{ format: 'cjs', platform: 'neutral' }` FILE_URL case would lock the single-warning / `{}.url` behavior for that format as well.

This does not block this approval.

hyfdev (approved):
The double-count path is gone: `resolve_file_urls` no longer pre-warns, and the finalizer owns a single `EMPTY_IMPORT_META` for default FILE_URL rewrites by recording `RolldownFileUrl` on the original span and reusing that span on the generated `import.meta`. Non-node CJS plain / optional / computed accesses warn correctly, and node CJS polyfills stay silent.

I left one non-blocking note about covering non-node CJS FILE_URL with a fixture.

— AI review by grok-4.5

crates/rolldown/src/module_finalizers/finalizer_context.rs
hyfdev: Verified on this head: `resolve_file_urls` no longer emits, and the finalizer's single-owner + shared-span path keeps one `EMPTY_IMPORT_META`. The previously failing Node fixtures (`iife-fallback-warns`, `umd-fallback-warns`, `computed-key-iife-warns`, `rollup-file-url-non-esm-format`) pass.

— AI review by grok-4.5

crates/rolldown/src/module_finalizers/mod.rs
sapphi-red: Added a test for this. Also extracted the function for the condition so that it's always synced.

PR #10224, opened 2026-07-11
refactor: extract external import symbol merging into a method

IWANABETHATGUY:
### Description

Behavior-neutral extraction: move the chunk-level "merge external import namespaces" pass out of `generate_chunks` into a dedicated `merge_external_import_symbols` method.

The stacked PR for #3427 adds a second chunk-level merge pass right next to this one (deduplicating re-exported external named imports), so this PR gives that logic a home first. No output changes — zero snapshot drift.

PR #10225, opened 2026-07-11
fix: deduplicate re-exported external named imports within chunks

### LINKED ISSUE #3427: Polish external symbol merging.
https://github.com/rolldown/rolldown/pull/3405 fixed a bug related to external symbol merging. But the output seems sub-optimal since it blocks the nuxt integration, merge it for now and investigate it later.

IWANABETHATGUY:
### Description

close #3427

Supersedes #8218.

#### Root cause

#3405 stopped merging external named imports whose local binding is re-exported (`export { a } from 'ext'`), because eager merging happens before chunks exist and can move a symbol's canonical ref where a chunk can't render it: in ESM output a chunk renders external imports only from its own modules' `named_imports` (`direct_imports_from_external_modules`), and `compute_chunk_imports` intentionally skips external-owned symbols. That fixed the Nuxt crash, but it left every re-export of the same `(external module, name)` pair as its own symbol:

```js
// tests/rolldown/function/external/export_external, before
import { a, a as a$1, a as a1, b, b as b$1, b as b1 } from "external";
```

#### Fix

Defer the merge until chunk assignment is known, then merge strictly per chunk (next to the existing namespace merger in `merge_external_import_symbols`):

- `bind_imports_and_exports` now collects re-exported external named imports into `deferred_external_binding_merges`, keyed by `(external module, imported name)`, instead of skipping them.
- At chunk level, each group merges within its chunk. The canonical is the first used symbol by `(exec_order, symbol index)`, preferring a member whose name equals the imported name so the output reads `import { a }` rather than `import { a as a$1 }`.
- The facade created by the eager merger (for plain, non-re-exported imports) is folded into that canonical only when every module owning those plain imports lives in the same chunk. A plain import is only referenced from its own module, so in that case all of the facade's references stay inside the chunk; otherwise the facade stays a root and each chunk keeps rendering its own import.

The per-chunk restriction preserves #3405's constraint by construction: a merged group's canonical always stays a symbol declared inside that chunk, so cross-chunk references keep flowing through the regular cross-chunk import/export machinery. (This is also where #8218 went wrong: it linked re-export symbols from every chunk into the external-owned facade, which reintroduces the #3405 scenario for indirect importers.)

```js
// tests/rolldown/function/external/export_external, after
import { a, b } from "external";
```

#### Tests

Two node-executed fixtures assert runtime identity of the merged bindings:

- `external/reexport_external_binding_dedup`: single chunk; plain import + re-export + pass-through re-export (named and default) collapse to one specifier each.
- `external/splitting_reexport_external_dedup`: the #3405 splitting shape; the shared chunk collapses its duplicate re-exports into one import/one cross-chunk export, while the entry chunks' facade correctly does *not* fold across chunks.

Rollup compat suite: 1214 passed / 0 failed.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/code_splitting.rs
@@ -248,6 +248,13 @@ impl GenerateStage<'_> {
   /// stays inside the chunk that declares its members, so every chunk keeps rendering its own
   /// `import ... from 'ext'` statement and cross-chunk references keep flowing through the
   /// regular cross-chunk import/export machinery (the constraint behind #3405).
+  ///
+  /// Two kinds of groups are handled:
+  /// - namespace imports (`import * as ns from 'ext'`), collected in
+  ///   `external_import_namespace_merger`;
+  /// - named imports whose local binding is re-exported (`export { a } from 'ext'`), collected in
+  ///   `deferred_external_binding_merges` because `bind_imports_and_exports` cannot merge them
+  ///   eagerly like plain named imports (#3427).
   fn merge_external_import_symbols(
     &mut self,
     chunk_graph: &ChunkGraph,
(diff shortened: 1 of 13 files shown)


IWANABETHATGUY:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10225?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#10225** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10225?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10225?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10224** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10224?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10225).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

IWANABETHATGUY:
@codex[agent]

PR #10227, opened 2026-07-11
fix: deterministic emission order under preserveModules

pjsny:
## What this fixes

Under `preserveModules: true` (a.k.a. tsdown's `unbundle: true`), the leading `import { … } from "./sibling.mjs";` sequence at the top of each emitted chunk permuted across repeated builds when a plugin `await`ed inside its `resolveId` hook — same source, same lockfile, different bytes.

Same source, same lockfile, empirical drift observed in production against a real ~60-module SDK graph on `tsdown@0.22.4 + rolldown@1.1.5` under `bunx turbo run build --filter=./packages/*` (4-way parallel). 60 sequential parallel builds produced 5 unique variants of `dist/`. After this fix, 60/60 identical.

## The race, top to bottom

The bug lives inside rolldown but was triggered by any plugin that `await`s in `resolveId` — for example [tsdown's `DepsPlugin.resolveId`](https://github.com/rolldown/tsdown/blob/v0.22.4/src/features/deps.ts) which does `await this.resolve(...)` plus optionally `await resolveDepSubpath(...)` on every non-relative specifier.

1. **`try_spawn_new_task`** (`crates/rolldown/src/module_loader/module_loader.rs`) mints `ModuleIdx` via `alloc_ecma_module_idx()` as tasks arrive. The outer user-defined-entry loop calls `try_spawn_new_task` synchronously per entry, but `tokio::spawn`s the resolution task. On a multi-worker runtime those spawned tasks call further `try_spawn_new_task` for transitive imports and race with the outer loop's remaining iterations. **Different runs → different `ModuleIdx` assignments for the same set of entries.**

2. **`LinkStage::new`** (`crates/rolldown/src/stages/link_stage/mod.rs`) had a comment claiming *"We need to preserve the original order of user defined entry points"* and sorted only non-user-defined entries. The intent was to preserve the user's `input` array positions, but `scan_stage_output.entry_points` is actually in `ModuleIdx` allocation order — which is racy from step 1. The claimed contract was already broken.

3. **`sort_modules`** (`crates/rolldown/src/stages/link_stage/sort_modules.rs`) DFSes from `self.entries.keys()` and assigns `exec_order` monotonically. Racy entry ordering → racy `exec_order` for the same source module.

4. **`compute_cross_chunk_links`** (`crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs`) correctly sorts `imports_from_other_chunks` by importee `exec_order` — but the sort key is now unstable.

5. **`render_esm_chunk_imports`** (`crates/rolldown/src/ecmascript/format/esm.rs`) iterates the sorted `FxIndexMap` at emit time. Emit order inherits the race.

### Debug traces

I instrumented `render_esm_chunk_imports` and `sort_modules` in a debug build; the divergent runs' `dist/internal/kernel.mjs` sections showed:

```
Run A  (kernel.mjs hash 7eb8490b…, kernel.mjs exec_order=38, 11 chunk imports)
  importee=util/log.mjs                                     exec_order=1  items=1
  importee=util/global.mjs                                  exec_order=13 items=1
  importee=util/stringify.mjs                               exec_order=14 items=1
  importee=internal/page-lifecycle.mjs                      exec_order=20 items=1
  importee=internal/config.mjs                              exec_order=24 items=2
  …
Run B  (kernel.mjs hash c99e189e…, kernel.mjs exec_order=35, 11 chunk imports)
  importee=util/log.mjs                                     exec_order=1  items=1
  importee=internal/page-lifecycle.mjs                      exec_order=12 items=1
  importee=util/global.mjs                                  exec_order=16 items=1
  importee=internal/config.mjs                              exec_order=17 items=2
  importee=util/stringify.mjs                               exec_order=27 items=1
  …
```

Same source, same 11 importees, different `exec_order` values → different iteration order → different leading `import{…}from"./…";` sequence in `kernel.mjs`.

Instrumenting `sort_modules` showed the entry iteration order was consistent (user-array order) across runs — but the `ModuleIdx` values assigned to transitively-discovered entries shifted, and that shift propagated to `exec_order` via `.import_records()` DFS.

### Observed drift shape

Two adjacent variants of the same chunk:

```diff
- import{createLogger}from"../util/log.mjs";import{onPageHidden}from"./page-lifecycle.mjs";
- import{getGlobal}from"../util/global.mjs";import{resolveTargets,setIngestHost}from"./config.mjs";
- import{SessionTracker}from"../tracking/api.mjs";import{safeStringify}from"../util/stringify.mjs";
+ import{createLogger}from"../util/log.mjs";import{safeStringify}from"../util/stringify.mjs";
+ import{onPageHidden}from"./page-lifecycle.mjs";import{getGlobal}from"../util/global.mjs";
+ import{resolveTargets,setIngestHost}from"./config.mjs";import{SessionTracker}from"../tracking/api.mjs";
```

Bodies after the leading imports are byte-identical. Only the run of `import{…}from"./sibling.mjs";` statements at the top permutes.

## The fix

Sort every entry in `scan_stage_output.entry_points` by a stable `(kind, id)` key — extending the pattern that was already applied to non-user-defined entries.

```diff
-    // We need to preserve the original order of user defined entry points.
-    let mut rest = scan_stage_output
-      .entry_points
-      .extract_if(0.., |item| !matches!(item.kind, EntryPointKind::UserDefined))
-      .collect_vec();
-    rest.sort_by_cached_key(|item| {
+    scan_stage_output.entry_points.sort_by_cached_key(|item| {
       (item.kind, scan_stage_output.module_table.modules[item.idx].id().as_str())
     });
-    scan_stage_output.entry_points.extend(rest);
```

Downstream, `exec_order` becomes deterministic → chunk `exec_order` becomes deterministic → the sort already in `compute_cross_chunk_links` produces a stable `imports_from_other_chunks` iteration → deterministic emit order for each chunk's leading imports.

## Consequence for users

`input: ["b.ts", "a.ts"]` now emits `a.ts, b.ts` instead of `b.ts, a.ts`. Users who rely on positional array semantics (e.g. output filename keyed to `input[0]`) may notice.

The comment removed above wanted to preserve user-array order, but the ordering it actually preserved was racy `ModuleIdx` allocation order — the intended contract was already broken.

**Alternative that fully preserves user-array order** — happy to swap to this if reviewers prefer:

Attach an `input_index: u32` to each `EntryPoint` when it's first inserted at `module_loader.rs:414`, then sort by `(kind, input_index)`. Non-user-defined entries would keep the current `(kind, id)` fallback since they have no input position. That path is more invasive (touches `rolldown_common::types::entry_point`, both `EntryPoint` construction sites, and this sort call) but preserves the original semantic intent.

## Snapshot change

One integration-test snapshot at `crates/rolldown/tests/rolldown/issues/9993/artifacts.snap` moved — the test fixture's emit order shifted, causing rolldown's base-54 name generator to assign different letters (`n` vs `t`) to two `require_*` symbols. Test's runtime assertions still pass; the snapshot was auto-updated by `just test-rust`.

## Testing

`just lint-rust` and `just test-rust` clean locally (1806 passed, 0 failed).

**I did not add an in-tree stability test in this PR.** I tried both:

- A snapshot fixture under `crates/rolldown/tests/rolldown/issues/<n>/` — but the artifacts snapshotter sorts assets alphabetically by filename before rendering (`build_artifacts_snapshot.rs:33`), so it can't detect emit-order drift inside a chunk.
- A `packages/rolldown/tests/stability/`-style test mirroring `issue-3453` — a fixture with 20 entries + 30 shared leaves + a plugin doing `await this.resolve` + `await new Promise(setImmediate)` per non-relative specifier, run 12× in parallel worker processes. It passed both with and without the fix — the sequential-single-process WASI-safe pattern used by `issue-3453` doesn't produce enough concurrent resolver pressure to trigger the race at synthetic-fixture scale.

The production shape that does trigger the race is the full `bunx tsdown` plugin stack (`DepsPlugin` + `dtsPlugin` + `NodeProtocolPlugin` + others, each with its own await points) against a ~60-module graph run in parallel via `bunx turbo`. Available on request as a sanitized repro. Happy to iterate on a fixture that reproduces at synthetic scale if there's a preferred angle — for example, a rolldown-native regression test that patches the module loader to inject deterministic completion-order variance rather than trying to synthesize it through plugin awaits.

## AI disclosure

Investigation, tracing, and this write-up were done in collaboration with Claude (Anthropic). All code, verification steps, and empirical claims have been reproduced and reviewed by me.

### DIFF
--- a/crates/rolldown/src/stages/link_stage/mod.rs
+++ b/crates/rolldown/src/stages/link_stage/mod.rs
@@ -1,13 +1,12 @@
 use arcstr::ArcStr;
-use itertools::Itertools;
 use oxc::span::Span;
 use oxc_index::IndexVec;
 #[cfg(debug_assertions)]
 use rolldown_common::common_debug_symbol_ref;
 use rolldown_common::{
-  ConstExportMeta, DependedRuntimeHelperMap, EntryPoint, EntryPointKind, FlatOptions, ImportKind,
-  ModuleIdx, ModuleTable, PreserveEntrySignatures, RetainedExportSymbols, RuntimeModuleBrief,
-  SymbolRef, SymbolRefDb, UsedExternalSymbols, UsedSymbolRefsBuilder,
+  ConstExportMeta, DependedRuntimeHelperMap, EntryPoint, FlatOptions, ImportKind, ModuleIdx,
+  ModuleTable, PreserveEntrySignatures, RetainedExportSymbols, RuntimeModuleBrief, SymbolRef,
+  SymbolRefDb, UsedExternalSymbols, UsedSymbolRefsBuilder,
   dynamic_import_usage::DynamicImportExportsUsage,
 };
 use rolldown_error::Diagnostics;
@@ -145,17 +144,12 @@ impl<'a> LinkStage<'a> {
       FxHashMap::default()
     };
 
-    // We need to preserve the original order of user defined entry points.
-    let mut rest = scan_stage_output
-      .entry_points
-      .extract_if(0.., |item| !matches!(item.kind, EntryPointKind::UserDefined))
-      .collect_vec();
-
-    rest.sort_by_cached_key(|item| {
+    // Canonicalize all entry_points by (kind, id). User-defined entries were
+    // previously left in ModuleIdx allocation order, which is racy under any
+    // plugin that awaits in resolveId (e.g. tsdown DepsPlugin).
+    scan_stage_output.entry_points.sort_by_cached_key(|item| {
       (item.kind, scan_stage_output.module_table.modules[item.idx].id().as_str())
     });
-
-    scan_stage_output.entry_points.extend(rest);
 
     Self {
       sorted_modules: Vec::new(),
--- a/crates/rolldown/tests/rolldown/issues/9993/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/issues/9993/artifacts.snap
@@ -6,7 +6,7 @@ source: crates/rolldown_testing/src/integration_test.rs
 ## entry-1.js
 
 ```js
-import { n as require_node4 } from "./v.js";
+import { t as require_node4 } from "./v.js";
 export default require_node4();
 
 ```
@@ -15,7 +15,7 @@ export default require_node4();
 
 ```js
 import { n as __esmMin, r as __exportAll } from "./rolldown-runtime.js";
-import { t as require_node3 } from "./v.js";
+import { n as require_node3 } from "./v.js";
 //#region node5.js
 var node5_exports = /* @__PURE__ */ __exportAll({ node_5: () => node_5 });
 var node_5;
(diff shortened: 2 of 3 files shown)


pjsny:
I can share the repro if needed

IWANABETHATGUY:
> I can share the repro if needed

Please give me the minimum reproduction repo

PR #10235, opened 2026-07-12
fix: treat rooted drive-less module ids as absolute in preserveModules naming

### LINKED ISSUE #10186: [Bug]: `preserveModules` produces an invalid `[name]` for file path looking ids on Windows

[REPL](https://repl.rolldown.rs/#eNqVUk1rwzAM/SvCDLJCcSnslC6DsQ/ooaeWXZaxZrXSeUvs4NhdIeS/T7aXdp+HmeBE0pOe9JSOlSztmFQC9/yl9d+KpUd7zDZkTiawkMZo04J9RrhzCMvbK9DONs5CqQ2cy3oLrdlkOZuUxU5utOJ05eyC54rSV5TWVG4rFdBjdFUJ/aY4wUq55bYFg62udhgJpKAsWTfaWPCloDS6huRz5WSWK9wHBHla68khC2iKUN/IUmsc9mP2k20Y85dIHDiWvFwub1aP82uq+xe3wLJwlYUuV0CTkRwpxMZ5jbbgwTX2wTh9m8K9tyBm+KOKGlNIDDZGJwHqz4cec3EqxegIJpISyAVZlh36GxHaOkPSihmQ2K+ITdCxeKIiznpBoYjS1lq4yjuGiv2BstLFv9jW3zRITroB0ydr0ugrQXw/hDv+OenA1NC0aHa4CM2RRn5zAeiTer/Qnha5o81M+ZSfsf4d6EvfZw==)

**Note that this only happens on Windows, you need to download the project and run it locally to reproduce**

### What is expected?

No error happens like Rollup

The following error happens
```
[INVALID_OPTION] Invalid substitution "/favicon" for placeholder "[name]" in "entryFileNames" pattern, can be neither absolute nor relative paths.
```

### System Info

```Shell
REPL 1.1.4
```

### Any additional comments?

The same error happens with `\\favicon.ico`. But doesn't happen with `C:\\favicon.ico`.

Originally reported at https://github.com/vitejs/vite/issues/22889

IWANABETHATGUY:
## Summary

Fixes #10186.

On Windows, Node (and therefore Rollup) treat a rooted-but-drive-less id such
as `/favicon` or `\favicon` as absolute — it's implicitly anchored to the
current drive. Rust's `Path::is_absolute()` disagrees because the path has no
drive prefix, so `preserveModules` path math (`preserveModulesRoot` stripping,
relativizing against the input base) mishandled these ids and could crash on
the `[name]` placeholder.

`node_style_absolute` now anchors such an id to `cwd`'s drive so downstream
path math treats it like any other absolute path.

## Commits

- **fix:** the actual #10186 fix, with two integration fixtures.
- **refactor:** small follow-up cleanup — `cwd.components().next()` is an
  `Option`, so `?` handles the empty-`cwd` case directly and the `let … else`
  only has to check the `Prefix` variant. Same behavior.

## Test plan

- New fixtures under `packages/rolldown-tests/.../preserve-modules/issue-10186*`
  (RED without the fix, GREEN with it).
- `cargo check -p rolldown_std_utils` passes.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/code_splitting.rs
@@ -1,4 +1,4 @@
-use std::{cmp::Ordering, collections::VecDeque, path::Path};
+use std::{borrow::Cow, cmp::Ordering, collections::VecDeque, path::Path};
 
 use crate::{
   chunk_graph::ChunkGraph,
@@ -11,15 +11,18 @@ use itertools::Itertools;
 use oxc_index::{IndexVec, index_vec};
 use rolldown_common::{
   Chunk, ChunkIdx, ChunkKind, ChunkMeta, EntryPointKind, ExportsKind, ImportKind, ImportRecordIdx,
-  ImportRecordMeta, IndexModules, Module, ModuleIdx, ModuleNamespaceIncludedReason, ModuleTag,
-  ModuleTagBitSet, ModuleTagRegistry, PostChunkOptimizationOperation, PreserveEntrySignatures,
-  RetainedExportSymbols, SymbolRef, UsedSymbolRefs, UsedSymbolRefsBuilder, WrapKind,
+  ImportRecordMeta, IndexModules, Module, ModuleId, ModuleIdx, ModuleNamespaceIncludedReason,
+  ModuleTag, ModuleTagBitSet, ModuleTagRegistry, PostChunkOptimizationOperation,
+  PreserveEntrySignatures, RetainedExportSymbols, SymbolRef, UsedSymbolRefs, UsedSymbolRefsBuilder,
+  WrapKind,
 };
 use rolldown_error::BuildResult;
+use rolldown_std_utils::PathBufExt as _;
 use rolldown_utils::{
   BitSet, IndexBitSet, commondir,
   index_vec_ext::IndexVecRefExt,
   indexmap::FxIndexMap,
+  node_style_absolute,
   rayon::ParallelIterator,
   rustc_hash::{FxHashMapExt, FxHashSetExt},
 };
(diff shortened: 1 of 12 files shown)


shulaoda:
@codex review

crates/rolldown/src/stages/generate_stage/code_splitting.rs
```diff
@@ -740,14 +757,14 @@ impl GenerateStage<'_> {
...
-          item.id.is_path().then_some(item.id.as_ref())
+          common_dir_input(&item.id, &self.options.cwd)
         } else {
           None
         }
       }
       Module::External(external_module) => {
         if self.options.preserve_modules {
-          external_module.id.is_path().then_some(external_module.id.as_ref())
+          common_dir_input(&external_module.id, &self.options.cwd)
```
hyfdev: External modules should not participate in `input_base`. This branch now runs rooted-but-volume-less external IDs through `common_dir_input` on Windows. With an unnamed `C:\project\main.js` importing `/favicon`, `resolveId` returning `{ id: '/favicon', external: true }`, and `preserveModules: true`, the base emits `main.js`; this PR anchors the external ID to `C:\favicon`, lowers `input_base` to `C:\`, and emits `project/main.js`. Rollup excludes `ExternalModule` from `getAbsoluteEntryModulePaths`, so it still emits `main.js`.

Could we return `None` for `Module::External` here and add a regression fixture? External modules do not produce preserved chunks, so they should not affect preserved chunk names.

hyfdev (approved):

IWANABETHATGUY:
### Merge activity

* **Jul 14, 6:07 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 14, 7:20 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 14, 7:26 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10237, opened 2026-07-12
fix(rolldown): drop the unused runtime module after entry-level external flattening

### LINKED ISSUE #9374: [Bug]: rolldown/runtime is false-positively injected

IWANABETHATGUY:
### What is this PR solving?

Top of the #9374 stack (#10238 ← #10239 ← this).

For a star re-export chain that ends at an external module:

```js
// entry_a.js
export * from './entry_b.js';
export const a = 'a';

// entry_b.js
export * from 'external';
export const b = 'b';
```

both entry chunks emit a literal `export * from "external"` and never call a runtime helper — yet the build still ships a dead `rolldown-runtime-*.js` chunk that exports `__reExport`/`__exportAll` to nobody, with a bare import of it in every entry.

Root cause: tree-shaking includes the runtime helpers at link time, when `entry_b` still counts as having dynamic exports. The generate stage later proves the whole chain can be represented statically (`find_entry_level_external_module` marks the records `EntryLevelExternal` and re-propagates `has_dynamic_exports = false` — including to transitive star importers since #10239; `finalized_module_namespace_ref_usage` drops the namespace objects that only served the chain), so the module finalizer emits no helper call — but by then the runtime module is already included and already placed in a chunk, and nothing un-includes it.

### The fix

Add `sweep_unused_runtime_module` (new pass at the tail of `generate_chunks()`, after the walk-back passes and before chunk exec-order assignment). It re-derives runtime helper demand from the same post-walk-back facts the finalizer renders from, through four channels: per-module `depended_runtime_helper` flags (with `ReExport` discounted unless some included star importee still has `has_dynamic_exports` — CommonJS importees always do, so wrapped-CJS star re-exports are unaffected), the namespace-object channel gated on `namespace_included` (calling the `LinkingMetadata::ns_star_external_re_export_emitted` rule extracted in #10238, so the prediction cannot drift from the emission), runtime-owned symbols referenced by included statements, and entry-chunk synthesized references.

If zero demand remains, the runtime module is un-included, its symbols are purged from `used_symbol_refs` (which severs every cross-chunk edge — all downstream consumers are liveness-filtered), and its now-empty chunk is tombstoned via the same `PostChunkOptimizationOperation::Removed` lifecycle the chunk optimizer uses. Any surviving demand, or any bail-out (tree-shaking disabled, side-effectful dev/HMR runtime), leaves everything exactly as tree-shaking decided.

### What other alternatives have you explored?

Teaching the link stage to skip the eager helper inclusion for the statically decidable cases was rejected: the flattening decision depends on chunk shapes (dynamic entries, `SimulateFacadeChunk`), so a link-time copy of the predicate would have to stay in sync with the generate stage forever, and drift in the optimistic direction produces broken output (a rendered helper that was never included) rather than dead bytes. Recomputing demand after the walk-back, where all facts are final, fails conservative instead.

### Effects on existing tests

Six snapshots change, all in the fixed direction: `6992`, `7115`, `7233`, `7233_chain` lose their dead runtime files and the bare imports/requires of them (the same bug shape under `preserveModules`, in both ESM and CJS output); `4472`/`5923` lose the last phantom `//#region` markers (the bulk of the marker cleanup happened in #10239).

### Tests

New fixture `crates/rolldown/tests/rolldown/issues/9374/` — fails on this PR's base (#10239's head still ships the runtime chunk; verified) and passes here. It uses `minify: false` so vestigial namespace declarations would also surface in the snapshot. Verified the legitimate cases still keep the runtime: `import * as ns` observation over the same chain (helpers are genuinely used to merge the external's exports into the namespace), and star re-exports of CJS modules (`__commonJSMin` wrapper kept).

`internal-docs/code-splitting/implementation.md` is updated with the new pipeline ordering and the sweep's liveness invariants.

fixes #9374

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/code_splitting.rs
@@ -165,6 +165,15 @@ impl GenerateStage<'_> {
 
     chunk_graph.sort_chunk_modules(self.link_output, self.options);
 
+    self.find_entry_level_external_module(&mut chunk_graph);
+
+    self.finalized_module_namespace_ref_usage();
+
+    // Runs after the two passes above invalidated the link-time reasons for including the
+    // runtime module, and before chunk exec-order assignment so a dropped runtime chunk is
+    // tombstoned like any other removed chunk.
+    self.sweep_unused_runtime_module(&mut chunk_graph, used_symbol_refs);
+
     chunk_graph
       .chunk_table
       .iter_mut_enumerated()
@@ -227,8 +236,6 @@ impl GenerateStage<'_> {
     // [order by chunk index]               [order by exec order]
 
     chunk_graph.rebuild_sorted_chunk_idx_vec();
-
-    self.find_entry_level_external_module(&mut chunk_graph);
 
     Ok(chunk_graph)
   }
--- a/crates/rolldown/src/stages/generate_stage/mod.rs
+++ b/crates/rolldown/src/stages/generate_stage/mod.rs
@@ -111,6 +111,7 @@ mod minify_chunks;
 mod on_demand_wrapping;
 mod post_banner_footer;
 mod render_chunk_to_assets;
+mod runtime_module_sweep;
 
 pub struct GenerateStage<'a> {
   link_output: &'a mut LinkStageOutput,
@@ -144,8 +145,9 @@ impl<'a> GenerateStage<'a> {
     self.plugin_driver.render_start(self.options).await?;
     let mut chunk_graph = self.generate_chunks(&mut used_symbol_refs).await?;
 
-    // The chunk optimizer's re-run of the inclusion pass (inside `generate_chunks`) was the
-    // last writer; sealing consumes the builder, so nothing downstream can mutate the set.
+    // The unused-runtime-module sweep (inside `generate_chunks`, after the chunk optimizer's
+    // re-run of the inclusion pass) was the last writer; sealing consumes the builder, so
+    // nothing downstream can mutate the set.
     let used_symbol_refs = used_symbol_refs.seal();
 
     // Count only live chunks. Chunks merged away during chunk optimization (e.g.
@@ -159,8 +161,6 @@ impl<'a> GenerateStage<'a> {
     if live_chunk_count > 1 {
       validate_options_for_multi_chunk_output(self.options)?;
     }
-
-    self.finalized_module_namespace_ref_usage();
 
     self.compute_retained_export_symbols(&used_symbol_refs);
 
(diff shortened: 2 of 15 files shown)


IWANABETHATGUY:
* **#10237** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10237?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10237?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10239** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10239?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10238** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10238?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10237).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyfdev (approved):
LGTM. I performed an additional adversarial review of the runtime sweep across helper demand, namespace handling, wrappers, side effects, and cross-chunk references, and found no correctness issues. The relevant regression fixtures also pass.

IWANABETHATGUY:
### Merge activity

* **Jul 12, 4:23 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 12, 4:23 PM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 12, 4:29 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

Issue #10070, opened 2026-07-01
[Bug]: Sourcemap loses mappings for ESM modules when CJS file is present in the chunk


sapphi-red:
### Reproduction link or steps

[REPL](https://repl.rolldown.rs/#eNptUMtuhDAM/BXLp10JpWrVE4hTf6HHXFIwS1bBifLoQyj/XkFg2a16sz2eGXtmHLCekTj6H3ENS81YH32FHdaoJ2d9hBlGMsZChsHbCSSKp56c6K5BYiP5tqWcu99Rzolt5WLshzLvow5ihLbIPY4VtAu/kYwVEtbRJ8oVFo39vK0rx9H3ajsk7qK2vLBPZ5glA9zpBt0TtPDcLHNPMXmG15dGcpaMucLtkd1hb4vFZPtkSBSnIEoK7WG5+O2iEkctsYHcFGFvjentF4vO8qAvIt48/kH+xN3ToJneVvhIdOetiW7f9zSoZOID47RmoNmlWEPRFBNFJdZRtYA2xRWdIdjkO5qUq2GJHHIlOZ/XH3KFn1ijUZFCxPwLwDa++Q==)

### What is expected?

`globalThis.side = 1` and `return 42` to have a mapping.

### What is actually happening?

`globalThis.side = 1` and `return 42` does not have a mapping.
<img width="1103" height="515" alt="Image" src="https://github.com/user-attachments/assets/633c27fc-53b4-4084-860e-8b5f75154e78" />

### System Info

```Shell
REPL 1.1.3
```

hyfdev:
Fixed by #10074 — but it's a consumer-side fix, not the root fix.

**How:** the CJS require-init is prepended into a concatenated ESM module as a MagicString mutation whose map was built with the default `Hires::False` (one column-0 token per line). `collapse_sourcemaps` walks that coarse map and looks each token up in the codegen map per line; for an indented line the codegen token is after the indent, so the column-0 query misses and the line is dropped. The PR builds the mutation map with `Hires::Boundary` so its columns cover codegen's.

**Limit:** `Hires` only covers maps rolldown generates; the same drop still happens for a coarse map returned by a plugin hook. The root cause — `oxc_sourcemap`'s lookup dropping on a same-line miss instead of approximating (as Rollup's `traceSegment` does) — should be fixed in oxc.

sapphi-red:
[...]

PR #10249, opened 2026-07-13
fix(sourcemap): preserve coarse mappings during composition

hyfdev:
refs #10070.

## The behavior to review

This PR makes one choice: when a coarse outer map points to a column before the first mapping on the same line of a more detailed inner map, use that first same-line mapping instead of dropping the generated position.

Rollup makes the same choice when tracing segments. The fallback never crosses a line.

### User input

`main.js` is the source written by the user:

```js
export function app() {
  globalThis.side = 1;
  return 42;
}
```

### Generated output

Rolldown produces this chunk. Region comments and the `sourceMappingURL` line are omitted here, but the code shape is otherwise the fixture's real output:

```js
function app() {
	globalThis.side = 1;
	return 42;
}
export { app };
```

The generated JavaScript is identical before and after this PR. Only its source map changes.

### Before this PR

Asking the composed map where the two generated statements came from returns no original position:

```text
generated 3:1  globalThis.side = 1  ──X──>  null
generated 4:1  return 42             ──X──>  null
```

In a source-map visualizer, these generated statements have no colored counterpart in the Original pane. A debugger or stack-trace consumer cannot take these positions back to `main.js`.

### After this PR

The same generated positions resolve to the corresponding user-written statements:

```text
Generated output                              User input: main.js

3:1  globalThis.side = 1  ─────────────────>  2:2  globalThis.side = 1
4:1  return 42             ─────────────────>  3:2  return 42
```

Source-map consumer lines are one-based and columns are zero-based. These are the actual positions asserted by the integration fixture.

In a visualizer, each generated statement now shares a color with its corresponding statement in the Original pane. Selecting either side can identify the other side. Debuggers and stack traces can return to the correct original line; the column is the best same-line position available from the coarse map.

## Why the lookup changes

The map chain contains two different levels of detail:

```text
detailed map:  intermediate column 2 -> user source
coarse map:    final column 0        -> intermediate column 0
```

Before:

```text
final column 0
  -> intermediate column 0
  -> detailed map has no mapping at or before column 0
  -> drop the generated position
```

After:

```text
final column 0
  -> intermediate column 0
  -> detailed map's first mapping on this line is column 2
  -> use column 2 and preserve the generated position
```

Lookups between or after existing mappings still use the nearest mapping at or before the requested column. Missing and empty lines still fail. One-field explicitly unmapped segments are skipped in this PR, matching Rollup; #10254 separately preserves those boundaries.

#10074 remains necessary because it makes Rolldown-owned mutation maps more precise. This PR covers coarse external maps and other cases where the producer did not provide matching columns.

## Why the snapshots change

The snapshot visualizer writes mappings as:

```text
(original line:column) "original text" --> (generated line:column) "generated text"
```

Unlike the source-map consumer positions above, these snapshot coordinates are zero-based for both lines and columns. The three snapshots only add mappings; no existing mapping is removed or changed.

| Snapshot | Newly restored generated mapping | Previous first mapping on that generated line | Original position used |
| --- | --- | --- | --- |
| `misc/wrapped_esm` | `60:1 "("`, `60:2 "{"` | `60:3 "e} = "` | `foo.js 4:4` |
| `misc/wrapped_esm` | `61:1 "("`, `61:2 "{"` | `61:3 "h: "` | `foo.js 7:2` |
| `misc/wrapped_esm` | `73:1 "("`, `73:2 "{"` | `73:3 "destructuring} = "` | `foo.js 30:6` |
| `wrapped_esm_default_function` | `19:0 "init_foo("` | `19:9 ");"` | `bar.js 0:23` |
| `wrapped_esm_export_named_function` | `18:0 "init_foo("` | `18:9 ");"` | `bar.js 0:27` |

There are eight added rows in total: six in `misc/wrapped_esm`, one in each deconflict snapshot. Every addition follows the behavior under review:

```text
new generated column < previous first mapped column on that line
new original position = previous first mapping's original position
```

For example:

```diff
+(0:23) ";\n" --> (19:0) "init_foo("
 (0:23) ";\n" --> (19:9) ");\n"
```

The coarse map requested generated column 0, while the detailed map's first mapping was at column 9. The fallback associates column 0 with the same original position already used at column 9. It does not move or replace the existing mapping.

## Suggested review order

1. Review the lookup change and `source_id` guards in `crates/rolldown_sourcemap/src/lib.rs`.
2. Review the Rust test where a column-0 coarse token meets a column-2 detailed token.
3. Review the `renderChunk` fixture, which exercises the public plugin path and asserts the exact positions shown above.
4. Check the snapshot table against the eight added snapshot rows.

## User impact

- generated JavaScript, runtime behavior, and bundle size do not change
- users of source maps regain original locations for lines previously dropped during composition
- the change matters when a plugin or another transform provides a less column-detailed map than Rolldown's preceding map
- already resolvable mappings are unchanged

## Validation

- `cargo test -p rolldown_sourcemap`
- `just test-rust`
- `just test-node-rolldown-only fixtures-concurrent.test.ts -t plugin/render-chunk/coarse-sourcemap-composition`
- `cargo clippy -p rolldown_sourcemap --all-targets -- --deny warnings`


hyfdev:
* **#10255** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10255?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10254** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10254?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10249** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10249?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10249?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10249).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyfdev:
Blocking: #10249 must not merge independently of #10254.

The approximate lookup itself looks correct: when a coarse map points before a line's first detailed segment, clamping to that first same-line segment preserves the mapping and matches Rollup's composition behavior.

However, the new `source_id` guards drop source-less segments. Those segments are explicit unmapped boundaries, not disposable entries: a source-map mapping stays active until the next segment. For example, `AAAA,K` maps from column 0 and explicitly stops mapping at column 5. If the column-5 segment is removed, the output becomes effectively `AAAA`, so code after column 5 is incorrectly attributed to the original source.

#10254 restores those unmapped boundaries and adds coverage for both final-map and intermediate-map cases. Since it is stacked on this PR, please merge the complete stack or fold #10254 into this PR before merging #10249.

Also, #10074 already fixed the direct #10070 reproducer. This PR generalizes the fix to coarse external maps; closing #10070 is appropriate only once the unmapped-boundary fix lands as well.

crates/rolldown_sourcemap/src/lib.rs
```diff
@@ -82,13 +82,19 @@ pub fn collapse_sourcemaps(sourcemap_chain: &[&oxc_sourcemap::SourceMap<'_>]) ->
...
+          // A coarse map may point before the first token on the same line. Clamp to that first
+          // token instead of dropping the mapping, matching Rollup's composition behavior.
+          let traced = sourcemap.lookup_source_view_token_approx(
             lookup_table,
             token.get_src_line(),
             token.get_src_col(),
-          )
+          )?;
+          traced.get_source_id()?;
+          Some(traced)
```
sapphi-red, line 96:
```suggestion
let original_token =
sourcemap_and_lookup_table.iter().try_fold(token, |token, (sourcemap, lookup_table)| {
          // A coarse map may point before the first token on the same line. Clamp to that first
          // token instead of dropping the mapping, matching Rollup's composition behavior.
          sourcemap.lookup_source_view_token_approx(
lookup_table,
token.get_src_line(),
token.get_src_col(),
          )
```
These `token.get_source_id()?` calls are no-op, right?

packages/rolldown/tests/fixtures/plugin/render-chunk/coarse-sourcemap-composition/_config.ts
```diff
@@ -0,0 +1,68 @@
+import { defineTest } from 'rolldown-tests';
+import { getOutputChunk } from 'rolldown-tests/utils';
+import { SourceMapConsumer, type RawSourceMap } from 'source-map-js';
```
sapphi-red: I think we should use the modern `@jridgewell/trace-mapping`.

crates/rolldown_sourcemap/src/lib.rs
sapphi-red: Ah, no it's bailing out

crates/rolldown_sourcemap/src/lib.rs
```diff
@@ -82,13 +82,19 @@ pub fn collapse_sourcemaps(sourcemap_chain: &[&oxc_sourcemap::SourceMap<'_>]) ->
   let tokens: Box<[Token]> = last_map
     .get_source_view_tokens()
     .filter_map(|token| {
+      token.get_source_id()?;
       let original_token =
         sourcemap_and_lookup_table.iter().try_fold(token, |token, (sourcemap, lookup_table)| {
-          sourcemap.lookup_source_view_token(
+          token.get_source_id()?;
```
sapphi-red: Is this possible to be None? I guess it's covered by the `token.get_source_id()?;` above and `traced.get_source_id()?;` below.

sapphi-red (approved):

packages/rolldown/tests/fixtures/plugin/render-chunk/coarse-sourcemap-composition/_config.ts
hyfdev: Updated the regression fixtures to use `@jridgewell/trace-mapping` (`TraceMap` and `originalPositionFor`) and added it as a direct test dev dependency.

crates/rolldown_sourcemap/src/lib.rs
hyfdev: Yes. The outer guard covers the first iteration, and `traced.get_source_id()?` covers every subsequent iteration, so this check was redundant. Removed it.

sapphi-red:
### Merge activity

* **Jul 15, 6:55 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 15, 6:55 AM UTC**: `sapphi-red` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 15, 7:05 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10257, opened 2026-07-13
fix(watch): drop tsconfig-merged transform options on each rebuild

shulaoda:
Part of #9598 (rebuild when tsconfig changes), on top of #10205.

`RawTransformOptions` memoizes the transform options merged from each tsconfig for the lifetime of the bundler. The watch task already clears the resolver cache before every rebuild, so edited tsconfig files are re-read from disk, but the memoized merge results were never dropped, so a rebuild kept transforming with the old tsconfig values. Concretely: edit `useDefineForClassFields` in tsconfig.json, touch a source file to trigger a rebuild, and the output still uses the old semantics.

The fix adds `clear_cache` on `RawTransformOptions` and wires it through `Bundler::clear_transform_tsconfig_cache`, which the watch task now calls next to the existing `clear_resolver_cache`. Thanks to #10205 the tsconfig contents themselves live in the cache shared with the main resolver, so this only needs to drop the memoized merge results, not a second resolver cache.

Note that tsconfig files are not registered as watch files yet, so editing a tsconfig alone still does not trigger a rebuild. That part comes in a follow-up PR. This PR makes sure that once a rebuild is triggered, it picks up the edits.

PR #10261, opened 2026-07-13
fix(dev): force a full rebuild when a tsconfig changes

shulaoda:
Part of #9598, on top of #10258. Addresses the review comment there about the dev API.

After #10258 the dev engine watches tsconfig files, but being watched is not enough: a tsconfig is neither a module nor a transform dependency, so the HMR mapper observes the change and maps it to `Noop`, and dev output keeps the old transform and resolution results. A tsconfig edit affects every module the tsconfig governs, which neither an HMR patch nor a partial scan can represent, so the only correct reaction is a full rebuild with fresh caches.

Discovered tsconfig paths are remembered in a set on `RawTransformOptions` that `discover_tsconfig_file` fills during watch registration. The set deliberately survives `clear_cache`, so the bundling task can still classify a changed path after the caches were dropped. When a changed file hits the set, the bundling task clears the resolver and tsconfig caches and upgrades itself to `TaskInput::FullBuild`.

Registering the tsconfig as a transform dependency was considered and rejected: the HMR mapper would patch every module the tsconfig governs instead of doing one clean rebuild, and a pure `Hmr` task never clears the caches, so those patches would still use the stale options.

Review notes: detection lives in `BundlingTask`, not the coordinator, because the coordinator must not lock the bundler while a running build holds the lock. The upgrade happens after the `watchChange` hooks so they still see the changed files. The coordinator reads `requires_full_rebuild()` only at scheduling time, so a task that upgrades itself mid-run completes through the normal `InProgress` path.

Client notification (a `FullReload` update so browsers refresh automatically) and the dev fixtures come in a follow-up.

### DIFF
--- a/crates/rolldown_dev/src/bundling_task.rs
+++ b/crates/rolldown_dev/src/bundling_task.rs
@@ -114,6 +114,24 @@ impl BundlingTask {
       }
     }
 
+    // A tsconfig edit affects every module the tsconfig governs, which HMR
+    // patches and partial scans cannot represent. Clear the caches and fall
+    // back to a full rebuild.
+    {
+      let bundler = self.bundler.lock().await;
+      let changed_tsconfig = self
+        .input
+        .changed_files()
+        .keys()
+        .any(|path| bundler.options().transform_options.is_known_tsconfig(path));
+      if changed_tsconfig {
+        tracing::trace!("[BundlingTask] detects a tsconfig change, upgrading to a full rebuild");
+        bundler.clear_resolver_cache();
+        bundler.clear_transform_tsconfig_cache();
+        self.input = TaskInput::FullBuild;
+      }
+    }
+
     let mut has_full_reload_update = false;
     if self.input.require_generate_hmr_update() {
       tracing::trace!("[BundlingTask] starts to generate HMR updates");
--- a/crates/rolldown_common/src/inner_bundler_options/types/transform_options.rs
+++ b/crates/rolldown_common/src/inner_bundler_options/types/transform_options.rs
@@ -9,7 +9,7 @@ use oxc::transformer::{ESFeature, EngineTargets, TransformOptions as OxcTransfor
 use oxc_resolver::ResolverGeneric;
 use rolldown_error::{BuildDiagnostic, BuildResult};
 use rolldown_fs::OsFileSystem;
-use rolldown_utils::dashmap::FxDashMap;
+use rolldown_utils::dashmap::{FxDashMap, FxDashSet};
 
 use super::tsconfig_merge::merge_transform_options_with_tsconfig as merge_tsconfig;
 use crate::BundlerTransformOptions;
@@ -34,6 +34,9 @@ pub struct RawTransformOptions {
   /// Derived from the main resolver and shares its cache, so tsconfig
   /// lookups here and in module resolution stay consistent.
   resolver: Arc<ResolverGeneric<OsFileSystem>>,
+  /// Every tsconfig file discovered so far. Survives `clear_cache` so
+  /// watchers can still recognize tsconfig files when routing file changes.
+  known_tsconfig_paths: FxDashSet<PathBuf>,
 }
 
 impl RawTransformOptions {
@@ -41,7 +44,12 @@ impl RawTransformOptions {
     base_options: BundlerTransformOptions,
     resolver: Arc<ResolverGeneric<OsFileSystem>>,
   ) -> Self {
-    Self { base_options: Arc::new(base_options), cache: FxDashMap::default(), resolver }
+    Self {
+      base_options: Arc::new(base_options),
+      cache: FxDashMap::default(),
+      resolver,
+      known_tsconfig_paths: FxDashSet::default(),
+    }
   }
 
   /// Drop the merged transform options so the next build re-merges them.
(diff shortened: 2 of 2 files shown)


shulaoda:
* **#10262** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10262?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10261** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10261?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10261?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10258** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10258?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10257** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10257?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10205** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10205?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10200** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10200?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10261).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown_dev/src/bundling_task.rs
```diff
@@ -114,6 +114,24 @@ impl BundlingTask {
...
+    // patches and partial scans cannot represent. Clear the caches and fall
+    // back to a full rebuild.
+    {
+      let bundler = self.bundler.lock().await;
+      let changed_tsconfig = self
+        .input
+        .changed_files()
+        .keys()
+        .any(|path| bundler.options().transform_options.is_known_tsconfig(path));
+      if changed_tsconfig {
```
h-a-n-a, line 127: we might need to clean cache for `TaskInput::FullBuild` as well?

shulaoda:
### Merge activity

* **Jul 14, 6:10 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 14, 7:44 AM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 14, 7:49 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

crates/rolldown_dev/src/bundling_task.rs
shulaoda: Didn't it clear the cache right below? 🤔
```
bundler.clear_resolver_cache();
bundler.clear_transform_tsconfig_cache();
```

h-a-n-a: What I mean is `if changed_tsconfig || self.input.requires_full_rebuild()`. Shouldn't we clear the cache for this case as well?

shulaoda: Oh, I see what you mean. The caches here mainly refer to our own `package_json_cache: FxDashMap<PathBuf, Arc<PackageJson>>` and the resolver caches:

```rust
pub(crate) paths: DashMap<CachedPath, (), BuildHasherDefault<IdentityHasher>>,
/// Cache for raw/unbuilt tsconfigs (used when extending).
pub(crate) tsconfigs_raw: DashMap<PathBuf, Arc<TsConfig>, BuildHasherDefault<FxHasher>>,
/// Cache for built/resolved tsconfigs (used for resolution).
pub(crate) tsconfigs_built: DashMap<PathBuf, Arc<TsConfig>, BuildHasherDefault<FxHasher>>,
```

For the tsconfig caches, `self.input.requires_full_rebuild()` is irrelevant here. We only need to check whether a tsconfig file has changed.

h-a-n-a: > We only need to check whether a tsconfig file has changed.

Agreed — but for a bare `TaskInput::FullBuild` that check cannot be answered: `changed_files()` is empty by construction, so `changed_tsconfig` is `false` even when a tsconfig *did* change. Concrete case:

1. Start dev with `tsconfig.json` v1 = `{ "compilerOptions": { "target": "es5" } }`. It's valid JSON, so `find_tsconfig` loads it into the caches you listed (`tsconfigs_raw` / `tsconfigs_built`), but the per-file merge rejects it ("Rolldown only supports ES2015 and above"). The initial build fails → coordinator state = `FullBuildFailed`.
2. The user fixes the tsconfig (v2, `"target": "esnext"`). The change event arrives while in `FullBuildFailed`, and that arm of `handle_file_changes` pushes a bare `TaskInput::FullBuild` — the changed file name is dropped here.
3. The task runs: `changed_files()` is empty → `changed_tsconfig` is `false` → no cache clear. The full scan calls `find_tsconfig`, which answers from `tsconfigs_built` with **v1**, the merge fails again → back to `FullBuildFailed`.
4. Every subsequent edit repeats 2–3, because in `FullBuildFailed` *every* change becomes a bare `FullBuild`. Pressing `r` goes through `trigger_full_build`, which also produces a bare `FullBuild` — same result. The dev server can't recover without a restart, even though the tsconfig on disk has been correct since step 2.

The same erasure happens on two other paths: `trigger_full_build` clears `queued_tasks` (dropping a queued `Hmr { [tsconfig] }`), and `merge_with` has `FullBuild` absorb other inputs and discard their file lists.

So the `|| self.input.requires_full_rebuild()` isn't about clearing for unrelated full builds — it's the fallback for exactly the inputs whose file list has already been erased, where "did a tsconfig change" is unanswerable. Since full builds only happen at startup, on `r`, and in failure recovery, the extra clears are cheap. The alternative would be to keep the changed-file info alive through those three coordinator paths, but that's a much bigger change to the state machine.

shulaoda: Okay, that makes sense. I’ll open a follow-up PR in this stack.

h-a-n-a (approved):

PR #10270, opened 2026-07-14
fix: skip init-call transfer for circular wrapper deps

### LINKED ISSUE #10265: [Bug]: `onDemandWrapping` emits circular init_* calls before initialization

https://repl.rolldown.rs/#eNqNUk1PwzAM/StWDnSTpu5ehBCD/8Cll7R1t0DrltSFoan/HTtt2ScSUg6xHb/3/JyDKU1yMI4K3Mfc6Z1McoxXJtewbhvPcNgiPw1Q+qaGKF7b+K2L7lPy+NE7j4s5tZRcSrgPPXlDHYPHrq8YHkARFst7wUWTsO9xWBnfVFXRfFEsb0u3PZFxo3IuCAosHeFzKMMsbW6LTnTIQ6sSThsWh5QAHLU9JzBCxjWyjUNqBes18A6hdBV20GGFOWMB2XfIZj1zQyAHbb4DtpmCSUvTspOhNRJu9K5GYlslENhAOl6wtlS8etu2jrYJqA9aG1IalikZ8aS1+bvdotjZiBGjG2e50YcAmZoCW5SNUe6wS40wKdIIpBuZAcL9aqOb40azaaNztUJbHqsaTQ9+bS17ynXcabHjjB659yG1gbs70D54hMhGkEDUy38Rw2xWoSINKlJ5Z5Hh/p9vd1PD5lrDk3Jnf3JPU830czgquGTQ6gVDpLkJTeA+FcMydmyGH6ZOIOY=

### What is expected?

No error during evaluation

```sh
TypeError: init_a is not a function
```

### System Info

```Shell
rolldown 1.1.5
```

### Any additional comments?

This resembles #7449 at the symptom level, but does not require code splitting or a cross-chunk cycle. The reproduction emits a single chunk, and the undefined function is a Rolldown-generated `init_*` wrapper called before its later `var init_* = __esmMin(...)` assignment.

It is also distinct from #9946, which is specific to Vite `bundledDev` and emits a source-level cyclic import as an entirely unbound identifier.

#10008 tracks chunking-caused issues, while this failure occurs without multiple output chunks.

marvinhagemeister:
When a wrapped ESM module imports another wrapped module that appears later in the output (circular imports), the init-call transfer logic in `ensure_lazy_module_initialization_order` can move an `init_*()` call to an earlier unwrapped module's top-level position. The call then runs before the later wrapper's `var init_* = __esmMin(...)` is assigned, causing `TypeError: init_* is not a function`.

This was reported in #10265 with `onDemandWrapping`, but the underlying issue can trigger whenever unwrapped modules coexist with circular wrapped dependencies.

Fixes #10265

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/code_splitting.rs
@@ -429,14 +429,12 @@ impl GenerateStage<'_> {
                 // later in the output — calling it eagerly here would run before
                 // that later wrapper is assigned (circular imports: a↔b).
                 let target_exec_order = chunk_module_to_exec_order[&module_idx];
-                let has_later_wrapped_dep = self.link_output.metas[midx]
-                  .dependencies
-                  .iter()
-                  .any(|dep_idx| {
+                let has_later_wrapped_dep =
+                  self.link_output.metas[midx].dependencies.iter().any(|dep_idx| {
                     matches!(self.link_output.metas[*dep_idx].wrap_kind(), WrapKind::Esm)
-                      && chunk_module_to_exec_order.get(dep_idx).is_some_and(
-                        |&dep_order| dep_order > target_exec_order,
-                      )
+                      && chunk_module_to_exec_order
+                        .get(dep_idx)
+                        .is_some_and(|&dep_order| dep_order > target_exec_order)
                   });
                 if has_later_wrapped_dep {
                   continue;
--- a/crates/rolldown/src/stages/generate_stage/code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/code_splitting.rs
@@ -419,10 +419,26 @@ impl GenerateStage<'_> {
             {
               let transfer_item = pending_transfer
                 .extract_if(0.., |(midx, _, _)| wrapped_modules[0..*deps_length].contains(midx));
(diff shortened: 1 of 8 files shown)


hyfdev:
`onDemandWrapping` is under a big refactor https://github.com/rolldown/rolldown/pull/10104. I'll test your case under the new implementation. Make sure it's covered. Will ship the new implementation next week.

marvinhagemeister:
Sweet, that's perfect! FWIW: I was able to reproduce it even in a larger codebase without `onDemandWrapping` enabled but wasn't able to get a clean minimal reproduction out of that. Using `onDemandWrapping: true` was the best I could do.

IWANABETHATGUY:
> Sweet, that's perfect! FWIW: I was able to reproduce it even in a larger codebase without `onDemandWrapping` enabled but wasn't able to get a clean minimal reproduction out of that. Using `onDemandWrapping: true` was the best I could do.

Then I guess you probably did not fix the original issues in the large codebase, but thanks for reporting the bug with `onDemandWrapping`.

hyfdev:
Thanks for your effort. Ive added a test case under the new impl https://github.com/rolldown/rolldown/pull/10104 to ensure it's covered. New impl would be shipped this Wendesaday's release.

PR #10276, opened 2026-07-14
fix(dev): clear tsconfig caches for bare full builds

### LINKED ISSUE #10269: We need to clean cache for `TaskInput::FullBuild`
> We only need to check whether a tsconfig file has changed.

Agreed — but for a bare `TaskInput::FullBuild` that check cannot be answered: `changed_files()` is empty by construction, so `changed_tsconfig` is `false` even when a tsconfig *did* change. Concrete case:

1. Start dev with `tsconfig.json` v1 = `{ "compilerOptions": { "target": "es5" } }`. It's valid JSON, so `find_tsconfig` loads it into the caches you listed (`tsconfigs_raw` / `tsconfigs_built`), but the per-file merge rejects it ("Rolldown only supports ES2015 and above"). The initial build fails → coordinator state = `FullBuildFailed`.
2. The user fixes the tsconfig (v2, `"target": "esnext"`). The change event arrives while in `FullBuildFailed`, and that arm of `handle_file_changes` pushes a bare `TaskInput::FullBuild` — the changed file name is dropped here.
3. The task runs: `changed_files()` is empty → `changed_tsconfig` is `false` → no cache clear. The full scan calls `find_tsconfig`, which answers from `tsconfigs_built` with **v1**, the merge fails again → back to `FullBuildFailed`.
4. Every subsequent edit repeats 2–3, because in `FullBuildFailed` *every* change becomes a bare `FullBuild`. Pressing `r` goes through `trigger_full_build`, which also produces a bare `FullBuild` — same result. The dev server can't recover without a restart, even though the tsconfig on disk has been correct since step 2.

The same erasure happens on two other paths: `trigger_full_build` clears `queued_tasks` (dropping a queued `Hmr { [tsconfig] }`), and `merge_with` has `FullBuild` absorb other inputs and discard their file lists.

So the `|| self.input.requires_full_rebuild()` isn't about clearing for unrelated full builds — it's the fallback for exactly the inputs whose file list has already been erased, where "did a tsconfig change" is unanswerable. Since full builds only happen at startup, on `r`, and in failure recovery, the extra clears are cheap. The alternative would be to keep the changed-file info alive through those three coordinator paths, but that's a much bigger change to the state machine.

_Originally posted by @h-a-n-a in https://github.com/rolldown/rolldown/pull/10261#discussion_r3576921302_

shulaoda:
Closes #10269.

The tsconfig detection in `BundlingTask` checks `changed_files()` against the known tsconfig paths, but a bare `TaskInput::FullBuild` carries no changed-file list by construction. Three coordinator paths produce such inputs and erase the file names on the way: `handle_file_changes` in the `FullBuildFailed` state, `trigger_full_build` (which also clears queued tasks), and `merge_with` where `FullBuild` absorbs other inputs. On those paths "did a tsconfig change" cannot be answered, so the caches were never cleared and the full scan reused the stale tsconfig. In the failure-recovery case this locks the dev server up for good: every edit in `FullBuildFailed` becomes a bare `FullBuild`, each one rebuilds with the cached tsconfig, fails again, and returns to `FullBuildFailed`, even though the file on disk has long been fixed.

The fix clears the resolver and tsconfig caches for every input with `requires_full_rebuild()`, alongside the existing detection for inputs that still carry their file list. Full builds only happen at startup, on restart, and in failure recovery, so the extra clears are cheap. The `FullReload` client notification still fires only when a tsconfig change is positively detected. Keeping the changed-file information alive through the three coordinator paths would be the precise alternative, but that is a much larger change to the state machine for no practical gain.

The regression fixture reproduces the erasure path: the initial build fails on an unresolved import, and the recovery step fixes the import and edits the tsconfig in the same batch, which the `FullBuildFailed` state folds into a bare `FullBuild`. With the fix the recovered output uses the new tsconfig (`useDefineForClassFields` flips the emitted class field), and without it the output keeps the stale semantics, which was verified by temporarily reverting the fix.

### DIFF
--- a/crates/rolldown/tests/rolldown/topics/hmr/error_recovery/fix_broken_file_and_tsconfig/_config.json
+++ b/crates/rolldown/tests/rolldown/topics/hmr/error_recovery/fix_broken_file_and_tsconfig/_config.json
@@ -1,0 +1,15 @@
+{
+  "config": {
+    "input": [
+      {
+        "name": "main",
+        "import": "main.ts"
+      }
+    ],
+    "tsconfig": true,
+    "experimental": {
+      "devMode": {}
+    }
+  },
+  "expectError": true
+}
--- a/crates/rolldown/tests/rolldown/topics/hmr/error_recovery/fix_broken_file_and_tsconfig/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/topics/hmr/error_recovery/fix_broken_file_and_tsconfig/artifacts.snap
@@ -1,0 +1,39 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+---
+# Errors
+
+## UNRESOLVED_IMPORT
+
+```text
+[UNRESOLVED_IMPORT] Could not resolve './missing' in main.ts
+   ╭─[ main.ts:1:8 ]
+   │
+ 1 │ import './missing';
+   │        ─────┬─────  
+   │             ╰─────── Module not found.
+───╯
+
+```
+
+# HMR Step 0
+
+## Build Output
+
+### Assets
+
+#### main.js
+
+```js
+// HIDDEN [\0rolldown/runtime.js]
+//#region main.ts
+var main_exports = /* @__PURE__ */ __exportAll({ Foo: () => Foo });
+__rolldown_runtime__.createModuleHotContext("main.ts");
+__rolldown_runtime__.registerModule("main.ts", { exports: main_exports });
+var Foo = class {
+	bar = 1;
+};
+//#endregion
+export { Foo };
+
+```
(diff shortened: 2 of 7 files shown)


h-a-n-a (approved):

shulaoda:
### Merge activity

* **Jul 15, 2:04 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 15, 2:04 AM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 15, 2:10 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10282, opened 2026-07-14
chore(docs): migrate Netlify redirects to Void config

tux-tn:
The migration to Void in #9509 omitted the redirects added in #7834, leaving old documentation links broken. For example, these URLs currently return 404 instead of redirecting to the correct page:

   - [`https://rolldown.rs/about/`](https://rolldown.rs/about/) -> `/guide/`
   - [`https://rolldown.rs/options/input`](https://rolldown.rs/options/input) -> `/reference/InputOptions.input`
   - [`https://rolldown.rs/in-depth/code-splitting`](https://rolldown.rs/in-depth/code-splitting) -> `/in-depth/automatic-code-splitting`

This ports the existing rules from `docs/netlify.toml` to `docs/void.json`. The changes follow Void's [redirect](https://void.cloud/guide/edge/redirects) and [rewrite](https://void.cloud/guide/edge/rewrites) documentation.

The configuration has not been tested on Void Cloud since it's currently in private beta. The JSON syntax was validated locally, and all Netlify rules were verified against the resulting Void redirects and rewrites.

**Disclosure:** AI assistance was used to verify the redirect-matching configuration.

### DIFF
--- a/docs/void.json
+++ b/docs/void.json
@@ -4,11 +4,10 @@
   },
   "routing": {
     "redirects": {
-      "/about/": { "to": "/guide/", "status": 301 },
-      "/plugins/hook-filters": { "to": "/apis/plugin-hook-filters", "status": 301 },
-      "/guide/plugin-development": { "to": "/plugins/", "status": 301 },
-      "/reference/": { "to": "/apis/", "status": 301 },
-      "/guide/in-depth/": { "to": "/in-depth/", "status": 301 },
+      "/about/": { "to": "/guide/introduction", "status": 301 },
+      "/plugins/hook-filters": { "to": "/apis/plugin-api/hook-filters", "status": 301 },
+      "/guide/plugin-development": { "to": "/apis/plugin-api", "status": 301 },
+      "/guide/in-depth/": { "to": "/in-depth/why-bundlers", "status": 301 },
       "/apis/config-options/": { "to": "/reference/", "status": 301 },
       "/options/": { "to": "/reference", "status": 301 },
       "/apis/plugin-hook-filters": { "to": "/apis/plugin-api/hook-filters", "status": 301 },
@@ -48,13 +47,13 @@
       "/in-depth/advanced-chunks": { "to": "/in-depth/manual-code-splitting", "status": 301 },
       "/in-depth/code-splitting": { "to": "/in-depth/automatic-code-splitting", "status": 301 },
       "/options/output-clean-dir": { "to": "/reference/OutputOptions.cleanDir", "status": 301 },
-      "/*.txt": { "to": "/:splat.md", "status": 301 },
       "/llms.md": { "to": "/llms.txt", "status": 301 }
     },
     "rewrites": {
-      "/llms.txt": "/llms.txt",
-      "/llms-full.md": "/llms-full.txt",
-      "/llms-full.txt": "/llms-full.txt"
+      "/llms-full.md": "/llms-full.txt"
+    },
+    "fallbacks": {
+      "/*.txt": "/:splat.md"
     }
   },
   "inference": {
--- a/docs/void.json
+++ b/docs/void.json
@@ -4,39 +4,75 @@
   },
   "routing": {
     "redirects": {
+      "/about": { "to": "/guide/introduction", "status": 301 },
       "/about/": { "to": "/guide/introduction", "status": 301 },
       "/plugins/hook-filters": { "to": "/apis/plugin-api/hook-filters", "status": 301 },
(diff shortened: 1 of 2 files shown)


docs/void.json
```diff
@@ -2,6 +2,61 @@
...
+        "status": 301
+      },
+      "/options/output-advanced-chunks": {
+        "to": "/reference/OutputOptions.codeSplitting",
+        "status": 301
+      },
+      "/in-depth/advanced-chunks": { "to": "/in-depth/manual-code-splitting", "status": 301 },
+      "/in-depth/code-splitting": { "to": "/in-depth/automatic-code-splitting", "status": 301 },
+      "/options/output-clean-dir": { "to": "/reference/OutputOptions.cleanDir", "status": 301 },
+      "/*.txt": { "to": "/:splat.md", "status": 301 },
```
hyfdev: `routing.redirects` are merged before `routing.rewrites` in Void 0.10.8, and the first matching pre-asset rule wins. As a result, `/llms.txt` matches this wildcard and redirects to `/llms.md`; the next rule redirects it back to `/llms.txt`, creating an infinite 301 loop. The `/llms.txt` self-rewrite below is unreachable, and `/llms-full.txt` is also redirected away from its existing static asset. Please preserve the asset-miss behavior of the old Netlify rule, for example with a fallback rewrite if a 200 rewrite is acceptable, or another approach that keeps existing `llms*.txt` files ahead of the wildcard.

docs/void.json
```diff
@@ -2,6 +2,61 @@
   "worker": {
     "compatibility_date": "2026-02-24"
   },
+  "routing": {
+    "redirects": {
+      "/about/": { "to": "/guide/", "status": 301 },
+      "/plugins/hook-filters": { "to": "/apis/plugin-hook-filters", "status": 301 },
+      "/guide/plugin-development": { "to": "/plugins/", "status": 301 },
+      "/reference/": { "to": "/apis/", "status": 301 },
```
hyfdev: Netlify's old rule had `force = false` by default, so the existing `reference/index.html` shadowed it. Void redirects always run before static asset lookup, so this turns the working `/reference/` page into a permanent redirect to `/apis/`, which has no index and returns 404. It also makes `/apis/config-options/ → /reference/ → /apis/` end in 404. Please remove this obsolete redirect.

docs/void.json
```diff
@@ -2,6 +2,61 @@
   "worker": {
     "compatibility_date": "2026-02-24"
   },
+  "routing": {
+    "redirects": {
+      "/about/": { "to": "/guide/", "status": 301 },
```
hyfdev: These copied destinations are stale: the built site has no `guide/index.html`, `plugins/index.html`, or `in-depth/index.html`. Consequently `/about/`, `/guide/plugin-development`, and `/guide/in-depth/` still end at 404 after their 301. Please retarget them to current pages—for example `/guide/introduction`, `/apis/plugin-api`, and `/in-depth/why-bundlers`, subject to the intended content—and verify that the final response after following redirects is 200.

hyfdev (commented):
Three routing regressions need to be fixed before these rules can be deployed.

docs/void.json
hyfdev: Fixed in 59c651d87. The LLM routes now use ordered rules in `docs/public/_redirects`: the existing `llms*.txt` assets are protected by `200!` rewrites before the explicit root-level `.txt` redirects. This removes the loop and avoids matching nested documentation paths.

docs/void.json
hyfdev: Fixed in 3b501baa5. Removed the obsolete `/reference/ -> /apis/` redirect. `/reference/` now resolves to the existing static page, and both forms of `/apis/config-options` redirect to it.

docs/void.json
hyfdev: Fixed in ed617390d. Retargeted both trailing-slash forms of `/about`, `/guide/plugin-development`, and `/guide/in-depth` to `/guide/introduction`, `/apis/plugin-api`, and `/in-depth/why-bundlers`, respectively. All destinations exist in the built site.

hyfdev (approved):
The prior routing regressions on this branch are fixed: always-force no longer collides with the live `/reference/` page, LLM routes no longer loop, and the retargeted destinations resolve to real pages on the current site. Trailing-slash sources are covered, and the root `.txt` enumeration keeps the existing `llms*.txt` assets ahead of the page aliases.

`docs/netlify.toml` is still the Netlify-preview baseline and still points a few legacy sources at 404 destinations; that does not affect Void production and can be cleaned up separately if previews should match.

hyfdev:
Thank you!

PR #10285, opened 2026-07-15
refactor: replace take_in-then-write-back with ReplaceWith and by-value moves

Boshen:
## What

Converts the `take_in` → build → write-back patterns in the module finalizer, scan pre-processor, HMR finalizer, and lazy-export generation to `oxc_allocator::ReplaceWith` (available since oxc 0.139) or plain by-value moves.

## Why

Each `take_in` writes a `Dummy` node into the vacated slot; for AST enums that dummy is a real arena allocation (16 bytes boxed) that is overwritten an instant later but stays in the pooled module arena until the generate stage ends. It advances the bump pointer (raising the pool's retained high-water mark) and lowers live-node density for every later walk over the same arena — scanner, DCE, finalizer, codegen. oxc's own parser (oxc-project/oxc#24018) and minifier (oxc-project/oxc#24017) made the same move in 0.140.0.

## How

- **Slot-level `replace_with`**: keep-name wraps (finalizer ×3), both dynamic-`import()`-in-CJS rewrites, the `.then((m) => __toESM(...))` interop wrapper, `replace_first_expr_stmt`, and the PreProcessor's `require(a?b:c)` / `import(a?b:c)` transposes (those destructure the conditional by value — 3 dummies each, now 0).
- **By-value with box reuse**: the finalizer's export-default block and the HMR finalizer's `handle_top_level_stmt` now match the owned statement directly, so `export default function/class` reuses the existing `Box<Function>`/`Box<Class>` instead of take-dummy-rebox (also saves the ~200-byte struct copy). `get_transformed_class_decl` takes the class box by value and returns `Result<Declaration, Box<Class>>`, giving the box back when no transform applies.
- **Plain moves where no trait is needed**: var-hoisting drains `declarations` by value (`Vec`'s dummy is an empty vec — no allocation), JSON prop inlining uses `Option::take`, and paren unwrapping in lazy export works by value.

Left as-is: the `Vec` drain-and-rebuild `take_in`s (free dummy, structurally required) and the `take_in_box` in the dynamic-import-to-other-chunk path, whose `Option`-returning protocol tolerates a partially consumed node on the warn paths.

One behavior note: the finalizer's export-default fallback arm is now `unreachable!()` for `TSInterfaceDeclaration` (TS is stripped before finalization), matching the convention already used by `visit_declaration` and the HMR finalizer for TS variants.

Output is byte-identical: the snapshot suite passes with zero snapshot updates.

### DIFF
--- a/crates/rolldown/src/hmr/hmr_ast_finalizer.rs
+++ b/crates/rolldown/src/hmr/hmr_ast_finalizer.rs
@@ -1,7 +1,7 @@
 use oxc::allocator::GetAllocator;
 use oxc::ast::ast::Str;
 use oxc::{
-  allocator::{Box as ArenaBox, IntoIn, TakeIn},
+  allocator::IntoIn,
   ast::{
     NONE,
     ast::{self, ExportDefaultDeclarationKind, Expression, ObjectPropertyKind, Statement},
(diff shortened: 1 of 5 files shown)


hyfdev (approved):
LGTM

IWANABETHATGUY:
### Merge activity

* **Jul 15, 3:24 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 15, 3:24 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 15, 3:29 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

Issue #7555, opened 2025-12-17
[Bug]: Sourcemap -> Sourcemap remapping is less accurate than Rollup


johnpyp:
### Reproduction link or steps

When running a transformation on the code using a tool which outputs another sourcemap in a plugin, for example an obfuscator, Rolldown doesn't seem to properly remap the sourcemap and loses some granularity.

Example file:
```js
function throwError() {
  const inner = () => {
    // pad so the throw stays on line 4
    throw new Error("TEST_ERROR"); // line 4
  };
  inner();
}

throwError();
```

[...]

PR #10289, opened 2026-07-15
fix(output): skip separator for empty import prelude

hyfdev:
Refs #7555.

## Why

The wrapped-format renderer appended `import_code` to `SourceJoiner` even when the generated import prelude was empty. `SourceJoiner` treats every appended value as a source slot and inserts a separator between slots, so an empty prelude still caused an unwanted separator: before the first module in the CJS reproduction and at the corresponding module boundary in affected IIFE and UMD output.

## What changed

- skip the generated import prelude when it is empty
- update the affected IIFE and UMD snapshots
- add a CJS `renderChunk` fixture that verifies the hook input starts with the first module, without a separator from an empty prelude

## Source-map scope

The final composed map observed in #7555 was wrong, but Rolldown's producer map and composition were correct for the inputs they received. Rolldown's producer map correctly described the exact `renderChunk` input, including the leading separator. JavaScript Obfuscator trimmed that input before parsing and returned a map for the trimmed code without representing the trim relative to the hook input ([source](https://github.com/javascript-obfuscator/javascript-obfuscator/blob/35860ec7087b1e53e89233419f91336f6d4f0bd5/src/code-transformers/preparing-transformers/HashbangOperatorTransformer.ts#L54-L63)).

This PR does not change Rolldown's source-map generation or composition logic; there was no bug in either for this reproduction. It removes the unintended output separator. As a consequence, the exact #7555 reproduction no longer triggers the coordinate mismatch in the plugin-returned map. Plugins remain responsible for returning maps relative to the exact code passed to `renderChunk`.

The regression fixture intentionally checks the hook input rather than source-map composition. A valid plugin map constructed from the token's actual hook position composes to the correct original position both before and after this change, so that assertion would not detect this regression.

## Before and after

For the CJS reproduction, the `renderChunk` input changes from:

```text
Before: "\nfunction throwError() { ..."
After:  "function throwError() { ..."
```

For affected IIFE and UMD output, the removed separator is at the wrapper or factory boundary rather than necessarily at the start of the whole chunk.

## Validation

- targeted `plugin/render-chunk/issue-7555` Node fixture
- `vp check packages/rolldown/tests/fixtures/plugin/render-chunk/issue-7555/_config.ts`
- all seven affected Rust fixtures
- `cargo fmt --all --check`
- `cargo check -p rolldown --all-features --all-targets --locked`

### DIFF
--- a/crates/rolldown/src/ecmascript/format/utils/mod.rs
+++ b/crates/rolldown/src/ecmascript/format/utils/mod.rs
@@ -144,7 +144,11 @@ pub fn render_modules_with_peek_runtime_module_at_first<'a>(
     _ => {}
   }
 
-  source_joiner.append_source(import_code);
+  // An empty import prelude is not a source slot. Appending it would make `SourceJoiner` insert a
+  // separator line before the modules.
+  if !import_code.is_empty() {
+    source_joiner.append_source(import_code);
+  }
 
   // chunk content
   module_sources_peekable.for_each(
--- a/crates/rolldown/tests/rolldown/function/extend/iife/namespace_default/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/function/extend/iife/namespace_default/artifacts.snap
@@ -12,7 +12,6 @@ class A {
 this.test = this.test || {};
 this.test.module = (function() {
 
-
 //#region main.js
 	var main_default = "default";
 
--- a/crates/rolldown/tests/rolldown/function/extend/iife/namespace_named/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/function/extend/iife/namespace_named/artifacts.snap
@@ -13,7 +13,6 @@ this.test = this.test || {};
 (function(exports) {
 
 Object.defineProperty(exports, Symbol.toStringTag, { value: 'Module' });
-
 //#region main.js
 	const a = 1;
 	const b = 2;
--- a/crates/rolldown/tests/rolldown/function/extend/iife/non_namespace_default/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/function/extend/iife/non_namespace_default/artifacts.snap
@@ -11,7 +11,6 @@ class A {
   constructor() {
 this.module = (function() {
 
-
 //#region main.js
 	var main_default = "default";
 
--- a/crates/rolldown/tests/rolldown/function/extend/iife/non_namespace_named/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/function/extend/iife/non_namespace_named/artifacts.snap
@@ -12,7 +12,6 @@ class A {
 (function(exports) {
 
 Object.defineProperty(exports, Symbol.toStringTag, { value: 'Module' });
-
 //#region main.js
 	const a = 1;
 	const b = 2;
--- a/crates/rolldown/tests/rolldown/function/format/iife/namespaced_name/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/function/format/iife/namespaced_name/artifacts.snap
@@ -13,7 +13,6 @@ this.namespace = this.namespace || {};
 this.namespace.module = (function(exports) {
 
 Object.defineProperty(exports, Symbol.toStringTag, { value: 'Module' });
-
 //#region foo.js
 	const value = 1;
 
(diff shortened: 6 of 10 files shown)


hyfdev:
* **#10289** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10289?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10289?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10289).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyfdev:
### Merge activity

* **Jul 15, 12:34 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 15, 3:04 PM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 15, 3:09 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

IWANABETHATGUY:
None of this blocks merging.

**1. Same bug is still live in `esm.rs`.** Line 202:

```rust
source_joiner.append_source(hoisted_fns);        // unconditional
if !hoisted_vars.is_empty() {                    // ...but this one is guarded
  source_joiner.append_source(concat_string!("var ", hoisted_vars, ";"));
}
```

When a module group has no hoisted functions, `hoisted_fns` stays `""` and buys a separator, exactly like the import prelude did. The fact that `hoisted_vars` right below it *is* guarded makes me think this is just an oversight. Same shape of fix:

```rust
if !hoisted_fns.is_empty() {
  source_joiner.append_source(hoisted_fns);
}
```

I tried it: 1832 pass, 0 fail, and exactly one snapshot changes — `strict_execution_order/exports_chain_shared_barrel_fixpoint`, which loses a stray second blank line before `var init_main`. Only one, because it's the only affected fixture with `minify: false`; everywhere else the minifier hides the blank line. It's ESM + `onDemandWrapping` only, so it overlaps the #10104/#10287 work. Separate PR is fine, I'd rather keep this one small.

**2. Longer term, maybe fix the class in `SourceJoiner` instead.** An early return on empty content in `append_source` would cover both this and the above, and stop callers from having to remember:

```rust
if source.content().is_empty() {
  return; // an empty source is not a line
}
```

Not for this PR though — `append_source_dyn` would need the same, and an empty source carrying a sourcemap would get dropped (`accumulate_sourcemap_data_size` also flips `enable_sourcemap`, so it's not purely cosmetic). Just flagging it, since nothing in the tree seems to actually *want* an empty slot.

**3. Nit on the fixture: it passes if `renderChunk` never fires.** For a test whose whole job is guarding a regression that'd be a shame. `afterTest` gets you a counter cheaply:

```ts
let renderChunkCalls = 0;

export default defineTest({
  config: { /* ...unchanged... */ },
  afterTest: () => {
    expect(renderChunkCalls).toBe(1);
  },
});
```

(with `renderChunkCalls++` at the top of the hook). `render-chunk-external-import` has the same gap, so this isn't on you — it just matters more here.

**4. Nit on the assertion.** `expect(code.startsWith('function throwError')).toBe(true)` fails with `expected false to be true` and nothing else — the `startsWith` throws the string away before vitest can print it. Either of these keeps the code in the failure output:

```ts
expect(code).toMatch(/^function throwError/);  // same meaning
expect(code).not.toMatch(/^\n/);               // says why the test exists
```

If you go with the second one, keep `attachDebugInfo: 'none'` — it's load-bearing there. Under `'full'` a `//!` line gets prepended (`ecma_generator.rs:264`), which pushes the empty slot off the front, and the separator becomes a mid-file blank line that `/^\n/` won't see.

IWANABETHATGUY (approved):

PR #10290, opened 2026-07-15
fix(runtime): don't re-wrap require(esm) namespaces in __toESM

### LINKED ISSUE #10308: Default import of an ESM-only dependency is undefined in the CommonJS build of a "type": "module" package

Minimal repo: https://github.com/TheHolyWaffle/tsdown-cjs-esm-default-interop

```sh
git clone https://github.com/TheHolyWaffle/tsdown-cjs-esm-default-interop
cd tsdown-cjs-esm-default-interop
npm install
npm run build     # tsdown, exit 0
npm run repro     # node -e "require('./dist/index.cjs')"
# -> TypeError: chalk.default.red is not a function
```

`package.json` sets `"type": "module"`, `tsdown.config.ts` uses `format: ['esm', 'cjs']`, and `src/index.ts` is:

```ts
import chalk from 'chalk';   // chalk 5 is ESM-only with a default export

export const message = chalk.red('hello');
```

### What is expected?

`require()`-ing the CommonJS build resolves the default export correctly and `chalk.red` is callable — the same as when the package is `"type": "commonjs"`, where tsdown emits `chalk = __toESM(chalk)` (no `isNodeMode`) and the file loads fine.

In a `"type": "module"` package, the `.cjs` output default-imports the ESM-only external dependency with `isNodeMode = 1`:

```js
let chalk = require("chalk");
chalk = __toESM(chalk, 1);               // isNodeMode = 1
const message = chalk.default.red("hello");
exports.message = message;
```

Node's `require("chalk")` already returns a module namespace with a correct `.default` and `__esModule: true`. The `__toESM(mod, isNodeMode)` helper, when `isNodeMode` is truthy, unconditionally does `target.default = mod` — overwriting the real default with the whole namespace. So `chalk.default` is the namespace and `chalk.default.red` is `undefined`:

```
TypeError: chalk.default.red is not a function
```

The trigger is the package's `"type"` field:

| `package.json` `"type"` | emitted interop | `require()` result |
| --- | --- | --- |
| `"module"` | `__toESM(chalk, 1)` | TypeError |
| `"commonjs"` | `__toESM(chalk)` | works |

Only `"type"` changes between the two; source, config, and the external dependency are identical.

### Any additional comments?

Not a rolldown bug. Driving rolldown directly (`external: ['chalk']`, `format: 'cjs'`, with or without `preserveModules` / `platform: 'node'`) emits the correct `__toESM(chalk)` (no `isNodeMode`) and loads fine. rolldown has no notion of the consumer's package `type`; the `isNodeMode = 1` is introduced by tsdown for a `.cjs` emitted inside a `"type": "module"` package.

Real-world impact: a dual-published `"type": "module"` package that wraps an ESM-only dependency and exposes a `require` condition (`./dist/index.cjs`) crashes every CommonJS consumer the moment it touches the dependency's default export. A `"type": "module"` package cannot simply switch to `"commonjs"`, so today the only workarounds are bundling the dependency or recovering the default in source (`const dep = mod.default ?? mod`).

Versions:

```
tsdown    0.22.3
rolldown  1.1.4
node      24.16.0
npm       11.13.0
OS        macOS (darwin 25.5.0)
```

SliverKeigo:
### What is this PR solving?

On Node ≥ 20.19 / ≥ 22.12, `require()` of an ESM-only package returns a real module namespace. The node-mode branch of `__toESM` still overwrites its `default` with the namespace itself, so default imports of external ESM packages break in the CJS output of a `"type": "module"` package:

```js
let esm_pkg = require("esm-pkg");
esm_pkg = __toESM(esm_pkg, 1);
(0, esm_pkg.default)("hi");   // TypeError: esm_pkg.default is not a function
```

Native Node ESM behaves correctly; only the generated CJS bundle differs.

### The fix

In node mode, detect genuine module namespaces and leave their `default` intact. The check is structural — non-extensible, `null` prototype, `Symbol.toStringTag === "Module"`, in that order — so webpack-tagged CJS exports are not mistaken for real namespaces, proxies that throw on any property access are never probed, and primitive exports skip the checks. Verified against native Node `import` behavior for six dependency shapes. Code generation is unchanged; only the runtime behavior of `__toESM` changes.

### Prior art

#9076 reported this failure mode and was closed as fixed by #8987 — but #8987 addressed CJS externals with `__esModule`, and this case still reproduces on 1.2.0. The esbuild-parity concern raised there predates unflagged `require(esm)`; matching Node's behavior now requires recognizing real namespaces.

### Tests

The new fixture fails without this change and passes with it; the full workspace suite passes.

**AI disclosure:** this PR was developed with AI assistance (Claude). I've reviewed and validated every change myself.

Fixes #10308


IWANABETHATGUY:
Please read our contribution guide first https://github.com/rolldown/rolldown/blob/main/docs/contribution-guide/index.md

SliverKeigo:
Marking this ready for review now that the original report lives here as #10308.

Rebased onto latest main, so the diff no longer touches `runtime/index.js` (removed as unused in #10302) — the change is only in `runtime-base.js` now. Re-ran the new fixture and the full workspace suite on the rebased branch,
all passing.

SliverKeigo:
Rebased to resolve the conflict with #10453 — the two changes touch the same condition but different branches of it: #10453 guards the Babel branch (`__esModule` without an own `default`), this PR guards the node-mode branch (genuine require(esm) namespaces). Both are preserved.
worth noting #10453 also establishes the same principle this PR relies on: diverging from esbuild where its answer is a guaranteed `TypeError` at runtime.

PR #10214, opened 2026-07-10
fix: `[format]` in `*FileNames` option for ESM format should be `es` instead of `esm`

sapphi-red:
Rollup uses `es` instead of `esm` ([REPL](https://rollupjs.org/repl/?version=4.62.2&shareable=eyJleGFtcGxlIjpudWxsLCJtb2R1bGVzIjpbeyJjb2RlIjoiY29uc29sZS5sb2coXCJtYWluXCIpXG5pbXBvcnQoJy4vcXV4LmpzJykiLCJpc0VudHJ5Ijp0cnVlLCJuYW1lIjoibWFpbi5qcyJ9LHsiY29kZSI6ImNvbnNvbGUubG9nKFwicXV4XCIpIiwiaXNFbnRyeSI6ZmFsc2UsIm5hbWUiOiJxdXguanMifV0sIm9wdGlvbnMiOnsib3V0cHV0Ijp7ImVudHJ5RmlsZU5hbWVzIjoiW25hbWVdLVtmb3JtYXRdLmpzIiwiZm9ybWF0IjoiZXMifX19)).

Given that this change may have a larger impact than other bug fixes, **this should land on minor versions.**

PR #10291, opened 2026-07-15
feat: implement `resolveFileUrl` plugin hook

sapphi-red:
This PR implements the `resolveFileUrl` hook, which exists in Rollup.

1. Collect `import.meta.ROLLUP_FILE_URL_*` information in the AstScanner
1. Rewrite the `import.meta.ROLLUP_FILE_URL_*` places in GenerateStage with the information collected

https://rollupjs.org/plugin-development/#resolvefileurl

refs https://github.com/vitejs/vite/issues/22709


sapphi-red:
* **#10296** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10296?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#10297](https://github.com/rolldown/rolldown/pull/10297) <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10297?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#10330](https://github.com/rolldown/rolldown/pull/10330) <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10330?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#10291** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10291?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10291?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10214** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10214?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10291).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown/src/ast_scanner/impl_visit.rs
```diff
@@ -275,6 +276,20 @@ impl<'me, 'ast: 'me> Visit<'ast> for AstScanner<'me, 'ast> {
     walk::walk_new_expression(self, it);
   }
 
+  /// Records `import.meta.ROLLUP_FILE_URL_<referenceId>` for the `resolveFileUrl` hook.
+  ///
+  /// Deliberately not folded into `visit_meta_property`: that method returns early when
+  /// `keep_esm_import_export_syntax()` is set, but `ROLLUP_FILE_URL_` is rewritten
+  /// regardless of that option.
+  fn visit_static_member_expression(&mut self, it: &ast::StaticMemberExpression<'ast>) {
```
IWANABETHATGUY: I think the scanner and the finalizer now disagree on what counts as a file URL reference. The finalizer's rewrite goes through `try_rewrite_import_meta_prop_expr`, which uses `static_property_name()` — so it also recognizes `import.meta['ROLLUP_FILE_URL_x']` with a string key. But the collection here only sees `StaticMemberExpression`, so the computed spelling never reaches the hook and silently falls back to the default rewrite, even with a plugin installed.

I checked on this branch — same emitted asset, same hook returning `'HOOK'`, es format:

| spelling | hook calls | output |
|---|---|---|
| `import.meta.ROLLUP_FILE_URL_x` | 1 | `const url = "HOOK";` |
| `import.meta['ROLLUP_FILE_URL_x']` | 0 | `const url = new URL("assets/a-....txt", import.meta.url).href;` |

There's also a small warning regression hiding in the same gap. On main, `import.meta['ROLLUP_FILE_URL_x']` in iife output produces an `EMPTY_IMPORT_META` warning (the old scan-time check used `static_property_name()` too, so it covered computed keys). On this branch it produces broken `{}.url` output with no warning at all: the scan-time path now returns `false` for the prefix, and the new generate-stage warning only iterates the recorded references — which don't include the computed form.

| build | dot, iife, no hook | computed, iife, no hook |
|---|---|---|
| main | 1 warning | 1 warning |
| this PR | 1 warning | **0 warnings**, `{}.url` in output |

Two ways to close the gap, and I don't have a strong preference:

- record computed string-literal keys in the scanner too, or
- restrict the finalizer's rewrite to the static form so both stages agree — Rollup only supports the dot form anyway, so this would also be a Rollup-compat argument.

Either way, the scanner and the finalizer should recognize the same set of spellings.

packages/rollup-tests/src/ignored-by-unsupported-features.md
```diff
@@ -302,6 +299,9 @@
  - rollup@form@import-attributes@keeps-static-attributes-key-default: keeps any import attributes on input using import attributes with "with" key@generates es
  - rollup@form@import-attributes@keeps-static-attributes-key-with: keeps any import attributes on input using import attributes with "with" key@generates es
  - rollup@form@resolve-file-url-import-meta-attributes: adds attributes to file resolveFileUrl and resolveImportMeta hooks@generates es
+ - rollup@form@configure-file-url: allows to configure file urls@generates es
+ - rollup@chunking form@resolve-file-url: allows to configure file urls@generates es
+ - rollup@chunking form@resolve-file-url: allows to configure file urls@generates cjs
```
IWANABETHATGUY, line 304: The passed count drops from 1213 to 1211, and the two `rollup@chunking form@resolve-file-url` entries move from passing to ignored — filed under "Import attributes is not supported", which doesn't sound like what these tests exercise ("allows to configure file urls").

Is the actual blocker the AST-vs-text splicing divergence (i.e. behavior matches but the snapshot differs)? If so, the "behavior passed, snapshot different" bucket seems like the more accurate home. Either way, a sentence in the PR description explaining why implementing the hook *reduces* the passed count would help the next person who reads the ledger.

packages/rolldown/src/plugin/docs/plugin-hooks-resolvefileurl.md
```diff
@@ -0,0 +1,20 @@
+Allows customizing how Rolldown resolves URLs of files that were emitted by plugins via [`this.emitFile`](/reference/Interface.PluginContext#emitfile). By default, Rolldown will generate code for `import.meta.ROLLUP_FILE_URL_referenceId` that should correctly generate absolute URLs of emitted files independent of the output format and the host system where the code is deployed. See [File URLs](/apis/plugin-api/file-urls) for more details and an example.
+
+This hook can be used to customize the behavior of `import.meta.ROLLUP_FILE_URL_referenceId`.
+
+The returned string must be a single JavaScript expression. Also the returned expression must be side-effect free. If the URL is not used in the code, Rolldown will remove it.
```
IWANABETHATGUY, line 5: Nit: it might be worth documenting that the returned code becomes real AST — the docs say the return value must be a single expression, but not that it's parsed and then processed like module code. Two consequences a plugin author might not expect:

- `import.meta.url` inside the returned expression gets rewritten per format (polyfilled in node-cjs, `{}` in iife) — Rollup splices it verbatim, so this is different *and better*, but it's different;
- constant folding can fuse the returned expression into its surroundings (the `basic` fixture relies on `'resolved:' + "path"` collapsing into one literal).

One or two sentences here would pin both down as intended behavior rather than accidents.

IWANABETHATGUY (commented):
Take it or leave it — fixture coverage is already strong; two gaps if you want them closed:

- a hashed-filename case (`chunkFileNames: '[name]-[hash].js'` + an emitted chunk) proving the hash placeholder in `fileName`/`chunkId` round-trips through plugin-returned code;
- a cjs-format case — the hook is typed for `'es' | 'cjs' | 'iife' | 'umd'` but only es and iife are exercised.

crates/rolldown/src/stages/generate_stage/resolve_file_urls.rs
```diff
@@ -0,0 +1,109 @@
...
+      }
+      let Some(chunk_idx) = chunk_graph.module_to_chunk[module.idx] else { continue };
+      let chunk = &chunk_graph.chunk_table[chunk_idx];
+      let chunk_id = chunk
+        .preliminary_filename
+        .as_ref()
+        .expect("chunk should have a preliminary filename by now")
+        .as_str();
+
+      for (span, reference_id) in &module.ecma_view.rollup_file_url_references {
```
hyfdev: This iterates every scanner-recorded occurrence in an included module, including occurrences that tree-shaking removes. I reproduced three observable differences from Rollup 4.62.2:

- With only a dead reference, Rolldown calls the hook once while Rollup calls it zero times. If the hook throws, the Rolldown build fails while Rollup succeeds.
- With a dead reference followed by a live reference to the same emitted file, Rolldown calls twice and the live output receives `call-2`; Rollup calls once and it receives `call-1`.
- Even a dead declarator inside an otherwise retained declaration triggers the hook, so filtering only by module or `StmtInfoIdx` would not be sufficient.

Could this derive hook invocations from occurrences that remain after tree-shaking, at a granularity fine enough to exclude dead expressions within an otherwise retained statement? A regression fixture where a dead reference's hook throws would cover the build-failure case.

crates/rolldown_plugin/src/types/hook_resolve_file_url_args.rs
```diff
@@ -0,0 +1,15 @@
...
+  /// Preliminary filename of the chunk containing the reference.
+  pub chunk_id: &'a str,
+  /// Filename of the emitted file, relative to the output directory.
+  pub file_name: &'a str,
+  pub format: OutputFormat,
+  /// Id of the module containing `import.meta.ROLLUP_FILE_URL_<reference_id>`.
+  pub module_id: &'a str,
+  pub reference_id: &'a str,
+  /// Path from the chunk to the emitted file.
+  pub relative_path: &'a str,
```
hyfdev, line 14: Rollup's `resolveFileUrl` options include a required `attributes: Record<string, string>` field. For an ordinary module, Rollup passes `{}`, while this branch exposes `undefined`; plugins that destructure or inspect it will break, and the TypeScript API is incompatible as well.

Could we add the field through the Rust/binding/TypeScript layers and populate it from the module's attributes, using `{}` when there are none? This is also the reason the two `rollup@chunking form@resolve-file-url` tests moved from passing to unsupported.

hyfdev (changes requested):
The overall direction and the two-stage design make sense, but I found one behavior difference that can make an otherwise successful build fail, so I'm requesting changes.

In priority order:

1. **Blocker:** `resolveFileUrl` is called for file-URL occurrences that tree-shaking removes. See the inline comment.
2. **Rollup compatibility:** `ResolveFileUrlArgs` omits Rollup's required `attributes` field. See the inline comment. This is also why the two chunking tests moved from passing to unsupported: Rollup passes `{}` even for an ordinary module, while this branch passes no field.
3. **Rollup compatibility:** On the existing computed-property thread, Rollup 4.62.2 does support `import.meta['ROLLUP_FILE_URL_x']` via its string `propertyKey`, so the scanner should be extended rather than restricting the finalizer.
4. **Docs:** The AST-parsing behavior is worth documenting as already noted. The File URLs link should also target `Interface.Plugin#resolvefileurl`, and the "independent of the output format" wording should be qualified because the default IIFE/UMD fallback warns and produces `{}.url`.

I also verified that hash placeholders round-trip through hook-returned code, output-only plugins work, and repeated ES/CJS generation receives the correct `format`; I did not find issues in those paths.

packages/rollup-tests/src/ignored-by-unsupported-features.md
sapphi-red: These tests expect the `attribute` property to be `{}`.
https://github.com/rollup/rollup/blob/4b860802d9656a0d8ffccc0a87bb9c2f5d82cff8/test/chunking-form/samples/resolve-file-url/_config.js#L39

But I intentionally avoided adding that property to align with the behavior with other hooks like `resolveId`.
- Rollup: https://rollupjs.org/plugin-development/#resolveid
- Rolldown: https://rolldown.rs/reference/Interface.FunctionPluginHooks#extraoptions

The test passed previously because the `resolveFileUrl` hook never ran and the assertion inside never ran.

crates/rolldown_plugin/src/types/hook_resolve_file_url_args.rs
sapphi-red: https://github.com/rolldown/rolldown/pull/10291#discussion_r3592872671

crates/rolldown/src/stages/generate_stage/resolve_file_urls.rs
hyfdev: I dug further into what an exact fix would require. Rollup gets this behavior by calling the synchronous hook while rendering an AST node that has already been marked as included. In Rolldown, the async plugin hook currently runs before finalization and chunk assembly have determined the exact surviving occurrences, so filtering scanner records cannot recover every occurrence that later becomes dead inside an included top-level statement. Exact parity would require a two-phase render path that carries occurrence identity through finalization and chunk assembly, determines survivors before `renderChunk`, and only then invokes the hook and replaces its results. That is a substantial pipeline change rather than a local fix for this PR.

If we expect dead `ROLLUP_FILE_URL_*` references to be uncommon in real plugins, and are comfortable with the observable difference for throwing or stateful hooks, I’m okay treating this thread as non-blocking rather than holding this PR on that architectural work. We should still record the limitation explicitly in the Rollup compatibility status or a follow-up issue.

sapphi-red: > * Even a dead declarator inside an otherwise retained declaration triggers the hook, so filtering only by module or `StmtInfoIdx` would not be sufficient.

Would you elaborate this case? Isn't this the treeshaking difference rather than which phase `resolveFileUrl` runs?

crates/rolldown/src/ast_scanner/impl_visit.rs
sapphi-red: This is now fixed.

crates/rolldown/src/stages/generate_stage/resolve_file_urls.rs
sapphi-red: The first two case should be now fixed.

packages/rolldown/src/plugin/docs/plugin-hooks-resolvefileurl.md
sapphi-red: I added the first point to the docs. I didn't add the second point because that can happen in Rollup's case using plugins.

packages/rolldown/src/plugin/docs/plugin-hooks-resolvefileurl.md
```diff
@@ -0,0 +1,26 @@
...
+
+::: tip `import.meta.url` in the returned string
+
+If the returned string contains `import.meta.url`, it will be rewritten for non-ESM formats similarly to [when `import.meta.url` is used in the code directly](/in-depth/non-esm-output-formats#well-known-import-meta-properties). Unlike Rolldown, Rollup outputs `import.meta.url` as-is.
+
+:::
```
sapphi-red, line 11: If the plugin returns `import.meta.url` in the hook for non-ESM formats and `import.meta` is not polyfilled. Currently, no error happens. https://github.com/rolldown/rolldown/pull/10222 fixes that as well.

sapphi-red:
I also refactored the code to use `NodeId` instead of `Span`.

crates/rolldown/src/stages/generate_stage/resolve_file_urls.rs
hyfdev: You're right that my original “dead declarator” wording was inaccurate. On the latest commit, the ordinary mixed-declaration case I had in mind is filtered correctly, and the destructuring case I tried is not a difference from Rollup either.

A more precise remaining case is:

```js
export const kept = (import.meta.ROLLUP_FILE_URL_REF, 1);
```

On the current branch, Rolldown calls `resolveFileUrl` once and then emits `const kept = 1`; Rollup 4.62.2 calls the hook zero times and emits the same effective result. The same happens for a standalone file-URL expression inside an exported function body: both outputs remove the expression, but only Rolldown calls the hook.

So yes, this is tree-shaking-related. More specifically, nested-expression liveness becomes known after Rolldown's pre-finalization hook pass, while Rollup invokes the hook only while rendering included nodes. Exact parity would require carrying occurrence-level identity and liveness into a later render phase, which is a substantially larger pipeline change. Given how unusual these patterns are in practice, I’m treating this remaining difference as non-blocking for this PR.

hyfdev (commented):
I re-reviewed the latest head. The computed-key and common whole-statement tree-shaking cases are fixed. I’m treating the remaining nested-expression liveness timing difference as non-blocking and will clarify that in the existing thread.

I found one live-code issue that still blocks approval, plus one non-blocking implementation difference; see the inline comments.

Two documentation items from my previous review also remain:

- The File URLs page links `resolveFileUrl` to `Interface.PluginContext`; it should link to `/reference/Interface.Plugin#resolvefileurl`.
- The hook documentation says the default URL works “independent of the output format”, but the current IIFE/UMD fallback warns and emits `{}.url`. This should be qualified to mention the need for an `import.meta.url` polyfill or a custom hook.

Everything else I checked looks good. The focused fixtures, injected-expression unit tests, local build, and relevant CI all pass.

crates/rolldown/src/stages/generate_stage/resolve_file_urls.rs
```diff
@@ -0,0 +1,114 @@
...
+    let format_needs_warning =
+      matches!(self.options.format, OutputFormat::Iife | OutputFormat::Umd);
+
+    if !has_hook && !format_needs_warning {
+      return Ok((resolved, warnings));
+    }
+
+    let out_dir = self.options.cwd.as_path().join(&self.options.out_dir);
+
+    for module in &self.link_output.module_table.modules {
```
hyfdev, line 46: This invokes `resolveFileUrl` in module-table/loading order rather than module render order. I reproduced it with one chunk where `entry` imports `dep` and both modules contain a live file URL. Current Rolldown calls the hook as `entry, dep`, even though it emits `dep` before `entry`; Rollup 4.62.2 calls `dep, entry`. A hook returning `call-${++n}` therefore assigns opposite live values.

This is separate from the nested dead-expression limitation: no occurrence is tree-shaken here. `chunk.modules` is already ordered as modules are rendered, so this should be locally fixable by iterating live chunks through `chunk_graph.sorted_chunk_idx_vec`, skipping entries in `post_chunk_optimization_operations`, and then iterating `chunk.modules` instead of `module_table`. A two-module regression fixture should assert both the call order and the output values.

crates/rolldown_ecmascript_utils/src/injected_expression.rs
```diff
@@ -0,0 +1,129 @@
...
+fn parse_in<'ast>(allocator: &'ast Allocator, code: &str) -> Result<Expression<'ast>, String> {
+  // Parenthesized so `{ a: 1 }` parses as an object literal rather than a block
+  // statement. Allocated into the arena so the AST's `&'ast str` slices outlive this
+  // function.
+  //
+  // `parse()` rather than `parse_expression()`: the latter stops at the first
+  // expression and silently discards whatever follows, so `1; 2` yields `1` and
+  // `foo() } evil` yields `foo()`. Parsing a whole program lets trailing garbage
+  // surface as a diagnostic or an extra statement, and fail the build.
+  let source = allocator.alloc_str(&format!("({code})"));
```
hyfdev: Nit (non-blocking): The wrapper rejects a valid expression that ends in a line comment. For example, `resolveFileUrl() { return '42 // trailing comment'; }` fails because `(${code})` becomes `(42 // trailing comment)`, so the closing `)` is commented out. Rollup 4.62.2 emits it successfully.

This is an implementation difference I’m okay leaving for now. If we want to align with Rollup later, using `format!("({code}\n)")` and adding a regression test should preserve the strict single-expression validation while accepting this case.

crates/rolldown/src/stages/generate_stage/resolve_file_urls.rs
hyfdev: Too nit.

crates/rolldown_ecmascript_utils/src/injected_expression.rs
hyfdev: NIt

hyfdev (approved):
LGTM

shulaoda (approved):

sapphi-red:
> * The File URLs page links `resolveFileUrl` to `Interface.PluginContext`; it should link to `/reference/Interface.Plugin#resolvefileurl`.
> * The hook documentation says the default URL works “independent of the output format”, but the current IIFE/UMD fallback warns and emits `{}.url`. This should be qualified to mention the need for an `import.meta.url` polyfill or a custom hook.
> * resolveFileUrl() { return '42 // trailing comment'; }

Fixed these in this PR

crates/rolldown/src/stages/generate_stage/resolve_file_urls.rs
sapphi-red: Fixed in https://github.com/rolldown/rolldown/pull/10330.

PR #10295, opened 2026-07-15
fix(magic-string): return UTF-16 code units from length()

IWANABETHATGUY:
### What this solves

`RolldownMagicString.length()` returns the UTF-8 **byte** count instead of the UTF-16 code unit count that JavaScript — and `magic-string` — report. Any non-ASCII source over-reports:

| input | `length()` before | `magic-string` |
| --- | --- | --- |
| `"abc"` | 3 | 3 |
| `"é"` | **2** | 1 |
| `"🤷"` | **4** | 2 |
| `"中文"` | **6** | 2 |
| `"abc™def"` | **9** | 7 |

There is no panic and no warning — just a silently wrong number, so a plugin doing arithmetic on `length()` (bounds checks, slicing decisions) quietly misbehaves on any file containing an accented character, CJK text, or an emoji.

### Root cause

`MagicString::len()` sums `&str::len()` over the chunk fragments, which is UTF-8 bytes. The binding returned it unconverted:

```rust
pub fn length(&self) -> u32 {
  self.inner.len() as u32
}
```

`length()` was the only index-based method that skipped the `Utf16ToByteMapper` the binding already builds for every other position API (`appendLeft`, `overwrite`, `slice`, …), so the UTF-16 boundary was handled everywhere except here.

### The fix

Add `MagicString::len_utf16`, covering exactly the same span as `len` — chunk content only, global `prepend`/`append` intro/outro excluded, matching `magic-string`'s `length()` — and use it from the binding. `len()` keeps its byte semantics for Rust callers and now documents that explicitly.

### Why this wasn't caught

The vendored upstream conformance suite (`packages/rolldown/tests/magic-string/MagicString.test.ts`) exercises `length` only with ASCII, where bytes and code units coincide. The new tests live next to the other hand-written binding cases in `rolldown-magic-string.test.ts`, which is where the UTF-16↔UTF-8 boundary coverage already sits.

### Tests

Three cases in `packages/rolldown/tests/magic-string/rolldown-magic-string.test.ts`: the non-ASCII table above, `length()` after an edit that inserts a surrogate pair (asserted against `toString().length`), and a guard that the global intro/outro stays excluded. All three fail on `main` and pass here.

### DIFF
--- a/crates/string_wizard/src/magic_string/mod.rs
+++ b/crates/string_wizard/src/magic_string/mod.rs
@@ -99,11 +99,27 @@ impl<'text> MagicString<'text> {
     self.ignore_list
   }
 
-  /// Returns the length of the content within chunks (intro + content + outro per chunk),
-  /// excluding the global intro/outro from `prepend`/`append`.
+  /// Returns the length in UTF-8 bytes of the content within chunks (intro + content + outro
+  /// per chunk), excluding the global intro/outro from `prepend`/`append`.
   /// This aligns with the reference `magic-string` behavior.
+  ///
+  /// Note this counts *bytes*. Callers that need the length JavaScript would report (UTF-16
+  /// code units, as the reference `magic-string` returns) must use [`Self::len_utf16`].
   pub fn len(&self) -> usize {
     self.iter_chunks().flat_map(|c| c.fragments(&self.source)).map(|f| f.len()).sum()
+  }
+
+  /// Returns the length in UTF-16 code units of the content within chunks, excluding the
+  /// global intro/outro — the same range [`Self::len`] covers.
+  ///
+  /// This is what the reference `magic-string` `length()` returns, since JavaScript string
+  /// length is measured in UTF-16 code units.
+  pub fn len_utf16(&self) -> usize {
+    self
+      .iter_chunks()
+      .flat_map(|c| c.fragments(&self.source))
+      .map(|f| f.chars().map(char::len_utf16).sum::<usize>())
+      .sum()
   }
 
   /// Returns `true` if all chunk content (intro + content + outro) is whitespace or empty.
--- a/crates/rolldown_binding/src/types/binding_magic_string.rs
+++ b/crates/rolldown_binding/src/types/binding_magic_string.rs
@@ -745,10 +745,11 @@ impl BindingMagicString<'_> {
 
   #[napi]
   pub fn length(&self) -> u32 {
-    // MagicString::len() returns usize (length of generated output)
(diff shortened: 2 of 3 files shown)


IWANABETHATGUY:
* **#10295** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10295?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10295?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10295).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/string_wizard/src/magic_string/mod.rs
```diff
@@ -99,13 +99,29 @@ impl<'text> MagicString<'text> {
...
   pub fn len(&self) -> usize {
     self.iter_chunks().flat_map(|c| c.fragments(&self.source)).map(|f| f.len()).sum()
   }
 
+  /// Returns the length in UTF-16 code units of the content within chunks, excluding the
+  /// global intro/outro — the same range [`Self::len`] covers.
+  ///
+  /// This is what the reference `magic-string` `length()` returns, since JavaScript string
+  /// length is measured in UTF-16 code units.
+  pub fn len_utf16(&self) -> usize {
```
hyfdev, line 117: **[P2] Count positional insertions on an empty original**

`len_utf16()` still returns `0` after a positional insertion at index `0` on an empty source:

```js
const s = new RolldownMagicString('');
s.appendLeft(0, 'é');
s.toString(); // 'é'
s.length();   // 0; magic-string@0.30.21 returns 1
```

All four positional methods reproduce this; an emoji returns `0` instead of `2`. `by_start_mut(0)` / `by_end_mut(0)` return `None` for an empty source, so these inserts are stored in global intro/outro and excluded here. Upstream keeps its zero-length chunk and counts them. Please keep exact-index-0 positional inserts on that chunk and add tests contrasting them with `append` / `prepend`, which should remain excluded.

hyfdev (approved):

shulaoda (approved):

crates/string_wizard/src/magic_string/mod.rs
IWANABETHATGUY: addressed in https://github.com/rolldown/rolldown/pull/10306

PR #10296, opened 2026-07-15
feat: support `import.meta.ROLLDOWN_FILE_URL_*`

sapphi-red:
Support `import.meta.ROLLDOWN_FILE_URL_*` as an alias of `import.meta.ROLLUP_FILE_URL_*` to make it less confusing.

PR #10297, opened 2026-07-15
feat: `import.meta.ROLLDOWN_FILE_URL_<referenceId>_<urlId>` support to pass context to `resolveFileUrl` hook

sapphi-red:
This PR adds a new feature to `import.meta.ROLLDOWN_FILE_URL_<referenceId>`.

Currently, these file URL references does not have a way to differentiate each references when `resolveFileUrl` hook is called. But that is sometimes needed. For example, in Vite, there are three cases:

- The wasm plugin emits the reference to the wasm binary. In SSR, this reference to the wasm binary should be rendered in a way that accesses as a relative path (as it's a file read). But in other plugins, the wasm binary should be rendered as the URL that can be accessed from the browser (this may sound strange, but it's the current behavior). So the `import.meta.ROLLDOWN_FILE_URL_<referenceId>` needs to be differentiated with type it is.
- Vite has `?inline` query which forces the asset to be output as `data` URL. The same asset can be imported with the query in some places while it can also be imported without the query in other places. This means the same asset needs to be referenced differently depending on the reference.
- Vite never inlines assets specified as the entrypoint. This information is tied to a single reference.

[rendered docs](https://github.com/rolldown/rolldown/blob/07-15-feat_import.meta.rolldown_file_url__referenceid___urlid_support_to_pass_context_to_resolvefileurl_hook/docs/apis/plugin-api/file-urls.md#passing-a-urlid)

In this PR, I made the `urlId` generated on the plugin side, but maybe that should be generated on the Rolldown side so that a conflict never happens. This can be done by adding `this.generateFileUrlId`.

The API can also be expanded to store the metadata in the Rolldown side. Currently, The metadata lives on the plugin side, so the referenceId emit and the `resolveFileUrl` hook needs to be coupled. This works for Vite's case, but won't work in general. For now, I think it's ok to go without it.

refs https://github.com/vitejs/vite/pull/22962

### DIFF
--- a/crates/rolldown/src/utils/file_url.rs
+++ b/crates/rolldown/src/utils/file_url.rs
@@ -1,14 +1,105 @@
 /// Prefixes recognized on `import.meta.<prefix><referenceId>` file URL references.
 ///
 /// `ROLLDOWN_FILE_URL_` is a rolldown-specific alias of Rollup's `ROLLUP_FILE_URL_`.
-pub const FILE_URL_PREFIXES: [&str; 2] = ["ROLLDOWN_FILE_URL_", "ROLLUP_FILE_URL_"];
+const ROLLDOWN_FILE_URL_PREFIX: &str = "ROLLDOWN_FILE_URL_";
+const ROLLUP_FILE_URL_PREFIX: &str = "ROLLUP_FILE_URL_";
 
-/// Returns the `<referenceId>` when `name` begins with a recognized file URL prefix.
-pub fn strip_file_url_prefix(name: &str) -> Option<&str> {
-  FILE_URL_PREFIXES.iter().find_map(|prefix| name.strip_prefix(prefix))
+/// Reference ids are produced by `FileEmitter::assign_reference_id` as the base64url
+/// encoding of a 128-bit xxhash (with `-` swapped for `$`), so they are always exactly
+/// this many characters. The optional `_<urlId>` part of
+/// `import.meta.ROLLDOWN_FILE_URL_<referenceId>_<urlId>` therefore begins right after them.
+const REFERENCE_ID_LEN: usize = 22;
+
+/// A parsed `import.meta.<prefix><referenceId>[_<urlId>]` reference.
+#[derive(Debug, PartialEq, Eq)]
+pub struct FileUrlReference<'a> {
+  pub reference_id: &'a str,
+  pub url_id: Option<&'a str>,
+}
+
+/// Parses `name` when it begins with a recognized file URL prefix.
+///
+/// For `ROLLDOWN_FILE_URL_<referenceId>`, an optional `_<urlId>` suffix is split off and
+/// exposed as [`FileUrlReference::url_id`]. The `ROLLUP_FILE_URL_` alias is parsed for
+/// backward compatibility with Rollup and never carries a `urlId`.
+pub fn strip_file_url_prefix(name: &str) -> Option<FileUrlReference<'_>> {
+  if let Some(rest) = name.strip_prefix(ROLLDOWN_FILE_URL_PREFIX) {
+    // Reference ids are fixed-length, so the `_<urlId>` suffix (if any) begins right after
+    // them. `rest` is ASCII (base64url + `$`), so byte indexing lands on char boundaries.
+    if rest.len() > REFERENCE_ID_LEN && rest.as_bytes()[REFERENCE_ID_LEN] == b'_' {
+      let url_id = &rest[REFERENCE_ID_LEN + 1..];
+      return Some(FileUrlReference {
+        reference_id: &rest[..REFERENCE_ID_LEN],
+        url_id: if url_id.is_empty() { None } else { Some(url_id) },
+      });
+    }
+    return Some(FileUrlReference { reference_id: rest, url_id: None });
+  }
+  name
+    .strip_prefix(ROLLUP_FILE_URL_PREFIX)
+    .map(|reference_id| FileUrlReference { reference_id, url_id: None })
 }
 
 /// Whether `name` begins with a recognized file URL prefix.
 pub fn starts_with_file_url_prefix(name: &str) -> bool {
   strip_file_url_prefix(name).is_some()
+}
+
+#[cfg(test)]
+mod test {
+  use super::{FileUrlReference, starts_with_file_url_prefix, strip_file_url_prefix};
+
+  #[test]
+  fn strips_rolldown_prefix_without_url_id() {
+    // A 22-char reference id with no `_<urlId>` suffix.
+    let reference_id = "aaaaaaaaaaaaaaaaaaaaaa";
+    assert_eq!(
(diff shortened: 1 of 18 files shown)


sapphi-red:
* **#10297** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10297?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10297?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10296** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10296?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#10330](https://github.com/rolldown/rolldown/pull/10330) <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10330?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#10291** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10291?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10214** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10214?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10297).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

docs/apis/plugin-api/file-urls.md
```diff
@@ -116,3 +116,37 @@ export const size = 6;
 :::
 
 If you build this code, both the main chunk and the worklet will share the code from `config.js` via a shared chunk. This enables us to make use of the browser cache to reduce transmitted data and speed up loading the worklet.
+
+## Passing a `urlId`
+
+Rolldown extends the syntax with an optional `urlId` (`import.meta.ROLLDOWN_FILE_URL_referenceId_urlId`). The `urlId` is an arbitrary identifier that is forwarded to the [`resolveFileUrl`](/reference/Interface.PluginContext#resolvefileurl) hook as `args.urlId`, so a single plugin can resolve the same emitted file differently depending on where it is referenced from:
```
hyfdev: This link points at `Interface.PluginContext#resolvefileurl`, but `resolveFileUrl` is a plugin hook, not a `PluginContext` method — the generated anchor lives under `Interface.Plugin#resolvefileurl` (which line 3 of this page already uses). As written, readers land on the `PluginContext` page with no matching section.

The example below also calls `path.basename` and `fs.readFileSync` without imports, while the example at the top of this page includes them.

packages/rolldown/tests/fixtures/plugin/resolve-file-url/url-id/_config.ts
```diff
@@ -0,0 +1,51 @@
...
+            `export const b = import.meta.ROLLDOWN_FILE_URL_${referenceId};`,
+            `export const c = import.meta.ROLLUP_FILE_URL_${referenceId};`,
+          ].join('\n');
+        },
+      },
+      {
+        name: 'resolve-file-url',
+        resolveFileUrl(args) {
+          seen.push(args.urlId);
+          return JSON.stringify(args.urlId ?? null);
```
hyfdev, line 36: Because this hook returns a value for every occurrence, no test exercises the default rewrite for a urlId-carrying reference — the path where the finalizer strips `_myUrlId` and resolves the emitted file to `new URL(<relative>, import.meta.url).href`. That is also the path behind the docs claim that default resolution ignores `urlId`.

Returning `null` for one of the occurrences (and asserting the `new URL(...)` output) would cover it.

crates/rolldown/src/utils/file_url.rs
```diff
@@ -1,14 +1,100 @@
...
+/// exposed as [`FileUrlReference::url_id`]. The `ROLLUP_FILE_URL_` alias is parsed for
+/// backward compatibility with Rollup and never carries a `urlId`.
+pub fn strip_file_url_prefix(name: &str) -> Option<FileUrlReference<'_>> {
+  if let Some(rest) = name.strip_prefix(ROLLDOWN_FILE_URL_PREFIX) {
+    // Reference ids are fixed-length, so the `_<urlId>` suffix (if any) begins right after
+    // them. `rest` is ASCII (base64url + `$`), so byte indexing lands on char boundaries.
+    if rest.len() > REFERENCE_ID_LEN && rest.as_bytes()[REFERENCE_ID_LEN] == b'_' {
+      return Some(FileUrlReference {
+        reference_id: &rest[..REFERENCE_ID_LEN],
+        url_id: Some(&rest[REFERENCE_ID_LEN + 1..]),
```
hyfdev: `import.meta.ROLLDOWN_FILE_URL_<id>_` (trailing lone `_`) parses here as `url_id: Some("")`, so the hook receives `urlId: ''` rather than `undefined` — the TS docs describe `urlId` as `undefined` when the reference has none. Only reachable with a hand-written degenerate reference, but treating the empty slice as `None` would keep `Some(_)` meaning "a urlId is present".

hyfdev (approved):
The fixed-length split is sound: reference ids are always exactly 22 chars of base64url from `assign_reference_id`, and the new lockstep test in `file_emitter.rs` guards that invariant. Per-occurrence results keyed by `(ModuleIdx, NodeId)` mean the same file referenced with different `urlId`s resolves independently, and the `ROLLUP_FILE_URL_` alias keeps Rollup semantics by never parsing a suffix.

Generating the `urlId` on the plugin side (rather than a Rolldown-side `generateFileUrlId`) covers the use cases described: the plugin controls both the reference spelling and the hook, so `urlId` conflicts stay within one plugin's namespace.

Left a few small notes on the docs and test coverage; nothing blocking.

packages/rolldown/tests/fixtures/plugin/resolve-file-url/url-id/_config.ts
sapphi-red: Added the case

crates/rolldown/src/utils/file_url.rs
sapphi-red: Added a test case

sapphi-red:
marked as experimental (may change in minors)

sapphi-red:
### Merge activity

* **Jul 22, 10:39 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 22, 10:43 AM UTC**: `sapphi-red` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 22, 10:49 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10299, opened 2026-07-15
fix(magic-string): don't report a change when edits cancel out

IWANABETHATGUY:
> Stacked on #10295 — review that first; this PR's diff is just the second commit.

### What this solves

`hasChanged()` reports `true` for edits that cancel out, i.e. when the generated output is byte-identical to the source:

```js
s.remove(0, 1);
s.prepend('a');
s.toString();     // 'abc'  — identical to the original
s.hasChanged();   // true   — magic-string says false
```

Plugins commonly gate on `hasChanged()` to decide whether to hand back a transformed result and a sourcemap at all, so a false positive means pointless downstream work for a module that wasn't really touched.

### Root cause

```rust
self.source.len() != self.len() || self.source.as_ref() != self.to_string()
```

The first clause is meant to be a cheap "definitely changed" fast path for the string comparison after it — but the two measure different spans. `len()` counts **only chunk content**, while `to_string()` also includes the global intro/outro from `prepend`/`append`.

So any edit that shifts bytes between the chunks and the global intro/outro trips the fast path even when the two sides balance out. Above: `remove(0, 1)` drops a byte from the chunks (`len()` 3 → 2) and `prepend("a")` puts it back in the intro, which `len()` cannot see — `3 != 2` short-circuits to `true`.

`magic-string` has no fast path at all; it is just `original !== toString()`.

### The fix

Compare the whole output — global intro/outro included — so the fast path covers the same span as the comparison it short-circuits. Kept as a length check rather than dropped entirely, since it still avoids the `to_string()` allocation in the common "definitely changed" case.

### Tests

Three cases, one per branch of the resulting condition:

| case | branch exercised |
| --- | --- |
| `remove(0,1)` + `prepend("a")` → `false` | lengths match, strings match — **the bug** |
| `overwrite(0,3,"XYZ")` → `true` | lengths match, falls through to the string comparison |
| `prepend("x")` → `true` | the fast path itself fires (output longer than source) |

Only the first fails without this change. The two `is true` cases are deliberate: this PR *removes* a fast path, and a suite pinning only the bug case would equally accept a `has_changed()` that always returns `false` — I verified that by mutation, and those two are what catch it.

### DIFF
--- a/crates/string_wizard/src/magic_string/mod.rs
+++ b/crates/string_wizard/src/magic_string/mod.rs
@@ -122,6 +122,14 @@ impl<'text> MagicString<'text> {
       .sum()
   }
 
+  /// Returns the length in UTF-8 bytes of the whole generated output, including the global
+  /// intro/outro. Equivalent to `self.to_string().len()` without the allocation.
+  fn output_len(&self) -> usize {
+    self.intro.iter().map(|f| f.len()).sum::<usize>()
+      + self.len()
+      + self.outro.iter().map(|f| f.len()).sum::<usize>()
+  }
+
   /// Returns `true` if all chunk content (intro + content + outro) is whitespace or empty.
   /// This aligns with the reference `magic-string` behavior where `isEmpty()` uses `.trim()`.
   pub fn is_empty(&self) -> bool {
@@ -129,8 +137,13 @@ impl<'text> MagicString<'text> {
   }
 
   /// Indicates if the string has been changed.
+  ///
+  /// The length check is only a fast path for "definitely changed", so it has to measure the
+  /// same span as the string comparison below it: the whole output, global intro/outro
+  /// included. Comparing against [`Self::len`] instead would report a change for edits that
+  /// cancel out, e.g. `remove(0, 1)` followed by `prepend("a")`.
   pub fn has_changed(&self) -> bool {
-    self.source.len() != self.len() || self.source.as_ref() != self.to_string()
+    self.output_len() != self.source.len() || self.source.as_ref() != self.to_string()
   }
 
   /// Returns the last character of the generated string, or `None` if empty.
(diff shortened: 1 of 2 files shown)


IWANABETHATGUY:
* **#10299** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10299?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10299?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10299).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyfdev (approved):

IWANABETHATGUY:
### Merge activity

* **Jul 15, 3:06 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 15, 5:52 PM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 15, 6:00 PM UTC**: The [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) couldn't merge this PR because **it had merge conflicts**.
* **Jul 16, 2:40 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 16, 2:40 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 16, 2:46 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10301, opened 2026-07-15
fix(magic-string): throw instead of panicking when append/prepend splits an edited chunk

IWANABETHATGUY:
> Stacked on #10299 → #10295 — review those first; this PR's diff is just the third commit.

### What this solves

Four `MagicString` methods abort the process instead of throwing. Three lines of plugin code are enough:

```js
const s = new RolldownMagicString('abcdef');
s.overwrite(0, 6, 'XYZ');
s.appendLeft(3, '!');   // Rolldown panicked. This is a bug in Rolldown, not your code.
```

```
thread '<unnamed>' panicked at crates/string_wizard/src/magic_string/append.rs:23:39:
Cannot split a chunk that has already been edited
```

This is not a recoverable error — a Rust panic across the FFI boundary takes the whole process down. A vitest worker running the repro dies outright, and there is nothing a plugin author can catch. `magic-string` throws a plain `Error` for the same input.

### Root cause

`append_left`, `append_right`, `prepend_left` and `prepend_right` each called `.expect(...)` on the result of `by_end_mut`/`by_start_mut`:

```rust
// Note: by_end_mut only errors when splitting an already-edited chunk,
// but append operations don't require splitting edited chunks in practice.
// We use expect here as this is an internal invariant.
match self.by_end_mut(text_index).expect("append_left: unexpected split error") {
```

The comment is simply wrong. `by_end_mut` calls `split_at`, which calls `Chunk::split`, which genuinely returns `Err("Cannot split a chunk that has already been edited")` — and appending *inside* a range that was previously overwritten reaches it directly.

The repo already knew, in a way: the conformance skip list carries `'split point', // split point errors cause panic` and `'should replace then remove', // causes split chunk panic`.

### The fix

Return `Result` from the four methods and propagate. This cascades:

- `indent_with` calls `prepend_right`, so it propagates too. Indentation only ever splits *unedited* chunks (the call sits in the `edited_content.is_none()` branch), so the error is unreachable there — but threading it through beats asserting an invariant that this very change shows can be wrong. It also makes the crate uniform: `update`, `remove`, `relocate`, `snip`, `reset` and `slice` already return `Result`.
- The binding maps each to `napi::Error`. No TypeScript surface change — napi maps `Result<T>` to `T` and throws on `Err`.
- Existing `string_wizard` tests gain `.unwrap()`; no behaviour change there.

Also replaces the `unwrap` on the chunk walk in `update_with` with an explicit error. Not reachable today (the binding bounds-checks first), but that function already returns `Result`, so propagating costs nothing.

### Scope note

I probed every other `unwrap`/`expect` reachable from the binding rather than converting them wholesale. Only these four are reachable; `update.rs` and `movement.rs` are already guarded by the binding's bounds checks and the split-point check, and return clean errors (`Invalid start character index`, `Cannot move a selection inside itself`). Converting a panic to an error is only safe where the function has not already mutated shared state — so the remaining sites are handled deliberately, not swept.

### Tests

Five cases in `rolldown-magic-string.test.ts`. **Each one aborts the test worker on the parent commit**, so they have to be run one per process to see it — a batch run just reports the file as crashed.

| test | panic site it covers |
| --- | --- |
| `appendLeft() throws instead of panicking` | `append.rs:23` |
| `appendRight() throws instead of panicking` | `append.rs:46` |
| `prependLeft() throws instead of panicking` | `prepend.rs:15` |
| `prependRight() throws instead of panicking` | `prepend.rs:26` |
| `leaves the instance usable after the error` | asserts `toString()` still works afterwards |

The four look near-identical but each covers a distinct call site. The fifth pins that the failed operation left the instance intact — a different property, and the one that would catch "propagates the error but corrupts state".

### DIFF
--- a/crates/rolldown_binding/src/types/binding_magic_string.rs
+++ b/crates/rolldown_binding/src/types/binding_magic_string.rs
@@ -155,6 +155,20 @@ impl Utf16ToByteMapper {
     self.get(utf16_offset).map(|e| e.byte_offset)
   }
 
+  /// For a UTF-16 index that lands on a low surrogate — i.e. strictly inside a surrogate
+  /// pair — returns the byte span of the supplementary character containing it.
+  /// Character-boundary indices return `None`.
+  fn surrogate_interior_char_span(&self, utf16_offset: u32) -> Option<(u32, u32)> {
+    let entry = self.get(utf16_offset)?;
+    if !entry.is_low_surrogate() {
+      return None;
+    }
+    // A low surrogate is always preceded by its high surrogate, whose byte offset is the
+    // character's start; the low surrogate's own byte offset is the character's end.
+    let start = self.entries[utf16_offset as usize - 1].byte_offset;
+    Some((start, entry.byte_offset))
+  }
+
   /// Returns the UTF-16 code unit count of the original string.
   /// This matches JavaScript's `String.prototype.length`.
   fn utf16_len(&self) -> i64 {
@@ -579,6 +593,21 @@ impl BindingMagicString<'_> {
     index.saturating_add(self.offset)
   }
 
+  /// A UTF-16 index inside a surrogate pair has no byte equivalent. `magic-string` splits
+  /// the chunk there (into lone surrogates, which UTF-8 cannot hold), so on unedited content
+  /// we round to the character boundary instead — but on an edited chunk `magic-string`
+  /// throws, and rounding would swallow that error and place content at a boundary the
+  /// caller never named. `utf16_index` must already have `self.offset` applied.
+  fn reject_surrogate_split_of_edited_chunk(&self, utf16_index: u32) -> napi::Result<()> {
+    if let Some((start, end)) = self.utf16_to_byte_mapper.surrogate_interior_char_span(utf16_index)
+    {
+      if self.inner.is_range_within_edited_chunk(start, end) {
+        return Err(napi::Error::from_reason("Cannot split a chunk that has already been edited"));
+      }
+    }
+    Ok(())
+  }
+
   #[napi]
   pub fn replace<'s>(
     &'s mut self,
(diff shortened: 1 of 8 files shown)


IWANABETHATGUY:
* **#10301** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10301?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10301?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10301).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown_binding/src/types/binding_magic_string.rs
```diff
@@ -675,7 +675,7 @@ impl BindingMagicString<'_> {
     // Match original magic-string: out-of-bound indices fall through to append_intro
     match self.utf16_to_byte_mapper.utf16_to_byte(self.apply_offset_u32(index)?) {
       Some(byte_index) => {
-        self.inner.append_left(byte_index, content);
+        self.inner.append_left(byte_index, content).map_err(napi::Error::from_reason)?;
```
hyfdev, line 713: Non-blocking edge case: this error path is still bypassed for a UTF-16 position inside a surrogate pair:

```js
const s = new RolldownMagicString('🤷');
s.overwrite(0, 2, 'X');
s.appendLeft(1, '!');
```

`magic-string@0.30.21` throws `Cannot split a chunk that has already been edited`, but Rolldown succeeds and produces `X!`. `utf16_to_byte(1)` maps the low-surrogate position to the byte offset after the emoji, so `append_left` sees the end boundary instead of an index inside the edited chunk. Could we preserve the split error for low-surrogate positions and add this case for all four methods?

hyfdev (approved):

IWANABETHATGUY:
### Merge activity

* **Jul 16, 3:56 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 16, 3:57 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 16, 4:03 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #2919, opened 2024-11-25
chore(node): rename `runtime-without-comments.js` with `runtime-base.js`

hyf0:
### Description

Part of https://github.com/rolldown/rolldown/issues/2685.

PR #375, opened 2023-11-23
feat(rust): rendering runtime module in snapshot testing

hyf0:
### Test Plan

---

PR #10302, opened 2026-07-15
chore(runtime): remove unused `runtime/index.js`

IWANABETHATGUY:
### Description

`crates/rolldown/src/runtime/index.js` is not part of the build, and hasn''t been for a long time. This PR deletes it.

The runtime module is assembled in [`runtime_module_task.rs`](https://github.com/rolldown/rolldown/blob/main/crates/rolldown/src/module_loader/runtime_module_task.rs#L33-L36) from exactly four `include_str!`ed files:

```rust
const RUNTIME_BASE_JS: &str = include_str!("../runtime/runtime-base.js");
const RUNTIME_TAIL_JS: &str = include_str!("../runtime/runtime-tail.js");
const RUNTIME_HEAD_NODE_JS: &str = include_str!("../runtime/runtime-head-node.js");
const RUNTIME_TAIL_NODE_JS: &str = include_str!("../runtime/runtime-tail-node.js");
```

`index.js` is not among them. Nothing else in the repo references it either — there''s no `build.rs` in the crate, and no path reference to it in Rust, JS, or `knip.jsonc`. Its only remaining coverage is incidental: the sibling `tsconfig.json` globs `./**/*`, so it gets type-checked and formatted alongside the files that actually ship.

### How it got here

It''s the original esbuild runtime port from #37. #375 added `runtime-without-comments.js` as the copy actually fed to the bundler; that file was renamed to `runtime-base.js` in #2919 and later split with `runtime-tail.js`. `index.js` was never wired back up — it survived as a copy that edits were hand-mirrored into.

### Why delete rather than leave it

The mirroring has fallen behind, so the file now actively misinforms:

| | `runtime-base.js` (ships) | `index.js` (dead) |
|---|---|---|
| `cb = null` release in `__commonJS` (#9067) | present | **missing** |
| `__exportAll` + `Symbol.toStringTag` | present | still the old `__export` |
| `__require` error text | rewritten to link the rolldown docs (in `runtime-tail.js`) | still esbuild''s "Dynamic require of …" |

Someone grepping for a runtime helper can land here and draw the wrong conclusion about what ships — the missing #9067 fix reads as a live memory leak. It isn''t, because the file is inert. Deleting it removes the trap and the ongoing cost of keeping a second copy in sync.

### Test Plan

Deletion only; no shipped output changes, so no snapshots move. `include_str!` resolves at compile time, so a stale reference would break the build:

- `cargo check -p rolldown --all-targets` — passes
- `tsc --noEmit -p crates/rolldown/src/runtime/tsconfig.json` — passes


hyfdev (approved):

hyfdev:
### Merge activity

* **Jul 15, 3:17 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 15, 3:17 PM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 15, 3:24 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10305, opened 2026-07-15
feat(dev): support `hotUpdate` hook

h-a-n-a:
Adds a dev-only `hotUpdate` plugin hook to the HMR engine. It works like Vite's hook of the same name. The Vite side is vitejs/vite#22956, which runs Vite's `hotUpdate` / `handleHotUpdate` hooks on top of this hook in bundled dev.

## Problem

When a file changes, the engine can only find modules the graph knows about: the file's own module, and modules that watched the file with `addWatchFile`. Plugins often know more. A config file may affect many modules. A content file may not be a module at all. Vite plugins express this knowledge with the `hotUpdate` hook. Without the same hook, these plugins cannot work on the rolldown dev engine.

## Design

```mermaid
flowchart LR
  A[file change] --> B[default affected set<br/>own module and transform deps]
  B --> C[hotUpdate chain<br/>each plugin may replace the set]
  C --> D[re-fetch, walk, patch<br/>existing steps, unchanged]
```

The hook edits the input of the update pipeline. It does not change how the pipeline works. It runs once per changed file, after the engine builds the default affected set, and before any module is re-fetched. Everything after that point — re-fetch, boundary walk, patch, delivery — stays the same.

The main design rules:

- **Vite behavior, engine-level types.** The chain works like Vite's: plugins run in hook order, each plugin gets the current set and may replace it, an empty array suppresses the update for this file, and no return passes the set through. But the engine uses plain module ids, not module node objects. Node objects belong to the framework layer — the adapter in vitejs/vite#22956 translates between the two.
- **The engine does not override an explicit plugin choice.** Modules returned by a hook skip the "output unchanged" check. That check is only safe when "same output" means "nothing happened". That is true for a normal edit, but not for a hook result — the change may live outside the module's code. Example: a module reads `config.json` at runtime. When a plugin maps a `config.json` edit to that module, re-running the module is the whole point, and its unchanged code proves nothing. Vite always ships hook-returned modules for the same reason.
- **Engine internals stay hidden.** Hooks never see lazy-compilation proxies or runtime modules, and cannot return them. Unknown ids are dropped with a debug log. A plugin cannot break the graph.
- **A hook cannot block error recovery.** Files queued for retry after a failed build are added before the suppress decision. So clients stuck on an error overlay always get the recovery build.
- **No hook, no cost.** If no plugin registers the hook, the whole path is skipped. If plugins return nothing, the default flow — including the "output unchanged" check — works as before.
- **No `read()` on the engine hook.** Reading the changed file from disk is a framework-layer concern — the disk content may not match what the plugin pipeline produces. Per review, `read` lives only in the Vite adapter (vitejs/vite#22956), which keeps Vite's retry behavior (vitejs/vite#610).
- **No time limit on the chain.** Same as Vite: a hook that never finishes blocks that update.

## What this enables

Through the Vite adapter, Vite plugins with either hook run on bundled dev without changes. The browser tests in this PR cover the common plugin patterns: replacing the set (tailwind-style config mapping), buffering with `moduleGraph.invalidateModule` plus an empty return, custom messages with `read()` and `hot.send` (markdown-pages style), the legacy shared-context `read` chain (unocss with plugin-vue), file-to-submodule expansion (SFC style), and `invalidateAll` as a full rebuild plus reload.

## Tests

- Unit: hidden-id filtering (runtime and lazy-proxy ids are hidden; real files and virtual modules are not).
- Browser end-to-end, six playgrounds:
  - `hmr-hot-update-hook` — the raw hook: replace, suppress, unknown ids dropped, an empty default set for a file with no module, decline, proof that a hook-selected module ships even when its output did not change, and proof that a suppressing hook cannot block error recovery.
  - `hmr-hot-update-hook-chain` — each plugin receives the set as edited by the plugins before it.
  - `hmr-hot-update-hook-delete` — delete events reach the hook; a replacement returned for a delete ships as-is.
  - `hmr-hot-update-hook-error` — a throwing hook fails the round, and the failed edit is not retried.
  - `hmr-hot-update-hook-vite` — the Vite adapter (8 specs), covering the plugin patterns listed above.
  - `hmr-delete-self-watched` — a module that watches its own file adds no extra failure mode on delete.
- A skipped engine test for #10487 (deleting a still-imported file on the raw engine): it asserts the desired behavior and is un-skipped by the resolver-cache fix.

## Not in this PR, planned next

- Directory support for `addWatchFile` (Rollup allows directories; needed to watch files that do not exist yet).
- `import.meta.glob` updates in the native plugin, built on top of that.


h-a-n-a:
* **#10305** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10305?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10305?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10305).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown/src/hmr/hmr_stage.rs
```diff
@@ -89,27 +100,82 @@ impl<'a, Fs: FileSystem + Clone + 'static> HmrStage<'a, Fs> {
...
         }
       }
 
-      // Check if any modules have this file as a transform dependency
-      for entry in self.plugin_driver.transform_dependencies.iter() {
-        let module_idx = *entry.key();
-        let deps = entry.value();
-        if deps.contains(&changed_file_path) {
+      for module_idx in affected_modules {
+        if *event == WatcherChangeKind::Delete && Some(module_idx) == own_module_idx {
```
hyfdev, line 184: When a module registers its own file with `addWatchFile`, deleting that file now behaves differently even with no `hotUpdate` hook installed: the baseline's transform-dep loop re-inserted the module itself, so the engine refetched it and surfaced the scan failure, while the dedup into `affected_modules` plus this guard now routes it to importer expansion only.

Skipping the doomed refetch of a deleted module looks like the better behavior, but it is an observable no-hook-path change, and a delete-event fixture would pin it as intended rather than accidental. This does not block this approval.

crates/rolldown/src/hmr/hmr_stage.rs
```diff
@@ -89,27 +100,82 @@ impl<'a, Fs: FileSystem + Clone + 'static> HmrStage<'a, Fs> {
...
+          affected_modules.insert(module_idx);
+        }
+      }
+
+      let mut hook_replaced = false;
+      if hot_update_hook_registered {
+        // Plugins receive raw module ids and may replace the set; an empty return suppresses
+        // this file's update. The chain also runs when the default set is empty — content
+        // plugins claim files no module maps to. Ids the graph doesn't know are dropped.
+        let default_ids = affected_modules
```
hyfdev, line 151: `args.file` reaches the hook slash-normalized (line 115) while these ids keep native separators on Windows, so one args object mixes two path conventions. A plugin deriving paths from `ctx.file` — as both playgrounds do with `path.join(path.dirname(ctx.file), …)` — then compares mixed forms; the specs only survive on Windows because they match with `endsWith`.

Normalizing the default ids with `to_slash` as well would hand the hook one consistent convention; the return path already tolerates that, since the lookup keys are slashed and returned ids are re-slashed at line 157.

A related observability point on the same construction: when several modules share a transform dependency, their order here follows `transform_dependencies` (a DashMap) iteration, so `args.modules` order can vary run to run. That order was engine-internal before (later re-sorted); this hook makes it plugin-visible, and Vite's equivalent set has stable insertion order.

crates/rolldown/tests/rolldown/topics/hmr/hot_update_hook/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+    if args.id.ends_with("entry.js") {
+      // Register both control files as watched transform dependencies, so changing them maps
+      // to `entry.js` in the engine's default affected set.
+      ctx.add_watch_file(ctx.cwd().join("./input.txt").normalize().to_str().unwrap());
+      ctx.add_watch_file(ctx.cwd().join("./suppress.txt").normalize().to_str().unwrap());
+    }
+    Ok(None)
+  }
+
+  async fn hot_update(
```
hyfdev: This fixture pins replace, suppress, and the unchanged-output exemption; the rest of the documented contract has no engine-level test: delete and create events through the chain, unknown-id dropping (which silently becomes suppression when it empties the set), a multi-plugin chain where a later plugin re-adds modules after an empty return, a multi-file batch with mixed outcomes (one file suppressed while another replaces), a hook that declines on a real module edit (pinning that the default flow, including unchanged-output suppression, still applies), and a throwing hook.

The throw path has a subtlety worth pinning: a hook error adds nothing to `pending_rescans` (a failed scan does), so the batch's edits reach the graph only when a later change retries them — consistent with `watchChange` semantics, though "a hook error fails the whole batch, like a build error" in the internal docs reads stronger than that. Follow-up tests are fine here.

hyfdev (approved):
The engine-level contract matches Vite's where it must — sequential replace-chain, per-file suppression, hook-returned modules shipping even when their rebuilt output is byte-identical — while keeping plain module ids at the boundary, and the no-hook path is otherwise unchanged from the baseline. The fixture proves the suppression exemption the design leans on, and the two playgrounds cover the raw contract and the Vite adapter patterns end to end.

The inline notes are follow-up material: one observable no-hook delete corner, a Windows path-convention nit, and engine-level test gaps. One more without a good inline anchor: a plugin from `defineParallelPlugin` that declares `hotUpdate` sets the usage bit — so the per-file chain runs — but `ParallelJsPlugin` never forwards the hook, so it is silently ignored, same as `closeWatcher` today; masking the bit there or documenting the limitation would close that gap.

— Review by @hyfdev, assisted by Claude Fable 5 (`claude-fable-5`)

packages/rolldown/src/plugin/bindingify-watch-hooks.ts
```diff
@@ -8,6 +9,62 @@ import {
...
+          args.pluginContextData,
+          args.onLog,
+          args.logLevel,
+          args.watchMode,
+        ),
+        {
+          type: hookArgs.kind as ChangeEvent,
+          file: hookArgs.file,
+          modules: hookArgs.modules,
+          read: () => readModifiedFile(hookArgs.file),
```
sapphi-red: I think we can drop `read` on Rolldown side and add this on Vite side when wrapping `hotUpdate` in Vite. Reading the file directly is always wrong as the file may not contain the information needed (a plugin might be injecting it). For example, [this line in plugin-vue](https://github.com/vitejs/vite-plugin-vue/blob/b2b559dd01b245be5b09c954ac112833d16c3b8b/packages/plugin-vue/src/handleHotUpdate.ts#L39-L40) expects the file to be a Vue component. But, this may not be true.

Btw, is plugin-vue still able to do what it wants to do without using `read`?

packages/rolldown/src/plugin/index.ts
```diff
@@ -658,10 +648,43 @@ export interface FunctionPluginHooks {
...
 }
 
 export type ChangeEvent = 'create' | 'update' | 'delete';
 
+export interface HotUpdateOptions {
+  type: ChangeEvent;
+  /** Normalized absolute path of the changed file. */
+  file: string;
+  /** The affected module ids as currently computed (raw module ids). */
+  modules: string[];
```
sapphi-red: The type of `modules` and the return value is not compatible with Vite's one. It feels confusing to have a different shape. Vite normally only extends the hooks.

h-a-n-a: You're right, and I don't think any shape on the Rolldown side can make the two hooks truly compatible. `EnvironmentModuleNode` is a live object in Vite's module graph: plugins rely on node identity (the node from `getModulesByFile` is the same object they received), on `url` and `type`, and on Vite-managed state like `isSelfAccepting`. The engine can't produce those, so a translation layer in Vite is needed no matter what crosses the binding.

What's your take on this?

packages/rolldown/src/plugin/bindingify-watch-hooks.ts
h-a-n-a: Agreed — I'll drop `read` here and let Vite add it in its wrapper. The wrapper already builds its own `read`, so nothing changes for Vite plugins.

On plugin-vue: yes, it still works, because it gets `read` from the wrapper. It does need the new content at `hotUpdate` time: it parses the file into a new descriptor ([L39-L40](https://github.com/vitejs/vite-plugin-vue/blob/b2b559dd01b245be5b09c954ac112833d16c3b8b/packages/plugin-vue/src/handleHotUpdate.ts#L39-L40)), then diffs it against the previous one block by block — script ([L53-L55](https://github.com/vitejs/vite-plugin-vue/blob/b2b559dd01b245be5b09c954ac112833d16c3b8b/packages/plugin-vue/src/handleHotUpdate.ts#L53-L55)), template ([L58-L71](https://github.com/vitejs/vite-plugin-vue/blob/b2b559dd01b245be5b09c954ac112833d16c3b8b/packages/plugin-vue/src/handleHotUpdate.ts#L58-L71)), styles ([L75-L105](https://github.com/vitejs/vite-plugin-vue/blob/b2b559dd01b245be5b09c954ac112833d16c3b8b/packages/plugin-vue/src/handleHotUpdate.ts#L75-L105)) — to pick only the affected modules (e.g. a style-only edit updates just the style module, so component state is kept). That diff has to happen before the update is sent, and at that point the new content only exists on disk. So it needs `read` (or an fs read of its own) — but that's a Vite-layer concern, and the wrapper is the right place for it.

h-a-n-a: Done — `read` is removed from the Rolldown-side options in bafd7077a. The Vite adapter keeps its own `read` (vitejs/vite#22956, updated in 67e9251).

sapphi-red: > Btw, is plugin-vue still able to do what it wants to do without using `read`?

I mean, if we remove `read` from both Rolldown and Vite, how should plugin-vue implement it's logic? and what needs to be exposed from Rolldown?

packages/rolldown/src/plugin/index.ts
sapphi-red: I think the only way is to introduce an flag in the top-level of the plugin or add a new hook. But it also depends on the answer of https://github.com/rolldown/rolldown/pull/10305#discussion_r3644287281.

h-a-n-a: Sorry, I may not understand this correctly. AFAIK, there's two parts of this: whether we need `read()` and in which API are we going to implement `hotUpdate` hook. 

For the first one, it depends on the answer of https://github.com/rolldown/rolldown/pull/10305#discussion_r3644287281. I'm not sure if I should investigate and implement now, since removing the API is a breaking change.

For the second one, what exactly does introducing an flag in the top-level mean? and why do we need to make Rolldown's API exactly the same as Vite's given that we can add a wrapper (adapter) in Vite?

sapphi-red: I had an assumption that we have to expose this hook publicly. With that in mind, it is better to avoid introducing a new hook after it turned out we need a different one for the plugin-vue's usecase. That's the reason I asked about how to achieve plugin-vue's logic without `read`. The flag in the top-level is something like:
```js
export const p = {
  name: 'my-plugin',
  hotUpdateVersion: 'v2',
  hotUpdate() {}
}
```
If the hook is public from the Rolldown side as well, it would be confusing if Vite and Rolldown had a different interface. It's not about the technical feasibility.

***But*** that assumption is wrong and we can hide it. I think we should hide it instead so that it's only exposed through Vite. That can solve all these issues temporarily.

hyfdev (approved):
Re-reviewed after the `read` drop. The engine contract still holds: sequential plain-id replace-chain, per-file suppress, hook-selected modules shipping past the unchanged-output check, and the no-hook path otherwise intact.

On the public-surface discussion: string ids at the engine boundary are the right shape for the Vite adapter. Hiding the TypeScript export so only Vite surfaces a plugin-facing hook would address the dual-shape concern without changing this engine work — a reasonable follow-up, not a merge blocker here.

packages/rolldown/src/plugin/index.ts
hyfdev: Agreed on the temporary approach: keep the engine-level string-id hook for the adapter boundary, and hide the public TypeScript export so plugins do not meet two different `hotUpdate` shapes. That can be a small follow-up on top of this PR and does not need to hold the engine work.

Vite keeps `read` in its wrapper, so plugin-vue's descriptor diff still has a content source without Rolldown owning `read`.

h-a-n-a: Done in 4a78eab5c — the hook is now hidden on the Rolldown side:

- `hotUpdate` is removed from `FunctionPluginHooks` / `Plugin`, and `HotUpdateOptions` is no longer exported.
- The name no longer appears anywhere in the published `.d.ts` or the generated API docs. It was also leaking as a string literal through the hook-name list type (`ENUMERATED_INPUT_PLUGIN_HOOK_NAMES` → `DefinedHookNames`), so I moved it to an internal name list.
- The runtime path is unchanged: the engine still reads `hotUpdate` off the plugin object, so Vite's adapter keeps working as-is (it declares the shape structurally and attaches the hook at runtime).
- The types now live in a source-only module (`plugin/internal-hooks.ts`) that is not exported from any entry point. It is set up as a small registry, so if a future dev-only hook also needs to stay Vite-only, hiding it is one added entry rather than a new mechanism.

The branch is also rebased onto latest main; the hook wiring is ported to the new `bindingifyHook` helper from #10428. Both hot-update playgrounds still pass.

packages/rolldown/src/plugin/bindingify-watch-hooks.ts
h-a-n-a: I got you. I will split this as a separate task. AFAIK, this does not block this PR, does it?

crates/rolldown_plugin/src/types/hook_hot_update_args.rs
```diff
@@ -0,0 +1,18 @@
...
+/// module ids (absolute paths or virtual ids), never the escaped stable-id encoding used by
+/// client-facing payloads.
+#[derive(Debug)]
+pub struct HookHotUpdateArgs {
+  pub kind: WatcherChangeKind,
+  /// Normalized absolute path of the changed file.
+  pub file: ArcStr,
+  /// The affected module ids as currently computed. Plugins earlier in the chain may have
+  /// replaced the engine's default mapping.
+  pub modules: Vec<ArcStr>,
```
sapphi-red, line 17: non blocking: I guess we need to change this to `Vec<(ArcStr, IsModuleOrAsset)>` to support the test cases added in https://github.com/vitejs/vite/pull/20391. Especially when a file is imported both as a normal module and as an asset (`this.emitFile`).

sapphi-red (approved):

crates/rolldown/src/hmr/hmr_stage.rs
h-a-n-a: Both points are addressed in 6800b1cf3:

- The default ids handed to the hook are now `to_slash`ed like `args.file`, so one args object no longer mixes path conventions on Windows. The return-path lookup keeps its own normalization, so ids a plugin builds with `path.join` still round-trip.
- The transform-dependency matches are now sorted by `stable_id` before joining the set, so `args.modules` has a stable order: own module first, then registrants in sorted order. I sorted by the id string rather than `ModuleIdx` because module indices are assigned during the parallel scan, so they are not stable across runs either.

crates/rolldown/src/hmr/hmr_stage.rs
h-a-n-a: Coverage added in 4eaec2740, and looking into this corner surfaced a bigger issue:

- A dev-server playground (`hmr-delete-self-watched`) now covers the self-watched delete end to end through Vite: deleting the still-imported file surfaces the unresolved-import error in the overlay, and recreating it recovers. At that level the two-route corner is not separately observable — the rebuild's resolve error dominates whether the guard is right or wrong.
- At the raw engine level, deleting a still-imported file is a silent Noop today. That is not a deliberate contract: it only works because the resolver cache is never invalidated by watch events (oxc_resolver caches every lookup; dev clears the cache only on a full rebuild or a tsconfig change). Vite masks this because `vite:resolve` re-checks the filesystem on every call. I filed this as #10487 with a repro branch (`test/dev-resolver-cache-deleted-file`), and this PR adds a skipped acceptance test in `dev-watch.test.ts` that asserts the desired behavior (delete → resolve error, recreate → recover); the cache-invalidation fix un-skips it and also flips `delete_file_used`'s step-0 `noop` snapshot deliberately.

crates/rolldown/tests/rolldown/topics/hmr/hot_update_hook/mod.rs
h-a-n-a: Added the remaining contract coverage as browser e2e specs in `packages/test-dev-server` (one scenario per gap), driving a real dev server, watcher, and client instead of engine snapshots:

- `hmr-hot-update-hook` (extended) — an unknown id next to a known one is dropped and the known module still ships; returning only unknown ids empties the set and the round ends as a noop; an unmapped watched file (`notes.txt`, plain `buildStart` watch, no module) runs the chain with an empty default set and the hook can claim modules for it; a declining hook leaves the default flow fully intact, including the unchanged-output suppression for a whitespace-only edit; and a mixed round where one file is suppressed while another file's default flow ships untouched.
- `hmr-hot-update-hook-chain` — two plugins: the second sees the first's edit, can re-add modules after an empty return, and its decline keeps the first's replacement instead of the default.
- `hmr-hot-update-hook-delete` — real watcher delete events: the deleted module itself arrives in the hook's default set (proven through a hook-side log the spec reads back), a decline falls back to the default delete flow, and a replacement returned for a delete ships as-is even though its code is unchanged.
- `hmr-hot-update-hook-error` — a throwing hook fails the whole round (error overlay in the browser); the failed round's edit is not queued for retry — a later unrelated edit ships alone and the lost content reaches the browser only when its own file changes again, the `watchChange` semantics you noted.

Also reworded the internal-docs sentence — instead of "fails the whole batch, like a build error" it now spells out that nothing is queued in `pending_rescans` and lost edits land only when the file changes again.

crates/rolldown_plugin/src/types/hook_hot_update_args.rs
h-a-n-a: Agreed. Right now the dev engine's graph has no asset representation for a changed file to map to, so the flag would always say "module". Since the hook is hidden from the public plugin API, changing `modules` to `Vec<(ArcStr, IsModuleOrAsset)>` later is cheap and non-breaking. I'd do it as a follow-up together with asset support in dev, and port the vitejs/vite#20391 cases as engine tests at that point.

h-a-n-a: tracked in https://github.com/rolldown/rolldown/issues/10491

h-a-n-a:
### Merge activity

* **Jul 28, 6:30 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 28, 6:30 AM UTC**: `h-a-n-a` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 28, 6:37 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10306, opened 2026-07-15
fix(magic-string): count positional inserts at index 0 on an empty source

IWANABETHATGUY:
> Stacked on #10295 — review that first; this PR's diff is just the second commit.
>
> Reported by @hyfdev in https://github.com/rolldown/rolldown/pull/10295#discussion_r3588401894. Split out of #10295 rather than folded into it: the cause is unrelated (chunk placement, not unit conversion) and it also changes `isEmpty()`, so it deserves its own entry in the history. It can't sit *below* #10295 though — see "Why it stacks here".

### What this solves

A positional insert at index 0 on an **empty** source is not counted:

```js
const s = new RolldownMagicString('');
s.appendLeft(0, 'é');
s.toString();  // 'é'
s.length();    // 0  — magic-string returns 1
s.isEmpty();   // true — magic-string returns false
```

All four positional methods reproduce it, and an emoji returns `0` instead of `2`. Pre-existing: `main` returns `0` here too, since `length()` there also walks chunks only.

### Root cause

`by_start_mut` and `by_end_mut` answered *before* consulting the chunk maps:

```rust
fn by_start_mut(..) { if text_index as usize == self.source.len() { Ok(None) } .. }
fn by_end_mut(..)   { if text_index == 0 { Ok(None) } .. }
```

Both are shortcuts for "no chunk at this index". That holds for a non-empty source — nothing ends at 0, nothing starts at `len`. It is false for an empty source: its sole `[0, 0)` chunk both **starts and ends at 0**, and the constructor registers it under `chunk_by_start[0]` *and* `chunk_by_end[0]`. The guard fired first, so the content was routed to the global intro/outro, which `length()` and `isEmpty()` deliberately exclude.

`magic-string` has no such special case — `appendLeft` is a plain `byEnd[index]` lookup, `appendRight` a plain `byStart[index]` lookup, and on an empty source both hit the zero-length chunk.

### The fix

Delete both shortcuts and let the map answer. Correct for both cases — on `'abc'`, `chunk_by_end.get(0)` returns `None` anyway, so out-of-chunk-range inserts still go to the global intro/outro exactly as before.

### Why it stacks here and not below #10295

On its own this fix would still be wrong. With the guards gone but no UTF-16 change, `''` + `appendLeft(0, 'é')` puts `é` on the chunk and `len()` counts its **2 bytes** — so `length()` returns 2, not 1. The correct answer only appears with both changes, which is why this sits directly on top of the `length()` UTF-16 fix.

### Tests

Eight cases. **Five fail without this change**: the four positional methods at index 0, plus the emoji UTF-16 case.

The other three are the contrast @hyfdev asked for, and they pass either way by design:

- `append('é')` / `prepend('é')` on an empty source stay **excluded** (`length() === 0`) — they are not positional, so they belong in the global intro/outro.
- Non-empty out-of-range inserts (`'abc'` + `appendLeft(0, 'X')` / `appendRight(3, 'X')`) stay **excluded**.

They earn their place: the tempting alternative fix is to make `len_utf16` count the global intro/outro, which would satisfy all five failing tests while breaking exactly these three.

### Verification

Differential run against `magic-string@0.30.21` over 11 cases (`toString`, `length`, `isEmpty`): **0 divergences**, down from 5 before this change. The 199-test vendored upstream conformance suite still passes.

### DIFF
--- a/crates/string_wizard/src/magic_string/mod.rs
+++ b/crates/string_wizard/src/magic_string/mod.rs
@@ -320,24 +320,27 @@ impl<'text> MagicString<'text> {
     Ok(())
   }
 
+  /// Returns the chunk starting at `text_index`, or `None` if no chunk does — in which case
+  /// the caller sends the content to the global outro.
+  ///
+  /// Do not short-circuit on `text_index == self.source.len()`: for an empty source the sole
+  /// `[0, 0)` chunk both starts *and* ends at 0, so index 0 has a chunk even though it is also
+  /// the end of the source. The map lookup already answers this correctly for both cases, and
+  /// matches the reference `magic-string`, which does a plain `byStart[index]` lookup.
   fn by_start_mut(&mut self, text_index: u32) -> Result<Option<&mut Chunk<'text>>, String> {
-    if text_index as usize == self.source.len() {
-      Ok(None)
-    } else {
-      self.split_at(text_index)?;
-      let idx = self.chunk_by_start.get(&text_index);
-      Ok(idx.map(|idx| &mut self.chunks[*idx]))
-    }
+    self.split_at(text_index)?;
+    let idx = self.chunk_by_start.get(&text_index);
+    Ok(idx.map(|idx| &mut self.chunks[*idx]))
   }
 
+  /// Returns the chunk ending at `text_index`, or `None` if no chunk does — in which case the
+  /// caller sends the content to the global intro.
+  ///
+  /// Do not short-circuit on `text_index == 0`: see [`Self::by_start_mut`].
   fn by_end_mut(&mut self, text_index: u32) -> Result<Option<&mut Chunk<'text>>, String> {
-    if text_index == 0 {
-      Ok(None)
-    } else {
-      self.split_at(text_index)?;
-      let idx = self.chunk_by_end.get(&text_index);
-      Ok(idx.map(|idx| &mut self.chunks[*idx]))
-    }
+    self.split_at(text_index)?;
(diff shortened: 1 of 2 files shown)


IWANABETHATGUY:
* **#10306** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10306?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10306?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10306).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

shulaoda (approved):

shulaoda:
### Merge activity

* **Jul 15, 5:37 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 15, 5:51 PM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 15, 5:58 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10310, opened 2026-07-16
fix(output): skip separator for empty hoisted functions

hyfdev:
Follow-up to #10289 and the [post-merge review](https://github.com/rolldown/rolldown/pull/10289#issuecomment-4982081640).

## Why

The ESM renderer unconditionally appended the accumulated `hoisted_fns` string for an on-demand-wrapped module group. When the group had no hoisted functions, `SourceJoiner` still treated the empty string as a source slot and inserted a separator before the next source.

This is the same empty-slot behavior fixed for the generated import prelude in #10289.

## What changed

- skip `hoisted_fns` when it is empty
- update the single affected ESM snapshot, removing the extra blank line before `init_main`
- make the #7555 `renderChunk` fixture assert that the hook ran exactly once
- preserve the actual hook input in assertion failures by matching `code` directly
- add the same call-count guard to the existing external-import `renderChunk` fixture

## Scope

This keeps the fix at the caller instead of changing `SourceJoiner` globally. Empty sources can carry sourcemaps, and a class-level change would need coherent handling for `append_source_dyn` and `prepend_source` as well.

## Impact

ESM output with `onDemandWrapping` no longer contains the stray separator when a concatenated group has no hoisted functions. The affected plugin fixtures can no longer pass if `renderChunk` is never invoked.

## Validation

- `just test-rust`
- `cargo test -p rolldown --test integration strict_execution_order -- --ignored` (65 passed)
- targeted Vitest run for both changed `renderChunk` fixtures
- `vp check` for both changed TypeScript fixtures
- `cargo fmt --all --check`
- `cargo check -p rolldown --all-features --all-targets --locked`
- `git diff --check`

### DIFF
--- a/crates/rolldown/src/ecmascript/format/esm.rs
+++ b/crates/rolldown/src/ecmascript/format/esm.rs
@@ -230,7 +230,9 @@ fn render_chunk_content<'code>(
     // into `make_esm_wrapper_stmt` for the non-concatenated wrapper.
     let is_async = ctx.link_output.metas[group.entry].is_tla_or_contains_tla_dependency;
 
-    source_joiner.append_source(hoisted_fns);
+    if !hoisted_fns.is_empty() {
+      source_joiner.append_source(hoisted_fns);
+    }
     if !hoisted_vars.is_empty() {
       source_joiner.append_source(concat_string!("var ", hoisted_vars, ";"));
     }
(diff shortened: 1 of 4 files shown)


hyfdev:
### Merge activity

* **Jul 16, 3:31 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 16, 3:38 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 16, 3:44 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

IWANABETHATGUY (approved):

PR #10311, opened 2026-07-16
fix(magic-string): reject non-contiguous moves before rewiring the chunk list

IWANABETHATGUY:
### What this solves

`move()` corrupts the chunk list when the requested range is no longer contiguous:

```js
s.move(0, 1, 2); // list is now B -> A -> C
s.move(1, 3, 0); // asks to move B..C — A sits between them
```

`relocate` assumes `[first..last]` is a contiguous run in the linked list. An earlier move can break that: in the state above, `first.prev` and `last.next` are both `None`, so lifting the range out leaves nothing to re-attach to, and the code fell off the end of the list and unwrapped a `None` — a Rust panic that aborts the whole process across the FFI boundary.

Worse than the panic itself: `relocate` rewires `next`/`prev` **before** reaching that point, so any failure there leaves a chunk pointing at itself — a caller that survives and calls `toString()` loops forever. (`magic-string` reaches the same state and TypeErrors, but its `toString()` still terminates.)

### The fix

Detect the condition up front, from the `prev`/`next` captured **before any mutation**, and return an error while the list is still intact.

The two later `unwrap` sites deliberately keep a `debug_assert` + `if let` instead of becoming `?`: the pre-check guarantees they cannot fire, and returning an error from the middle of the rewiring would be exactly the corrupting behaviour this PR removes. That is also why this can't be fixed by the blanket `expect`→`Result` conversion used in #10301 — converting a panic into a catchable error is only safe when the function hasn't already mutated shared state.

### Tests

- JS: the reproduction above throws a catchable error, and the instance stays usable (`toString()` terminates).
- Rust: unit test for the same sequence at the `string_wizard` level.

Both fail on the parent commit (the JS one by killing the vitest worker — the panic takes the process down) and pass with the fix.

### DIFF
--- a/crates/string_wizard/src/magic_string/movement.rs
+++ b/crates/string_wizard/src/magic_string/movement.rs
@@ -46,6 +46,18 @@ impl MagicString<'_> {
     let old_left_idx = self.chunks[first_idx].prev;
     let old_right_idx = self.chunks[last_idx].next;
 
+    // The moved range occupies both ends of the linked list, so lifting it out would leave no
+    // chunk to re-attach to. Reachable once an earlier move has reordered chunks — after
+    // `relocate(0, 1, 2)` on "abc" the list is B->A->C, so `relocate(1, 3, 0)` asks to move
+    // B..C, which is no longer a contiguous run.
+    //
+    // This has to be caught here, before the rewiring below: that rewiring is not atomic, and
+    // bailing out part-way through leaves a chunk pointing at itself, which makes `to_string`
+    // loop forever.
+    if old_left_idx.is_none() && old_right_idx.is_none() {
+      return Err("Cannot move a range that spans the entire string".to_string());
+    }
+
     let new_right_idx = self.chunk_by_start.get(&to).copied();
 
     // `new_right_idx` is `None` means that the `to` index is at the end of the string.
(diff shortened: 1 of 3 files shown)


IWANABETHATGUY:
* **#10326** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10326?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10312** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10312?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10311** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10311?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10311?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10311).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

shulaoda (approved):

IWANABETHATGUY:
### Merge activity

* **Jul 16, 4:46 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 16, 4:46 PM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 16, 4:54 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

hyfdev:
The guard added here rejects only one non-contiguous topology; other reachable layouts still corrupt the chunk list. This merged while the review was still in flight, so flagging it post-merge.

The predicate at https://github.com/rolldown/rolldown/blob/2c66e98a469f1d5d12b062e3b796c86135056612/crates/string_wizard/src/magic_string/movement.rs#L57 only catches the case where `first` has no predecessor and `last` has no successor. Reachable bypass, verified at current `main` (`2c66e98`):

```rust
let mut s = MagicString::new("abc");
s.relocate(0, 2, 3).unwrap(); // "cab", list C -> A -> B
s.relocate(1, 3, 0).unwrap(); // unexpectedly returns Ok
```

For the second call, `first = B` and `last = C`, while both `old_left_idx` and `old_right_idx` are `Some(A)`, so the guard passes. The rewiring then assigns `A.next = A`, and `to_string()` never returns — I confirmed both with assertion tests at `2c66e98` (the second `relocate` returns `Ok`, and `to_string` still hangs at a 3s timeout). The existing valid-move suite passes, so this is exactly the pre-existing failure class this PR set out to reject, reachable through a different endpoint layout.

One possible direction is to validate every original-position adjacency from `first_idx` through `last_idx` before rewiring — each selected chunk's `next` must be the chunk starting at its `end`. I verified the reproduction, not any particular fix; one-ended and interior layouts need covering tests either way.

PR #10312, opened 2026-07-16
feat(magic-string): support storeName on overwrite/update

IWANABETHATGUY:
> Stacked on #10311 — review that first; this PR's diff is just the top commit.

### What this solves

`overwrite`/`update` accepted a `storeName` option in their TypeScript signatures, but it did nothing: the binding hardcoded `keep_original: false`, so `generateMap().names` was always empty regardless of the option.

```js
const s = new RolldownMagicString('var foo = 1;');
s.overwrite(4, 7, 'bar', { storeName: true });
s.generateMap({}).names; // [] — magic-string returns ['foo']
```

### An underlying bug, not just plumbing

Wiring the flag through exposed a real bug in `string_wizard`: it took the stored name from the **start chunk's span** rather than the replaced range. When an earlier edit had already split the range, the start chunk covers only part of it — `'f'` got stored instead of `'foo'`, and the mapping pointed at that wrong name. `magic-string` keys `storedNames` off `original.slice(start, end)`.

Fixed with a `stored_names` map on `MagicString` (name → insertion order), populated in `inner_update_with` from the whole `[start, end)` range — the single code path both `update` and `overwrite` converge on. The sourcemap builder seeds `names` from it in insertion order and resolves each edited chunk's name by lookup, which also reproduces magic-string's quirk that a split range leaves its name in `names` with no mapping referencing it (upstream's per-mapping lookup is `names.indexOf(chunk.original)`, which misses once the chunk is narrower than the range).

### Verification against `magic-string@0.30.21`

- **Differential run, 18 scenarios, all byte-identical** on full `generateMap()` output — raw `mappings` VLQ string, `names` content *and* order, `sourcesContent` — plus `toString()`. Covers: both edit orders, duplicate-name dedup, re-storing the same range, split ranges, `contentOnly` interaction, multiline sources, non-BMP characters before the name, stored-then-removed ranges, `hires: true` and `hires: 'boundary'`.
- **Un-skipped the vendored upstream test** `should recover original names` — its skip-list entry said *"storeName option not implemented"*, which this PR makes untrue. It passes: `SourceMapConsumer.originalPositionFor()` recovers the original name from our map.
- Rollup's test suite: 0 failed / 1213 passed at this branch.

### Tests

Six hand-written cases, consolidated against the vendored suite so nothing is covered twice (the basic overwrite case lives in the un-skipped upstream test):

- `update({ storeName: true })` — `update` has its own napi options struct (`BindingUpdateOptions`); a missing `store_name` field there is invisible to every overwrite-based test.
- names stay empty when `storeName` is omitted / `false` — flag-gating pins (pass either way by design; an implementation that stored every replaced range unconditionally would pass every positive test).
- `contentOnly` + `storeName` are independent.
- the split-range regression (`'foo'`, not `'f'`).
- a split range leaves the mapping unnamed, like magic-string — decoded-mappings assertion, split vs unsplit.

All except the two documented flag-gating pins fail on the parent commit and pass with this change.

### DIFF
--- a/crates/rolldown_binding/src/types/binding_magic_string.rs
+++ b/crates/rolldown_binding/src/types/binding_magic_string.rs
@@ -218,12 +218,16 @@ pub struct BindingMagicStringOptions {
 #[derive(Default)]
 pub struct BindingUpdateOptions {
   pub overwrite: Option<bool>,
+  /// Stores the replaced content in the generated sourcemap's `names` field.
+  pub store_name: Option<bool>,
 }
 
 #[napi(object)]
 #[derive(Default)]
 pub struct BindingOverwriteOptions {
   pub content_only: Option<bool>,
+  /// Stores the replaced content in the generated sourcemap's `names` field.
+  pub store_name: Option<bool>,
 }
 
 #[napi(object)]
@@ -757,14 +761,15 @@ impl BindingMagicString<'_> {
       .utf16_to_byte_mapper
       .utf16_to_byte(self.apply_offset_u32(end)?)
       .ok_or_else(|| napi::Error::from_reason("Invalid end character index"))?;
-    let content_only = options.and_then(|o| o.content_only).unwrap_or(false);
+    let (content_only, store_name) = options
+      .map_or((false, false), |o| (o.content_only.unwrap_or(false), o.store_name.unwrap_or(false)));
     self
       .inner
       .update_with(
         start_byte,
         end_byte,
         content,
-        string_wizard::UpdateOptions { overwrite: !content_only, keep_original: false },
+        string_wizard::UpdateOptions { overwrite: !content_only, keep_original: store_name },
       )
       .map_err(napi::Error::from_reason)?;
     Ok(this)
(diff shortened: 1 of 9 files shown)


IWANABETHATGUY:
* **#10326** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10326?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10312** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10312?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10312?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10311** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10311?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10312).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

shulaoda (approved):

IWANABETHATGUY:
### Merge activity

* **Jul 16, 4:46 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 16, 4:48 PM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 16, 4:56 PM UTC**: The [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) couldn't merge this PR because **it had merge conflicts**.
* **Jul 16, 5:00 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 16, 5:03 PM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 16, 5:09 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10316, opened 2026-07-16
refactor: enable oxc_ast's `disable_old_builder` feature

Boshen:
Enables oxc_ast's `disable_old_builder` cargo feature — the endgame of the oxc#23043 `AstBuilder` redesign, ahead of oxc removing the old builder methods entirely. All construction call sites were already on the new per-type-constructor API, so this is the enforcement pass; it also stops compiling the deprecated 599 KB `generated/ast_builder.rs`.

The `oxc` umbrella crate has no passthrough for the feature, so it is enabled through a direct `oxc_ast` dependency carried by `rolldown_ecmascript`; cargo feature unification applies it to the copy the umbrella re-exports. The dep is cargo-shear-ignored — no code references `oxc_ast::` paths directly, and without the ignore, CI autofix would remove the dep and silently disable the feature.

Changes forced by the feature:

- `AstBuilder` is no longer `Copy`/`Clone`, so `AstFactory` isn't either. The two by-value copy sites now reconstruct a handle (`AstFactory::new(alloc)` is a free wrapper over the arena reference) or borrow it.
- `ConstantValue::to_expression` and `json_value_to_expression` take `&AstBuilder`; the `HmrAstBuilder::builder` impls construct a fresh `AstBuilder::new(allocator())` instead of deref-copying, leaving their call sites untouched.
- The top-level `oxc::ast::{AstBuilder, NONE}` re-exports are removed by the feature; imports now use `oxc::ast::builder::{AstBuilder, NONE}`.

Also updates `internal-docs/ast-construction/implementation.md` (migration completion note, corrected sample) and adds `oxc_ast` to the `upgrade-oxc` skill's version-bump lists so the pin moves in lockstep on future upgrades.

A cleaner follow-up on the oxc side would be an umbrella passthrough feature (`disable_old_builder = ["oxc_ast/disable_old_builder"]`), after which the direct dep here can be dropped.

### DIFF
--- a/.claude/skills/upgrade-oxc/SKILL.md
+++ b/.claude/skills/upgrade-oxc/SKILL.md
@@ -15,8 +15,8 @@ CRITICAL: Run each step sequentially ONE AT A TIME. Wait for each command to FUL
 4. Edit `pnpm-workspace.yaml`: update `@oxc-project/runtime`, `@oxc-project/types`, `oxc-minify`, `oxc-parser`, `oxc-transform` to the version from step 3 (use `=x.y.z` format)
 5. `cargo search oxc_allocator --limit 1` - note the version
 6. `cargo search oxc_resolver --limit 1` - note the version
-7. Edit `Cargo.toml`: update `oxc`, `oxc_allocator`, `oxc_ecmascript`, `oxc_minify_napi`, `oxc_parser_napi`, `oxc_transform_napi`, `oxc_traverse` to version from step 5; update `oxc_resolver`, `oxc_resolver_napi` to version from step 6
-8. `cargo update oxc oxc_allocator oxc_ecmascript oxc_minify_napi oxc_parser_napi oxc_transform_napi oxc_traverse oxc_resolver oxc_resolver_napi oxc_sourcemap oxc_index`
+7. Edit `Cargo.toml`: update `oxc`, `oxc_allocator`, `oxc_ast`, `oxc_ecmascript`, `oxc_minify_napi`, `oxc_parser_napi`, `oxc_transform_napi`, `oxc_traverse` to version from step 5; update `oxc_resolver`, `oxc_resolver_napi` to version from step 6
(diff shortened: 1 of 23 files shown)


shulaoda (approved):

IWANABETHATGUY (approved):

hyfdev (approved):
Reviewed at `f8100a991928fcac05a3f4704669b1447bb8d315`.

The feature-only `oxc_ast` dependency correctly enables `disable_old_builder` through Cargo feature unification, and the import/ownership migration preserves the pinned Oxc 0.140 semantics. I verified the full diff and dependency graph, `cargo check --workspace --all-features --all-targets --locked`, the JSON AST tests, formatting, and cargo-shear.

The red Vite Test Ubuntu job is the same pre-existing source-map snapshot mismatch on the base commit. The branch is still behind `main`, so the normal update/merge-gate rerun remains. No blocking findings.

overlookmotel (approved):
We can clean this up further, shortening code. I'd also like to propose changes to `AstFactory` - I'll open an issue for that.

We intend to imminently remove the old `AstBuilder` from Oxc entirely, and with it remove the `disable_old_builder` Cargo feature - so that complication will go away shortly.

I'd recommend merging this, I'll open an issue to discuss the changes I have in mind for `AstFactory`, and I can work on it next week.

Not sure why Vite Test Ubuntu job is failing CI. I don't have the context to say if it's related to these changes, but I suspect not.

overlookmotel:
Opened issue about proposed changes to `AstFactory`: #10328

hyfdev:
### Merge activity

* **Jul 17, 2:55 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 17, 2:55 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 17, 3:01 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10322, opened 2026-07-16
fix(code-splitting): treat class definition-time global reads as order-sensitive

hyfdev:
## Summary

Fixes the class order-sensitivity gap found while reviewing #10104. A class can be side-effect-free for tree shaking while its definition still reads mutable global state from an extends expression, computed key, static initializer, or static block. Reordering that class across a module that patches the global can change behavior.

When Oxc certifies the whole class as side-effect-free, Rolldown now runs an independent definition-time visitor that collects only GlobalVarAccess and PureAnnotation reasons. It never invokes or short-circuits on tree-shaking analysis. Therefore a PureCJS target write remains tree-shaking-only, while its RHS and later siblings are still visited.

The visitor covers evaluated binding defaults, computed keys, assignment RHS, direct IIFEs, and nested classes. It skips deferred function and method bodies and instance initializers.

The tree-shaking channel is structurally unchanged: the original whole-class Oxc boolean remains its only input, and the new visitor can only add order-sensitive reasons.

## Regression coverage

- PureCJS-only and literal binding-default cases remain order-insensitive.
- PureCJS followed by a global read or pure-annotated call is order-sensitive in either operand order.
- Static-block binding defaults, computed member keys, assignment RHS, direct IIFEs, heritage expressions, and nested class-definition work are covered.
- Deferred arrow/function bodies, method bodies, and instance initializers remain order-insensitive.

## Validation

- cargo test -p rolldown --lib — 87 passed.
- cargo test -p rolldown — 87 unit and 49 integration tests passed; the only failure was the existing missing local test262 submodule.
- cargo clippy -p rolldown --lib --tests -- -D warnings — clean.
- cargo fmt --all -- --check — clean.

Follow-up to #10104; addresses Major 1 from https://github.com/rolldown/rolldown/pull/10104#issuecomment-4992442574.

Draft, rebased directly onto main.

### DIFF
--- a/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs
+++ b/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs
@@ -6,7 +6,8 @@ use oxc::ast::ast::{
   IdentifierReference, UnaryOperator, VariableDeclarationKind,
 };
 use oxc::ast::match_member_expression;
-use oxc::semantic::{NodeId, SymbolId};
+use oxc::ast_visit::{Visit, walk};
+use oxc::semantic::{NodeId, ScopeFlags, SymbolId};
 use oxc_ecmascript::GlobalContext;
 use oxc_ecmascript::side_effects::{
   MayHaveSideEffects, MayHaveSideEffectsContext, PropertyReadSideEffects,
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY (approved):

hyfdev:
### Merge activity

* **Jul 21, 1:49 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 21, 1:49 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 21, 1:55 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10324, opened 2026-07-16
test(code-splitting): pin flag-off wrapped-esm init emission

hyfdev:
## Summary

Adds three executable **flag-off** regression fixtures for wrapped-ESM init emission changes that landed with #10104. None enables `strictExecutionOrder`, and this PR changes no production source.

The earlier version of this PR required each barrel directly, which wrapped the immediate importee and bypassed the shared init-target collector it claimed to cover. The two collector fixtures now keep the immediate barrel at `WrapKind::None` in a shared chunk, while later `require()` calls wrap the leaf owners and make their init symbols reachable from the main chunk.

### Importer-local namespace liveness

`namespace_liveness_gating` imports the barrel namespace but reads only `ns.a()`. A second entry places the unwrapped barrel in a shared chunk; later requires make both wrapped leaf init symbols reachable.

- Current output emits only `init_leaf_a()` at the namespace-import position. `init_leaf_b()` stays at its later explicit require position.
- Pre-#10104 output emitted `init_leaf_a(), init_leaf_b()` together at the namespace-import position.
- The shared barrel still evaluates both leaves exactly once; runtime order is `A, B, BEFORE_REQUIRE:a, MAIN`.

### Sorted, non-ambiguous star-export routing

`ambiguous_star_reexport` has two wrapped leaves that both export `x`. The entry uses a real `export * from` record; a dynamic CommonJS re-export keeps that record included, while the second entry keeps the immediate barrel unwrapped and in another chunk.

- Current output excludes ambiguous `x` from the collector's sorted, non-ambiguous export map and emits no owner init before `__reExport(...)`.
- Pre-#10104 output selected an ambiguous owner and emitted `init_mod_a()` before `__reExport(...)`.
- Runtime confirms `x` is absent and both owners still evaluate exactly once through the shared barrel.

### Excluded re-export to a TLA wrapper

`excluded_reexport_tla` uses `codeSplitting: false`, so an inline dynamic import wraps the barrel and its static dependency chain even with strict execution order disabled. With `moduleSideEffects: false`, the carrier's unused re-export is excluded while the same TLA leaf remains live through a separate `used` import.

- Current output emits `await init_leaf()` in the carrier's async wrapper before its later statement.
- Pre-#10104 output emitted bare `init_leaf()`, producing `CARRIER:undefined` before the leaf's top-level await completed.
- Current runtime order is `LEAF:before, LEAF:after, CARRIER:ready, BARREL:used:marker`.

## Behavior / snapshot impact

Tests only: two existing fixtures are corrected, one TLA fixture is added, and their generated snapshots are pinned. No existing production code or unrelated snapshot changes.

## Validation

- Rebased onto `main` at `cccf1d0d849c71e80c02fe3ff17a97192f3bf258`.
- Generated all three fixtures with the macOS native binding from current-main CI [run 29795703407](https://github.com/rolldown/rolldown/actions/runs/29795703407); every `_test.mjs` passed under Node.
- Compared each committed snapshot byte-for-byte with those generated assets after applying the integration harness's runtime-module hiding.
- Re-ran the same sources with the pre-#10104 binding from CI [run 29559371389](https://github.com/rolldown/rolldown/actions/runs/29559371389): both collector fixtures show the obsolete extra init, and the TLA fixture logs `CARRIER:undefined`.
- `vp fmt ... --check`, `typos`, JS syntax checks, JSON parsing, and `git diff --check` pass. The pre-commit `vp check --fix` hook also completed cleanly.

- CI [run 29819995357](https://github.com/rolldown/rolldown/actions/runs/29819995357): Repo, Rust, Node, Cargo, WASI, native-build, and Node 20/22/24 test jobs passed. Vite Test Ubuntu and the three Ubuntu Dev Server jobs stopped before tests because the Vite `rolldown-canary` branch no longer rebases cleanly onto `vite/main` (`pnpm-lock.yaml` conflict); this is unrelated Vite checkout noise.

## AI usage disclosure

Per the [AI Usage Policy](https://rolldown.rs/contribution-guide/#ai-usage-policy), OpenAI Codex was used to rebase and review the PR, design the corrected regression shapes, and verify generated output. I reviewed the final diff and validation evidence.

---

Follow-up to #10104; addresses Major 2b/2c from the review discussion: https://github.com/rolldown/rolldown/pull/10104#issuecomment-4992442574

### DIFF
--- a/crates/rolldown/tests/rolldown/function/wrapped_esm_init_interop/ambiguous_star_reexport/_config.json
+++ b/crates/rolldown/tests/rolldown/function/wrapped_esm_init_interop/ambiguous_star_reexport/_config.json
@@ -1,0 +1,14 @@
+{
+  "config": {
+    "input": [
+      {
+        "name": "main",
+        "import": "./main.js"
+      },
+      {
+        "name": "wrapper",
+        "import": "./wrapper.js"
+      }
+    ]
+  }
+}
--- a/crates/rolldown/tests/rolldown/function/wrapped_esm_init_interop/ambiguous_star_reexport/_test.mjs
+++ b/crates/rolldown/tests/rolldown/function/wrapped_esm_init_interop/ambiguous_star_reexport/_test.mjs
@@ -1,0 +1,12 @@
+import assert from 'node:assert';
+
+globalThis.__log = [];
+const ns = await import('./dist/main.js');
+
+// Runtime check: ambiguous x stays absent and both owners evaluate once; the snapshot pins that neither owner init is added at main's star re-export.
+assert.strictEqual('x' in ns, false, 'ambiguous names are omitted from the namespace');
+assert.deepStrictEqual(
+  globalThis.__log,
+  ['A', 'B', 'BEFORE_REQUIRE', 'MAIN'],
+  'both ambiguous owners evaluate once before the main body',
+);
--- a/crates/rolldown/tests/rolldown/function/wrapped_esm_init_interop/ambiguous_star_reexport/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/function/wrapped_esm_init_interop/ambiguous_star_reexport/artifacts.snap
@@ -1,0 +1,55 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+---
+# Assets
+
+## barrel.js
+
+```js
+// HIDDEN [\0rolldown/runtime.js]
+//#region mod_a.js
+var mod_a_exports = /* @__PURE__ */ __exportAll({ x: () => "ax" });
+var init_mod_a = __esmMin((() => {
+	globalThis.__log.push("A");
+}));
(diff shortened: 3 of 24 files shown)


shulaoda (approved):

hyfdev:
### Merge activity

* **Jul 21, 1:36 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 21, 1:36 PM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 21, 1:42 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10319, opened 2026-07-16
test(test-dev-server): drop the vite submodule, track rolldown-canary

shulaoda:
Stacked on the vite-tests PR. Together they make both Vite harnesses run on the same code: the latest `rolldown-canary` rebased onto the latest `main`.

## Problem

The dev-server test harness built the vite submodule at its pinned commit. Same pin-rot problem as vite-tests: `rolldown-canary` is force-pushed on every update, so a pinned commit ends up on no branch, and canary-side fixes need a manual pin bump to reach CI. The pin mechanism fundamentally conflicts with tracking a rebased branch.

## Change

- **Remove the vite git submodule entirely**: the `.gitmodules` entry and the `vite` gitlink are gone. `vite/` at the repo root is now a plain, gitignored clone.
- **`scripts/src/setup-vite/checkout.ts`**: when `vite/` is missing (fresh clone, CI), it clones vitejs/vite `rolldown-canary` and rebases it onto `origin/main`. An existing checkout is never moved: updating it stays a manual git step inside `vite/`, so local work in the checkout is never trampled.
- **`reusable-node-dev-server-test.yml`**: adds a "Configure Git" step, since the rebase needs a committer identity in CI (the vite-test jobs in ci.yml already had one).
- Wording updates everywhere the submodule or the pin was mentioned: `setup-vite` scripts, `vite-server.ts` comments and error messages, `.gitignore`, `justfile`, `testing.md`, `repo-structure.md`, `implementation.md`, `tests/AGENTS.md`.

Fixes and test adjustments for the Vite side now always belong on the vitejs/vite `rolldown-canary` branch, for both harnesses.

Jobs that check out with `submodules: true` are unaffected: only the `rollup` and `test262` submodules remain.

## Verification

- Fresh `just setup-vite` end to end: clone `rolldown-canary`, rebase onto `main`, `vp install`, vite build, rolldown symlink swap and resolution check all pass.
- Dev-server browser suite smoke (`hmr-client-invalidate`, 4 tests) passes on the resulting checkout with a current rolldown `main` build.
- `just lint` fully green (fmt, oxlint/tsc, knip, publint, ls-lint).
- A repo-wide sweep found no remaining references to the submodule or the pin mechanism.

PR #10318, opened 2026-07-16
test(vite-tests): track the latest rolldown-canary rebased onto vite main

shulaoda:
## Problem

Since #10164 (and #10293 which unified the checkouts), `packages/vite-tests/run.ts` ran Vite's test suite on the commit pinned by the vite submodule gitlink. The `rolldown-canary` branch of vitejs/vite is regularly rebased and force-pushed, so any pinned commit rots quickly: the current pin `4d1480ca` is no longer on any branch. Worse, test adjustments landing on `rolldown-canary` cannot reach rolldown CI without a manual pin bump. That is exactly why CI on `main` is red right now: the `js-sourcemap` inline snapshot was updated on `rolldown-canary` on 2026-07-15 (for the output change from #10249), but the pin predates that update.

## Change

Restore the pre-#10164 approach from #7633. `run.ts` now does:

```
git clone --branch rolldown-canary https://github.com/vitejs/vite.git
git rebase origin/main
```

so test adjustments landing on `rolldown-canary` take effect right away, and new tests from Vite `main` surface incompatibilities with rolldown early. The rebase identity comes from the existing "Configure Git" step in the vite-test CI jobs.

All inline spec patches are removed from `run.ts`. Test adjustments belong on the `rolldown-canary` branch itself, not in a patch layer inside this repo:

- The `css-codesplit` style-/style2- patch was already dead code: upstream Vite `main` has had the swapped assertions since vitejs/vite#22922.
- The `assets` raw-query skip (#8839) is obsolete: the test passes on current rolldown `main`, in both serve and build.
- The `hmr-full-bundle-mode` invalidate patch is removed. This one still needs a spec fix on `rolldown-canary`: with client-side HMR there is no "hmr invalidate" server log anymore, so the spec should assert that `.invalidation-parent` becomes `child updated` instead. Until that lands on the canary branch, `test-serve` fails this single test.

Also reverts the `tsconfig.json` include added for the now-removed `checkout.ts` import, and updates the `repo-structure.md` description.

## Verification

Full local run against current rolldown `main` (debug build):

| Suite | Result |
| --- | --- |
| test-unit | 65 files passed |
| test-serve | 1 real failure: `hmr-full-bundle-mode > invalidate` (expected, see above) |
| test-build | 93 files passed, `js-sourcemap` snapshot green again |

`environment-react-ssr > deps reload` failed once under full-suite load but passes in isolation (timing flake with the slow debug binding, not a regression).

PR #10325, opened 2026-07-16
test(vite-tests): reuse the shared root vite checkout

shulaoda:
Follow-up to the review comments on #10318 and #10319: both test projects should be set up from the single `vite/` checkout at the workspace root, via the `just setup-vite` checkout step.

## Change

`packages/vite-tests/run.ts` no longer clones vitejs/vite from GitHub on its own. It now:

1. calls `ensureViteCheckout()` from `scripts/src/setup-vite/checkout.ts` (the checkout step of `just setup-vite`), which guarantees the root `vite/` checkout exists and is at the latest `rolldown-canary` rebased onto the latest `main`, exactly what #10319 set up for the dev-server tests.
2. clones that checkout locally into `./repo` and runs Vite's test suite there. The local clone shares objects via hardlinks, so no extra network download. It stays a throwaway copy because this suite must edit tracked files (rewrite the pnpm overrides to point `rolldown` at the workspace package, which also rewrites the lockfile on install), while `vite/` itself must stay unpatched: the dev-server setup installs it with `--frozen-lockfile` and builds it with Vite's own pinned rolldown, so the two harnesses would corrupt each other's state if they shared one working copy.

Both harnesses therefore run on exactly the same Vite code: one checkout, one rebase result, no chance of the two suites drifting apart because `rolldown-canary` or `main` moved between two independent clones.

Per the review discussion on #10319, an existing checkout is kept up to date as well: when it is clean and on `rolldown-canary` (the managed state), every setup fetches and re-derives the latest `rolldown-canary` rebased onto `main` (kept as-is when the fetch fails, e.g. offline). A checkout taken over by the developer (dirty, or switched to another branch) is used exactly as-is, with a notice, so local experiments inside `vite/` are never trampled and can still be tested through both harnesses.

Also restores the `checkout.ts` include in `packages/vite-tests/tsconfig.json` and updates the docs accordingly.

## Verification

- Local smoke run: `run.ts` picks up the shared checkout, clones it locally, and proceeds through install, build, and into the test suites.
- All three checkout states exercised live: a clean checkout on `rolldown-canary` was updated to the latest upstream state; a dirty checkout and a detached HEAD were left as-is with a notice.
- `just lint` fully green.

### DIFF
--- a/packages/vite-tests/run.ts
+++ b/packages/vite-tests/run.ts
@@ -2,6 +2,7 @@ import fs from 'node:fs';
 import path from 'node:path';
 import { styleText } from 'node:util';
 import { x } from 'tinyexec';
+import { ensureViteCheckout, viteDir } from '../../scripts/src/setup-vite/checkout.js';
 
 const REPO_PATH = path.resolve(import.meta.dirname, './repo');
 const OVERRIDES = [
@@ -44,16 +45,18 @@ async function runCmdAndPipeOrExit(title: string, cmdOptions: Parameters<typeof 
 
 fs.rmSync(REPO_PATH, { recursive: true, force: true });
 
-// Test the latest `rolldown-canary` rebased onto the latest `main`, so that
-// test adjustments landing on `rolldown-canary` take effect right away and
-// new tests from Vite `main` surface incompatibilities with rolldown early.
+// Reuse the shared `vite/` checkout at the repo root (kept at the latest
+// `rolldown-canary` rebased onto the latest `main`, see
+// scripts/src/setup-vite/checkout.ts), the same code the dev-server tests
+// run on. The tests run on a throwaway LOCAL clone of it, never on the
+// checkout itself: this suite edits tracked files (pnpm overrides) and the
+// checkout must stay unpatched. The clone shares objects via hardlinks, so
+// no network is needed beyond the checkout itself.
+printTitle('# Ensuring the vite checkout...');
+ensureViteCheckout();
 await runCmdAndPipeOrExit(
-  '# Cloning vite repo (rolldown-canary branch)...',
-  ['git', ['clone', '--branch', 'rolldown-canary', 'https://github.com/vitejs/vite.git', REPO_PATH]],
-);
-await runCmdAndPipeOrExit(
-  '# Rebasing rolldown-canary onto main...',
-  ['git', ['rebase', 'origin/main'], { nodeOptions: { cwd: REPO_PATH } }],
+  '# Cloning the local vite checkout...',
+  ['git', ['clone', viteDir, REPO_PATH]],
 );
 
 printTitle('# Updating pnpm-workspace.yaml to link to local rolldown...');
--- a/scripts/src/setup-vite/checkout.ts
+++ b/scripts/src/setup-vite/checkout.ts
@@ -1,8 +1,10 @@
-// Handling of the Vite checkout (`vite/` at the repo root, gitignored), used
-// by the dev-server test harness (`packages/test-dev-server`, see `main.ts`).
-// It is a plain clone of vitejs/vite on the `rolldown-canary` branch rebased
-// onto the latest `main`, the same code `packages/vite-tests` runs on, so
-// both harnesses track the canary line instead of a pinned commit.
+// Handling of the Vite checkout (`vite/` at the repo root, gitignored), the
+// single Vite checkout shared by the dev-server test harness
+// (`packages/test-dev-server`, see `main.ts`) and Vite's own test suite
+// (`packages/vite-tests/run.ts`, which clones this checkout locally). It is
+// a plain clone of vitejs/vite on the `rolldown-canary` branch rebased onto
+// the latest `main`, so both harnesses track the canary line instead of a
+// pinned commit.
 
 import { execSync } from 'node:child_process';
 import nodeFs from 'node:fs';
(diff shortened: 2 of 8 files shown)


shulaoda:
* **#10325** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10325?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10325?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10319** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10319?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10318** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10318?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10325).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

h-a-n-a:
There is a staleness window in how `run.ts` reuses the shared checkout.

`packages/vite-tests/run.ts` calls `ensureViteCheckout()`, which can move `vite/` to a new commit — but nothing rebuilds the dev-server side afterwards. `loadVite()` in `test-dev-server` only checks that `vite/packages/vite/dist` **exists**, not that it matches the current checkout. Whenever upstream `rolldown-canary` or `main` moved since the last setup (the normal case), this happens:

```
just setup-vite          → vite/ at commit A, dist built from A     ✔ consistent
run vite-tests           → ensureViteCheckout() moves vite/ to B
                           dist is still built from A
next browser-test run    → loadVite() finds dist → silently tests A
```

So a vite-tests run can silently invalidate the dev-server test results.

Let's make the setup follow one rule instead:

- **`just setup-vite` is the only entry point that touches `vite/`.** It sets the commit (latest `rolldown-canary` rebased onto `main`) and finishes the whole setup — install, vite build, rolldown link — in one pass, for both test projects.
- **Scripts in `test-dev-server` and `vite-tests` never do setup themselves.** They use the prepared checkout as-is and go straight to testing. `run.ts` should drop the `ensureViteCheckout()` call and just clone `vite/`; when `vite/` is missing, fail with a "run `just setup-vite` first" hint, the same way `loadVite()` already does. (The `checkout.ts` import and the tsconfig include can then be reverted again.)

With that rule the checkout and the built dist can never drift apart: the only command that moves the commit also rebuilds everything right after.

For CI this means the vite-tests jobs get a setup step before `run.ts`, like the dev-server workflow already has.

h-a-n-a (approved):
Approved with minor changes in comment section above

packages/vite-tests/run.ts
```diff
@@ -44,16 +45,18 @@ async function runCmdAndPipeOrExit(title: string, cmdOptions: Parameters<typeof
...
-// new tests from Vite `main` surface incompatibilities with rolldown early.
+// Reuse the shared `vite/` checkout at the repo root (kept at the latest
+// `rolldown-canary` rebased onto the latest `main`, see
+// scripts/src/setup-vite/checkout.ts), the same code the dev-server tests
+// run on. The tests run on a throwaway LOCAL clone of it, never on the
+// checkout itself: this suite edits tracked files (pnpm overrides) and the
+// checkout must stay unpatched. The clone shares objects via hardlinks, so
+// no network is needed beyond the checkout itself.
+printTitle('# Ensuring the vite checkout...');
+ensureViteCheckout();
```
hyfdev, line 56: This call can advance a managed `vite/` HEAD without rebuilding `vite/packages/vite/dist`.

`ensureViteCheckout()` (via `updateViteCheckout`) does `fetch` + `checkout -B rolldown-canary origin/rolldown-canary` + `rebase origin/main` when the tree is clean and on `rolldown-canary`. This suite then clones that source into `./repo` and builds there, so vite-tests themselves are fine. The browser suite, however, only checks that `vite/packages/vite/dist/node/index.js` exists (`loadVite` in `packages/test-dev-server/src/vite-server.ts`) and never compares it to HEAD.

Local sequence:

1. `just setup-vite` → `vite/` @ A, dist built from A
2. `vp run --filter vite-tests test` → `vite/` moved to B, dist still from A
3. browser suite → loads dist A while the shared checkout source is B

That is a silent harness divergence for the sequential local workflow this PR documents. CI jobs are separate and always hit the missing-checkout path, so they do not exercise this.

This would have blocked merge. Auto-updating only inside the full `just setup-vite` path (so every HEAD move is followed by install + build + link), or failing loudly when dist does not match HEAD, are possible directions.

scripts/src/setup-vite/checkout.ts
```diff
@@ -23,19 +25,40 @@ export const run = (cmd: string, cwd: string): void => {
...
+    );
+    return;
+  }
+  try {
+    run('git fetch origin main rolldown-canary', viteDir);
+  } catch {
+    console.log('[setup-vite] fetch failed, using the existing checkout as-is');
+    return;
+  }
+  run('git checkout -B rolldown-canary origin/rolldown-canary', viteDir);
```
hyfdev, line 62: `git checkout -B rolldown-canary origin/rolldown-canary` resets the branch tip. A clean local tree that is on `rolldown-canary` with committed-but-unpushed work is treated as managed and those commits are discarded (recoverable only via reflog).

"Taken over" today means dirty working tree or not on the branch name. Local canary experiments that were committed but not pushed are not protected. Documenting that managed checkouts are reset to `origin/rolldown-canary`, or also treating unpushed commits as takeover, would make the contract explicit.

This does not block this approval.

scripts/src/setup-vite/checkout.ts
```diff
@@ -23,19 +25,40 @@ export const run = (cmd: string, cwd: string): void => {
...
+  run('git rebase origin/main', viteDir);
```
hyfdev, line 63: If this rebase fails (conflict, or no committer identity), `execSync` throws and the process exits with a rebase-in-progress or conflicted tree. The next `ensureViteCheckout` sees a non-empty `git status --porcelain` and takes the "using it as-is" branch, so a one-shot failure becomes a sticky broken checkout that later setups treat as intentional developer takeover.

Aborting a failed rebase (or refusing to proceed when a rebase is in progress) would keep the failure recoverable. This does not block this approval.

hyfdev (commented):
Post-merge flag: if this review had landed before merge, the verdict would have been Request changes on one local correctness issue with the shared-checkout design.

`packages/vite-tests/run.ts` calls `ensureViteCheckout()`, which on a managed clean `rolldown-canary` checkout fetches and rebases `vite/` HEAD, but never rebuilds `vite/packages/vite/dist`. The browser suite only checks that the dist file exists, so after a local vite-tests run the two harnesses can diverge: vite-tests on source B, browser suite still loading dist from A. That breaks the "both harnesses run on the same Vite code" invariant this PR states for the local sequential workflow. CI is fine because each job starts without a prior checkout.

Keeping auto-update only in the full `just setup-vite` path (checkout + install + build + link), and having `run.ts` ensure the checkout exists without moving HEAD, is one possible fix. Making any HEAD move force a rebuild before either harness proceeds is another.

I also left two non-blocking notes on the update path (local commits wiped by `checkout -B`, and failed-rebase sticky "as-is").

PR #10326, opened 2026-07-16
fix(magic-string): refuse to use a MagicString after sendMagicString consumed it

IWANABETHATGUY:
> Stacked on #10312 — review that first; this PR's diff is just the top commit.

### What this solves

Returning a `MagicString` from the `transform` hook hands it to `sendMagicString`, which **moves** its contents out to the native sourcemap channel. The JS object outlives that move, so a plugin holding on to the reference got an empty `MagicString` back with no error — every read returned `''`, every edit silently applied to nothing:

```js
transform(code, id, meta) {
  stashed = meta.magicString;
  stashed.append('...');
  return { code: stashed };  // contents moved out here
},
buildEnd() {
  stashed.toString();  // '' — silently, no error
}
```

### The fix

`BindingMagicString` tracks a `consumed` flag: `take_inner()` (the `sendMagicString` path) sets it, and every method that touches the inner `MagicString` goes through an `ensure_live()` check first, erroring with:

> This MagicString was already passed to `sendMagicString()`, which moves its contents into the native sourcemap channel. Finish reading and editing it before returning it from the `transform` hook.

Reads and edits alike refuse — reporting `''` for a string the caller populated is worse than an error naming what happened and what to do instead.

### Tests

An integration fixture (`plugin/native-magic-string-consumed`) drives the real plugin flow: the `transform`-time use still works, the emitted chunk contains the appended code, and post-handoff reuse throws across the API surface (`toString`, `length`, `hasChanged`, `generateMap`, `append`, `overwrite`).

On the parent commit the fixture fails at the first `toThrow` — the consumed instance silently returns instead of refusing. The full JS suite at this branch shows no other change (the only failures are the two fixtures already failing on `main`).

### DIFF
--- a/crates/rolldown_binding/src/options/plugin/binding_transform_context.rs
+++ b/crates/rolldown_binding/src/options/plugin/binding_transform_context.rs
@@ -36,12 +36,15 @@ impl BindingTransformPluginContext {
   pub fn send_magic_string(
     &self,
     magic_string: &mut BindingMagicString<'static>,
-  ) -> Option<String> {
-    let internal_magic_string = std::mem::take(&mut magic_string.inner);
+  ) -> napi::Result<Option<String>> {
+    // This moves the contents out, leaving `magic_string` unusable from JS onwards.
+    let internal_magic_string = magic_string.take_inner();
 
-    // If the the message is not send to main thread correctly, we should panic immediately.
-    self.inner.send_magic_string(internal_magic_string).expect(
-      "TransformPluginContext: failed to send MagicString to sourcemap worker - sourcemap generation thread terminated unexpectedly during transform"
-    )
+    self.inner.send_magic_string(internal_magic_string).map_err(|_| {
+      napi::Error::from_reason(
+        "TransformPluginContext: failed to send MagicString to sourcemap worker - sourcemap \
+         generation thread terminated unexpectedly during transform",
+      )
+    })
   }
 }
--- a/crates/rolldown_binding/src/types/binding_magic_string.rs
+++ b/crates/rolldown_binding/src/types/binding_magic_string.rs
@@ -405,11 +405,36 @@ impl BindingDecodedMap {
 
 #[napi]
 pub struct BindingMagicString<'a> {
-  pub(crate) inner: MagicString<'a>,
+  inner: MagicString<'a>,
   utf16_to_byte_mapper: Utf16ToByteMapper,
   pub(crate) offset: i64,
   indent_exclusion_ranges: Option<IndentExclusionRanges>,
   ignore_list: bool,
+  /// Set once `sendMagicString` has moved `inner` out to the native sourcemap channel.
+  /// The JS object outlives that move, so every method that touches `inner` has to refuse
+  /// rather than silently operate on the empty `MagicString` left behind.
+  consumed: bool,
+}
+
+impl<'a> BindingMagicString<'a> {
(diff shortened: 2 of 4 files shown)


IWANABETHATGUY:
* **#10326** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10326?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10326?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10312** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10312?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10311** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10311?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10326).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

shulaoda (approved):

IWANABETHATGUY:
### Merge activity

* **Jul 16, 5:06 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 16, 5:07 PM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 16, 5:16 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

hyfdev:
The consumed-state guard added here misses the one entry point it exists to protect: repeating the ownership transfer itself. This merged while the review was still in flight, so flagging it post-merge.

[`take_inner`](https://github.com/rolldown/rolldown/blob/1cc2b707470a3b9a8ae1b5e26740f546900742dc/crates/rolldown_binding/src/types/binding_magic_string.rs#L422-L425) sets `consumed = true` but never checks it, and the `sendMagicString` path ([binding_transform_context.rs#L41](https://github.com/rolldown/rolldown/blob/1cc2b707470a3b9a8ae1b5e26740f546900742dc/crates/rolldown_binding/src/options/plugin/binding_transform_context.rs#L41)) calls it without `ensure_live()`. So a second `this.sendMagicString(ms)` on the same instance returns normally, takes the `MagicString::default()` left behind by `std::mem::take`, and queues the empty string into the sourcemap channel.

Reproduced at `1cc2b707` with an exact-head native build, `experimental.nativeMagicString: true` and `sourcemap: true`: one send produced non-empty mappings (`;AAAA,MAAA,QAAA`); sending twice in the same transform hook returned normally and left the final map with mappings `""` — the real map is silently replaced. Every other inner-dependent API correctly throws the new consumed-state error (33 probed); the added fixture enables neither option, so it only exercises the fallback path and cannot catch this.

One possible direction is to have `take_inner` (or the `send_magic_string` binding before it) call `ensure_live()?` and return `napi::Result`, so a repeated send fails with the same error the rest of the API now throws. I verified the failure and the static chain, not a patched implementation; one-send and double-send native-channel cases both need covering tests either way.

PR #10327, opened 2026-07-16
fix(magic-string): validate relocate ranges before rewiring

IWANABETHATGUY:
### What this solves

Follow-up to #10311, addressing [the post-merge report](https://github.com/rolldown/rolldown/pull/10311#issuecomment-4994594426): the guard added there rejects only one non-contiguous topology — `first` having no predecessor and `last` no successor. Other reachable layouts pass it and still corrupt the chunk list:

```js
s.move(0, 2, 3); // list is C -> A -> B
s.move(1, 3, 0); // first = B (the tail), last = C (the head) — guard passes
```

The rewiring then assigns `A.next = A` and `toString()` never returns. With the break in the middle of the list instead, the failure is quieter and worse: `toString()` terminates with wrong output (`"acdb"` becomes `"b"`) and leaves orphaned chunk cycles behind.

Fixing that surfaced two more corrupting inputs that no contiguity check can catch, plus a panic:

- a zero-width range corrupts a pristine string with no prior move at all (`move(1, 1, 3)` on `"abc"` prints `"a"`): the chunk ending at `start` is the one *before* the range, so the code treats a backwards pair as a range;
- after an earlier reorder, moving a chunk to the slot it already occupies drops it (`move(4, 5, 1)` twice on `"abcdef"` prints `"abcdf"`): the splice point resolves to the range's own last chunk, which then links to itself;
- inverted or past-the-end ranges panicked on a map lookup, because `split_at` silently ignores out-of-range positions.

### The fix

Every check runs before any pointer is rewired — the rewiring is not atomic, so bailing out half-way through is exactly the corruption being removed:

- walk the selected range in original order and require each chunk's list successor to be the chunk starting at its end, else error. This subsumes the #10311 guard: a contiguous range open at both ends is the entire string, which is already rejected as moving a selection inside itself;
- treat zero-width ranges as a no-op, mirroring `remove` — with one deliberate exception: the containment guard runs first, so `relocate(n, n, n)` keeps the "inside itself" error, matching both magic-string and this method's previous behavior;
- error on inverted and out-of-bounds ranges instead of panicking;
- generalize the "moving the tail range to the end" early-return to "the chunk at `to` is already the range's list successor", which covers the already-in-position case.

### Tests

One regression test per failure class at the `string_wizard` level; each fails on the parent commit (one by hanging, two by silently wrong output, one by panicking). The #10311 JS test only needed its message regex updated: the old "spans the entire string" message described the list topology rather than the requested range, and the accurate "non-contiguous" error replaces it.

### DIFF
--- a/crates/string_wizard/src/magic_string/movement.rs
+++ b/crates/string_wizard/src/magic_string/movement.rs
@@ -35,6 +35,26 @@ impl MagicString<'_> {
     if to >= start && to <= end {
       return Err("Cannot move a selection inside itself".to_string());
     }
+    // A zero-width range has nothing to move. It also breaks the lookups below — the chunk
+    // ending at `start` is the one *before* the range — and used to self-cycle two chunks.
+    // Deliberately checked after the containment guard above: `relocate(n, n, n)` keeps the
+    // "inside itself" error, matching both magic-string and this method's previous behavior.
+    // Every other zero-width move is a no-op (magic-string corrupts on those, so there is no
+    // upstream behavior to mirror).
+    if start == end {
+      return Ok(self);
+    }
+    if start > end {
+      return Err(format!("end must be greater than start, got start: {start}, end: {end}"));
+    }
+    // Past-the-end positions have no chunk to look up in `chunk_by_end`; indexing the map
+    // with one was a panic.
+    if end > self.source.len() as u32 {
+      return Err(format!(
+        "Cannot move the range ({start}, {end}): it is out of bounds (source length is {})",
+        self.source.len()
+      ));
+    }
 
     self.split_at(start)?;
     self.split_at(end)?;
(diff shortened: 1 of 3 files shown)


IWANABETHATGUY:
* **#10327** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10327?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10327?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10327).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/string_wizard/src/magic_string/movement.rs
```diff
@@ -35,6 +35,22 @@ impl MagicString<'_> {
     if to >= start && to <= end {
       return Err("Cannot move a selection inside itself".to_string());
     }
+    // A zero-width range has nothing to move. It also breaks the lookups below — the chunk
+    // ending at `start` is the one *before* the range — and used to self-cycle two chunks.
+    if start == end {
```
hyfdev, line 44: `relocate(1, 1, 1)` still returns an error before this zero-width no-op runs.

That makes zero-width moves depend on `to`: `relocate(1, 1, 0)` succeeds, while `relocate(1, 1, 1)` returns an error. The PR describes zero-width moves as no-ops, although the latter matches upstream `magic-string`.

Adding a test for the same-point case would make the intended behavior clear. If all zero-width moves are meant to be no-ops, moving this check above the containment guard is one possible way to make that behavior consistent.

hyfdev (approved):
The range validation now rejects non-contiguous layouts before the relocation splice runs, and the regression tests cover the reported failure modes. This looks good overall.

I left one note about the same-point zero-width case.

hyfdev (approved):
The contiguity walk is the real splice precondition, so the layouts that slipped past the #10311 endpoint-only guard are rejected before any list pointers change. Zero-width and already-in-position moves no longer self-link chunks, and inverted or out-of-bounds ranges error instead of panicking. The new `string_wizard` cases cover each failure class from the post-merge report.

The open note on same-point zero-width (`relocate(n, n, n)` still hits the inside-itself guard) still stands and does not block this.

crates/string_wizard/src/magic_string/movement.rs
IWANABETHATGUY: Good catch — the order was accidental as written. I kept it and made it deliberate rather than moving the check: `relocate(n, n, n)` already errored before this PR (the containment guard fires first), and upstream magic-string throws there too, so keeping the error preserves both. The other zero-width moves have no upstream behavior to mirror — magic-string corrupts on them — which is what the no-op replaces.

3f76c55ef adds `zero_width_relocate_to_its_own_position_still_errors` pinning the same-point case next to the zero-width no-op test, plus a comment on the check spelling out the precedence. Also updated the PR description to name the exception.

IWANABETHATGUY:
### Merge activity

* **Jul 18, 10:00 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 18, 10:00 AM UTC**: `IWANABETHATGUY` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 18, 10:05 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

Issue #10019, opened 2026-06-30
Tracking: Full Bundle Mode — Phase 2 (Complete CSR support)


h-a-n-a:
Sub-tracker of #9030 (🎯 Full Bundle Mode) for **Phase 2 — Complete CSR support**. Mirrors Milestone 2 of the FBM roadmap (#153, Milestone 6) / Phase 2 of the Vite [bundled dev mode roadmap](https://github.com/vitejs/vite/discussions/22746).

## Goal

Stabilize FBM for real-world client-side-rendering projects beyond the basic SPA scenario:

- Land the HMR refactor that unblocks plugin support
- Get incremental rebuilds correct
- Expand support for advanced features (globs, watcher, memory)

Priority order: internal-state correctness → HMR refactor (blocks plugin support) → globs & memory.

## Week of July 6th

- [x] Reuse Vite's bundledDev server in test harness @h-a-n-a  https://github.com/rolldown/rolldown/pull/10081
- [x] [properly update internal state for incremental builds](https://github.com/rolldown/rolldown/issues/7416) @shulaoda 
- [x] [**HMR boundary calculation refactor**](https://github.com/rolldown/rolldown/issues/10039) — move boundary computation from server to client; also fixes the lazy-compilation duplicate-execution bug @h-a-n-a 
- [x] [Investigate supporting `hotUpdate` / `handleHotUpdate`, or whether a different mechanism fits](https://github.com/rolldown/rolldown/issues/10065) @h-a-n-a 
- [ ] [rebuild when `tsconfig.json` changes](https://github.com/rolldown/rolldown/issues/9598) @shulaoda 

## Week of July 13th
- [x] [Skip HMR / reload when the output doesn't change (should be relatively easy in FBM)](https://github.com/rolldown/rolldown/issues/10024) cc @h-a-n-a 
- [ ] [Proper `import.meta.glob` support](https://github.com/rolldown/rolldown/issues/10059) — requires watching directories @shulaoda 
- [x] [vitejs/vite#22642](https://github.com/vitejs/vite/pull/22642) — add `input` option (explicit FBM entrypoint instead of overloading `build.rollupOptions.input`) @h-a-n-a 
- [ ] [vitejs/vite#20374](https://github.com/vitejs/vite/pull/20374) — process HTML with the plugin container in dev (aligns `transformIndexHtml` unbundled ↔ FBM / build) @shulaoda 

## Week of July 20th

[...]

PR #10333, opened 2026-07-17
feat(dev): skip the HMR update when a module's rebuilt output is unchanged

### LINKED ISSUE #10024: [Feature Request]: skip the HMR update when a module's rebuilt output is unchanged

In Full Bundle Mode, **every** watched edit that maps to a module renders and ships a full HMR patch. The only early-out today is when a changed file maps to *zero* modules. So a no-op save — or a whitespace / comment-only edit, anything that produces identical rebuilt output — still re-runs the module on the client: `import.meta.hot.accept` handlers re-fire and module-local state is discarded, for an update that changes nothing.

In dev, Vite always triggers HMR/reload. FBM's single bundled module graph makes "did this module's output actually change?" cheap to answer, so FBM can suppress these no-op updates — fewer needless client re-executions, less flicker / state loss on incremental rebuilds. End-user experience: editing a comment, reformatting whitespace, or saving a file unchanged should not blow away component / module state in dev.

This is the (currently unlinked) **"Skip HMR / reload when the output doesn't change"** item from the FBM Phase 2 tracker, #10019.

No public API change — an internal optimization in the HMR stage.

**Proposed behavior:** after rendering each changed / boundary module's new code, compare it against the previously-rendered code; drop the modules whose output is unchanged, and emit `HmrUpdate::Noop` when nothing remains to patch.

**Reproduction** — branch **`test/skip-hmr-when-output-unchanged`**: https://github.com/rolldown/rolldown/tree/test/skip-hmr-when-output-unchanged/crates/rolldown/tests/rolldown/topics/hmr/skip_hmr_when_output_unchanged

```bash
git fetch origin && git switch test/skip-hmr-when-output-unchanged
cargo test -p rolldown --test integration skip_hmr_when_output_unchanged
# then open: crates/rolldown/tests/rolldown/topics/hmr/skip_hmr_when_output_unchanged/artifacts.snap
```

`mod.js` is a self-accepting module:

```js
export const value = 'hello';
import.meta.hot.accept((mod) => {
  assert.strictEqual(mod.value, 'hello');
});
```

The fixture applies three HMR edits to it:

| Step | Edit | Rendered patch output |
| --- | --- | --- |
| 0 | re-save with identical content (no-op save) | identical to the loaded module |
| 1 | whitespace-only change | byte-identical to step 0 |
| 2 | real change (`'hello'` → `'world'`) | different (control) |

What `artifacts.snap` shows today: **HMR Step 0 and Step 1 both emit `update type: patch` with byte-identical `## Code`** even though neither edit changed the bundled output (the HMR printer re-prints the AST, so source whitespace never reaches the output). Step 2 — a real change — produces a different patch, confirming HMR itself works.

**Expected once implemented:** steps 0 and 1 become `update type: noop` (no patch sent, no client re-run); step 2 stays a `patch`. The committed snapshot is the regression guard — that delta is exactly what the fix should flip.

**Suspected root cause** — `crates/rolldown/src/hmr/hmr_stage.rs`:

- `compute_hmr_update_for_file_changes()` maps changed files → `changed_modules`; the sole early-out is `if changed_modules.is_empty() → HmrUpdate::Noop` (~`hmr_stage.rs:197`).
- `compute_hmr_update()` → `render_hmr_patch_from_prerequisites()` (`hmr_stage.rs:299`, `:816`) renders the patch unconditionally; nothing compares the freshly-rendered module code against its previously-rendered form. The comparison proposed above is the missing piece.

**Environment for the repro:** in-repo Rust fixture (no published rolldown package) · Rolldown commit `aa32751cf` (origin/main) · rustc 1.96.0 · Node 24.12.0 · macOS 26.5.1 (arm64).

h-a-n-a:
### Description

In Full Bundle Mode dev, every watched edit shipped a full HMR patch — a no-op save or a whitespace/comment-only edit still re-ran the module on the client and discarded its state.

Now each changed module is rendered once before the cache merge and once after, with the same settings (`unique_index` pinned to 0, no sourcemap). If the two strings are byte-identical, the module is dropped before stamping, so it never marks clients stale, never seeds the importer walk, and never enters `changed_ids`. A client left with nothing gets `HmrUpdate::Noop` instead of an empty patch.

```
watch event: save mod.js (whitespace-only edit)

  render(old AST) ──► "const value = ..."   ┐
                                            ├─ equal? ──► drop from changed_modules
  merge, re-scan                            │
  render(new AST) ──► "const value = ..."   ┘
                                                └─ changed_modules empty ──► Noop, nothing ships
```

Example: `export const value = 'hello'` re-saved with blank lines removed prints the same output → `noop`. Changing `'hello'` to `'world'` prints different output → `patch`.

Dep identity and interop kind are part of the compared string (`loadExports("<stable id>")`, `registerFactory(id, kind, ...)`), so a resolution shift under an unchanged source still ships.

One exception: after an errored build the suppression is disabled (`DevContext::last_task_errored`), so a recovery that rebuilds byte-identical output still reaches clients stuck on the error overlay.

Fixes #10024 (from the FBM Phase 2 tracker #10019).

Note: the `hmr-full-bundle-mode > invalidate` failure in the Vite Test job also fails on `main` — canary drift that needs a canary-side fix, not a change here.


h-a-n-a:
* **#10333** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10333?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10333?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10333).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

h-a-n-a:
### Merge activity

* **Jul 17, 9:56 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 17, 11:34 AM UTC**: `h-a-n-a` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 17, 11:40 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

crates/rolldown/src/hmr/hmr_stage.rs
```diff
@@ -143,6 +144,36 @@ impl<'a, Fs: FileSystem + Clone + 'static> HmrStage<'a, Fs> {
       }
     }
 
+    // After an errored build (`last_build_errored`) the capture is skipped, so
+    // every changed module ships. A failed scan merges nothing, so undoing the
+    // broken edit rebuilds byte-identical output — the graph can't tell "broke,
+    // then fixed" from "nothing happened", but clients are stuck on the error
+    // (overlay / fallback page) and a suppressed update would leave them there.
+    let pre_rebuild_renders = if last_build_errored {
```
hyfdev, line 160: `last_build_errored` is cleared whenever the previous task succeeds, including a successful empty `Noop` that returns before `pending_rescans` are folded into `changed_modules`. After a failed partial scan, an intervening watched non-module path can therefore clear the latch; a later restore of the broken file to its original bytes is then suppressed as unchanged, and clients stuck on the error overlay never get a recovery patch.

The direct recovery path still works when the next edit is the broken file itself (the `error_recovery` fixtures). The hole needs that intervening empty success between the failure and the identical restore.

Folding `pending_rescans` into `changed_modules` before the empty early-return, or keeping the recovery latch sticky until a real patch is delivered, would close the gap. This does not block this approval.

hyfdev (approved):
Factory output is compared with fixed render settings before stamping, so no-op and whitespace-only edits drop cleanly, mixed batches only ship real changes, and the direct error-recovery path still forces a patch. This looks good overall.

I left one note about clearing the recovery latch when an empty successful task runs before pending rescans are retried.

sapphi-red (commented):
The tests lgtm

crates/rolldown/src/hmr/hmr_stage.rs
h-a-n-a: Thanks! This has been fixed by putting `pending_rescans` before `changed_modules`.

PR #10346, opened 2026-07-19
fix(generator): filter DISABLE_JS_HOOK bits from generated HookUsage

Brooooooklyn:
### Summary

Extracted from #10268 as a standalone fix (part of the breakup of that PR).

The generated TS `HookUsageKind` enum included `transformAst = 1 << 16` even though `transform_ast` is listed in `DISABLE_JS_HOOK` and can never be registered from a JS plugin (`extractHookUsage` already skipped it). This change filters disabled hooks out of the generated enum too, so the JS side no longer exposes a usage bit for a hook it cannot run.

### Changes

- `tasks/generator/src/generators/hook_usage.rs`: filter `DISABLE_JS_HOOK` entries out of the `HookUsageKind` enum generation. The filter runs after `enumerate()`, so the remaining variants keep their bit positions aligned with the Rust `HookUsage` bitflags (which intentionally keep the `TransformAst` bit). Adds a unit test locking in both properties, and drops a stray trailing-whitespace line from the `extractHookUsage` template.
- `tasks/generator/Cargo.toml`: `[lib] test = true` so the new unit test actually runs.
- `packages/rolldown/src/plugin/generated/hook-usage.ts`: regenerated (`cargo run --bin generator`) — only the dead `transformAst = 1 << 16,` member is removed. It was referenced nowhere else in the repo.

The three files land together because CI's repo-validation job re-runs the generator and fails on drift.

### Verification

- `cargo test -p generator --lib`: 1 passed (`disabled_js_hooks_preserve_the_shared_bit_positions`)
- `cargo run --bin generator && git diff --exit-code`: clean (regeneration is idempotent)
- `cargo clippy -p generator --all-targets`, `cargo fmt --check`: clean
- `crates/rolldown_plugin/src/generated/hook_usage.rs` is untouched by regeneration (Rust bits unchanged)

---

> [!NOTE]
> **Low Risk**
> Codegen-only alignment between TS and existing Rust bitflags; no change to which hooks run or how usage bits are computed for supported hooks.
> 
> **Overview**
> Aligns generated JS **`HookUsageKind`** with hooks JS plugins can actually register: **`transformAst`** is no longer emitted even though it sits in **`DISABLE_JS_HOOK`** and **`extractHookUsage`** already ignored it.
> 
> The generator filters disabled hooks **after** **`enumerate()`**, so remaining enum members keep the same bit shifts as Rust **`HookUsage`** (e.g. **`banner`** stays **`1 << 17`**). A unit test locks that behavior in, and **`generator`**’s **`[lib] test = true`** enables it. **`hook-usage.ts`** is regenerated so CI’s generator check stays clean.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a8ce54867849b668bcaeeba2c321f97f89f839f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/rolldown/src/plugin/generated/hook-usage.ts
```diff
@@ -18,7 +18,6 @@ export enum HookUsageKind {
   closeBundle = 1 << 13,
   watchChange = 1 << 14,
   closeWatcher = 1 << 15,
-  transformAst = 1 << 16,
   banner = 1 << 17,
```
IWANABETHATGUY, line 21: We should reorder the rest hook, after `transformAst` hook is eliminated

PR #10349, opened 2026-07-19
refactor: descriptor-based parallel-plugin detection (getParallelPluginInfo)

Brooooooklyn:
### Description

Standalone extraction from #10268 (part of breaking that PR into reviewable pieces).

Replaces the `'_parallel' in plugin` sniffing at all four call sites with one shared helper, `getParallelPluginInfo` (new `packages/rolldown/src/utils/parallel-plugin.ts`). Detection is descriptor-based: only an **own data property** named `_parallel` whose value is an object with a string `fileUrl` counts as a parallel plugin marker, and the helper returns the validated payload so callers never re-read it unchecked.

Why the `in` operator is a problem (both reproduced on main):

- A regular plugin carrying a malformed own `_parallel` (e.g. `_parallel: null`) crashes `initializeParallelPlugins`:
  `TypeError: Cannot destructure property 'fileUrl' of 'plugin._parallel' as it is null.`
- A plugin that merely *inherits* `_parallel` from its prototype chain is silently routed into the parallel worker registry (its real hooks are dropped and the workers try to import a bogus `fileUrl`).

Converted call sites (all of main's occurrences, so every site agrees on one predicate):

- `src/plugin/plugin-driver.ts` (`getObjectPlugins`)
- `src/utils/normalize-plugin-option.ts` (`normalizePlugins`)
- `src/utils/bindingify-input-options.ts`
- `src/utils/initialize-parallel-plugins.ts`

Compared to the #10268 version, the helper here is trimmed to the clean detection refactor: no `try/catch` proxy paranoia, no per-field descriptor walking of the marker payload, and none of the runtime-branch capability gates.

### Test Plan

- `tsc -p tsconfig.check.json` — passes
- `vitest run fixture.test.ts fixtures-concurrent.test.ts error/error.test.ts` — 427 passed
- `pnpm run test:cli` — 61 passed, 1 skipped
- `vitest run exports-consistency.test.ts validate-option.test.ts` — 5 passed
- Manual end-to-end smoke (no vitest suite covers parallel plugins today):
  - a genuine `defineParallelPlugin` plugin is still routed into the worker registry exactly as on main
  - a plugin with own `_parallel: null` now builds and runs as a regular plugin (crashed with the destructuring `TypeError` on main)
  - a plugin inheriting `_parallel` from its prototype now runs as a regular plugin (was misrouted to workers on main)

Note: while smoking this on Node 24 I hit a pre-existing failure in the parallel worker flow itself (`napi` ThreadsafeFunction `code: 'Closing'` errors once the unref'd workers exit). It reproduces identically with unmodified main and with the published `rolldown@1.2.0` from npm, so it is unrelated to this refactor — flagging it since parallel plugins have no CI coverage.

---

> [!NOTE]
> **Medium Risk**
> Touches plugin routing and parallel worker initialization; behavior changes for edge-case plugins but genuine `defineParallelPlugin` markers should behave the same.
> 
> **Overview**
> Centralizes how Rolldown recognizes **parallel plugins** by introducing `getParallelPluginInfo` and using it everywhere that used to check `'_parallel' in plugin`.
> 
> Detection now requires an **own data** `_parallel` property whose value is an object with a string `fileUrl`, instead of treating any inherited or malformed `_parallel` as parallel. That fixes crashes when `_parallel` is `null` and stops regular plugins with a prototype `_parallel` from being sent to worker threads.
> 
> Call sites updated: `getObjectPlugins`, `normalizePlugins`, `bindingify-input-options`, and `initializeParallel-plugins` (which now destructures the validated payload from the helper).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a7a3e6df1f8feedf1fdb19bae9eee243b0e260c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/rolldown/src/plugin/plugin-driver.ts
+++ b/packages/rolldown/src/plugin/plugin-driver.ts
@@ -5,6 +5,7 @@ import { getLogger, getOnLog } from '../log/logger';
 import { LOG_LEVEL_INFO, type LogLevelOption } from '../log/logging';
 import { normalizeHook } from '../utils/normalize-hook';
 import { normalizePluginOption } from '../utils/normalize-plugin-option';
+import { getParallelPluginInfo } from '../utils/parallel-plugin';
 import type { Plugin } from './';
 import { MinimalPluginContextImpl } from './minimal-plugin-context';
 
@@ -73,7 +74,7 @@ export function getObjectPlugins(plugins: RolldownPlugin[]): Plugin[] {
     if (!plugin) {
       return undefined;
     }
-    if ('_parallel' in plugin) {
+    if (getParallelPluginInfo(plugin)) {
       return undefined;
     }
     if (plugin instanceof BuiltinPlugin) {
--- a/packages/rolldown/src/utils/bindingify-input-options.ts
+++ b/packages/rolldown/src/utils/bindingify-input-options.ts
@@ -18,7 +18,7 @@ import type { LogHandler } from '../log/log-handler';
 import type { LogLevelOption } from '../log/logging';
 import type { AttachDebugOptions, DevModeOptions, InputOptions } from '../options/input-options';
 import type { OutputOptions } from '../options/output-options';
-import type { RolldownPlugin } from '../plugin';
+import type { Plugin, RolldownPlugin } from '../plugin';
 import { bindingifyPlugin } from '../plugin/bindingify-plugin';
 import type { PluginContextData } from '../plugin/plugin-context-data';
 import { arraify } from './misc';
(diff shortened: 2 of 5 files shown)


hyfdev:
### Merge activity

* **Jul 20, 10:31 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 20, 10:32 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 20, 10:38 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

hyfdev (approved):
parallel plugin is currently in a not-active status. We haven't found value in this part, so no effort putting on this.

PR #10353, opened 2026-07-19
refactor: remove AstFactory in favor of new_* construction traits

### LINKED ISSUE #10328: Proposal: Remove `AstFactory`
## `AstBuilder` pattern

The pattern for Oxc's new version of `AstBuilder` is:

* All builder methods are implemented on AST types themselves.
* They all take a `&B where B: GetAstBuilder`.
* `GetAstBuilder` is implemented on all structs which hold an `AstBuilder` e.g. `ParserImpl`, `TraverseCtx`.

```rs
impl<'a> UnaryExpression<'a> {
    pub fn new<B: GetAstBuilder<'a>>(
        span: Span,
        operator: UnaryOperator,
        argument: Expression<'a>,
        builder: &B,
    ) -> Self {
        let builder = builder.builder();
        UnaryExpression { node_id: Cell::new(builder.node_id()), span, operator, argument }
    }
}
```

The `B: GetAstBuilder` generic shortens call sites - you can pass `self` rather than e.g. `&self.ast_builder`:

```rs
// `self` here is e.g. `&TraverseCtx`
let unary_expr = UnaryExpression::new(span, argument, self);
```

## Proposal

I propose to migrate `AstFactory` to the same pattern.

Before:

```rs
self.ast_factory.make_id_ref_expr(SPAN, &binding_name)
```

After:

```rs
Expression::new_id_ref(SPAN, &binding_name, self)
```

All methods which are currently methods of `AstFactory` would become e.g.:

```rs
pub trait ExpressionMethods<'ast> {
    fn new_id_ref<B: GetAstBuilder<'ast>>(span: Span, name: &str, builder: &B) -> Expression<'ast> {
        Expression::new_identifier(span, Str::from_str_in(name, builder), builder)
    }
}

impl<'ast> ExpressionMethods<'ast> for Expression<'ast> {}
```

Having to use traits is annoying, but it's the only way Rolldown can implement methods on Oxc's foreign types.

We can, however, make it painless at call sites. API users would just add at top of file:

```rs
use rolldown_ecmascript_utils::ast_factory::*;
```

Once this change is made, `AstFactory` can be deleted altogether - Rolldown would just use Oxc's `AstBuilder` directly.

## Why?

The main reasons we went with the new design of `AstBuilder` in Oxc are:

1. Enables making `AstBuilder` stateful to support automatically assigning unique `NodeId`s to all AST nodes.
2. Shorter syntax - pass `self`, not `self.ctx.ast_builder`.
3. Passing `self` as *last* param avoids borrow-checker problems. Patterns like this pass borrow-check where they wouldn't before:

```rs
// `generate_id` takes `&mut self`

// New pattern - this compiles fine
let id = Expression::new_identifier(span, self.generate_id(), self);

// Old pattern - borrow-check refuses (`self` already borrowed)
let id = self.ast_builder.identifier(span, self.generate_id());
```

IMO it's important to make assignment of unique `NodeId`s consistent and automatic across both Oxc and Rolldown, so that `NodeId`s can be used reliably as unique identifiers. This proposed change to `AstFactory` would make this goal easier to achieve.

### Further changes to come

There is a further change I'd like to make to `AstBuilder`. It'd be ideal if methods took a `&mut AstBuilder` not `&AstBuilder`. This would allow us to make `AstBuilder` stateful without having to use interior mutability (`Cell`s), which would be both a performance gain and a simplification.

That change *should* work with few code changes to either Oxc or Rolldown if we use the new methods-on-AST-types pattern, but would be *really* unpleasant with the old pattern - borrow-checker will go bananas.

## Alternative

If Rolldown team strongly feel that you want to keep `AstFactory` as it is, we *could* work around its limitations and achieve unique `NodeId`s.

However, we'd either have to lose the "further changes to come" idea, or make a *lot* of tortuous changes to Rolldown to work around borrow-checker problems.

## Who does it?

If you accept my proposal, I would be very happy to make a PR for the changes. The codemod we used to migrate Oxc over to new `AstBuilder` could be easily adapted to do this migration too.

Boshen:
Removes the `AstFactory` newtype (rolldown's wrapper over oxc's `AstBuilder`) and moves rolldown's recurring AST constructions onto per-node-type extension traits, matching oxc's own methods-on-types construction shape. Rolldown now holds and passes oxc's `AstBuilder` directly.

## What

- `AstFactory`'s ~28 inherent `make_*` methods become `new_*` associated functions on `*FactoryExt` traits — `ExpressionFactoryExt`, `StatementFactoryExt`, `MemberExpressionFactoryExt`, `CallExpressionFactoryExt`, `BindingIdentifierFactoryExt`, `IdentifierNameFactoryExt`, `ObjectPropertyKindFactoryExt`, `ClassElementFactoryExt` — each generic over `B: GetAstBuilder<'ast> + GetAllocator<'ast>` and taking the builder as its **last** argument (e.g. `Expression::new_id_ref_expr(SPAN, name, builder)`).
- `BindingPatternExt` / `BindingPropertyExt` are made generic over the builder the same way, decoupling them from the deleted type.
- Every holder struct's `ast_factory: AstFactory` field becomes `ast_builder: AstBuilder`; call sites go from `self.ast_factory.make_x(..)` to `Type::new_x(.., &self.ast_builder)`.
- `internal-docs/ast-construction/implementation.md` is updated to describe the new convention.

## Why

- **Matches oxc.** Construction lives on the AST types for both oxc and rolldown nodes, builder passed last.
- **Builder-last unblocks `&mut` construction.** Passing the builder as an argument rather than a `self` receiver is what lets oxc later switch builder methods to `&mut AstBuilder` — dropping the `Cell`s from `Allocator` (a hot-path win) and enabling stateful builders that auto-assign unique `NodeId`s.
- **Statefulness stays reintroducible.** Because every constructor is generic over the builder, a future stateful wrapper (`impl GetAstBuilder + GetAllocator`) can be threaded through with zero call-site changes.

## Notes vs the proposal

Two intentional differences from the sketch in the issue:

- Method names keep the operation suffix (`new_id_ref_expr`, not `new_id_ref`); the `new_*` names are operation-based (`new_keep_name_call`), disjoint from oxc's variant-based constructors (`new_call_expression`), so none collide.
- Holder structs keep an `ast_builder` field and pass `&self.ast_builder`, rather than implementing `GetAstBuilder` themselves to pass `self`. The borrow-check benefit still holds via argument evaluation order; the shorter `self` call syntax is a possible follow-up.

Closes #10328

### DIFF
--- a/crates/rolldown/src/ecmascript/format/esm.rs
+++ b/crates/rolldown/src/ecmascript/format/esm.rs
@@ -275,7 +275,7 @@ fn render_chunk_content<'code>(
     // The concatenated closure holds the bodies of every module in the group. If the group's
     // entry is TLA-tainted (has top-level `await`, or awaits a TLA importee's `init` call), those
     // `await`s land inside this closure, so it must be `async`. This mirrors the flag threaded
-    // into `make_esm_wrapper_stmt` for the non-concatenated wrapper.
+    // into `new_esm_wrapper_stmt` for the non-concatenated wrapper.
     let is_async = ctx.link_output.metas[group.entry].is_tla_or_contains_tla_dependency;
 
     if !hoisted_fns.is_empty() {
--- a/crates/rolldown/src/hmr/hmr_ast_finalizer.rs
+++ b/crates/rolldown/src/hmr/hmr_ast_finalizer.rs
@@ -3,18 +3,25 @@ use oxc::ast::ast::Str;
 use oxc::{
   allocator::IntoIn,
   ast::{
-    ast::{self, ExportDefaultDeclarationKind, Expression, ObjectPropertyKind, Statement},
+    ast::{
+      self, BindingIdentifier, ExportDefaultDeclarationKind, Expression, IdentifierName,
+      ObjectPropertyKind, Statement,
+    },
     builder::NONE,
   },
   semantic::{IsGlobalReference, Scoping, SymbolId},
   span::{SPAN, Span},
 };
 
+use oxc::ast::builder::AstBuilder;
 use rolldown_common::{
   ExternalModule, ImportRecordIdx, ImportRecordMeta, IndexModules, Module, ModuleIdx, NormalModule,
 };
 use rolldown_ecmascript::CJS_REQUIRE_REF_STR;
-use rolldown_ecmascript_utils::{AstFactory, ExpressionExt};
+use rolldown_ecmascript_utils::{
+  BindingIdentifierFactoryExt as _, ExpressionExt, ExpressionFactoryExt as _,
+  IdentifierNameFactoryExt as _, ObjectPropertyKindFactoryExt as _, StatementFactoryExt as _,
+};
 use rolldown_utils::{
   ecmascript::is_validate_identifier_name,
   indexmap::{FxIndexMap, FxIndexSet},
@@ -25,7 +32,7 @@ use crate::hmr::utils::{HmrAstBuilder, MODULE_EXPORTS_NAME_FOR_ESM};
 
 pub struct HmrAstFinalizer<'me, 'ast> {
   // Outside input
-  pub ast_factory: AstFactory<'ast>,
+  pub ast_builder: AstBuilder<'ast>,
   pub modules: &'me IndexModules,
   pub module: &'me NormalModule,
   pub use_pife_for_module_wrappers: bool,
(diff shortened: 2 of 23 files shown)


shulaoda (approved):

shulaoda:
### Merge activity

* **Jul 19, 3:20 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 19, 3:20 PM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 19, 3:25 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10364, opened 2026-07-20
feat(devtools): metrics mode - agent-readable build metrics report

IWANABETHATGUY:
`devtools: { mode: 'metrics' }` folds the devtools event stream in-memory into a build report under `node_modules/.rolldown/metrics/` instead of writing the JSON-lines event log, so that a coding agent can read sizes, per-hook timings and the module graph without parsing the event stream. `mode: 'full'` is unchanged.

Draft until I have build-overhead numbers for `mode: 'metrics'` vs devtools off, and docs.


Boshen (commented):
`metrics` or `markdown`? Matching for format instead of usecase.

Also, does it make sense to output both formats?

IWANABETHATGUY:
> `metrics` or `markdown`? Matching for format instead of usecase.
> 
> Also, does it make sense to output both formats?


The original `devtools` generated large raw JSON. However, always producing raw JSON did not make sense because you never want the agent to read the full raw JSON—that would increase token usage significantly. The current strategy is that when `metric` mode is used, a markdown summary plus trimmed JSON is generated, optimized for the agent.

Regarding the first question, I agree with you that we should use a better name than `metric`.

webfansplz:
Thanks for the PR! We're building MCP tooling for vite-devtools on top of rolldown's build data, and there are a couple of things we'd like to discuss for that use case:

## 1. Could full mode also emit the metrics report?

The devtools MCP is read-only by design — it never triggers builds, it just reads whatever the user's build produced. With full and metrics being mutually exclusive, users end up choosing between having the DevTools UI and having the MCP work, but those two consumers are usually the same person, same project, same build.

Since the metrics aggregation happens in memory over the same event stream, the marginal cost of emitting metrics on top of full mode should mostly be writing a few small files at the end. So instead of exclusive modes, could the two be composable — e.g. full mode also emits metrics, by default or opt-in? One build would then serve both the UI (event logs) and agents (metrics). Metrics-only still makes total sense for agent loops where you want to skip the full-log overhead — we really like that part.

One small related ask: a reliable way to correlate the metrics output with the corresponding session. When both are present, the flow we have in mind is "use metrics to locate the problem fast, then drill into that session's event logs for details" — which needs a link between the two. That could be nesting the metrics output under the session folder, or including the session id in `metrics.json` and `history.jsonl` — whichever fits the design better (though `history.jsonl` probably wants it either way, since `buildIndex` resets per session and lines from different sessions would collide).

## 2. A few extra fields

We checked the metrics output against what our existing analysis tools need. The following seem to already exist in the internal aggregation structs — they're just not exported:

**module-graph.json** (it's already the "filter it, never load it whole" file, so per-node fields shouldn't conflict with the condensed-report design):

- `package` on each module node (the package it belongs to, `name@version`). The event stream already has this via `MPackage.modules`. We need it to answer "why is package X in my bundle" (package-target dependency trace) — without it we're stuck inferring from `node_modules` paths, which falls apart with pnpm and monorepos.
- `chunk` on each module node (the chunk file it ended up in). We need it for chunk/asset-target traces, "which chunks contain this module", and initial/async classification. One field covers all of these, and avoids exporting a full chunk→modules list in `metrics.json` (which admittedly wouldn't fit the condensed design).
- Keep the import kind (or at least a static/dynamic flag) on `imports` edges — if they're currently serialized down to bare module ids. We need it to mark dynamic boundaries along trace paths.

**metrics.json**:

- `entryModule` on `chunks.largest[]` (`ChunkAgg` already has it). Lets us go from a chunk or asset back to its entry module.
- Truncation markers on the top-N lists (`truncated: true` or a `totalCount`). An agent needs to know whether something is missing or just not listed, so it can decide when to drill into `module-graph.json` or the full logs. Related question: is the `assets` array complete, or capped at top-N as well?
- The full plugin list in `session` (`[{ index, name }]`). Right now there's only `pluginCount`, so plugins that never had a hook called are invisible, and we can't reliably answer "which plugins were in this build".

There are two more module-level things we'd use — the per-module transform size chain (before/after, for attributing transform bloat and estimating tree-shaking) and per-module resolve/load timings (for pinpointing modules under load/resolve-heavy plugins) — but we get that those might be out of scope for a condensed report. Happy to open a separate issue for them; on the MCP side we can fall back to the full logs for now.

PR #10254, opened 2026-07-13
fix(sourcemap): preserve unmapped boundaries during composition

hyfdev:
## Why

This PR is stacked on #10249.

A source-map segment changes the mapping state from its generated column until the next segment. It is not just a label for one character. A segment with a source starts a mapped range; a one-field segment with only a generated column starts an explicitly unmapped range.

For example, the transform fixture returns `AAAA,K` for this generated line:

```text
columns    012345678901234567
generated  foo(); injected();

column 0   start mapping to main.js
column 5   stop mapping
```

The intended ranges are therefore:

```text
Mapped    generated 0:0–0:5   -> main.js 0:0
Unmapped  generated 0:5–EOL
```

### What a source-map visualizer shows

The blocks below represent the colored spans that a visualizer associates with the original source. `█` is mapped and `░` is uncolored or explicitly unmapped.

Before this PR, composition dropped the one-field segment at column 5:

```text
input map      AAAA,K
collapsed map  AAAA

generated  foo(); injected();
visualized ██████████████████
```

With no boundary after column 0, the previous mapping remains active to the end of the line. `injected()` is colored as if it came from `foo()`; asking a source-map consumer for the original position of `injected()` returns the earlier `main.js` mapping.

After this PR, composition preserves the boundary:

```text
input map      AAAA,K
collapsed map  AAAA,K

generated  foo(); injected();
visualized █████░░░░░░░░░░░░░
```

The generated `foo()` remains associated with the original source, while the semicolon and `injected()` have no original position. A visualizer leaves that range uncolored, and a debugger or other source-map consumer no longer attributes it to `foo()`.

The same rule applies when the unmapped boundary appears in an intermediate map instead of the final map.

Rollup currently drops these one-field segments while collapsing maps. This PR deliberately differs from that behavior so an explicit unmapped range remains explicit after composition.

## What changed

- stop tracing when a final or intermediate token has no source
- emit an unmapped token at the corresponding final generated position
- cover final and intermediate boundaries with Rust tests
- cover a plugin transform returning `AAAA,K`

## Validation

- `cargo test -p rolldown_sourcemap`
- `just test-rust`
- `just test-node-rolldown-only fixtures-concurrent.test.ts -t plugin/transform/explicit-unmapped-boundary`
- `cargo clippy -p rolldown_sourcemap --all-targets -- --deny warnings`

PR #10368, opened 2026-07-21
test(sourcemap): fix Windows-only failures in composition fixtures

shulaoda:
### Description

The `node-test-windows` job has been failing on every `main` run since 2026-07-15 (for example [this run](https://github.com/rolldown/rolldown/actions/runs/29795703407/job/88527096422)). The two fixtures added in #10249 and #10254 guard their `load` and `transform` hooks with `id.endsWith('/main.js')`. On Windows, module ids use backslashes, so the guard never matches and the hooks are silently skipped.

With the hooks skipped, `coarse-sourcemap-composition-issue-6399` bundles the on-disk comment-only `main.js` into an empty chunk whose `map` is `null`, so `new TraceMap(null)` throws `Cannot read properties of null (reading '_decodedMemo')`. In `explicit-unmapped-boundary`, the transform never injects `injected()`, so `code.indexOf('injected()')` returns -1 and the assertion fails.

This PR drops the leading slash so the checks become `id.endsWith('main.js')`, which is the pattern every other fixture in the repo already uses and works with both path separators. No production code is touched.

### DIFF
--- a/packages/rolldown/tests/fixtures/plugin/transform/coarse-sourcemap-composition-issue-6399/_config.ts
+++ b/packages/rolldown/tests/fixtures/plugin/transform/coarse-sourcemap-composition-issue-6399/_config.ts
@@ -46,7 +46,7 @@ export default defineTest({
       {
         name: 'load-detailed-source-map',
         load(id) {
-          if (!id.endsWith('/main.js')) {
+          if (!id.endsWith('main.js')) {
             return;
           }
           return {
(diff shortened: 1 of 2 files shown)


hyfdev (approved):

hyfdev:
### Merge activity

* **Jul 21, 9:59 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 21, 9:59 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 21, 10:05 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10374, opened 2026-07-22
feat(minify): support property name mangling

Dunqing:
Adds `output.minify.mangleProps` to Rolldown's JavaScript API, native and WASI bindings, normalized output options, and chunk minification.

### Feature

- `include` and `exclude` accept JavaScript `RegExp` values.
- `reserved`, `quoted`, `debug`, and `cache` are passed to Oxc.
- Normalized options keep each `RegExp` source and flags.
- Rolldown uses Oxc to mangle properties in a single JavaScript output chunk.
- The updated cache is returned as `RolldownOutput.mangleCache`.

Pass the returned cache to `mangleProps.cache` when minifying a later unminified version of the same source to keep property names stable.

### Benchmark

I benchmarked [Sveltia CMS at `a27c736`](https://github.com/sveltia/sveltia-cms/tree/a27c736dbb05c564dcecf734aa12f60407e24ba6), a real project linked by its maintainer in a [Vite issue comment](https://github.com/vitejs/vite/issues/22779#issuecomment-4850586465). Its production IIFE output contains one JavaScript chunk, matching the current limitation of property mangling.

The benchmark used the stable releases available on 2026-08-27: Vite 8.2.2 with Rolldown explicitly pinned to 1.2.6. Both variants used the same minifier configuration; the enabled variant added only:

```js
mangleProps: {
  include: /^_(?:i18n|type|file|thumbnailFieldNames|fileMap)$/,
  quoted: true,
}
```

Three fresh Node.js 24.20.0 processes each ran 10 warm-up pairs and 30 measured pairs, for 90 measured pairs on an Apple M4 Max.

| Measurement | Disabled | Enabled | Change |
| --- | ---: | ---: | ---: |
| In-memory `vite.build()` median | 987.01 ms | 998.63 ms | +11.63 ms |
| Raw JavaScript | 2,001,264 B | 2,000,778 B | -486 B (-0.0243%) |
| gzip level 9 | 613,427 B | 613,317 B | -110 B (-0.0179%) |
| Brotli quality 11 | 491,776 B | 491,791 B | +15 B (+0.0031%) |

The median paired timing change was +1.25%, with a bootstrap 95% confidence interval of +0.70% to +1.49% on this machine. The output-size difference was effectively neutral for this conservative five-property allowlist.

The timer covered Vite transformation, Rolldown bundling and minification, source-map generation, and in-memory output generation for the IIFE. It did not include configuration loading, disk writes, compression, packaging-only plugins, or Sveltia's second ESM output, so these numbers are not full `pnpm build` wall time.

The pinned source tree passed all 8,202 upstream tests. Both generated artifacts also passed the same headless-browser CMS flow with no console, page, or network errors, and an artifact diff found only the 135 expected occurrences of the five allowlisted property names.

### Limitation

Property mangling currently supports only one JavaScript output chunk. Rolldown returns an error when `mangleProps` is used with multiple JavaScript chunks.

Property mangling recognizes property syntax and annotated key strings. It cannot find property names in arbitrary runtime strings.

Reserve names accessed indirectly, through module namespace objects, by external or host code, or by code added after minification.

Tests cover the single-chunk output, returned cache and reuse, reserved names, regular expression flags, source maps, normalized options, and the multi-chunk error.

AI assistance disclosure: OpenAI Codex and Claude were used for research, implementation, adversarial review, testing, and PR preparation. The author will review and take responsibility for the changes before marking the draft ready.


izaakschroeder:
Appreciate this change 😄

packages/rolldown/src/options/output-options.ts
```diff
@@ -100,9 +104,26 @@ export type CodeSplittingNameFunction = (
 /** @inline @category Code Splitting */
 export type CodeSplittingTestFunction = (id: string) => boolean | undefined | void;
```
sapphi-red, line 106: Not directly related but we should also update https://oxc.rs/docs/guide/usage/minifier.html

packages/rolldown/src/options/output-options.ts
```diff
@@ -100,9 +104,26 @@ export type CodeSplittingNameFunction = (
...
 
-// `mangleProps` is not plumbed through the bundler yet — `normalize_binding_options` only
-// reads `mangle`, `compress` and `codegen`, so keep it off the public type.
-export type MinifyOptions = Omit<BindingMinifyOptions, 'module' | 'sourcemap' | 'mangleProps'>;
+export interface ManglePropertiesOptions extends Omit<BindingManglePropertiesOptions, 'cache'> {
+  /**
+   * Stable mappings from original names to output names. `false` reserves an original name.
+   * String targets must be valid identifiers and cannot be `__proto__`, `constructor`, or
+   * `prototype`. Rolldown does not return generated mappings.
```
sapphi-red: > Rolldown does not return generated mappings.

Probably we can expose this in `RolldownOutput`
https://rolldown.rs/reference/Interface.RolldownOutput

sapphi-red (approved):

sapphi-red:
> Property mangling currently supports only one JavaScript output chunk.

(just in case) If this is expected to take time, it would be nice to have an issue so that users can track the feature.

Dunqing:
> > Property mangling currently supports only one JavaScript output chunk.
> 
> (just in case) If this is expected to take time, it would be nice to have an issue so that users can track the feature.

I will do it soon

Dunqing:
> > > Property mangling currently supports only one JavaScript output chunk.
> > 
> > 
> > (just in case) If this is expected to take time, it would be nice to have an issue so that users can track the feature.
> 
> I will do it soon

created https://github.com/rolldown/rolldown/issues/10771

packages/rolldown/src/options/output-options.ts
Dunqing: Good idea. Oxc already returns the updated property cache, so I’ll expose it as `RolldownOutput.mangleCache?: Record<string, string | false>`.

It will contain the supplied entries and newly generated mappings, so it can be passed directly to `mangleProps.cache` in a later build.


What do you think?

packages/rolldown/src/options/output-options.ts
Dunqing: https://github.com/oxc-project/website/pull/1274

packages/rolldown/src/options/output-options.ts
sapphi-red: Sounds good to me!

h-a-n-a (approved):

packages/rolldown/src/options/output-options.ts
Dunqing: Done in https://github.com/rolldown/rolldown/pull/10374/commits/71669ff1b3e87a15c21768dfd3b313e089609db7

sapphi-red (approved):

hyfdev:
### Merge activity

* **Aug 26, 8:14 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Aug 26, 8:15 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Aug 26, 8:22 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10386, opened 2026-07-22
fix(dev): sourcemaps of lazily compiled chunks

### LINKED ISSUE #10378: [Bug]: `DevEngine.compileEntry` ignores input source maps, and its `sourceMappingURL` points at a file that is never emitted
## Reproduction link or steps

Both problems are about the source map of a chunk produced by `DevEngine.compileEntry` (lazy compilation, `experimental.devMode.lazy`). The eagerly bundled output of the same build is correct, which is what makes them look like bugs rather than missing features.

```
npm i rolldown@1.2.0 magic-string @jridgewell/trace-mapping
```

```js
// entry.js
globalThis.__lazy = () => import('./lazy.js')
import { eager } from './eager.js'
globalThis.__eager = eager
```

```js
// eager.js
export function eager() {
  throw new Error('from eager')
}
```

```js
// lazy.js
export function lazy() {
  throw new Error('from lazy')
}
```

```js
// run.mjs
import { dev } from 'rolldown/experimental'
import MagicString from 'magic-string'
import { originalPositionFor, TraceMap } from '@jridgewell/trace-mapping'

// shift both modules down by three lines and hand rolldown a correct source map
const shift = {
  name: 'shift',
  transform(code, id) {
    if (!/\/(eager|lazy)\.js$/.test(id)) {
      return
    }
    const s = new MagicString(code)
    s.prepend('\n\n\n')
    return { code: s.toString(), map: s.generateMap({ hires: 'boundary', source: id }) }
  },
}

const outputs = []
const engine = await dev(
  { input: { entry: './entry.js' }, plugins: [shift], experimental: { devMode: { lazy: true } } },
  { sourcemap: 'inline', entryFileNames: 'assets/[name].js', chunkFileNames: 'assets/[name].js', minify: false },
  { onOutput: r => !(r instanceof Error) && outputs.push(...r.output), watch: { skipWrite: true } },
)
await engine.run()
await engine.ensureCurrentBuildFinish()

const mapOf = (code) => {
  const m = /sourceMappingURL=data:application\/json[^,]+base64,([\w+/=]+)\s*$/.exec(code)
  return m ? JSON.parse(Buffer.from(m[1], 'base64').toString()) : null
}
const at = (code, needle) => {
  const lines = code.split('\n')
  const line = lines.findIndex(l => l.includes(needle)) + 1
  return { line, column: lines[line - 1].indexOf('throw') }
}

const chunk = outputs.find(o => o.type === 'chunk' && o.code.includes('from eager'))
console.log('EAGER ->', originalPositionFor(new TraceMap(mapOf(chunk.code)), at(chunk.code, 'from eager')))

await engine.registerClient('c1')
const lazy = await engine.compileEntry(new URL('./lazy.js', import.meta.url).pathname + '?rolldown-lazy=1', 'c1')
console.log('LAZY  ->', originalPositionFor(new TraceMap(mapOf(lazy.code)), at(lazy.code, 'from lazy')))

await engine.close()
```

```
node run.mjs
```

For the second problem, run the same script with `sourcemap: true` instead of `'inline'` and print `outputs.map(o => o.fileName)` and the last line of `lazy.code`.

## What is expected?

1. `compileEntry` composes the input source maps the plugins returned, the same way the eagerly bundled chunk does. Both `throw` statements are on line 2 of their source file, so both should map to line 2.
2. With `sourcemap: true`, the map referenced by the emitted `sourceMappingURL` is reachable — either emitted as an output file, or returned alongside the code.

**1. Input source maps are dropped**

```
EAGER -> { source: '../../eager.js', line: 2, column: 2 }   ✅
LAZY  -> { source: '../lazy.js',     line: 5, column: 2 }   ❌
```

Line 5 is where the `throw` sits in the plugin's *output* — the map of the lazily compiled chunk is relative to the transformed source rather than to the original file. The eager chunk of the same build, produced by the same plugin, composes correctly.

The practical effect is that any transform that changes line positions — TypeScript type stripping, JSX, a framework's SFC compiler — makes every stack frame and every devtools jump inside a lazily compiled module point at the wrong line.

**2. The `sourceMappingURL` points at a file that is never emitted**

With `sourcemap: true`:

```
outputs:        assets/entry.js, assets/lazy.js, assets/entry.js.map, assets/lazy.js.map
lazy filename:  lazy_compile_0.js
lazy code tail: //# sourceMappingURL=lazy_compile_0.js.map      <- not in outputs
```

There is no `lazy_compile_0.js.map` in the output, and `BindingLazyChunkOutput` is `{ code, filename }`, so the caller has no way to obtain it. A dev server serving `compileEntry` output verbatim therefore serves a dangling `sourceMappingURL`, and the browser cannot map the module at all. Setting `sourcemap: 'inline'` is a workaround, but it is not what a consumer configuring `sourcemap: true` would expect.

This is how it surfaces in Vite: `experimental.bundledDev` hard-codes `output.sourcemap = true`, so `/@vite/lazy` responses reference `lazy_compile_N.js.map`, which the dev server never serves.

## System Info

```Shell
  System:
    OS: macOS 26.4
    CPU: (16) arm64 Apple M3 Max
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.21.1
    npm: 10.9.4
    pnpm: 11.6.0
  npmPackages:
    rolldown: 1.2.0 (also reproduced on 1.1.5, the version Vite 8.1.5 pins)
```

## Any additional comments?

Found while building a dev server integration on top of `experimental.devMode`. Lazy compilation is what keeps module boundaries intact there — every cross-module read goes through `loadExports(id)` — so turning it off is not an option, which leaves the source map of `compileEntry` output as the only route back to the original sources.

Happy to split this into two issues if you would rather track them separately — they are filed together because both are about the map of the same output.

tbvjaos510:
Fixes #10378.

Two bugs, both about the sourcemap of a chunk produced by `DevEngine::compile_entry` (lazy compilation). The eagerly bundled output of the same build is correct in both cases, which is what makes them look like bugs rather than missing features.

## 1. Input sourcemaps were dropped

`render_module_code` returned the oxc codegen map as-is. That map is only the last element of a module's `sourcemap_chain`, so on its own it points at what the plugins produced rather than at the file the user wrote. `render_ecma_module` collapses the whole chain for a module rendered into a chunk; this path did not.

The practical effect: any transform that shifts positions — TypeScript type stripping, JSX, an SFC compiler — made every stack frame and every devtools jump inside a lazily compiled module land on the wrong line.

```
before                              after
EAGER -> { line: 2 }  ✅            EAGER -> { line: 2 }  ✅
LAZY  -> { line: 5 }  ❌            LAZY  -> { line: 2 }  ✅
```

`render_module_code` serves both the lazy-chunk and the HMR-patch path, so this fixes HMR patches as well.

The collapse is guarded on `enable_sourcemap`, matching `render_ecma_module`. Without the guard a render that asked for no sourcemap could still come back with one, since a non-empty chain collapses to a map even when the codegen step contributed none.

## 2. The `sourceMappingURL` named a map nobody could reach

`process_code_and_sourcemap` returns the sourcemap as an output asset, and `render_hmr_patch` carries it on `HmrPatch` so the consumer can serve it. The lazy-chunk path dropped that asset while still emitting a `sourceMappingURL` naming it — and a lazy chunk is never written to disk, so nothing could produce the file.

With `sourcemap: true` this left a dangling `sourceMappingURL=lazy_compile_N.js.map` and the browser could not map the module at all. `HmrLazyChunkOutput` now carries `sourcemap` / `sourcemap_filename` the way `HmrPatch` already does.

## Tests

Two tests in `packages/rolldown/tests/dev/dev-lazy-compile.test.ts`, one per bug. They drive `dev()` + `compileEntry` directly (no browser) and use a plugin that shifts lines and returns a sourcemap — the shape of a real type-stripping transform.

Both fail without this PR:

```
lazy chunk sourcemap maps through the plugin sourcemap chain
  → expected 2, received 5

lazy chunk returns the sourcemap its sourceMappingURL names
  → expected 'lazy_compile_0.js.map', received undefined
```

`cargo test --workspace` and the Node.js suite pass. No snapshot updates were needed.

I also verified the fix end to end by building the first commit on top of `v1.1.5` and dropping the binding into a dev-server integration that runs on Vite's `experimental.bundledDev`: a TypeScript test file whose failure was reported one line off now reports the exact line and column, with no other behavior change across its suites.

## Alternatives considered

- **Turning lazy compilation off.** Not an option for consumers that rely on it: lazy compilation is what keeps module boundaries intact, since every cross-module read goes through `loadExports(id)`.
- **`sourcemap: 'inline'` as a workaround for the second bug.** It does make the map reachable, but it is not what a consumer configuring `sourcemap: true` would expect, and Vite's bundled dev mode hard-codes `output.sourcemap = true`.

## Worth a closer look

- **`HmrLazyChunkOutput` and `BindingLazyChunkOutput` gain two fields.** Additive, and `binding.d.cts` is regenerated, but it is a binding-surface change.
- **The first commit changes HMR patch sourcemaps too**, not just lazy chunks, since both render through `render_module_code`. That is intentional — the same bug affected both. No repository snapshot changed, because the affected tests do not enable sourcemaps.
- **Collapse warnings are dropped** in `render_module_code` (`&mut Vec::new()`). The reasoning is that a module rendered here was rendered by the full build as well, which already reports a broken chain for it. Happy to thread them through if you would rather they surface.

### DIFF
--- a/crates/rolldown/src/hmr/hmr_stage.rs
+++ b/crates/rolldown/src/hmr/hmr_stage.rs
@@ -29,9 +29,15 @@ use rustc_hash::{FxHashMap, FxHashSet};
 use sugar_path::SugarPath;
 
 use crate::{
-  SharedOptions, SharedResolver, hmr::hmr_ast_finalizer::HmrAstFinalizer,
-  module_loader::ModuleLoader, type_alias::IndexEcmaAst, types::scan_stage_cache::ScanStageCache,
-  utils::process_code_and_sourcemap::process_code_and_sourcemap,
+  SharedOptions, SharedResolver,
+  hmr::hmr_ast_finalizer::HmrAstFinalizer,
+  module_loader::ModuleLoader,
+  type_alias::IndexEcmaAst,
+  types::scan_stage_cache::ScanStageCache,
+  utils::{
+    process_code_and_sourcemap::process_code_and_sourcemap,
+    render_ecma_module::collapse_module_sourcemap,
+  },
 };
 
 pub struct HmrStageInput<'a, Fs: FileSystem + Clone + 'static> {
@@ -810,10 +816,34 @@ impl<'a, Fs: FileSystem + Clone + 'static> HmrStage<'a, Fs> {
       },
     );
 
-    match codegen.map {
+    // The codegen map is the last element of the module's sourcemap chain, so on its
+    // own it maps back to what the plugins produced rather than to the original file.
+    // Collapsing the chain is what lets a position in a patch or a lazy chunk resolve
+    // to the source the user wrote, the same way `render_ecma_module` does for a
+    // module rendered into a chunk.
+    //
+    // Warnings are dropped: a module rendered here was rendered by the full build too,
+    // which reports a broken chain for it already.
+    let (code, codegen_map) = match codegen.map {
       Some(map) => (codegen.code, Some(map.into_owned())),
       None => (codegen.code, None),
-    }
+    };
+
+    // Guarded, because a chain collapses to a map even when the codegen step
+    // contributed none — which would hand back a sourcemap for a render that asked
+    // not to have one.
+    let map = enable_sourcemap
+      .then(|| {
+        collapse_module_sourcemap(
+          &module.sourcemap_chain,
+          codegen_map,
+          module.id.as_str(),
+          &mut Vec::new(),
+        )
+      })
+      .flatten();
+
+    (code, map)
   }
 }
 
--- a/crates/rolldown/src/hmr/hmr_stage.rs
+++ b/crates/rolldown/src/hmr/hmr_stage.rs
@@ -599,7 +599,7 @@ impl<'a, Fs: FileSystem + Clone + 'static> HmrStage<'a, Fs> {
 
     let file_dir = self.options.cwd.as_path().join(&self.options.out_dir);
 
-    if let Some(map) = map.as_mut() {
+    let sourcemap_asset = if let Some(map) = map.as_mut() {
       process_code_and_sourcemap(
         &self.options,
         &mut code,
@@ -610,8 +610,10 @@ impl<'a, Fs: FileSystem + Clone + 'static> HmrStage<'a, Fs> {
         /*is_css*/ false,
         None,
       )
-      .await?;
-    }
+      .await?
+    } else {
+      None
+    };
 
     let carried = modules_to_be_updated
       .iter()
@@ -621,7 +623,13 @@ impl<'a, Fs: FileSystem + Clone + 'static> HmrStage<'a, Fs> {
       })
       .collect();
 
-    Ok(HmrLazyChunkOutput { code, filename, carried })
+    Ok(HmrLazyChunkOutput {
+      code,
+      filename,
+      sourcemap_filename: sourcemap_asset.as_ref().map(|asset| asset.filename.to_string()),
+      sourcemap: sourcemap_asset.map(|asset| asset.source.try_into_string()).transpose()?,
+      carried,
+    })
   }
 
   async fn render_hmr_patch(
--- a/crates/rolldown/src/utils/render_ecma_module.rs
+++ b/crates/rolldown/src/utils/render_ecma_module.rs
@@ -65,7 +65,7 @@ pub fn render_ecma_module(
 ///
 /// A `SOURCEMAP_BROKEN` warning is pushed onto `warnings` for every plugin that
 /// omitted its sourcemap, since that breaks the mapping chain for `module_id`.
-fn collapse_module_sourcemap(
+pub fn collapse_module_sourcemap(
   sourcemap_chain: &[SourcemapChainElement],
   codegen_map: Option<SourceMap>,
   module_id: &str,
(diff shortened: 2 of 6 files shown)


h-a-n-a (approved):

h-a-n-a:
Thanks for the contribution.

Would you also create a follow-up commit to Vite's bundled-dev server so that it serves the generated sourcemap?

tbvjaos510:
Opened the Vite follow-up: vitejs/vite#23026. It's a draft until this lands in a rolldown release — it needs compileEntry to return sourcemap / sourcemapFilename, which isn't on npm yet. Any plan for a release (or a nightly I could pin) that includes it?

h-a-n-a:
> Opened the Vite follow-up: [vitejs/vite#23026](https://github.com/vitejs/vite/pull/23026). It's a draft until this lands in a rolldown release — it needs compileEntry to return sourcemap / sourcemapFilename, which isn't on npm yet. Any plan for a release (or a nightly I could pin) that includes it?

Thanks! We're releasing Rolldown next week

tbvjaos510:
@h-a-n-a opened the Vite follow-up: vitejs/vite#23026

PR #10391, opened 2026-07-22
fix: reject path-fragment emitted chunk names at emit time

### LINKED ISSUE #9994: Build error provides insufficient diagnostics (INVALID_OPTION in chunkFileNames)
I'm building my Qwik app, and the build fails with the following error:

```text
✓ 599 modules transformed.
✗ Build failed in 1.69s
error during build:
Build failed with 1 error:

[INVALID_OPTION] Invalid substitution "../node_modules/@builder.io/qwik-city/lib/index.qwik.mjs_entry_ErrorBoundary" for placeholder "[name]" in "chunkFileNames" pattern, can be neither absolute nor relative paths.

    at aggregateBindingErrorsIntoJsError (file:///home/dev/theme001/site/node_modules/rolldown/dist/shared/error-B68YLzl3.mjs:48:18)
    at unwrapBindingResult (file:///home/dev/theme001/site/node_modules/rolldown/dist/shared/error-B68YLzl3.mjs:18:128)
    at #build (file:///home/dev/theme001/site/node_modules/rolldown/dist/shared/rolldown-build-DR0wzp0V.mjs:3256:34)
    at async buildEnvironment (file:///home/dev/theme001/site/node_modules/vite/dist/node/chunks/node.js:32575:66)
    at async Object.build (file:///home/dev/theme001/site/node_modules/vite/dist/node/chunks/node.js:32997:19)
    at async Object.buildApp (file:///home/dev/theme001/site/node_modules/vite/dist/node/chunks/node.js:32994:153)
    at async CAC.<anonymous> (file:///home/dev/theme001/site/node_modules/vite/dist/node/cli.js:777:3) {
  errors: [Getter/Setter]
}

❌ Error: Client build failed: ExecaError: Command failed with exit code 1: npm run build.client
```

I'm stuck because the error doesn't indicate where the problem originates.

* Which file is causing this?
* Which line?
* Is this my configuration, a plugin, or an internal Qwik/Vite/Rolldown issue?
* Is there a way to enable more detailed diagnostics?

The current error message doesn't provide enough information to troubleshoot the issue.

IWANABETHATGUY:
## What

Reject a plugin-emitted chunk whose `name`/`fileName` is a path fragment (absolute or relative, e.g. `../node_modules/...`) synchronously at `emitFile` time, instead of accepting it and failing later during chunk rendering.

Fixes #9994.

## Why

The reporter's Qwik build failed with:

```
[INVALID_OPTION] Invalid substitution "../node_modules/@builder.io/qwik-city/lib/index.qwik.mjs_entry_ErrorBoundary" for placeholder "[name]" in "chunkFileNames" pattern, can be neither absolute nor relative paths.
```

The message is misleading on two counts:

- **It blames the `chunkFileNames` pattern, but the pattern is innocent.** The same error fires even with a static `chunkFileNames: 'chunk-[hash].js'` that has no `[name]` placeholder — `FilenameTemplate::render` validates the name substitution regardless of whether the template references `[name]`.
- **It names no plugin.** There's no way to tell which plugin emitted the bad name. The real culprit is a plugin in the Qwik / rolldown-vite pipeline that derives a chunk name from a module path made relative to the project root, for a dependency hoisted into a parent `node_modules` — so the name comes out as `../node_modules/...`.

Rollup rejects such a name synchronously inside `emitFile()`, so its failure carries plugin/hook attribution and points at the offending call site.

## What changed

`FileEmitter::emit_chunk` now validates `name`/`fileName` up front via a new `EmittedChunk::has_valid_name` (mirroring the existing `EmittedAsset::has_valid_name`) and the existing `InvalidEmittedFileName` diagnostic — whose message already reads *"emitted chunks and assets…"*, it just wasn't wired into the chunk path. The render-time check stays as a backstop for names not derived from a plugin emit.

**No plugin identity needs to be stored.** Because `emit_chunk` always runs inside a plugin hook, the existing `CausedPlugin` hook-error machinery attributes the error to the emitting plugin automatically; for JS plugins the surfaced error also carries the module id and a stack trace pointing at the plugin's own `emitFile` call.

### Before / after (JS plugin)

Before — late, misleading, no attribution:

```
[INVALID_OPTION] Invalid substitution "../node_modules/…_entry_ErrorBoundary" for placeholder "[name]" in "chunkFileNames" pattern …
```

After — fail-fast, correct message, attributed to the plugin:

```
[plugin emit-invalid-chunk-name] …/dep.js?chunk-url
Error: The "fileName" or "name" properties of emitted chunks and assets must be strings that are neither absolute nor relative paths, received "../node_modules/…_entry_ErrorBoundary".
    at …emitFile (…)
    at …plugin transform (…)
```

## Tests

- **`tests/fixtures/plugin/context/emit-chunk-invalid-name`** — `emitFile({ type: 'chunk' })` rejects path-fragment `name`/`fileName` (`./`, `../`, `/`, and Windows `C:/`) synchronously at emit time; valid names and subdirectory `fileName`s are still accepted. Mirrors the existing asset test `emit-file-invalid-name`.
- **`tests/error/error.test.ts`** — the surfaced diagnostic is attributed to the offending plugin (cross-platform inline snapshot; the machine/OS-specific stack trace is stripped).

### DIFF
--- a/crates/rolldown_common/src/file_emitter.rs
+++ b/crates/rolldown_common/src/file_emitter.rs
@@ -53,6 +53,20 @@ pub struct EmittedChunk {
   pub preserve_entry_signatures: Option<PreserveEntrySignatures>,
 }
 
+impl EmittedChunk {
+  /// Returns true if the emitted chunk has a valid name (not an absolute or relative path).
+  /// Mirrors [`EmittedAsset::has_valid_name`] and Rollup's `hasValidName`.
+  pub fn has_valid_name(&self) -> bool {
+    let validated_name = self.file_name.as_deref().or(self.name.as_deref());
+    validated_name.is_none_or(|name| !is_path_fragment(name))
+  }
+
+  /// Returns the validated name (fileName or name) if present.
+  pub fn validated_name(&self) -> Option<&str> {
+    self.file_name.as_deref().or(self.name.as_deref())
+  }
+}
+
 pub struct EmittedChunkInfo {
   pub reference_id: ArcStr,
   pub filename: ArcStr,
@@ -121,6 +135,18 @@ impl FileEmitter {
   }
 
   pub fn emit_chunk(&self, chunk: Arc<EmittedChunk>) -> anyhow::Result<ArcStr> {
+    // Reject a path-fragment name here, like the asset path in `emit_file` and Rollup's
+    // `emitFile()`. Deferring to `FilenameTemplate::render` loses the plugin context and
+    // blames the `chunkFileNames` pattern; erroring inside the hook lets the driver
+    // attribute it to the emitting plugin. https://github.com/rolldown/rolldown/issues/9994
+    if !chunk.has_valid_name() {
+      return Err(
+        BuildDiagnostic::invalid_option(InvalidOptionType::InvalidEmittedFileName(
+          chunk.validated_name().unwrap_or_default().to_string(),
+        ))
+        .into(),
+      );
+    }
     // Must stay synchronous: making this async would force the napi binding back
     // onto `block_on`, pinning the JS thread while `send().await` waits on channel
     // capacity — but the consumer draining the channel itself needs the JS thread
--- a/packages/rolldown/tests/error/error.test.ts
+++ b/packages/rolldown/tests/error/error.test.ts
@@ -204,6 +204,37 @@ describe('Error output format', () => {
   });
 });
 
+// https://github.com/rolldown/rolldown/issues/9994
+// Emitting a chunk with a path-fragment name is rejected at emit time and attributed
+// to the offending plugin, instead of failing late with a message that blamed the
+// `chunkFileNames` pattern and named no plugin.
+test('emitFile chunk with a path-like name is rejected and attributed to the plugin', async () => {
+  const error = await buildWithPlugin({
+    name: 'emit-invalid-chunk-name',
+    buildStart() {
+      this.emitFile({
+        type: 'chunk',
+        id: './main.js',
+        name: '../node_modules/@builder.io/qwik-city/lib/index.qwik.mjs_entry_ErrorBoundary',
+      });
+    },
+  });
+  // Keep only the plugin attribution + message; drop everything from the first stack
+  // frame onward. That is either the JS `    at …` frames (absolute paths / OS-specific
+  // separators) or the Rust `Stack backtrace:` block that `RUST_BACKTRACE=1` (set by the
+  // test scripts) inserts before them — both make the snapshot machine-dependent.
+  const rendered = removeAnsiColors(error!.message)
+    .replace(/\r\n/g, '\n')
+    .split(/\n\s*Stack backtrace:|\n\s+at /)[0]
+    .trimEnd();
+  expect(rendered).toMatchInlineSnapshot(`
+    "Build failed with 1 error:
+
+    [plugin emit-invalid-chunk-name]
+    Error: The "fileName" or "name" properties of emitted chunks and assets must be strings that are neither absolute nor relative paths, received "../node_modules/@builder.io/qwik-city/lib/index.qwik.mjs_entry_ErrorBoundary"."
+  `);
+});
+
 // oxlint-disable no-control-regex
 function removeAnsiColors(str: string) {
   return str.replace(/\x1b\[[0-9;]*m/g, '');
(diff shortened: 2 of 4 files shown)


Nic-Polumeyv:
The description here restates the diagnosis I posted on #9994 yesterday: the pattern being innocent, `FilenameTemplate::render` validating without `[name]`, rollup rejecting synchronously, the Qwik-side plugin deriving a relative name. #10377 has been open since before this one, implementing that fix. Neither is referenced.

It also fixes it at the wrong layer. Validating in `emit_chunk` surfaces `INVALID_OPTION`; rollup throws inside `emitFile()` with `code: 'PLUGIN_ERROR'` and `pluginCode: 'VALIDATION_ERROR'`, which is what `emit-file@invalid-asset-name` and `invalid-asset-name3` assert. Both stay ignored under this PR. #10377 produces rollup's shape, un-ignores both, and covers the asset path with the same mismatch on main.

Rebased and green. #10391 shouldn't land over it.

Nic-Polumeyv:
Why validate in `emit_chunk` rather than the JS plugin context?

From Rust the failure can only surface as `INVALID_OPTION`. Rollup throws inside `emitFile()` with `code: 'PLUGIN_ERROR'` and `pluginCode: 'VALIDATION_ERROR'`, and `emit-file@invalid-asset-name` and `invalid-asset-name3` assert that, so both stay ignored here. The asset path keeps the mismatch it has on main.

#10377 validates in the plugin context, produces rollup's shape, and un-ignores both.

IWANABETHATGUY:
Fair points — and you're right that I should've cross-linked #10377 and your #9994 write-up, my bad.

It's worth anchoring on what @Nefcanto actually asked for, though:

> I'm stuck because the error doesn't indicate where the problem originates. Which file is causing this? Which line? Is this my configuration, a plugin, or an internal Qwik/Vite/Rolldown issue? ... The current error message doesn't provide enough information to troubleshoot the issue.

The bar they set is "tell me which plugin — or whether it's rolldown itself." Validating in `emit_chunk` clears that: the error comes out as `[plugin <name>] <module id>` with a stack pointing straight at the offending `emitFile` call. And it does that for *every* emitter, not just JS ones — `plugin-context.ts` is only the JS entry point, but rolldown's own native plugins call `ctx.emit_chunk()` directly and never pass through it, so a check there answers "which plugin?" for user plugins while leaving the "…or is it internal?" half of their question back at the opaque late error. `emit_chunk` is the one choke point they all share, which is also why the asset path already validates there rather than in the binding — and reusing `is_path_fragment` avoids a second copy of rollup's `isPathFragment` in TS to keep in sync.

You're right about the shape: this surfaces `code: PLUGIN_ERROR` with the plugin attributed, but `pluginCode` is `GenericFailure`, not `VALIDATION_ERROR`, so it doesn't un-ignore `invalid-asset-name`/`name3`. Real gap — but it's one field. Tagging the native `InvalidEmittedFileName` diagnostic with `VALIDATION_ERROR` gets rollup's exact shape and un-ignores those tests without moving validation into the binding.

So I don't really see it as native-vs-JS — happy to fold `VALIDATION_ERROR` in here so we get rollup parity at the layer that covers every emitter, or bring that coverage into #10377. Whatever the maintainers prefer; not trying to land over your work.

IWANABETHATGUY:
Closing in favor of #10377 — after digging into both, it's the better base:

- It produces rollup's exact error shape (`code: PLUGIN_ERROR` / `pluginCode: VALIDATION_ERROR`) and un-ignores `invalid-asset-name` / `invalid-asset-name3`, which this PR can't do from the native layer — the napi boundary only ever surfaces a Status name (`GenericFailure`), never `VALIDATION_ERROR`.
- Its `isPathFragment` is a faithful, OS-independent port of rollup's, so it catches the Windows drive case (`F:\…`) on any host. This PR leaned on `Path::is_absolute()`, which misses it on Unix.
- It covers the asset path too, not just chunks.

The one idea from here worth keeping is the native `emit_chunk` backstop (so native/Rust emitters are covered, not just JS plugins), plus aligning `rolldown_common::is_path_fragment` with rollup's predicate so the render-time path agrees with the new JS check. I'll spin that up as a small complementary follow-up on top of #10377 rather than a competing change.

Thanks @Nic-Polumeyv — sorry again for the drive-by overlap.

Nic-Polumeyv:
No worries at all! Just making sure we get the best result for everyone.

Issue #10393, opened 2026-07-22
Build peak memory scales with module count; lazyBarrel is off-by-default and doesn't defer `export *` barrels


kHorozhanov:
### What problem does this address?

For a moderate module graph, rolldown's **peak build RSS is reached during the scan phase** — every module's AST (bumpalo arena) is allocated and held — and it is several times esbuild's for the same graph, widening as the graph grows. On real apps this OOMs builds that webpack/turbopack/esbuild handle: a ~2,500-module Next.js App Router build reaches ~9–10 GB and is OS-killed on 16 GB.

Minimal, dependency-free reproduction with a bundler baseline and phase-boundary peak-RSS traces: https://github.com/kHorozhanov/rolldown-build-memory (`npm i && npm run matrix`).

### 1. Per-module retention vs esbuild (identical reachable graph)

| N | esbuild | rolldown | ratio |
|--:|--:|--:|--:|
| 1,000 | 55 MB | 137 MB | 2.5× |
| 4,000 | 60 MB | 297 MB | 5.0× |
| 8,000 | 65 MB | 493 MB | 7.6× |

esbuild is ~flat; rolldown adds ~55 MB per 1,000 trivial modules and the gap widens. The phase trace shows `post-scan` is already the peak — every module's AST is live at once. This looks like the user-facing side of the AST-arena retention in #9516 / the footprint gap in #9330. Is any of the arena right-sizing / per-module release / streaming follow-up planned? The repro suggests it's `O(modules)` with a large constant, not just the #9516 fixture.

### 2. `experimental.lazyBarrel` — the one lever that lowers the scan peak

Deferring *loading* of a side-effect-free barrel's unused members is the only thing that reduces the peak (the ASTs are never allocated):

| N=8000, used=50, `sideEffects:false` barrel | post-scan peak |
|---|--:|
| **named** barrel, lazyBarrel off | 599 MB |
| **named** barrel, lazyBarrel on | **211 MB** |
| **star** (`export *`) barrel, lazyBarrel on (before) | 599 MB — no effect |
| **star** barrel, lazyBarrel on (**with #PR**) | **196 MB** |

Two gaps, one addressed here:
[...]

PR #10394, opened 2026-07-22
feat(lazy-barrel): defer `export *` star re-export targets

kHorozhanov:
## What this solves

`experimental.lazyBarrel` defers *loading* unused **named** re-exports of a side-effect-free barrel, but loaded every `export *` **star** target eagerly — because a requested name can't be attributed to a specific star target without loading it. For a barrel that `export *`s many modules (a library index doing `export * from './x'` for hundreds of files — e.g. `date-fns`, 245 star re-exports) and is imported for a few names, this parses and retains the AST of every target, which dominates peak build memory (ASTs are held for the whole build).

Part of #10393.

## Approach

Resolve star names on demand instead of eagerly: probe star targets one at a time in source order and stop once every requested name is found. Deferred targets are never loaded, so their ASTs are never allocated. `import * as ns` (an `All` request) still loads every target. Only active when `experimental.lazyBarrel` is on **and** the barrel is `sideEffects:false` (`UserDefined(false)`) — the assertion that makes skipping unused members safe.

- `BarrelInfo::take_needed_records` records missing names in a new `pending_star_names` set instead of pushing every star record as "needed".
- The loader gains `drive_star_probes` / `on_star_probe_target_loaded`: it loads star targets sequentially, subtracting the names each provides as it resolves, and stops when the set drains. Probe bookkeeping is transient per-scan state on `ModuleLoader`, not the cached `BarrelState`. Deferred star records keep `resolved_module = None`, which `NormalModule::star_export_records()` already skips during linking.

## Result

On an 8000-target star barrel imported for 50 names (`sideEffects:false`): peak RSS **~649 MB → ~196 MB**, **byte-identical output** (`cmp` clean). Matches the named-barrel deferral it mirrors.

## Tests

- New fixture `crates/rolldown/tests/rolldown/issues/lazy_barrel_star_reexport_deferral/` (star barrel, partial-name import; snapshot shows only resolved values survive and the unused target is never loaded).
- `cargo test --workspace --exclude rolldown_binding` green (the only failure, `test262::test262_module_code`, is the unchecked-out test262 submodule and fails identically on `main`). The existing lazyBarrel fixtures pass with **no snapshot changes**.
- clippy `--deny warnings` + rustfmt clean.

## Needs reviewer attention (opened as draft)

Strict `export *` **ambiguity** — a name exported by two star targets, which rolldown normally excludes-and-diagnoses — requires visiting all targets. Greedy stop-when-found resolves it to the first provider instead of erroring (verified: eager build errors, lazy build resolves greedily; it does not crash). This is gated behind `sideEffects:false` + opt-in `experimental.lazyBarrel`, and documented in a code comment. I'd like guidance on whether that trade-off is acceptable or whether ambiguity should force a full scan. A barrel that star-re-exports *another* barrel may also over-probe (extra loading only, never wrong/under-loading).

### DIFF
--- a/crates/rolldown/src/module_loader/module_loader.rs
+++ b/crates/rolldown/src/module_loader/module_loader.rs
@@ -125,6 +125,14 @@ pub struct ModuleLoader<'a, Fs: FileSystem + Clone + 'static> {
   /// of the first TLA keyword. Stored here instead of on every `EcmaView`
   /// because top-level await is rarely used in real-world apps.
   tla_keyword_span_map: FxHashMap<ModuleIdx, Span>,
+  /// On-demand `export *` deferral (transient per scan, not build-cache state):
+  /// maps a star-re-export *target* module to the barrel modules that
+  /// speculatively loaded it while probing for a requested name. When the target
+  /// finishes loading, the loader subtracts the names it provides from each
+  /// awaiting barrel's `pending_star_names` and probes the barrel's next star
+  /// target if names remain. See `BarrelInfo::take_needed_records` for why star
+  /// targets can't be attributed to a requested name up front.
+  star_probe_targets: FxHashMap<ModuleIdx, FxHashSet<ModuleIdx>>,
 }
 
 pub struct ModuleLoaderOutput {
@@ -218,6 +226,7 @@ impl<'a, Fs: FileSystem + Clone + 'static> ModuleLoader<'a, Fs> {
       magic_string_tx,
       tla_module_count: 0,
       tla_keyword_span_map: FxHashMap::default(),
+      star_probe_targets: FxHashMap::default(),
     })
   }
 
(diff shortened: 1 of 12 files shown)


kHorozhanov:
Closing this draft. On review the early-stop star probing here doesn't preserve `export *` ambiguity detection: when a requested name is found in the first star target, remaining targets aren't probed, so a name exported by two `export *` sources would resolve to the first match instead of being flagged ambiguous. That's the invariant the existing `lazy-barrel/star-export` fixture guards by loading every target, so this can't land as-is without either checking the remaining targets for conflicts (which gives back part of the deferral cost) or narrowing to the single-star-target case. The downstream use case that motivated it no longer needs it, so closing rather than reworking. The per-module retention findings in #10393 still stand.

PR #10398, opened 2026-07-23
fix(common): align is_path_fragment with Rollup's isPathFragment

IWANABETHATGUY:
Follow-up polish on #10377 (thanks @Nic-Polumeyv), which ported Rollup's `isPathFragment` to the JS `emitFile` validation. This brings the Rust `is_path_fragment` — used by the native render-time and emit-time checks — in line with it, so both layers agree on every input.

## Problem

`is_path_fragment` matched `./` and `../` by prefix and then fell back to `Path::new(name).is_absolute()`, which resolves differently depending on the **host OS** and is narrower than Rollup's predicate in both directions:

| name | Rollup | before (Unix host) | before (Windows host) | after |
| --- | --- | --- | --- | --- |
| `C:/windows`, `F:\test.ext` | reject | **accept** | reject | reject |
| `\server\share`, `\|foo` | reject | **accept** | **accept** | reject |
| `..y`, `..y.js` | reject | **accept** | **accept** | reject |
| `.x`, `.foo` | accept | accept | accept | accept |
| `C:relative` | accept | accept | accept | accept |

Two consequences:

1. **The same config could pass on Linux CI and fail on a Windows machine** (or vice versa) — a `chunkFileNames` of `F:\out\[name].js` was silently accepted on Unix.
2. **The native check disagreed with the JS check** added in #10377: `emitFile({ name: '..y' })` was rejected in JS, but the same string flowing through the render-time `[name]` substitution was accepted by Rust.

## Change

Port Rollup's predicate directly, matching on bytes so the result no longer depends on the host OS:

- `rollup/src/utils/relativeId.ts` — `isPathFragment`
- `rollup/src/utils/path.ts` — `ABSOLUTE_PATH_REGEX = /^(?:\/|(?:[A-Za-z]:)?[/\\|])/`

The doc comment on each copy now points at the Rollup source and at the other copy, so the two are maintained as a pair. The JS side in `packages/rolldown/src/utils/misc.ts` is unchanged in behavior — only the cross-reference comments were added.

Unit tests in `filename_template.rs` are extended with the Rollup-parity cases above, including the ones that only differ on one host (`F:\test.ext`, `\server\share`, `\|foo`) so they're now pinned on every platform.

## Behavior change

`..y` is now a path fragment, which is what Rollup does — verified against `rollup@4.62.3` (latest), which fails the identical build with `code: VALIDATION_ERROR`:

```
Invalid substitution "..y" for placeholder "[name]" in "output.entryFileNames" pattern, can be neither absolute nor relative path.
```

Two test updates fall out:

- **`issues/5557`** imported both `./.x.js` and `./..y.js`. The issue is about `.`-prefixed specifiers keeping their `./` prefix, which `./.x.js` covers on its own; the `..y.js` half is moved out.
- **`errors/preserve_modules_dot_dot_name`** (new) pins the `..y.js` case as an error under `preserveModules`, where the derived `[name]` is `..y`:

  ```
  [INVALID_OPTION] Invalid substitution "..y" for placeholder "[name]" in "entryFileNames" pattern, can be neither absolute nor relative paths.
  ```

## Known gap this does *not* close

Aligning the predicate aligns *detection*, not the whole `preserveModules` behavior. Comparing against Rollup on a matrix of names × module contents shows two remaining divergences worth flagging for reviewers:

- **Rolldown inlines modules Rollup preserves.** Under `preserveModules`, a module whose export is a const literal is inlined into its importer, so no `[name]` is ever derived for it and the check never runs — the build succeeds where Rollup errors. Only when the module survives inlining (a `function` export, or `optimization.inlineConst: false`) does rolldown reject it too. That's why the new fixture sets `inlineConst: false`. So whether rolldown accepts a `..y.js` module currently depends on its *content*, not just its name.
- **The error code and wording differ.** Both reject at the same point, but:

  | | Rollup 4.62.3 | rolldown |
  | --- | --- | --- |
  | `code` | `VALIDATION_ERROR` | `INVALID_OPTION` |
  | pattern named as | `output.entryFileNames` | `entryFileNames` |
  | tail | `…neither absolute nor relative path.` | `…neither absolute nor relative path**s**.` |

Both are pre-existing and out of scope here, but they're the reason this PR is framed as aligning the predicate rather than "matching Rollup's `preserveModules`".


IWANABETHATGUY:
* **#10400** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10400?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10399** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10399?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10398** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10398?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10398?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10398).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyfdev (approved):
The Rust `is_path_fragment` is a faithful, OS-independent port of Rollup’s `isPathFragment` / `ABSOLUTE_PATH_REGEX`, so it now agrees with the JS `emitFile` check from #10377 and no longer depends on host `Path::is_absolute()`. The unit cases pin the Windows-only and `..y` parity rows, and the preserveModules error fixture correctly keeps `inlineConst` off so the derived `[name]` is actually checked.

shulaoda (approved):

PR #10399, opened 2026-07-23
fix(plugin): validate emitted prebuilt-chunk file names

IWANABETHATGUY:
Stacked on #10398. Follow-up polish on #10377.

## Problem

#10377 added `isPathFragment` validation to `emitFile`, but the `prebuilt-chunk` branch returns before the check runs:

```ts
public emitFile: PluginContext['emitFile'] = (file): string => {
  if (file.type === 'prebuilt-chunk') {
    return this.context.emitPrebuiltChunk({ ... });   // <- returns early
  }
  // isPathFragment validation lives below this point
  ...
```

So this is accepted today:

```js
this.emitFile({
  type: 'prebuilt-chunk',
  fileName: '../prebuilt.js',
  code: 'export default 1;',
});
```

The bad name is not caught at emit time, so the plugin that emitted it is no longer on the stack when it eventually surfaces — as a write outside `outDir`, or as a later native error that doesn't name the plugin. That is the same diagnostic problem #10377 fixed for the other emit types (#9994), just on the branch it missed.

## Change

Run the same `isPathFragment` check inside the `prebuilt-chunk` branch, with a message specific to prebuilt chunks — they only carry `fileName`, no `name`, so reusing the shared message would be inaccurate:

> The "fileName" property of emitted prebuilt chunks must be strings that are neither absolute nor relative paths, received "../prebuilt.js".

The `typeof file.fileName === 'string'` guard keeps the existing behavior for a missing/non-string `fileName`, which is already reported downstream.

Covered by a new case in the existing `plugin/context/emit-file-invalid-name` fixture.

### DIFF
--- a/packages/rolldown/src/plugin/plugin-context.ts
+++ b/packages/rolldown/src/plugin/plugin-context.ts
@@ -371,6 +371,24 @@ export class PluginContextImpl extends MinimalPluginContextImpl {
 
   public emitFile: PluginContext['emitFile'] = (file): string => {
     if (file.type === 'prebuilt-chunk') {
+      // Rollup validates prebuilt chunks in `FileEmitter#emitPrebuiltChunk` — `code`
+      // first, then `fileName` — before either reaches the bundler. Without these the
+      // values fall through to N-API, which reports the napi `Status` (`InvalidArg`,
+      // `StringExpected`) as `pluginCode` instead of `VALIDATION_ERROR`.
+      if (typeof file.code !== 'string') {
+        return error(
+          logFailedValidation(
+            `Emitted prebuilt chunks need to have a valid string code, received "${file.code}".`,
+          ),
+        );
+      }
+      if (typeof file.fileName !== 'string' || isPathFragment(file.fileName)) {
+        return error(
+          logFailedValidation(
+            `The "fileName" property of emitted prebuilt chunks must be strings that are neither absolute nor relative paths, received "${file.fileName}".`,
+          ),
+        );
+      }
       return this.context.emitPrebuiltChunk({
         fileName: file.fileName,
         name: file.name,
--- a/packages/rolldown/tests/fixtures/plugin/context/emit-file-invalid-name/_config.ts
+++ b/packages/rolldown/tests/fixtures/plugin/context/emit-file-invalid-name/_config.ts
@@ -66,6 +66,41 @@ export default defineTest({
             'The "fileName" or "name" properties of emitted chunks and assets must be strings that are neither absolute nor relative paths, received "../node_modules/some-lib/entry".',
           );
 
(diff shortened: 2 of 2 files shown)


IWANABETHATGUY:
* **#10400** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10400?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10399** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10399?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10399?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10398** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10398?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10399).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

rollup
```diff
@@ -1 +1 @@
-Subproject commit a80a1974c584bfa8b694fb5d1a20f3fa75ebaf0a
+Subproject commit 8faa18777374582bb813d54ce3623f4acf1f9e0b
```
hyfdev: This moves the `rollup` submodule from `a80a197` (4.62.3, what main pins after #10494) back to `8faa187` (4.62.2). That is a downgrade, not an update, and it is what fails the required Node Test jobs on this head.

With 4.62.2, samples such as `plugin-parse` and `error-after-transform-should-throw-correct-location` still do `const MagicString = require('magic-string')`. The monorepo resolves `magic-string` v1 (ESM default export), so `new MagicString(...)` throws `TypeError: MagicString is not a constructor`. 4.62.3 already switched those fixtures to `await import('magic-string')`.

The commit message says verification was against rollup 4.62.3, which matches the parent pin; this gitlink change is not needed for the prebuilt-chunk validation or its fixtures. Keep `a80a197` / 4.62.3.

This blocks merge.

hyfdev (changes requested):
The prebuilt-chunk validation itself matches Rollup's `FileEmitter#emitPrebuiltChunk` (code first, then `fileName` / path-fragment, with the prebuilt-specific messages), and the fixture covers the cases this branch used to miss. That part looks ready.

The required Node Test jobs are red because this head moves the `rollup` submodule backward from 4.62.3 to 4.62.2. Restoring the pin on main is enough to unblock merge once CI re-runs green.

hyfdev (approved):

PR #10406, opened 2026-07-23
fix(rolldown): apply output.paths to CJS export-star of external modules

### LINKED ISSUE #10402: [Bug]: CJS output ignores output.paths for export * re-exports of external modules

Minimal repro with rolldown 1.1.5:

```js
// src/main.js
export * from './sub/mod.js';

// src/sub/mod.js
export const foo = 1;
```

```js
// build.mjs

    input: 'src/main.js',
    external: (id) => id.includes('sub/mod'),
});

await bundle.write({
    format: 'esm',
    file: 'dist/esm/main.mjs',
    paths: () => './sub.mjs',
});

await bundle.write({
    format: 'cjs',
    file: 'dist/cjs/main.cjs',
    paths: () => './sub.cjs',
});
```

### What is expected?

`output.paths` should be applied to the external module reference in **both** formats (matching Rollup behavior):

```js
// dist/esm/main.mjs
export * from "./sub.mjs";

// dist/cjs/main.cjs
var sub = require("./sub.cjs");
// ...
```

ESM output correctly applies `paths`, but the CJS output emits the **original source specifier** instead:

```js
// dist/esm/main.mjs — correct
export * from "./sub.mjs";

// dist/cjs/main.cjs — wrong: ignores output.paths
var src_sub_mod_js = require("./sub/mod.js");
Object.keys(src_sub_mod_js).forEach(function(k) { ... });
```

The `paths` function does receive the resolved absolute id (e.g. `/abs/path/to/src/sub/mod.js`) and its return value is used for the ESM output, but the CJS lowering of `export *` emits the raw import specifier `./sub/mod.js`.

This breaks multi-entry library builds that use `external` + `output.paths` to reference sibling chunks (e.g. subpath exports of a package referencing each other): CJS consumers end up with `require()` calls pointing at source files that do not exist in the published package.

### System Info

```
rolldown: 1.1.5
Node: v26.5.0
OS: Linux x64
```

### Any additional comments?

- Possibly a gap left by #6923, which applied `output.paths` to "dynamic imports, require calls, and re-export statements in generated code" — the CJS `export *` lowering path appears to have been missed.
- Also observed: `chunk.imports` in the `renderChunk` hook is `[]` for these pure `export *` facade chunks in both formats (unsure if intended).
- Current workaround: a `renderChunk` plugin that string-replaces `require("<original specifier>")` with the transformed path in `.cjs` chunks.
- Encountered in a real-world setup: Vite 8 (rolldown-vite) library-mode multi-entry build with per-entry `external` + `output.paths` mapping.

IWANABETHATGUY:
## Summary

Fixes #10402. `output.paths` was ignored by CJS output when re-exporting an external module via `export * from '<external>'`, even though ESM output applied it correctly.

The CJS/IIFE/UMD lowering of `export * from '<external>'` renders `var ns = require("<path>")` in `render_chunk_exports.rs`, but it called `external.get_import_path(chunk, None)` — passing `None` for the pre-resolved paths, so the `paths` map was never consulted and the **original source specifier** was emitted. This was the only `get_import_path`/`get_file_name` call site passing `None`; every other one (ESM export-star, CJS direct imports, `require()` calls, dynamic imports) already threads `ctx.resolved_paths`. It's the gap left by #6923.

## Fix

```rust
// crates/rolldown/src/utils/chunk/render_chunk_exports.rs
- external.get_import_path(chunk, None)
+ external.get_import_path(chunk, ctx.resolved_paths)
```

`ctx.resolved_paths` is the correct value: `output.paths` may be an async function, and the generate stage pre-resolves it into an `FxHashMap` before sync rendering (calling the raw `Fn` variant synchronously hits a `debug_assert!`). When no `paths` option is set, `resolved_paths` is `None`, so behavior is unchanged.

## Before / after

Input: `export * from 'ext-mod'` (external), `paths: { "ext-mod": "./mapped-ext.js" }`

| Format | Before | After |
|---|---|---|
| ESM | `export * from "./mapped-ext.js"` ✅ | `export * from "./mapped-ext.js"` ✅ |
| CJS | `require("ext-mod")` ❌ | `require("./mapped-ext.js")` ✅ |

## Rollup parity

Verified against **rollup 4.62.2** with the issue's exact reproduction — Rollup applies `output.paths` in both formats, which this change now matches:

```js
// ESM
export * from './sub.mjs';
// CJS
var mod_js = require('./sub.cjs');
```

## Tests

- New regression fixture `tests/rolldown/function/output_paths_export_star_external/` snapshotting both ESM and CJS variants.
- Existing `export_star_from_external` / `cjs_module_lexer` fixtures unchanged (no `paths` ⇒ identical output).

### DIFF
--- a/crates/rolldown/src/utils/chunk/render_chunk_exports.rs
+++ b/crates/rolldown/src/utils/chunk/render_chunk_exports.rs
@@ -276,7 +276,7 @@ pub fn render_chunk_exports(
           s.push('\n');
           // Only generate require statement if this external module hasn't been imported yet
           if imported_external_modules.insert(external.namespace_ref) {
-            writeln!(s, "var {} = require(\"{}\");", binding_ref_name, external.get_import_path(chunk, None)).unwrap();
+            writeln!(s, "var {} = require(\"{}\");", binding_ref_name, external.get_import_path(chunk, ctx.resolved_paths)).unwrap();
           }
           s.push_str(&import_stmt);
         });
--- a/crates/rolldown/tests/rolldown/function/output_paths_export_star_external/_config.json
+++ b/crates/rolldown/tests/rolldown/function/output_paths_export_star_external/_config.json
@@ -1,0 +1,11 @@
+{
+  "expectExecuted": false,
+  "config": {
+    "external": ["ext-mod"],
+    "paths": { "ext-mod": "./mapped-ext.js" }
+  },
+  "configVariants": [
+    { "format": "esm", "_configName": "esm" },
+    { "format": "cjs", "_configName": "cjs" }
+  ]
+}
--- a/crates/rolldown/tests/rolldown/function/output_paths_export_star_external/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/function/output_paths_export_star_external/artifacts.snap
@@ -1,0 +1,41 @@
+---
+source: crates/rolldown_testing/src/integration_test.rs
+---
+# Assets
+
+## main.js
+
+```js
+export * from "./mapped-ext.js";
+
+```
+
+# Variant: esm: [format: Esm]
+
+## Assets
+
+### main.js
+
+```js
+export * from "./mapped-ext.js";
+
+```
+
+# Variant: cjs: [format: Cjs]
+
+## Assets
+
+### main.js
+
+```js
+var ext_mod = require("./mapped-ext.js");
+Object.keys(ext_mod).forEach(function(k) {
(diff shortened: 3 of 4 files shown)


hyfdev (approved):
This completes the #6923 pattern: the CJS star-export require was the last external-reference render site not consulting the pre-resolved paths map. Since the generate stage pre-resolves `output.paths` over every external module before rendering, the function form is covered as well, and with `paths` unset `resolved_paths` stays `None`, so output is unchanged — consistent with no existing snapshots moving. I confirmed the new fixture fails without the source change: reverting only that line makes the CJS variant render `require("ext-mod")` and dirties the snapshot.

— Review by @hyfdev, assisted by Claude Fable 5

PR #10408, opened 2026-07-23
perf: speed up JS hook filter evaluation

IWANABETHATGUY:
## What this solves

The plugin hook filter interpreter (`filter_exprs_interpreter` in `rolldown_utils`) runs on the Rust side once per (module, hook, plugin-with-filter) — it's the cheap gate that decides whether to cross into a JS hook. This speeds up that gate.

I started from the question "should the tree-walk interpreter become a stack VM?" — but a local criterion benchmark showed the cost is in the **leaf tests**, not the interpretation dispatch (the trees are tiny and short-circuit already). A VM would optimize the ~1% that is dispatch, and a naïve postfix VM would even regress by losing short-circuiting. So this PR targets the actual cost instead.

## Reorder `And`/`Or` operands cheapest-first (`optimize_filter_expr`)

Compile-time (once, in `parse`) stable sort of `And`/`Or` children by a static cost rank (`ModuleType` < `Query` < `Id` < `Code`-string < `Code`-regex). Short-circuit evaluation then skips an expensive `Code` regex — which scans the **entire module source** — whenever a cheaper operand already decides the result.

- Pathological `and(code, id)`: **461 ns → 292 ns (−36%)** (8 ids/iter), landing on the ideal `and(id, code)` reference.
- Semantics-preserving: the boolean operators are pure over side-effect-free predicates (the only state, `parsed_url_cache`, is order-independent), and it only ever moves *cheaper* operands earlier — so it can only make short-circuiting skip an expensive leaf *more* often, never less.

## Why a VM was the wrong lever

- Dispatch is ~1% of the cost; the trees are tiny and stay hot in L1. A single `Code` regex over module source is 100–1000× a node dispatch.
- Short-circuiting is load-bearing: `And` stops on first false, `Or` on first true — that's what lets a cheap `Id` test skip an expensive `Code` regex. A naïve postfix stack VM evaluates every leaf and loses that (net slower); preserving it needs jump opcodes for no dispatch win that matters.

## Follow-up (not in this PR)

There's a second win — hoisting the per-call `Path::join(cwd, glob)` + normalize allocation in `get_matcher_string` for relative-glob `id` filters (measured ~1246 ns → ~546 ns). An earlier revision cached the cwd-resolved matcher on the leaf, but that changed `Id`/`ImporterId` matching semantics for relative globs (it baked in the first `cwd`). It's reverted here and will be stacked on top as a separate PR that caches the **compiled glob** rather than the cwd-resolved string.

## Tests

- `optimize_reorders_operands_cheapest_first` — verifies the reorder.
- `optimize_preserves_semantics` — asserts reordered vs. original agree across all input combinations.

No public signatures changed (`parse`, `filter_exprs_interpreter`, `FilterExprKind`), so `rolldown_binding` is untouched. Net diff is a single file, `crates/rolldown_utils/src/filter_expression.rs`.

### DIFF
--- a/crates/rolldown_utils/src/filter_expression.rs
+++ b/crates/rolldown_utils/src/filter_expression.rs
@@ -42,15 +42,22 @@ pub enum FilterExprKind {
 fn cost_rank(expr: &FilterExpr) -> u32 {
   match expr {
     FilterExpr::ModuleType(_) => 0,
-    FilterExpr::Query(..) => 1,
-    FilterExpr::Id(_) | FilterExpr::ImporterId(_) => 2,
+    FilterExpr::Id(_) | FilterExpr::ImporterId(_) => 1,
+    // The first `Query` eval scans/decodes the id's URL and collects its params into a
+    // map, so rank it *above* the cheap `Id`/`ModuleType` checks — they should get to
+    // short-circuit before that parse is forced.
+    FilterExpr::Query(..) => 2,
     // `memmem` over the full source, cheaper than a regex over it.
     FilterExpr::Code(StringOrRegex::String(_)) => 3,
     FilterExpr::Code(StringOrRegex::Regex(_)) => 4,
     // A wrapper costs whatever its (single) inner expression costs.
     FilterExpr::Not(inner) | FilterExpr::CleanUrl(inner) => cost_rank(inner),
-    // `And`/`Or` short-circuit, so the entry cost is the cheapest operand's.
-    FilterExpr::And(args) | FilterExpr::Or(args) => args.iter().map(cost_rank).min().unwrap_or(0),
+    // Rank a compound by its *costliest* reachable leaf, not its cheapest: an operand
+    // whose cheap leaf fails still falls through to its expensive one, so `min` would let
+    // e.g. `or(moduleType, code)` pose as free (rank 0) and jump ahead of a cheaper
+    // sibling that would have rejected first — forcing the very full-source scan this
+    // reordering exists to avoid.
+    FilterExpr::And(args) | FilterExpr::Or(args) => args.iter().map(cost_rank).max().unwrap_or(0),
   }
 }
 
@@ -310,7 +317,7 @@ pub fn parse(mut tokens: Vec<Token>) -> anyhow::Result<FilterExprKind> {
 #[cfg(test)]
 mod test {
   use crate::{
-    filter_expression::{FilterExpr, InterpreterCtx, Token, filter_expr_interpreter},
+    filter_expression::{FilterExpr, InterpreterCtx, QueryValue, Token, filter_expr_interpreter},
     pattern_filter::StringOrRegex,
   };
 
@@ -393,6 +400,50 @@ mod test {
       None,
       ".",
     ));
+  }
+
+  #[test]
+  fn optimize_keeps_query_after_id() {
+    // The first `Query` eval parses the id's URL and builds a param map, so a cheap
+    // `Id` check must run (and get to short-circuit) before that parse is forced —
+    // even though `Query` has fewer nodes, it is not cheaper to evaluate.
+    let mut expr = FilterExpr::And(vec![
+      FilterExpr::Id(StringOrRegex::Regex("target".into())),
+      FilterExpr::Query("raw".to_string(), QueryValue::Boolean(true)),
+    ]);
+    optimize_filter_expr(&mut expr);
+    match &expr {
+      FilterExpr::And(args) => {
+        assert!(matches!(args[0], FilterExpr::Id(_)), "Id must precede Query, got {args:?}");
+        assert!(matches!(args[1], FilterExpr::Query(..)));
+      }
+      other => panic!("expected And, got {other:?}"),
+    }
+  }
+
+  #[test]
+  fn optimize_ranks_compound_by_costliest_leaf() {
+    // A nested `or` containing a `Code` regex must not be ranked cheap by its
+    // `ModuleType` leaf: a non-matching module has to reject at the cheap `Id` check
+    // first, not fall into the full-source `Code` scan inside the `or`.
+    let mut expr = FilterExpr::And(vec![
+      FilterExpr::Id(StringOrRegex::Regex("target".into())),
+      FilterExpr::Or(vec![
+        FilterExpr::ModuleType("css".to_string()),
+        FilterExpr::Code(StringOrRegex::Regex("needle".into())),
+      ]),
+    ]);
+    optimize_filter_expr(&mut expr);
+    match &expr {
+      FilterExpr::And(args) => {
+        assert!(
+          matches!(args[0], FilterExpr::Id(_)),
+          "Id must precede the Code-bearing Or, got {args:?}"
+        );
+        assert!(matches!(args[1], FilterExpr::Or(_)));
+      }
+      other => panic!("expected And, got {other:?}"),
+    }
   }
 
   #[test]
--- a/crates/rolldown_utils/src/filter_expression.rs
+++ b/crates/rolldown_utils/src/filter_expression.rs
@@ -34,6 +34,52 @@ pub enum FilterExprKind {
   Exclude(FilterExpr),
 }
 
+/// Relative, static cost estimate for evaluating a [`FilterExpr`], used to order
+/// `And`/`Or` operands cheapest-first (see [`optimize_filter_expr`]). Lower is
+/// cheaper. These are ordinal ranks, not measured costs — only their relative
+/// order matters. The dominant distinction is that `Code` tests scan the entire
+/// module source, whereas `Id`/`ImporterId` only test the (short) path.
+fn cost_rank(expr: &FilterExpr) -> u32 {
+  match expr {
+    FilterExpr::ModuleType(_) => 0,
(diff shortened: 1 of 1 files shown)


crates/rolldown_utils/src/filter_expression.rs
```diff
@@ -34,6 +34,60 @@ pub enum FilterExprKind {
...
+/// [`InterpreterCtx::parsed_url_cache`], is keyed by the (fixed) id's query and so
+/// is order-independent. Reordering only ever moves *cheaper* operands earlier, so
+/// it can only make short-circuiting skip an expensive leaf more often, never less.
+fn optimize_filter_expr(expr: &mut FilterExpr) {
+  match expr {
+    FilterExpr::And(args) | FilterExpr::Or(args) => {
+      for arg in args.iter_mut() {
+        optimize_filter_expr(arg);
+      }
+      args.sort_by_cached_key(cost_rank);
```
hyfdev, line 87: `renderChunk` calls this interpreter with `id = None`, so sorting predicates whose inputs may be missing here can turn an authored short-circuit into a panic. With `filter: [include(and(code('__never__'), id(/x/)))]`, the base evaluates `code` first and returns `false`; this head moves `Id` first and hits ``id` should not be none``. The public `renderChunk` filter type accepts `TopLevelFilterExpression[]`, and binding validation only rejects `importerId`, so this changes a type-accepted configuration instead of merely optimizing it.

The reordering needs to preserve short-circuit behavior when an operand's input is unavailable, or validation needs to reject the expression before evaluation. This blocks merge.

crates/rolldown_utils/src/filter_expression.rs
```diff
@@ -34,6 +34,60 @@ pub enum FilterExprKind {
...
+/// order matters. The dominant distinction is that `Code` tests scan the entire
+/// module source, whereas `Id`/`ImporterId` only test the (short) path.
+fn cost_rank(expr: &FilterExpr) -> u32 {
+  match expr {
+    FilterExpr::ModuleType(_) => 0,
+    FilterExpr::Id(_) | FilterExpr::ImporterId(_) => 1,
+    // The first `Query` eval scans/decodes the id's URL and collects its params into a
+    // map, so rank it *above* the cheap `Id`/`ModuleType` checks — they should get to
+    // short-circuit before that parse is forced.
+    FilterExpr::Query(..) => 2,
```
hyfdev: `Query` is only more expensive than `Id` while the URL cache is empty. `InterpreterCtx` is shared across top-level rules, so an earlier rule can populate the cache. With `[exclude(query('warm', true)), include(and(query('missing', true), id(/target/)))]` on `/target.js?raw`, the base lets the cached `query` reject before evaluating `id`, while this rank moves the ID regex first.

In a release microbenchmark, this changed 54–57 ns/op to 61–68 ns/op, and the stacked head remains at 61–66 ns/op. Keeping `Id` and `Query` at the same rank is one way to preserve authored order while keeping both ahead of `Code`. This does not block the review by itself.

hyfdev (changes requested):
The operand reordering can evaluate a filter leaf whose hook input is unavailable even when the authored expression would have short-circuited before it. That turns a type-accepted `renderChunk` filter from a normal non-match into a panic.

crates/rolldown_utils/src/filter_expression.rs
```diff
@@ -244,8 +315,16 @@ pub fn parse(mut tokens: Vec<Token>) -> anyhow::Result<FilterExprKind> {
   }
 
   match tokens.pop() {
-    Some(Token::Include) => Ok(FilterExprKind::Include(rec(&mut tokens)?)),
-    Some(Token::Exclude) => Ok(FilterExprKind::Exclude(rec(&mut tokens)?)),
+    Some(Token::Include) => {
+      let mut expr = rec(&mut tokens)?;
+      optimize_filter_expr(&mut expr);
```
hyfdev, line 320: Deleting both `optimize_filter_expr` calls here leaves all ten `filter_expression` tests green, so nothing pins the reorder to the only path that reaches production — `pre_compile_filter_expr` → `parse`. The six `optimize_*` tests call `optimize_filter_expr` directly, and the `parse`-based tests assert booleans that the reorder is designed not to change.

Asserting on `parse`'s output shape would close that: for example, that `include(and(code(/x/), id(/y/)))` parses to an `And` whose first operand is the `Id`.

`missing_id_is_a_non_match_not_a_panic` has the same blind spot in the other direction. Without the reorder, `Code(/__never__/)` returns `false` and the `Id` leaf is never reached, so it passes on the merge base too — its `missing_code_` sibling is the one that fails there. Asserting `include(id(/target/))` with `id: None` directly would pin leaf totality on its own.

crates/rolldown_utils/src/filter_expression.rs
```diff
@@ -34,6 +34,73 @@ pub enum FilterExprKind {
...
+}
+
+/// Reorder `And`/`Or` operands cheapest-first (stably) so that short-circuit
+/// evaluation skips expensive leaves — most importantly a `Code` regex over the
+/// full module source — whenever a cheaper operand already decides the result.
+///
+/// This is a semantics-preserving transform: the boolean operators are pure over
+/// side-effect-free predicates. The only interpreter state, the lazily-populated
+/// [`InterpreterCtx::parsed_url_cache`], is keyed by the (fixed) id's query and so
+/// is order-independent. Reordering only ever moves *cheaper* operands earlier, so
```
hyfdev, line 73: The id is not fixed across an evaluation: the `CleanUrl` arm re-enters with `clean_url(id)` while passing the same `ctx`, so a `Query` leaf underneath it caches an empty param map that every later `Query` then reads. That makes `parsed_url_cache` order-dependent, and reordering decides which `Query` populates it.

`and(and(query('raw', true), code(/x/)), cleanUrl(query('raw', true)))` on `/t.js?raw` is `true` in authored order and `false` after `cost_rank` sorts the rank-1 `cleanUrl` operand ahead of the rank-2 inner `and`.

Nothing reachable through the public API builds that shape — `bindingify-hook-filter.ts` emits `CleanUrl` only directly around an `Id`/`ImporterId` leaf, and those are its only two producers — so this does not block this approval. Stating that as the actual precondition, rather than claiming the cache is order-independent, would keep the invariant honest for whoever generalizes `cleanUrl` later; clearing the cache when entering `CleanUrl` would make the claim true as written.

crates/rolldown_utils/src/filter_expression.rs
```diff
@@ -79,8 +148,10 @@ pub fn filter_expr_interpreter<'a>(
       ctx,
     ),
     FilterExpr::Query(key, value) => {
+      // When id is None (e.g. the `renderChunk` hook) there is no query to test
+      let Some(id) = id else { return false };
```
hyfdev, line 152: `query('raw', false)` and `not(query('raw', true))` both mean "the id does not carry `?raw`", and on `renderChunk` they now disagree: this early return makes the first `false`, so the hook never fires, while the second is `true` and it fires for every chunk. With an id present the two spellings always agree.

Both panicked before, so nothing regresses. Returning `*v == false` for the boolean form is one way to make them agree; a comment saying the "missing input is a non-match" rule deliberately wins over the boolean's natural reading would settle it just as well.

One compatibility datapoint for whichever way you settle it: `exprInterpreter` in `@rolldown/pluginutils` 1.0.1 still throws for a missing `id` or `code`, so the two interpreters of the same filter language now disagree where they previously both errored. `moduleType` already diverged that way, so this follows the existing Rust convention rather than breaking a new one.

hyfdev (approved):
The leaves are now total over the inputs each hook supplies, so the expression is a pure boolean function of its leaves and reordering `And`/`Or` operands cannot change its value. Restoring the three old `expect` calls makes both new `missing_*` tests fail, and every behavior change I could find is confined to configurations that panicked before — nothing that worked behaves differently. Ranking `Query` alongside `Id`/`ImporterId` also keeps the authored order when the URL cache may already be warm.

Three non-blocking notes inline: no test pins the reorder to `parse`, the order-independence claim in the `optimize_filter_expr` doc comment does not hold for a `Query` under a `CleanUrl`, and `query(k, false)` and `not(query(k, true))` disagree on `renderChunk`. The description is also stale — it still lists the cost rank as `ModuleType < Query < Id < Code-string < Code-regex`, and it does not mention the leaf-totality change, which is the most user-visible part of the PR.

The red checks on this head were cancelled before executing a single step, so they need a rerun rather than investigation.

IWANABETHATGUY:
Closing this one.

@hyfdev's review was right that the operand reordering can evaluate a leaf whose hook input is unavailable, and rather than patch around that I'm dropping the reordering optimisation for now.

The two pieces worth keeping have been split out onto `main`:

- **#10443** — the latent panic the review surfaced. A `renderChunk` filter using `id(...)`/`query(...)` hits `expect(```id``` should not be none)` and crashes; this exists on `main` today, independent of any reordering. `Code`/`Id`/`Query` now report "no match" for a missing input, matching `ImporterId`.
- **#10410** — the `Id`/`ImporterId` glob-matcher cache, now rebased directly onto `main` instead of stacking on this PR.

Thanks for the careful review — both the reordering hazard and the `Query`-vs-`Id` cost point were correct.

PR #10409, opened 2026-07-23
fix(code-splitting): initialize pure definers behind star re-export barrels under strict execution order

hyfdev:
### Problem

First P0 item of #10294 (Family A): under `strictExecutionOrder: true` — in both wrap modes — an inferred-pure ESM definer reached through an `export *` barrel by a split namespace consumer is never initialized. The definer's `init_*` ends up with zero call sites, its exported binding stays `undefined`, and a namespace reader observes it (NaN fold / TypeError). `strictExecutionOrder: false` is correct.

Minimal shape (now `strict_execution_order/pure_definer_star_barrel_split_readers`):

```js
// definer.js — inferred pure, non-inlinable value
export const vDef = /* @__PURE__ */ build();
// barrel.js
export * from './definer.js';          // star — essential
export { vSib } from './sibling.js';   // side-effectful sibling keeps the barrel wrapped
// entry a: import * as ns from './barrel.js';  ns.vDef   ← undefined
// entry b: import * as ns from './barrel.js';  ns.vSib   ← the split
```

### Root cause

The consumer side deliberately delegates: a non-transparent wrapped barrel is pushed as the only init target, trusting the barrel's own `init_*` to forward (the soundness comment on `wrapper_is_reachable_in_chunk` states exactly this contract). But the barrel's excluded `export *` statement only forwards with **retained-re-export evidence**, and the retention index (`star_reexport_records_by_imported_symbol`) was populated **only for named imports** (`Specifier::Literal` in `match_imports_with_exports`). A statically resolved namespace member read (`ns.vDef`) resolves through `resolve_member_expr_refs` and recorded nothing — so the hop had no evidence anywhere: statement excluded, no overlay, namespace not included. Nobody called `init_definer`.

Two adjacent holes fall out of the same evidence model:

- **Nested barrels:** `collect_order_wrap_esm_init_targets` stops a retained-path walk at the first init-owning barrel and delegates the remainder to that barrel's `init_*` — but `root_paths` was keyed only by the outermost record, so the interior owning barrel had no evidence for its own hop and the delegated remainder was silently dropped.
- **Opaque namespaces:** a whole-consumed namespace (`import * as ns` used opaquely) retains *every* export, including chains no named import or resolved read records; behind an owning barrel those chains had no evidence either (pre-existing on main, confirmed red under wrap-all).

### Fix (all strict-mode gated; flag-off output untouched)

1. **`resolve_member_expr_refs`** records the star re-export path of every statically resolved member step, keyed by the final canonical symbol — exactly the ref the inclusion pass marks used for the read, so `collect_frozen_reexport_usage`'s existing usedness gate applies unchanged (an inlined constant is never marked used and needs no init).
2. **`collect_frozen_reexport_usage`** segments each retained path at every init-owning barrel: the suffix becomes that barrel's own retained root, completing the delegation contract.
3. **`record_namespace_consumed_star_reexport_paths`** (new, link stage) records breadth paths for namespace-consumed modules, keyed by the importee's **namespace object** and gated on `namespace_included` — route-accurate, so a leaf used through a direct import elsewhere cannot retain a barrel path nobody consumes. Symmetrically, `transitive_esm_init_targets` no longer lets a recorded path restrict the walk below the record's own namespace evidence (`namespace_reexport_is_retained` ⇒ unrestricted walk) — without this, a path recorded for one resolved read would starve the definers only the opaque namespace observes.

The on-demand emergent-cycle projection consumes the same `collect_frozen_reexport_usage` / overlay population, so lowering and projection stay in lockstep by construction.

### Tests

Four new fixtures (each red without the fix, green with it, in both wrap modes — except `opaque_and_resolved_reads`, which is green on main and guards repair 3's gate against the path-restriction regression):

- `pure_definer_star_barrel_split_readers` — the minimal Family-A conjunction
- `pure_definer_nested_star_barrel_split_readers` — pure outer barrel, owning inner barrel (delegated remainder)
- `pure_definer_star_barrel_opaque_and_resolved_reads` — mixed resolved + opaque reads on one record
- `pure_definer_delegated_breadth_nested_barrels` — breadth demand crossing a delegation boundary (main is red under wrap-all)

### Validation

- Full integration suite: zero failure-list changes vs unmodified main; **zero existing `artifacts.snap` modified** (byte-identical outputs outside the new fixtures).
- Execution-order fuzzer (rolldown-order-fuzzer@77f196b) against this build: pure-ESM 100-case cell **24/100 RED → 0/100**, mixed cell **22/100 → 3/100**, identical in both wrap modes; the 3 residuals are the `sideEffects:false`-package event-reorder arm with signatures identical before and after this change (a distinct root, tracked separately in #10294). Pure-ESM seed `200004` and the shadcn transplant overlay pass in both wrap modes; the `strictExecutionOrder: false` control still passes.

Part of #10294.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/compute_wrapped_esm_init_metadata.rs
+++ b/crates/rolldown/src/stages/generate_stage/compute_wrapped_esm_init_metadata.rs
@@ -264,9 +264,18 @@ fn transitive_esm_init_targets(
       }
       let mut targets = vec![];
       if ctx.order_wrap {
-        let retained_reexport_path = overlay
-          .filter(|overlay| !overlay.retained_reexport_path.is_empty())
-          .map(|overlay| overlay.retained_reexport_path.as_slice());
+        // A recorded retained path restricts the hop walk to the chains resolved reads consumed.
+        // That is only sound when the path is the record's whole evidence: an included namespace
+        // (or forwarded dynamic exports) retains EVERY non-ambiguous export of this star record —
+        // including chains no resolved read recorded — so the walk must stay unrestricted there or
+        // the off-path pure definers lose their only init call site.
+        let retained_reexport_path = if namespace_reexport_is_retained {
+          None
+        } else {
+          overlay
+            .filter(|overlay| !overlay.retained_reexport_path.is_empty())
+            .map(|overlay| overlay.retained_reexport_path.as_slice())
+        };
         let mut retained_path_visited = FxHashSet::default();
         collect_order_wrap_esm_init_targets(
           ctx.modules,
--- a/crates/rolldown/src/stages/generate_stage/order_wrapping.rs
+++ b/crates/rolldown/src/stages/generate_stage/order_wrapping.rs
@@ -687,18 +687,50 @@ pub(super) fn collect_frozen_reexport_usage(input: &OrderLoweringInput<'_>) -> F
   let mut root_paths =
     FxHashMap::<(ModuleIdx, ImportRecordIdx), Vec<(ModuleIdx, ImportRecordIdx)>>::default();
   for (imported_as_ref, paths) in input.star_reexport_records_by_imported_symbol {
+    // A namespace-keyed path (recorded for a whole consumed namespace by
+    // `record_namespace_consumed_star_reexport_paths`, or a member read resolving to a
+    // namespace-valued binding) is consumed exactly when that namespace object is materialized:
+    // an included namespace retains every non-ambiguous export, so its star chains are
+    // execution-relevant; symbol-level usedness would conflate routes (a leaf used through a
+    // direct import elsewhere must not retain a barrel path nobody consumes).
+    let key_is_namespace = input.modules[imported_as_ref.owner]
+      .as_normal()
+      .is_some_and(|module| module.namespace_object_ref == *imported_as_ref);
     for path in paths {
       let Some(root) = path.first().copied() else {
         continue;
       };
-      let consumer_is_used = input.used_symbols.contains(imported_as_ref)
-        || consumed_facades.contains(imported_as_ref)
-        || input.linking[root.0]
-          .referenced_symbols_by_entry_point_chunk
-          .iter()
-          .any(|(symbol_ref, _)| symbol_ref == imported_as_ref);
+      let consumer_is_used = if key_is_namespace {
+        // `namespace_included` here is the provisional pre-wrap value: `finalize_chunk_plan` runs
+        // `finalized_module_namespace_ref_usage` before order analysis/lowering and re-runs it
+        // only after. The skew is safe — the post-wrap refinement can only ADD namespaces
+        // demanded by import overlays (`requires_namespace`: `export *` of a dynamic-exports
+        // importee, `require` interop, splitting-disabled dynamic import), and those routes
+        // discharge their breadth at runtime through `__reExport`/`__toCommonJS` glue rather
+        // than statically routed init forwarding. An opaque `import * as` consumer — the demand
+        // this gate exists for — is a link-time fact the provisional pass already observes.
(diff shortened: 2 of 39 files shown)


IWANABETHATGUY:
Nice work tracking this down — the evidence-model framing makes the three repairs easy to follow, and the fixtures are great. A few small things, nothing blocking:

**1. Doc comment on `record_namespace_consumed_star_reexport_paths`**

The comment says it records paths for "(`import * as ns` or a dynamic import)", but the scan only looks at `named_imports` with `Specifier::Star`, and a plain `import()` never creates one of those. Dynamic imports are actually handled by the existing recorder in `create_exports_for_ecma_modules.rs` (the dynamic-entry export path from #10104). Behavior is fine — I built the branch and ran a dynamic-import variant of the split-readers shape (pure definer behind `export *`, read only through the awaited namespace) and it's green in both wrap modes. But the comment would send someone debugging a dynamic-route issue to the wrong function, so I'd reword it to say which recorder owns which route.

**2. `key_is_namespace` in `collect_frozen_reexport_usage`**

Tiny one: it only depends on `imported_as_ref`, but it's recomputed inside the per-path loop. Could hoist it up one level to the key loop. Purely cosmetic.

**3. `namespace_included` is read at two different points in its lifecycle**

The new gate in `collect_frozen_reexport_usage` reads `namespace_included` as of the *first* `finalized_module_namespace_ref_usage` call (provisional topology, empty order state), while `namespace_reexport_is_retained` in `compute_wrapped_esm_init_metadata` reads it after the *post-wrap* refinement, where `order_state.requires_namespace` can flip additional namespaces on. For a namespace that flips on late, the record's own hop self-heals (namespace evidence at metadata time gives it the unrestricted walk), but the delegation suffixes from repair 2 would be missing, since the gate already rejected the path.

I couldn't construct a failing shape — the late flips come from overlay registration, which seems to presuppose the consumer evidence already existed — so I don't think anything is wrong today. But it took a while to convince myself, so a one-line comment on the gate saying which refinement pass it's allowed to observe (and why the skew is safe) would help the next reader.

IWANABETHATGUY (approved):

hyfdev:
Thanks — all three addressed in 6477b6c9a.

1. Reworded. This recorder only scans static `import * as` namespaces (keyed by the importee's namespace object); dynamic-import consumption is owned by the entry-export recorder in `create_exports_for_ecma_modules.rs`, which records per referenced entry export and is gated by the entry-chunk arm of `collect_frozen_reexport_usage`. The comment now names both recorders and their routes.
2. Hoisted to the key loop.
3. Added the note on the gate. It observes the provisional pre-wrap `namespace_included` (`finalize_chunk_plan` re-runs `finalized_module_namespace_ref_usage` only after lowering), and the skew is safe because the refinement can only add overlay-demanded namespaces (`requires_namespace`: star of a dynamic-exports importee, `require` interop, splitting-disabled dynamic import), whose breadth is discharged at runtime through `__reExport`/`__toCommonJS` glue rather than statically routed init forwarding — while an opaque `import * as` consumer is a link-time fact the provisional pass already sees. Same conclusion you reached; now it's written down next to the gate.

hyfdev:
### Merge activity

* **Jul 24, 3:35 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 24, 3:35 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 24, 3:42 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10410, opened 2026-07-23
perf: cache the resolved glob matcher for Id/ImporterId filters

IWANABETHATGUY:
> **Stacked on top of #10408** (base branch `perf/js-hook-filter-eval`). Review/merge that one first; the diff here is just the glob-matcher cache.

## What this solves

For a `string` (glob) `id`/`importerId` filter, `get_matcher_string` allocates on **every module tested** when the glob is *relative*: it does `Path::join(cwd, glob)` + normalize to anchor the glob to cwd. `IdMatcher` now memoizes the resolved matcher so that resolution happens **once** instead of once per module.

- Relative-glob `id` filter: **~1246 ns → ~562 ns** per 8 ids (measured locally; no bench committed).
- `**`-prefixed / absolute globs and regex ids: unchanged (they never allocated per call).

## The cwd-correctness design (the fix for the earlier revert)

The first attempt cached the cwd-resolved matcher on a plain `OnceLock` keyed on the **first cwd seen**, which baked in a "cwd constant per compiled filter" assumption and changed relative-glob matching if that filter was ever evaluated under a different cwd. That was reverted.

This version caches by **classifying the glob**, so nothing cwd-dependent is ever assumed:

```rust
enum ResolvedGlob {
  CwdIndependent(Box<str>),                       // `**` / absolute: valid for any cwd
  ScopedToCwd { cwd: Box<str>, matcher: Box<str> }, // relative: valid only for this cwd
}
```

- **`**` / absolute globs** resolve independently of cwd → cached once, reused for any cwd.
- **Relative globs** are cached together with the cwd they were resolved for. If a later call presents a different cwd, we **re-resolve from scratch** (never returning a stale, wrong-cwd matcher).

So matching is **behaviorally identical to resolving on every call** — the cache only removes the repeated allocation in the common case where cwd is constant for a build. It's a lock-free `OnceLock`, so it's safe under the concurrent (rayon) filter path, and the common `**`/absolute case is routed around the cwd comparison entirely (zero regression).

## Note on `fast_glob`

`fast_glob` (1.0.1) exposes only `glob_match(pattern, path)` — there's no compiled-matcher type to cache, so the only hoistable per-call cost is the string resolution above. (If we later switch to a crate with a compiled `GlobMatcher`, the glob *parsing* per call could also be hoisted — that's a separate change.)

## Tests

- `id_matcher_relative_glob_is_cwd_correct` — evaluates one `IdMatcher` under two different cwds and asserts each resolves correctly. This **fails** against the earlier baking cache and passes here.
- Existing `test_filter_expr_interpreter` / `parse_test` continue to exercise the `Id`/`ImporterId` path through `IdMatcher`.

### DIFF
--- a/crates/rolldown_utils/src/filter_expression.rs
+++ b/crates/rolldown_utils/src/filter_expression.rs
@@ -1,10 +1,14 @@
-use std::borrow::Cow;
+use std::{borrow::Cow, sync::OnceLock};
 
+use fast_glob::glob_match;
 use rustc_hash::FxHashMap;
 
 use crate::{
   js_regex::HybridRegex,
-  pattern_filter::{StringOrRegex, StringOrRegexMatchKind, normalize_path},
+  pattern_filter::{
+    StringOrRegex, StringOrRegexMatchKind, get_matcher_string, glob_matcher_depends_on_cwd,
+    normalize_path,
+  },
   url::{clean_url, get_query},
 };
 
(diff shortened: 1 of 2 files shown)


hyfdev (changes requested):
The cwd-scoped glob cache itself preserves matching across repeated and concurrent calls, but this layer inherits #10408's missing-input panic, so the stack cannot merge until the base layer preserves authored short-circuit behavior.

crates/rolldown_utils/src/filter_expression.rs
```diff
@@ -14,13 +18,78 @@ pub enum FilterExpr {
...
+  }
+}
+
+impl IdMatcher {
+  fn test(&self, value: &str, cwd: &str) -> bool {
+    match &self.pattern {
+      StringOrRegex::Regex(re) => re.matches(value),
+      StringOrRegex::String(glob) => {
+        match self.glob_cache.get_or_init(|| ResolvedGlob::resolve(glob, cwd)) {
+          ResolvedGlob::CwdIndependent(matcher) => glob_match(matcher.as_bytes(), value.as_bytes()),
```
hyfdev, line 70: Replacing this arm with a `panic!` leaves all 69 crate tests green, so nothing exercises the cwd-independent half of the cache. It is the only arm that reuses one entry under any `cwd`, and its correctness rests entirely on `get_matcher_string` and `glob_matcher_depends_on_cwd` agreeing about which globs read `cwd` — if that coupling ever loosens, this arm starts matching against the wrong directory while every committed test still passes.

Extending `id_matcher_relative_glob_is_cwd_correct` with a `**`-prefixed glob asserted under two different cwds would cover it.

Nothing observes that resolution now happens once per matcher rather than once per module either, so a later refactor could drop the win with the suite green — harder to pin, and lower stakes than the correctness half. Neither point blocks this approval.

hyfdev (approved):
The earlier request on this PR was about #10408, which is now approved; this layer is unchanged apart from the restack.

The cache is behaviorally identical to resolving the matcher on every call: a `**`-prefixed or absolute glob resolves without reading `cwd`, so one entry stays valid everywhere, and a relative glob is stored with the `cwd` it was resolved under and re-resolved whenever a later call presents a different one. Because `get_matcher_string` now dispatches on the same `glob_matcher_depends_on_cwd` the classifier uses, the two cannot drift apart.

Bypassing the `resolved_cwd == cwd` guard makes `id_matcher_relative_glob_is_cwd_correct` fail on its third assertion, so that test really does pin the design that was reverted.

I left one non-blocking note about the half of the cache no test reaches.

Docs Validation has no result on this head because the Deploy Website run never started, so it needs a rerun before the merge queue will take this.

crates/rolldown_utils/src/filter_expression.rs
IWANABETHATGUY: Good catch — reproduced it. `id_matcher_relative_glob_is_cwd_correct` uses a relative glob, so it always takes the `ScopedToCwd` path and never reaches the arm that reuses one entry across cwds.

Added a sibling test in `75a922792` using a `**`-prefixed glob asserted under two different cwds — kept it separate from the existing one so each test pins a single arm. With `CwdIndependent` replaced by a `panic!` it's now the only failing test, so the `get_matcher_string` / `glob_matcher_depends_on_cwd` agreement is actually pinned rather than just true by construction.

On the perf half: you're right that nothing observes it. Pinning it needs either a test-only counter or `OnceLock` introspection, which is more test surface than the regression it'd catch — happy to add it if you'd prefer, otherwise I'd leave that to the benchmark.

Also — the rebase onto `main` after #10443 landed dismissed your approval. Mind taking another look when you get a chance?

PR #10412, opened 2026-07-23
fix(resolveId): reject string `id` filters (must be a RegExp)

IWANABETHATGUY:
> **Bottom of the stack.** Base is `main`. The filter perf PRs (#10408 reorder, #10410 glob-cache) stack on top of this.

## What this solves

In `resolveId`, the `id` given to the filter is the **raw import specifier** (`./foo`, `lodash`, `virtual:x`), not a resolved path. But a `string` `id` filter is compiled to a **glob** — which is path/cwd-oriented (`get_matcher_string` even anchors relative globs onto cwd). Matching a specifier against a cwd-anchored glob is meaningless and silently never matches the way a user expects. Per the docs, resolveId `id` filters must be a `RegExp` (`hook-filters.md:80`).

The **object form** already enforces this at the type level — `HookFilterExtension<'resolveId'>` types `id` as `GeneralHookFilter<RegExp>` (`index.ts:774`). But the **composable form** (`filter: [include(id('glob'))]`, typed `TopLevelFilterExpression[]`) is not hook-restricted: `@rolldown/pluginutils`'s `id()` accepts `string | RegExp`, and `bindingifyResolveIdFilter` passed it straight through to the Rust glob matcher.

## The fix

Add `assertNoStringId` in `bindingifyResolveIdFilter`, mirroring the existing `assertNoImporterId` guard. It walks the filter expressions and throws a clear error if any `id` node uses a string, covering **both** forms (the composable form, and the object form defensively in case the type is bypassed):

```
A string `id` filter is not supported for the `resolveId` hook, because its
`id` is the import specifier rather than a resolved path. Use a RegExp instead.
```

`importerId` is intentionally **not** rejected — for resolveId it matches the *importer's* resolved id, which is a real path, so a glob is appropriate there.

## Tests

- `tests/fixtures/plugin/resolve-id/filter-string-id-error/` — a resolveId hook with `filter: [include(id('**/main.js'))]`, asserting the build throws an error mentioning `id`, `resolveId`, and `RegExp`. Mirrors the existing `filter-importer-id-error` fixture.

Typecheck passes. (The fixture runs in CI; it wasn't run locally because no napi binding is built in this worktree.)

### DIFF
--- a/packages/rolldown/src/plugin/bindingify-hook-filter.ts
+++ b/packages/rolldown/src/plugin/bindingify-hook-filter.ts
@@ -138,6 +138,33 @@ function assertNoImporterId(
   }
 }
 
+function containsStringId(expr: FilterExpression | TopLevelFilterExpression): boolean {
+  switch (expr.kind) {
+    case 'and':
+    case 'or':
+      return expr.args.some(containsStringId);
+    case 'not':
+    case 'include':
+    case 'exclude':
+      return containsStringId(expr.expr);
+    case 'id':
+      return typeof expr.pattern === 'string';
+    default:
+      return false;
+  }
+}
+
+function assertNoStringId(
+  filterExprs: filter.TopLevelFilterExpression[] | undefined,
+  hookName: string,
+): void {
+  if (filterExprs?.some(containsStringId)) {
+    throw new Error(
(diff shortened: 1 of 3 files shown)


hyfdev:
LGTM @sapphi-red cc

sapphi-red (approved):

shulaoda (approved):

PR #10414, opened 2026-07-23
fix(code-splitting): fold the runtime chunk back after order lowering

hyfdev:
### Problem

First P1 item of #10294: `ensure_runtime_module_for_order_wraps` unconditionally evicts a co-hosted runtime module onto a standalone `rolldown-runtime` chunk (and mints one when the runtime was unplaced), even when every helper consumer lives in a single chunk. `issues/9463` ships a `rolldown-runtime.js` whose only content is two helper re-exports consumed exclusively by `common~a~b~shared.js` — one extra request per page load for nothing, in both wrap modes.

The eviction itself is justified: the baseline merge proof (`try_merge_runtime_chunk`, called from `generate_chunks` and the chunk optimizer) runs **before** order analysis, so nothing it proved covers the helper demand order lowering adds (`__esm`/`__esmMin` per wrapper, `__toCommonJS`/`__reExport` per lowered overlay). What was missing is re-running the proof afterwards: by the end of lowering the full demand is materialized in `OrderWrapState`, so the complete post-lowering consumer set is knowable.

### Fix

`ensure_runtime_module_for_order_wraps` keeps its normalize-to-standalone behavior unchanged, then re-runs the merge proof with the post-lowering consumer set:

- New `OrderWrapState::runtime_helper_consumer_chunks` collects the order-introduced consumers: each wrapper's synthetic `init_*` statement renders in its assigned chunk, and each overlay's lowered import/require glue renders in the importer's chunk. Pre-lowering demand is re-collected by `collect_runtime_consumer_chunks`'s own scan (chunk helper bits, interop wrappers, statement-level runtime symbol references), and stale/tombstoned chunks are filtered by the existing `post_chunk_optimization_operations` guard.
- `try_merge_runtime_chunk` gains a `RuntimeMergeCascade` parameter. The two pre-existing call sites use `Full` (byte-identical cascade). The post-lowering fold uses `SingleConsumerOnly`, and two deliberate restrictions apply:
  - **Sole-consumer hosts only.** A dominator merge turns transitive reachability into a direct import whose exec-order sort position can hoist the host past sibling imports, and a bitset host can gain a brand-new consumer edge — either way the chunk-evaluation order deviates from what the on-demand analysis proved against the pre-fold edge set. A sole-consumer merge only removes that consumer's edge to an inert runtime-only chunk, so evaluation order is preserved by construction (this also makes the TLA/signature/cycle guards trivially satisfied, and keeps the existing sole-consumer allowance for manual group chunks).
  - **Esm output only.** Under cjs output, `compute_cross_chunk_links` later gives every ESM-exports entry chunk — including the zero-module order facades — a `__toCommonJS` demand that is invisible at fold time; folding would hand an entry chunk with no visible demand a brand-new `require` edge into a user chunk. Every other post-fold demand source was audited to be either Esm-excluded or already represented in the fold-time consumer set.
- In `SingleConsumerOnly` mode the merge tail skips `target_chunk.bits.union(...)`: the sole consumer's own bits are already exact (the runtime becomes internal to it), while the order-time standalone chunk carries synthetic all-live-union bits whose widening would leak into bits-derived chunk names.

Non-esm formats and `strictExecutionOrder: false` builds are byte-for-byte untouched; the fold only runs from the order-wrap path (non-empty plan).

### Output shape

`issues/9463` (acceptance): `rolldown-runtime.js` disappears and the helpers inline into `common~a~b~shared.js` in both cells; same for `issues/9463_plain_group`, `issues/9463_three_entries`, and `strict_execution_order/concatenate_basic` (wrap-all cells fold into `main2.js`). 40 snapshots change, all strict-mode fixtures, all the same shape: the standalone runtime chunk is gone and its (actually demanded) helpers render inside the sole consumer — `issues/5870` and `mixed_static_dynamic_same_target` also drop a previously exported-but-unconsumed `__exportAll`, since the host renders only demanded statements.

One snapshot-cosmetic note: in fixtures where the runtime region folds ahead of a synthetic wrapper rendered without its own region markers (pre-existing quirk), the `HIDDEN [\0rolldown/runtime.js]` replacement visually swallows the following wrapper in the snapshot (`top_level_await_syntax`). The raw emitted chunk was dumped and verified correct — `init_foo` is present; only the snapshot display collapses it.

### Validation

- Full integration suite + workspace tests: failure list byte-identical to unmodified main (verified by stash-and-rerun on the same machine; the only local failures are the pre-existing css/legal-comments/esbuild-diff environment set).
- Execution-order fuzzer (rolldown-order-fuzzer) against this build, 300 cases across three cells: mixed 112/120, pure-ESM 83/100, wrap-all 71/80. The same three campaigns (same seeds and flags) were then run against an unmodified-main build of this worktree: the failing seed sets are identical and every one of the 34 failures carries a byte-identical verdict signature (the known Family-A / event-reorder arms tracked in #10294) — zero failure signatures introduced or altered by this change.

Part of #10294.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/order_wrapping.rs
+++ b/crates/rolldown/src/stages/generate_stage/order_wrapping.rs
@@ -7,8 +7,8 @@ use itertools::Itertools;
 use oxc::ast::ast::{Declaration, ExportDefaultDeclarationKind, Statement};
 use rolldown_common::{
   Chunk, ChunkIdx, ChunkKind, ChunkMeta, ImportKind, ImportRecordIdx, ImportRecordMeta,
-  IndexModules, ModuleIdx, PostChunkOptimizationOperation, RuntimeHelper, StmtInfoIdx, SymbolRef,
-  SymbolRefDb, UsedSymbolRefsBuilder, WrapKind,
+  IndexModules, ModuleIdx, OutputFormat, PostChunkOptimizationOperation, RuntimeHelper,
+  StmtInfoIdx, SymbolRef, SymbolRefDb, UsedSymbolRefsBuilder, WrapKind,
 };
 use rolldown_ecmascript::EcmaAst;
 use rolldown_ecmascript_utils::StatementExt;
@@ -17,6 +17,7 @@ use rustc_hash::{FxHashMap, FxHashSet};
 use super::{
   GenerateStage,
   chunk_ext::{ChunkCreationReason, ChunkDebugExt},
+  chunk_optimizer::RuntimeMergeCascade,
   order_analysis::{OrderAnalysis, OrderWrapPlan},
   order_wrap_state::{OrderImportKey, OrderImportOverlay, OrderWrapState},
 };
@@ -146,7 +147,7 @@ impl GenerateStage<'_> {
     self.place_order_wrap_modules(chunk_graph, plan, order_state);
     self.create_order_wrap_entry_facades(chunk_graph, analysis);
     self.restore_order_wrap_entry_facades(chunk_graph, plan);
-    self.ensure_runtime_module_for_order_wraps(chunk_graph);
+    self.ensure_runtime_module_for_order_wraps(chunk_graph, order_state);
     chunk_graph.sort_chunk_modules(self.link_output, self.options);
     self.renumber_live_chunks(chunk_graph);
     true
@@ -392,7 +393,19 @@ impl GenerateStage<'_> {
     }
   }
 
-  fn ensure_runtime_module_for_order_wraps(&mut self, chunk_graph: &mut ChunkGraph) {
+  /// Normalize the runtime module onto a standalone chunk, then re-run the runtime-chunk merge
+  /// proof with the post-lowering consumer set.
+  ///
+  /// The baseline `try_merge_runtime_chunk` calls run before order analysis, so a pre-lowering
+  /// merge never proved anything about the helper demand the wrappers and overlays above added.
+  /// Evicting a co-hosted runtime first restores the standalone shape that proof requires; by this
+  /// point lowering has materialized every order-introduced demand in [`OrderWrapState`], so the
+  /// re-proof sees the complete consumer set.
+  fn ensure_runtime_module_for_order_wraps(
+    &mut self,
+    chunk_graph: &mut ChunkGraph,
+    order_state: &OrderWrapState,
+  ) {
     let runtime_idx = self.link_output.runtime.id();
     if let Some(runtime_chunk_idx) = chunk_graph.module_to_chunk[runtime_idx] {
       if self.options.code_splitting.is_disabled() {
@@ -401,6 +414,9 @@ impl GenerateStage<'_> {
       let runtime_chunk = &chunk_graph.chunk_table[runtime_chunk_idx];
       if runtime_chunk.modules.len() == 1 {
         self.clear_module_symbol_chunk_indices(runtime_idx);
+        // Facade restoration above can tombstone chunks the pre-lowering merge counted as
+        // consumers, so a standalone runtime may only now have a sole consumer left.
+        self.fold_runtime_chunk_after_order_lowering(chunk_graph, order_state);
         return;
       }
       let mut bits = runtime_chunk.bits.clone();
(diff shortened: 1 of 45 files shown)


crates/rolldown/src/stages/generate_stage/order_wrapping.rs
```diff
@@ -483,6 +500,32 @@ impl GenerateStage<'_> {
       self.link_output.metas[runtime_idx].depended_runtime_helper,
     );
     self.clear_module_symbol_chunk_indices(runtime_idx);
+    self.fold_runtime_chunk_after_order_lowering(chunk_graph, order_state);
+  }
+
+  /// Re-run the runtime-chunk merge proof against the post-lowering consumer set: the
```
hyfdev, line 506: This adds a second, ESM-only runtime merge after order lowering, but `internal-docs/code-splitting/implementation.md` still describes `try_merge_runtime_chunk` only in `split_chunks`, lists only the pre-lowering/facade-elimination consumer sources, and documents the full bitset/dominator cascade as the candidate policy.

That file is the current implementation contract, and `AGENTS.md` requires affected internal docs to be updated in the same change. It needs to record the `OrderWrapState` consumer source, the sole-consumer/ESM restrictions, and why this mode does not union the runtime bits. This blocks merge.

crates/rolldown/tests/rolldown/function/experimental/strict_execution_order/top_level_await_syntax/artifacts.snap
```diff
@@ -30,13 +30,7 @@ await init_main();
...
 
 ```js
-import { t as __esmMin } from "./rolldown-runtime.js";
-function init_foo() {
-	return (init_foo = __esmMin((async () => {
-		await 1e3;
-	})))();
-}
-//#endregion
+// HIDDEN [\0rolldown/runtime.js]
```
hyfdev, line 33: The snapshot scrubber now consumes `init_foo` together with the runtime region, so this fixture only shows `init_main` calling `init_foo`; it no longer proves that the TLA wrapper was emitted. Because `expectExecuted` is false, deleting or breaking that declaration would leave this snapshot unchanged and would not fail execution.

The raw chunk on this head contains `init_foo` and executes correctly, so this is a coverage gap rather than a correctness failure. Exposing the runtime region for this fixture or adding a structural/execution assertion would preserve that coverage. This does not independently block merge.

crates/rolldown/src/stages/generate_stage/order_wrapping.rs
hyfdev: Documented in 58e262bc7. The pipeline diagram now shows `ensure_runtime_module_for_order_wraps` inside `apply_order_wraps()`; the consumer-set formula's last bullet is generalized to the caller-supplied channel (facade elimination during chunk optimization, order-introduced consumers during the fold); the cascade paragraph records the caller-selected `RuntimeMergeCascade` depth; and a new "Post-Order-Lowering Runtime Fold" section records the `OrderWrapState` consumer source, the sole-consumer/ESM-only restrictions with their rationale, why this mode does not union the runtime bits, and the regression fixtures.

crates/rolldown/tests/rolldown/function/experimental/strict_execution_order/top_level_await_syntax/artifacts.snap
hyfdev: This anchors to the pre-rebase head (8548a60), where the dce pass ate the region markers and the scrubber swallowed `init_foo` — the bug fixed by #10431. After rebasing onto main with #10431, the current snapshot shows the full wrapper under a visible `//#region foo.js`, so deleting or breaking `init_foo` now changes the snapshot. No extra assertion added.

IWANABETHATGUY (approved):

hyfdev:
### Merge activity

* **Jul 26, 3:39 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 26, 3:39 PM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 26, 3:45 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10416, opened 2026-07-23
feat(watch): allow function filters for watch include/exclude

### LINKED ISSUE #5799: [Feature Request]: allow function for including/excluding files for `watch`

Right now, including and excluding files from watching is only allowed via string or regex (array).

https://github.com/rolldown/rolldown/blob/5b89c5d26a4ca1451d2c6d4be55f882338b46797/crates/rolldown_common/src/inner_bundler_options/types/watch_option.rs#L25-L33

Right now, Vite & Rollup rely on [chokidar](https://rollupjs.org/configuration-options/#watch) which allows an [ignored](https://github.com/paulmillr/chokidar#path-filtering) function to ignore files from watching. To align, would it be possible to pass a function in as well (and would that have severe perf implications)?

Allowing a function (or multiple) in addition to string & regexp for include and exclude.

unknownjedi:
## Summary
This PR adds function-based file filters for watch mode so `watch.include` and `watch.exclude` can accept callbacks in addition to string/regex patterns.

This aligns Rolldown with the common Rollup/Vite watch filtering style and allows consumers to express dynamic filtering rules that static patterns cannot cover.

Closes #5799.

## What changed
1. **Watch filter API extended**
- `watch.include` and `watch.exclude` now accept:
  - string/regex patterns (existing behavior)
  - function filters `(id: string) => boolean`
  - mixed arrays of patterns and functions

2. **End-to-end option plumbing**
- Threaded function filters through:
  - TypeScript options/types
  - input option validator
  - NAPI binding layer
  - Rust watch options

3. **Watcher filtering logic updated**
- File registration now evaluates both pattern filters and function filters.
- Exclude-first semantics are preserved:
  - exclude pattern / exclude function checked first
  - include pattern / include function checked after
- Default include behavior is preserved (when no include filter matches, watch behavior stays consistent with existing semantics).

4. **Error behavior hardened**
- If a watch include/exclude callback throws, watcher now emits a proper `ERROR` event instead of silently failing the build path.

5. **Coverage updates**
- Added tests for:
  - function filters
  - mixed filter arrays
  - thrown callback regression path (ensuring `ERROR` is surfaced)

## Why this change
Before this PR, watch include/exclude only supported static patterns. That made it hard for integrators to implement context-dependent filtering rules.

This change enables dynamic filtering while keeping existing pattern behavior intact and preserving watch-mode reliability by surfacing callback errors explicitly.

## Compatibility and impact
- Existing string/regex watch filters continue to work unchanged.
- New callback filters are additive.
- No core watch architecture changes were introduced; this is a scoped feature + error-path fix.

## AI disclosure
This PR was developed with AI assistance. All generated changes were reviewed, tested, and adjusted by the author before submission.


unknownjedi:
Hi @shulaoda, just a friendly ping on this PR — it's been open about a month. Whenever you have time, would appreciate a review. Thanks!

shulaoda:
> Hi @shulaoda, just a friendly ping on this PR — it's been open about a month. Whenever you have time, would appreciate a review. Thanks!

Thanks for the PR, and sorry for the delay.

I'm currently focused on watch-mode work. I still need to think through whether `watch.include` / `watch.exclude` should accept functions.

PR #10427, opened 2026-07-24
fix(treeshake): preserve eagerly evaluated child effects of manual pure chains

### LINKED ISSUE #10397: [Bug]: `manualPureFunctions`: a pure call inside a member/`new` chain drops eagerly-evaluated child effects (arguments, computed keys)
With `treeshake.manualPureFunctions`, a call to a listed function that sits inside a member or `new` chain is deleted together with subexpressions that JavaScript evaluates eagerly — call arguments, computed property keys, and the `new` callee chain. Their side effects are silently dropped. A bare pure call (`make(effect())`) is handled correctly; the bug is specific to chains.

### Repro

rolldown 1.2.0 (latest release); the responsible short-circuit is unchanged on current `main` (4930a00).

```js
// main.mjs
import { make } from 'ext';

function effect() {
  console.log('effect');
  return 'k';
}

make()[effect()];
make(effect()).value;
new (make()[effect()].Box)();
```

```js
// rolldown.config.mjs
export default {
  input: 'main.mjs',
  external: ['ext'],
  treeshake: { manualPureFunctions: ['make'] },
  output: { file: 'out.mjs', format: 'esm' },
};
```

Actual output — the whole module body is gone, including the three `effect()` calls:

```js
import "ext";
```

Expected: deleting the pure calls themselves is fine, but the eagerly-evaluated `effect()` subexpressions must be kept. Each isolated statement reproduces on its own.

Controls:

- Without `manualPureFunctions`, all three statements are retained.
- A bare call is already handled correctly: `make(effect());` compiles to `effect();`.

### Comparison with Rollup

Rollup 4.62.2 with the identical input and config retains all three statements. Rollup's documentation promises this in the [`manualPureFunctions` warning](https://rollupjs.org/configuration-options/#treeshake-manualpurefunctions):

> If you pass arguments to such a pure function, those arguments are still checked for direct side effects like mutating a variable or calling a global function, in which case the call to the pure function is retained.

### Cause

- `analyze_member_expr` short-circuits to "no side effects" when the member chain's root resolves to a manual pure function, before analyzing anything inside the chain: [stmt_eval_analyzer/mod.rs#L192-L194](https://github.com/rolldown/rolldown/blob/4930a00b00c224560f21bff9d541a5e67a7b3e39/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs#L192-L194). The namespace branch directly below shows the intended pattern — its comment notes that a computed key is still evaluated and may have its own side effects, and it recurses into `ComputedMemberExpression` keys ([#L195-L203](https://github.com/rolldown/rolldown/blob/4930a00b00c224560f21bff9d541a5e67a7b3e39/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs#L195-L203)).
- The pure-function match is by chain root only: `extract_first_part_of_member_expr_like` walks through computed members, static members, and calls down to the leftmost identifier ([#L832-L863](https://github.com/rolldown/rolldown/blob/4930a00b00c224560f21bff9d541a5e67a7b3e39/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs#L832-L863)).
- The same root-match is exposed to Oxc via the `MayHaveSideEffectsContext::manual_pure_functions` impl ([#L880-L882](https://github.com/rolldown/rolldown/blob/4930a00b00c224560f21bff9d541a5e67a7b3e39/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs#L880-L882)), which is how the `new (make()[effect()].Box)()` callee-chain effects get dropped.

### Provenance

Found by the execution-order fuzzer while auditing manually-pure calls; reduced cases and current-target signatures are recorded in [rolldown-order-fuzzer@77f196b](https://github.com/hyfdev/rolldown-order-fuzzer/commit/77f196bd11e6e388c3fab14c612514ee5dce7b31). It reproduces with fully default options and is an ordinary tree-shaking bug, unrelated to `strictExecutionOrder` or wrapping. Extracted from the fuzzer-findings list in https://github.com/rolldown/rolldown/issues/10294#issuecomment-5054455313.

Nic-Polumeyv:
Fixes #10397, at both drop paths named in the issue's Cause section.

One note for review. Two unit tests from #10387 encoded the old behavior as their premise, their comments said so explicitly, and they now expect conservative retention in both analysis dimensions.

### DIFF
--- a/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs
+++ b/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs
@@ -193,7 +193,7 @@ impl<'a> StmtEvalAnalyzer<'a> {
 
   fn analyze_member_expr(&self, member_expr: &ast::MemberExpression) -> StmtEvalFacts {
     if self.is_expr_manual_pure_functions(member_expr.object()) {
-      let mut facts = self.analyze_member_expr_like_eager_children(member_expr.object());
+      let mut facts = self.analyze_eager_chain_children(member_expr.object());
       if let ast::MemberExpression::ComputedMemberExpression(expr) = member_expr {
         facts |= self.analyze_expr(&expr.expression);
       }
@@ -218,9 +218,9 @@ impl<'a> StmtEvalAnalyzer<'a> {
     facts
   }
 
-  fn analyze_member_expr_like_eager_children(&self, expr: &Expression) -> StmtEvalFacts {
+  fn analyze_eager_chain_children(&self, expr: &Expression) -> StmtEvalFacts {
     let mut facts = StmtEvalFacts::default();
-    walk_member_expr_like(expr, |child| {
+    for_each_eager_chain_child(expr, |child| {
       facts |= child
         .map_or_else(StmtEvalFacts::from_unknown_side_effect, |child| self.analyze_expr(child));
     });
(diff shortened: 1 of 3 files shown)


IWANABETHATGUY:

IWANABETHATGUY (approved):

PR #10431, opened 2026-07-24
test: stop snapshots from silently hiding user code inside the runtime region

hyfdev:
Artifact snapshots can silently present user code as part of the hidden runtime region — six fixtures on main do this today. This PR keeps fixture output's region structure intact through the default dce pass, so the hide applies exactly to the runtime and everything else stays visible. `topics/runtime/transform` before and after:

```js
// before: the fixture's own code is invisible
// HIDDEN [\0rolldown/runtime.js]
export { value };
```

```js
// after: the chunk contained this all along
// HIDDEN [\0rolldown/runtime.js]
//#region dep.cjs
//#endregion
//#region entry.js
const value = (/* @__PURE__ */ __commonJSMin(((exports) => {
	exports.value = "cjs-value";
})))().value;
//#endregion
export { value };
```

### Problem

The snapshot harness hides the runtime by matching `//#region \0rolldown/runtime.js … //#endregion`, trusting the markers to delimit it. But markers are ordinary comments, and under the default `minify: 'dce-only'` a statement removed at a module boundary takes the adjacent `//#endregion` + `//#region` pair with it — the surviving markers still pair up textually, so the hide silently extends across the next module's code.

The comment loss is oxc behavior: a comment rides the statement that follows it, and the minifier drops a removed statement's comments (the pre-dce output is fully balanced, verified with `minify: false`). Filed upstream as oxc-project/oxc#24885.

Blast radius: the snapshot review surface. Production builds and the public API are untouched by this PR — the change lives entirely in the test harness. Real builds keep the corrupted markers until the upstream fix.

### Fix

Keep the markers alive across the dce pass, from inside the harness only: a plugin registered for every fixture build rewrites marker lines to legal-comment form (`//!#region`) in `renderChunk`, which runs before the internal minify pass — legal comments survive the removal of their host statement, that is their license-text contract — and rewrites them back in `generateBundle`. The rewrite is line-anchored (only a line starting with a marker token matches) and gated to dce-only builds; full minification strips all normal comments on purpose and is left alone.

Two alternatives were rejected: preserving the markers inside the minify pipeline itself would put workaround logic in production code paths, and hardening the snapshot hide heuristically (#10429, superseded) cannot recover a boundary the text no longer contains.

Non-goal: fixing emitted output for real builds — that is oxc-project/oxc#24885; this plugin gets deleted when it lands.

### Sourcemap impact

No mapping is invalidated, by construction: the rewrite is line-preserving (one `!` per marker line, line count unchanged), marker lines are comments carrying no mapping segments, and every other position is untouched. `renderChunk` returns `map: Null` — the explicit "no remapping needed" form, covered by an existing test asserting it raises no `SOURCEMAP_BROKEN`. Known precision edge: the emitted `.map` is finalized against the disguised text and the restore runs after, so marker lines' columns differ from the map by one character — those lines have no segments, so no lookup can observe it.

### Validation

- Unit test pins the rewrite as line-anchored and reversible (`//# sourceMappingURL` and string contents never match).
- Full fixture suite: the failure list is byte-identical to unmodified main on the same machine (the pre-existing local environment set). Every snapshot change was classified: restored marker pairs (the bulk), the runtime hide turning precise again (previously-leaked helper definitions hidden), the six swallow fixtures revealing their user code (`esbuild/default/external_es6_converted_to_common_js`, `function/module_types/asset`, `resolve/attach_correct_package_json`, `sourcemap/issue_6099`, `topics/runtime/transform`, `issues/4782`), and line/hash shifts that follow mechanically (visualizer sections shift by the restored lines; content hashes in eight hash-bearing fixtures change deterministically, since hashing runs while the disguise is in place).

Supersedes #10429. Long-term fix tracked in oxc-project/oxc#24885.

### DIFF
--- a/crates/rolldown_testing/src/integration_test.rs
+++ b/crates/rolldown_testing/src/integration_test.rs
@@ -26,6 +26,7 @@ use crate::hmr_files::{
   apply_hmr_edit_files_to_hmr_temp_dir, collect_hmr_edit_files,
   copy_non_hmr_edit_files_to_hmr_temp_dir, get_changed_files_from_hmr_edit_files,
 };
+use crate::preserve_region_markers::PreserveRegionMarkersPlugin;
 use crate::types::{
   BuildArtifactsSnapshot, BuildRoundOutput, DevArtifactsSnapshot, DevRoundOutput, HmrStepOutput,
 };
@@ -560,8 +561,12 @@ impl IntegrationTest {
   pub async fn run_multiple(
     &self,
     mut multiple_options: Vec<NamedBundlerOptions>,
-    plugins: Vec<SharedPluginable>,
+    mut plugins: Vec<SharedPluginable>,
   ) {
+    // Registered last so it runs right before the internal dce pass; see the module docs of
+    // `preserve_region_markers` for why snapshots need markers kept intact.
+    plugins.push(Arc::new(PreserveRegionMarkersPlugin));
+
     let test_folder_path = &self.test_folder_path;
 
     // Detect HMR mode by checking for HMR edit files
--- a/crates/rolldown_testing/src/preserve_region_markers.rs
+++ b/crates/rolldown_testing/src/preserve_region_markers.rs
@@ -1,0 +1,120 @@
+//! Test-harness plugin that keeps `attachDebugInfo` region markers intact through rolldown's
+//! default `dce-only` minify pass.
+//!
+//! Region markers are ordinary line comments, and oxc attaches a comment to the statement that
+//! follows it (blank lines don't detach it). When the dce pass removes that statement, its
+//! comments go with it — a removed statement at a module boundary silently takes the adjacent
+//! `//#endregion` + `//#region` pair out of the output. That corrupts the region structure the
+//! snapshot tooling relies on: the runtime-hide replacement matches `//#region … //#endregion`,
+//! so an eaten pair at the runtime boundary makes snapshots silently hide user code.
+//!
+//! Legal comments survive the removal of their host statement, so this plugin disguises the
+//! markers as legal comments right before the internal minify pass (`renderChunk` runs before it)
+//! and restores them after everything else (`generateBundle`). Both rewrites are line-anchored
+//! and reversible; codegen prints comments verbatim, so only the disguise prefix ever changes.
+//! The plugin is registered after every fixture-defined plugin, so a fixture's own
+//! `generateBundle` hook runs before the restore and observes the disguised `//!#region` form.
+//!
+//! This lives in the test harness on purpose: production builds and the public API are
+//! untouched. The proper long-term fix is for oxc to re-attach orphaned comments of removed
+//! statements, at which point this plugin can be deleted.
+
+use std::borrow::Cow;
+use std::sync::Arc;
+
+use rolldown::plugin::{
+  HookGenerateBundleArgs, HookNoopReturn, HookRenderChunkArgs, HookRenderChunkOutput,
+  HookRenderChunkReturn, HookTransformOutputMap, HookUsage, Plugin, PluginContext,
+};
+use rolldown_common::{MinifyOptions, Output};
+
+#[derive(Debug)]
+pub struct PreserveRegionMarkersPlugin;
+
+impl Plugin for PreserveRegionMarkersPlugin {
+  fn name(&self) -> Cow<'static, str> {
+    "rolldown-testing:preserve-region-markers".into()
+  }
+
+  fn register_hook_usage(&self) -> HookUsage {
+    HookUsage::RenderChunk | HookUsage::GenerateBundle
+  }
+
+  async fn render_chunk(
+    &self,
+    _ctx: &PluginContext,
+    args: &HookRenderChunkArgs<'_>,
+  ) -> HookRenderChunkReturn {
+    // Only the dce-only pass eats markers selectively; full minification strips all normal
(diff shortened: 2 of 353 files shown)


crates/rolldown_testing/src/preserve_region_markers.rs
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  async fn generate_bundle(
+    &self,
+    _ctx: &PluginContext,
+    args: &mut HookGenerateBundleArgs<'_>,
+  ) -> HookNoopReturn {
+    for output in args.bundle.iter_mut() {
+      if let Output::Chunk(chunk) = output {
+        if chunk.code.contains("//!#") {
```
h-a-n-a, line 76: nit: the disguise only runs for dce-only builds, but this restore runs for every build. If a fixture's own source ever contains a legal comment line starting with `//!#region`, this would rewrite it to `//#region` even though no disguise step produced it — and the runtime hide could then swallow it. `args.options.minify` is available here, so the restore could use the same `DeadCodeEliminationOnly` gate as `render_chunk` to keep both sides symmetric. Very unlikely to matter in practice.

crates/rolldown_testing/src/integration_test.rs
```diff
@@ -560,8 +561,12 @@ impl IntegrationTest {
   pub async fn run_multiple(
     &self,
     mut multiple_options: Vec<NamedBundlerOptions>,
-    plugins: Vec<SharedPluginable>,
+    mut plugins: Vec<SharedPluginable>,
   ) {
+    // Registered last so it runs right before the internal dce pass; see the module docs of
+    // `preserve_region_markers` for why snapshots need markers kept intact.
+    plugins.push(Arc::new(PreserveRegionMarkersPlugin));
```
h-a-n-a, line 568: nit: because this plugin is appended last, a fixture-defined `generate_bundle` hook runs before the restore and would see the disguised `//!#region` text. No current fixture does that, but one sentence about this in the module docs of `preserve_region_markers` would save a future fixture author some confusion.

crates/rolldown_testing/src/preserve_region_markers.rs
```diff
@@ -0,0 +1,112 @@
...
+      }
+    }
+    Ok(())
+  }
+}
+
+/// Rewrite the marker token of every line that starts (after indentation) with
+/// `{from}region` or `{from}endregion`. Line-anchored, so `//# sourceMappingURL` and marker-like
+/// text inside string content never match; only the first token on the line is touched.
+fn map_region_marker_lines(source: &str, from: &str, to: &str) -> String {
```
h-a-n-a, line 88: nit: this allocates a new `String` for every line, including untouched ones. Building one output `String` and appending with `push_str` would avoid that. Test-only path, so fine as is.

h-a-n-a (approved):
I checked the mechanism against the source and it holds up:

- Plugin `render_chunk` runs before the internal `minify_chunks` pass, and `generate_bundle` runs after hashing, so the disguise and restore surround the dce pass exactly (`render_chunk_to_assets.rs`).
- The dce-only pass keeps legal comments (`legal: Inline` in `minify_chunks.rs`), so the disguised markers survive statement removal.
- `map: Null` is the right signal here: an omitted map would raise a SOURCEMAP_BROKEN warning, and marker lines carry no mapping segments.
- Every fixture path goes through `run_multiple`, so the plugin covers all fixture builds.

One thing to accept on purpose: snapshots now show intact region markers (and different content hashes) that a real dce-only build would not produce until oxc-project/oxc#24885 lands. I think that trade is right — a snapshot that silently labels user code as hidden runtime is worse than one that differs from the shipped bytes in comment lines.

A few small nits inline, none blocking.

hyfdev:
### Merge activity

* **Jul 24, 6:58 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 24, 7:02 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 24, 7:06 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 24, 7:12 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

crates/rolldown_testing/src/preserve_region_markers.rs
hyfdev: Good catch — the restore now uses the same `DeadCodeEliminationOnly` gate as the disguise (25cde9ec4), so the two sides stay symmetric and fixture-authored `//!#region`-shaped content is never rewritten on other builds.

crates/rolldown_testing/src/integration_test.rs
hyfdev: Added to the module docs in 25cde9ec4: the plugin registers after every fixture-defined plugin, so a fixture's own `generateBundle` runs before the restore and observes the disguised form.

crates/rolldown_testing/src/preserve_region_markers.rs
hyfdev: Reworked in 25cde9ec4 to append into a single output buffer.

PR #10432, opened 2026-07-24
fix(treeshake): preserve side-effect-free spread arguments in manual-pure chains

IWANABETHATGUY:
Follow-up to #10427.

#10427 restored the eagerly-evaluated child effects of manual-pure chains, but its eager-child walk maps *every* spread argument to an unknown side effect. So a manual-pure chain wrapping a statically-safe spread — `make(...[]).value`, `make(...'').value`, `make(...[])()` — is now retained wholesale, even though spreading an array, string, or template literal only runs built-in iterators. Rollup drops these.

## Fix

Surface spread arguments distinctly from plain eager children (new `EagerChild` enum) and analyze them with `analyze_spread_argument`, mirroring oxc's `Argument::may_have_side_effects`:

- array / string / template-literal operands are analyzed through the (oxc-gated) expression analyzer, so a side-effectful element (`make(...[effect()])`) or a nested spread (`make(...[...x])`) still retains the statement;
- any other operand stays conservatively unknown, since evaluating or iterating it may run a user-defined `Symbol.iterator`.

## Tests

New unit test `test_manual_pure_chains_drop_side_effect_free_spreads` covers both directions: safe literal spreads become removable, while effectful or opaque spreads stay retained — including the nested-spread (`...[...x]`) and bare-identifier (`...x`) cases that must not under-retain.

### DIFF
--- a/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs
+++ b/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs
@@ -3,7 +3,7 @@ use std::ops::{BitOr, BitOrAssign};
 use bitflags::bitflags;
 use oxc::ast::ast::{
   self, Argument, AssignmentTarget, BindingPattern, CallExpression, ChainElement, Expression,
-  IdentifierReference, UnaryOperator, VariableDeclarationKind,
+  IdentifierReference, SpreadElement, UnaryOperator, VariableDeclarationKind,
 };
 use oxc::ast::match_member_expression;
 use oxc::ast_visit::{VisitJs, walk_js};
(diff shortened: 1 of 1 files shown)


hyfdev (approved):

hyfdev:
### Merge activity

* **Jul 24, 6:07 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 24, 6:07 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 24, 6:13 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10433, opened 2026-07-24
fix(code-splitting): avoid the dynamic-entry facade when the same-chunk collapse carries the trigger

hyfdev:
Under `strictExecutionOrder`, dynamically importing a module that chunking already placed in the importer's own chunk still shipped a dedicated facade chunk, and the `import()` stayed a real network fetch of that file. After this PR the import collapses inline and carries the module's deferred initialization with it, so the facade chunk disappears — matching what the same layout already produced without strict mode. A facade is still emitted whenever any other chunk dynamically imports the module.

```js
// before — the importer fetches a file whose whole job is the trigger:
import("./lib.js").then((mod) => ...)
// dist/lib.js:
//   import { n as a, r as init_lib } from "./entry2.js";
//   init_lib();
//   export { a };

// after — dist/lib.js is gone; the collapse carries the trigger:
Promise.resolve().then(() => (init_lib(), lib_exports)).then((mod) => ...)
```

### Problem

Second P1 item of #10294. Strict-order builds defer each module's execution behind an initializer, and a dynamically imported module needs that initializer called when the import resolves. When code splitting co-locates the import target with its importer — a `codeSplitting` group matching both, or `experimental.chunkOptimization` merging the dynamic entry into its host — the build resurrected the eliminated facade chunk to carry that call, even though nothing else would ever load it. `issues/9463` wrap-all ships a `shared.js` chunk containing only a noop initializer call plus a re-export; the same shape recurs with real, non-noop initializers in `manual_group_phantom_dynamic`, `mixed_static_dynamic_same_target`, and `chunk_merging/dynamic_entry_merged_in_user_defined_entry`. Wrap-all hits every such entry; on-demand hits whichever targets the analysis wraps.

Root cause: the restore pass (`restore_order_wrap_entry_facades`, which un-tombstones facade chunks the chunk optimizer eliminated) ran unconditionally for every wrapped entry, because the finalizer's merged-entry rewrite (`rewrite_dynamic_import_for_merged_entry`, which turns `import()` of a merged dynamic entry into an inline namespace access) could not express an execution-order trigger — its non-interop arms asserted such entries never reach them. The interop arm already carries its wrapper's trigger through the collapse, and the cross-chunk links stage already publishes execution-order wrapper exports; the machinery was half-built, so the restore undid every merge the optimizer had proved.

Blast radius: output shape only — one extra chunk and one extra request per affected dynamic entry; runtime behavior was already correct. Not affected: non-strict builds, single-chunk builds, `emitFile`'d chunks, and any dynamic entry with a cross-chunk dynamic importer.

### Fix

Keep the facade eliminated exactly when every live dynamic importer sits in the chunk that hosts the entry's implementation, and let the collapse carry the trigger.

- The merged-entry rewrite now routes both arms through the shared `esm_init_target` view, which presents interop and execution-order wrappers uniformly, and emits `Promise.resolve().then(() => (init_x(), ns))` for the same-chunk case. A noop initializer call gets the `/* @__PURE__ */` annotation, so the default dce-only minify drops it — `issues/9463` wrap-all becomes byte-identical to its on-demand cell. A new `debug_assert!(!target.tla_tainted)` pins why the non-awaiting collapse is sound: facade elimination is globally disabled when any module is TLA-tainted.
- The restore still runs when any cross-chunk dynamic importer exists: a facade runs the trigger synchronously within its own module evaluation, while a `.then` trigger would run a microtask after the host chunk settles — an observable reordering that `m4_dynamic_facade_race` and `wrapped_dynamic_entry_keeps_facade_after_manual_chunk_merge` pin (both stay green). Emitted-chunk facades are always restored so `emitFile` reference ids keep resolving to a real file. The importer scan is one pass over the module table shared by all restore candidates.
- The plan-applied debug assertion now accepts an entry whose entry chunk is another module's entry chunk containing it (dynamic entry merged into a user-defined entry), a state the unconditional restore previously made unreachable.

Why not the ledger's narrower alternative — skip the facade only when the initializer is a noop: three of the four anchor fixtures have real initializers (`init_m6`, `init_dep_b`, `init_lib`), so only trigger-carrying collapse removes their facades.

Non-goals: the cross-chunk `esm_init_target` arm is dead-by-construction today (any cross-chunk importer forces the restore) and is kept for symmetry, not as a functional path; routing the interop arm through the shared init-call helper widens the PURE-when-noop annotation to merged interop entries in non-strict builds (no existing snapshot changes); the wrap-all-side unconditional entry split stays as tracked in #10294.

### Validation

- New fixture `strict_execution_order/cross_chunk_dynamic_importer_keeps_facade` pins the boundary in minimal form (two entries, one target, one group): one same-chunk plus one cross-chunk dynamic importer keeps the restored facade with its synchronous trigger, in both wrap modes. Verified it bites: neutralizing the cross-chunk guard flips this fixture's snapshot, while its runtime assertions alone would not catch it — the microtask-race semantics stay pinned by `m4_dynamic_facade_race`.
- Acceptance and anchors, green with runtime `_test.mjs` assertions: `issues/9463` wrap-all loses the `shared.js` facade (byte-identical to on-demand), `issues/9463_plain_group` likewise; `manual_group_phantom_dynamic` drops `m6.js` in both wrap modes; `mixed_static_dynamic_same_target` drops `dep-b.js`; `chunk_merging/dynamic_entry_merged_in_user_defined_entry` drops `lib.js` in the esm and cjs wrap-all cells and stops publishing the now-dead exports.
- Full `cargo test -p rolldown --test integration`: 1889 passed; the only failures are the pre-existing environment set (`cjs_module_lexer_compat` ×3, `npm_packages/util_deprecate`, `test262`), verified byte-identical by rerun on unmodified main on the same machine.
- Execution-order fuzzer differential, 80 mixed cases per cell (seeds 400001–400080) in on-demand and wrap-all against this build and an unmodified-main build: identical results in all four cells (78/80); the two failing seeds are the known packaged pure-definer-behind-barrel residue tracked in #10294, with byte-identical verdict signatures on both builds — zero failure signatures introduced or altered.
- Not exercised by a fixture: a TLA-tainted entry reaching the collapse. It is unreachable today (facade elimination is globally disabled when any module is TLA-tainted) and the new debug assertion pins that coupling.

Part of #10294.

### DIFF
--- a/crates/rolldown/src/ast_scanner/impl_visit.rs
+++ b/crates/rolldown/src/ast_scanner/impl_visit.rs
@@ -214,6 +214,9 @@ impl<'me, 'ast: 'me> VisitJs<'ast> for AstScanner<'me, 'ast> {
           meta.set(ImportRecordMeta::IsTopLevel, self.is_root_scope());
           meta.set(ImportRecordMeta::IsUnspannedImport, expr.source.span().is_empty());
           meta.set(ImportRecordMeta::InTryCatchBlock, self.in_side_try_catch_block());
+          // Mirrors the bail condition in `try_rewrite_import_expression` exactly: any options
+          // argument at all makes it leave the specifier untouched.
+          meta.set(ImportRecordMeta::DynamicImportWithOptions, expr.options.is_some());
           meta
         },
         Some(expr.node_id()),
--- a/crates/rolldown/src/module_finalizers/mod.rs
+++ b/crates/rolldown/src/module_finalizers/mod.rs
@@ -2242,70 +2242,52 @@ impl<'me, 'ast> ScopeHoistingFinalizer<'me, 'ast> {
     if is_importer_importee_in_same_chunk {
       let importee_meta = &self.ctx.linking_infos[importee.idx];
 
-      let finalized_expr = match importee_meta.wrap_kind() {
-        WrapKind::Cjs => {
-          let importee_wrapper_ref = self.ctx.linking_infos[importee.idx].wrapper_ref.unwrap();
+      let finalized_expr = if matches!(importee_meta.wrap_kind(), WrapKind::Cjs) {
+        let importee_wrapper_ref = self.ctx.linking_infos[importee.idx].wrapper_ref.unwrap();
 
-          let (finalized_importee_wrapper_ref, _) =
-            self.finalized_expr_for_symbol_ref(importee_wrapper_ref, false, false);
+        let (finalized_importee_wrapper_ref, _) =
+          self.finalized_expr_for_symbol_ref(importee_wrapper_ref, false, false);
 
-          let finalized_to_esm = self.finalized_expr_for_runtime_symbol("__toESM");
+        let finalized_to_esm = self.finalized_expr_for_runtime_symbol("__toESM");
 
-          // require_xxx()
-          let wrapper_ref_call_expr = ast::Expression::new_call_expression(
-            SPAN,
-            finalized_importee_wrapper_ref,
-            NONE,
-            [],
-            false,
-            self,
-          );
+        // require_xxx()
+        let wrapper_ref_call_expr = ast::Expression::new_call_expression(
+          SPAN,
+          finalized_importee_wrapper_ref,
+          NONE,
+          [],
+          false,
+          self,
+        );
 
-          // __toESM(require_xxx(), isNodeMode)
-          Expression::new_to_esm_wrapper(
-            finalized_to_esm,
-            wrapper_ref_call_expr,
-            self.ctx.module.should_consider_node_esm_spec_for_dynamic_import(),
-            self,
-          )
-        }
-        WrapKind::Esm => {
-          // (init_xxx(), namespace_exports)
-          let importee_wrapper_ref = self.ctx.linking_infos[importee.idx].wrapper_ref.unwrap();
+        // __toESM(require_xxx(), isNodeMode)
+        Expression::new_to_esm_wrapper(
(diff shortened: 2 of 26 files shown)


hyfdev:
* **#10485** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10485?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10456** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10456?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10433** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10433?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10433?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10433).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

shulaoda:
@hyfdev please resolve the conflicts

hyfdev:
> @hyfdev please resolve the conflicts

Done

shulaoda:
@codex review

hyfdev:
Ive changed the priority to p1 as this needs to to merged before next release.

shulaoda (approved):

PR #10438, opened 2026-07-24
fix(import-glob): escape generated string literals

### LINKED ISSUE #10437: [Bug]: viteImportGlobPlugin does not escape quotes in matched file paths

[REPL](https://stackblitz.com/edit/rolldown-rolldown-starter-stackblitz-fnqwuvsf?file=src%2Fmain.js)

Build fails with the error

```

[PARSE_ERROR] Expected a semicolon or an implicit semicolon after a statement, but found none
   ╭─[ src/main.js:1:43 ]
   │
 1 │ import * as __import_glob__0_0_ from "./a"b.js";export default /* #__PURE__ */ Object.assign({"./a"b.js": __import_glob__0_0_});
   │                                           │ 
   │                                           ╰─ 
   │ 
   │ Help: Try inserting a semicolon here
───╯
```

### What is expected?

bundle.generate() should succeed. The generated module specifier and glob object key should escape the double quote.

The native Vite import-glob plugin interpolates the path without escaping it:

```js
 import * as __import_glob__0_0_ from "./a"b.js";
```

Rolldown then reports:

```
 [PARSE_ERROR] Expected a semicolon or an implicit semicolon after a statement
```

Affects eager imports, lazy imports, and generated object keys.

### System Info

```Shell
System:
    OS: macOS 26.6
    CPU: (8) arm64 Apple M2
    Memory: 163.66 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 26.3.0 - /Users/shantanu/.local/share/mise/installs/node/26/bin/node
    Yarn: 4.12.0 - /Users/shantanu/.local/share/mise/installs/node/26/bin/yarn
    npm: 11.16.0 - /Users/shantanu/.local/share/mise/installs/node/26/bin/npm
    bun: 1.3.10 - /Users/shantanu/.bun/bin/bun
    Deno: 2.9.4 - /opt/homebrew/bin/deno
    Watchman: 2026.07.20.00 - /opt/homebrew/bin/watchman
  Browsers:
    Firefox Developer Edition: 154.0
    Safari: 27.0
```

### Any additional comments?

I noticed this during the upgrade from Astro 6 to 7 which switched to rolldown I have a bunch of images where the name contains quotes eg. `2023-05-17-15-24-00-43°30'30.2"N 16°26'29.0"E.jpg`

shantanuraj:
- Escape module specifiers and object keys emitted by the native Vite import-glob plugin.
- Unescaped quotes in matched file names produced invalid JavaScript during builds.

Closes #10437

### DIFF
--- a/crates/rolldown_plugin_vite_import_glob/src/utils.rs
+++ b/crates/rolldown_plugin_vite_import_glob/src/utils.rs
@@ -9,7 +9,10 @@ use oxc::ast::ast::{
 use oxc::ast_visit::{VisitJs, walk_js};
 use rolldown_ecmascript_utils::ExpressionExt;
 use rolldown_plugin::{LogWithoutPlugin, PluginContext};
-use rolldown_plugin_utils::constants::{ViteImportGlob, ViteImportGlobValue};
+use rolldown_plugin_utils::{
+  constants::{ViteImportGlob, ViteImportGlobValue},
+  to_string_literal,
+};
 use rolldown_std_utils::relative_path_to_slash;
 use string_wizard::MagicString;
 use sugar_path::SugarPath;
@@ -185,6 +188,7 @@ impl<'ast> GlobImportVisit<'_> {
         Cow::Borrowed(import_path)
       };
 
+      let formatted_file = to_string_literal(&formatted_file);
       let value: Cow<'_, str> = if matches!(omit_type, ImportGlobOmitType::Values) {
         Cow::Borrowed("0")
       } else if options.eager {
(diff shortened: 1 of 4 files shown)


hyfdev (approved):
Matched paths are now emitted through `to_string_literal` at the same sites Vite covers with `JSON.stringify` (eager `from`, lazy `import()`, object keys, and the lazy `import` option property access), so a quote in a filename no longer splits the generated source.

The `escaped-paths` fixture is a good regression for #10437: it creates real quoted names so the glob still discovers them, covers eager / lazy / `Object.keys`, and asserts both parse-time success and runtime key identity. The virtual resolve plugin cannot mask the original PARSE_ERROR, and the Windows skip is appropriate because `"` cannot appear in Windows filenames.

hyfdev:
LGTM. Wait for @shulaoda 's second look.

shulaoda (approved):

PR #10439, opened 2026-07-25
docs: clarify that `esmExternalRequirePlugin` must own its externals

TheAlexLichter:
While migrating https://github.com/cloudflare/kumo I noticed that the docs of `esmExternalRequirePlugin` around externalized modules are a bit ambiguously worded and need a highlight around "what to do" for both humans and agents 😋


shulaoda (approved):

hyfdev:
Thank you!

PR #10440, opened 2026-07-25
fix(binding): forward resolveId options.kind to callable builtin plugins

### LINKED ISSUE #10403: [Bug]: Callable builtin plugin resolveId hardcodes ImportKind::Import — programmatic resolveId can never select the "require" condition

Three files, `npm i`, `node repro.mjs` (rolldown 1.1.4, Node 24.18.0, macOS):

`package.json`

```json
{"name":"repro","private":true,"type":"module","dependencies":{"rolldown":"1.1.4","dual-pkg":"file:./dual-pkg"}}
```

`dual-pkg/package.json` (+ trivial `index.mjs`/`index.cjs`)

```json
{"name":"dual-pkg","version":"1.0.0","exports":{".":{"import":"./index.mjs","require":"./index.cjs"}}}
```

`repro.mjs` — constructs `viteResolvePlugin` from `rolldown/experimental` the way Vite 8 does for an unbundled server environment, then calls its `resolveId` twice:

```js
import path from "node:path";
import { fileURLToPath } from "node:url";
import { viteResolvePlugin } from "rolldown/experimental";

const root = path.dirname(fileURLToPath(import.meta.url));
const importer = path.join(root, "src/importer.js");

function makePlugin(staticIsRequire) {
  const plugin = viteResolvePlugin({
    resolveOptions: {
      isBuild: false, isProduction: false, asSrc: false, preferRelative: false,
      ...(staticIsRequire === undefined ? {} : { isRequire: staticIsRequire }),
      root, scan: false,
      mainFields: ["module", "jsnext:main", "jsnext", "main"],
      conditions: ["node"], externalConditions: ["node"],
      extensions: [".mjs", ".js", ".mts", ".ts", ".jsx", ".tsx", ".json"],
      tryIndex: true, preserveSymlinks: false, tsconfigPaths: false,
    },
    environmentConsumer: "server", environmentName: "ssr",
    builtins: [], external: true, noExternal: true, dedupe: [],
    resolveSubpathImports: () => undefined,
  });
  const hook = plugin.resolveId;
  return (id, imp, opts) =>
    typeof hook === "function" ? hook(id, imp, opts) : hook.handler(id, imp, opts);
}

const viaKind = await makePlugin(undefined)("dual-pkg", importer, { kind: "require-call" });
console.log('per-call kind "require-call" ->', viaKind?.id);

const viaStatic = await makePlugin(true)("dual-pkg", importer, {});
console.log("static isRequire: true       ->", viaStatic?.id);
```

Output:

```
per-call kind "require-call" -> …/dual-pkg/index.mjs
static isRequire: true       -> …/dual-pkg/index.cjs
```

### What is expected?

A resolveId call that declares itself a require-call (`kind: "require-call"`, the same options shape Vite's plugin container forwards to plugin hooks) should reach the native plugin as `ImportKind::Require`, so exports/imports condition resolution includes `require` and returns `index.cjs`. The second arm shows the resolver itself fully supports this — only the per-call signal never arrives.

`options.kind` is silently discarded at the JS→native boundary: `BindingHookJsResolveIdOptions` carries only `isEntry`/`scan`/`custom` (`crates/rolldown_binding/src/options/plugin/binding_callable_builtin_plugin.rs:198`), and `BindingCallableBuiltinPlugin::resolve_id` hardcodes `kind: ImportKind::Import` (`:101`; both unchanged at current main, `4930a00b0`). So `vite_resolve_plugin.rs:338`'s `is_require: self.resolve_options.is_require.unwrap_or(args.kind == ImportKind::Require)` is always false on this path, and `resolver.rs` pushes the `import` condition.

Downstream, this means Vite 8's `environment.pluginContainer.resolveId` has no working way to resolve a `require()` call: Vite 7's `custom["node-resolve"].isRequire` handling no longer exists in Vite 8, and the `kind` Vite forwards dies here. A concrete casualty is `@cloudflare/vitest-pool-workers`' module-fallback service for workerd (cloudflare/workers-sdk#12984): `require()` of dual-format packages (e.g. `pg` → `pg-protocol`) resolves to the ESM entry inside workerd, throwing `Cannot use import statement outside a module`; it's currently worked around downstream with a post-hoc exports-map rescan (cloudflare/workers-sdk#13062).

### System Info

```shell
System:
  OS: macOS 26.5.2
  CPU: (14) arm64 Apple M4 Pro
  Memory: 5.67 GB / 48.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 24.18.0 - ~/.local/share/mise/installs/node/24.18.0/bin/node
  npm: 11.16.0 - ~/.local/share/mise/installs/node/24.18.0/bin/npm
  pnpm: 11.11.0 - ~/.local/share/mise/installs/pnpm/11.11.0/pnpm
npmPackages:
  rolldown: 1.1.4 => 1.1.4
```

### Any additional comments?

Proposed fix (per the contribution guide's discuss-first path — seeking agreement before a PR): add an optional `kind` field to `BindingHookJsResolveIdOptions`, map it onto `ImportKind` (defaulting to `Import` when absent, preserving current behavior), and use it in place of the hardcoded value in `resolve_id`. Happy to send the PR with tests once the direction is agreed.

martijnwalraven:
### Description

Programmatic `resolveId` calls on callable builtin plugins (`rolldown/experimental`, the path Vite 8's plugin container uses) hardcoded `kind: ImportKind::Import` at the JS→native boundary: `BindingHookJsResolveIdOptions` carried only `isEntry`/`scan`/`custom`, so `options.kind` was silently discarded and a `require-call` resolution could never select the `require` exports condition (details and repro in #10403).

Per the direction agreed in the issue, this PR:

- adds an optional `kind` field to `BindingHookJsResolveIdOptions`, with the same vocabulary and doc comment as the existing `BindingHookResolveIdExtraArgs.kind` and `BindingPluginContextResolveOptions.importKind`;
- maps it onto `ImportKind` with the existing `TryFrom<&str>` impl and uses it in place of the hardcoded value in `BindingCallableBuiltinPlugin::resolve_id` — absent `kind` still defaults to `ImportKind::Import`, so existing callers are unaffected;
- rejects an invalid `kind` with an `InvalidArg` error (raised before the async block, so the async-runtime acquire/release pairing is untouched).

The `packages/rolldown/src/binding.d.cts` hunk is the regenerated napi output from `just build-rolldown`.

Fixes #10403.

### Tests

A new `packages/rolldown/tests/builtin-plugin/vite-resolve.test.ts` drives `viteResolvePlugin(...)`'s `resolveId` against a new dual-format fixture package (`exports: { import, require }`): `kind: 'require-call'` resolves the `require` entry (fails without this PR — it resolved the `import` entry), `kind: 'import-statement'` and absent `kind` resolve the `import` entry, and an invalid `kind` rejects. The fixture follows the existing `package-fixtures` + `link:` convention, with the matching `knip.jsonc` ignore entry.

### AI disclosure

Authored with Claude Code (Claude Fable 5), under my direction and review.


shulaoda (approved):
LGTM! Thanks!

PR #10441, opened 2026-07-25
fix(code-splitting): only split an entry facade when something else can load its chunk

hyfdev:
## Summary

Under `strictExecutionOrder` (SEO), an entry's body moves into `init_E()` and the entry chunk carries a top-level `init_E()` call. That inline trigger is only safe while the chunk hosting `E` is loaded *solely to enter* `E`. Once another chunk imports that chunk, the entry's program runs during an unrelated load — which is exactly what SEO exists to prevent. The facade split moves the trigger into its own file when that happens.

`create_order_wrap_entry_facades` asked that question only of interop-wrapped entries, and only under `onDemandWrapping`. **Every order-wrapped entry split unconditionally**, so each one cost an extra chunk even when nothing but the entry itself could ever load it — including every dynamic entry, i.e. one extra file per lazy route.

This PR gates all three facade sources on one predicate — *can anything else load this chunk?* — answered from the real cross-chunk link computation instead of a prediction.

Measured on a generated 120-lazy-route app:

| | `main` | this PR |
| --- | --- | --- |
| SEO off | 146 chunks | 146 |
| `strictExecutionOrder` (wrap-all) | **268** (+122) | **147** (+1) |
| `strictExecutionOrder` + `onDemandWrapping` | **267** (+121) | **147** (+1) |

Raw JS bytes in wrap-all: 96,326 → 82,870 (−14%). The `+1` in both fixed columns is the runtime chunk. The executed output of all three modes is byte-identical before and after.

Relates to #10294. Independent of #10433, which addresses a different facade case.

## Why the decision could not read facts before

The facade decision sits in `finalize_chunk_plan`, which runs *before* `compute_cross_chunk_links`. The only edge set available was `predicted_static_import_edges` — computed with an **empty** `OrderWrapState`, i.e. "what would the edges be if nothing were wrapped" — and in wrap-all mode not computed at all (`wrap_all_order_analysis` filled `import_edges` with empty sets, which the `!on_demand ||` short-circuit papered over).

## Alternatives explored

Two ways of re-deriving the missing edges were tried and rejected, both unsound in the same way:

1. **Scan the module graph** for chunks that some included module in another chunk statically imports into. Misses `init_*` obligations that reach their target *through a non-included forwarder* — `collect_order_wrap_esm_init_targets` descends into the import records of a tree-shaken forwarder, so both hops have a non-included endpoint and the scan skips them. Produces an entry whose program runs during another entry's load.
2. **Reuse `post_lowering_import_edges`** (the emergent-cycle fixpoint's projection). That projection deliberately omits interop `WrapKind::Esm` overlay edges and consumption-gated hops, because for *cycle detection* projecting them would fabricate a cycle and over-wrap. Omissions that are safe there are unsafe here: it drops a needed facade and breaks `retained_star_renamed_cycle` at runtime.

The lesson is that any hand-rolled reachability scan silently misses an edge source, and a missed edge is an execution-order bug with no signal. So this PR asks the real computation rather than re-deriving it.

## How the query is made possible

`collect_depended_symbols` does not only read the chunk graph — it writes each declared symbol's owning chunk back into the shared symbol database, and `compute_chunk_imports` reads that ownership back to resolve which chunk a depended symbol is imported from. The write is load-bearing *within* one pass, so the pass cannot be made read-only by dropping it.

It can be scoped instead. `SymbolChunkOwnership::Scoped` records the previous owner of every symbol the pass assigns; the caller restores them afterwards, leaving the symbol database exactly as it found it. `lowered_static_import_edges` drives the pass with the fully lowered `OrderWrapState` and the real final init metadata, so every edge source the final pass registers — value references, `init_*` forwarding, retained re-export overlays, transitive init obligations — is present, with no approximation.

**Why deciding before the facades exist is sound.** A facade is an empty chunk that takes over the entry role from the implementation chunk: it adds exactly one importer of that chunk (itself), and it inherits — never widens — the outgoing edges the implementation chunk had as an entry chunk. So for every *other* chunk the relation "some chunk other than me imports me" is unchanged, and all facade decisions can be taken together from one pre-facade snapshot.

## Ordering changes

Three passes move ahead of the query, none of which depends on facades existing:

- `ensure_runtime_module_for_order_wraps` — resolving a depended symbol to its chunk requires the runtime module to already sit in one, and order wrappers demand runtime helpers. A facade chunk copies its implementation chunk's bits, so the runtime chunk's bit union is unchanged.
- `restore_order_wrap_entry_facades` — `compute_chunk_imports` **skips chunks marked `Removed`**, so restoring afterwards would hide the edge from a resurrected facade to the chunk hosting its implementation, and an entry sharing that chunk would keep an inline trigger that the restored facade's load then fires. This one is not hypothetical: an earlier revision of this PR hit it, and the fixture suite did not catch it.
- `finalized_module_namespace_ref_usage` — so the query sees the namespaces the final link pass will see rather than the provisional pre-lowering set. No output changes from this today; it closes the gap rather than fixing an observed failure.

Restoring stays **ungated**. Unlike splitting, it exists because a chunk can host only one entry's top-level trigger, so an entry merged into another entry's chunk has no trigger at all without its own file.

### The regression test that was missing

The over-split was invisible to every existing test, and that is the more important thing to fix: an
output-correctness test cannot see it, because the over-split bundle still *runs* correctly — it just
ships one extra file per entry. `packages/rolldown/tests/behaviors/strict-execution-order-entry-facade.test.ts`
pins the **shape** of the output instead, keying on the property that defines a facade: a chunk
holding zero modules, existing only to host an entry's `init_*()` trigger.

Both directions are pinned, so the budget check cannot be satisfied by simply never splitting:

| | assertion |
| --- | --- |
| six lazy routes over a shared library | no facade, and no chunk beyond the shared runtime |
| two entries, one sharing a chunk with a module the other imports | the facade is still emitted, and the chunk holding the entry's own module is no longer an entry chunk |

Each runs under wrap-all and `onDemandWrapping`. **Against the parent commit all four fail** — seven
facades for the lazy-route app, and `['b', 'e']` instead of `['e']` for the shared-chunk app.

The manual group in the lazy-route app is load-bearing: it makes the chunk-driven evaluation order
differ from source order, which is what puts modules on the on-demand plan. Without it on-demand
finds no hazard, wraps nothing, and the shape does not exercise entry splitting at all.

### Execution-order coverage

New fixture `function/experimental/strict_execution_order/entry_shared_chunk_facade`: entry `e` shares a chunk with `shared.js`, which entry `b` imports. The facade must be kept, and `_test.mjs` asserts `e`'s program does not run while `b` is loading. Forcing the predicate to return an empty edge set makes it fail with `['E body', 'S body', 'B body S']` — the violation this PR must not introduce.

Existing coverage that pins the discrimination in both directions: `m4_dynamic_facade_race` keeps `target.js`'s facade (its chunk also hosts `observer.js`) while dropping `a2.js`/`b2.js`; `dynamic_target_in_entry_chunk` keeps `a.js`'s and drops `c.js`'s.

89 snapshots update, removing 207 emitted assets. No fixture loses a facade it needs — every strict fixture executes its output under Node.

- Rust integration suite: 1896 passed / 0 failed, identical to `main` on the same machine.
- Node suite: 913 passed / 0 failed.

## Reviewer attention

- **The soundness argument for a pre-facade snapshot** (the "How the query is made possible" section) is the load-bearing claim. If a facade can widen the set of chunks that import some *other* chunk, the decision would need a fixpoint.
- **The restore/query ordering** — the removed-chunk skip in `compute_chunk_imports` is easy to miss and produces a silent wrong answer.
- **Build time.** Strict mode now runs one extra `compute_wrapped_esm_init_metadata` and one extra `compute_cross_chunk_link_state`. On the 120-route app, wrap-all median 63.9ms → 67.6ms (~+6%); SEO-off is unchanged. Partly offset downstream by 45% fewer chunks to name, render and hash; the balance on a large app is worth a look.

## Unrelated pre-existing bug found while testing

`preserveModules` + `strictExecutionOrder` panics on `main` at `chunk.entry_module(...).unwrap()` in `render_chunk_exports`, because strict lowering demotes an entry chunk to `Common` while `preserve_modules` assumes every chunk has an entry module. Behaviour is identical before and after this PR; filing separately.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/order_wrapping.rs
+++ b/crates/rolldown/src/stages/generate_stage/order_wrapping.rs
@@ -5,6 +5,7 @@ use crate::{
 };
 use itertools::Itertools;
 use oxc::ast::ast::{Declaration, ExportDefaultDeclarationKind, Statement};
+use oxc_index::{IndexVec, index_vec};
 use rolldown_common::{
   Chunk, ChunkIdx, ChunkKind, ChunkMeta, ImportKind, ImportRecordIdx, ImportRecordMeta,
   IndexModules, ModuleIdx, OutputFormat, PostChunkOptimizationOperation, RuntimeHelper,
@@ -111,8 +112,19 @@ impl GenerateStage<'_> {
     let plan = &analysis.plan;
     if plan.is_empty() {
       // Entry-trigger facades are needed even with an empty plan: a pure interop graph can
-      // still share one entry's chunk with another entry.
-      if !self.create_order_wrap_entry_facades(chunk_graph, analysis) {
+      // still share one entry's chunk with another entry. With nothing order-wrapped, the only
+      // candidates are interop-wrapped entries, so a pure-ESM graph has none and asks nothing.
+      //
+      // The three passes hoisted ahead of the query below are skipped here, which is safe only
+      // because an empty plan makes each one a no-op: nothing demands an order-wrap runtime
+      // helper, restoring is filtered by the plan, and `finalize_chunk_plan` already ran the
+      // namespace pass against this same default `order_state`.
+      let candidates = self.entry_facade_candidates(plan);
+      if candidates.is_empty() {
+        return false;
+      }
+      let import_edges = self.entry_facade_import_edges(chunk_graph, used_symbol_refs, order_state);
+      if !self.create_order_wrap_entry_facades(chunk_graph, candidates, &import_edges) {
         return false;
       }
       chunk_graph.sort_chunk_modules(self.link_output, self.options);
@@ -145,12 +157,67 @@ impl GenerateStage<'_> {
       &self.link_output.symbol_db,
     );
     self.place_order_wrap_modules(chunk_graph, plan, order_state);
-    self.create_order_wrap_entry_facades(chunk_graph, analysis);
+    // Restoring runs before everything below it. A restored facade is an optimizer-removed chunk
+    // brought back to life, so it changes both the live-chunk count the runtime placement reads
+    // and the edge set the facade decision reads — and `compute_chunk_imports` skips removed
+    // chunks, so asking first would hide the edge from a revived facade to the chunk hosting its
+    // implementation, leaving an entry that shares that chunk with an inline trigger the restored
+    // facade's load then fires. Restoring is not itself a candidate for the necessity gate: it
+    // exists because a chunk can host only one entry's top-level trigger, so an entry merged into
+    // another entry's chunk has no trigger at all without its own file.
     self.restore_order_wrap_entry_facades(chunk_graph, plan);
-    self.ensure_runtime_module_for_order_wraps(chunk_graph, order_state);
+    // Placement runs before the facade decision so that decision can ask the real link computation
+    // which chunks import which: the order wrappers demand runtime helpers, and resolving a
+    // depended symbol to its chunk requires the runtime module to already sit in one. The merge
+    // re-proof this normalizes for is deferred to `fold_runtime_chunk_after_order_lowering` below,
+    // because it counts the runtime chunk's consumers and must see the final facade topology.
+    let fold_runtime_chunk = self.ensure_runtime_module_for_order_wraps(chunk_graph);
+    // Refresh the namespace facts against the lowered order state before querying the edges.
+    // `finalize_chunk_plan` re-runs this once more after lowering anyway; doing it here too means
+    // the query sees the namespaces the final link pass will see, instead of the provisional
+    // pre-lowering set, so a namespace an import overlay demands cannot become an edge the query
+    // missed. Facades do not move modules between chunks, so the result is the same before and
+    // after they are created.
+    self.finalized_module_namespace_ref_usage(chunk_graph, order_state);
+    // Collecting the candidates reads only the module graph, so it can run before the edges and
+    // spare the query entirely when no entry could need a facade in the first place.
+    let candidates = self.entry_facade_candidates(plan);
+    if !candidates.is_empty() {
+      let import_edges = self.entry_facade_import_edges(chunk_graph, used_symbol_refs, order_state);
+      self.create_order_wrap_entry_facades(chunk_graph, candidates, &import_edges);
+    }
+    if fold_runtime_chunk {
+      self.fold_runtime_chunk_after_order_lowering(chunk_graph, order_state);
+    }
     chunk_graph.sort_chunk_modules(self.link_output, self.options);
     self.renumber_live_chunks(chunk_graph);
     true
+  }
+
+  /// The chunk->chunk static import edges the entry-facade decision is taken against.
+  ///
+  /// Computed from the fully lowered order state through the same cross-chunk link pass that
+  /// produces the final edges, so the decision reads facts rather than a re-derived approximation
+  /// of them. See [`GenerateStage::lowered_static_import_edges`].
+  fn entry_facade_import_edges(
+    &mut self,
+    chunk_graph: &ChunkGraph,
+    used_symbol_refs: &UsedSymbolRefsBuilder,
+    order_state: &OrderWrapState,
+  ) -> IndexVec<ChunkIdx, FxHashSet<ChunkIdx>> {
+    if self.options.code_splitting.is_disabled() {
+      // A single-chunk build has no other chunk to load the entry's implementation, and
+      // `create_order_wrap_entry_facades` bails out on it anyway.
+      return index_vec![FxHashSet::default(); chunk_graph.chunk_table.len()];
+    }
+    let final_esm_init_metadata =
+      self.compute_wrapped_esm_init_metadata(&self.ast_table, chunk_graph, order_state);
+    self.lowered_static_import_edges(
+      chunk_graph,
+      used_symbol_refs,
+      order_state,
+      &final_esm_init_metadata,
+    )
   }
 
   fn place_order_wrap_modules(
@@ -174,17 +241,16 @@ impl GenerateStage<'_> {
     }
   }
 
-  fn create_order_wrap_entry_facades(
-    &self,
-    chunk_graph: &mut ChunkGraph,
-    analysis: &OrderAnalysis,
-  ) -> bool {
+  /// Entries whose inline `init_E()` trigger might have to move into a facade — an order-wrapped
+  /// entry, or an interop-wrapped one, whether or not anything imports it.
+  ///
+  /// Reads only the module graph, never the chunk topology, so the caller can collect these before
+  /// computing the import edges and skip that query entirely when there is nothing to decide.
+  fn entry_facade_candidates(&self, plan: &OrderWrapPlan) -> Vec<ModuleIdx> {
     if self.options.code_splitting.is_disabled() {
-      return false;
+      return vec![];
     }
-    let plan = &analysis.plan;
-
-    let mut entries_to_split = plan
+    let mut candidates = plan
       .modules()
       .filter(|module_idx| self.link_output.entries.contains_key(module_idx))
       .collect_vec();
@@ -195,7 +261,7 @@ impl GenerateStage<'_> {
       if !meta.is_included {
         continue;
       }
-      entries_to_split.extend(module.import_records.iter().filter_map(|rec| {
+      candidates.extend(module.import_records.iter().filter_map(|rec| {
         if !matches!(rec.kind, ImportKind::Import | ImportKind::Require) {
           return None;
         }
@@ -206,19 +272,30 @@ impl GenerateStage<'_> {
         .then_some(importee_idx)
       }));
     }
+    candidates.extend(self.link_output.entries.keys().copied().filter(|entry_module_idx| {
+      !matches!(self.link_output.metas[*entry_module_idx].wrap_kind(), WrapKind::None)
+    }));
+    candidates
+  }
 
-    // Move an interop entry trigger to a facade when another chunk imports its implementation.
-    // Wrap-all mode computes no prediction and splits unconditionally. The wrapping policy is
-    // carried on the analysis (decided once in `analyze_execution_order`) rather than re-read here.
-    let on_demand = analysis.on_demand;
+  fn create_order_wrap_entry_facades(
+    &self,
+    chunk_graph: &mut ChunkGraph,
+    facade_candidates: Vec<ModuleIdx>,
+    import_edges: &IndexVec<ChunkIdx, FxHashSet<ChunkIdx>>,
+  ) -> bool {
+    if self.options.code_splitting.is_disabled() {
+      return false;
+    }
+
     let mut imported_chunks = FxHashSet::default();
-    for (chunk_idx, importee_chunks) in analysis.import_edges.iter_enumerated() {
+    for (chunk_idx, importee_chunks) in import_edges.iter_enumerated() {
       imported_chunks
         .extend(importee_chunks.iter().copied().filter(|importee| *importee != chunk_idx));
     }
     // A dynamic import evaluates its target's chunk, so an inline entry trigger hosted there
     // would run the entry's whole program during that load (e.g. a manual group placing a
-    // dynamic target next to an entry). Predicted static edges cannot see these loads; collect
+    // dynamic target next to an entry). The static edge set cannot see these loads; collect
     // the cross-chunk dynamic-import targets directly.
     let mut dynamic_target_modules_by_chunk: FxHashMap<ChunkIdx, FxHashSet<ModuleIdx>> =
       FxHashMap::default();
@@ -230,7 +307,15 @@ impl GenerateStage<'_> {
       }
       let importer_chunk = chunk_graph.module_to_chunk[module.idx];
       for rec in &module.import_records {
-        if rec.kind != ImportKind::DynamicImport {
+        // Only records the emitted code still executes count as loads. Tree shaking flags a
+        // dynamic record whose entry it dropped as `DeadDynamicImport`, and the finalizer rewrites
+        // exactly those to an inert `Promise.resolve().then(...)` stub that loads nothing — so a
+        // dead record must not force the split. The importee can still be included through a
+        // static route, which is precisely when it may share an entry's chunk. Same exclusion as
+        // `dynamic_already_loaded`.
+        if rec.kind != ImportKind::DynamicImport
+          || rec.meta.contains(ImportRecordMeta::DeadDynamicImport)
+        {
           continue;
         }
         let Some(importee_idx) = rec.resolved_module else { continue };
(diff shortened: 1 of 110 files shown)


hyfdev:
> **Model review** — generated by Claude Fable 5 (`claude-fable-5`); posted by the author for reviewer reference.

I reviewed this PR focusing on whether the fix approach is sound — specifically, whether gating the facade decision on a pre-facade run of the real link computation can ever miss a needed facade (the dangerous direction; an unnecessary facade is only a wasted file). Conclusion: **the approach is sound and the implementation matches its stated invariants.** What was independently verified, keyed to code:

**The scoped query's rollback is complete.** The only shared-state mutation inside `compute_cross_chunk_link_state` is `assign_symbol_chunk_ownership` writing `symbol_data.chunk_idx`; `compute_chunk_imports` is `&self`, and the parallel phase of `collect_depended_symbols` writes only pass-local structures. Across the codebase the only writers of `chunk_idx` are the journaled assign, the rollback itself, and `clear_module_symbol_chunk_indices` (runtime module only, runs before the query). No interior mutability on these paths, and `canonical_ref_for` is the non-compressing `&self` variant. Reverse journal replay restores exact previous values.

**The pre-facade snapshot argument holds.** Enumerating the edge sources in `compute_chunk_imports` that a new facade can trigger: the entry-block depended symbols ran identically for the implementation chunk in the query (it was the EntryPoint chunk pre-split), the entry-in-other-chunk edge targets only the implementation chunk, and the facade side-effect block duplicates edges the per-module loop already produced. `Chunk::new` gives a facade empty `imports_from_other_chunks`, and synthetic `init_*` statements stay keyed to the implementation chunk. So the only genuinely new inbound edge post-split is facade→implementation, whose entry's decision is already "split" — no other entry's predicate can flip after the snapshot, and no fixpoint is needed.

**No dangerous input skew between the query and the final pass.** Between them the state changes are facade creation (above), `sweep_unused_runtime_module` (gated on no order-wrap runtime demand; removal-only, so final edges can only shrink — worst case an unnecessary facade, never a missing one), and the re-runs of `find_entry_level_external_module` / `finalized_module_namespace_ref_usage` (idempotent relative to the hoisted runs: facades hold no modules, and both read only module-level facts). `compute_wrapped_esm_init_metadata` reads only `module_to_chunk`/liveness/metas/order state — all facade-invariant — so the query's sealed metadata equals the final pass's.

**The restore-before-query ordering is load-bearing, as its comment claims.** `compute_chunk_imports` skips chunks marked plain `Removed`, so restoring after the query would hide a revived facade's edge into a shared implementation chunk; `RemovedWithPreservedExports` facades still contribute their edges, which covers the merged-entry topologies.

**Tests pin both directions.** The shape tests assert no facade for self-contained lazy routes and a retained facade when another chunk imports the entry's chunk, each under wrap-all and on-demand; the new fixture executes under Node asserting `e`'s program does not run while `b` loads. Spot-checked snapshots (`dynamic_target_in_entry_chunk`, `m4_dynamic_facade_race`) confirm dropped facades collapse into their implementation chunk (which takes over the entry filename) while needed ones remain.

**Non-blocking nit.** In the `plan.is_empty()` branch, `entry_facade_import_edges` runs a full metadata + link pass even when there are no wrapped entries at all (the common on-demand no-hazard path); computing `facade_candidates` first and skipping the query when empty would avoid it. The same branch skips the three hoisted passes — vacuously safe today (restore is plan-filtered; the namespace pass already ran with the same default state), but that is an implicit invariant worth a one-line comment.

The extra strict-mode link pass (+~6% wrap-all on the 120-route benchmark) is a reasonable trade for the 45% chunk-count reduction and is honestly flagged in the PR body.

hyfdev:
Both nits addressed in f2b2aa2ce.

**Skipping the query when nothing can need a facade.** `entry_facade_candidates` is split out of `create_order_wrap_entry_facades`. It reads only the module graph, never the chunk topology, so it now runs before the edges and the metadata + link pass is skipped outright when the set is empty — the case the nit names: an empty plan on a pure-ESM graph, where the only candidates would be interop-wrapped entries.

One correction to the motivation, though: **I could not measure a win.** My first attempt benchmarked the 120-route app from the PR body, which turns out never to reach that path — `ROLLDOWN_ORDER_DEBUG=1` reports `one-shot plan: 145 modules` there, so it takes the lowering branch and the candidate set is non-empty. On an app that does reach it (600 modules, pure ESM, no hazard, confirmed `one-shot plan: 0 modules`), the difference does not clear run-to-run noise:

| | median | min |
| --- | --- | --- |
| with the skip | 134.3ms | 101.1ms |
| without | 134.9ms | 106.2ms |

With an empty plan and a single chunk, the link pass is cheap next to parsing and rendering 600 modules. So the change stands on doing strictly less work and on making the branch's reasoning explicit, not on a measured speedup.

**The implicit invariant.** Now stated in the empty-plan branch: it may skip the three passes hoisted ahead of the query because an empty plan makes each one a no-op — nothing demands an order-wrap runtime helper, restoring is filtered by the plan, and `finalize_chunk_plan` has already run the namespace pass against this same default `order_state`.

Both changes are refactor-only: full fixture suite 1896 passed / 0 failed, zero snapshot drift.

hyfdev:
* **#10454** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10454?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10449** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10449?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10441** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10441?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10441?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10441).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

IWANABETHATGUY (approved):

crates/rolldown/src/stages/generate_stage/order_wrapping.rs
```diff
@@ -205,19 +267,30 @@ impl GenerateStage<'_> {
...
-    for (chunk_idx, importee_chunks) in analysis.import_edges.iter_enumerated() {
+    for (chunk_idx, importee_chunks) in import_edges.iter_enumerated() {
       imported_chunks
         .extend(importee_chunks.iter().copied().filter(|importee| *importee != chunk_idx));
     }
     // A dynamic import evaluates its target's chunk, so an inline entry trigger hosted there
     // would run the entry's whole program during that load (e.g. a manual group placing a
-    // dynamic target next to an entry). Predicted static edges cannot see these loads; collect
+    // dynamic target next to an entry). The static edge set cannot see these loads; collect
     // the cross-chunk dynamic-import targets directly.
```
IWANABETHATGUY, line 299: **[P2] A dead dynamic import still counts as a load and forces the facade this PR exists to drop**

This scan walks raw `import_records`, so it also counts records tree shaking has already killed. When a dynamic entry is dropped (`is_dynamic_entry_alive`), its creating records get `ImportRecordMeta::DeadDynamicImport` and the finalizer lowers them to an inert stub — `Promise.resolve().then(() => /* @__PURE__ */ Object.freeze({ __proto__: null }))` (`module_finalizers/mod.rs:1539`) — which can never load the importee's chunk. The importee can still be `is_included` through a static route, which is exactly the case where it shares an entry's chunk: the dead record then keeps the facade.

Reproduced on this branch under wrap-all: entry `e` statically imports `t.js` (non-inlinable export), entry `b`'s only reference to it is a top-level `import('./t.js')` with an unused result, `treeshake.moduleSideEffects: false`. The output ships an empty `e.js` facade plus an `e2.js` implementation chunk, while `b.js` contains only the stub — nothing in the emitted code can load `e2.js` except entering `e`.

Fix verified locally (full integration suite 1898 passed / 0 failed, no snapshot drift outside the new fixture), following the precedent in `dynamic_already_loaded.rs:347`:

```rust
      for rec in &module.import_records {
        // A dead dynamic import lowers to an inert `Promise.resolve().then(...)` stub, so it can
        // never load the importee's chunk — only records the emitted code still executes count as
        // loads. The importee can still be included here through a static route (which is exactly
        // when it may share the entry's chunk).
        if rec.kind != ImportKind::DynamicImport
          || rec.meta.contains(ImportRecordMeta::DeadDynamicImport)
        {
          continue;
        }
```

Skipping flagged records is safe in the dangerous direction (a missed facade) because the flag is a direct contract with the finalizer: flag ⇒ stub ⇒ no load.

<details>
<summary>Drop-in regression fixture: <code>function/experimental/strict_execution_order/dead_dynamic_target_in_entry_chunk/</code></summary>

`_config.json`:

```json
{
  "_comment": "b.js's only dynamic import of t.js is a top-level pure dynamic import with an unused result, so tree shaking drops the dynamic entry and the finalizer emits an inert `Promise.resolve().then(...)` stub — the emitted code can never load t.js's chunk through it. t.js stays included because entry e uses it statically, so it shares e's chunk. A dead record must not count as a dynamic-import load of e's chunk, so e keeps its inline trigger and no facade is emitted.",
  "configName": "wrap-all",
  "config": {
    "input": [
      { "name": "b", "import": "./b.js" },
      { "name": "e", "import": "./e.js" }
    ],
    "strictExecutionOrder": true,
    "treeshake": {
      "moduleSideEffects": false
    }
  },
  "configVariants": [{ "_configName": "on-demand", "onDemandWrapping": true }]
}
```

`e.js`:

```js
import { t } from './t.js';

console.log('E', t);
```

`t.js` (the `String(...)` call keeps the export from being constant-inlined, which would otherwise drop `t.js` entirely and dodge the scan):

```js
export const t = String(globalThis.__t ?? 't');
```

`b.js`:

```js
import('./t.js');

console.log('B');
```

Before the fix, the wrap-all snapshot contains `e.js` = `import { t as init_e } from "./e2.js"; init_e();` plus an `e2.js` implementation chunk; after, `e.js` hosts `init_t`/`init_e` inline and `e2.js` is gone. `b.js` is byte-identical in both — only the stub.

</details>

One residual door, fine as a follow-up: a dynamic import whose whole *statement* is tree-shaken (e.g. inside a dropped function) is not flagged either and still counts here. A complete gate could be `!self.link_output.entries.contains_key(&importee_idx)` — a live dynamic record always keeps its target in the entry set, so a non-entry target implies every record to it is dead — but that rests on a derived global invariant rather than a finalizer contract, so it wants its own fixture before shipping.

crates/rolldown/src/stages/generate_stage/order_wrapping.rs
```diff
@@ -245,20 +318,30 @@ impl GenerateStage<'_> {
...
-    }));
+    // An entry's `init_E()` trigger may stay inline at the top of the chunk hosting `E` exactly
+    // while that chunk is loaded only to enter `E`. Once anything else can load it — another chunk
+    // imports it, or it also hosts some other chunk's dynamic-import target — the trigger has to
+    // move into a facade or `E`'s program runs during that unrelated load.
+    //
+    // This is the same question for an order-wrapped entry and for an interop-wrapped one, so all
+    // candidate sources are gated by it. It used to be asked only of interop entries, and only in
+    // on-demand mode; an order-wrapped entry always split, which cost one extra chunk per entry
+    // (including every dynamic entry) even when nothing but the entry itself could load the chunk.
```
IWANABETHATGUY, line 342: **[P2] `internal-docs/code-splitting/design.md` now contradicts this gate**

AGENTS.md is explicit — "**Keep in sync:** If your changes affect a doc, update it in the same change" — and two spots in `internal-docs/code-splitting/design.md` still describe the rule this PR replaces:

- The **Entry-trigger facades** bullet under *Conservative decisions* says the split is asked only of an *interop-wrapped* entry, gated on *predicted* static imports, "(unconditionally in wrap-all mode, which has no predicted edges)". All three clauses are exactly what this PR removes: the gate now covers order-wrapped entries too, in both strict modes, and reads the real lowered edges (`lowered_static_import_edges`) instead of a prediction.
- The **Trigger placement** table row "user or dynamic entry activation → entry facade prologue": the activation now lives in the entry chunk's prologue, which is a *facade* only when another chunk can load the implementation chunk.

The sentence directly under that table — "A trigger must never sit inline in a chunk body that other chunks can evaluate as a dependency; that is the facade rule's content." — is already the exact post-PR rule, so only the mechanism description around it needs updating.

<details>
<summary>Suggested wording for both spots</summary>

*Conservative decisions* bullet:

```md
- **Entry-trigger facades**: an inline entry trigger fires whenever its chunk is evaluated,
  so an entry whose chunk anything else *can* load — order-wrapped or interop-wrapped, in
  both strict modes — moves its trigger to a facade. The question is answered by running the
  real cross-chunk link computation against the fully lowered order state
  (`lowered_static_import_edges`), not a prediction, so an entry whose chunk only it can
  load keeps its trigger inline and costs no extra file. "Loads" covers both static imports
  from other chunks and cross-chunk dynamic imports of any other module hosted in the entry
  chunk — e.g. a manual group placing a dynamic target next to an entry, where the
  `import()` evaluates the entry chunk. A dynamic import of the entry module itself must run
  its program, and a dead dynamic import lowers to an inert stub, so neither forces the
  split; any other surviving record counts as a possible load even if never executed.
```

(The dead-dynamic-import clause assumes the sibling finding on this review lands; drop it otherwise.)

*Trigger placement* table row:

```md
| user or dynamic entry activation | entry chunk prologue (a facade only when other chunks can load the implementation chunk) | `create_order_wrap_entry_facades` / `restore_order_wrap_entry_facades` |
```

</details>

IWANABETHATGUY (commented):
> **Model review** — findings generated and verified by Claude Fable 5 (`claude-fable-5`); posted by the reviewer.

Two P2 findings, both verified against f2b2aa2 by reproducing locally: a residual case of the over-split this PR removes (dead dynamic imports still count as loads — reproduction, one-line fix, and a drop-in regression fixture in the inline comment), and a doc-sync obligation (`internal-docs/code-splitting/design.md` still documents the unconditional/predicted-edges rule this PR replaces). Neither changes the verdict from the earlier review — the approach and its soundness argument hold.

crates/rolldown/src/stages/generate_stage/order_wrapping.rs
hyfdev: Fixed in 7b5fe959b. I verified the contract at both ends before taking it rather than on the reproduction alone: `include_statements.rs:397` sets the flag on exactly the records whose dynamic entry tree shaking dropped, and `module_finalizers/mod.rs:1539` rewrites exactly those to the stub unconditionally once the record resolves. So flag ⇒ stub ⇒ no load holds as a contract, not as an observation, which is what makes skipping them safe in the dangerous direction.

One addition to the fixture you supplied: it is in as `dead_dynamic_target_in_entry_chunk`, but on its own it only fails CI through snapshot drift, since a local run rewrites the snapshot. So the case is also asserted in `packages/rolldown/tests/behaviors/strict-execution-order-entry-facade.test.ts`, which hard-fails on the facade count. Confirmed it discriminates: with the gate reverted, an `e2.js` implementation chunk reappears.

The residual door is left open as you suggested, with your reasoning recorded in the commit message — the entry-set gate would cover it but rests on a derived global invariant rather than a finalizer contract, so it wants its own fixture first.

crates/rolldown/src/stages/generate_stage/order_wrapping.rs
hyfdev: Updated in 7b5fe959b, using your wording for both spots. The dead-dynamic-import clause is kept, since the sibling finding landed in the same commit.

You were right that the sentence under the table already stated the post-PR rule — only the mechanism description around it needed to move.

hyfdev:
Rebased onto `main` (12 commits). Both review findings are addressed in 7b5fe959b — replies in the inline threads.

## The one real conflict

34 of the 35 conflicted files were snapshots. The one that mattered was #10414's `fold_runtime_chunk_after_order_lowering`, and it was a genuine ordering conflict rather than a signature clash: its comment states it must run *after* facade restoration, because it counts the runtime chunk's consumers, while this PR needs the runtime module *placed* in a chunk before the edge query, since resolving a depended symbol to its chunk requires it.

Resolved by splitting the two. `ensure_runtime_module_for_order_wraps` now does placement only and returns whether the merge re-proof is still owed; `apply_order_wraps` discharges it once the facade topology is final. Both preconditions hold, and the fold now also sees *created* facades, not only restored ones.

One further ordering change fell out of it: `restore_order_wrap_entry_facades` has to run before placement, not just before the query. A revived facade changes the live-chunk count, and in the "runtime not yet in a chunk, exactly one live chunk" case that count is what selects the placement branch.

## `late_order_wrapping_revalidates_output_file` changed shape

Worth a look, since it is the one existing assertion this PR invalidates rather than preserves.

The test built `issue_4782` with `output.file` and asserted the multi-chunk diagnostic. That graph is no longer multi-chunk after order lowering — the extra chunk was the unconditional entry facade — so `output.file` now succeeds there. That is the fix working, not a regression: a build that used to be rejected under `strictExecutionOrder` now produces a single file.

The invariant is still worth pinning, so the test moved to a dedicated fixture (944a3ea4d): a dynamically imported `lib.js` the optimizer merges into the user entry's chunk, whose facade `restore_order_wrap_entry_facades` revives because a chunk can host only one entry's top-level trigger. Single-chunk when `output.file` is first validated, multi-chunk only afterwards.

Note for whoever reviews the fixture: `lib.js` importing a side-effectful `probe.js` is deliberate, not filler. My first attempt reused `optimization/chunk_merging/dynamic_entry_merged_in_user_defined_entry`, which passed here but broke #10454 upstack — that PR skips wrapping a merged-away entry whose load closure holds nothing order-sensitive, which is exactly that fixture. The `probe.js` side effect keeps the closure order-sensitive, so the fixture does not depend on an entry being wrapped when wrapping buys nothing. Verified green on this branch and on #10454.

## Stack

#10449 and #10454 were restacked and force-pushed with the same base. #10454 had no source conflicts, only snapshots.

- Rust integration suite: 1899 passed / 0 failed, on each of the three branches.
- Node suite: 917 passed / 0 failed.
- `cargo fmt --check` and `clippy --deny warnings` clean.

hyfdev:
### Merge activity

* **Jul 26, 4:51 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 26, 4:51 PM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 26, 4:57 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10442, opened 2026-07-25
fix(preserve-modules): don't assume every chunk mirrors a module

hyfdev:
## Summary

`preserveModules` emits one chunk per module, and the export-naming pass in `compute_cross_chunk_links` relies on that: to decide whether a chunk export should be named `default`, it reaches for `chunk.entry_module(&self.link_output.module_table).unwrap()`.

Synthetic chunks break the assumption. `strictExecutionOrder` splits the shared `rolldown-runtime` chunk that holds `__esmMin`; that chunk mirrors no user module and is `ChunkKind::Common`, so the lookup returns `None`:

```
thread 'rolldown-worker' panicked at crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs:1299:65:
called `Option::unwrap()` on a `None` value
```

Any `preserveModules` build with `strictExecutionOrder` and more than one emitted chunk hits it — a two-module entry is enough:

```js
// dep.js
export default function dep() { return 'D'; }
// main.js
import dep from './dep.js';
console.log(dep());
```

```js
await (await rolldown({ input: './main.js' }))
  .generate({ format: 'esm', preserveModules: true, strictExecutionOrder: true }); // panics
```

## Fix

A chunk with no entry module has no module default export to rename, so ask the question instead of asserting the answer — `is_some_and` in place of `unwrap`.

Behaviour for chunks that *do* mirror a module is unchanged: the condition is identical whenever `entry_module()` is `Some`, and no existing snapshot moves. The only chunks reaching the new branch are ones that previously panicked.

The sibling `entry_module(...).expect(...)` a little earlier in the same function is already guarded by an `if let ChunkKind::EntryPoint { .. }`, so it is sound and left alone.

## Tests

New fixture `misc/preserve_modules/strict_execution_order_runtime_chunk`, run under both wrap-all and `onDemandWrapping`. It builds a static import with a default export plus a dynamic import, so the output has a per-module file, a `default` export to name, and the split runtime chunk. `_test.mjs` asserts the modules evaluate in source order.

Without the source change the fixture panics at the line above; with it, it passes. `dep.js` still exports `default`, so the naming path this code exists for is covered rather than merely bypassed.

Full Rust integration suite: 1892 passed, with the only 4 failures (`cjs_module_lexer_compat` ×3, `npm_packages/util_deprecate`) failing identically on unpatched `main` in the same tree — they need npm packages this worktree does not have installed.

## Relationship to #10441

Independent, and this PR is based on `main` rather than stacked. The panicking chunk is the synthetic runtime chunk, which #10441 does not touch; the panic reproduces identically with and without that PR, and the fix is in a different function.

One thing reviewers should know: this fixture's snapshot shows `preserveModules` currently emitting *two* files for an order-wrapped entry — `main.js` plus `_virtual/main.js` — because the entry facade splits unconditionally. That contradicts the one-file-per-module contract, and it is exactly what #10441 fixes. Whichever lands second will regenerate this snapshot with the facades gone. That is a snapshot update, not a conflict.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
+++ b/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
@@ -1295,9 +1295,15 @@ impl GenerateStage<'_> {
         // single default export per chunk must be named `default`. Otherwise use the
         // `default_export_ref` representative name. The `&&` keeps the `entry_module`
         // lookup guarded behind the `preserve_modules` check.
+        //
+        // `preserve_modules` emits one chunk per module, so a chunk normally carries the module it
+        // mirrors. Synthetic chunks are the exception: the shared `rolldown-runtime` chunk that
+        // strict execution order splits out mirrors no user module and is `ChunkKind::Common`, so
+        // it has no entry module and none of its exports can be a module's default export.
         let base = if self.options.preserve_modules
-          && chunk.entry_module(&self.link_output.module_table).unwrap().default_export_ref
-            == *chunk_export
+          && chunk
+            .entry_module(&self.link_output.module_table)
+            .is_some_and(|entry_module| entry_module.default_export_ref == *chunk_export)
         {
           CompactStr::new_const("default")
         } else {
--- a/crates/rolldown/tests/rolldown/misc/preserve_modules/strict_execution_order_runtime_chunk/_config.json
+++ b/crates/rolldown/tests/rolldown/misc/preserve_modules/strict_execution_order_runtime_chunk/_config.json
@@ -1,0 +1,19 @@
+{
+  "configName": "wrap-all",
+  "config": {
+    "input": [
+      {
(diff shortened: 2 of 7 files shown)


shulaoda (approved):

shulaoda:
### Merge activity

* **Jul 26, 6:49 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 26, 6:49 AM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 26, 6:54 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10443, opened 2026-07-25
fix(filter): treat missing hook inputs as non-matches, not panics

IWANABETHATGUY:
## What

`filter_expr_interpreter` unwrapped `id` and `code` with `expect`, but hooks pass `None` for the inputs they don't have:

| hook | `id` | `code` | `moduleType` |
|---|---|---|---|
| `resolveId` | specifier | **None** | **None** |
| `load` | id | **None** | **None** |
| `transform` | id | code | moduleType |
| `renderChunk` | **None** | code | **None** |

Each hook's `filter` is typed as an arbitrary `TopLevelFilterExpression[]`, and binding validation only rejects `importerId` (`assertNoImporterId`). So a leaf reading an input its hook never supplies is reachable, and panics rather than simply not matching:

```js
// plugin with a renderChunk filter -- `id` is None here
{ renderChunk: { filter: [include(id(/\.css$/))] } }
// => panics: `id` should not be none
```

This makes `Code`/`Id`/`Query` report "no match" when their input is unavailable, matching what `ImporterId` already did for `None`:

```rust
// When importer_id is None (e.g., entry files), return false (no match)
match importer_id {
  Some(importer) => id_pattern.test(importer, &StringOrRegexMatchKind::Id(cwd)),
  None => false,
}
```

## Why "no match" rather than an error

Every case that changes behaviour is one that panics today, so nothing that currently works can regress. Rejecting these at bindingify time (the way `importerId` and string-`id` already are) is defensible too, but that is a wider validation surface — it would also have to decide what to do about `moduleType` in a `load` filter, which is silently non-matching today. This change fixes the crash without expanding validation; tightening it can follow separately.

## Tests

Two regression tests covering both directions — a missing `id` (`renderChunk`) and a missing `code` (`resolveId`/`load`). Both fail with the original `expect` panic and pass with the fix.

### DIFF
--- a/crates/rolldown_utils/src/filter_expression.rs
+++ b/crates/rolldown_utils/src/filter_expression.rs
@@ -34,6 +34,11 @@ pub enum FilterExprKind {
   Exclude(FilterExpr),
 }
 
+/// Every leaf is total over the inputs a hook can supply. Hooks pass `None` for the
+/// inputs they don't have — `renderChunk` has no `id`, `resolveId`/`load` have no
+/// `code` — and each hook's filter is typed as an arbitrary `TopLevelFilterExpression[]`,
+/// so a leaf reading an input its hook never supplies is reachable. Such a leaf reports
+/// "no match" instead of panicking.
 pub fn filter_expr_interpreter<'a>(
   expr: &FilterExpr,
   id: Option<&'a str>,
@@ -54,10 +59,12 @@ pub fn filter_expr_interpreter<'a>(
       !filter_expr_interpreter(inner, id, code, module_type, importer_id, cwd, ctx)
     }
     FilterExpr::Code(pattern) => {
-      pattern.test(code.expect("`code` should not be none"), &StringOrRegexMatchKind::Code)
+      // When code is None (e.g. the `resolveId`/`load` hooks), return false (no match)
+      code.is_some_and(|code| pattern.test(code, &StringOrRegexMatchKind::Code))
     }
     FilterExpr::Id(id_pattern) => {
-      id_pattern.test(id.expect("`id` should not be none"), &StringOrRegexMatchKind::Id(cwd))
+      // When id is None (e.g. the `renderChunk` hook), return false (no match)
+      id.is_some_and(|id| id_pattern.test(id, &StringOrRegexMatchKind::Id(cwd)))
     }
     FilterExpr::ImporterId(id_pattern) => {
       // When importer_id is None (e.g., entry files), return false (no match)
@@ -79,8 +86,10 @@ pub fn filter_expr_interpreter<'a>(
       ctx,
     ),
     FilterExpr::Query(key, value) => {
+      // When id is None (e.g. the `renderChunk` hook) there is no query to test
+      let Some(id) = id else { return false };
       if ctx.parsed_url_cache.is_none() {
-        let query_string = get_query(id.expect("`id` should not be none"));
+        let query_string = get_query(id);
         let cache = form_urlencoded::parse(query_string.as_bytes())
           .into_iter()
           .map(|(k, v)| (k.to_string(), v))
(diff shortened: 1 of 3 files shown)


hyfdev:
For the long term, I feel more correct direction is to have domain-sepcific matchers like "moduleId(), chunkId()", `moduleId()` in `renderChunk` will simply get ignored.

crates/rolldown_utils/src/filter_expression.rs
```diff
@@ -79,8 +86,10 @@ pub fn filter_expr_interpreter<'a>(
       ctx,
     ),
     FilterExpr::Query(key, value) => {
+      // When id is None (e.g. the `renderChunk` hook) there is no query to test
+      let Some(id) = id else { return false };
```
hyfdev: `query(k, false)` means "the id does not carry `?k`", so on every hook that supplies an `id` it is `true` when the key is absent. This early return makes it `false` on `renderChunk`, the one place there is no query at all — while the identical `not(query(k, true))` returns `true` there. That also leaves `query(k, true)` and `query(k, false)` both false at once, a state no real id can produce.

Replacing the early return with an empty query map — `let query_string = id.map(get_query).unwrap_or("")` — flips exactly that one cell; the `true`, string, and regex variants still return false, and the committed tests still pass.

`missing_id_is_a_non_match_not_a_panic` covers `Boolean(true)`, and the `false` variant is the one that behaves inconsistently. Nothing here blocks this approval.

crates/rolldown_utils/src/filter_expression.rs
```diff
@@ -340,4 +349,47 @@ mod test {
       ".",
     ));
   }
+
+  #[test]
+  fn missing_id_is_a_non_match_not_a_panic() {
```
hyfdev, line 358: For an `exclude`-only or negated filter, "no match" comes out as "run the handler for everything". `renderChunk: { filter: [exclude(id(/vendor/))] }` and `[include(not(id(/vendor/)))]` both return `true` for every chunk, so the handler runs on exactly the chunks the author declared it should skip; on `transform`, where the id exists, the same `exclude` correctly returns `false`. A handler has no reason to re-check what its own filter already expressed, so this direction can produce wrong output rather than merely doing nothing.

These tests assert on leaves through `filter_expr_interpreter`, so nothing pins the hook-level outcome that `filter_exprs_interpreter`'s `include_count == 0` rule produces, and nothing exercises the JS route that makes the crash reachable in the first place. `packages/rolldown/tests/fixtures/plugin/render-chunk/filter/` already asserts a handler ran 0 times and `resolve-id/filter-string-id-error/` is the same shape, so a sibling fixture for `[include(id(/x/))]` would fail on `main` and pin the user-visible outcome.

#10412 closed the same loophole class — a composable-form leaf the hook cannot support — by rejecting at bindingify time, because it "silently never matches the way a user expects". There is now no error, warning, or trace for `renderChunk` + `id`, which is what the follow-up validation the description mentions would restore. None of this blocks this approval.

hyfdev (approved):
Every behaviour change here is confined to inputs that panicked on `main` — the three rewritten arms are identical wherever the input is `Some` — so nothing that works today can regress. Both new tests panic against the base rather than merely failing an assertion. The `Query` half needed its own check, since `missing_id_is_a_non_match_not_a_panic` aborts at its first assertion; it panics on the base too.

I left two non-blocking notes: `query(k, false)` is a negative predicate that this rule makes wrong on `renderChunk`, and for `exclude`-only or negated filters "no match" means the handler runs for everything rather than nothing.

This PR is what establishes "a leaf whose input the hook does not supply is a non-match", and `docs/apis/plugin-api/hook-filters.md` is where a plugin author would look for that. The page still omits `renderChunk` entirely and marks only `importerId` as hook-restricted while listing `id`, `code`, and `query` unrestricted.

The red Vite Test Ubuntu is the `playground/wasm` browser timeout, unrelated to this change — that job also failed twice on `main` in the last thirty runs.

PR #10445, opened 2026-07-26
fix(link): resolve circular star reexports as null

### LINKED ISSUE #9175: [Bug]: incorrect `CIRCULAR_REEXPORT` error

[REPL](https://repl.rolldown.rs/#eNp1j8EKwjAMhl+l5DKVWTcRDwPfZBetnShdI12VQem7m3RUGbJD4U+T72sToIMmwN1e9SgfA2cLza8uQXHZP9F5EUSHKKLoHPaikLszTRStay0fhXZAo6XB24rGSnHYr/meFBoa7146lsBEfiPlya/H5N98zZds/uuo3AHS8VjWpTzTLX6XUfZkNOUZarRP8Im2mAhC3tSvZSWrrVOyPkL8AOyAYVw=)

### What is expected?

No `CIRCULAR_REEXPORT` error happens

`CIRCULAR_REEXPORT` error happens

### System Info

```Shell
REPL 1.0.0-rc.16
```

### Any additional comments?

Found in test262 tests (https://github.com/rolldown/rolldown/pull/9174). The test was added by https://github.com/tc39/test262/pull/5017

Nic-Polumeyv:
Fixes #9175. A circular star-reexport path now resolves as null per ResolveExport instead of failing the build; CIRCULAR_REEXPORT still fires when a binding has no non-circular resolution.

### DIFF
--- a/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
+++ b/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
@@ -73,6 +73,7 @@ impl Hash for MatchImportKindNormal {
 }
 
 #[derive(Debug, PartialEq, Eq, Hash)]
+#[expect(clippy::box_collection)]
 pub enum MatchImportKind {
   // "sourceIndex" and "ref" are in use
   Normal(MatchImportKindNormal),
--- a/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
+++ b/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
@@ -38,6 +38,7 @@ struct ImportTracker {
 #[derive(Debug)]
 pub struct MatchingContext {
   tracker_stack: Vec<ImportTracker>,
+  circular_reexport: Option<(String, String)>,
 }
 
 impl MatchingContext {
@@ -987,7 +988,7 @@ impl BindImportsAndExportsContext<'_> {
     // from an empty stack (and the ambiguous-reexport recursion gets its own cloned stack), so
     // clearing before each call preserves behavior while reusing this allocation instead of
     // allocating a fresh `Vec` per import.
-    let mut matching_ctx = MatchingContext { tracker_stack: Vec::new() };
+    let mut matching_ctx = MatchingContext { tracker_stack: Vec::new(), circular_reexport: None };
     for (imported_as_ref, named_import) in &module.named_imports {
       let match_import_span = tracing::trace_span!(
         "MATCH_IMPORT",
@@ -1040,6 +1041,7 @@ impl BindImportsAndExportsContext<'_> {
         }
       }
       matching_ctx.tracker_stack.clear();
+      matching_ctx.circular_reexport = None;
       let ret = self.match_import_with_export(
         self.index_modules,
         &mut matching_ctx,
@@ -1052,7 +1054,15 @@ impl BindImportsAndExportsContext<'_> {
       );
       tracing::trace!("Got match result {:?}", ret);
       match ret {
-        MatchImportKind::Cycle => {}
+        MatchImportKind::Cycle => {
+          let (importer_id, imported_specifier) = matching_ctx
+            .circular_reexport
+            .take()
+            .expect("cycle should always have a circular reexport diagnostic");
+          self
+            .diagnostics
+            .push(BuildDiagnostic::circular_reexport(importer_id, imported_specifier));
+        }
         MatchImportKind::Ambiguous { symbol_ref, potentially_ambiguous_symbol_refs } => {
           let importee = self.index_modules[resolved_module_idx].stable_id().to_string();
 
@@ -1235,18 +1245,17 @@ impl BindImportsAndExportsContext<'_> {
 
     let mut ambiguous_results = vec![];
     let mut reexports = vec![];
-    let ret = loop {
+    let ret = 'tracking: loop {
       for prev_tracker in ctx.tracker_stack.iter().rev() {
         if prev_tracker.importer == tracker.importer
           && prev_tracker.imported_as == tracker.imported_as
         {
+          // ResolveExport treats an in-progress (module, binding) request as null.
           let importer_module = &index_modules[tracker.importer];
           let importer_id = importer_module.id().to_string();
           let imported_specifier = tracker.imported.to_string();
-          self
-            .diagnostics
-            .push(BuildDiagnostic::circular_reexport(importer_id, imported_specifier));
-          return MatchImportKind::Cycle;
+          ctx.circular_reexport = Some((importer_id, imported_specifier));
+          break 'tracking MatchImportKind::Cycle;
         }
       }
       ctx.tracker_stack.push(tracker.clone());
@@ -1294,7 +1303,10 @@ impl BindImportsAndExportsContext<'_> {
                 if let Some(resolved_module_idx) = rec.resolved_module {
                   let ambiguous_result = self.match_import_with_export(
                     index_modules,
-                    &mut MatchingContext { tracker_stack: ctx.tracker_stack.clone() },
+                    &mut MatchingContext {
+                      tracker_stack: ctx.tracker_stack.clone(),
+                      circular_reexport: None,
+                    },
                     ImportTracker {
                       importer: ambiguous_ref_owner.idx(),
                       importee: resolved_module_idx,
@@ -1366,6 +1378,12 @@ impl BindImportsAndExportsContext<'_> {
 
     tracing::trace!("ambiguous_results {:#?}", ambiguous_results);
     tracing::trace!("ret {:#?}", ret);
+    ambiguous_results.retain(|result| !matches!(result, MatchImportKind::Cycle));
+    let ret = if matches!(ret, MatchImportKind::Cycle) && !ambiguous_results.is_empty() {
+      ambiguous_results.pop().unwrap()
+    } else {
+      ret
+    };
 
     for ambiguous_result in &ambiguous_results {
       if *ambiguous_result != ret {
--- a/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
+++ b/crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
@@ -72,6 +72,21 @@ impl Hash for MatchImportKindNormal {
   }
 }
 
+/// Bookkeeping for a collected star branch so it can be promoted if the primary path turns out to
+/// be a cycle: the outer re-export hops accumulated up to the collection point, the branch's own
+/// export symbol, and where the branch was collected — the star-exporting module (and the name
+/// requested there) whose cached `resolved_exports` winner points into the cycle.
+#[derive(Debug)]
+struct StarBranchPromotion {
+  reexports_prefix: Vec<SymbolRef>,
+  export: SymbolRef,
+  star_module: ModuleIdx,
+  imported: Specifier,
+}
+
+/// Non-cycle star branches keyed by resolution, in collection order.
+type DedupedStarBranches = FxIndexMap<MatchImportKind, StarBranchPromotion>;
+
 #[derive(Debug, PartialEq, Eq, Hash)]
 #[expect(clippy::box_collection)]
 pub enum MatchImportKind {
@@ -1274,6 +1289,48 @@ impl BindImportsAndExportsContext<'_> {
     }
   }
 
+  /// Apply `ResolveExport`'s null-cycle rule to the collected star branches. Cycle branches are
+  /// dropped from the agreement set: `ResolveExport` skips a star branch that resolves to null
+  /// instead of failing the whole lookup. Walking a cycle also re-visits the same module once per
+  /// lap, which collects the surviving branches repeatedly — those duplicates would otherwise
+  /// show up as repeated exporters, and repeated files, in the ambiguity diagnostic. An
+  /// insertion-ordered map keeps source order for the "first branch wins" rule and the first
+  /// collection's promotion state while deduplicating in linear time, since a name supplied by N
+  /// distinct `export *` sources would otherwise make the scan quadratic.
+  ///
+  /// When the primary path was itself a cycle it resolves to null, so the binding is whatever the
+  /// remaining branches say: per `ResolveExport` the first non-null branch wins and the rest only
+  /// have to agree with it. The promoted branch's state is returned (its prefix already spliced
+  /// into the result) so the caller can realign the collection module's cached winner.
+  fn promote_first_surviving_star_branch(
+    ret: MatchImportKind,
+    ambiguous_results: Vec<(MatchImportKind, StarBranchPromotion)>,
+  ) -> (MatchImportKind, DedupedStarBranches, Option<StarBranchPromotion>) {
+    let mut deduped_ambiguous_results = DedupedStarBranches::default();
+    for (result, promotion_state) in ambiguous_results {
+      if !matches!(result, MatchImportKind::Cycle { .. }) {
+        deduped_ambiguous_results.entry(result).or_insert(promotion_state);
+      }
+    }
+
+    let (ret, promoted) = if matches!(ret, MatchImportKind::Cycle { .. }) {
+      if let Some((mut promoted, mut promotion)) = deduped_ambiguous_results.shift_remove_index(0) {
+        if let MatchImportKind::Normal(normal) = &mut promoted {
+          let mut reexports = std::mem::take(&mut promotion.reexports_prefix);
+          reexports.append(&mut normal.reexports);
+          normal.reexports = reexports;
+        }
+        (promoted, Some(promotion))
+      } else {
+        (ret, None)
+      }
+    } else {
+      (ret, None)
+    };
+
+    (ret, deduped_ambiguous_results, promoted)
+  }
+
   fn match_import_with_export(
     &mut self,
     index_modules: &IndexModules,
@@ -1290,14 +1347,20 @@ impl BindImportsAndExportsContext<'_> {
 
     let mut ambiguous_results = vec![];
     let mut reexports = vec![];
+    // The ambiguous-branch recursion below inherits a clone of the caller's stack, so a cycle can
+    // match a frame belonging to the *outer* request. Resolving that as null is still right for
+    // THIS request, but such a resolution is context-dependent and must not be persisted into the
+    // context-free `resolved_exports` cache: standalone, the collection module may resolve the
+    // name to a different branch, or find it ambiguous. Frames at `stack_base` and beyond were
+    // pushed by this invocation, so only a cycle hitting them justifies the cache write below.
+    let stack_base = ctx.tracker_stack.len();
+    let mut cycle_in_own_frames = false;
     let ret = loop {
-      if ctx
-        .tracker_stack
-        .iter()
-        .rev()
-        .any(|prev| prev.importer == tracker.importer && prev.imported_as == tracker.imported_as)
-      {
+      if let Some(matched) = ctx.tracker_stack.iter().rposition(|prev| {
+        prev.importer == tracker.importer && prev.imported_as == tracker.imported_as
+      }) {
         // ResolveExport treats an in-progress (module, binding) request as null.
+        cycle_in_own_frames = matched >= stack_base;
         break MatchImportKind::Cycle {
           importer: tracker.importer,
           imported: tracker.imported.clone(),
@@ -1358,7 +1421,15 @@ impl BindImportsAndExportsContext<'_> {
                   );
                   let mut reexports_prefix = reexports.clone();
                   reexports_prefix.push(another_named_import.imported_as);
-                  ambiguous_results.push((ambiguous_result, (reexports_prefix, *ambiguous_ref)));
+                  ambiguous_results.push((
+                    ambiguous_result,
+                    StarBranchPromotion {
+                      reexports_prefix,
+                      export: *ambiguous_ref,
+                      star_module: tracker.importee,
+                      imported: tracker.imported.clone(),
+                    },
+                  ));
                 }
               }
               _ => {
@@ -1367,7 +1438,12 @@ impl BindImportsAndExportsContext<'_> {
                     symbol: *ambiguous_ref,
                     reexports: vec![],
                   }),
-                  (reexports.clone(), *ambiguous_ref),
+                  StarBranchPromotion {
+                    reexports_prefix: reexports.clone(),
+                    export: *ambiguous_ref,
+                    star_module: tracker.importee,
+                    imported: tracker.imported.clone(),
+                  },
                 ));
               }
             }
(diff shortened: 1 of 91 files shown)


IWANABETHATGUY:
Pushed a follow-up commit (6b713cbc8) to this branch. The approach here is right and the `retain` is the correct root-cause fix — but letting a cycle fall through to the ambiguity merge made two previously-unreachable paths live, and both panic on input that builds fine on `main`.

### 1. `unreachable!("symbol should always exist")`

When the surviving star branch resolves through an external or CommonJS module the resolution is `NormalAndNamespace`, not `Normal`, so it fell past the `if let Normal(..) = ret` into the `unreachable!`. Pure-ESM repro (`"external": ["ext-pkg"]`, `"format": "cjs"`):

```js
// index.js
import { foo } from './a.js';
// a.js
export * from './b.js'; export * from './d.js'; export * from './c.js';
// b.js
export { foo } from './a.js';   // cycle
// c.js
export let foo = 1;
// d.js
export { foo } from 'ext-pkg';
```

`main` reports a clean `CIRCULAR_REEXPORT`; this branch panicked. `unreachable!` is not compiled out in release, so via napi that aborts the host process. Fixed by extracting the bound symbol through `MatchImportKind::bound_symbol()`, so namespace-backed resolutions can be reported as ambiguous instead of falling off the end.

### 2. `debug_assert!(!self.files.contains_key(..))` in `Diagnostic::add_file`

`ambiguous_results` is declared outside `'tracking: loop` but pushed to inside it, so walking a cycle collects the surviving branches once per lap. Those duplicates reached `AmbiguousExternalNamespace`, whose `on_diagnostic` calls `add_file` once per exporter. (The doubled `CIRCULAR_REEXPORT` messages on `main` are that same second lap showing through.) Same repro with `d.js` as `export let foo = 2;` — no external or CJS needed. Fixed by deduplicating through an insertion-ordered set, which also keeps that scan linear when one name is supplied by many `export *` sources.

### Smaller changes

- Folded the diagnostic payload into `MatchImportKind::Cycle(Box<CircularReexportInfo>)`. That removes the `MatchingContext.circular_reexport` side channel, its manual reset, and the `expect`, so a `Cycle` can no longer exist without the data needed to report it.
- `pop()` → first element, since `ResolveExport` step 7 makes `starResolution` the *first* non-null branch. This one is spec alignment only; no test distinguishes it from `pop()`.

### Tests

Two fixtures, both failing on 582aea97e with the duplicate-file assertion and passing after:

- `circular_star_reexport_with_ambiguous_alternatives` — pure ESM, cycle + two disagreeing branches
- `circular_star_reexport_with_external_alternative` — cycle + an external branch; the only one that pins the `unreachable!` fix

Worth noting your own fixture already covers the "no duplicated `CIRCULAR_REEXPORT`" case — on `main`'s linking code that exact input emits the error twice, so a successful build asserts it.

`cargo test -p rolldown`: 1890 passed, 0 failed, `test262_module_code` included.

IWANABETHATGUY:
I didn’t notice you had self-assigned; the fix was written before you self-assigned.

crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
```diff
@@ -72,13 +81,37 @@ pub enum MatchImportKind {
...
+    match *self {
+      MatchImportKind::Normal(MatchImportKindNormal { symbol, .. }) => Some(symbol),
+      MatchImportKind::Namespace { namespace_ref }
+      | MatchImportKind::NormalAndNamespace { namespace_ref, .. } => Some(namespace_ref),
+      MatchImportKind::Cycle(_) | MatchImportKind::Ambiguous { .. } | MatchImportKind::NoMatch => {
+        None
+      }
+    }
+  }
+}
```
hyfdev, line 131: ### Blocking: a promoted namespace resolution panics the ambiguity consumer

`bound_symbol()` returns the `namespace_ref` for `Namespace` / `NormalAndNamespace`. When a cycle promotes such a branch to be `ret`, that ref is handed to `MatchImportKind::Ambiguous { symbol_ref, .. }`, and the consumer does:

```rust
// match_imports_with_exports, ~line 1003 — and again in the filter_map at ~1016
let named_export = &owner.named_exports[name];
```

on that ref's owner — a module that took the namespace fallback *precisely because* it has no such named export. The index panics.

Repro. No `external` needed: `has_dynamic_exports` is set for a star from any CommonJS module too (`wrapping.rs:62-84`).

```js
// index.js
import { foo } from './a.js';
console.log(foo);

// a.js
export * from './b.js';
export * from './d.js';
export * from './c.js';

// b.js
export { foo } from './a.js';   // cycle -> null

// c.js
export let foo = 1;

// d.js
export { foo } from './e.js';

// e.js
export * from './cjs.cjs';      // has_dynamic_exports, no named `foo`

// cjs.cjs
module.exports = { bar: 1 };
```

with `{ "input": [{ "name": "index", "import": "index.js" }] }`:

| | result |
|---|---|
| `main` (ce804c60) | clean `2 × [CIRCULAR_REEXPORT] 'foo' cannot be exported from 'b.js' …` |
| this branch | `panicked at bind_imports_and_exports.rs:1003:54: no entry found for key` |

An `external` variant (`e.js: export * from 'ext-pkg'`) panics identically.

**One thing worth being explicit about**, because it changes what the right fix is: this is not an assumption `bound_symbol()` invented. Drop the cycle branch from the graph above and `main` **also** panics — at the old `unreachable!("symbol should always exist")`. That `unreachable!` was the guard documenting the invariant. Replacing it with `if let Some(symbol_ref) = ret.bound_symbol()` made the path silent, and cycle promotion then added a new way into it.

So I'd rather not fix this by skipping namespace results during promotion — that only restores `main`'s behaviour and leaves `main`'s hole open. Make the consumer total instead, at both sites:

```rust
// ~1003
if let Some(named_export) = owner.named_exports.get(name) {
    exporter.push(AmbiguousExternalNamespaceModule { /* … */ });
}

// ~1016, inside the filter_map closure
let named_export = normal_module.named_exports.get(name)?;
```

That fixes the pre-existing panic at the same time.

crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
```diff
@@ -1270,30 +1306,35 @@ impl BindImportsAndExportsContext<'_> {
...
+      .filter(|result| !matches!(result, MatchImportKind::Cycle(_)))
+      .collect::<FxIndexSet<_>>();
+
+    // A cycle resolves to null, so the binding is whatever the remaining star branches say. Per
+    // `ResolveExport` the first non-null branch wins and the rest only have to agree with it.
+    let ret = if matches!(ret, MatchImportKind::Cycle(_)) {
+      ambiguous_results.shift_remove_index(0).unwrap_or(ret)
+    } else {
+      ret
+    };
```
hyfdev: ### Blocking: promotion drops the outer walk's `reexports` chain

`shift_remove_index(0)` substitutes a `MatchImportKindNormal` produced by the **nested** `match_import_with_export` call at ~1235. That nested call starts its own `reexports` at `vec![]` from its own tracker, so every hop the outer walk accumulated is lost. The result then flows into:

```rust
// match_imports_with_exports, ~1040
MatchImportKind::Normal(MatchImportKindNormal { symbol, reexports }) => {
  for r in &reexports {
    if self.side_effects_modules.contains(&r.owner) {
      self.metas[module_idx].dependencies.insert(r.owner);
    }
  }
  self.normal_symbol_exports_chain_map.insert(*imported_as_ref, reexports);
```

and `preserve_reexported_interfaces` (`tree_shaking/passes.rs:111-137`) never re-includes the skipped facade's re-export statement.

It's observable. Two fixtures with an **identical five-file module set** and `preserveModules: true`, differing only in the order of two `export *` lines:

```js
// e.js
export let foo = globalThis.SEED;

// d.js
export { foo } from './e.js';   // the facade hop that gets lost

// b.js
export { foo } from './a.js';   // cycle -> null

// index.js
import { foo } from './a.js';
console.log(foo);
```

```js
// a.js — variant 1: cycle branch first
export * from './b.js';
export * from './d.js';

// a.js — variant 2: cycle branch second
export * from './d.js';
export * from './b.js';
```

| | `a.js` output | `d.js` output |
|---|---|---|
| variant 1 | `import "./e.js"; import "./d.js";` — **no `export { foo }`**, and `b.js` is not emitted at all | `import "./e.js";` — **no `export { foo }`** |
| variant 2 | `import { foo } from "./e.js"; …; export { foo };` | `import { foo } from "./e.js"; export { foo };` |

Both source modules export `foo`. Under `preserveModules` each output file is supposed to mirror its source module's interface, and variant 1 loses it on both. (`main` errors with `CIRCULAR_REEXPORT` on both graphs, so this is new behaviour this PR enables rather than a regression on working input — but it's still wrong output.)

Instrumenting the insert at ~1046 shows it directly:

```
variant 1:  chain=[]                    <- the outer hop through d.js is gone
variant 2:  chain=[d.js#SymbolId(3)]
```

Related, same root: `reexports` is excluded from `MatchImportKindNormal`'s `PartialEq`/`Hash`, so the `FxIndexSet` dedup also silently keeps whichever chain happened to be inserted first — and the branch at ~1249 deliberately constructs `reexports: vec![]`.

The promoted value needs the outer walk's `reexports` prefix spliced onto it, or promotion should carry only the symbol and let the outer walk keep owning the chain.

crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
```diff
@@ -72,13 +81,37 @@ pub enum MatchImportKind {
...
   NoMatch,
 }
 
+/// Where a circular re-export was detected, so `CIRCULAR_REEXPORT` can be reported by whoever ends
+/// up keeping the `Cycle` as the final resolution.
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub struct CircularReexportInfo {
+  importer_id: String,
+  imported_specifier: String,
+}
```
hyfdev: ### `Cycle` doesn't need the box, the struct, or the strings

I measured `size_of::<MatchImportKind>()` four ways:

| shape | size |
|---|---|
| `Cycle` (unit, on `main`) | 40 |
| `Cycle(Box<CircularReexportInfo>)` (this PR) | 40 |
| `Cycle { importer: ModuleIdx, imported: Specifier }` | 40 |
| `Cycle(CircularReexportInfo)` — two `String`s inline | 48 |

`Normal` (32) and `NormalAndNamespace { SymbolRef, CompactStr }` (32) already set the payload ceiling, and `ModuleIdx` (4) + `Specifier` (24) pads to 32, under it. So "Boxed to keep the enum small" only holds against the two-inline-`String`s strawman. Against cheap identifiers the box saves zero bytes and costs a heap allocation plus two `to_string()`s at every detection — and `importer_id` is the expensive one, since `Module::id()` wraps an `ArcStr` and `.to_string()` copies a full absolute path.

That cost lands on a value consumed at exactly one site and discarded everywhere else: the filter below strips every `Cycle` out of `ambiguous_results`, and a top-level `Cycle` is replaced whenever any branch survived — which is the case this PR exists for.

```rust
// The import could not be evaluated due to a cycle; carries where it was detected so
// CIRCULAR_REEXPORT can be reported if no other branch resolves the binding.
Cycle { importer: ModuleIdx, imported: Specifier },
```

detection:

```rust
break MatchImportKind::Cycle { importer: tracker.importer, imported: tracker.imported.clone() };
```

consumption:

```rust
MatchImportKind::Cycle { importer, imported } => {
  self.diagnostics.push(BuildDiagnostic::circular_reexport(
    self.index_modules[importer].id().to_string(),
    imported.to_string(),
  ));
}
```

`Specifier` already derives `Hash`/`PartialEq`/`Eq` and implements `Display`, so the enum derives still work and the message text is unchanged. `CircularReexportInfo` goes away with it, along with its `Hash`/`Eq` — which are never actually executed, since every `Cycle` is filtered out before anything is hashed.

Minor, if you'd rather keep a payload struct: `rolldown_error` already has a field-identical `CircularReexport { importer_id, imported_specifier }`, it's just not re-exported from `lib.rs`. `RequireTla` and `AmbiguousExternalNamespaceModule` are the precedent for building an error payload in the producing crate.

crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
```diff
@@ -1270,30 +1306,35 @@ impl BindImportsAndExportsContext<'_> {
...
+    // lookup, so cycles must not take part in the agreement check below. Walking a cycle also
+    // re-visits the same module once per lap, which collects the surviving branches repeatedly —
+    // those duplicates would otherwise show up as repeated exporters, and repeated files, in the
+    // ambiguity diagnostic. An insertion-ordered set keeps source order for the "first branch
+    // wins" rule below while deduplicating in linear time, since a name supplied by N distinct
+    // `export *` sources would make a scan of the collected results quadratic.
+    let mut ambiguous_results = ambiguous_results
+      .into_iter()
+      .filter(|result| !matches!(result, MatchImportKind::Cycle(_)))
+      .collect::<FxIndexSet<_>>();
```
hyfdev: ### The dedup treats the symptom — two notes on where the fix belongs

Both non-blocking.

**1. The duplicates come from the walk, not from the collection.** The comment names the cause exactly: a cycle re-visits the same module once per lap. For a cycle of length L over a module supplying the name from k `export *` sources, `ambiguous_results` ends up with L×k entries and only k distinct ones. Each extra lap also re-runs `advance_import_tracker`'s `potentially_ambiguous_symbol_refs` clone (~1143) and, for every ambiguous ref that is itself a named import, another recursive call plus a full `tracker_stack` clone (~1237).

The main loop already has a resolve-set — `ctx.tracker_stack`. The branch expansion at ~1235 doesn't. Guarding that on `(importee, imported)` stops the duplicates being produced at all, which makes this dedup unnecessary and lets the plain `retain` stay with no allocation.

To be clear: the re-walk is pre-existing and this PR doesn't add traversals. What it adds is the hashing and allocation needed to consume the duplicates.

**2. The `debug_assert` this works around isn't the linker's.** The dedup is here because duplicate exporters trip:

```rust
// rolldown_error/src/build_diagnostic/diagnostic.rs:79
debug_assert!(!self.files.contains_key(&filename));
```

That assert is already reachable without this PR — `AmbiguousExternalNamespace::on_diagnostic` adds the importer's file and then each exporter's file, so any ambiguity where the importer is also one of the exporters trips it on `main`. I reproduced that separately, no cycle and none of this PR's code involved.

`Diagnostic` keeps `files` and `labels` as separate fields, so several labels in one file are already representable and ariadne renders them fine; the assert is the only thing forbidding it. Making `add_file` idempotent (`entry().or_insert_with(...)`, return the id) fixes it for every producer at once instead of for this one caller.

**Small things in the same block.** Realistic `n` here is ~4 (it's 4 in this PR's own ambiguous fixture), where a `Vec` beats two allocations plus hashing. And `shift_remove_index(0)` on an `IndexSet` is O(n) — it shifts every later entry and fixes up the index table; `swap_remove_index` isn't a substitute because it would destroy the ordering the "first branch wins" rule depends on.

**On that ordering.** It does hold today, but it rests on four facts spread across three modules, none asserted or mentioned here: `star_export_records()` iterates `import_records` by index (= source order), `add_exports_for_export_star` pushes in transitive DFS pre-order, lap 1's list is transitively complete so it dominates index 0, and `FxIndexSet` keeps first-insertion order. Commit 1 of this PR used `pop()` (the *last* branch) and commit 2 changed it to index 0, with no test distinguishing them — which is roughly the failure mode I'd expect this to have again later. If `ResolvedExport` handed back a single ordered `candidates` list instead of a privileged `symbol_ref` plus a bag of alternatives, "first non-null wins" would be the shape of the loop rather than an index, and the promotion step wouldn't be needed at all.

crates/rolldown/tests/rolldown/misc/circular_star_reexport_with_external_alternative/a.js
```diff
@@ -0,0 +1,3 @@
+export * from './b.js';
+export * from './d.js';
```
hyfdev, line 2: ### This fixture is one module hop away from the panic

Worth calling out, because it reads as covering the namespace case when it actually dodges it: this graph is exactly the shape that panics. It survives only because `ext-pkg` is external, so at ~1001 `self.index_modules[symbol_ref.owner].as_normal()` returns `None` and the `named_exports[name]` index is never reached.

Move the external one hop further away — `d.js` from:

```js
export { foo } from 'ext-pkg';
```

to:

```js
// d.js
export { foo } from './e.js';

// e.js
export * from 'ext-pkg';
```

and the owner becomes a normal module, and it panics. A relative CommonJS star does the same with no `external` config at all.

Once the ambiguity consumer is total, this graph would be a good fixture to keep next to the current one.

crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
```diff
@@ -1139,18 +1176,17 @@ impl BindImportsAndExportsContext<'_> {
...
-          self
-            .diagnostics
-            .push(BuildDiagnostic::circular_reexport(importer_id, imported_specifier));
-          return MatchImportKind::Cycle;
+          break 'tracking MatchImportKind::Cycle(Box::new(CircularReexportInfo {
+            importer_id: importer_module.id().to_string(),
+            imported_specifier: tracker.imported.to_string(),
+          }));
         }
       }
```
hyfdev: Nit: the label is only needed to escape the inner `for`. As a predicate a plain `break` works, matching the four other `break` sites in the same loop, and the borrow of `tracker_stack` ends before the `push`:

```rust
let ret = loop {
  if ctx.tracker_stack.iter().rev().any(|prev| {
    prev.importer == tracker.importer && prev.imported_as == tracker.imported_as
  }) {
    // ResolveExport treats an in-progress (module, binding) request as null.
    break MatchImportKind::Cycle { importer: tracker.importer, imported: tracker.imported.clone() };
  }
  ctx.tracker_stack.push(tracker.clone());
```

crates/rolldown/src/stages/link_stage/bind_imports_and_exports.rs
```diff
@@ -63,7 +64,15 @@ impl PartialEq for MatchImportKindNormal {
...
 }
 
-#[derive(Debug, PartialEq, Eq)]
+// Must agree with the `PartialEq` above: `reexports` is deliberately excluded from both, so that
+// two resolutions of the same symbol hash and compare as one.
+impl Hash for MatchImportKindNormal {
+  fn hash<H: Hasher>(&self, state: &mut H) {
+    self.symbol.hash(state);
+  }
+}
```
hyfdev, line 73: Nit: this comment is the only thing keeping the two impls in sync. Declaring the exclusion once on the field makes it structural instead:

```rust
/// The re-export hops walked to reach the symbol. Carried for dependency tracking only, so it is
/// deliberately invisible to equality and hashing: two resolutions of the same symbol are one.
#[derive(Debug, Default)]
struct ReexportChain(Vec<SymbolRef>);

impl PartialEq for ReexportChain { fn eq(&self, _: &Self) -> bool { true } }
impl Eq for ReexportChain {}
impl Hash for ReexportChain { fn hash<H: Hasher>(&self, _: &mut H) {} }

#[derive(Debug, PartialEq, Eq, Hash)]
pub struct MatchImportKindNormal { symbol: SymbolRef, reexports: ReexportChain }
```

Both impls are then derived and can't drift apart. Roughly line-count-neutral — it buys the invariant, not brevity.

Two things that may make this moot, so probably worth deciding after those: if the resolve-set guard removes the dedup, `MatchImportKind: Hash` isn't needed at all; and if fixing the dropped `reexports` chain makes that field load-bearing for *which* branch wins, excluding it from equality needs revisiting anyway.

hyfdev (changes requested):
Thanks for digging into this — the spec reading is right and I think it's the direction we want. `ResolveExport` step 7/8 does treat an in-progress `(module, exportName)` as null and take the *first* non-null star branch, and with this change rolldown is actually closer to the spec than esbuild is here (esbuild runs the agreement check unconditionally, so a cycle on the primary branch poisons the whole result).

I can't take it as-is though. The promotion in `match_import_with_export` has two bugs and they share a root: `shift_remove_index(0)` lifts a resolution produced by a **nested** `match_import_with_export` call and uses it as the **outer** result — without checking it is a shape the outer caller can handle, and without the outer walk's accumulated state.

1. A promoted `Namespace` / `NormalAndNamespace` reaches a consumer that assumes it is a named export → panic on graphs that error cleanly on `main`. No external package required; a relative CommonJS star is enough.
2. A promoted nested `Normal` carries `reexports: vec![]` instead of the outer walk's chain → under `preserveModules`, `export { … }` silently disappears from the output, and *which* one disappears depends on the source order of two `export *` lines.

Repros inline. Everything else I flagged is quality-level and fine as follow-up.

Recording what I checked and found clean, so nobody re-derives it:

- The core resolution is spec-correct. Hand-traced `ResolveExport` against all three new fixtures plus a nested-cycle probe; outputs match.
- `CIRCULAR_REEXPORT` still fires when no branch survives. The `shift_remove_index(0)` → `None` → `unwrap_or(ret)` path is already covered — `rolldown/errors/circular_reexport` and `esbuild/default/export_infinite_cycle{1,2}` contain no `export *`, so they hit exactly that branch, and their snapshots are untouched and still pass.
- The `shift_remove_index(0)` ordering holds. `potentially_ambiguous_symbol_refs` is built by ordered `push` in star-record index order (= source order) and lap-1 entries always precede later laps, so index 0 really is the first surviving branch. Deterministic. I went looking for a counterexample and didn't find one — see the note below on making it not depend on that.
- `ctx.tracker_stack` is consistent: the cycle `break` happens before the `push`, same state the old `return` left.
- test262 accounting reconciles: 543 → 545, the 6 removed `test262_failures.json` lines are exactly the two flipped cases, and the remaining `instn-star-binding.js` fails for an unrelated reason.
- Full integration suite on `main` vs this branch: no test that passes on `main` fails here.

Nic-Polumeyv:
Both blockers are fixed in c1e016f69; your four graphs are in as fixtures and all fail on the parent commit. One choice to check: when no contributor has the named export, the consumer now reports MISSING_EXPORT
  rather than an ambiguity with no exporters. The resolve-set guard, idempotent add_file and ReexportChain newtype are left as follow-ups. The CI red was the refactor commit dropping the pre-existing box_collection expectation; restored.

hyfdev:
I found a issue in case

  ```js
  // a.js
  export * from './b.js';
  export * from './d.js';
  // b.js
  export { foo } from './c.js';
  // c.js
  export { foo } from './a.js';   // closes the circle
  // d.js
  export let foo = globalThis.SEED;
  // z.js
  import { foo } from './c.js';
  export const zfoo = foo;
  ```

  If the entry reaches `c.js` only through `z.js`, `a.js`'s output file is correct. If the entry *also* does `import { foo
  } from './a.js'` — resolved first — `a.js`'s output loses its interface:

  ```js
  // a.js output — before, entry also imports from a.js
  import "./d.js";

  // a.js output — before, entry reaches c.js via z.js only (and after, in both orders)
  import { foo } from "./d.js";
  export { foo };
  ```
  
  I pushed the fix andtest into this PR to fix it and avoid back and forth loops.
  
Overall, Thank you!

hyfdev (approved):

PR #10449, opened 2026-07-26
refactor(code-splitting): derive symbol chunk ownership as pass-local link data

hyfdev:
Asking the cross-chunk link computation a question used to change the state it answers from: the what-if passes — the wrapping prediction and the entry-facade edge query of #10441 — had to write into the shared ownership records they read, then undo their own writes by hand. This PR makes asking side-effect-free by construction: a pass derives ownership as data local to itself, only the final pass commits it, and the compiler now enforces what a doc comment used to promise.

```rust
// before — the edge query undoes its own writes, in reverse, by hand
let mut journal = Vec::new();
let state = self.compute_cross_chunk_link_state(…, Scoped(&mut journal));
for (symbol_ref, previous_chunk_idx) in journal.into_iter().rev() {
  symbols.get_mut(symbol_ref).chunk_idx = previous_chunk_idx;
}

// after — asking cannot write, so there is nothing to undo
let state = self.compute_cross_chunk_link_state(…); // &self
```

## Motivation

The cost of the old shape was a silent-failure class, not a bug today. The undo was complete only while the pass contained exactly one write; nothing stopped a future edit from adding a second one the journal does not cover. An incomplete undo leaves values behind that later reads can pick up, and what ships is misresolved cross-chunk imports or a wrong facade decision — with no test signal unless a fixture happens to traverse the exact path. Reviewing #10441 meant manually enumerating every mutation site in the pass to prove the undo complete, and that proof expires with every future edit to the pass.

The old shape also leaked between passes even when nothing went wrong. The prediction pass committed provisional ownership it had no way to retract, so stale values sat in the shared records until the final pass overwrote them — the runtime-module spot-clean exists solely to patch the cases where overwriting was not enough (its three call sites now overwrite `None` with `None`; see non-goals).

The coupling itself: the collect step (`collect_depended_symbols`, which derives each declared symbol's owning chunk) wrote its derivation into the shared symbol database because the import-resolution step (`compute_chunk_imports`) reads ownership back later in the same pass. The write was load-bearing within one pass, so no caller could simply skip it — every caller had to either keep it (and pollute) or undo it (and be provably complete).

Emitted output is unaffected. The downstream consumers of committed ownership — the CJS cross-chunk reference rendering in the module finalizer and the chunk-export generator — read identical values at identical times.

## Change

Ownership becomes a value the pass returns instead of state it writes, and committing becomes the final caller's single explicit act. The collect step builds that value — a pass-local `SymbolChunkTable` — from the per-chunk symbol lists it already materialized.

- The import-resolution step resolves a depended symbol through the table, falling back to the database value on a miss — exactly the write-then-read union it replaces, so behavior is unchanged by construction.
- A pass whose answer is never committed (the prediction, the facade edge query) drops its table on return. The `SymbolChunkOwnership` enum, the journal, and its reverse replay are deleted.
- The final `compute_cross_chunk_links` flushes its table in one place, ahead of every downstream reader, and the whole pass chain down from `compute_cross_chunk_link_state` (including `analyze_execution_order`) becomes `&self`.
- The one-chunk-per-symbol debug assert now checks within the pass's own derivation instead of against whatever an earlier pass left behind, removing its only false-positive source.

Alternative weighed: a two-impl ownership view — the committing pass reading the flushed database directly, monomorphized separately from the table-reading query — would keep the final pass's resolve a direct field read instead of an `FxHashMap` probe. Not taken: it doubles the read-path surface for a cost that should be noise (one probe per depended-symbol resolve plus one map build per pass). CodSpeed on this PR is the check, and the view design stays the escalation path if the strict-mode benchmarks disagree.

Non-goal: removing the now-inert runtime-module spot-clean (`clear_module_symbol_chunk_indices`). With no provisional writes left to clean up, its three call sites overwrite `None` with `None`; deleting them wants its own verification, so they stay untouched here.

## Validation

A behavior-preserving refactor pins itself on the existing suite, and the suite is built to catch exactly the drift this change could introduce: any output difference surfaces as a snapshot diff. 1896 integration tests pass with zero snapshot modifications; `cargo clippy --all-targets` and `cargo fmt --check` are clean. (`test262_module_code` fails locally only because this worktree has no test262 submodule checkout — the panic fires before any bundling runs; CI executes it with submodules.) Unverified locally and delegated to CI: the CodSpeed benchmarks for the probe-vs-field-read cost above.

Stacked on #10441.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
+++ b/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
@@ -13,8 +13,8 @@ use oxc_str::CompactStr;
 use rolldown_common::{
   ChunkIdx, ChunkKind, ChunkMeta, CrossChunkImportItem, EntryPointKind, ExportsKind, ImportKind,
   ImportRecordMeta, Module, ModuleIdx, NamedImport, OutputFormat, PostChunkOptimizationOperation,
-  PreserveEntrySignatures, RUNTIME_HELPER_NAMES, RuntimeHelper, SymbolRef, TaggedSymbolRef,
-  UsedSymbolRefs, UsedSymbolRefsBuilder, WrapKind,
+  PreserveEntrySignatures, RUNTIME_HELPER_NAMES, RuntimeHelper, SymbolRef, SymbolRefDb,
+  TaggedSymbolRef, UsedSymbolRefs, UsedSymbolRefsBuilder, WrapKind,
 };
 use rolldown_utils::index_vec_ext::IndexVecRefExt as _;
 use rolldown_utils::indexmap::{FxIndexMap, FxIndexSet};
@@ -39,6 +39,7 @@ struct CrossChunkLinkState {
   index_cross_chunk_imports: IndexCrossChunkImports,
   index_cross_chunk_dynamic_imports: IndexCrossChunkDynamicImports,
   order_live_symbols: FxHashSet<SymbolRef>,
+  symbol_chunk_table: SymbolChunkTable,
 }
 
 #[derive(Clone, Copy)]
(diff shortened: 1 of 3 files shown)


hyfdev:
* **#10454** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10454?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10449** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10449?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10449?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10441** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10441?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10449).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

shulaoda (approved):

shulaoda:
@codex review

hyfdev:
### Merge activity

* **Jul 27, 6:20 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 27, 6:20 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 27, 6:26 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #9601, opened 2026-05-29
feat: add `SOURCEMAP_BROKEN` warning for renderChunk hook

sapphi-red:
Emit `SOURCEMAP_BROKEN` warning for renderChunk hook calls.

refs #6780

PR #10450, opened 2026-07-26
refactor(plugin): share asset emission helpers between copy_module and asset_module

Nic-Polumeyv:
#9601 patched the same placeholder machinery in both plugins at once. This moves it into rolldown_plugin_utils, behavior unchanged. The parse half is split out per review.

### DIFF
--- a/crates/rolldown_plugin_utils/src/asset_emission.rs
+++ b/crates/rolldown_plugin_utils/src/asset_emission.rs
@@ -13,8 +13,6 @@ pub async fn emit_asset(
   ctx: &PluginContext,
   clean_id: &str,
   read_error: impl FnOnce(std::io::Error) -> anyhow::Error,
-  after_emit: impl FnOnce(&str),
-  add_watch_file: impl FnOnce(&str),
 ) -> anyhow::Result<ArcStr> {
   let path = Path::new(clean_id);
   let bytes = tokio::fs::read(clean_id).await.map_err(read_error)?;
@@ -31,8 +29,7 @@ pub async fn emit_asset(
     })
     .await?;
 
-  after_emit(&reference_id);
-  add_watch_file(clean_id);
+  ctx.add_watch_file(clean_id);
 
   Ok(reference_id)
 }
--- a/crates/rolldown_plugin_utils/src/asset_emission.rs
+++ b/crates/rolldown_plugin_utils/src/asset_emission.rs
@@ -29,8 +29,6 @@ pub async fn emit_asset(
     })
     .await?;
 
-  ctx.add_watch_file(clean_id);
-
   Ok(reference_id)
 }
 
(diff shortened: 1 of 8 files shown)


shulaoda:
This involves two separate pieces of duplicated logic. Could you split them into two PRs to make the review easier?

crates/rolldown_plugin_utils/src/asset_emission.rs
```diff
@@ -0,0 +1,94 @@
...
+  let reference_id = ctx
+    .emit_file_async(EmittedAsset {
+      name: Some(file_name),
+      original_file_name: Some(original_file_name),
+      source: StrOrBytes::Bytes(bytes),
+      ..Default::default()
+    })
+    .await?;
+
+  ctx.add_watch_file(clean_id);
```
hyfdev: This registers the file only through `PluginContext::add_watch_file`, so the load-hook call site no longer records a per-module transform dependency.

`AssetModulePlugin::load` used to call `add_watch_file` on `SharedLoadPluginContext`, which resolves to `LoadPluginContext::add_watch_file` and also runs `plugin_driver.add_transform_dependency`. `emit_asset` takes `&PluginContext`, so deref coercion drops that specialization. Global `watch_files` still update, but HMR maps cleaned asset paths to modules via `transform_dependencies` as well as `module_idx_by_abs_path` (`hmr_stage.rs`). When the module id carries a query (the path this plugin already strips before watching), a change to the cleaned file can miss the module.

Plugin-provided load output also sets `is_read_from_disk = false`, so core does not auto-watch the module path; the plugin watch path is intentional.

Keeping `add_watch_file` on the real load context after emit is one way to restore the previous behavior; leaving watch out of `emit_asset` and letting each caller register it is another.

This blocks merge.

hyfdev (changes requested):
The shared `rewrite_emitted_asset_references` path matches the previous `renderChunk` logic in both plugins. The problem is `emit_asset` taking ownership of watch registration through `&PluginContext`, which drops `LoadPluginContext`'s HMR transform-dependency side effect used by `AssetModulePlugin::load`.

See the inline note for details.

hyfdev (approved):
The follow-up moves `add_watch_file` back to the callers, so `AssetModulePlugin::load` again goes through `LoadPluginContext` and keeps the HMR transform dependency. The shared emit and rewrite helpers otherwise match the previous plugin behavior.

hyfdev (approved):
The shared `emit_asset` and `rewrite_emitted_asset_references` helpers match the previous emit and renderChunk logic in both plugins. Watch registration stays on the callers, so `AssetModulePlugin::load` still hits `LoadPluginContext::add_watch_file` and keeps the HMR transform dependency.

This branch needs a rebase onto current `main` to pick up #10452 before it can merge.

hyfdev:
Thank you!

PR #8445, opened 2026-02-24
chore(ai): make CLAUDE.md a symlink of AGENTS.md

hyf0:

PR #10451, opened 2026-07-26
chore(repo): repair dangling CLAUDE.md symlink

hanayashiki:
## What this solves

The root `CLAUDE.md -> AGENTS.md` symlink has been broken since #8445. Its target string is `"AGENTS.md\n"` — a 10-byte path with a literal trailing newline — so nothing resolves it:

```console
$ git cat-file -s $(git rev-parse :CLAUDE.md)
10                      # "AGENTS.md" is 9 characters

$ node -e 'console.log(require("fs").existsSync("CLAUDE.md"))'
false
```

The bad byte is in the committed blob (`c3170642`, mode `120000`) and is present on `main`, so **every clone on every platform gets a dangling link** — a newline is a legal filename character on Linux and macOS alike. It went unnoticed because `ls -l` renders the newline as invisible whitespace, `git status` stays clean, and the link *looks* correct in every listing.

The practical effect: agent tooling that looks for `CLAUDE.md` at the repo root stats it, receives `ENOENT`, and **silently** loads nothing. `AGENTS.md` has not been reaching agents through that path since February — which rather defeats the purpose of #8445.

## The change

One line — recreate the link with an exact 9-byte target:

```sh
ln -sf AGENTS.md CLAUDE.md
```

The diff is just the stray byte:

```diff
-AGENTS.md
+AGENTS.md
\ No newline at end of file
```

Verified after: `readlink` gives `AGENTS.md` (9 bytes), `existsSync` is `true`, and reading `CLAUDE.md` returns the contents of `AGENTS.md`.

## On tests

No test accompanies this. `just lint-repo` runs `typos`, `cargo ls-lint`, and `vp fmt` — none validate that symlinks resolve, which is why this sat unnoticed. A regression guard (failing `lint-repo` on any dangling `120000` entry) would be cheap, but it touches shared tooling and felt like it belonged in its own PR rather than bundled here. Happy to add it in either place if you'd prefer.

For what it's worth, the only other symlink tracked in the repo — `packages/rolldown/tests/fixtures/builtin-plugin/import-glob/follow-symlinks/linked/my-lib` — is clean at exactly 18 bytes and resolves fine. This was the only broken one.

> [!NOTE]
> **AI usage disclosure**, per the [contribution guide](https://rolldown.rs/contribution-guide/): the diagnosis and this PR were produced with Claude Code. I have reviewed and verified the change and its reasoning myself.


hyfdev (approved):
Thank you

PR #10396, opened 2026-07-23
perf: use VisitJs in JavaScript-only AST passes

Boshen:
Use Oxc's `VisitJs` for AST passes that run after ECMAScript transformation, and assert in debug builds that the transformed source type is JavaScript.

Validation: `just lint-rust` and targeted Rust suites.
Binary size: −49,648 bytes (−0.304%) for the macOS arm64 release binding.

PR #10057, opened 2026-07-01
perf: disable preserve_parens across all parse paths

Dunqing:
### What

Sets `preserve_parens: false` on every parser rolldown controls, so the oxc parser no longer emits a `ParenthesizedExpression` wrapper node for each `(expr)` in source:

- `EcmaCompiler::parse` — the primary JS/TS module parse
- `EcmaCompiler::parse_expr_as_program` — text / base64 / data-url modules
- `EcmaCompiler::dce_or_minify` — DCE / minify input (DCE already strips these, so this is a micro-opt that avoids allocating nodes that get removed anyway)
- `enhanced_transform` — the oxc transform path
- the vite `transform`, `import-glob`, and `dynamic-import-vars` plugins

Test-only and benchmark parsers (`cjs_export_analyzer` tests, sourcemap tests, side-effect-detector tests) are intentionally left untouched.

### Why

The wrapper nodes carry nothing codegen needs. In oxc, `GenExpr for ParenthesizedExpression` just delegates to the inner expression and re-derives parentheses from operator precedence:

```rust
impl GenExpr for ParenthesizedExpression<'_> {
    fn gen_expr(&self, p: &mut Codegen, precedence: Precedence, ctx: Context) {
        self.expression.print_expr(p, precedence, ctx);
    }
}
```

So emitted code is identical whether or not the wrappers exist — the flag only affects the in-memory AST shape. Dropping them means fewer AST nodes to allocate and traverse in every scanner / finalizer / visitor pass, and removes an AST-shape inconsistency (DCE already stripped these nodes when tree-shaking ran, so the parsed AST previously differed depending on whether tree-shaking was enabled).

### Note

The existing paren-peeling helpers (`ast_scanner`, `generate_lazy_export`, the vite build-import-analysis plugin, `side_effect_detector::get_inner_expression`) are kept: paren nodes can still enter the AST via plugin `transform_ast` hooks that parse with their own options, and via internal synthesis in `rolldown_ecmascript_utils/src/ast_factory.rs`. `get_inner_expression` also peels TS `as` / `satisfies` / `!` wrappers, so it is needed regardless.

### Test

- Rust integration suite (`cargo test -p rolldown --test integration`, 1778 fixtures) with snapshots on: all pass with **zero** snapshot changes — output is byte-for-byte identical.
- Node fixtures for the three vite plugins (`import-glob`, `dynamic-import-vars`, `transform` / `ecma-transform`) and the transform paths all pass.

PR #10452, opened 2026-07-26
refactor(plugin): share the oxc parse preamble between import-glob and dynamic-import-vars

Nic-Polumeyv:
#10057 and #10396 each had to patch this identical parse preamble in both plugins. Split from #10450.


hyfdev (approved):
Thank you!

PR #10453, opened 2026-07-26
fix(runtime): don't leave default undefined for cjs that fakes __esModule

### LINKED ISSUE #10360: [Bug]: cjs interop: `__toESM(mod).default` is undefined for a cjs module that fakes `__esModule` but has no `default`

minimal repro:

https://stackblitz.com/edit/rolldown-rolldown-starter-stackblitz-b8jfywo8

note: there is no `type: module` so `main.js` is classified as a script/cjs importer

actual issue is caused by `@supabase/functions-js`, which ships `dist/module/*.js` written in esm syntax (`import { __awaiter } from "tslib"`) but its `package.json` has no `"type": "module"` and no `exports` map, so those files are classified as script/cjs importers and their `tslib` import resolves to tslib's cjs umd build. that surfaces the crash in nuxt 4.5 ssr builds via `@nuxtjs/supabase`, specifically under pnpm's node_modules layout (npm's flat hoisting dedupes `tslib` to its esm build and sidesteps it). on the nuxt side this bisects to the vite 8 migration ([nuxt/nuxt#34256](https://github.com/nuxt/nuxt/pull/34256)).

there's a prior user-facing report over in [rolldown-vite#599](https://github.com/vitejs/rolldown-vite/issues/599), auto-closed as `needs reproduction`.

### What is expected?

rollup's `getDefaultExportFromCjs` guards with an extra `('default' in exports)` check so this doesn't crash

```
TypeError: __toESM(...).default is undefined
```

or, we're getting upstream:

```
TypeError: Cannot destructure property '__extends' of '__toESM(...).default' as it is undefined.
```

### System Info

```Shell
n/a
```

### Any additional comments?

_No response_

IWANABETHATGUY:
### Description

Fixes #10360

Some CommonJS modules set `__esModule: true` on `module.exports` without shipping a `default` export — tslib's UMD build (`tslib.js`) is the canonical offender: it fakes the flag via `Object.defineProperty(exports, "__esModule", { value: true })` so that Babel/TS-interop consumers resolve named imports directly off `module.exports`, but it never sets `default`.

When such a module is default-imported from an ESM file that is *not* in node-mode (plain `.js`, no `"type": "module"` — e.g. `@supabase/functions-js`'s `dist/module/*.js`), `__toESM` takes the Babel branch, trusts the faked flag, and skips defining `default` on the interop namespace. Every `.default` access then evaluates to `undefined`:

```
TypeError: Cannot destructure property '__extends' of '__toESM(...).default' as it is undefined.
```

In the wild this breaks Nuxt 4.5 SSR builds via `@nuxtjs/supabase` under pnpm's node_modules layout (see the issue for the resolution chain that lands on the UMD build). Raw Node.js runs the same unbundled graph fine because it ignores `__esModule` entirely, so the crash is a bundling-induced behavior change.

### Fix

Only respect `__esModule` when the module actually has an own `default` property; otherwise fall back to `default = module.exports`:

```js
isNodeMode || !mod || !mod.__esModule || !__hasOwnProp.call(mod, 'default')
  ? __defProp(target, 'default', { value: mod, enumerable: true })
  : target
```

This is the same guard `@rollup/plugin-commonjs` uses in `getDefaultExportFromCjs`:

```js
function getDefaultExportFromCjs (x) {
	return x && x.__esModule && Object.prototype.hasOwnProperty.call(x, 'default') ? x['default'] : x;
}
```

The guard only changes behavior for modules with a faked `__esModule` and no own `default` — the case that previously produced `default === undefined`. It now yields `module.exports`, which is what both node-mode interop and rollup produce, so the three interop regimes converge instead of crashing. Genuine Babel-transpiled CJS (which defines an own `exports.default` whenever a default export exists) is unaffected. `__toESM` is the single shared helper behind static imports, dynamic imports, and external requires, so one change covers all paths.

### ⚠️ Deliberate divergence from esbuild/webpack (review this part)

esbuild produces the identical crash on this input (verified locally), and evanw's [bundler-esm-cjs-tests](https://github.com/evanw/bundler-esm-cjs-tests) conformance suite pins the Babel interpretation. This PR **flips 8 synced conformance expectations** (cases 38, 39, 48, 49, 56, 57, 58, 59 — all the "fake `__esModule` without `default`" shapes) from the esbuild/webpack answer to the rollup answer, both in `scripts/misc/gen-bundler-esm-cjs-tests.mjs` (with NOTE comments) and in the generated fixtures. Observable changes:

- `import foo from './fake.js'` → `module.exports` instead of `undefined` (cases 48/49)
- `import * as ns` → `ns.default` is `module.exports` instead of `undefined`, and `'default'` now appears as an own, enumerable key of the namespace (cases 38/39/56–59)

It also updates the secondary assertions in the `issues/4289` fixture, which pinned `default === undefined` for this shape on the dynamic-import path (the primary named-export assertions from that issue are untouched and still pass).

The trade-off: matching rollup (and the only non-crashing behavior for real-world packages like tslib's UMD build) at the cost of diverging from esbuild/webpack on a corner where their answer is a guaranteed `TypeError` for any consumer that actually uses the default import. Code that feature-detects (`x.__esModule ? x.default : x`) gets the same result under both interpretations.

Docs: the default-interop heuristic in `docs/in-depth/bundling-cjs.md` gains the new condition.

### Test

Added `cjs_compat/script_esm_import_cjs_with_fake_esmodule_flag`: a type-less package whose entry default-imports (plus namespace- and dynamic-imports) a CJS module shaped like tslib's UMD build. Pre-fix the executed output crashes with the exact TypeError from the issue; post-fix `default` resolves to `module.exports` (verified both directions by stashing the runtime change). Full integration suite passes.


crates/rolldown/src/runtime/runtime-base.js
```diff
@@ -61,7 +61,11 @@ export var __reExport = (target, mod, secondTarget) => (
 export var __toESM = (mod, isNodeMode, target) => (
   (target = mod != null ? __create(__getProtoOf(mod)) : {}),
   __copyProps(
-    isNodeMode || !mod || !mod.__esModule
+    // Some CommonJS modules set `__esModule` without shipping a `default`
+    // export (e.g. tslib's UMD build). Treating them as Babel output would
+    // leave `default` undefined, so only respect `__esModule` when an own
+    // `default` actually exists, like rollup's `getDefaultExportFromCjs`.
```
shulaoda: We should remove this comment to make runtime code clean

shulaoda (approved):

IWANABETHATGUY:
This pr is not fully ready; I need to verify the behavior on other bundlers.

IWANABETHATGUY:
I didn't want to argue the "this diverges from esbuild" part of this PR from memory, so I measured it. Harness and raw output here: https://github.com/IWANABETHATGUY-reproduction/rolldown-10360-cjs-interop-matrix

It takes the 10 fixtures whose expectations this PR changes — the 8 synced `bundler_esm_cjs_tests` cases, the `issues/4289` non-node-mode assertions, and the new tslib-shaped one — rebuilds each as a standalone package with no `"type"` field, bundles it with node / webpack / esbuild / rollup / rolldown, runs the output, and records what actually comes back. It reports observed values rather than pass/fail, since the synced fixtures bake in esbuild's answer by construction.

| case | what's being asked | webpack · esbuild · rolldown today | node · rollup · rolldown with this PR |
| --- | --- | --- | --- |
| 38, 39 | `ns.default` | `undefined` | the exports object (`{__esModule, bar}`) |
| 48, 49 | `import foo from './foo.js'` | `undefined` | `{__esModule, bar}` |
| 56, 57 | is `default` among `Object.keys(ns)`? | no | yes |
| 58, 59 | `hasOwnProperty(ns, 'default')` | `false` | `true` |
| 4289 | `ns.default` via dynamic `import()` | `undefined` | `{parse}` |
| tslib | `import x from` a tslib-shaped UMD file | `undefined`, then TypeError on use | `{__awaiter, __extends}` |

Within each group every tool produced the same answer in every case.

Two things I think are worth knowing before reviewing:

**Unbundled Node lands on this PR's side in all 10 cases.** I hadn't checked that before opening this, and it changes how I'd frame the change: less "swap esbuild's convention for rollup's", more "stop emitting a value you can't get by running the same source without a bundler". In the tslib case what webpack/esbuild/rolldown-today produce is `undefined`, so anything that actually uses the default import gets `TypeError: Cannot destructure property '__extends' of '…default' as it is undefined` — the error from #10360. Node runs the same unbundled graph fine.

**The control is PR base vs PR head.** Published 1.2.0 is 99 commits behind this branch's base, so "published vs PR" doesn't isolate the change on its own. The matrix has a separate column built from the parent commit `87278129e`: it matches published on all 10 cases and differs from `4aed17f1b` on all 10, so the delta really is this patch and nothing else in between. The head numbers come from the pkg.pr.new build of this PR; I also reproduced them from a local build of the same commit.

Two caveats so the table isn't over-read:

- This is a claim about `default` only, not full output equality. Node still differs from everyone on the rest of the namespace in cases 56, 57, 59 and 4289 — it adds its own `module.exports` named export (Node 23.0.0+) and cjs-module-lexer can't see computed `exports[...]` assignments. Both are orthogonal to this change, and the per-case sections in `results.md` list Node separately where that happens.
- The tslib case is synthetic: it reproduces tslib's UMD shape (`__esModule` via `defineProperty`, helpers as named properties, no `default`) without pulling in the real dependency tree. The npm/pnpm hoisting variance described in the issue is a resolution question rather than an interop one.

Happy to add cases if anyone wants a particular shape covered.

crates/rolldown/src/runtime/runtime-base.js
IWANABETHATGUY: Condensed it to one line in 42b30ad94 rather than dropping it outright:

```js
// `__esModule` alone is not enough: the module must own a `default` (#10360).
```

The extra clause reads as redundant at a glance, and this is the one spot where the helper intentionally differs from esbuild's, so it's easy to "simplify" back and silently reintroduce the bug — which would also re-flip the 8 synced conformance cases. Runtime comments are stripped from emitted output, so it costs nothing in the bundle.

Happy to remove it entirely if you'd still prefer.

PR #10454, opened 2026-07-26
fix(code-splitting): don't order-wrap a merged-away entry with an unobservable load

hyfdev:
Under `output.strictExecutionOrder`, order-wrapping an entry whose chunk the post-chunk optimizer merged away does not defer anything. It makes `restore_order_wrap_entry_facades` resurrect a file for that entry, because a merged entry has nowhere else to put its trigger. On an app that declares a lazy entry per route, that is one extra chunk per route. This skips the wrapper in the one case where it cannot buy anything: no module in the entry's own static load closure is order-sensitive, so running that whole closure early is not observable.

lobehub, same Vite, only the rolldown package swapped:

| | JS files | raw | gzip | brotli |
| --- | ---: | ---: | ---: | ---: |
| `strictExecutionOrder` off | 1518 | — | — | — |
| before | 2585 (+1067) | +0.80% | +2.12% | +2.04% |
| after | 1600 (+82) | +0.42% | +0.94% | +0.91% |

ComfyUI_frontend and plane are byte-identical before and after — neither has an entry the optimizer merged away, so nothing is skipped.

## Motivation

The chunk growth is a regression against released behaviour, not an inherent cost of the guarantee. `order_wrapping.rs` does not exist at `v1.2.0`; it arrived with #10104. Released 1.2.0 emits 1471 JS files on lobehub with `strictExecutionOrder` on, 47 *fewer* than with it off. Current `main` emits 2585.

Nor is the growth the price of the correctness fixes in that series. #10441 gates `create_order_wrap_entry_facades` on the real cross-chunk edges, and on lobehub that gate creates 59 facades. The other 1008 files come from `restore_order_wrap_entry_facades`, whose candidate set is every order-wrapped entry with no necessity test — and the default plan wraps every entry.

Restoring is not the thing to gate. A resurrected chunk is the only place a merged entry's trigger can live:

- Dropping the restore leaves the entry with no trigger at all, so it never initializes. That is a missing evaluation, not a mis-ordering, which is why gating it broke `optimization/chunk_merging/dynamic_entry_merged_in_user_defined_entry`.
- One file can host only one entry's top-level trigger, so two merged entries sharing a host chunk cannot share one.
- Moving the trigger to the `import()` call site fires a microtask too late even when the host chunk's top level is inert. `function/strict_execution_order_invariants/m4_dynamic_facade_race` pins that with a fully wrapped host chunk.

So the wrapper is what to skip, and it is safe to skip exactly when the entry's execution position is unobservable. `is_order_sensitive` already answers that per module from module content — `EcmaViewMeta::ExecutionOrderSensitive`, plus eagerly triggering an interop module. When no module in the entry's static load closure is order-sensitive, an unwrapped body inline in the host chunk runs that closure whenever the chunk loads and nothing an importer can observe differs.

That question reads module content, never chunk reachability, so it does not approximate a graph that does not exist yet when the plan is built — the failure mode that sank three earlier attempts at narrowing this cost, recorded in #10294.

## What comes off, concretely

`issues/9463` is the fixture this facade mechanism exists for: entry `b` must not run entry `a`'s top-level side effect. Its `shared.js` is a dynamic-import target the optimizer merged into the shared chunk, so it was order-wrapped and restored as its own file:

```js
// shared.js — resurrected so this call has somewhere to live
import { i as init_shared, r as foo } from "./common~a~b~shared.js";
init_shared();
export { foo };
```

`init_shared` was `__esmMin((() => {}))`. `shared.js` only declares a function, so the wrapper deferred nothing and the file existed to call an empty init. Afterwards the module is a namespace object in the host chunk and the dynamic import collapses to a same-chunk `Promise.resolve()`. `init_a` and `init_b` are untouched, and the fixture's `_test.mjs` — which asserts that loading `b` never runs `a`'s side effect — still passes.

`optimization/chunk_merging/dynamic_entry_merged_in_user_defined_entry` goes from three emitted files to one for the same reason.

## The sensitivity signal has to be complete

`add_top_level_eager_order_reasons` ran only when `experimental.onDemandWrapping` was set, which left `ExecutionOrderSensitive` less complete in the default plan than in the selective one. Any decision about what a wrapper buys reads that flag, so the reasons are now collected whenever `strictExecutionOrder` is on.

Measured on its own, with the entry rule disabled: ComfyUI_frontend and plane byte-identical, lobehub raw byte-identical with gzip and brotli moving by 30 and 59 bytes out of 19 MB.

## Non-goals

- The selective wrapping plan stays behind `experimental.onDemandWrapping`. The predicted chunk edges, the expected-versus-actual order comparison and the emergent-cycle fixpoint are untouched, and which non-entry modules the default plan defers is unchanged.
- The default plan is not brought to parity with the selective one on bytes. Both emit 1600 files on lobehub, but the selective plan is 28 KB smaller raw there, and on ComfyUI_frontend it is +4.13% raw against this PR's +7.12%. That remainder is wrapper code rather than chunk proliferation.
- 1471 is not a target. Counting execution-order violations directly in the emitted output — a chunk whose inline `__esm` init another chunk statically imports, so loading that other chunk runs it — gives 58 for `strictExecutionOrder` off (pre-existing, from interop wrapping), 56 for released 1.2.0, and 0 both before and after this PR. Part of why the released file count is lower is that the build was not emitting those triggers.

## Tests

`test_extended_top_level_reasons_do_not_leak_to_the_non_strict_default` replaces `..._do_not_leak_to_default_or_wrap_all`: the reasons still must not reach a non-strict build, and both strict plans now collect the same ones. The snapshot updates are the fixtures whose entries the rule now leaves unwrapped.

### DIFF
--- a/crates/rolldown/src/ast_scanner/impl_visit.rs
+++ b/crates/rolldown/src/ast_scanner/impl_visit.rs
@@ -83,7 +83,10 @@ impl<'me, 'ast: 'me> VisitJs<'ast> for AstScanner<'me, 'ast> {
         Some(&self.namespace_object_symbol_ids),
       );
       let mut stmt_eval_facts = analyzer.analyze_stmt(stmt);
-      if self.immutable_ctx.options.is_strict_on_demand_wrapping_enabled()
+      // Every strict build reads `ExecutionOrderSensitive` to decide what a wrapper would buy, so
+      // the reasons have to be collected whenever strict execution order is on. Collecting them
+      // only for the selective plan left the default plan with a weaker signal than it needs.
+      if self.immutable_ctx.options.is_strict_execution_order_enabled()
         && !self.result.ecma_view_meta.contains(EcmaViewMeta::ExecutionOrderSensitive)
         && !stmt_eval_facts.is_order_sensitive()
       {
--- a/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs
+++ b/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs
@@ -1211,7 +1211,7 @@ mod test {
       .map(|stmt| {
         let analyzer = StmtEvalAnalyzer::new(&ast_scopes, flags, options, None, None);
         let mut facts = analyzer.analyze_stmt(stmt);
-        if options.is_strict_on_demand_wrapping_enabled() && !facts.is_order_sensitive() {
+        if options.is_strict_execution_order_enabled() && !facts.is_order_sensitive() {
           analyzer.add_top_level_eager_order_reasons(stmt, &mut facts);
         }
         (facts.tree_shaking_flags(), facts.is_order_sensitive())
@@ -2277,7 +2277,7 @@ mod test {
   }
 
   #[test]
-  fn test_extended_top_level_reasons_do_not_leak_to_default_or_wrap_all() {
+  fn test_extended_top_level_reasons_do_not_leak_to_the_non_strict_default() {
     let manual_pure_tag = "function tag() { return (/* @__PURE__ */ globalThis.__orderRead?.() ?? 5); } class C { static value = tag``; }";
     let manual_pure_treeshake = || InnerOptions {
       manual_pure_functions: Some(std::iter::once("tag".to_string()).collect()),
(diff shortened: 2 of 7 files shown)


hyfdev:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10454?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#10454** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10454?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10454?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10449** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10449?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10441** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10441?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10454).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyfdev:
I'm also investigating the extra 82 chunks to check if it's reasonable or not.

hyfdev:
Wait. It seems there's a better solution

PR #10456, opened 2026-07-26
fix(code-splitting): let a cross-chunk dynamic importer carry the order trigger

hyfdev:
#10433 lets a dynamic import collapse inline when chunking already placed the target in the importer's own chunk, and the collapse carries the target's deferred initialization. A cross-chunk dynamic importer was left keeping its facade chunk. It does not need one: the cross-chunk rewrite emits `import('./host.js').then(n => (n.init_x(), n.namespace))`, which carries the trigger just as well. Both arms already share that shape through the `esm_init_target` view, so this only stops asking for a file.

Emitted JS files, same Vite, only the rolldown package swapped:

| project | `strictExecutionOrder` off | before | after |
| --- | ---: | ---: | ---: |
| lobehub | 1517 | 2584 (+1067) | **1576 (+59)** |
| ComfyUI_frontend | 521 | 521 (+0) | 521 (+0) |
| plane | 923 | 924 (+1) | 924 (+1) |

The 1008 files that come off lobehub are one restored facade per lazy route. ComfyUI_frontend and plane are unchanged: neither has a dynamic entry the chunk optimizer merged away, so neither restored a facade to begin with.

lobehub's residual +59 is the other facade source, `create_order_wrap_entry_facades`, and is out of scope here. Those entries still have a chunk of their own, so declining to split one leaves its inline `init_E()` at the top of a chunk other chunks import — the hazard the split exists for. Reaching them needs the chunk demoted without a facade emitted, which is a separate change.

## Motivation

A restored facade exists so a merged entry's trigger has a load site of its own. When every live importer of that entry is a dynamic one, every entry path into it is an `import()` whose rewrite can run the trigger, so the load site is redundant — the facade is a file whose whole body is `init_x()`.

This is also what `strictExecutionOrder` did before #10104. That release drove SEO through `WrapKind::Esm`, so a merged dynamic entry took the same cross-chunk rewrite and shipped no facade; `v1.2.0` emits `import(\`./FileViewer-9VJj0tgR.js\`).then(e => (e.r(), e.n))` on lobehub and 1471 JS files, 47 fewer than with the flag off. #10104 moved order wrapping onto its own plan, where `wrap_kind` stays `None`, and routed those entries to facade restoration instead.

Unchanged: a module with no dynamic importer has no call site to carry anything, and an emitted chunk still needs a real file for its reference id to resolve to. Two more shapes cannot carry the trigger either, and keep their facade for the same reason: a two-argument `import()`, whose specifier `try_rewrite_import_expression` never rewrites, so a file has to stay at the name the source wrote; and a host chunk that could expose a callable `then` to a cross-chunk importer, where promise resolution assimilates the chunk namespace as a thenable and the rewrite's extraction callback never receives it. The `then` check reads both the source-level export alias and the declaring symbol name, since a common chunk emits cross-chunk exports under the latter.

## What this gives up

The trigger now runs in the rewrite's `.then`, one microtask after the host chunk settles, instead of inside a facade's own evaluation. A microtask already queued when the chunk settles therefore observes the target uninitialized. `m4_dynamic_facade_race` measures exactly that and its runtime oracle changes:

```
before: ['target', 'checkpoint:true']
after:  ['checkpoint:false', 'target']
```

What an importer sees is unaffected — by the time its own `await` resumes, the target has run — and `cross_chunk_dynamic_importer_uses_call_site_trigger` still asserts at runtime that entering `b` initializes the target without running entry `a`. Counting execution-order violations in lobehub's emitted output (a chunk whose inline `__esm` init another chunk statically imports) gives 58 with the flag off and 0 both before and after this PR.

So the trade is 1008 files against that one microtask, on the same side of it that shipped in 1.2.0.

## Tests

- `wrapped_dynamic_entry_keeps_facade_after_manual_chunk_merge` becomes `wrapped_dynamic_entry_uses_the_call_site_trigger_after_manual_chunk_merge`: `a` imports the host chunk, the rewrite carries the trigger, and no `target.js` is emitted.
- `cross_chunk_dynamic_importer_keeps_facade` becomes `cross_chunk_dynamic_importer_uses_call_site_trigger`. Its runtime assertions are unchanged; only the facade in its snapshot is gone.
- `m4_dynamic_facade_race` records the microtask ordering above.
- `dynamic_import_options`, `dynamic_import_then_export` and `dynamic_import_then_export_alias` cover the two keep-facade cases above at runtime; all three came out of review feedback on this PR.
- Full suite: no other test changes, and no new failures.


hyfdev:
* **#10485** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10485?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10456** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10456?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10456?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10433** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10433?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10456).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

IWANABETHATGUY:
I probed the "every live dynamic importer can carry the trigger" assumption with two runtime fixtures. **Both pass on the parent commit (`14f6c01ce`) and fail on this branch** — each breaks the call-site rewrite a different way, so posting both.

| fixture | parent (#10433 tip) | this PR |
| --- | --- | --- |
| `dynamic_import_options` | ✅ | ❌ `ERR_MODULE_NOT_FOUND: ./target.js` |
| `dynamic_import_then_export` | ✅ | ❌ `TypeError: n.init_target is not a function` |

Both live under `crates/rolldown/tests/rolldown/function/strict_execution_order_invariants/` with `"snapshot": false`; repro with `cargo test -p rolldown --test integration dynamic_import_`.

## 1. A two-argument `import()` never carries the trigger

`try_rewrite_import_expression` bails as soon as `expr.options.is_some()` (`module_finalizers/mod.rs:2407`) — before the specifier finalization *and* before the trigger-carrying rewrite. So the emitted code still contains the raw

```js
const targetPromise = import("./target.js", {});
```

while this PR eliminated the facade that used to exist at that path:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '…/dist/target.js' imported from …/dist/a.js
```

The importer scan in `restore_order_wrap_entry_facades` counts dynamic import records without checking for options, so it classifies this importer as one whose rewrite carries the trigger. I suspect the same blind spot reaches #10433's same-chunk collapse (a two-arg importer inside the host chunk), so the fix probably belongs in the shared scan downstack — e.g. a scanner-set `ImportRecordMeta` flag for `import()` with options, treated as facade-requiring.

**`dynamic_import_options/_config.json`**

```json
{
  "_comment": "A live two-argument import() is not rewritten by the module finalizer. If its wrapped dynamic entry is merged into the manually assigned host chunk, the entry facade must remain available at the original stable specifier so the import resolves and runs the entry trigger.",
  "snapshot": false,
  "config": {
    "input": [
      {
        "name": "a",
        "import": "./a.js"
      },
      {
        "name": "b",
        "import": "./b.js"
      }
    ],
    "entryFilenames": "[name].js",
    "chunkFilenames": "[name].js",
    "strictExecutionOrder": true,
    "experimental": {
      "onDemandWrapping": true
    },
    "codeSplitting": {
      "groups": [
        {
          "name": "dyn",
          "test": "[\\\\/](?:target|observer)\\.js$"
        }
      ]
    }
  }
}
```

**`a.js`**

```js
globalThis.events = [];
export const targetPromise = import('./target.js', {});
```

**`b.js`**

```js
import './observer.js';
```

**`target.js`**

```js
globalThis.events.push('target');
globalThis.ready = true;
export const value = 1;
```

**`observer.js`**

```js
globalThis.events.push(`observer:${Boolean(globalThis.ready)}`);
```

**`_test.mjs`**

```js

const { targetPromise } = await import('./dist/a.js');
const target = await targetPromise;

assert.strictEqual(target.value, 1);
```

## 2. The host chunk namespace can be a thenable

The cross-chunk rewrite resolves `import('./dyn.js')` through the promise resolution procedure. When the host chunk re-exports a sibling's callable `then` (here preserved by `minifyInternalExports: false`, and kept live by `b.js` importing it cross-chunk), the chunk namespace is assimilated as a thenable — the extraction callback never receives the namespace at all:

```js
const targetPromise = import("./dyn.js").then((n) => (n.init_target(), n.target_exports));
// TypeError: n.init_target is not a function
// (n is 'not the host namespace' — the value the hijacking `then` resolved with)
```

The facade was the safe import boundary: its namespace holds only the entry's own exports, so an unrelated chunk-mate can't hijack the promise. Note the target itself exporting `then` would be the user's own spec-quirk to own — the bug here is that a *sibling's* export changes the observable result of `import('./target.js')`.

(The same `.then(n => n.ns)` shape predates this stack in the flag-off merged-entry rewrite, so plain builds with `minifyInternalExports: false` are probably exposed on `main` too — that part deserves its own issue. But strict-mode wrapped entries were previously always behind a facade, so for this path it's a regression of this PR.)

**`dynamic_import_then_export/_config.json`**

```json
{
  "_comment": "A wrapped dynamic entry shares its host chunk with an unrelated callable export named `then`. With internal export names preserved, dynamically importing that host would assimilate its namespace as a thenable before the generated extraction callback runs, so the target facade must remain the dynamic-import boundary.",
  "snapshot": false,
  "config": {
    "input": [
      {
        "name": "a",
        "import": "./a.js"
      },
      {
        "name": "b",
        "import": "./b.js"
      }
    ],
    "entryFilenames": "[name].js",
    "chunkFilenames": "[name].js",
    "minifyInternalExports": false,
    "strictExecutionOrder": true,
    "experimental": {
      "onDemandWrapping": true
    },
    "codeSplitting": {
      "groups": [
        {
          "name": "dyn",
          "test": "[\\\\/](?:target|observer|then-export)\\.js$"
        }
      ]
    }
  }
}
```

**`a.js`**

```js
globalThis.events = [];
export const targetPromise = import('./target.js');
```

**`b.js`**

```js
import './observer.js';
import { then } from './then-export.js';

globalThis.hostThen = then;
```

**`then-export.js`**

```js
export function then(resolve) {
  resolve('not the host namespace');
}
```

**`target.js`**

```js
globalThis.events.push('target');
globalThis.ready = true;
export const value = 1;
```

**`observer.js`**

```js
globalThis.events.push(`observer:${Boolean(globalThis.ready)}`);
```

**`_test.mjs`**

```js

const { targetPromise } = await import('./dist/a.js');
const target = await targetPromise;

assert.strictEqual(target.value, 1);
```

hyfdev:
Both confirmed, both fixed here — your fixtures are in verbatim, plus a third one for an alias case the fix review caught: `export { then as hostThen }` still emits a literal `then` cross-chunk, so the guard matches the emitted name, not just the alias.

On this branch: a two-argument `import()` now marks its target facade-required, and the cross-chunk collapse is refused when the host chunk could expose a callable `then`.

You were right about `main` — I re-ran both fixtures with the flag off and got the same two failures. Root fixes are coming as separate PRs: one keeps the entry chunk alive when a two-argument `import()` targets it, one stops internal chunk exports from ever being named `then`. I'll link them here. The guards on this branch stay as backstop for what renaming can't reach (an entry's own public `then`, `export * from` an external).

shulaoda:
@codex review

IWANABETHATGUY (approved):

hyfdev:
Root fixes are up: #10479 keeps the entry chunk a two-argument `import()` names, #10480 stops internal chunk exports from ever being named `then`.

hyfdev:
### Merge activity

* **Jul 27, 10:45 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 27, 5:17 PM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 27, 5:27 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10458, opened 2026-07-26
feat(plugin): estimate hook cost against measured phase wall clocks

IWANABETHATGUY:
Stacked on #10411.

## Problem

`[PLUGIN_TIMINGS]` summed each hook call's `Instant::elapsed()` and reported every plugin's share of that sum. Concurrent calls queue behind each other on the single JS thread, so a call's measured duration includes the wait behind its predecessors: N calls dispatched together and served serially sum to `c·N(N+1)/2` against a true occupancy of `c·N` — overcounting by roughly half the queue depth.

Module loading spawns a task per module (`module_loader.rs:297`), so its hooks inflate by orders of magnitude. A serially-invoked callback like the `advancedChunks` `groups[].name` classifier added in #10411 is measured exactly. Ranking those two against each other — and against a denominator that was itself the inflated sum — meant the classifier could never surface no matter how much wall time it cost.

## Approach

Assign every hook to a `TimingSection` describing how it is dispatched, and measure what each concurrent section really took. Module loading, chunk instantiation, `renderChunk` and `augmentChunkHash` are each bounded by a single awaited call, so four boundaries cover all of it.

A hook's estimate is its share of its section's measured hook time, scaled by that section's wall clock:

```
est(hook) = hook_micros / section_hook_micros * section_wall_micros
```

The inflation factor is a property of the section's queue depth, so it is common to every hook in the section and cancels in the ratio; multiplying by a real wall clock re-grounds the result in seconds.

Serial sections deliberately get **no** boundary and are reported as measured. Any window wide enough to enclose a serial call site also contains unrelated Rust work that apportioning would credit to the callback — bounding the classifier by `generate_chunks()` would have handed it that call's entire chunk-graph construction.

Estimates are therefore additive, comparable across sections, and bounded by the build, so percentages are shares of build time and each row carries a duration:

```
Your build spent significant time in plugin hooks. Estimated share of the build:
  - my-plugin (61%, ~6.2s)
      transform (58%, ~5.8s)
      resolveId (3%, ~350ms)
  - output options (21%, ~2.1s)
```

Rows are keyed on plugin **identity**, not display name, and a name is resolved only for rendering. Two registrations of one plugin are separate culprits; merging them could produce a row clearing the one-second floor that neither reached alone, with their hook breakdowns fused. It also means a plugin named `output options` cannot collide with the label used for core-invoked callbacks.

`HookKind` maps statically to its section via a `const fn`, so call sites pass only the hook and cannot pair one with the wrong section.

## Every hook inside a boundary must be measured

Apportioning a wall clock has a consequence the old scheme did not: anything unmeasured inside a boundary is not omitted, it is **redistributed**. Under the previous denominator (the sum of recorded hooks), an untimed hook was simply absent from the accounting. Here, the section's whole span is divided among whoever *is* recorded, so an untimed hook's cost lands on unrelated ones.

`transform_ast` was exactly this case — the one driver hook loop with no `record_timing` call, untimed on main too, running inside `fetch_modules` via `parse_to_ecma_ast`. Registered in-tree plugins hit it: the registration filter only skips `builtin:` names, and `lazy-compilation` implements the hook. It is now timed and reported.

Finding it required the inverse of the sweep that missed it — enumerating hooks that *execute inside* each section rather than existing `record_timing` call sites, which structurally cannot see a hook with no call site. All 22 driver hook loops are now instrumented. `groups[].test` is timed as well: it sits in the same serial loop as the `name` classifier and was invisible for the same reason, which undercut the point of surfacing output-option callbacks at all.

The remaining core-invoked callbacks (`sanitizeFileName`, the filename functions) run in `generate_chunk_name_and_preliminary_filenames`, outside every apportioned boundary, so they stay omissions and cannot be redistributed. Worth timing eventually; not a regression here.

## What this deliberately does not change

`plugins_are_slow` is left alone. Gating on the estimate would be circular: it apportions a section's whole wall clock to the hooks in it, so it cannot tell a plugin-bound phase from a Rust-bound one.

Worth flagging for reviewers that the existing gate is weaker than it looks. `(total - link) / link > 100` only establishes that the link stage is under ~1% of the build, not that plugins dominate — everything except linking (oxc parsing, codegen, minification) sits in the numerator. What now carries the weight is the second condition, an owner clearing 1s of *estimated* time; the old inflated sums cleared that almost automatically, so the report is meaningfully harder to trip than before. Measuring per-section occupancy (in-flight hook count per section, accumulating wall time while ≥ 1) would close the gap properly, serving both as the gate and as a cap on how much of a section gets apportioned. Happy to fold that in here or as a follow-up.

## Notes for reviewers

- Two registrations of the same plugin now render as two rows with identical names. That is truthful and is what fixes the threshold bug; no index suffix was added, since `PluginIdx` is not meaningful to a user.
- `transformAst` has no JS binding — it is Rust-only. The actionable signal is the plugin row; the hook name is the nested detail.

## Tests

- Estimator unit tests: serial hooks reported as measured, concurrent hooks scaled to the section wall clock, a serial hook not drowned out by an overlapping section, `transformAst` taking its share of module loading, same-named plugins staying distinct owners, sections without a boundary dropped, `clear()` semantics, and unregistered plugins skipped.
- A rendering test pinning the report format.
- `checks.pluginTimings` docs updated — they described the old accumulated-time semantics explicitly, including a warning block about sums not being wall-clock time.


shulaoda:
@codex review

IWANABETHATGUY:
Superseded by #10508 and #10509, which take a different approach to the same problem.

Both of these tried to salvage the Rust-side measurement — this one by apportioning phase
wall clocks, #10484 by restricting the report to serially-invoked hooks. Ground-truth A/B on
a real build showed neither could rank correctly: the callback's execution *start* lives in
the JS event loop and is never sampled from Rust, so no arithmetic over dispatch/completion
pairs recovers it.

The replacement measures inside the JavaScript callback and hands the result back to Rust,
which still owns the gate and the diagnostic. Everything worth keeping came across:
per-hook rows, plugin identity as the row key, and the `codeSplitting groups[].name`
classifier — plus sixteen more option callbacks that were never timed at all.

Closing to keep the queue honest; the discussion here is still readable.

PR #10463, opened 2026-07-26
fix(code-splitting): collect top-level eager order reasons in every strict build

hyfdev:
`add_top_level_eager_order_reasons` ran only when `experimental.onDemandWrapping` was set, which left `EcmaViewMeta::ExecutionOrderSensitive` less complete in the default strict plan than in the selective one. That walk exists because a pure annotation answers a different question than execution order asks: it says a statement is safe to *drop* when nothing uses it, not that the statement is safe to *run at a different time* when it is kept. A top-level statement tree-shaking treats as pure — a `manualPureFunctions`-tagged call in a class static initializer that reads mutable global state, say — still observes its own position once the module is included. So the plan that defers every eligible module was working from a weaker signal than the plan that derives its wrapping from the analysis, for a fact that is a property of the module rather than of the plan.

## What reads the signal

`ExecutionOrderSensitive` is not read only by the wrap planner. `sort_chunk_modules` splits a chunk's modules into the leaf ones — no import records and no `ExecutionOrderSensitive` — and the rest, then hoists that leaf group and sorts it by module id rather than by execution order under `experimental.chunkModulesOrder: 'module-id'`. That consumer has no idea which plan produced the wrap set, so as things stand the same module lands in the leaf group under the default plan and outside it under `onDemandWrapping`, on a question about the module's own content.

Under the default plan the difference is textual for now, and deliberately so: `wrap_all_order_analysis` defers every eligible module, so nothing evaluates at chunk top level and hoisting a leaf cannot move an evaluation. It stops being textual as soon as something reads sensitivity to make a decision under that plan, which is what #10454's exemption for a merged-away entry with an unobservable load does. Landing the completeness first keeps that PR's rule reading the same signal the selective plan already reads.

## The change

Gate the walk on `strictExecutionOrder` rather than on strict plus on-demand wrapping. Non-strict builds keep the old behaviour on purpose: `add_top_level_eager_order_reasons` only ever adds order reasons and never touches the tree-shaking flags, so running it without the flag could only make a build that never asked for an order guarantee pay for one. The three comments that scoped the walk to on-demand builds are updated to match.

Measured on its own, with #10454's entry rule disabled: ComfyUI_frontend and plane are byte-identical, and lobehub is byte-identical raw, with gzip and brotli moving by 30 and 59 bytes out of 19 MB.

Extracted from #10454, which has since been closed in favor of #10456's call-site-trigger line; the completeness argument above stands on its own, and any future default-plan decision that reads sensitivity needs it. Until then the accessor has no in-repo caller, so its doc comment now describes the mode it names rather than the walk it used to gate.

## Tests

`test_extended_top_level_reasons_do_not_leak_to_default_or_wrap_all` becomes `test_extended_top_level_reasons_do_not_leak_to_the_non_strict_default` and now covers both halves at the analyzer level: the non-strict default still collects no reasons, and the two strict plans — wrap-all and on-demand — collect the same one. The full `-p rolldown` suite passes with no snapshot changes.

### DIFF
--- a/crates/rolldown/src/ast_scanner/impl_visit.rs
+++ b/crates/rolldown/src/ast_scanner/impl_visit.rs
@@ -83,7 +83,10 @@ impl<'me, 'ast: 'me> VisitJs<'ast> for AstScanner<'me, 'ast> {
         Some(&self.namespace_object_symbol_ids),
       );
       let mut stmt_eval_facts = analyzer.analyze_stmt(stmt);
-      if self.immutable_ctx.options.is_strict_on_demand_wrapping_enabled()
+      // `ExecutionOrderSensitive` is read outside the wrap planner too — it gates which leaf
+      // modules `sort_chunk_modules` may sort by id — so it has to mean the same thing in every
+      // strict plan. Collecting the reasons only for the selective plan left the default weaker.
+      if self.immutable_ctx.options.is_strict_execution_order_enabled()
         && !self.result.ecma_view_meta.contains(EcmaViewMeta::ExecutionOrderSensitive)
         && !stmt_eval_facts.is_order_sensitive()
       {
--- a/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs
+++ b/crates/rolldown/src/ast_scanner/stmt_eval_analyzer/mod.rs
@@ -26,7 +26,7 @@ bitflags! {
         /// Reads from an unresolved global or a member chain rooted at one.
         const GlobalVarAccess = 1;
         /// A call/new/tagged-template expression was treated as pure by an annotation or
-        /// cross-module analysis. Only the strict on-demand statement walk additionally sets it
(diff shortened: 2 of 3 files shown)


shulaoda (approved):

hyfdev:
### Merge activity

* **Jul 28, 2:42 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 28, 2:42 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 28, 2:49 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10473, opened 2026-07-26
fix(watch): preserve plugin attribution in warnings

### LINKED ISSUE #10472: [Bug]: watch mode warnings lose their plugin attribution
### Reproduction

On rolldown 1.2.0, the same config reports `log.plugin` as the plugin name through `build()` and as `null` through `watch()`:

```js
import { build, watch } from 'rolldown';

const plugin = {
  name: 'my-named-plugin',
  // returning code without a map while sourcemaps are on triggers SOURCEMAP_BROKEN
  transform: (code) => ({ code: code + '\n// touched\n' }),
};

const config = (tag) => ({
  input: 'src/main.js',
  plugins: [plugin],
  onLog: (level, log) => console.log(tag, log.code, log.plugin ?? null),
  output: { dir: `dist-${tag}`, sourcemap: true },
});

await build(config('build'));
const watcher = watch(config('watch'));
watcher.on('event', (e) => e.code === 'END' && watcher.close());
```

prints

```
build SOURCEMAP_BROKEN my-named-plugin
watch SOURCEMAP_BROKEN null
```

### Diagnosis

`WatchTask::emit_warnings` hard-codes `plugin: None` (crates/rolldown_watcher/src/watch_task.rs:239 on main). It was forked from the binding's `handle_warnings` in #8897, and #9600 later changed the binding copy to `warning.plugin()` (crates/rolldown_binding/src/utils/mod.rs:128) without touching the watcher copy, so the two paths have drifted. SOURCEMAP_BROKEN is currently the only warning kind that carries a plugin name, so it is the visible case.

I'd like to take this. Instead of patching the one line, the fix I have in mind routes both copies through a single `Log` constructor that reads `diag.plugin()` itself, so the two paths cannot drift again.

Nic-Polumeyv:
The watcher's warning loop was forked from the binding's before #9600 picked up plugin attribution, so watch-mode logs report `plugin: null`. Both copies now build the `Log` through one constructor.

Fixes #10472

### DIFF
--- a/crates/rolldown_binding/src/utils/mod.rs
+++ b/crates/rolldown_binding/src/utils/mod.rs
@@ -14,7 +14,7 @@ use napi::{
   bindgen_prelude::{PromiseRaw, ToNapiValue},
 };
 use napi_derive::napi;
-use rolldown::{LogLevel, NormalizedBundlerOptions};
+use rolldown::{Log, LogLevel, NormalizedBundlerOptions};
 use rolldown_error::{
   BuildDiagnostic, Diagnostic, DiagnosticOptions, filter_out_disabled_diagnostics,
 };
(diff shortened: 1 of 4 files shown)


shulaoda (approved):
LGTM. Thanks for the contribution!

PR #10474, opened 2026-07-27
feat(dev): expose `enabled` through the dev engine bindings

shulaoda:
## Problem

`DevWatchOptions` has a `disableWatcher` option that swaps the file system watcher for a no-op one (`dev_engine.rs:101-105`). It works, and rolldown's own Rust integration tests use it, but a JS caller cannot reach it: `BindingDevWatchOptions` has no such field and the binding hard-codes `disable_watcher: None` when it builds the options.

So `dev()` has no way to turn the watcher off. Vite's bundled dev needs exactly that to honor `server.watch: null`, which today silently keeps watching in full bundle mode.

## Related Issues

Unblocks the "expose watcher options on the Vite side" item in #10019. The Vite side maps `server.watch` onto these options and needs this released before it can honor `server.watch: null`.

### DIFF
--- a/packages/rolldown/src/binding.d.cts
+++ b/packages/rolldown/src/binding.d.cts
@@ -2041,6 +2041,7 @@ export interface BindingDevtoolsOptions {
 }
 
 export interface BindingDevWatchOptions {
+  enabled?: boolean
   skipWrite?: boolean
   usePolling?: boolean
   pollInterval?: number
--- a/crates/rolldown_binding/src/binding_dev_engine.rs
+++ b/crates/rolldown_binding/src/binding_dev_engine.rs
@@ -48,6 +48,7 @@ impl BindingDevEngine {
       dev_options.as_ref().and_then(|opts| opts.rebuild_strategy).map(Into::into);
     // Take ownership of watch so we can consume Vec fields (include/exclude).
     let watch_options = dev_options.and_then(|opts| opts.watch);
+    let watcher_enabled = watch_options.as_ref().and_then(|watch| watch.enabled);
     let skip_write = watch_options.as_ref().and_then(|watch| watch.skip_write);
     let use_polling = watch_options.as_ref().and_then(|watch| watch.use_polling);
     let poll_interval = watch_options.as_ref().and_then(|watch| watch.poll_interval);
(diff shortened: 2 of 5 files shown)


packages/rolldown/src/binding.d.cts
```diff
@@ -2041,6 +2041,7 @@ export interface BindingDevtoolsOptions {
 }
 
 export interface BindingDevWatchOptions {
+  disableWatcher?: boolean
```
hyfdev: Does current vite have similar feature?

hyfdev: Vite has the same *behavior* (`server.watch: null` → noop watcher), not a nested `disableWatcher` flag. Build-side is closer to `watch: false` / `null`.

On the public JS name: `watch: { disableWatcher: true }` reads awkwardly. Nesting a `disable*` flag under `watch` is a double negative, and it reverses the polarity of the Vite check (`server.watch === null` → `disableWatcher: true`).

`enabled` (default `true`; set `false` to turn the watcher off) fits this position better and maps as `enabled: config.server.watch !== null`. The binding/Rust side can keep `disable_watcher` internally if that is convenient.

hyfdev: In short, let's use `DevWatchOptions.enabled` in js world for better naming. You could later change rust to `enabled` if you feel necessary.

hyfdev (approved):

packages/rolldown/src/binding.d.cts
sapphi-red: Yeah, I think it's better to use `enabled` so the polarity is the same.

h-a-n-a (approved):

shulaoda:
### Merge activity

* **Jul 27, 6:33 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 27, 6:36 AM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 27, 6:48 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10476, opened 2026-07-27
fix(dev): keep the export name of `export * as ns from`

tbvjaos510:
In dev mode, `export * as ns from './dep.js'` silently loses its export name. The eagerly bundled output of the same build is correct, which is what makes this look like a bug rather than a missing feature.

```js
// dep.js
export const value = 'from-dep'
// reexport.js
export * as NS from './dep.js'
// main.js
import { NS } from './reexport.js'
NS.value // TypeError: Cannot read properties of undefined
```

## The two statements share one AST node

`export * from './dep.js'` and `export * as ns from './dep.js'` both parse to `ExportAllDeclaration`; they differ only by whether `exported` is set. `HmrAstFinalizer` never read that field, so the named form fell through to the anonymous star re-export path and was finalized as:

```js
var __rolldown_exports__ = __rolldown_runtime__.__exportAll({})
__rolldown_runtime__.__reExport(__rolldown_exports__, import_dep)
```

That is the semantics of the *unnamed* form — `dep`'s own exports are spread onto this module, and the name `ns` never comes into existence. A consumer importing `ns` gets `undefined`, with no error at build time.

The fix binds the importee's namespace object to that single export name instead, and only when `exported` is present. The unnamed form keeps taking `create_re_export_call_stmt`, byte for byte.

Export names that are not valid identifiers are handled: `export * as "not-an-ident" from './dep.js'` is legal ES2022, so the property is emitted in computed form via the existing `is_validate_identifier_name` check rather than as a bare key.

## Tests

Two separate paths reach this finalizer, and both were broken. Each is pinned by its own test:

- `crates/rolldown/tests/rolldown/topics/hmr/export_star_as` — the HMR patch path. The snapshot shows the export name surviving into the patch.
- `packages/rolldown/tests/dev/dev-lazy-compile.test.ts` — the lazy-chunk path, where the re-exported module is reachable *only* through the namespace re-export, so nothing else can pull it in.

Both fail on `main` and pass with this change; reverting only the finalizer hunk turns both red again.

The lazy-chunk case is a dev-engine test rather than a Rust fixture because lazy compilation is not reachable from the fixture config (`lazyBarrel` is a different feature). `cargo test --workspace` and the Node.js suite pass, with no other snapshot changes.

## Documentation

`internal-docs/lazy-compilation/implementation.md` gains Issue 11, following the format already used there for Issues 1-10, per `AGENTS.md` ("Keep in sync: if your changes affect a doc, update it in the same change").

## Alternatives considered

- **Emitting `__reExport` and then aliasing the name on top.** Works for the common case, but it leaks `dep`'s own exports onto the re-exporting module as a side effect — the exact bug, just harder to see.
- **Handling this in the shared finalizer instead.** `ScopeHoistingFinalizer` already gets this right; the defect is specific to the HMR/dev path, so widening the change would touch a correct code path for no reason.

## Worth a closer look

- **Only the dev/HMR finalizer changes.** Production scope hoisting was already correct, so there is no user-facing change to `build`.
- **How it surfaced.** Found through Vite's `experimental.bundledDev` with lazy compilation, where a UI library re-exported as a namespace arrived as `undefined` at runtime. The HMR patch path is affected identically, so this is not Vite-specific.

## AI disclosure

This PR was developed with AI assistance. The diagnosis, the fix, and both tests were reviewed and validated by me against a real project before submission.

### DIFF
--- a/internal-docs/lazy-compilation/implementation.md
+++ b/internal-docs/lazy-compilation/implementation.md
@@ -454,7 +454,7 @@ The flow is:
 
 ### Issue 11: `export * as ns from` Is Not `export * from`
 
-**Problem**: `export * as ns from './dep'` and `export * from './dep'` are the same oxc AST node (`ExportAllDeclaration`), distinguished only by whether `exported` is set. The HMR finalizer ignored that field and rendered both as a star re-export — `__reExport(__rolldown_exports__, import_dep)` — so the re-exporting module's namespace object never carried `ns`, and every consumer read `undefined`. Only the module wrappers were affected (lazy chunks and HMR patches); the scope-hoisted build resolves the same source correctly, which is why the bug appeared only under `experimental.devMode.lazy`.
+**Problem**: `export * as ns from './dep'` and `export * from './dep'` are the same oxc AST node (`ExportAllDeclaration`), distinguished only by whether `exported` is set. The HMR finalizer ignored that field and rendered both as a star re-export — `__reExport(__rolldown_exports__, import_dep)` — so the re-exporting module's namespace object never carried `ns`, and every consumer read `undefined`. Only the module wrappers were affected (lazy chunks and HMR patches); the scope-hoisted build resolves the same source correctly.
 
 **Solution**: When `exported` is present, bind the importee's `loadExports` result under that single name in the namespace object (`{ ns: () => import_dep }`, computed when the name is not a valid identifier) and emit no `__reExport`. Pinned by `crates/rolldown/tests/rolldown/topics/hmr/export_star_as/`.
 
--- a/internal-docs/lazy-compilation/implementation.md
+++ b/internal-docs/lazy-compilation/implementation.md
@@ -452,6 +452,12 @@ The flow is:
 
 **Solution**: The stub template awaits the **re-registered proxy's own `'rolldown:exports'` promise** (`return await loadExports($STABLE_PROXY_MODULE_ID)['rolldown:exports']`) rather than handing back a namespace — so a rejection anywhere in the chain rejects `lazyExports` and the consumer's import promise. Pinned by the two lazy-init-error specs (cold and warm paths).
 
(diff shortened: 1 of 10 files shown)


internal-docs/lazy-compilation/implementation.md
```diff
@@ -452,6 +452,12 @@ The flow is:
 
 **Solution**: The stub template awaits the **re-registered proxy's own `'rolldown:exports'` promise** (`return await loadExports($STABLE_PROXY_MODULE_ID)['rolldown:exports']`) rather than handing back a namespace — so a rejection anywhere in the chain rejects `lazyExports` and the consumer's import promise. Pinned by the two lazy-init-error specs (cold and warm paths).
 
+### Issue 11: `export * as ns from` Is Not `export * from`
+
+**Problem**: `export * as ns from './dep'` and `export * from './dep'` are the same oxc AST node (`ExportAllDeclaration`), distinguished only by whether `exported` is set. The HMR finalizer ignored that field and rendered both as a star re-export — `__reExport(__rolldown_exports__, import_dep)` — so the re-exporting module's namespace object never carried `ns`, and every consumer read `undefined`. Only the module wrappers were affected (lazy chunks and HMR patches); the scope-hoisted build resolves the same source correctly, which is why the bug appeared only under `experimental.devMode.lazy`.
```
h-a-n-a:
```suggestion
**Problem**: `export * as ns from './dep'` and `export * from './dep'` are the same oxc AST node (`ExportAllDeclaration`), distinguished only by whether `exported` is set. The HMR finalizer ignored that field and rendered both as a star re-export — `__reExport(__rolldown_exports__, import_dep)` — so the re-exporting module's namespace object never carried `ns`, and every consumer read `undefined`. Only the module wrappers were affected (lazy chunks and HMR patches); the scope-hoisted build resolves the same source correctly.
```

This happens in HMR patches as well.

h-a-n-a (approved):
Thank you!

PR #10477, opened 2026-07-27
refactor(plugin): reuse rolldown_utils::stabilize_id in bundle analyzer

Nic-Polumeyv:
The analyzer's private copy is behavior-identical: under its `is_absolute` guard the shared helper reduces to the same `relative_path_to_slash` call, and the virtual-module and passthrough branches match exactly.


IWANABETHATGUY:
Thanks for your contribution! But I plan to remove this plugin in a few days, so I prefer not to accept this pr.

PR #10479, opened 2026-07-27
fix(code-splitting): point a two-argument dynamic import at the chunk that holds the module

hyfdev:
A dynamic import written with a second argument — `import('./target.js', {})`, or anything carrying `with { type: … }` — keeps the specifier the source wrote, while the module it names ships inside the bundle under a different name. Loading such a build fails with `ERR_MODULE_NOT_FOUND`. This makes a two-argument call produce exactly what the same call without the second argument produces.

```js
// input
export const loaded = import('./target.js', {}).then((target) => target.value);
```

```js
// before — the bundle emits target-oUvUsQEi.js, and nothing is at ./target.js
const loaded = import("./target.js", {}).then((target) => target.value);

// after
const loaded = import("./target-oUvUsQEi.js").then((target) => target.value);
```

## Problem

Nothing unusual is needed to hit this: default output settings, no code-splitting configuration, one call site. The failure gets quieter when the target is inlined rather than given its own chunk — with

```js
export const loaded = import('./data.json', { with: { type: 'json' } });
```

the JSON module is inlined into the importing chunk and no file is emitted for it at all, yet the call still names `./data.json`. In a directory where the source tree sits next to the output, that resolves to the original file and the build silently ends up with two copies of the module instead of failing. CJS output has a third shape: the same call also escapes the `import()` → `require()` conversion, so a CJS bundle ships a native `import()` at that site.

The dynamic-import rewrite in the module finalizer — the pass that redirects a call at the chunk that ends up holding the module, collapses it when importer and target share a chunk, and converts it for CJS output — returns early whenever the call has a second argument, before it touches the specifier. Nothing downstream picks the call up again. The early return carries no recorded rationale; it predates the current control flow and reads as "not handled yet".

That leaves two-argument calls as the only place where import attributes mean anything for a bundled module. Everywhere else they are inert: they are read only when rendering an import of an *external* module, they select no loader (a `.txt` file holding JSON and marked `type: 'json'` still loads as text), they do not affect module identity (the same module imported with and without attributes yields one instance), and a static attributed import of a bundled module already drops them.

Only call sites with a second argument are affected. External and unresolved targets are not — that is where attributes are load-bearing and the current behavior is correct — and neither are calls whose specifier is not statically known.

## Fix

When the target is a module rolldown bundles, drop the second argument and let the call take the ordinary path.

The specifier is then redirected at the chunk that emits the module, the merged-entry collapse and namespace extraction apply, and CJS output converts the call like any other — none of that machinery changes, it simply stops being skipped. Dropping the attributes is what the rest of the pipeline already does for bundled modules, so the result is byte-identical to writing the call without its second argument. External targets, unresolved targets, and computed specifiers keep bailing, and an external call is now explicitly protected from the CJS `require()` conversion, which cannot carry attributes.

**Why not keep a file at the name the source wrote.** The first revision of this PR did that: it stopped the chunk optimizer from merging away an entry named by a two-argument call. That only helps when the emitted filename happens to equal the path the source wrote — with hashed names or a chunk subdirectory the specifier still dangles — it costs a merge that was otherwise sound, and it cannot address the inlined case at all, where by construction no file carries that name. This supersedes it: the guard is removed, along with the equivalent guard merged in #10456 and the import-record flag both used.

**Non-goals.** Making attributes meaningful for bundled modules — selecting a module type, distinguishing instances per attribute, warning on conflicting attributes the way Rollup does — is a separate and much larger change; this only stops treating a two-argument call as unrewritable. An options *expression* with side effects is no longer evaluated when the target is bundled; the argument is spec'd as an object literal carrying `with`, and such a call site does not load today anyway.

## Validation

New fixtures, each red before the finalizer change and green after:

- `topics/dynamic_import_options/hashed_chunk_names` — default hashed names, nothing merged. `ERR_MODULE_NOT_FOUND` before; resolves and returns the value after. This is the case the removed guard could never fix.
- `topics/dynamic_import_options/json_module_attribute` — `with { type: 'json' }` on an inlined JSON module. Points at the emitting chunk with no `with` clause, and asserts the data at runtime.
- `topics/dynamic_import_options/cjs_format` — pins that the call converts like a one-argument one instead of staying a native `import()`. Its CJS cell was runtime-green before, because Node's ESM loader resolves the CJS chunk anyway; only the snapshot catches this one.
- `topics/dynamic_import_options/external_target_keeps_attributes` and `…/non_static_specifier_in_cjs` — the two deliberate bails. Byte-identical output before and after.
- `chunk_merging/dynamic_import_with_options_rewrites_specifier` (repurposed from `…_keeps_entry`) — the merged case now collapses into the host chunk instead of keeping a facade.
- `packages/rolldown/tests/…/dynamic-imports-metadata/options-import` — `dynamicImports` names the chunk the call now points at.
- `strict_execution_order_invariants/dynamic_import_options` — updated: no facade, specifier rewritten, trigger carried, runtime assertions unchanged.

`chunk_merging/tree_shaken_options_import_does_not_keep_entry` is deleted; it pinned the removed guard's liveness rule.

One test on `main` depended on the old behavior and had to be rebuilt: `late_order_wrapping_revalidates_output_file` used a two-argument import to keep a graph single-chunk through chunk optimization and multi-chunk only after order lowering. It now gets that shape from an emitted chunk for the entry module plus a manual group, and is verified non-vacuous — the same build succeeds with `strictExecutionOrder` off.

Exactly one pre-existing snapshot changes (the repurposed fixture above); everything else is new files. Full Rust suite matches the baseline failure set by name, and the Node suite is green (883 passed).


hyfdev:
### Merge activity

* **Jul 27, 12:45 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 28, 2:19 PM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 28, 2:28 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

hyfdev:
Note for the next touch: `collect_unrewritable_dynamic_import_targets` has the same gap #10433's review caught (fixed there in a226daeb6) — it filters by module inclusion and `DeadDynamicImport` only, so a record whose statement tree-shaking excluded still counts. The consequence here is only over-conservatism: a phantom two-argument record keeps the entry chunk alive. Same fix applies: walk included statements via `stmt_info_included`, like `is_included_dynamic_import_record` does.

shulaoda:
@codex review

IWANABETHATGUY (commented):
Good catch on the root cause — "two-arg ⇒ never rewritten ⇒ a file has to stay at that name" is the right way to frame it, and the new fixture is a real red/green. Dropping the guard locally gives exactly the reported failure:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../dist/target.js'
  imported from .../dist/a.js
```

One blocker, then a few small things.

## The two existing snapshots need to go back

`e38f106c9` rewrote `dynamic_entry_merged_in_common_chunk/artifacts.snap` and `dynamic_entry_merged_in_common_chunk2/artifacts.snap` to show *more* merging than main does — `imp.js` disappears, `import("./imp.js").then(...)` collapses to `Promise.resolve().then(...)`, and chunk2 loses its `INEFFECTIVE_DYNAMIC_IMPORT` warning.

Neither fixture uses a two-argument `import()`, so the guard is a no-op for both and they shouldn't have moved at all. Running them at HEAD rewrites both files back to main's exact blobs (`39bb870bb`, `18c4f37a0`), and CI agrees — `cargo-test (ubuntu-latest)` is red on precisely those two:

```
test ...__dynamic_entry_merged_in_common_chunk2__config_json ... FAILED
test ...__dynamic_entry_merged_in_common_chunk__config_json  ... FAILED
snapshot assertion for 'artifacts' failed in line 697
test result: FAILED. 1906 passed; 2 failed
```

Looks like they got regenerated against a different working state. `git checkout main -- <the two artifacts.snap>` and a re-run of `just test-rust` should settle it. Worth doing before `graphite: merge-when-ready` picks this up.

## Non-blocking

**The scan now diverges from `order_wrapping`'s.** `order_wrapping.rs:454-492` computes the same "live two-argument dynamic importer" notion via a whole-program statement walk, and its comment claims the same discipline this one does. After `e38f106c9` they agree on neither method nor result, and nothing will catch the drift. Since `is_included_dynamic_import_record` is `pub(super)` now anyway — worth folding both onto one helper?

(FWIW `e38f106c9` does resolve the O(total statements) scan Codex flagged — that thread just got re-anchored onto the new call site at line 844, so it reads as still open. `order_wrapping` is where the walk it described actually still lives.)

**`internal-docs/code-splitting/implementation.md:183-190` should get a line.** That section already spells out the facade-elimination rules in prose, including the analogous #9463 user-entry exception. This is a third exception and the least self-evident of them.

**Small one in the predicate.** `is_included_dynamic_import_record` already resolves the importer and the record, and then the closure at `chunk_optimizer.rs:819-823` redoes both just to read one bitflag. It's correct — the `&&` short-circuit makes the direct index safe — but a variant returning `Option<&ImportRecord>` would let both callers share the lookup.

**No CJS variant on the new fixtures.** The neighbouring `dynamic_entry_merged_in_common_chunk*` ones carry `"configVariants": [{ "format": "cjs" }]`, and CJS is arguably the more interesting case here: the `expr.options.is_some()` bail sits *above* the `import()` → `require()` transform (`module_finalizers/mod.rs:2407`), so a two-arg `import()` stays a native ESM `import()` inside CJS output. Cheap to pin while the fixtures are fresh.

hyfdev:
All four addressed, but the PR changed shape underneath them — worth re-reading rather than diffing against the old version.

The snapshots are back (they were regenerated in a worktree whose path contains `imp`, which those two fixtures group by, so a local run silently regrouped every module; `cargo-test` is green now).

Your CJS observation is what sent me back to the premise. If the bail sits above the `require()` transform, then it also sits above the specifier rewrite — and it does. So a two-argument call never points at the emitted chunk, with or without merging:

```js
const p = import("./target.js", {});          // bundled as target-CXX96dh0.js
const q = import("./other-BTQ0PaZP.js");      // one-argument, rewritten
```

Nothing merged there, default names. Worse with `import('./data.json', { with: { type: 'json' } })`: the JSON is inlined, no file is emitted for it, and the call still names `./data.json` — next to a source tree that resolves, so you get a second copy instead of an error.

Attributes turn out to be inert for bundled modules everywhere else — read only when rendering external imports, no loader selection, no effect on identity, and static attributed imports already drop them. So the guard was patching one symptom of a call site that is simply never rewritten, and only in the configs where the emitted name happens to equal the source path.

This revision drops the second argument for bundled targets and lets the call take the ordinary path, which makes it byte-identical to the one-argument form. The guard is gone, so is the one from #10456 and the record flag both used — which also dissolves the divergence you flagged between the two scans. External and computed-specifier calls keep bailing, each pinned by a fixture. Coverage for hashed names, CJS, and the attributed-JSON case is in.

Two things you should look at specifically: `late_order_wrapping_revalidates_output_file` used a two-argument import to build its late-multi-chunk shape and had to be rebuilt on an emitted chunk plus a manual group; and dropping the attributes is reasoning from rolldown's own treatment of static attributed imports, not from any precedent — Rollup's suite only covers external ones. If you think attributes should instead become meaningful for bundled modules, that is the decision to push back on.

IWANABETHATGUY (approved):

PR #10480, opened 2026-07-27
fix(code-splitting): never name an internal chunk export `then`

hyfdev:
A chunk is loaded through `import('./chunk.js')`, whose promise resolves with the chunk namespace. Promise resolution assimilates any value with a callable `then` as a thenable, so a chunk that exports `then` hijacks every dynamic import of itself — including the `.then((n) => n.ns)` the finalizer generates to reach a merged dynamic entry, whose callback then receives whatever the exported `then` resolved with instead of the namespace. With `minifyInternalExports: false` an internal cross-chunk export named `then` reaches the host namespace directly or through an alias, since a cross-chunk export is emitted under the declaring symbol's name (`export { then as hostThen }` still emits `then`).

`deconflict_exported_names` now reserves `then` before naming bundler-owned internal exports, so they deconflict to `then$1` like any other collision, and the minified-name generator skips the literal `then` (unreachable in practice at four characters, but the generator is the only other source of internal names). Names the user can observe keep whatever they declare: an entry chunk's public exports arrive with predefined names and are emitted verbatim, and an `emitFile`-preserved export keeps `then` only when `then` is the exported name itself — a preserve-name module whose local `then` leaves under an alias deconflicts like any other symbol, so it cannot hand `then` to the chunk (and two preserved modules cannot emit a duplicate one; the second falls back to the resolver).

No snapshot churn: the full fixture sweep produces exactly the same set of snapshots as `main`, so no existing fixture had an internal export named `then`.

#10456's strict-side guard stays for what renaming can't reach: an entry's own public `then` and `export * from` an external module.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
+++ b/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
@@ -31,6 +31,17 @@ type IndexCrossChunkDynamicImports = IndexVec<ChunkIdx, FxIndexSet<ChunkIdx>>;
 type IndexImportsFromOtherChunks =
   IndexVec<ChunkIdx, FxHashMap<ChunkIdx, Vec<CrossChunkImportItem>>>;
 
+/// A chunk is loaded through `import('./chunk.js')`, whose promise resolves with the chunk's
+/// namespace object. Promise resolution assimilates any value with a callable `then` as a
+/// thenable, so a chunk exporting `then` hijacks every dynamic import of that chunk — including
+/// the `.then((n) => n.ns)` the finalizer generates to reach a merged dynamic entry, whose
+/// callback then receives whatever the exported `then` resolved with instead of the namespace.
+///
+/// Internal cross-chunk export names are bundler-owned, so simply never hand out `then` for one.
+/// Names the user can observe — an entry chunk's public exports and the export names an
+/// `emitFile` consumer relies on — are a contract and keep whatever they declare.
+const THENABLE_HAZARD_EXPORT_NAME: &str = "then";
+
 struct CrossChunkLinkState {
   index_chunk_exported_symbols: IndexChunkExportedSymbols,
   index_chunk_direct_imports_from_external_modules: IndexChunkImportsFromExternalModules,
@@ -1338,7 +1349,10 @@ impl GenerateStage<'_> {
           loop {
             named_index += 1;
             export_name = generate_minified_names(named_index).into();
-            if !used_names.contains(&export_name) {
+            // Unreachable in practice — base54 needs about 14M names before it produces a
+            // four-character `then` — but the generator is the only other source of internal
+            // export names, so make it impossible rather than improbable.
+            if !used_names.contains(&export_name) && export_name != THENABLE_HAZARD_EXPORT_NAME {
               break;
             }
           }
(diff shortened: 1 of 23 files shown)


hyfdev:
### Merge activity

* **Jul 27, 8:54 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 28, 6:48 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 28, 6:55 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

IWANABETHATGUY:
The new emitted-chunk branch can hand `then` to two different symbols, and the chunk that comes out doesn't parse — so nothing importing it loads at all.

`minifyInternalExports: false` plus two `AllowExtension` emitted entries merged into the same group is enough to hit it. Repro is `allow_extension_merge_same_exports` with the libs swapped out: each module declares a local `then` but exports it under a different alias, so neither module exports the *name* `then`:

```js
// lib1.js
function then(resolve) {
  resolve('lib1');
}
export { then as first };

// lib2.js
function then(resolve) {
  resolve('lib2');
}
export { then as second };

// index.js
import { first } from './lib1.js';
import { second } from './lib2.js';

assert.strictEqual(typeof first, 'function');
assert.strictEqual(typeof second, 'function');
```

`mod.rs` is `allow_extension_merge_same_exports/mod.rs` verbatim apart from the test name and one added option:

```rust
       BundlerOptions {
         input: Some(vec![InputItem { name: Some("index".into()), import: "./index.js".into() }]),
+        minify_internal_exports: Some(false),
         code_splitting: Some(CodeSplittingMode::Advanced(ManualCodeSplittingOptions {
```

(It has to be a Rust fixture rather than a `_config.json` one — the branch only fires for chunks carrying `ChunkMeta::EmittedChunk`, and the declarative harness has no way to emit one.)

On this branch:

```
dist/libs.js:11
export { then$1 as then, then };
                         ^^^^

SyntaxError: Duplicate export of 'then'
    at compileSourceTextModule (node:internal/modules/esm/utils:318:16)
```

On `main` the same fixture emits `export { then, then$1 }` and passes.

The mechanism is that `preserved_export_refs` collects every export ref of the preserve-export-names modules and throws the export name away, so the branch matches on `base` — the *declaring symbol's* name — rather than on the name the consumer relies on. Both symbols are declared `then`, so both match, and returning `base` skips the one thing that would have recorded `then` as taken.

Filtering the set by export name fixes this case and, I think, is what the comment above the branch already claims it does:

```rust
self.link_output.metas[module_idx]
  .canonical_exports(false)
  .filter(|(name, _)| name.as_str() == THENABLE_HAZARD_EXPORT_NAME)
  .map(|(_, export)| self.link_output.symbol_db.canonical_ref_for(export.symbol_ref))
```

That gives `export { then as then$1, then$1 as then$2 }` — valid, and neither export is named `then`, so the thenable hazard is closed rather than just deduplicated. With it applied: the fixture above passes, both new `dynamic_import_host_exporting_*` fixtures pass, both existing `allow_extension_*` fixtures pass, and the full integration suite is green (1904) with no snapshot churn.

Same filter also covers the aliased single-module case — `export { then as hostThen }` from an emitted entry currently keeps `then` on the chunk, which is the shape `order_wrap_host_can_expose_then_export` explicitly calls out as a hazard in its doc comment.

One thing the filter doesn't cover: two preserved modules that both genuinely export the name `then` would still both skip the resolver. That contract is unsatisfiable anyway, but it'd be nice to fail on something clearer than unparseable output — an "already handed out `then` in this chunk" flag falling back to `resolver.resolve` would do it.

Unrelated to the duplicate, but noticed while poking at this: on `main` the `first`/`second` aliases aren't honored either — the non-minify branch emits the declaring symbol's name, so an emitted entry's export names only survive under `minifyInternalExports: true`. Looks pre-existing and separate from this PR.

IWANABETHATGUY:
A few smaller things alongside the duplicate-export one — none of these are blocking.

### `_test.mjs`, both new fixtures

Both only load `dist/a.js`, so `b.js` — the chunk that actually consumes the renamed binding — never gets linked. The snapshot shows `import { then$1 as then } from "./dyn.js"`, but nothing at runtime proves that resolves; if the two sides ever disagreed you'd get a `SyntaxError: does not provide an export named 'then$1'` and this test would still be green.

Two more lines would close it:

```js
await import('./dist/b.js');
assert.strictEqual(typeof globalThis.hostThen, 'function');
```

### `compute_cross_chunk_links.rs:1352`

The 14M figure looks off by about 32x — I think that's where *five*-character names start (54·64³ = 14,155,776). Four-character names start at 221,184, and `then` specifically lands at index 443,179: `t` is `FREQUENT_CHARS[1]`, `h` is `[15]`, `e` is `[0]`, `n` is `[2]`, so `1 + 54·(15 + 64·(0 + 64·2))`.

Doesn't change the conclusion — nobody is shipping 443k internal exports in one chunk — but since the comment commits to a number it may as well be the right one.

### `internal-docs/code-splitting/`

Would you mind adding a note here? The rule now spans two files and two PRs — rename the bundler-owned names in `deconflict_exported_names`, guard the user-observable ones in `order_wrap_host_can_expose_then_export` — and from either side alone it's not obvious why that split is where it is, or which cases deliberately keep `then`. There's nothing on the thenable hazard in `internal-docs/` today, and this feels like exactly the kind of "why" AGENTS.md asks to be written down.

### nit — `compute_cross_chunk_links.rs:43`

`THENABLE_HAZARD_EXPORT_NAME` here and `THEN` in `order_wrapping.rs:28` are the same string for the same reason. Not worth churning this PR over, but they could share one definition once it lands.

IWANABETHATGUY (approved):

PR #5632, opened 2025-08-06
feat: output better error message for callable plugins

sapphi-red:
Before
```
19:12:22 [vite] Internal server error: foo!
```

After
```
19:15:43 [vite] Internal server error: foo!
  Plugin: builtin:vite-resolve
      at async EnvironmentPluginContainer.resolveId (file:///D:/documents/GitHub/rolldown-vite/packages/vite/dist/node/chunks/dep-BpEN-Pj2.js:30512:19)
      at async EnvironmentModuleGraph._resolveUrl (file:///D:/documents/GitHub/rolldown-vite/packages/vite/dist/node/chunks/dep-BpEN-Pj2.js:34726:39)
      at async EnvironmentModuleGraph.getModuleByUrl (file:///D:/documents/GitHub/rolldown-vite/packages/vite/dist/node/chunks/dep-BpEN-Pj2.js:34533:19)
      at async file:///D:/documents/GitHub/rolldown-vite/packages/vite/dist/node/chunks/dep-BpEN-Pj2.js:26903:20
      at async Promise.all (index 0)
      at async BasicMinimalPluginContext.devHtmlHook (file:///D:/documents/GitHub/rolldown-vite/packages/vite/dist/node/chunks/dep-BpEN-Pj2.js:26902:25)
      at async applyHtmlTransforms (file:///D:/documents/GitHub/rolldown-vite/packages/vite/dist/node/chunks/dep-BpEN-Pj2.js:26472:15)
      at async viteIndexHtmlMiddleware (file:///D:/documents/GitHub/rolldown-vite/packages/vite/dist/node/chunks/dep-BpEN-Pj2.js:26952:24)
```

PR #10482, opened 2026-07-27
fix(binding): stop mishandling per-call data at the callable builtin plugin boundary

hyfdev:
Calling a builtin plugin's hooks directly — the path `rolldown/experimental` exposes and Vite's JS plugin container uses — mishandles three kinds of per-call data: a resolve that carries plugin metadata can come back unresolved, a successful resolve hides the `package.json` the resolver already found, and a failing transform blames the wrong file. Each now behaves the way the same call behaves through the bundler.

```js
// `resolve.tsconfigPaths` on; tsconfig maps `@/*` -> `./src/*`
await resolveId('@/dir/a', importer, {})
// '/app/src/dir/a.js'

await resolveId('@/dir/a', importer, { custom: { 'my-plugin': { … } } })
// before: '/app/src/dir/a'      <- raw mapping; extension never resolved, file never checked
// after:  '/app/src/dir/a.js'
```

Follow-up to #10440, which fixed a fourth defect on this boundary (a `require()` resolve could never select the `require` export condition). These three came out of looking for the same failure mode elsewhere on it.

## 1. Plugin metadata is read as an `import.meta.glob` request

A glob pattern has no file on disk, so when a resolve is marked as coming from `import.meta.glob` the Vite resolver hands back the tsconfig `paths`/`baseUrl` mapping verbatim instead of resolving it. That marker is one key inside `custom`: `vite:import-glob`.

It goes wrong when `resolve.tsconfigPaths` is on (off by default), the importer belongs to a tsconfig with `paths` or `baseUrl`, and the call carries any `custom` payload at all. Passing metadata between plugins through `custom` is ordinary Rollup practice, so nothing about the calling side looks unusual — the caller just gets a path that was never checked against the filesystem.

Root cause: the marker was derived from whether `custom` was present, not from whether the key was in it. The decoded payload (`BindingVitePluginCustom`, which declares only the one key) keeps its outer value non-empty when the key is absent, and the conversion into the native custom-field map tested that outer value.

Nothing that goes through the bundler is affected: the `ctx.resolve` boundary (`BindingPluginContextResolveOptions`) already keys off the field itself. Only direct hook calls are — Vite dev, `createIdResolver`, the deps scan.

**Fix:** key off `custom['vite:import-glob']`, matching the `ctx.resolve` boundary.

## 2. A resolve result hides the `package.json` the resolver found

A successful resolve reports the id and nothing about which `package.json` governs it, even though the Vite resolver looks that up on its main resolution path and rolldown needs it to decide whether the file is ESM or CJS.

```js
await resolveId('some-dual-pkg', importer, {})
// before: { id: '/app/node_modules/some-dual-pkg/index.mjs' }
// after:  { id: '…/index.mjs', packageJsonPath: '/app/node_modules/some-dual-pkg/package.json' }
```

Root cause: `HookResolveIdOutput::package_json_path` had no counterpart on the callable output struct, so the conversion dropped it. Every other route keeps it — the resolver fills it in (`resolver.rs`, `normalize_oxc_resolver_result`), rolldown feeds it to module-format inference (`resolve_id_with_plugins.rs`), and a JS plugin's own resolve result carries it back in (`bindingify-build-hooks.ts`).

Nothing observable changes for existing callers; a caller that forwards the result back into rolldown gains the ability to preserve what the resolver already computed, instead of forcing it to be re-inferred from the id.

**Fix:** carry `packageJsonPath` on the callable resolve result.

## 3. A failing transform blames an options object instead of the file

```js
// builtin:vite-json transform on malformed JSON
// before: { plugin: 'builtin:vite-json', hook: 'transform', id: { moduleType: 'json' } }
// after:  { plugin: 'builtin:vite-json', hook: 'transform', id: '/app/src/data.json' }
```

Root cause: the wrapper that turns a builtin plugin into a callable one reads the module id out of the hook arguments when reporting an error, and took the third argument. The hook is called `(code, id, options)`. Wrong since #5632, which introduced the error metadata.

**Fix:** read the second argument.

## Not fixed here

Three more asymmetries on the same boundary have no producer or consumer today, so fixing them would add API surface nothing exercises: `HookLoadOutput::module_type` and `HookResolveIdOutput::normalize_external_id` are dropped on the way out (no callable plugin sets either); `load`'s `asserted_module_type` and `module_idx` are hardcoded with no JS-side parameter; and the transform context is built once in the constructor with an empty id and source (no callable transform plugin reads it). Happy to fold any of them in if you'd rather have the boundary lossless by construction.

## Tests

Three tests, each run red on `main` and green with this PR — the Rust change was reverted, rebuilt, and the suite re-run to confirm, then restored:

- `vite-resolve.test.ts` — against a new `fixtures/tsconfig-paths` fixture (`@/*` -> `./src/*`), `@/dir/a` must resolve to `src/dir/a.js` both with and without an unrelated `custom` payload, while `custom['vite:import-glob']` still yields the raw mapping `src/dir/a`. Red form was exactly the unresolved `src/dir/a` for the unrelated-`custom` call.
- `vite-resolve.test.ts` — the resolve result reports `packageJsonPath` (`undefined` before).
- `vite-json.test.ts` (new file) — a `builtin:vite-json` transform error carries `plugin`, `hook` and the module id, plus a baseline case pinning that a valid JSON transform still returns a module.

Regression evidence: the full `rolldown` Node suite passes (885 tests; one watch-mode CLI e2e case flaked once under full-suite load and passes on its own). `cargo clippy -p rolldown_binding --all-targets -- --deny warnings`, `cargo fmt --check`, `cargo check --workspace --all-features --all-targets` and `vp check` + knip are clean.

Not verified: Windows, and the Rust test suite — `just test-rust` excludes `rolldown_binding`, which is the only crate this touches.

The `packages/rolldown/src/binding.d.cts` hunk is regenerated napi output from `just build-rolldown`.


hyfdev:
### Merge activity

* **Jul 27, 9:36 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.

shulaoda:
I just draft it first, I will make a deep review when I have spare review time

PR #10484, opened 2026-07-27
fix(plugin): rank plugin timings by what the bundler can measure

IWANABETHATGUY:
Supersedes #10458, which this now contains. Rebased onto `main` and squashed into one commit — the stack's history was a record of an investigation that ends by deleting most of what it built, which is honest but a poor thing to review.

## Problem

`[PLUGIN_TIMINGS]` lists one row per plugin, ranked by the elapsed time between dispatching a hook call and its result arriving back. For the hooks the core dispatches concurrently — `resolveId`, `load`, `transform`, `moduleParsed`, `renderChunk` — that number is mostly **queue wait**, and the queue is deepest behind whichever callback is doing the most work.

The failure is not noise, it is a sign inversion:

- A hook that blocks the JS thread creates the backlog that dilutes its own credit.
- A cheap hook called thousands of times sits in that backlog, collecting credit for work it is not doing.

At hand-checkable scale: one callback holds the thread for 100ms while a hundred trivial ones queue behind it and then drain in 1ms total. Truth — the expensive one is 99% of the work. Measured — every queued call's stopwatch ran for the whole 100ms it spent waiting, so the trivial hook appears to cost a hundred times what the expensive one did.

On a real build, gated A/B so the cost reads straight off the wall clock, a hook worth ~15s ranked *below* one worth ~0s.

## Why no arithmetic fixes it

The driver brackets *dispatch* and *completion*. The callback's execution **start** lives in the JS event loop and is never sampled, so two hooks with identical measurements can have entirely different real costs. Two schemes were built on this branch and both failed in the same direction:

- **Apportion a phase's wall clock by each hook's measured share.** A pure linear rescale — every row moves by the same factor, so it cannot reorder anything.
- **A processor-sharing clock** charging each in-flight call `dt / n`. Conserves the total exactly, but Node is run-to-completion: one callback runs at 100% while the others run at 0%, so `dt / n` charges the *running* callback a `1/n` sliver of what it is actually consuming.

Measuring these needs a timestamp inside the JS wrapper, where queue wait is excluded by construction. That is a separate change.

## Approach

Report only the hooks the core invokes **one at a time**, where dispatch-to-completion *is* execution and the measurement stands on its own. `HookKind::is_serially_invoked` is an exhaustive match, so a hook added later has to answer the question rather than default into being reported, and `HookTimingCollector::record` drops an unmeasurable hook even if a call site forgets to check.

The warning still fires on total plugin cost. A build slowed entirely by `transform` still warns — with nothing itemised, and a pointer at `node --cpu-prof`. Staying silent would hide a real problem; naming a plugin from unusable measurements would point at the wrong one.

## What gets better

The rows that survive are more useful than the rows that went away:

- **Per-hook breakdown.** Rows no longer stop at the plugin.
- **`output.advancedChunks` `groups[].name` and `groups[].test` are timed**, under `output options`. The Rust core calls these user callbacks directly, so they have never appeared under any plugin — and a chunk-name classifier can dominate a build. These are serial, so their numbers are trustworthy; they were the one part of the original report that ground truth confirmed.
- **`transformAst` is timed** instead of being folded into its caller.
- **Rows key on plugin identity, not display name.** Two registrations of one plugin stay separate culprits instead of fusing into a row neither earned.
- **Builtin plugins count toward the build total but never get a row** — a user cannot act on `builtin:` work.

Watch and dev hooks (`watchChange`, `closeWatcher`, `hotUpdate`) are not timed at all: they fire between builds, so they have no build wall clock to be a share of.

## Tests

- `a_queued_hook_is_never_credited_over_the_one_blocking_it` — the fixture above, asserting that neither hook is reported so neither can outrank the other.
- `every_hook_is_classified_deliberately` — pins each of the 22 kinds to the serial/concurrent answer its call site was read for.
- `same_named_plugins_stay_distinct_owners`, `unregistered_plugins_get_no_row`, `serial_hooks_are_measured_and_reported`, `clear_keeps_registrations_and_drops_measurements`.
- `still_warns_when_nothing_can_be_attributed` — the empty-rows path renders a warning rather than a header with nothing under it.

## Docs

`checks-plugin-timings.md` now states which hooks are absent and why, rather than leaving a reader to assume the report is exhaustive.

## For reviewers

The judgement call worth checking is the serial/concurrent classification in `hook_kind.rs` — each `true` is a claim about a call site. `resolveFileUrl` (nested `for` loops awaiting each call) and the code-splitting pair (one serial loop) are the two least obvious.


IWANABETHATGUY:
Superseded by #10508 and #10509, which take a different approach to the same problem.

Both of these tried to salvage the Rust-side measurement — this one by apportioning phase
wall clocks, #10484 by restricting the report to serially-invoked hooks. Ground-truth A/B on
a real build showed neither could rank correctly: the callback's execution *start* lives in
the JS event loop and is never sampled from Rust, so no arithmetic over dispatch/completion
pairs recovers it.

The replacement measures inside the JavaScript callback and hands the result back to Rust,
which still owns the gate and the diagnostic. Everything worth keeping came across:
per-hook rows, plugin identity as the row key, and the `codeSplitting groups[].name`
classifier — plus sixteen more option callbacks that were never timed at all.

Closing to keep the queue honest; the discussion here is still readable.

PR #10485, opened 2026-07-27
fix(code-splitting): avoid creating call-site-triggered entry facades

hyfdev:
With strict execution order, a dynamically imported entry can still emit a one-purpose facade file even when every live import can initialize it directly. This PR removes those redundant files in both strict modes while preserving the value and initialization observed by the importing promise.

```js
// before: the import loads a separate file whose only effect is activation
await import("./page-b2.js");
// page-b2.js
import { t as init_page_b } from "./page-b.js";
init_page_b();

// after: the import activates the implementation directly
await import("./page-b.js").then((n) => (n.t(), n.n));
// page-b2.js is not emitted
```

## Problem

Each affected dynamic entry adds one emitted chunk and an extra module load without adding user code. On LobeHub wrap-all, #10456 still emits 1577 chunks, 59 more than the non-strict build; 58 of those are these activation-only facades.

The extra file appears when strict execution order wraps a pure dynamic entry whose implementation still owns its entry chunk. It can happen in both wrap-all and on-demand modes; wrap-all reaches it more often because it wraps every eligible module.

Strict lowering has two ways for such a facade to survive. The restore pass (which revives a facade removed by chunk optimization) already uses the call-site trigger policy from #10456. The creation pass, `create_order_wrap_entry_facades` (which splits an inline entry initializer into its own file), still decides from chunk loading alone and never asks whether every live `import()` can carry that initializer. The same output shape therefore keeps or removes a facade depending on which pass produced it.

Collapsing the facade also means the implementation chunk must publish the namespace value that the facade used to return. That interface cannot be reconstructed from a stale retained-symbol set: aliases can widen it beyond the dynamic import's actual export usage, and a getter such as `x: () => x` can be rendered even when `x` is neither declared nor imported in the chunk. Reading or spreading that namespace then throws a `ReferenceError`.

Non-strict builds are unaffected. Strict builds still retain a real facade for user or emitted entries, entries without a live `import()` call site, two-argument `import()`, TLA-tainted targets, host namespaces that may expose callable `then`, entries whose direct or transitive export-star chain reaches an external module, and facades already restored because another collapse proof failed.

For the dynamic entries that do collapse, output-hook metadata changes with the file topology: the implementation is reported as a common chunk, so `isDynamicEntry` becomes `false`, `facadeModuleId` becomes `null`, and `exports` reports the common chunk's emitted linkage names instead of the facade's entry export list. Manifest and plugin consumers that key on entry metadata will therefore stop seeing a separate dynamic entry for exactly these removed facades; this is the same contract choice already made for optimizer-removed facades in #10456.

## Fix

Use the same call-site proof when creating and restoring strict entry facades. When a pure dynamic entry has an initializer and every surviving `import()` can be rewritten safely, its implementation becomes a common chunk, each import points to that chunk and runs the initializer before resolving, and no separate facade is emitted.

Removing the facade also removes the namespace object it would have exported. Strict lowering now records the exact export names selected by `DynamicImportExportsUsage` and materializes that narrowed interface in the implementation chunk. Its synthetic statement references every non-inlined binding behind those getters, allowing normal cross-chunk linking to keep, export, and import each getter target after the entry becomes a common chunk. Smart inline constants omit that reference only when the individual binding is actually safe to inline. If the namespace also has a real semantic consumer, that complete interface overrides the narrowing. This keeps the `ExportAll` runtime demand without turning a synthetic narrow namespace into an opaque user namespace read or initializing unrelated re-exports.

Entry-level external stars are rendered on the facade chunk and keep that facade in every format. ESM would lose its chunk-level `export * from "external"` statement after reclassification. CJS-like formats would replace the facade's deduplicated `Object.keys` merge with one module-local `__reExport` per direct record; that differs for primitive module values and duplicate records, while a transitive external star is not available from the entry module's direct records at all. The provisional chunk's `entry_level_external_module_idx` covers both direct and transitive chains, so one format-independent guard preserves the complete interface and its existing merge behavior.

This deliberately takes the same ordering tradeoff as #10456: the initializer runs in the rewritten import's promise continuation, one microtask after the host chunk settles, instead of during facade-module evaluation. The importing promise still resolves only after initialization, but a microtask queued during host evaluation can observe the target before it runs. Keeping the facade would preserve that timing but keep all 58 files; using different timing solely based on which lowering pass produced the facade would be inconsistent.

Entries with no `import()` call site, including the remaining LobeHub HTML entry, are intentionally unchanged. LobeHub therefore moves from 1577 to 1519 chunks rather than to the 1518-chunk non-strict baseline.

The analogous external-star loss in the pre-existing non-strict chunk optimizer is not changed here. It already reproduces on `main` without strict execution order and is being fixed independently in #10492, rather than adding another dependency to this stack.

## Validation

- `retained_star_renamed_cycle` now spreads the emitted synthetic namespace at runtime and asserts its exact keys are `['_', 'common']`. This fails on the reviewed revision because the namespace contains a dangling `x` getter and an alias-widened `commonLeaf` getter.
- New `retained_star_opaque_cycle` spreads the source-level dynamic namespace and asserts the complete interface. Its snapshots pin the cross-chunk import that backs `x`, so a genuine semantic namespace consumer cannot be accidentally narrowed.
- New `retained_star_smart_constant_cycle` pins the smart inline-constant boundary in both strict modes: a getter's backing reference is omitted only when that exact constant can actually be inlined.
- The direct and transitive ESM external-star fixtures are red without the guard because the resolved `import()` namespace loses `externalValue`. With the guard, each snapshot retains a facade carrying `export * from "external"`, and runtime assertions observe `x === 1`, `externalValue === 42`, and the expected initialization order.
- New `cjs_dynamic_entry_external_star_keeps_facade` pins why direct CJS stars also keep the facade: a duplicated Proxy external is enumerated once, while a primitive external contributes its enumerable `0` and `1` keys. A collapsed module-local namespace would enumerate the Proxy twice and omit the primitive keys.
- New `cjs_dynamic_entry_transitive_external_star` reproduces the reported `target -> barrel -> external` loss. It is red before the guard with `undefined !== 42`; the fixed snapshot retains the facade's chunk-level external merge and the runtime assertion observes `externalValue === 42`.
- The related retained-star, fanout, namespace, and external-star fixtures pass, as do all 12 strict-order invariant tests.
- LobeHub wrap-all emits 1518 `.js` files and one `.mjs`; all 1519 pass `node --check`. Expanding the external-star guard to every format does not change this ESM measurement, so all 58 chunk reductions remain. This measurement covers production output topology and syntax, not browser E2E behavior.
- The output-hook metadata change is source-verified but is not pinned by a dedicated plugin or manifest fixture in this PR.
- Targeted clippy for Rolldown, the all-features workspace check, Rust and repository formatting, filename lint, and `git diff --check` pass. Full-workspace clippy remains locally blocked by the pre-existing reasonless `#[ignore]` generated in the isolated-declaration test.

Part of #10294.


hyfdev:
* **#10485** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10485?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10485?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10456** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10456?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10433** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10433?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10485).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hyfdev:
Reviewed the collapse end to end — the create/restore interplay, init ordering at the rewritten call sites, the semantic-vs-simulated namespace split, and the TLA / two-arg `import()` / callable-`then` exclusions all check out. One correctness regression needs fixing before this merges: **in ESM output, collapsing a dynamic entry that has `export * from 'external'` silently drops the external's exports from the `import()` namespace.**

### Repro

ESM twin of this PR's own `cjs_collapsed_dynamic_entry_external_star` fixture — same modules, same config minus `"format": "cjs"`:

<details>
<summary>Fixture files</summary>

```js
// main.js
export const loaded = import('./target.js').then((m) => m);

// target.js
import { read } from './reader.js';
globalThis.__externalStarLog.push('target');
export const x = 1;
read();
export * from 'external';

// reader.js
import { x } from './target.js';
export function read() {
  globalThis.__externalStarLog.push(`reader:${x}`);
}

// node_modules/external/index.js — only resolved at runtime ("external": ["external"])
export const externalValue = 42;
```

Config: `strictExecutionOrder: true` (wrap-all), `preserveEntrySignatures: "allow-extension"`, `external: ["external"]`, one manual group for `reader.js` so the implementation chunk is imported by another chunk and the entry becomes a facade candidate.

</details>

On the base branch, the create pass still splits a facade, and the facade is what carries the external star at chunk level:

```js
// target2.js (facade, base)
import { n as x, t as init_target } from "./target.js";
export * from "external";
init_target();
export { x };
```

```js
const ns = await main.loaded;
ns.externalValue; // 42 ✓
```

With this PR the collapse is taken and `external` vanishes from the entire output — no facade, no chunk-level `export *`, and no `__reExport` on the simulated namespace:

```js
// main.js (this PR)
loaded = import("./target.js").then((n) => (n.t(), n.n)).then((m) => m);

// target.js (this PR) — the simulated namespace only has the local export
var target_exports = /* @__PURE__ */ __exportAll({ x: () => 1 });
```

```js
const ns = await main.loaded;
ns.x;             // 1 ✓ — the collapse and init ordering themselves work
ns.externalValue; // undefined ✗
```

### Root cause

1. Pass 1 of `find_entry_level_external_module` runs while the entry chunk is still `ChunkKind::EntryPoint`. It marks the star record `ImportRecordMeta::EntryLevelExternal` (`code_splitting.rs:717`; the flag has no remove site) and re-propagates `has_dynamic_exports` to `false` — on the contract that the *entry chunk file* will emit a chunk-level `export * from "external"`.
2. The collapse then converts that chunk to `ChunkKind::Common` (`order_wrapping.rs:446`). The pass-2 recompute only walks `EntryPoint` chunks, so nothing emits the chunk-level star anymore.
3. The runtime-merge fallback is also off: for ESM, `ns_star_external_re_export_emitted` (`linking_metadata.rs:156`) returns `!EntryLevelExternal || Unknown`. The entry's reason set has neither `Unknown` nor is `SimulateFacadeChunk` consulted — the create path never inserts a reason at all — so both the gate's `ReExport` helper demand (`order_wrapping.rs:436`) and the finalizer's `__reExport(ns, ext)` are skipped.
4. CJS is unaffected because that predicate is unconditionally `true` there — which is exactly why the CJS fixture in this PR passes while its ESM twin fails.

### Not introduced here, but extended

The same hole already ships on `main` through the optimizer path: put `target.js` + `reader.js` into one manual group (empty facade → eliminated → `.then(n => n.ns)` rewrite) and the identical assertion fails on current `main` with `strictExecutionOrder` off entirely. So the root cause predates this PR; what is new is that the create path now reaches it for entry chunks that still own their implementation. The `main`-side half deserves its own issue.

### Suggested fix

Minimal and safe for this PR: have the collapse gate reject the entry when it carries an external star record for which `ns_star_external_re_export_emitted` is `false` (ESM + `EntryLevelExternal` without `Unknown`), keeping the facade. Pass 1 has already run by gate time, so the fact is queryable in place, and the facade is the only shape that carries the chunk-level `export *` without a runtime merge. This is the same trade the gate already makes for two-arg `import()`, TLA taint, and callable-`then` hosts: the rare shape keeps one extra file.

The thorough alternative — teach `ns_star_external_re_export_emitted` to honor `SimulateFacadeChunk` and insert that reason in the create path (as `chunk_optimizer.rs:1157` does on the optimizer path), forcing the runtime `__reExport` merge — would fix the pre-existing `main` hole too. But that predicate is also consumed by `runtime_module_sweep.rs:136`, and the `has_dynamic_exports = false` flattening would need a downstream audit, so it fits better as its own change.

### Minor

`cjs_collapsed_dynamic_entry_external_star/_test.mjs` only reads `{ x }` at runtime; adding a runtime read of an external-provided name would pin the CJS external-star guarantee beyond the snapshot.

hyfdev:
Re-reviewed the revision (f3a205714 → 089a0d681). **The fix is verified** — the ESM repro from my previous comment now keeps the facade and passes at runtime, and both new fixtures pass with `target2.js` carrying `export * from "external"`.

What the verdict rests on:

- **Gate-time freshness.** `entry_level_external_module_idx` has exactly one writer (`find_entry_level_external_module`: clear at `code_splitting.rs:653`, assign at `:731`). Pass 1 runs before `apply_order_wraps` (`finalize_chunk_plan.rs:27` vs `:33`) and pass 2 only after it returns, so both create-pass call sites (`order_wrapping.rs:139` empty-plan path, `:205` main path) read fresh pass-1 state. Facades revived by the restore pass carry empty vecs, but they are already rejected earlier in the same `&&` chain (`modules.contains` + `EntryPoint { meta == DynamicImported }`), so the new condition never decides for them.
- **The chunk-level fact is the right guard, and stronger than a per-record check.** The collapse removes exactly one emission — the chunk-level `export * from` render — and that render iterates the very vec the gate tests (`esm.rs:57-88`), computed by the same pass, so the guard is a complete proxy by construction. The transitive fixture pins the shape a per-record check on the entry module would miss: `target.js` has no direct external record; `barrel.js` owns the only `export * from 'external'`. Non-star routes (star over an internal dynamic-exports barrel, star over internal CJS, `export * as ns`) are module-body emissions in the implementation chunk and survive the collapse.
- **Format scoping is right.** IIFE/UMD cannot reach the collapse at all — multi-chunk output is rejected in `validate_options_for_multi_chunk_output`, and a dynamic entry implies a second chunk. CJS keeps collapsing and the strengthened fixture now runtime-verifies the external merge (`namespace.externalValue === 42`).
- The strict restore path and the non-strict optimizer path still reproduce the pre-existing loss, as the PR body now documents — consistent with tracking that half separately.

One remaining nit: the revision leaves `internal-docs/` untouched. The rejection list in `code-splitting/design.md` (L62-64) and the retain list in `implementation.md` (L102) — both authored by this PR — don't mention the new ESM external-star exclusion. When adding it, word it as create-path-only (or note the restore-path status), since the restore/optimizer path intentionally keeps the pre-existing behavior until the separate fix lands.

IWANABETHATGUY:
Went through the create-path collapse and most of it holds up, but I hit one correctness regression while checking the new snapshots: **a collapsed entry's simulated namespace can emit getters over bindings that no longer exist in the output, so touching the namespace throws.**

This PR's own updated `retained_star_renamed_cycle/artifacts.snap` pins it. `page-a.js` now emits:

```js
var page_a_exports = /* @__PURE__ */ __exportAll({
	_: () => _,
	common: () => common,
	commonLeaf: () => common,
	x: () => x
});
```

There is no `x` in that chunk — not declared, not imported. It comes from `bridge.js`, which sits in the page-b chunk, and in the new output its declaration is gone from there as well (the old output still carried `function x() { return 1; }` in `page-b.js`).

I ran the emitted chunk from the snapshot as-is:

```js
const ns = (await import('./page-a.js')).n;
ns.x;       // ReferenceError: x is not defined
({ ...ns }); // ReferenceError: x is not defined
```

So any consumer that reads `.x` — or just spreads/enumerates the resolved namespace, which is a pretty ordinary thing to do — throws. On the base branch this was `undefined` (the `page-a2.js` facade exported only `_` and `common`). CI stays green because the fixture's `_test.mjs` only touches `pageA.common` / `pageA._`.

What I believe is happening: the getter map is built from `canonical_exports` filtered by `retained_export_symbols` (`module_finalizers/mod.rs:770-787`), a link-time set. Before this PR, the liveness backing those getters ran through the facade — a real `EntryPoint` chunk computes entry-signature exports (`compute_cross_chunk_links.rs:962-1033`), which created the cross-chunk demand that kept the star-re-exported `x` alive and exported from the page-b chunk. After the collapse the chunk is `Common`, so that branch never runs; the absorbed-facade branch only publishes the namespace ref and `init_*` (`compute_cross_chunk_links.rs:1048-1074`), and the synthetic statement is registered with `referenced_symbols: vec![]` (`order_wrap_state.rs:259-265`). Nothing imports `x` into the page-a chunk or keeps its declaration alive in page-b, yet the finalizer still renders the getter from the stale retained set.

Worth noting the gate never consults `DynamicImportExportsUsage`, so source code that does `{ ...await import('./page-a.js') }` still takes the collapse and still throws — this isn't limited to exotic consumers.

Smaller observation, same mechanism: the map includes `commonLeaf` but silently drops `underscoreLeaf`, even though both canonically resolve to retained symbols (`common` / `_`). Membership depends on whether the export's own `symbol_ref` (rather than its canonical form) happens to be in `retained_export_symbols`, so the emitted interface is neither the old facade's list (`_`, `common`) nor the complete one.

IWANABETHATGUY:
> Went through the create-path collapse and most of it holds up, but I hit one correctness regression while checking the new snapshots: **a collapsed entry's simulated namespace can emit getters over bindings that no longer exist in the output, so touching the namespace throws.**
> 
> This PR's own updated `retained_star_renamed_cycle/artifacts.snap` pins it. `page-a.js` now emits:
> 
> ```js
> var page_a_exports = /* @__PURE__ */ __exportAll({
> 	_: () => _,
> 	common: () => common,
> 	commonLeaf: () => common,
> 	x: () => x
> });
> ```
> 
> There is no `x` in that chunk — not declared, not imported. It comes from `bridge.js`, which sits in the page-b chunk, and in the new output its declaration is gone from there as well (the old output still carried `function x() { return 1; }` in `page-b.js`).
> 
> I ran the emitted chunk from the snapshot as-is:
> 
> ```js
> const ns = (await import('./page-a.js')).n;
> ns.x;       // ReferenceError: x is not defined
> ({ ...ns }); // ReferenceError: x is not defined
> ```
> 
> So any consumer that reads `.x` — or just spreads/enumerates the resolved namespace, which is a pretty ordinary thing to do — throws. On the base branch this was `undefined` (the `page-a2.js` facade exported only `_` and `common`). CI stays green because the fixture's `_test.mjs` only touches `pageA.common` / `pageA._`.
> 
> What I believe is happening: the getter map is built from `canonical_exports` filtered by `retained_export_symbols` (`module_finalizers/mod.rs:770-787`), a link-time set. Before this PR, the liveness backing those getters ran through the facade — a real `EntryPoint` chunk computes entry-signature exports (`compute_cross_chunk_links.rs:962-1033`), which created the cross-chunk demand that kept the star-re-exported `x` alive and exported from the page-b chunk. After the collapse the chunk is `Common`, so that branch never runs; the absorbed-facade branch only publishes the namespace ref and `init_*` (`compute_cross_chunk_links.rs:1048-1074`), and the synthetic statement is registered with `referenced_symbols: vec![]` (`order_wrap_state.rs:259-265`). Nothing imports `x` into the page-a chunk or keeps its declaration alive in page-b, yet the finalizer still renders the getter from the stale retained set.
> 
> Worth noting the gate never consults `DynamicImportExportsUsage`, so source code that does `{ ...await import('./page-a.js') }` still takes the collapse and still throws — this isn't limited to exotic consumers.
> 
> Smaller observation, same mechanism: the map includes `commonLeaf` but silently drops `underscoreLeaf`, even though both canonically resolve to retained symbols (`common` / `_`). Membership depends on whether the export's own `symbol_ref` (rather than its canonical form) happens to be in `retained_export_symbols`, so the emitted interface is neither the old facade's list (`_`, `common`) nor the complete one.

I guess it should be fixed in https://github.com/rolldown/rolldown/pull/10492#issuecomment-5099701501, but it is not stacked on the current branch

hyfdev:
@IWANABETHATGUY Fixed in this PR already. I'll check what https://github.com/rolldown/rolldown/pull/10492 needs to be done separately.

---edited

#10492 fixes a separate optimizer issue. No need to connect to this PR.

IWANABETHATGUY:
Found a second silent-loss case while poking at the collapse: **in CJS output, a collapsed dynamic entry that reaches an external star only transitively (`target` → `export * from './barrel.js'` → `export * from 'external'`) loses the external's exports from the resolved `import()` namespace.**

The ESM guard at `order_wrapping.rs:429-430` is format-scoped, so CJS collapses unconditionally — which is fine for the direct case, because the simulated namespace compensates via `__reExport(target_exports, require("external"))`. But that compensation (the helper check at `order_wrapping.rs:442-448` and the CJS arm of the namespace renderer in `module_finalizers/mod.rs`) only iterates the entry's own `star_exports_from_external_modules`. The transitive fact lives in the chunk-level `entry_level_external_module_idx`, and in CJS that is only rendered by the `ChunkKind::EntryPoint` branch of `render_chunk_exports.rs` — the `Object.keys(require("external")).forEach(...)` merge the facade used to carry. The collapse turns the chunk into `Common`, so that render disappears, and the barrel chain can't step in either: pass 1 of `find_entry_level_external_module` already re-propagated `has_dynamic_exports = false` on the contract that the entry chunk file would carry the star, so the barrel's namespace is never included. Net result: `require("external")` appears nowhere in the output.

On this branch the namespace comes out as `__exportAll({ x: () => 1 })` and `namespace.externalValue` is `undefined`; on the merge-base the same graph keeps the facade with the chunk-level external merge and resolves `42`. The restore path has no equivalent condition either, though that half predates this PR.

Here's the repro as a ready-to-drop-in integration fixture — the CJS twin of this PR's own `esm_dynamic_entry_transitive_external_star_keeps_facade`. It currently fails with `undefined !== 42` and passes on the merge-base. I deliberately left out `artifacts.snap`, since its shape depends on which way the fix goes (facade retention vs. reconstructing the interface):

`crates/rolldown/tests/rolldown/function/experimental/strict_execution_order/cjs_dynamic_entry_transitive_external_star/`

<details>
<summary>Fixture files</summary>

`_config.json`

```json
{
  "_comment": "A wrapped CJS dynamic entry can reach an external star only transitively (target -> export * barrel -> export * external). The exports of that external must stay observable on the resolved import() namespace: the entry-level external fact covers the whole chain, while the simulated-namespace merge only sees the entry's direct records, so collapsing the facade must not drop the transitive external interface.",
  "configName": "wrap-all",
  "config": {
    "external": ["external"],
    "format": "cjs",
    "preserveEntrySignatures": "allow-extension",
    "strictExecutionOrder": true,
    "codeSplitting": {
      "includeDependenciesRecursively": false,
      "groups": [
        {
          "name": "reader",
          "test": "[\\\\/]reader\\.js$"
        }
      ]
    },
    "experimental": {
      "attachDebugInfo": "full"
    }
  },
  "hiddenRuntimeModule": false
}
```

`main.js`

```js
export const loaded = import('./target.js').then((namespace) => namespace);
```

`target.js`

```js
import { read } from './reader.js';

globalThis.__externalStarLog.push('target');

export const x = 1;
read();

export * from './barrel.js';
```

`barrel.js`

```js
export * from 'external';
```

`reader.js`

```js
import { x } from './target.js';

export function read() {
  globalThis.__externalStarLog.push(`reader:${x}`);
}
```

`node_modules/external/package.json`

```json
{
  "name": "external",
  "main": "./index.js"
}
```

`node_modules/external/index.js`

```js
module.exports = { externalValue: 42 };
```

`_test.mjs`

```js
import { createRequire } from 'node:module';

globalThis.__externalStarLog = [];

const main = require('./dist/main.js');
const namespace = await main.loaded;
assert.strictEqual(namespace.x, 1);
assert.strictEqual(
  namespace.externalValue,
  42,
  'transitive external star exports must survive on the resolved dynamic-import namespace',
);
assert.deepStrictEqual(globalThis.__externalStarLog, ['target', 'reader:1']);

delete globalThis.__externalStarLog;
```

</details>

Run with `cargo test -p rolldown --test integration cjs_dynamic_entry_transitive_external_star`.

One thing I checked so it doesn't get assumed away: #10492 does **not** cover this. I merged it onto this branch locally and the fixture still fails — its `mark_simulated_facade_export_star_chain` / `entry_level_external_star_keeps_dynamic_exports` machinery is only invoked from the chunk optimizer's elimination loop, and the create-path collapse here never sets `SimulateFacadeChunk` on the chain. Same result with #10488 stacked on top as well.

Unrelated heads-up from that experiment: #10488 and this PR merge cleanly but don't compile together — #10488 removes `OrderSyntheticStmt.owner` while `insert_simulated_facade_namespace` here constructs it (`E0560` at `order_wrap_state.rs:330`).

hyfdev:
https://github.com/rolldown/rolldown/pull/10488 is for another situation. let's focus on this one and see how to integrate https://github.com/rolldown/rolldown/pull/10488 after #10485.

IWANABETHATGUY (approved):

hyfdev:
### Merge activity

* **Jul 28, 5:19 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 28, 5:19 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 28, 5:33 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10488, opened 2026-07-27
fix(code-splitting): route CJS barrel initialization per consumer

hyfdev:
With `strictExecutionOrder`, importing one binding from an otherwise re-export-only barrel could initialize unrelated CommonJS re-exports and pull lazy-route code into the initial load. This PR makes each consumer initialize only the exports it can statically reach while preserving complete namespace and cycle semantics.

```js
// barrel.js
export { cn } from './cn.js';
export { default as cloneDeep } from './clone.cjs';

// main.js
import { cn } from './barrel.js';
export const loadRoute = () => import('./route.js');

// route.js
import { cloneDeep } from './barrel.js';

// Before: loading main.js also initialized clone.cjs.
// After: clone.cjs waits until loadRoute() loads route.js.
```

## Problem

The regression occurs in strict-order builds when tree shaking retains different exports from the same re-export-only barrel for eager and lazy consumers, especially when one re-export needs generated CommonJS interop. Both wrap-all and on-demand strict modes could put the lazy leaf in the eager entry's static closure.

Generated CommonJS interop previously gave the barrel one shared initialization body. Every consumer therefore initialized that body as a unit, and pre-chunk reachability followed the barrel's bundle-wide dependency union instead of the leaves selected by that consumer.

Non-strict builds are unaffected. The narrower routing is limited to tree-shaken source modules made entirely of direct re-exports; shapes whose evaluation cannot be safely split retain the existing behavior.

## Fix

Keep eligible re-export-only barrels as routing points and make generated CommonJS interop independently callable for each re-export record.

Each CommonJS re-export now has a memoized carrier beside its importee. Named imports and statically resolved namespace members select only their leaf and, when needed, its carrier; opaque namespace uses expand that consumer's complete statically known leaf/carrier list without disabling routing for other consumers. Effectful carriers become unconditional only when every forwarding hop retains their side effects, and selected leaves and carriers remain in source order.

The same importer-local resolution drives pre-chunk placement, on-demand cycle projection, cross-chunk registration, and final emission. Non-routing wrappers that materialize a consumer-local namespace initialize its complete ordered target list inside their existing module guard and at the original re-export position. A consumer-local entry instead invokes that list from its entry prologue, or from the rewritten `import()` call site when its dynamic-entry facade is collapsed, so namespace getters cannot observe uninitialized leaves or carriers.

The conservative monolithic path remains for tree-shaking-disabled builds, modules with local executable bodies, synchronous `import`/`require` SCCs, direct opaque CommonJS `require()` targets, top-level await, dynamic exports, CJS `export *`, missing-export shims, and concatenated wrappers. Keeping synchronous SCCs monolithic preserves one evaluating guard across dependency traversal and the module body.

## Validation

The directed fixtures cover the behavior at observable boundaries:

- `cjs_reexport_barrel_consumer_local` and `cjs_reexport_barrel_namespace_routing` are red on `main` and green here for named imports, static namespace members, opaque namespace expansion, and lazy-route isolation.
- `cjs_reexport_barrel_per_record_chunks` and `cjs_reexport_barrel_shared_importee` pin per-record carrier identity and placement, including multiple records that reach the same CommonJS module.
- `cjs_reexport_barrel_effectful_carrier` and `cjs_reexport_barrel_side_effect_boundary` pin source ordering and nested `moduleSideEffects` boundaries.
- `cjs_reexport_barrel_monolithic_require_namespace`, `cjs_reexport_barrel_monolithic_esm_namespace`, and `cjs_reexport_barrel_nested_namespace` were red in the initial implementation and are green in the final revision for complete namespaces above consumer-local routes and nested waypoint reachability.
- `cjs_reexport_barrel_cycle_fallback`, `cjs_reexport_barrel_require_cycle_fallback`, and `cjs_reexport_barrel_require_fallback` lock the conservative path. The require fallback now observes the unrelated CommonJS leaf before the eager entry body and confirms that the later lazy route does not initialize it twice.

On the recorded vue-vben-admin reproduction, initial-load gzip changes from 615,917 B before the fix to 81,271 B in wrap-all and 80,692 B in on-demand; strict mode off is 61,780 B.

Checks run on the final rebased commit:

- strict-execution-order fixtures: 95/95 passed
- strict-execution-order invariant fixtures: 5/5 passed
- strict-execution-order Rust invariant tests: 12/12 passed
- `cargo test -p rolldown --lib`: 104/104 passed
- `cargo clippy -p rolldown --all-targets -- --deny warnings`
- `cargo fmt --all -- --check`
- `just lint-repo`

The order fuzzer also found no new divergence for seeds 1000-1099 across wrap-all/on-demand and mixed/pure-ESM inputs; all four cells retained the known 97/3 baseline. Full Node, WASI, and Vite suites were not run locally and remain covered by CI.

Part of #10294.


hyfdev:
Reviewed this end to end — the problem is real and important, the per-record carrier model with consumer-local routing is the right shape, the fallback taxonomy matches the design doc one-to-one, and the fixture matrix is thorough (all 9 fixtures pass across their config variants here, including a probe where the barrel is simultaneously a dynamic entry with an opaque escape and a static named-import route). The vue-vben-admin numbers make this the key strict-mode adoption fix.

However, there is one defect family the taxonomy does not cover, with three runnable counterexamples — all regressions against `main` (each shape passes on `origin/main`, fails at this PR's head). **A monolithic (non-routing) consumer sitting above a consumer-local barrel needs the barrel's complete namespace, and nothing initializes the carriers (or even the ESM leaves) that its materialized namespace getters reference.**

### Counterexample 1 — `require()` above the barrel

```js
// outer.cjs                          // wrapper.js
const ns = require('./wrapper.js');   export * from 'pure-barrel';
module.exports = {
  cloned: ns.cloneDeep({ a: 1 }),     // pure-barrel/index.js
};                                    export { cn } from './cn.js';
                                      export { cloneDeep } from './clone-deep.cjs';
```

`wrapper.js` is rejected as a direct `require()` target (monolithic), but `pure-barrel/index.js` behind it stays consumer-local. Emitted output:

```js
//#region node_modules/pure-barrel/clone-deep.cjs
var import_clone_deep = void 0;            // carrier namespace declared, init function absent entirely
//#endregion
var init_pure_barrel = __esmMin(() => {
	init_cn();                             // no carrier call
});
var wrapper_exports = /* @__PURE__ */ __exportAll({
	cloneDeep: () => import_clone_deep.cloneDeep,   // TypeError: reading 'cloneDeep' of undefined
	cn: () => cn
});
```

### Counterexample 2 — the family is not require-specific

Same shape, pure ESM, zero `Require` records in the graph: `wrapper.js` is `export * from 'pure-barrel'` plus one side-effect statement (monolithic for a non-require reason), and the consumer does `import * as ns from './wrapper.js'` with a computed-key escape. Identical `TypeError` — and here even the pure-ESM leaf is lost: `init_wrapper` contains only the wrapper's own side effect, so `ns.cn` reads an uninitialized `var cn;` too. This rules out fixing the family by widening the `required_modules` set (transitively or otherwise): this graph has nothing to put in that set.

### Counterexample 3 — nested pure wrappers panic the bundler

Make `wrapper.js` a pure re-export (so wrapper and barrel are both consumer-local) and keep the opaque `import * as ns` escape:

```
panicked at crates/rolldown/src/stages/generate_stage/code_splitting.rs:1035:
Empty bits means the module is not reachable, so it should bail out with is_included: false
… "node_modules/pure-barrel/index.js"
```

The chunk-bits BFS special-cases consumer-local barrels — it enqueues only eager carriers and skips both `load_dependencies` and the waypoint chain — so a consumer-local module consumed only through another consumer-local waypoint keeps `is_included: true` with empty bits. In release builds the `debug_assert` compiles out and chunk assignment proceeds on empty reachability.

### Root cause, unified

The demand "this consumer needs the barrel's complete namespace" loses its connection to init-obligation registration:

- The overlay's direct wrapper reference is removed ("removes the unused direct-wrapper reference from non-empty retained-path overlays"), on the assumption that the resolver registers equivalent targets — but for a bare excluded `export *` record the resolver registers nothing: `import_record_has_live_binding_consumer` (`esm_init_obligations.rs`) iterates `named_imports` only, and a bare `export *` has none. The opaque demand lands on the rendered namespace getters, which the obligation machinery never sees.
- The complete-namespace expansion that would fix this already exists — `collect_wrapped_esm_init_targets_for_module_namespace` — but it is wired only to entry/dynamic-entry prologues.

### Suggested direction

Treat "a materialized namespace references plan symbols" as the registration trigger, wherever it occurs: wire `collect_wrapped_esm_init_targets_for_module_namespace` (or its equivalent) to every monolithic module whose rendered namespace exposes consumer-local leaves/carriers — the require-of-ESM path, monolithic star-re-exporters, and any future opaque consumer then share one rule, mirroring how entry prologues already work. Alternatively, broaden the rejection net from "direct require targets" to "modules whose namespace is demanded by any non-routing consumer", which trades output size for simplicity. The bits BFS separately needs to enqueue waypoint chains (counterexample 3) regardless of which direction the registration fix takes.

All three probe fixtures are two-to-four tiny files each in the shapes shown above (assertions: `cloneDeep({a:1})` deep-equals, `cn('x')`, plus `Object.keys(ns)` completeness) — happy to hand them over to fold into the fixture matrix.

### Logistics

`main` has moved under this branch (#10463 touches the same files), so a rebase is due; and the #10456/#10485 stack overlaps on `order_wrapping.rs` / `order_wrap_state.rs` / `design.md`, with the `esm_init_target` return type now an enum — worth deciding the merge order between this PR and that stack deliberately.

hyfdev:
Re-reviewed the revision (91ed0fb6f → 7bb418751, now rebased onto main with the merged facade-collapse stack). **Verified — all three counterexamples are fixed at the right level, and the fix is the generalized rule rather than three patches:** `consumer_local_namespace_targets` is precomputed for every consumer-local module, monolithic wrappers emit that complete leaf/carrier list at their retained `export *` record positions (per-target live-chunk filtering), and the chunk-bits BFS now enqueues waypoint chains. All three former counterexamples pass both as my original probes and as their new fixture twins (`cjs_reexport_barrel_monolithic_require_namespace` / `..._monolithic_esm_namespace` / `..._nested_namespace`), alongside the v1 fixture set and the adapted stack-interplay fixtures.

Beyond re-running the counterexamples, two deeper shapes were probed and hold:

- **Two `export *` records from one monolithic wrapper into two different consumer-local barrels** (with a leaf shared by both): emitted `init_wrapper` is exactly `init_alpha(); init_barrel_a_cjs_0(); init_shared_leaf(); init_beta(); init_barrel_b_cjs_1(); <own statement>` — each record position carries its barrel's complete list in source order, the shared leaf initializes once, and the runtime event order matches reference ESM semantics (`['alpha','cjs-a','shared','beta','cjs-b','wrapper']`) across wrap-all, on-demand, and CJS output.
- **Collapse interplay with the merged stack** — a consumer-local barrel that is itself a collapse-eligible dynamic entry, in both trigger forms: same-chunk rewrites to `Promise.resolve().then(() => (init_cn(), init_pure_barrel_cjs_0(), pure_barrel_exports))`, and the cross-chunk form exports the carrier init through the host chunk (`.then((n) => (n.o(), n.r(), n.i))`) where it survives deconfliction; a `phase: 'pre-load'` assertion proves the trigger itself (not any static route) runs the carrier. On-demand correctly omits `init_cn` where cn.js is unwrapped and eager — live-state filtering, not a dropped target.

One residual note for the record: the pre-chunk placement probe state is built from the wrap-all structural plan while the final state can be on-demand. Eligibility today depends only on plan-independent facts (cycles, require targets, AST shape, tree-shaking flag) and routes are fixed before the precompute, so the two cannot diverge; every probe above also ran its on-demand variant without tripping the `.expect`. If consumer-local eligibility ever becomes plan-dependent, that invariant needs a dedicated fixture.

With that, the earlier review's blocking findings are all closed from my side.

hyfdev:
### Test review note: `cjs_reexport_barrel_require_fallback`

The dual-use graph is the right fixture (barrel is both ESM-named-imported by main and `require()`’d from a lazy route → plan must keep the monolithic barrel path). The emitted snap already shows that: `init_required_barrel` runs `init_cn()` and `__toESM(require_clone())` on the main entry, with `clone` living in `main2.js`.

The runtime assertions do not lock that property:

- `clone.cjs` has no `__events` log, so the extra monolithic init cost is invisible.
- `_test.mjs` only checks entry events `['cn', 'main:cn']` and that the route still returns the clone result — which reads like a consumer-local success criterion and would still pass if main only initialized `cn` and deferred `clone` (with a matching snap update).

Contrast `cjs_reexport_barrel_monolithic_require_namespace`, which observes both `'cn'` and `'clone-deep'` at materialization time.

**Suggested fix (no new fixture needed):** instrument `clone.cjs` (e.g. push `'clone'`) and assert it runs on the main entry under the require-forced monolithic path; optionally also assert `staticClosure('main.js')` contains the clone module, same pattern as `per_record_chunks` / `side_effect_boundary`.

Other new barrel fixtures look necessary and reasonably aligned with the carrier / routing / fallback invariants; I would not drop or merge them for redundancy.

hyfdev:
Should be the last p1 fix.

IWANABETHATGUY:
Went through this end to end — planner/resolver, finalizer emission, cross-chunk linking, and the whole fixture family. Overall it's really solid. I spent most of my time trying to construct a case where a namespace getter reads an uninitialized carrier (`var import_x = void 0`) and couldn't: every reader path in the snapshots calls the owning carrier first, in both wrap-all and on-demand. Non-strict output is byte-identical to main, and the internal docs actually match what the code does. Nice work.

That said, here's what I'd like to see addressed — 1 and 3 are the ones I care about before merge, the rest can be follow-ups:

1. **The init-companion registration in `compute_cross_chunk_links.rs:704-749` doesn't know about carriers.** A reference that canonicalizes to a carrier namespace registers the barrel's *empty* wrapper as its init companion, never the carrier. This is currently fine only because of a non-local invariant: every consumer that Emit hands a carrier target is itself wrapped or an entry (trivially true in wrap-all; in on-demand only via the plan's backward closure). I couldn't build an input that escapes it — but if one ever exists, the failure is silent: `wrapper_is_reachable_in_chunk` just drops the init call and you get a runtime TypeError. Worse, the comment justifying that filter (`module_finalizers/mod.rs:277-282`, "the barrel's own lowered statements perform the init") is now simply false for consumer-local barrels. Could we consult `cjs_carrier_by_namespace` there as defense-in-depth, and fix the comment either way?

2. The consumer-local entry prologue in `render_chunk_exports.rs:58-84` emits bare `wrapper();` calls with no `debug_assert!(!target.tla_tainted)`, while the single-target path right below awaits TLA targets and the collapsed-rewrite twin pins the same invariant with an assert. Safety currently rests on the plan rejecting TLA barrels — fine today, but cheap to pin at the emission site too.

3. **The new per-record loop in `determine_reachable_modules_for_entry` runs in every build — including non-strict — and does the expensive check first** (`code_splitting.rs:1234-1263`). `import_record_has_live_binding_consumer` runs per record *before* the cheap `resolved_module` / `is_consumer_local_reexport_route` filters. It walks the module's whole `named_imports` map per record, and for each unused/flattened import it exhaustively scans all included stmt-infos. With strict off the state is empty, so 100% of that work is discarded — scaling with entries × modules × records. CodSpeed won't catch this shape (single entry, mostly-live imports). The fix is mechanical: check the importee route first, evaluate `has_live_binding` lazily, and skip the whole loop when the state has no consumer-local routes at all.

4. `consumer_local_target_order` is used as a `sort_by_key` key (`esm_init_obligations.rs:235,390`) — a recursive backtracking DFS with a Vec allocation per call, re-invoked per comparison, and the path enumeration is exponential in the worst case (barrel diamonds with an unresolvable target — exactly what the `vec![usize::MAX]` fallback anticipates). A 500-export barrel consumed opaquely is the realistic bad case. `sort_by_cached_key` + a record→position index + memoizing `(forwarder, target)` would bound it.

5. We recompute a lot per strict build: `pre_chunk_order_state` runs in both `generate_chunks` and `analyze_execution_order`, and `synchronous_cycle_modules` (a full petgraph Tarjan) runs ~3-5× per build across `order_wrapping.rs:462` and `order_analysis.rs:596,666`. Worth computing once and threading through.

6. `retained_star_renamed_cycle/_test.mjs` lost its teeth in the rewrite: it now only checks `Object.keys` and spread-no-throw. If the entry-prologue `init_page_a()` regressed away, `pageA.common` would be `undefined` and the test would still pass. One `assert.equal(ns.common, 'common')` closes it — the sibling fixtures already do this.

7. Five of the fallback triggers in the taxonomy — TLA, dynamic exports, CJS `export *`, missing-export shims, concatenated wrappers — have no fixture that combines them with a CJS re-export barrel, so a regression in those fallback *decisions* would only show up as unrelated snapshot drift. `inlineDynamicImports` is also unexercised. Fine as a follow-up fixture batch, just don't want to lose track of it.

8. The intentionally-empty shared wrapper is still emitted and exported with no importer in the tree-shaken cells — and in `cjs_reexport_barrel_per_record_chunks` it costs an entire extra chunk whose only content is an empty function. Consistent with the design as written, but seems like a natural follow-up to drop it when nothing routes to it.

9. The carrier guard in `ensure_lazy_module_initialization_order` (`code_splitting.rs:466`) is dead code: the function early-returns under strict mode, and carriers only exist under strict mode (they're only inserted via `apply_consumer_local_reexport_plan`, and `analyze_execution_order` returns `None` when strict is off — the probe states are separate locals). I verified this empirically: with the guard commented out, the full `cargo test -p rolldown` corpus (1934 executed tests, incl. all 63 strict-execution-order fixtures) produces byte-identical output — zero `artifacts.snap` changes — and a `debug_assert` planted on the same condition never fires across the corpus. Harmless, but it implies a reachability that doesn't exist — either drop it or leave a comment saying it's future-proofing for if the strict early-return is ever relaxed.

10. Small finalizer things, none load-bearing: the carrier namespace is declared as `var ns = void 0;` where the initializer-free `var ns;` is the safer in-repo precedent for closure-assigned bindings; the `FromCjsWrapKindEntry` branch in `append_order_cjs_carriers` is unreachable (the wrapper is always in the chunk being finalized); the per-iteration `AstBuilder::new` implies a borrow constraint that doesn't exist; and the exhaustion-rule comment in `wrapped_esm_init_stmt_for_import_record` only names one of the two dedup sets it now protects.

For the record, I re-checked the earlier rounds' items (the three monolithic-namespace counterexamples, the wrap-all/on-demand probe caveat, the `require_fallback` instrumentation) — all confirmed closed; everything above is new. CI is green on the head commit; might be worth the `ci: windows` label given this is p1 core, even though the logic is platform-independent.

hyfdev:
@IWANABETHATGUY Thanks for the deep pass — especially for pressuring the carrier interaction from the registration side; the earlier rounds had only covered that invariant from the emission side.

Everything is now either fixed on the head or explicitly deferred. The fixes went in as part of the squashed force-push, so the reviewable delta is `git diff 1bd09b10c..be292203` — 7 files, +51/−31, nothing else on the branch changed.

**Fixed in `be29220`:**

- **Item 1 (registration doesn't know about carriers):** `add_depended_symbol_with_wrapped_esm_init` now consults `order_cjs_carrier_key_for_namespace` first and registers that exact per-record carrier as an additional init companion, through the same `add_wrapped_esm_init_target_depended_symbol` path so live-chunk filtering still applies (`compute_cross_chunk_links.rs:725-735`). Defense-in-depth as you suggested: the emission-side invariant still holds, but a consumer shape that ever escapes it now gets a registered carrier import instead of a silently dropped init. The stale `wrapper_is_reachable_in_chunk` justification comment is rewritten to state the actual rule (every live referenced binding is paired with its init companion; a carrier namespace additionally registers its carrier wrapper).
- **Item 3 (per-record loop cost in every build):** exactly your prescribed mechanical fix. The whole per-statement loop is skipped when the state has no consumer-local routes (`code_splitting.rs:1151,1226`), so non-strict builds pay one `is_empty()` per entry and zero per-record work; within strict builds the cheap `is_consumer_local_reexport_route(importee)` lookup now runs before `import_record_has_live_binding_consumer` (`code_splitting.rs:1239`).
- **Item 2:** the consumer-local prologue path now pins `debug_assert!(!target.tla_tainted)` at the emission site (`render_chunk_exports.rs:66`), matching the single-target path and the collapsed-rewrite twin.
- **Item 6:** `retained_star_renamed_cycle/_test.mjs` now asserts `pageANamespace.common === 'common'`.
- **Item 9:** dropped — the dead guard and the `order_state` parameter are removed from `ensure_lazy_module_initialization_order`; your corpus experiment settled that the future-proofing wasn't buying anything.
- **Item 10 (partial):** the exhaustion-rule comment now names both dedup sets.

**Deferred as follow-ups**, queued for the #10294 ledger: item 4 (`sort_by_cached_key` + record-position index + `(forwarder, target)` memoization — agreed the backtracking `visited.remove` makes it path enumeration on barrel diamonds), item 5 (compute the wrap-all probe analysis and the Tarjan SCC pass once and thread them through), item 7 (fallback-trigger × barrel combination fixtures plus `inlineDynamicImports`; I'd fold one more into that batch: `effectful_carrier` / `side_effect_boundary` have no strict-off cell today, so "declared-false executes the same set as flag-off" is currently analytic rather than pinned), item 8 (drop the intentionally-empty shared wrapper when nothing routes to it), and the rest of item 10 (`var ns = void 0;` → `var ns;`, the unreachable `FromCjsWrapKindEntry` branch).

CI is green on `be29220` (cargo test, Vite Ubuntu, Node dev-server suites; Node Validation finished after your pass), and I've added the `ci: windows` label so the full matrix runs before merge.

IWANABETHATGUY (approved):

hyfdev:
### Merge activity

* **Jul 29, 6:00 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 29, 6:00 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 29, 6:19 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10489, opened 2026-07-27
fix(dev): resolve re-exports from externals through a real import

### LINKED ISSUE #10478: [Bug]: re-exports from external modules do not work in dev module wrappers (HMR patches, lazy chunks)

No public reproduction link is attached, but the behavior is verified empirically at `c50fac0`-equivalent source: lazy chunks were compiled through the dev engine (`experimental.devMode.lazy`) with `node:assert` as the external, and the emitted chunks were executed against the real `DevRuntime`. Steps that reach the broken path:

1. Build in dev mode (`experimental.devMode`), with lazy compilation enabled or with an HMR edit so a module is emitted as a module wrapper (HMR patch or lazy chunk).
2. Inside that module, re-export from an **external** module, in any of the three forms:

```js
export * from 'ext';
export * as ns from 'ext';
export { x } from 'ext';
```

3. Look at the wrapper output for that module. The external is referenced through the dev runtime registry instead of a real import (actual emitted code, trimmed):

```js
// input: export { ok as assertOk } from 'node:assert'  (external)
__rolldown_runtime__.registerFactory(".../lazy.js", "esm", (function(__rolldown_module_id__) {
	try {
		var __rolldown_exports__ = __rolldown_runtime__.__exportAll({ assertOk: () => import_node_assert_00.ok });
		__rolldown_runtime__.registerModule(__rolldown_module_id__, { exports: __rolldown_exports__ });
		const hot_lazy = __rolldown_runtime__.createModuleHotContext(__rolldown_module_id__);
		var import_node_assert_00 = __rolldown_runtime__.loadExports("node:assert");
	} finally {}
}));
```

There is no `import` statement anywhere in the chunk. Executing it against the dev runtime prints `Module node:assert not found` and `exports.assertOk` is `undefined`.

### What is expected?

Re-exports from an external module should work the same way plain imports of that external already do in the same finalizer: emit a real import statement and build the re-export from that binding. For comparison, a plain `import * as A from 'node:assert'` in a wrapper is emitted as `import * as import_node_assert_10 from "node:assert"` (visible in the snapshot added by #10476), and the host resolves it.

Two symptoms, both confirmed by executing the emitted lazy chunks against the real dev runtime:

1. **Empty re-exports.** All three re-export forms emit `var import_ext = loadExports("<external id>")` and no import statement (also no `__toESM` — interop is `None` for externals). External modules are never registered in the dev runtime registry (`registerModule` is only emitted for the wrapped modules themselves), so `loadExports` logs `Module <id> not found` and returns `{}` — it does not throw. Every re-exported name reads as `undefined`; for `export * from`, the resulting exports object has zero keys.

2. **A dead `var` shadows the real import when a module both imports and re-exports the same external.** Both the import arm and the re-export arms name the local binding through `ensure_static_import_info`, keyed by the importee, but they check different sets before emitting (`self.imports` vs `generated_static_import_stmts_from_external`), so both declarations are emitted with one shared name. This is **not** a syntax error (`node --check` passes): the real `import * as` sits at the chunk top level while the `var` sits inside the factory function, so the `var` legally shadows the import inside the wrapper. Every module-body reference to the external is rewritten to `import_X.…` and now reads from the empty `{}`:

```js
// input
import { ok } from 'node:assert';
export * from 'node:assert';
ok(true);

// wrapper output (actual, trimmed)
import * as import_node_assert_00 from "node:assert"; // chunk top level — correct
__rolldown_runtime__.registerFactory(".../lazy.js", "esm", (function(__rolldown_module_id__) {
	try {
		// ...
		var import_node_assert_00 = __rolldown_runtime__.loadExports("node:assert"); // shadows the import
		__rolldown_runtime__.__reExport(__rolldown_exports__, import_node_assert_00);
		import_node_assert_00.ok(true); // TypeError: import_node_assert_00.ok is not a function
	} finally {}
}));
```

### System Info

```Shell
Not environment-specific. Verified empirically at c50fac0-equivalent source:
lazy chunks compiled through the dev engine and executed against the real DevRuntime.
Affects the dev/HMR wrapper output (HMR patches, lazy chunks) on any platform.
```

### Any additional comments?

The import arm branches on the module kind. The two re-export arms do not:

```mermaid
flowchart TD
  A["plain import of an external module"] --> B["ImportDeclaration arm — has a Module::External branch"]
  B --> C["emits a real import statement"]
  C --> OK["works — the host resolves the package"]
  D["any re-export from an external module"] --> E["ExportAllDeclaration and ExportNamedDeclaration arms — no external branch"]
  E --> F["emits var import_ext = loadExports of the external id"]
  F --> BAD["fails — externals are never registered, loadExports warns and returns an empty object"]
```

Code locations (at `c50fac0`):

- Import arm with the external branch: [hmr_ast_finalizer.rs#L120-L135](https://github.com/rolldown/rolldown/blob/c50fac0639724199193b016a15845560e11e8428/crates/rolldown/src/hmr/hmr_ast_finalizer.rs#L120-L135)
- `export * from` / `export * as ns from` arm, no external branch: [hmr_ast_finalizer.rs#L288-L305](https://github.com/rolldown/rolldown/blob/c50fac0639724199193b016a15845560e11e8428/crates/rolldown/src/hmr/hmr_ast_finalizer.rs#L288-L305)
- `export { x } from 'ext'` path, no external branch: [hmr_ast_finalizer.rs#L178](https://github.com/rolldown/rolldown/blob/c50fac0639724199193b016a15845560e11e8428/crates/rolldown/src/hmr/hmr_ast_finalizer.rs#L178)
- `create_load_exports_call_stmt` emits `loadExports` for any module kind: [hmr_ast_finalizer.rs#L398](https://github.com/rolldown/rolldown/blob/c50fac0639724199193b016a15845560e11e8428/crates/rolldown/src/hmr/hmr_ast_finalizer.rs#L398)
- The gap is already marked in the code: [`// TODO reexport external module`](https://github.com/rolldown/rolldown/blob/c50fac0639724199193b016a15845560e11e8428/crates/rolldown/src/hmr/hmr_ast_finalizer.rs#L504)
- `loadExports` on a missing id warns and returns `{}`: [runtime-extra-dev-common.js#L217-L225](https://github.com/rolldown/rolldown/blob/c50fac0639724199193b016a15845560e11e8428/crates/rolldown_plugin_hmr/src/runtime/runtime-extra-dev-common.js#L217-L225)
- Where the hoisted external import statements are appended at chunk top level (which is why symptom 2 is shadowing, not a duplicate-declaration error): [impl_traverse_for_hmr_ast_finalizer.rs#L143](https://github.com/rolldown/rolldown/blob/c50fac0639724199193b016a15845560e11e8428/crates/rolldown/src/hmr/impl_traverse_for_hmr_ast_finalizer.rs#L143)

Scope notes:

- Only the wrapper path is affected. The initial dev bundle and production builds go through scope hoisting, which resolves externals correctly for all forms. Verified in the same run: a statically imported re-exporting module in the initial entry chunk got a real `import { ok as assertOk } from "node:assert"` and worked when executed.
- Found during the review of #10476 and then verified empirically. That PR does not change external handling — its new branch reads the same binding that `export * from` already used. (The verification run, on pre-#10476 source, also reproduced the #10476 bug itself: `export * as NS from 'node:assert'` lowered byte-identical to bare `export *`, dropping `NS`.)

Suggested direction:

- Give the two re-export arms the same `Module::External` branch the import arm has: emit a real import statement and build the re-export from that binding.
- Unify the two duplicate-check sets so the shared binding name is only declared once.

tbvjaos510:
Fixes #10478.

## What's broken

The dev runtime registry holds the modules this build wrapped. An external is by definition not one of them, so `loadExports('<external id>')` finds nothing, warns `Module <id> not found`, and returns `{}` — it does not throw. The import arm of `HmrAstFinalizer` knew that and emitted a real import statement, hoisted outside the wrapper. The three re-export arms asked the registry instead:

```js
// input: export { ok as assertOk } from 'node:assert'
__rolldown_runtime__.registerFactory("lazy.js", "esm", (function(__rolldown_module_id__) {
	var __rolldown_exports__ = __rolldown_runtime__.__exportAll({ assertOk: () => import_node_assert_00.ok });
	// …
	var import_node_assert_00 = __rolldown_runtime__.loadExports("node:assert");  // {}
}));
```

Every re-exported name reads as `undefined`, with nothing failing at build time.

The second symptom is worse. When a module both imports and re-exports the same external, both arms name the binding through `ensure_static_import_info` — so they agree on the name — but they deduplicate against different sets (`self.imports` vs `generated_static_import_stmts_from_external`), so both emit a declaration. The real import sits at chunk top level and the `var` inside the factory, which is legal shadowing rather than a redeclaration error. Every reference in the module body then reads the empty object, including the ones that were working before.

## Root cause

Four statement kinds need the same thing — a local binding for the importee's namespace — and each decided how to obtain it on its own. Only one of the four knew about externals.

## The fix

All four now route through one `create_importee_binding_stmt`: `loadExports` for a normal module, a real import statement for an external. The two deduplication sets become disjoint by construction, so the shadowing `var` has no path to being emitted.

`create_load_exports_call_stmt` now takes a `&NormalModule` rather than a `&Module`, which makes "ask the registry for an external" a compile error rather than a rule to remember. It also removes a `Module::External => None` interop arm that is now unreachable.

## Scope

Only the module-wrapper path (HMR patches, lazy chunks) changes. Scope hoisting already resolved all four forms correctly, so the initial dev bundle and production builds were never affected — in the new fixture's snapshot the eager output and the patch sit side by side, and the patch now matches what the eager output always had.

No existing snapshot moved, which is also the answer to why this survived: no fixture reached this path.

## Tests

- `crates/rolldown/tests/rolldown/topics/hmr/reexport_external/` — all three re-export forms plus the import-and-re-export shadowing case, against `node:assert` and `node:util`. The fixture executes, and it runs its assertions twice against the same expectations: first on the eagerly bundled output, then again once every module has been re-emitted as a wrapper. So the wrapper is held to what scope hoisting already satisfies, instead of to a second set of numbers written by hand.
- `packages/rolldown/tests/dev/dev-lazy-compile.test.ts` — the lazy-chunk path, which reaches the same finalizer by a different route. Asserts exactly one `import * as … from "external-dep"`, no `loadExports("external-dep")`, and no `var` redeclaring that binding.

Both fail against the finalizer as it stands on `main` and pass with this change; reverting only the finalizer hunk turns both red again.

## Alternatives considered

- **Give each re-export arm its own `Module::External` branch**, mirroring the import arm. Fixes the empty re-exports, but leaves the same decision written in four places — which is what produced the bug — and does not fix the shadowing, since the two deduplication sets stay independent.
- **Merge the two deduplication sets into one.** Fixes the shadowing but not the empty re-exports. Once the two module kinds route separately the sets are disjoint anyway, so merging them would hide the invariant rather than enforce it.
- **Register externals in the dev runtime registry** so `loadExports` works for them. An external is resolved by the host, so the runtime would have to import it anyway, and this would change the registry contract for every consumer to fix one caller.

## Worth a closer look

- The `&NormalModule` signature change is the part I would most like a second opinion on. It is what turns this from a fix into an invariant, but it does narrow a public-ish helper inside the crate.
- `export * from 'ext'` is lowered as a real `import * as` plus `__reExport`, matching what scope hoisting emits for the same input. It cannot be a top-level `export * from` because a wrapper's exports object is a runtime value.

## Docs

`internal-docs/lazy-compilation/implementation.md` gains Issue 12, in the format the file already uses for Issues 1–11.

## AI disclosure

Written with AI assistance. I reproduced the bug, reviewed the fix and both tests, and verified them against a real project before submitting.

### DIFF
--- a/crates/rolldown/src/hmr/hmr_ast_finalizer.rs
+++ b/crates/rolldown/src/hmr/hmr_ast_finalizer.rs
@@ -117,21 +117,10 @@ impl<'ast> HmrAstFinalizer<'_, 'ast> {
             }
           });
         });
-        match importee {
-          Module::Normal(_) => {
-            if let Some(stmt) =
-              self.create_load_exports_call_stmt(importee, &binding_name, import_decl.span)
-            {
-              program_body.push(stmt);
-            }
-          }
-          Module::External(importee_ext) => {
-            self.create_static_import_stmt_from_external_module(
-              importee_ext,
-              &binding_name,
-              import_decl.span,
-            );
-          }
+        if let Some(stmt) =
+          self.create_importee_binding_stmt(importee, &binding_name, import_decl.span)
+        {
+          program_body.push(stmt);
         }
       }
       ast::Statement::ExportNamedDeclaration(mut decl) => {
@@ -175,7 +164,7 @@ impl<'ast> HmrAstFinalizer<'_, 'ast> {
               &self.ast_builder,
             )
           }));
-          if let Some(stmt) = self.create_load_exports_call_stmt(importee, &binding_name, decl.span)
+          if let Some(stmt) = self.create_importee_binding_stmt(importee, &binding_name, decl.span)
           {
             program_body.push(stmt);
           }
@@ -293,7 +282,7 @@ impl<'ast> HmrAstFinalizer<'_, 'ast> {
         self.dependencies.insert(importee_idx);
         let binding_name = self.ensure_static_import_info(importee_idx, rec_id).to_string();
         if let Some(stmt) =
-          self.create_load_exports_call_stmt(importee, &binding_name, export_all_decl.span)
+          self.create_importee_binding_stmt(importee, &binding_name, export_all_decl.span)
         {
           program_body.push(stmt);
         }
@@ -407,23 +396,46 @@ impl<'ast> HmrAstFinalizer<'_, 'ast> {
     }
   }
 
-  fn create_load_exports_call_stmt(
+  /// Bind the importee's namespace object to `binding_name`, whatever brought the importee in -
+  /// a plain import or any of the three re-export forms. All of them read the binding the same
+  /// way afterwards, so they must all obtain it the same way.
+  ///
+  /// Only normal modules can come from the dev runtime registry: it holds the modules this build
+  /// wrapped, and an external is by definition not one of them. `loadExports` on an external id
+  /// finds nothing, warns, and returns `{}` - so an external has to keep a real import statement
+  /// and let the host resolve it. That statement is emitted outside the wrapper, hence no
+  /// statement to return here.
+  fn create_importee_binding_stmt(
     &mut self,
     importee: &Module,
     binding_name: &str,
     span: Span,
   ) -> Option<Statement<'ast>> {
-    if self.imports.contains(&importee.idx()) {
+    match importee {
+      Module::Normal(importee) => self.create_load_exports_call_stmt(importee, binding_name, span),
+      Module::External(importee) => {
+        self.create_static_import_stmt_from_external_module(importee, binding_name, span);
+        None
+      }
+    }
+  }
+
+  /// Takes a `&NormalModule` rather than a `&Module` on purpose: asking the registry for an
+  /// external is the defect this pairing exists to prevent, so it should not be expressible.
+  fn create_load_exports_call_stmt(
+    &mut self,
+    importee: &NormalModule,
+    binding_name: &str,
+    span: Span,
+  ) -> Option<Statement<'ast>> {
+    if self.imports.contains(&importee.idx) {
       return None;
     }
-    self.imports.insert(importee.idx());
+    self.imports.insert(importee.idx);
 
     // Use stable module ID for consistent runtime lookup
-    let id = importee.stable_id();
-    let interop = match importee {
-      Module::Normal(importee) => self.module.interop(importee),
-      Module::External(_) => None,
-    };
+    let id = importee.stable_id.as_ref();
+    let interop = self.module.interop(importee);
     let call_expr = Expression::new_call_with_arg(
       Expression::new_member_access_expr("__rolldown_runtime__", "loadExports", self),
       ast::Expression::new_string_literal(SPAN, Str::from_str_in(id, self), None, self),
@@ -513,8 +525,6 @@ impl<'ast> HmrAstFinalizer<'_, 'ast> {
     binding_name_for_namespace_object_ref: &str,
     scoping: &Scoping,
   ) -> Vec<ast::Statement<'ast>> {
-    // TODO reexport external module
-
     // construct `{ prop_name: () => returned, ... }`
     let mut arg_obj_expr = ast::ObjectExpression::boxed(
       SPAN,
--- a/crates/rolldown/tests/rolldown/topics/hmr/reexport_external/_config.json
+++ b/crates/rolldown/tests/rolldown/topics/hmr/reexport_external/_config.json
@@ -1,0 +1,8 @@
+{
+  "config": {
+    "external": ["node:assert", "node:util"],
+    "experimental": {
+      "devMode": {}
+    }
+  }
+}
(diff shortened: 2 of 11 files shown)


h-a-n-a:
Nit (non-blocking): `export { default as d } from 'ext'` is not covered by either test. It should work — the binding becomes `ns.default` on a real namespace object — but it is the one named-re-export shape with its own semantics. One extra line in `reexport.js` (plus a matching assertion in `hmr.js`) would pin it.

tbvjaos510:
Thanks. added in d53385895.

h-a-n-a (approved):
Perfect! Thank you so much for your contribution!

Issue #7115, opened 2025-11-18
Transitive export * statements not re-exported in `cjs` with `preserveModules`


logaretm:
### Summary

When using preserveModules: true with CJS output format, Rolldown does not generate re-export code for transitive `export *` statements.

The intermediate module correctly re-exports from an external dependency, but the top-level module that re-exports from the intermediate module produces an empty exports object.

In our use-case, we have layers of libraries building on top of each other, so there is transitive export statements in each level.

**Reproduction:**

https://repl.rolldown.rs/#eNptUD1PxDAM/StWlgAq7Z4bmdnYKENpXAhKk8pJj0pV/jtueukBuiWx/Z4/3lvFINQqjNO41DFssRPqmlei57Rp4OXTBCCcCAO6GGBnBKQzEhNbh8vkKcIDDORHkHWzY/LEM1CoSDOmSpC3VvtvV/feDebj18obyL7cjHnwChoH4/Apw5Aue0qbPLXuOIKZ3Wzzf3Tcra0DPnuao9rOKxJltdX9HDOQSQCDp7HbeP3XhQCgDXFBmxBLJZvBIp+9ni0GBZvIjKX84hKRXGcVvMoSP1rzLt9al+75YMGOHBYWI66FXf8/Y/8MEoknnJlku4ghivQD3FKcmQ==

You can also check this stackblitz and run node with `require('./dist')` and notice that the module doesn't export `zod`'s own exports.

https://stackblitz.com/edit/abveex9m-czrtm8oq?file=dist%2Fserver.js,dist%2Findex.js,server.ts,index.ts

**File structure:**

- index.ts: export * from `'./server'`;
- server.ts: `export * from 'external-lib'`;
- external-lib: External dependency

**Config**

```js
{
  format: 'cjs',
  preserveModules: true,
  external: ['external-lib']
}
```

### Expected

index.js should contain re-export code similar to, or re-use the exports from the `./server`.

```js
const require_server = require('./server.js');

Object.keys(require_server).forEach(function (k) {
  if (k !== 'default' && !Object.prototype.hasOwnProperty.call(exports, k)) {
    Object.defineProperty(exports, k, {
      enumerable: true,
      get: function () { return require_server[k]; }
    });
  }
});
```

### Actual Behavior

index.js produces an empty exports object:

```js
const require_server = require('./server.js');

var index_exports = {};
```

Meanwhile, `server.js` correctly generates the re-export code for external-lib.

---

**Workaround**

[...]

PR #10492, opened 2026-07-27
fix(chunk-optimizer): preserve external stars after facade elimination

hyfdev:
When chunk optimization eliminates a dynamically imported entry facade, an ESM external star can disappear from the namespace returned by `import()`. This reproduces on `main` without strict execution order: local exports remain, but exports reached through `export * from "external"` become `undefined`.

```js
const namespace = await import("./target.js");
namespace.x;             // 1
namespace.externalValue; // undefined before, 42 after
```

## Problem

Rolldown delivers an entry's `export * from "external"` to `import()` callers through one of two shapes, chosen exclusively per record:

- **File-level** — when the entry owns an output file, that file carries a plain `export * from "external"` statement, free at runtime. A generate-stage walk picks this shape by marking the record `EntryLevelExternal`, which tells the namespace renderer to skip the runtime merge.
- **Object-level** — when callers receive a materialized namespace object instead, the renderer emits `import * as ns from "external"` plus `__reExport(entry_exports, ns)` to copy the external's properties in at runtime.

Facade elimination breaks the either/or assumption: the optimizer deletes the entry's file — the file-level statement has nowhere to live — and hands callers a namespace object synthesized in the implementation chunk, but the `EntryLevelExternal` mark set for the file-level shape still suppresses the object-level merge. Neither shape is emitted; the external names silently vanish.

Skipping eliminated facades during the walk would not be enough, because the mark lives on the shared import record and the `has_dynamic_exports` re-propagation behind it is transitive: a live entry reaching the same internal barrel legitimately marks the same record for its own file-level shape, and an unrelated invalidation seed can pull a simulated barrel into its importer closure — either way the recomputation treats the marked record as static and severs an upstream `__reExport` the synthesized namespace still needs.

There is a second, independent loss on the object-level path: the synthesized `import * as ns from "external"` came from an AST factory that never emitted a with-clause, so an original `export * from "external" with { ... }` lost its import attributes.

## Fix

Teach the pipeline that a namespace standing in for an eliminated facade is an object-level consumer. When the optimizer replaces a dynamic facade with a namespace in its implementation chunk, it walks the internal `export *` chain that carries dynamic exports and marks every participating module's namespace with `SimulateFacadeChunk` — "this namespace stands in for a deleted entry file".

That reason drives both halves of the decision. On the renderer side, the shared emission predicate treats `SimulateFacadeChunk` like a genuinely observed namespace, so the synthesized namespace regains its runtime `import * as` + `__reExport` merge. On the propagation side, invalidation seed selection and recursive `has_dynamic_exports` propagation both use the new `LinkingMetadata::entry_level_external_star_keeps_dynamic_exports` helper: an `EntryLevelExternal` record keeps counting as a dynamic source exactly when its owning module's namespace is genuinely observed (`Unknown`) or simulates an eliminated facade. The flag therefore always agrees with what the finalizer will emit, no matter how the importer closure grows — a live entry sharing the same barrel keeps its free file-level `export *`, while the eliminated facade's namespace keeps the runtime merge, both at once.

The two predicates stay deliberately separate: the helper decides whether a namespace chain must remain dynamic (materialized and merged), while `ns_star_external_re_export_emitted` still decides how an already-materialized namespace merges the external in ESM versus CJS/IIFE/UMD. Ordinary entries continue to use their file-level `export *` without growing namespace proxy chains.

The synthesized runtime namespace import now rebuilds the original record's import attributes, preserving `with` versus `assert`, identifier versus string keys, and values.

This is the pre-existing optimizer bug called out in #10485. It is based directly on `main`, does not depend on #10485, and does not change facade-elimination eligibility or chunk topology.

## Validation

- Direct and transitive eliminated-facade fixtures fail before the fix with `undefined !== 42` and pass after it while preserving initialization order.
- The transitive fixture includes the importer-closure counterexample from review: `live -> x` creates a normal invalidation seed, `barrel` imports `x`, and the snapshot still contains both the barrel external merge and `__reExport(target_exports, barrel_exports)`. Runtime assertions cover both live and eliminated-facade namespaces.
- A separate shared-record fixture keeps both paths at once: the live entry emits chunk-level `export * from "external"`, while the eliminated dynamic facade emits runtime `import *` plus `__reExport`; both namespaces expose `externalValue === 42`.
- An import-attributes fixture is red before the fix and snapshots `import * as ... from "external" with { type: "json" }` after it.
- The CI-reported CJS `preserveModules` fixtures for issues #7115 and #7233, including the chained case, retain their existing output without extra namespace proxies.
- Existing strict external-facade and attribute fixtures, CJS external re-export, facade elimination, issue #9374, and runtime-sweep fixtures remain unchanged.
- All 103 `rolldown` library tests, targeted `rolldown` clippy across all targets, Rust formatting, diff checks, and targeted spelling checks pass.


hyfdev:
Reviewed against the machinery this derives from (#10485's findings). The parts this PR claims are verified working: all four new fixtures pass, both original repro shapes (non-strict and strict optimizer path) now resolve `externalValue === 42`, the record-global `EntryLevelExternal` insight and the shared-chain fixture pinning both emissions at once are exactly right, and the attributes reconstruction is faithful (with/assert keyword, identifier vs string keys, correct per-record keying). Ordering holds too: elimination marks the chain before `find_entry_level_external_module` reads the reasons, and graphs without any elimination can't be affected since `SimulateFacadeChunk` has a single insertion site.

However, one same-class shape still drops the external — the invalidation exemption guards seed insertion but not the importer-closure behind it, so **"preserve external stars after facade elimination" is not yet complete**.

### Counterexample

Take this PR's own passing `eliminated_dynamic_facade_transitive_external_star` fixture and add a second user entry plus two lines in `barrel.js`:

```js
// live.js (new user entry)
export const live = true;
export * from './x.js';

// x.js (new)
export const fromX = 7;
export * from 'external';

// barrel.js — was just `export * from 'external'`
import { fromX } from './x.js';   // ← the trigger
export const viaBarrel = fromX;
export * from 'external';
```

`target.js` (`export * from './barrel.js'`, facade eliminated into the manual group chunk) now emits:

```js
// dist/target.js — barrel still merges the external into ITS namespace…
import * as import_external from "external";
__reExport(barrel_exports, import_external);
// …but the entry-level merge is gone: no `__reExport(target_exports, barrel_exports)`
```

```js
const namespace = await import('./target.js');
namespace.externalValue; // undefined !== 42
```

Removing the single `import { fromX } from './x.js'` line makes it pass, isolating the trigger.

### Mechanism

1. `live.js → x.js` makes `x.js` an invalidation seed with no exempting reason — the new `Unknown | SimulateFacadeChunk` check (`code_splitting.rs:719-729`) applies only at seed insertion.
2. The importer-closure below it (`code_splitting.rs:741-751`) then inserts every transitive importer **unconditionally**, so `barrel.js` joins `invalidated_modules` despite carrying `SimulateFacadeChunk`.
3. `propagate_has_dynamic_exports` recomputes barrel with its ELE-marked `export * from 'external'` contributing `false` — the `EntryLevelExternal` skip at `code_splitting.rs:~1196` is unconditional — so barrel's `has_dynamic_exports` flips to `false`. (Barrel's record is ELE-marked because the eliminated facade chunk keeps its `EntryPoint` kind in `chunk_table` and its walk still runs.)
4. The entry-body merge `__reExport(target_exports, barrel_exports)` is gated on exactly that flag (`module_finalizers/mod.rs:1828`), so the chain is severed at the entry, while barrel's own predicate-gated merge still renders — now unreachable from the entry namespace. The same flipped flag gates every namespace merge of that barrel, not only the eliminated facade's.

### Suggested fix

Make the flattening contribution rule reason-aware at the propagation level: inside `propagate_has_dynamic_exports`, an `EntryLevelExternal` record whose **owning module's** reason intersects `Unknown | SimulateFacadeChunk` will runtime-merge (that is the same condition `ns_star_external_re_export_emitted` uses), so it must keep counting as a dynamic source. That keeps the flag in agreement with what the finalizer emits no matter how the closure grows; the current seed-insertion exemption then becomes a redundant safety rather than the only barrier. Guarding the closure insertion instead would also work but is bypassable by future callers of the propagation.

For scope: the shape also fails on `main` (there barrel's own merge is missing as well), so this is remaining incompleteness of the bug class rather than a regression introduced here — but since closing that class is this PR's stated purpose, it seems worth folding in, with the counterexample above as a fixture (it is two files and two lines away from the existing transitive one).

### Interplay with #10485

No file overlap with the stack's source diff, and the create path never sets `SimulateFacadeChunk`, so the broadened predicate leaves its demand computation unchanged. After both land, the create path conservatively keeps facades for ESM external-star entries while this path merges at runtime — asymmetric but each correct.

hyfdev:
Re-reviewed the revision (1f0f0fb84 → 478c1a8d9). **Verified — the counterexample class is closed at the right level.** Moving the reason check into `propagate_has_dynamic_exports` itself (via the shared `entry_level_external_star_keeps_dynamic_exports()` helper) makes the flag agree with what the finalizer emits no matter how the importer-closure grows, and the folded `eliminated_dynamic_facade_transitive_external_star` fixture pins the exact trigger shape with runtime assertions on both the live entry and the eliminated facade.

Checks behind the verdict:

- My counterexample probe and its control both pass at 478c1a8d9; the updated snapshot restores `__reExport(target_exports, barrel_exports)` alongside barrel's own external merge. All four fixtures, both optimizer-path repros, and the facade-elimination / external-reexport regression set stay green.
- **The `Unknown`-side effect of the new rule is also a fix, not a risk.** The delta set vs `main` is precisely: invalidation-closure members owning an `EntryLevelExternal` record whose reason intersects `Unknown | SimulateFacadeChunk` — and for every such module, the ESM predicate already emits the runtime merge into its namespace, so `main`'s flattened `false` contradicted the predicate's own promise and severed importer-side merges (`module_finalizers/mod.rs:1828`). A no-elimination probe (live entry → seed; an `Unknown`-observed barrel importing the seed; an `Unknown`-observed re-exporter above it) shows `main`'s severed `__reExport(mid_exports, barrel_exports)` restored while the chunk-level flattening stays — `externalValue === 42` end to end. There is no shape where `main` was correct and now differs: correctness of the flattened output would require the namespace to carry no dynamic props, contradicting the emission the predicate mandates.
- **Two-pass consistency under `strictExecutionOrder` holds.** Reasons are immutable between the two `find_entry_level_external_module` passes (inserted only at link time and at the elimination site), and `invalidated_modules`/`visited` are fresh per pass, so pass 2 reproduces pass-1 values. A strict variant of the folded fixture — elimination confirmed, order wraps applied, second pass exercised — keeps both merges and passes at runtime.
- Owner-keying in the recursive propagation is right: ELE bits exist only on records resolving to external modules, the fold judges them by the owning module's reason (the same receiver `ns_star_external_re_export_emitted` uses), and importers inherit transitively through recursion keyed to each successive owner.
- The new `internal-docs/code-splitting/implementation.md` paragraph accurately records the record-global hazard and the importer-closure interaction.

shulaoda:
@codex review

hyfdev:
Feel like too niche.

PR #10501, opened 2026-07-28
fix(error): align filename-pattern diagnostics with Rollup's wording

IWANABETHATGUY:
Stacked on #10400. Diagnostic wording only — no behavior change.

#10398 made rolldown's `is_path_fragment` agree with Rollup on *which* names are rejected. Running the resulting `preserve_modules_dot_dot_name` fixture against `rollup@4.62.3` shows the two bundlers now reject the same input at the same point but still *say* different things:

```
rollup    Invalid substitution "..y" for placeholder "[name]" in
          "output.entryFileNames" pattern, can be neither absolute nor relative path.

rolldown  Invalid substitution "..y" for placeholder "[name]" in
          "entryFileNames" pattern, can be neither absolute nor relative paths.
```

Two differences: the option is named without Rollup's `output.` prefix, and the trailing "path" is pluralised.

## Change

**1. Prefix every pattern name with `output.` at its origin.** Rollup passes `output.entryFileNames` / `output.chunkFileNames` / `output.assetFileNames` / `output.sourcemapFileNames` as `patternName` (`Chunk.ts:595-596,630`, `FileEmitter.ts:81`); rolldown passed the bare camelCase name. Updated at all six origins — chunk entry/non-entry, sourcemap, the two asset-template constructors, and the null-byte diagnostic in `bundle.rs`.

This is a message-only value: `pattern_name` is never compared or matched on, only interpolated, so nothing downstream depends on it. It reaches four diagnostics — `InvalidFilenamePattern`, `InvalidFilenameSubstitution`, `HashLengthTooLong`/`TooShort`, and `NulByteInFilename` — all of which now name the option the way the docs and Rollup do.

**2. Drop the plural** in `InvalidFilenameSubstitution`: `…nor relative paths.` → `…nor relative path.`, matching `renderNamePattern.ts:31`. (`InvalidFilenamePattern` is plural in Rollup too — `patterns can be neither…` — and already matched, so it is left alone.)

**3. Quote the option name in the hash-size messages.** Rollup writes `Check the "output.entryFileNames" option.` (`hashPlaceholders.ts:22,33`); rolldown used backticks. These strings changed anyway from (1), so this aligns the quoting at the same time.

Worth noting the other `InvalidOption` variants already named `output.`-prefixed options (`output.format`, `output.file`, `output.dir`, `output.codeSplitting`, `output.preserveModules`) — the four `*FileNames` ones were the exception, not the rule.

## Tests

Four Rust snapshots and two JS fixtures updated. The `sourcemap-filenames/invalid-pattern-diagnostic` fixture guards that the diagnostic names the public option rather than leaking the internal `sourcemap_filename`; its assertion is tightened to the Rollup-shaped `for "output.sourcemapFileNames"`, which keeps that guarantee.

`cargo test -p rolldown --test integration` and the Node fixture suite pass.

## What this does not change

The error *code* still differs — rolldown reports `INVALID_OPTION`, Rollup reports `VALIDATION_ERROR` — as does when the check is reachable under `preserveModules`, since rolldown inlines const modules that Rollup preserves. Both are noted in #10398 and are out of scope here.

### DIFF
--- a/crates/rolldown/src/bundle/bundle.rs
+++ b/crates/rolldown/src/bundle/bundle.rs
@@ -195,12 +195,12 @@ impl<Fs: FileSystem + Clone + 'static> Bundle<Fs> {
         let pattern_name = match chunk {
           rolldown_common::Output::Chunk(c) => {
             if c.is_entry {
-              "entryFileNames"
+              "output.entryFileNames"
             } else {
-              "chunkFileNames"
+              "output.chunkFileNames"
             }
           }
-          rolldown_common::Output::Asset(_) => "assetFileNames",
+          rolldown_common::Output::Asset(_) => "output.assetFileNames",
         };
         return Err(
           BuildDiagnostic::invalid_option(rolldown_error::InvalidOptionType::NulByteInFilename {
--- a/crates/rolldown/tests/rolldown/errors/invalid_option/invalid_filename_pattern/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/errors/invalid_option/invalid_filename_pattern/artifacts.snap
@@ -6,6 +6,6 @@ source: crates/rolldown_testing/src/integration_test.rs
 ## INVALID_OPTION
 
 ```text
-[INVALID_OPTION] Invalid pattern "/absolute/[name].js" for "entryFileNames", patterns can be neither absolute nor relative paths. If you want your files to be stored in a subdirectory, write its name without a leading slash like this: subdirectory/pattern.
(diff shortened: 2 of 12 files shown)


hyfdev (approved):
The change is message-only: every filename-pattern origin now names the public `output.*FileNames` option the way Rollup does, the substitution message uses singular "path", and the hash-size messages quote the option with double quotes. Snapshots and the sourcemap diagnostic fixture track those strings; rejection behavior is untouched.

PR #10502, opened 2026-07-28
docs(code-splitting): record the thenable-namespace rule

hyfdev:
Follow-ups from #10480's review (https://github.com/rolldown/rolldown/pull/10480#issuecomment-5100864763) that missed the merge.

- The two `then`-hazard fixtures only loaded the dynamic importer, so the chunk that imports the renamed binding was never linked at runtime — a producer/consumer disagreement on the new name would have stayed green. Both `_test.mjs` now load `b.js` and assert the re-exported binding arrived.
- `internal-docs/code-splitting/design.md` gains a "Thenable chunk namespaces" section: the assimilation mechanism, why the defense splits by name ownership (bundler-owned names are never `then`; user-observable ones keep the facade instead), and which case deliberately keeps `then` (the target's own export — native `import()` behaves the same). The reservation site in `deconflict_exported_names` points at it.
- The generator comment claimed `then` needs ~14M names; that is where five-character names begin. It first appears at value 443,179.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
+++ b/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs
@@ -1349,9 +1349,10 @@ impl GenerateStage<'_> {
           loop {
             named_index += 1;
             export_name = generate_minified_names(named_index).into();
-            // Unreachable in practice — base54 needs about 14M names before it produces a
-            // four-character `then` — but the generator is the only other source of internal
-            // export names, so make it impossible rather than improbable.
+            // Unreachable in practice — the generator first produces the four-character `then`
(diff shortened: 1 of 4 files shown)


hyfdev:
### Merge activity

* **Jul 28, 7:02 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 28, 7:34 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 28, 7:41 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

shulaoda (approved):

PR #10505, opened 2026-07-28
test(code-splitting): cover a single-chunk dynamic import of a wrapped module

hyfdev:
Last P1 item of #10294.

### Why

With `codeSplitting: false` there is no second chunk to load, so `determine_module_exports_kind` marks the dynamic target `WrapKind::Esm`, `wrap_modules` follows into its whole subgraph, and the finalizer inlines every `import()` call site to `Promise.resolve().then(() => (init_target(), target_exports))`. Generate-stage order lowering is almost entirely short-circuited on this path — six `code_splitting.is_disabled()` early returns in `order_wrapping.rs` — so `strictExecutionOrder` was relying on it without pinning it.

`strict_execution_order/` had no coverage of that shape, and no `iife` or `umd` coverage of any shape: of its 80+ fixtures, three use `codeSplitting: false` (none with a dynamic import) and none use `iife`/`umd`.

### What

One fixture, four cells: `iife`/`umd` × wrap-all/on-demand.

The snapshot difference between the wrap modes is real rather than cosmetic — wrap-all also wraps the entry, so both formats render `exports.ready` from a wrapped entry (`init_main(); exports.ready = ready;`, the `render_wrapped_entry_chunk` path) while on-demand renders it from an inline body. Both keep `init_lazy` behind the deferred `Promise.resolve().then(...)`.

`_test.mjs` runs in all four cells and asserts the full execution sequence, so the pin is behavioural and not just textual:

- the dependency shared with the entry keeps its eager static position,
- nothing the dynamic target pulls in has run while the entry body is still executing,
- the target initializes exactly once,
- both `import()` call sites settle to the same namespace object.

### Verification

- New fixture green; full `strict_execution_order` group green (101 tests).
- Negative check: flipping the `lazy-pending:true` expectation makes the fixture fail, confirming `_test.mjs` really executes in every cell rather than being skipped.

### DIFF
--- a/crates/rolldown/tests/rolldown/function/experimental/strict_execution_order/single_chunk_dynamic_import_wrapped_module/_config.json
+++ b/crates/rolldown/tests/rolldown/function/experimental/strict_execution_order/single_chunk_dynamic_import_wrapped_module/_config.json
@@ -1,0 +1,26 @@
+{
+  "_comment": "Single-chunk strict coverage for iife/umd: with `codeSplitting: false` there is no other chunk to load, so `determine_module_exports_kind` wraps the dynamic target and `wrap_modules` follows into its whole subgraph, and the finalizer inlines both `import()` call sites to `Promise.resolve().then(() => (init_lazy(), lazy_exports))`. Strict order must not disturb that: the target stays lazy (nothing it pulls in runs before the promise settles), initializes exactly once, and hands both call sites the same namespace object, while the dependency it shares with the entry keeps its eager static position. Wrap-all additionally wraps the entry itself, so the two formats also render `exports.ready` from a wrapped entry instead of from an inline body.",
+  "configName": "iife-wrap-all",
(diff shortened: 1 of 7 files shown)


hyfdev:
### Merge activity

* **Jul 28, 1:51 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 28, 1:57 PM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 28, 2:05 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

shulaoda (approved):

PR #10508, opened 2026-07-29
fix(plugin): measure plugin hooks in JavaScript, report them from Rust

IWANABETHATGUY:
Third of four, on #10521. #10509 sits on top and adds the remaining option-level callbacks.

Two prefactors were split out below to keep this one to its actual subject: #10520 gives the
build's clocks their own type, and #10521 moves the report to close (fixing a bug where
`closeBundle`'s cost was measured and discarded) and makes it account for multi-output
builds. Both stand on their own; this PR is then only *where the number comes from*.

## Problem

`[PLUGIN_TIMINGS]` is timed from Rust, which can only bracket *dispatch* and *completion*. For a hook the core dispatches concurrently — `resolveId`, `load`, `transform`, `moduleParsed`, `renderChunk` — those two points are separated mostly by the queue the call waited in on JavaScript's single thread, and the queue is deepest behind whichever callback is doing the most work.

The failure is a sign inversion, not noise:

- a hook that blocks the thread creates the backlog that dilutes its own credit;
- a cheap hook called thousands of times sits in that backlog collecting credit for work it is not doing.

At hand-checkable scale: one callback holds the thread for 100ms while a hundred trivial ones queue behind it and drain in 1ms total. Truth — the expensive one is 99% of the work. Measured — every queued call's stopwatch ran for the whole 100ms it waited, so the trivial hook looks a hundred times more expensive.

On a real build, gated A/B so the cost reads straight off the wall clock, **a hook worth ~15s ranked below one worth ~0s**.

## Why no arithmetic fixes it

The callback's execution *start* lives in the JS event loop and is never sampled. Two schemes were built and both failed in the same direction: rescaling every row by a common factor cannot reorder them, and dividing by the in-flight count charges the *running* callback a `1/n` sliver of what it is actually consuming (Node is run-to-completion — one callback runs at 100% while the others run at 0%).

## Approach

The missing quantity is known inside the JS wrapper, so the clock moves there. A wrapper around each handler records entry-to-exit, which excludes the dispatch queue by construction.

Collection is split by necessity — only Rust can see the link stage, only JavaScript can see when a callback began running — so the two halves meet at the sink rather than one side accumulating the other's numbers.

**Rust pulls; JavaScript does not push.** `pluginTimings` is a JS callback on the options, modelled on the existing `deferSyncScanData`, invoked while the build closes:

```rust
// crates/rolldown_binding/src/binding_bundler.rs
let res = cleanup_fut.await;                          // runs closeBundle
handle_result(res)?;
handle_result(report_plugin_timings(&handles).await)?; // asks JavaScript
```

Pulling is what lets the measurement include `closeBundle`, which runs *inside* `close()` and would already be over by the time a push could reach anyone.

Reporting therefore stays where the rest of the diagnostics live. `EventKind::PluginTimings` is constructed again rather than merely generating an option name, so `checks.pluginTimings` is honoured by `filter_out_disabled_diagnostics` like every other check, and the message renders through `Diagnostic::render_batch` like every other warning.

## The trust rule

Spans may be summed only if they never overlap. A synchronous callback cannot — it holds the thread until it returns. An `async` one can suspend at an `await` and let another call of the same hook begin, and then both spans cover the same wall clock. Overlap also changes what the span *measures*: a hook awaiting `this.resolve` is measuring the bundler. So overlap is counted per hook rather than assumed:

- **never overlapped** — the sum is execution time, and the row gets a number;
- **overlapped** — the row is named as needing `node --cpu-prof`, with no number.

That narrows "cannot measure" from a property of the *call site* to a property of the *callback*, which is where it lives. A concurrently dispatched hook whose body never overlaps is now measured exactly — the case a call-site rule has to reject.

## Also here, because it cannot be anywhere else

`output.codeSplitting.groups[].name` is timed today on `main`, from Rust. This PR deletes
that Rust-side timing, so the JS wrapping for it has to land in the same commit — wrapping
it earlier would double count against the Rust clock, and wrapping it later would leave a
window on `main` with no coverage of a callback that runs once per module. Its `groups[].test`
sibling is not timed on `main`, so that one waits for #10509.

## Unchanged

The gate: build over 3s, non-link time over 100× link time, summed across every output. It
lives on `BuildTimings` as of #10520 and gained the multi-output summing in #10521; this PR
does not touch it.

## Verification

Gated A/B against a plugin burning a known amount of CPU — same build with and without the burn, so the cost reads off the wall clock:

```
gated off   : wall   27ms, warned=false
gated on    : wall 5608ms, reported 5600ms
wall delta  : 5581ms   <- ground truth
error       :   19ms   (0.3%)
```

Plus: unit tests for the collector (overlap, busy time, settle-on-throw/reject, the payload shape), Rust tests for the gate and the rendered message, and end-to-end runs asserting a blocking sync `transform` ranks first while an overlapping `async moduleParsed` lands in the un-measurable list.

## Rows are keyed on identity, not name

Two configured instances of one plugin share a name — `normalizePlugins` allows it — but
are different culprits, and merging them would also let one instance's overlap mark the
other's clean measurement unrankable. `TimingOwner` carries `key` (the user's plugin object,
stable across a build's outputs) separately from the displayed `name`.

## For reviewers

- `report_plugin_timings` in `binding_bundler.rs` — the ordering against `cleanup_fut` is load-bearing; the close-time emission and clock summing themselves arrive in #10521.
- The overlap rule in `plugin-timings.ts`: `overlapMs` accumulates `(inFlight - 1) × dt` at each transition, which is exactly the double-counted amount, and a row stays reportable while that is under 1% of its span. A boolean on peak in-flight was too brittle — one incidental overlap among thousands of calls discarded a good measurement and made the report depend on scheduling.
- `busyMs` (union of spans) is maintained as the recorder's in-flight count crosses zero.
- Native `builtin:` plugins have no JS callback and so are not covered. They never got a row before either, being filtered by name, but their cost *is* honestly measurable from Rust — no shared single thread to queue on — and that is the natural next step.


crates/rolldown_binding/src/binding_bundler.rs
```diff
@@ -15,20 +15,25 @@ use crate::{
...
 use std::sync::Arc;
 
 #[napi]
 pub struct BindingBundler {
   inner: ClassicBundler,
-  last_bundle_handle: Option<BundleHandle>,
+  /// Every build this bundler has run, in order. A `RolldownBuild` may `generate` more
+  /// than once, and the plugin-timing measurement accumulates across all of them, so the
+  /// clocks it is judged against are summed over all of them too.
+  bundle_handles: Vec<BundleHandle>,
```
shulaoda, line 28: It seems that this PR changes several existing behaviors and implementation details. Could you split it into a few smaller PRs to make it easier to review?

shulaoda:
@cidex review

shulaoda (approved):

PR #10509, opened 2026-07-29
feat(plugin): time the user callbacks configured on the options

IWANABETHATGUY:
Stacked on #10508, which moves the measurement into the JavaScript callback. This adds the callbacks that are not plugin hooks.

## Problem

Plugin hooks are not the only JavaScript the bundler calls. Every option-level callback crosses the same boundary, and any of them can dominate a build while belonging to no plugin's rows — a heavy `codeSplitting.name` in a real project was what prompted this.

Sixteen of them reach napi unwrapped, so none could appear in `[PLUGIN_TIMINGS]` at all. (`groups[].name` is the seventeenth and lands in #10508, because `main` already times it there and that PR removes the Rust-side timing it relies on.)

## What is now measured

Audited from the Rust side — every field on `BindingInputOptions` / `BindingOutputOptions` typed as a JS callback — rather than by reading the TypeScript, so the list is complete rather than the ones I happened to think of:

| | |
|---|---|
| **output** | `codeSplitting` `groups[].test`, `entryFileNames`, `chunkFileNames`, `assetFileNames`, `sourcemapFileNames`, `sanitizeFileName`, `banner`/`footer`/`intro`/`outro` and their `post` variants, `globals`, `paths`, `sourcemapPathTransform`, `sourcemapIgnoreList` |
| **input** | `external`, `treeshake.moduleSideEffects` |

`external` runs once per import specifier and `treeshake.moduleSideEffects` once per module — the same fan-out as the classifier, so both are just as capable of dominating a build.

They report under `output options` / `input options`, and rows carry a `kind` discriminator supplied by the wrapping call site — not derived from the owner's name, since a plugin may legally be called `output options`.

## Two things worth review attention

**`assetFileNames` and `sanitizeFileName` have two call paths.** Besides the binding, `PluginContextImpl` calls the user's option *directly* for `this.emitFile`. Wrapping at each consumer would count one call in two places, so those two are measured once in `createBundlerOptions` and passed on already wrapped.

**Rows can now sum past 100%.** `this.emitFile` inside `buildStart` invokes `assetFileNames`, so the nested time belongs to both rows. The headline figure is the union of every span, so it cannot exceed the build. Documented in the check's docs.

## Verification

Each callback made individually expensive enough to clear the one-second row floor, so the wiring is checked by execution rather than by inspection:

```
  - virtual-fs buildStart                        - output options intro
  - output options paths                         - output options chunkFileNames
  - output options sanitizeFileName              - output options entryFileNames
  - input options treeshake.moduleSideEffects    - output options outro
  - input options external                       - output options footer
  - output options codeSplitting groups[].name   - output options banner
  - output options codeSplitting groups[].test   - output options assetFileNames
  - output options globals                       - output options sourcemapFileNames
```

The motivating case in isolation: 61 calls, 3.1s, `output options codeSplitting groups[].name` at 60% of a 5.1s build.

**Two exceptions, stated rather than implied:** `sourcemapPathTransform` and `sourcemapIgnoreList` are wrapped but I could not get the core to invoke them under `generate` with `sourcemap: true` — zero calls. They are used in `process_code_and_sourcemap.rs`, so they are presumably reached on another path; wrapped either way, but unverified by execution.

### DIFF
--- a/crates/rolldown_binding/src/types/binding_plugin_timings.rs
+++ b/crates/rolldown_binding/src/types/binding_plugin_timings.rs
@@ -1,12 +1,14 @@
-use rolldown_error::{PluginTiming, PluginTimingsMeasurement};
+use rolldown_error::{PluginTiming, PluginTimingKind, PluginTimingsMeasurement};
 
 /// What one callback cost, measured inside it on the JavaScript side — the one place the
 /// measurement exists, since this side can only bracket dispatch and completion.
 #[napi_derive::napi(object, object_to_js = false)]
 #[derive(Debug)]
 pub struct BindingPluginTiming {
-  /// The plugin the callback belongs to.
+  /// The plugin the callback belongs to, or the options it was configured on.
   pub owner: String,
+  #[napi(ts_type = "'plugin' | 'outputOption' | 'inputOption'")]
+  pub kind: String,
   pub hook: String,
   pub calls: u32,
   pub ms: f64,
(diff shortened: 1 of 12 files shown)


shulaoda:
@cidex review

shulaoda (approved):

PR #10512, opened 2026-07-29
perf(code-splitting): dedupe the strict-order cycle analysis and cache consumer-local sort keys

hyfdev:
Strict-execution-order builds recompute unchanged plan inputs: the module-graph cycle scan runs three to five times per build, and ordering one consumer's initialization targets re-derives each target's route position at every sort comparison — quadratic on a wide barrel consumed as a namespace. This PR computes the cycle scan once per build and each ordering key once per target. Emitted output is byte-identical: the full fixture corpus runs with zero snapshot changes.

## Motivation

Every strict build pays the same analysis several times over. The synchronous-cycle facts are derived independently by the pre-chunk placement probe, the on-demand projection, and lowering — three to five full-graph passes per build depending on the mode. Separately, whenever a consumer's initialization targets are sorted into source order, the sort key — the target's position along its re-export route — is recomputed for every comparison rather than once per target, so a barrel with many re-exports consumed as a namespace multiplies a graph walk by the comparison count.

Non-strict builds are untouched either way: with the flag off the order state is empty and neither the scan nor the sorts ever run.

## Changes

The synchronous-cycle scan (`synchronous_cycle_modules`, a Tarjan pass over the whole module graph) is computed lazily once per build on `GenerateStage` and shared by the probe, the projection, and lowering. Its only inputs are import-record kind/target and module normality, which no generate-stage pass mutates. The two consumer-local sorts switch to `sort_by_cached_key` — stable, so equal keys keep their order — and the carrier arm reads its position directly from the record index instead of scanning `import_records`.

Two further dedups are deliberately **not** taken:

- The wrap-all analysis still runs per call. The facade-chunk optimizer (default-on `experimental.mergeCommonChunks`) replays inclusion and rewrites `metas.is_included` between that analysis's two call sites, so caching would silently freeze the pre-optimizer view. That wants an explicit probe/final coherence decision first; the function now carries a comment naming the hazard.
- The route-order DFS keeps its backtracking shape. Memoizing failed subtrees is unsound on cyclic waypoint graphs without taint tracking, and the remaining worst case requires a legal star-re-export diamond among transparent waypoints.

## Validation

- Output is byte-identical: the strict-execution-order fixtures (101/101) and the full `--ignored` integration corpus run with a clean working tree afterward — zero snapshot changes.
- The full corpus failing-test name set is identical to `main` on this machine (62 pre-existing local failures: uninitialized `rollup`/`test262` submodules plus dot-disabled fixtures); no new failures.
- `cargo test -p rolldown --lib`: 104/104
- `cargo clippy -p rolldown --all-targets -- --deny warnings`; `cargo fmt --all -- --check`

Part of #10294.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/mod.rs
+++ b/crates/rolldown/src/stages/generate_stage/mod.rs
@@ -1,4 +1,5 @@
 use std::path::PathBuf;
+use std::sync::OnceLock;
 
 use arcstr::ArcStr;
 use futures::future::try_join_all;
@@ -128,6 +129,10 @@ pub struct GenerateStage<'a> {
   /// `paths` option, it is resolved asynchronously here before entering sync rendering code,
   /// avoiding the need for `invoke_sync` which can cause deadlocks.
   resolved_paths: Option<PathsOutputOption>,
+  /// Synchronous-cycle facts read only from `import_records` kind/target and module normality,
+  /// which no generate-stage pass mutates — computed lazily once per build and shared by
+  /// pre-chunk probing, order analysis, and lowering.
+  cached_synchronous_cycle_modules: OnceLock<FxHashSet<ModuleIdx>>,
 }
 
 impl<'a> GenerateStage<'a> {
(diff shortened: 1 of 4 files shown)


crates/rolldown/src/stages/generate_stage/mod.rs
```diff
@@ -128,6 +129,10 @@ pub struct GenerateStage<'a> {
   /// `paths` option, it is resolved asynchronously here before entering sync rendering code,
   /// avoiding the need for `invoke_sync` which can cause deadlocks.
   resolved_paths: Option<PathsOutputOption>,
+  /// Synchronous-cycle facts read only from `import_records` kind/target and module normality,
+  /// which no generate-stage pass mutates — computed lazily once per build and shared by
+  /// pre-chunk probing, order analysis, and lowering.
+  cached_synchronous_cycle_modules: OnceLock<FxHashSet<ModuleIdx>>,
```
hyfdev, line 135: This doesn't smell good.

hyfdev:
OnceLock

PR #10513, opened 2026-07-29
refactor(code-splitting): declare CJS carrier namespaces initializer-free

hyfdev:
A CJS re-export carrier's namespace was the one closure-assigned hoisted binding in strict output declared with an initializer. It is now declared initializer-free, like every other binding that is assigned inside an init closure:

```diff
- var import_clone_deep = void 0;
+ var import_clone_deep;
```

Semantics are unchanged — `var` declarations have no temporal dead zone, so both forms read `undefined` before the carrier runs — and the change exists only in strict-execution-order output.

## Motivation

The two forms are equivalent today because the declaration provably executes before any carrier call (carriers are never minted for cyclic importees; those shapes fall back to a monolithic barrel, pinned by the `*_cycle_fallback` fixtures). The initializer-free form is the safer shape by construction: `= void 0` is a real assignment, so any future statement reordering that ran the declaration after the carrier initialized would silently wipe the namespace back to `undefined` behind the memoized guard — a bare declaration cannot. It also matches how every other closure-assigned hoisted binding in scope-hoisted output is declared; the initializer only existed because the statement factory required one.

## Changes

The statement factory gains an initializer-free variant (`new_var_decl_no_init`) and the carrier declaration uses it. From the same review pass: the finalizer branch that handled a carrier's wrapper resolving through another chunk's CJS entry namespace — unreachable, because a carrier is emitted by its own importee's finalizer and is therefore always a same-chunk reference — is replaced by a `debug_assert!` stating that argument.

## Validation

- Snapshot-only: 28 changed lines across 9 strict fixtures, every one a `var x = void 0;` → `var x;` declaration.
- strict-execution-order fixtures: 101/101; `cargo test -p rolldown --lib`: 104/104
- `cargo clippy -p rolldown -p rolldown_ecmascript_utils --all-targets -- --deny warnings`; `cargo fmt --all -- --check`
- Non-strict output is untouched: carriers exist only in strict builds.

Part of #10294.

### DIFF
--- a/crates/rolldown/src/module_finalizers/impl_visit_mut.rs
+++ b/crates/rolldown/src/module_finalizers/impl_visit_mut.rs
@@ -42,28 +42,23 @@ impl<'ast> ScopeHoistingFinalizer<'_, 'ast> {
       debug_assert_eq!(importee, self.ctx.idx);
 
       let namespace_name = self.canonical_name_for(namespace_ref);
-      program.body.push(Statement::new_var_decl(
-        namespace_name,
-        ast::Expression::new_void_0(SPAN, &ast_builder),
-        &ast_builder,
-      ));
+      program.body.push(Statement::new_var_decl_no_init(namespace_name, &ast_builder));
 
       let importee_wrapper_ref =
         self.ctx.linking_info.wrapper_ref.expect("carrier importee should have a CJS wrapper");
       let (importee_wrapper_expr, hint) =
         self.finalized_expr_for_symbol_ref(importee_wrapper_ref, false, false);
-      let require_call = if hint.contains(FinalizedExprProcessHint::FromCjsWrapKindEntry) {
-        importee_wrapper_expr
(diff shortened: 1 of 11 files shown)


hyfdev:
### Merge activity

* **Jul 29, 7:25 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 29, 7:32 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 29, 7:39 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

shulaoda (approved):

PR #10514, opened 2026-07-29
test(code-splitting): pin CJS barrel fallback triggers and flag-off equivalence

hyfdev:
The fallback decisions from #10488 — which barrel shapes must give up per-consumer routing — were mostly untested: of the rejecting triggers, only cycles and direct `require()` targets had fixtures, so a regressed decision would surface only as snapshot drift in unrelated fixtures. This PR pins every constructible trigger, plus the claim that a `moduleSideEffects: false` barrel executes the same set with the flag on and off, as runnable assertions. For example:

```js
// barrel: one leaf carries top-level await → the whole barrel refuses routing
export { cn } from './cn.js';                            // contains `await Promise.resolve()`
export { default as cloneDeep } from './clone-deep.cjs';

// pinned: the entry initializes clone-deep eagerly —
// asserted as runtime events ['cn', 'clone-deep', 'main:a b'], not only as a snapshot
```

Stacked on #10513 (the carrier-declaration change) so the snapshots carry the final declaration shape.

## Motivation

The routing side of the barrel fix is heavily fixtured, but the refusing side is load-bearing semantics with one-sided coverage. A wrong fallback decision — routing a shape that must stay monolithic, or vice versa — would not fail any test today; it would only drift snapshots elsewhere, which reads as noise. The strict/non-strict same-execution-set claim for side-effect boundaries was likewise argued in review, not asserted by a test.

## New fixtures

Each fallback fixture runs a wrap-all and an on-demand cell and asserts both placement (the entry's static import closure) and runtime event order on Node:

- `cjs_reexport_barrel_tla_fallback` — a top-level-await taint keeps the barrel monolithic: the unrelated CJS leaf joins the eager entry.
- `cjs_reexport_barrel_external_star_fallback` — `export *` from an external module (dynamic exports) keeps the barrel monolithic; the external import and the CJS leaf both become eager.
- `cjs_reexport_barrel_cjs_export_star_fallback` — `export *` from a CJS module keeps the barrel monolithic; star bindings resolve through the namespace merge.
- `cjs_reexport_barrel_shimmed_export_fallback` — a shimmed missing export keeps the barrel monolithic; the shimmed binding reads `undefined`.
- `cjs_reexport_barrel_code_splitting_disabled` — single-chunk output (`codeSplitting: false`, the current spelling of `inlineDynamicImports`). One observation here: the plans diverge. Wrap-all still routes per consumer, so the CJS leaf stays lazy behind the inlined `import()` trigger; on-demand sees no cross-chunk hazard in one chunk, leaves the barrel out of its wrap plan, and the monolithic interop body makes the leaf eager. Source order holds in both, and the eager-set gap is in wrap-all's favor, so the "wrap-all never executes more user code than on-demand" invariant stands — flagged in case on-demand should match wrap-all's laziness here, which this fixture would now catch.
- `cjs_reexport_barrel_effectful_carrier` / `cjs_reexport_barrel_side_effect_boundary` gain a `flag-off` cell each: strict and non-strict execute the same eager and lazy sets, with one legitimate order difference the cells document — non-strict hoists generated CJS interop via lazy-init transfer, strict keeps each require at its re-export source position.

A concatenated-wrapper × barrel fixture is not constructible today: `ConcatenateWrappedModuleKind::Inner`/`Root` never arises on this branch (see the `concatenate_basic` fixture's `_comment`), so that eligibility guard remains future-proofing.

## Validation

All fixtures pin current behavior — green here by construction; their value is turning a future fallback regression from silent snapshot drift into a named failure.

- strict-execution-order fixtures: 106/106 (101 on the base branch, +5 new)
- full `--ignored` integration corpus: no new failures relative to the base branch on this machine; this PR adds 5 passes
- Test-only: no source files change in this PR.

Part of #10294.

### DIFF
--- a/crates/rolldown/tests/rolldown/function/experimental/strict_execution_order/cjs_reexport_barrel_cjs_export_star_fallback/_config.json
+++ b/crates/rolldown/tests/rolldown/function/experimental/strict_execution_order/cjs_reexport_barrel_cjs_export_star_fallback/_config.json
@@ -14,10 +14,6 @@
     {
       "_configName": "on-demand",
       "onDemandWrapping": true
-    },
-    {
-      "_configName": "flag-off",
-      "strictExecutionOrder": false
     }
   ]
 }
--- a/crates/rolldown/tests/rolldown/function/experimental/strict_execution_order/cjs_reexport_barrel_cjs_export_star_fallback/_test.mjs
+++ b/crates/rolldown/tests/rolldown/function/experimental/strict_execution_order/cjs_reexport_barrel_cjs_export_star_fallback/_test.mjs
@@ -30,28 +30,23 @@ function staticClosure(root) {
   return [...seen].map((name) => files.get(name)).join('\n');
 }
 
-const mainClosure = staticClosure('main.js');
-const routeFile = /import\(["']\.\/([^"']*route[^"']*)["']\)/.exec(mainClosure)?.[1];
-assert.ok(routeFile, 'main must retain the dynamic route boundary');
-assert.doesNotMatch(mainClosure, /effectful-clone/);
-assert.match(staticClosure(routeFile), /effectful-clone/);
-assert.match(staticClosure(routeFile), /nested-effectful-clone/);
+// A CJS `export *` gives the barrel a dynamic namespace, which rejects consumer-local routing;
+// the monolithic barrel keeps every leaf eager for the entry.
+assert.match(staticClosure('main.js'), /clone-deep\.cjs/);
 
 globalThis.__events = [];
 
 const entry = await import('./dist/main.js');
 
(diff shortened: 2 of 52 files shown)


shulaoda (approved):

hyfdev:
### Merge activity

* **Jul 29, 7:41 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 29, 7:42 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 29, 7:50 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10516, opened 2026-07-29
fix(external): keep `__toESM` when the importing module is tree-shaken

### LINKED ISSUE #10069: [Bug]: default-import a CJS builtin module, re-export it, import it again, then use it -> `Cannot read properties of undefined`

See https://gist.github.com/jameshfisher/18ca51d835d40b5f822e84de693385c9 for minimal repro.

```
// node-https.ts
import node_https from "node:https";
export { node_https };
```

```
// entry.ts
import { node_https } from "./node-https";

export function makeAgent(): import("node:https").Agent {
  return new node_https.Agent({ keepAlive: true });
}
```

Build and run as `cjs`.

### What is expected?

When a builtin module has default export, I can import it, then re-export it, then use that somewhere else.

Running rolldown-generated code emits `Cannot read properties of undefined`. In above example, `node_https` is `undefined` in `entry.ts`.

### System Info

```Shell
System:
    OS: macOS 26.3
    CPU: (12) arm64 Apple M4 Pro
    Memory: 112.80 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.14.0 - /Users/jim/.local/state/fnm_multishells/32887_1782853313885/bin/node
    npm: 11.9.0 - /Users/jim/.local/state/fnm_multishells/32887_1782853313885/bin/npm
    pnpm: 10.33.0 - /Users/jim/.local/state/fnm_multishells/1081_1782816033351/bin/pnpm
    bun: 1.3.9 - /Users/jim/.nix-profile/bin/bun
  Browsers:
    Chrome: 149.0.7827.199
    Firefox: 146.0.1
    Safari: 26.3
```

### Any additional comments?

Note this is NOT https://github.com/rolldown/rolldown/issues/9299. It looks _very_ similar, but that repro is fixed in latest, whereas mine is not

(A bisect claims rolldown 0.14.0 works, 0.15.0 breaks. Still broken in latest 1.1.3.)

IWANABETHATGUY:
Fixes #10069.

Default-importing a CJS external, re-exporting it, then reading it back produced `Cannot read properties of undefined` in CJS output.

## The bug

`import d from 'external'; export { d }` links the shim's symbol to a `NamespaceAlias`, so tree-shaking follows that alias straight to the external namespace and never enqueues the shim's declaring statement. With a side-effect-free external the shim is dropped entirely, and two things break:

- `RuntimeHelper::ToEsm` is registered against the now-dead import statement and discarded with it.
- `render_cjs_chunk_imports` derived `needs_interop` from `chunk.direct_imports_from_external_modules`, which is built only from modules present in the chunk, so the external landed in the indirect bucket unwrapped.

The finalizer still rendered `<external_ns>.default`, so the output read `.default` off the raw CommonJS exports and got `undefined`.

Introduced by #6751 ("skip `__toESM` when not needed to reduce output size"), first shipped broken in 1.0.0-beta.46 — verified by bisecting published npm builds.

## The fix

The question "is this external read as an ES module?" can only be answered *after* linking resolves the reference, so the inclusion pass records it — the one pass that walks the reference at that point. The record is keyed by **observer**: the module holding the surviving reference, not the importer that wrote the `import`, since the importer is routinely the shim that gets dropped while an observer has included code by construction and therefore lands in a chunk.

Rendering then resolves observers to chunks, which decides both questions per chunk:

- **Which chunks wrap** — only those an observer landed in. A chunk holding just a named import reads the CommonJS object directly; wrapping it would change namespace identity and descriptor shape, and fire extra `ownKeys` traps on a Proxy-based export.
- **Which mode** — modes are aggregated per observer. Unioning bundle-wide would make separate `.mjs` and `.js` chunks both look mixed-mode and emit `__toESM(mod, 1)` *and* `__toESM(mod)` for one live binding.

An observer that ends up without a live chunk cannot be attributed, so every chunk emitting that external honours it. This fallback is load-bearing: narrowing it away under-approximates, which is what produced the bug.

Deconflicting and the cjs/iife/umd renderers share one mode derivation (`chunk_external_interop_modes`). Deriving it in two places is how they drifted apart — deconflicting filtered out named-only specifiers and the renderers did not, so a chunk could be planned single-mode then rendered `__toESM(mod, 1)` because an unrelated *named-only* importer happened to be ESM.

## Notes for review

- `issues/4585` regains its `__toESM` call, restoring the output it had before #6751. That fixture sets `expectExecuted: false`, which is how it kept snapshotting the broken form for months. The new fixtures assert executably against real `node:https`.
- Node-mode provenance is recovered by walking the symbol link chain back to the importing module, since linking collapses the re-export chain onto the external's namespace symbol.
- **Known limitation, not fixed here:** one consumer reaching an external through both a `.mjs` and a `.js` shim collapses to a single canonical symbol, so one reference reads a wrapper that does not match its shim's format. A real fix needs per-reference provenance in the finalizer. Observable only for an external declaring `__esModule` *and* owning a `default`; `dual_provenance_through_shims_of_different_formats` characterizes it.

Design rationale is in `internal-docs/runtime-helpers/implementation.md`.


IWANABETHATGUY:
* **#10517** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10517?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10516** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10516?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10516?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10516).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

shulaoda:
@codex review

crates/rolldown/src/utils/external_import_interop.rs
```diff
@@ -24,3 +31,173 @@ pub fn import_record_needs_interop(module: &NormalModule, rec_idx: ImportRecordI
...
+) -> Option<ExternalInteropUse> {
+  let canonical_ref = link_output.symbol_db.canonical_ref_for(external_namespace_ref);
+  let observers = link_output.used_external_symbols.interop_uses_by_observer(&canonical_ref)?;
+  observers
+    .iter()
+    .filter(|(observer, _)| {
+      // An observer without a live chunk could be rendered anywhere, so every chunk has to honour
+      // it. Attributable ones only bind the chunk they landed in.
+      match assignments.live_chunk_of(**observer) {
+        Some(observer_chunk) => observer_chunk == chunk_idx,
```
hyfdev, line 77: `module_to_chunk` is not sufficient for entry modules that advanced chunking moves into another chunk. The moved entry is assigned to `common.js`, but its retained entry facade still emits its own external default export. With two entries grouped into a common chunk, `a.js: export { default as d } from "external-pkg"` produces a wrapped require in `common.js` while the `a.js` facade emits raw `require("external-pkg")` followed by `.default`. For `module.exports = { marker: "raw" }`, `require("./dist/a.js").d` is therefore `undefined`. Attribution must also cover the facade chunk that renders the entry export, with this case covered by an executable fixture.

crates/rolldown/src/stages/link_stage/tree_shaking/include_statements.rs
```diff
@@ -771,6 +783,93 @@ fn follow_cjs_namespace_alias(ctx: &mut IncludeContext, canonical_ref: &mut Symb
...
+/// the one holding the import. Returns `None` when the reference *is* the external namespace and no
+/// importing module is recoverable.
+fn external_import_writer(ctx: &IncludeContext, symbol_ref: SymbolRef) -> Option<SymbolRef> {
+  let mut cursor = symbol_ref;
+  let mut writer = None;
+  loop {
+    if ctx.modules[cursor.owner].as_normal().is_some() {
+      writer = Some(cursor);
+    }
+    match ctx.symbols.get(cursor).link {
```
hyfdev: `link` is a union-find parent, not a stable import-provenance chain. `SymbolRefDb::link` resolves both endpoints through `find_mut`, so after a `.mjs` shim is linked, linking its `.js` consumer can make the consumer point straight at the external namespace. This walk then records the `.js` consumer as the writer and loses the Node-mode `.mjs` import. With an external `{ __esModule: true, default: "inner", marker: "raw" }`, a `.mjs` namespace shim followed by a `.js` consumer returns `"inner"` from `ns.default` instead of raw `module.exports`; module allocation order makes this deterministic. Importer provenance needs a source that survives union-find compression, with this ordering covered by a fixture.

crates/rolldown/src/utils/external_import_interop.rs
```diff
@@ -24,3 +31,173 @@ pub fn import_record_needs_interop(module: &NormalModule, rec_idx: ImportRecordI
...
+    .used_external_symbols
+    .interop_uses_by_observer(&canonical_ref)
+    .is_some_and(|observers| {
+      observers.keys().any(|observer| {
+        // An unattributable observer could be rendered here, so it has to count.
+        let could_be_in_chunk = match assignments.live_chunk_of(*observer) {
+          Some(observer_chunk) => observer_chunk == chunk_idx,
+          None => true,
+        };
+        could_be_in_chunk && is_node_esm(*observer)
```
hyfdev, line 165: This also counts an ESM entry whose only remaining read is emitted by `render_chunk_exports`, but entry-export finalization does not select `node_mode_external_ns_names`. In CJS output with `entry.mjs -> shim.mjs -> external` plus a non-Node default importer, this reports a Node reader and emits both wrappers, yet the generated code discards `__toESM(external, 1)` and exports `external.default` from the non-Node wrapper. For an external with `__esModule` and its own `default`, the ESM entry export is `"inner"` instead of raw `module.exports`. Entry exports must select the Node-mode binding just like module finalization, with this mixed-mode case covered by an executable fixture.

crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
```diff
@@ -223,31 +225,46 @@ pub fn deconflict_chunk_symbols(
...
+          link_output,
+          chunk_assignments,
+          chunk_idx,
+          ext.namespace_ref,
+          named_imports,
+        )
+      {
         let canonical_ref = link_output.symbol_db.canonical_ref_for(ext.namespace_ref);
         let original_name = canonical_ref.name(&link_output.symbol_db);
         let node_name = renamer.create_conflictless_name(original_name);
```
hyfdev, line 270: `create_conflictless_name` only avoids names already reserved in the chunk-level resolver, while this synthetic name has no symbol entry for the later nested-scope pass to inspect. A Node-mode reader containing `function readNode(foo$3) { return foo }` can therefore receive the same synthetic `foo$3`; finalization rewrites the external reference to `foo$3.default`, so the parameter captures it and `readNode("shadow")` returns `undefined` instead of raw `module.exports`. The synthetic Node binding must participate in nested-scope deconfliction, or choose a name that checks nested bindings, with this case covered by a fixture.

hyfdev (dismissed):
The new observer and mode bookkeeping still loses external interop semantics in four reachable cases: entry facades can miss wrapping after advanced chunking, union-find compression can misidentify the import writer, mixed-mode entry exports discard the Node-mode binding, and a synthetic Node binding can be captured by a nested local. Each case changes the value of a default or namespace import for CommonJS exports that declare `__esModule`, so they need to be covered before this merges.

IWANABETHATGUY:
Thanks for digging into this — all four were real, and I reproduced every one with an executable fixture before touching anything.

I've fixed the union-find one here. You were right that `link` isn't a provenance chain: `find_mut` path-compresses and `link()` assigns to the target *root*, so once the shim is linked first the consumer points straight at the external and the shim drops off the walk entirely. Discovery order decides whether it bites, which is why the existing fixtures sailed past it. Provenance now comes from import records, which linking leaves alone. Worth saying plainly: that walk was my fix for the Node-mode provenance issue you raised a couple of rounds back, so this one's squarely on me.

The other three I've split out into #10519 rather than fixing here, because they all reproduce on `main` at `9fef0099f` with identical values — entry facade gives `undefined`, mixed-mode entry export gives `'inner'`, synthetic binding capture gives `undefined`. This branch neither introduces nor worsens any of them. I checked by running the same four fixtures against `main` directly, so it's not a reading of the code.

My reasoning for splitting: this PR is already five commits of interop rework, and folding in three unrelated pre-existing bugs would make it considerably harder to review — and would put them in the history as if this change caused them. #10519 has full runnable fixtures for all three plus the emitted output, so nothing is lost.

One thing your review surfaced that I think is the real story: findings 1 and 2 are the same gap. `node_mode_external_ns_names` and the "which chunk wraps this external" decision get consulted by the module finalizer but not by `render_chunk_exports` or entry facades. You proposed a single per-chunk interop plan that every render site reads a few rounds ago and I pushed back as premature — those two findings are the bill for that call. I'd like to do it properly in #10519 rather than teach each render site the rule separately.

Happy to pull any of the three back in here if you'd rather see them together — just say so.

shulaoda (approved):

PR #10517, opened 2026-07-29
perf(external): drop the runtime import from chunks that never wrap

IWANABETHATGUY:
Stacked on #10516 — review that one first.

A chunk that does not wrap an external still emitted a useless import of the runtime chunk:

```js
Object.defineProperty(exports, Symbol.toStringTag, { value: "Module" });
require("./rolldown-runtime.js");        // nothing is used from it
let node_https = require("node:https");
exports.foo = node_https.request;
```

## Why it happened

Not an empty cross-chunk import. The chunk really did import `__toESM`, so it rendered `const require_rolldown_runtime = require(...)`; DCE then dropped the unused binding and kept the side-effectful call. The import existed because `patch_module_dependencies` set `reads_external_as_esm` from the bundle-wide record, so a module reading only a *name* off the external still demanded the helper and put it in its chunk's depended symbols.

## The fix

Derive that edge from the same observer set the renderers use, so a module demands `__toESM` only when it is an observer. Both decisions now come from one source, which is what makes narrowing safe: the chunk that wraps is exactly the chunk that depends on the runtime.

Under-supplying is the dangerous direction — a chunk that wraps without the edge references a helper it never imported and finalization panics — so an observer that is not included keeps the previous bundle-wide behaviour, matching rendering's own unattributable fallback. Chunks do not exist yet at that point, so `is_included` stands in for "will have a chunk".

This only became safe once #10516 made rendering observer-attributed. Narrowing this edge before that reintroduces the panic, which is why it is a separate commit rather than part of the fix below it.

## Result

`external_interop_only_reachable_via_entry_export` shows the payoff — with only `entry-a` demanding the helper, the runtime stops being shared and folds into `entry-a`, so the separate runtime chunk disappears entirely:

```js
// entry-a.js
let node_https = require("node:https");
node_https = __toESM(node_https);

// entry-b.js
let node_https = require("node:https");
exports.foo = node_https.request;
```

`named-user` in `reexport_default_import_of_external_multi_chunk` loses its bare `require` the same way; that fixture's `_test.mjs` still asserts the values executably.

Those two snapshots are the only output changes. The fixture comment and internal doc previously stated this edge had to stay bundle-wide with the bare `require` as an accepted cost — true when written, so both are updated.


IWANABETHATGUY:
* **#10517** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10517?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10517?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10516** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10516?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10517).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

internal-docs/runtime-helpers/implementation.md
```diff
@@ -28,4 +28,6 @@ Inclusion runs before chunking, so the record cannot name a chunk directly. It s
 
 An observer that ends up without a live chunk cannot be attributed, so every chunk emitting the external honours it — contributing only that observer's mode, not the bundle union. This fallback is load-bearing: narrowing it away under-approximates, which is what produced #10069.
 
-`patch_module_dependencies` stays bundle-wide on purpose. It runs before chunking and cannot know which chunks will wrap; over-supplying a runtime edge is dead weight (a bindings-less `require` of the runtime chunk), while under-supplying panics in helper finalization.
+`patch_module_dependencies` derives the runtime edge from the same observer set, so a module demands `__toESM` only when it is an observer. The two decisions have to agree: a module that gains the edge without wrapping puts `__toESM` in its chunk's depended symbols, and the resulting cross-chunk import loses its binding to DCE while the side-effectful `require` of the runtime chunk survives — a bare `require("./rolldown-runtime.js")` for a helper nothing calls. Under-supplying is worse: a chunk that wraps without the edge references a helper it never imported, and finalization panics looking up the missing cross-chunk binding.
```
hyfdev: `a module demands __toESM only when it is an observer` understates where the helper can come from. A chunk that wraps through a dead (unused-binding) default or namespace import still gets the runtime edge: `chunk_external_interop_modes` counts such imports via `named_imports` with no liveness filter, and the edge arrives through the `reference_needed_symbols` registration on the kept import statement, not the observer set. The wrap/edge invariant holds because `specifier_needs_interop` and `import_record_needs_interop` use the same predicate over the same `named_imports`, not because both decisions come from one source.

As written, the "single source" framing invites a future refactor to drop the static-import channel (for example registering `ToEsm` only for live imports), which would reintroduce the under-supply panic for dead-import wraps. Stating the two channels would make the invariant's actual boundary visible and keep such a refactor safe.

crates/rolldown/tests/rolldown/function/external/external_interop_only_reachable_via_entry_export/_config.json
```diff
@@ -1,5 +1,5 @@
...
+  "_comment": "`entry-a` default-imports the external through a shim tree-shaking drops, so only `entry-a` observes it as an ES module. `entry-b` merely re-exports a *named* binding, and its sole live reference lives in `referenced_symbols_by_entry_point_chunk` rather than any included statement — the one path that never reaches the inclusion pass's statement walk. It must not wrap, since observer attribution is what keeps a named-only chunk off `__toESM` (see `chunk_recorded_external_interop`), and it must not demand the helper either, or its chunk gains a cross-chunk import whose binding dies in DCE and leaves a bare `require` of the runtime chunk. Both the wrap and the runtime edge are now derived from the same observer set (`patch_module_dependencies`), which is what makes narrowing them safe: `entry-b` neither wraps nor depends on the runtime, so nothing can reference a helper its chunk never imported — the panic this fixture was written to pin. With only `entry-a` demanding it, the runtime stops being shared and folds into `entry-a` outright.",
```
hyfdev: `nothing can reference a helper its chunk never imported — the panic this fixture was written to pin` pins only the named-only direction. The directions the safety argument depends on are untested: a chunk wrapping through a dead default import (backstop channel), and an observer recorded only through `referenced_symbols_by_entry_point_chunk` (a default re-export through a dropped shim). Both build correctly at this head. Adding a fixture for each would pin the invariant so a future change to either channel fails loudly instead of silently.

hyfdev (approved):
The narrowing keeps the two decisions in agreement: a chunk that wraps an external — via an observer, an unattributable observer, or a kept default/namespace import — depends on the runtime, and a named-only chunk drops both the wrap and the import together. The same invariant holds on the dead-import and entry-export paths.

Two notes on the documented safety argument below; neither blocks merge.

— Review by @hyfdev, assisted by deepseek-v4-flash

hyfdev (approved):
The observer-scoped edge matches rendering for the attributable cases this PR targets: named-only chunks drop both the wrap and the bare runtime import, while observer chunks keep `__toESM` with a live runtime edge. The unattributable `!is_included` fallback keeps the dangerous under-supply direction conservative, and the two snapshot deltas plus the multi-chunk executable assert match that contract.

The open notes on the dual wrap/edge channels and the untested safety paths still stand. One small leftover in the same function, outside the changed hunk: the entry-export block still says the recorded interop is bundle-wide; under the new predicate a named-only entry export neither wraps nor demands the helper.

PR #10520, opened 2026-07-29
refactor(plugin): give the build's own clocks their own type

IWANABETHATGUY:
First of four. Pure refactor — no behaviour change — to make the three PRs above it smaller.

## What

`HookTimingCollector` held two things that are not the same kind of thing:

- what each plugin's hooks cost, accumulated per call, keyed by `PluginIdx`;
- the build's total and link-stage wall clocks, set exactly once each.

The second pair also answers a different question. `plugins_are_slow` decides whether the build is worth reporting on *at all*, and it is asked before any per-plugin number is looked at — it does not read the plugin map.

So they split. `BuildTimings` owns the two clocks and the gate, and sits on `PluginDriver` beside the collector.

## Also

`is_plugin_bound(total, link)` exposes the same test over totals the caller supplies. Nothing needs that yet; #10521 does, because a build that produced several outputs has to sum its clocks before asking.

The gate now has unit tests — the fast-build floor, the ratio, and the cases where there is no ratio to take. It had none before.

## Not in scope

The threshold itself (3s, 100×) is unchanged, as is every caller. If this PR changes what any build reports, it is wrong.

### DIFF
--- a/crates/rolldown_plugin/src/lib.rs
+++ b/crates/rolldown_plugin/src/lib.rs
@@ -33,6 +33,7 @@ pub use crate::{
   },
   plugin_driver::{PluginDriver, PluginDriverFactory, SharedPluginDriver},
   pluginable::Pluginable,
+  types::build_timings::BuildTimings,
   types::custom_field::CustomField,
   types::hook_addon_args::HookAddonArgs,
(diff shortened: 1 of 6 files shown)


shulaoda:
@codex review

shulaoda (approved):

PR #10521, opened 2026-07-29
fix(plugin): report plugin timings after closeBundle has run

IWANABETHATGUY:
Second of four. Stacked on #10520.

## The bug

`closeBundle` is instrumented like every other hook — `output_hooks.rs` brackets the call and records it. But the warning is pushed into `output.warnings` from `Bundle::write` / `Bundle::generate`, and those return **before** `BundleHandle::close()` invokes the hook.

So its cost is measured and then thrown away. A plugin whose expense is concentrated in `closeBundle` can never appear in `[PLUGIN_TIMINGS]`, however slow it is.

## The fix

Emit from `BindingBundler::close` instead, after the cleanup future resolves. It still goes through the existing `handle_warnings`, so `checks.pluginTimings` filters it like any other check and the message renders through `Diagnostic::render_batch` as before.

## What deferring forces

One `RolldownBuild` may `generate` more than once. Until now each output emitted its own self-consistent report; a single report at close has to account for all of them. So `BindingBundler` keeps every `BundleHandle` rather than only the last, and:

- **the gate** sums the build and link clocks across all outputs — otherwise a slow first output followed by a fast one would fall under the 3s floor and be suppressed entirely;
- **the rows** merge the per-plugin summaries by name — each output has its own driver and so its own `PluginIdx`, so the same plugin arrives once per output and would otherwise compete with itself for the five-row cap and halve its own share.

## Verification

A plugin whose only cost is a slow `closeBundle`, alongside another that makes the build qualify:

```
[PLUGIN_TIMINGS] Your build spent significant time in plugin `slow-closebundle`.
```

Before this PR it is absent from the report entirely. Plus unit tests for the cross-output merge.

### DIFF
--- a/crates/rolldown/src/bundle/bundle.rs
+++ b/crates/rolldown/src/bundle/bundle.rs
@@ -55,7 +55,7 @@ impl<Fs: FileSystem + Clone + 'static> Bundle<Fs> {
     }
     .await;
     self.plugin_driver.set_total_build_time(start);
-    self.append_plugin_timings_warning(result)
+    result
   }
 
   #[tracing::instrument(level = "debug", skip_all, parent = &self.bundle_span)]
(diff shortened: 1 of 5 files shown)


shulaoda:
@cidex review

shulaoda (approved):

PR #10523, opened 2026-07-29
docs(development-guide): update config.json link line reference

dogledogle:
Fixes an outdated line number reference, updating it from `L36-L40` to `L40-L43` so the link points to the correct VS Code settings section.

### DIFF
--- a/docs/development-guide/testing.md
+++ b/docs/development-guide/testing.md
@@ -41,7 +41,7 @@ We use Rust's built-in test framework for writing and running tests. Test cases 
 
 A data-driven test case is a folder that contains a `_config.json` file. The test runner will read the configuration from `_config.json`, bundle the input files, and execute the output files to verify the behavior.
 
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY (approved):
Verified against current `main`: `.vscode/settings.json` L40-L43 is exactly the `json.schemas` entry mapping `_config.json` to `crates/rolldown_testing/_config.schema.json`, which is the config that actually provides the auto-completion described in the surrounding text. The old `L36-L40` range was stale — it covered the tail of `files.readonlyInclude`, `git.ignoreLimitWarning`, and only the opening bracket of `json.schemas`.

Docs-only, no behavioral risk. LGTM 👍

Non-blocking follow-up idea (pre-existing, not introduced here): this link targets `blob/main/…`, so the line numbers will drift again the next time `.vscode/settings.json` changes. Sibling links in the same doc (e.g. the Bundler Options link just below) pin a commit SHA — doing the same here, or dropping the line anchor entirely, would make it self-maintaining.

PR #10526, opened 2026-07-29
feat(code-splitting): support inlining of shared deps in dynamic entries

### LINKED ISSUE #10263: [Bug]: Excessive automatic code splitting when importing a module both statically and dynamically

This module graph, where dashed lines are dynamic imports.

Image

[REPL with Rolldown 1.1.5](https://repl.rolldown.rs/#eNqVj01ugzAQha8y8iamjVxo1Q1VehI2FrUrIhgjYxIS5LtnbHCIlFVW9puf772ZmWblzDrZoDgO4YusvMs9q0k1XW+s4zvxIfueyrvsp0LqKVY6Oyq/Z0s9ba/qcRlm0Dl40NZ0QKCTwj9jA4tQFdYGB9Mq0Zp/rnNeZMGhQi6HC9bAMzj8wlwhQBiMMCTYAeRZNg62fK28XlJAoCEePj6LD6Ocaz8FTfKFpGqKQ3rE2jUGo8cW7eGGr3iDX3zvkOS8FRbvJ2zOp5VrlRstwgRv8AnvUCRqwNJmIQrxzfwNILuPvA==)

### What is expected?

The execution order is guaranteed to be:
1. `main.js`
2. Maybe at some point later, `vendor.js` and `app.js`
3. Maybe at some point later, `lazy.js`

Hence there should be three chunks: `main.js`, `vendor.js`+`app.js`, `lazy.js`.

This is what Rolloup does too: [Rollup 4.62.2 REPL](https://rollupjs.org/repl/?version=4.62.2&shareable=eyJleGFtcGxlIjoiIiwibW9kdWxlcyI6W3siY29kZSI6ImltcG9ydCgnLi9hcHAuanMnKTsiLCJpc0VudHJ5Ijp0cnVlLCJuYW1lIjoibWFpbi5qcyJ9LHsiY29kZSI6ImltcG9ydCB7IGYwIH0gZnJvbSAnLi92ZW5kb3IuanMnO1xuXG5jb25zb2xlLmxvZyhmMCgxKSk7XG5cbihhc3luYyAoKSA9PiB7XG4gIGNvbnN0IHsgZm4gfSA9IGF3YWl0IGltcG9ydCgnLi9sYXp5LmpzJyk7XG4gIGZuKCk7XG59KSgpO1xuIiwiaXNFbnRyeSI6ZmFsc2UsIm5hbWUiOiJhcHAuanMifSx7ImNvZGUiOiJpbXBvcnQgeyBmMCB9IGZyb20gJy4vdmVuZG9yLmpzJztcblxuZXhwb3J0IGZ1bmN0aW9uIGZuKCkge1xuICBjb25zb2xlLmxvZyhmMCgzKSk7XG59XG4iLCJpc0VudHJ5IjpmYWxzZSwibmFtZSI6ImxhenkuanMifSx7ImNvZGUiOiJleHBvcnQgZnVuY3Rpb24gZjAoeCkge1xuICByZXR1cm4geCAqIDIgKyAxO1xufVxuIiwiaXNFbnRyeSI6ZmFsc2UsIm5hbWUiOiJ2ZW5kb3IuanMifV0sIm9wdGlvbnMiOnt9fQ%3D%3D).

Rolldown is generating 4 chunks, one per file.

### System Info

```Shell
Online repl with v1.1.15
```

### Any additional comments?

This is somewhat related to https://github.com/rolldown/rolldown/issues/9269. The "bad" case from that issue was fixed in v1.0.0 to match the "good case". However, since v1.0.2 both examples behave the "bad way", even though they don't have the problem that https://github.com/rolldown/rolldown/pull/9398 was trying to fix (I _think_, the graph view in the Repl seems to not update when I change rolldown version)

nicolo-ribaudo:
This PR fixes #10263 by aligning Rolldown's behavior with Rollup's: modules can be inlined into dynamic entry chunks, but when doing so we must generate a fake namespace object that we use at the dynamic import site, to avoid exposing the extra exports.

This is my first time contributing to Rolldown. I split this PR into three commits hoping that it makes things easier to review:
- the first one adds a test showing the current behavior (wrong)
- the second one moves some existing code into a shared helper because I need to re-use it for the third commit (it's the code to generate the fake namespace objects). I used an LLM to help extracting this code.
- the third commit implements the actual fix.

I also used an LLM to figure out where to even make this change, and to generate the extra tests included in the last commit.


IWANABETHATGUY:
Thanks for working on this — and for splitting it into three commits, that made it much easier to review. I compared the extracted `simulated_facade_inclusion.rs` helpers against the old inline code in `chunk_optimizer.rs` and the refactor is faithful, and the reworked decision logic in `can_use_reduced_dependent_entries` keeps every previously-reachable case unchanged. The `then` check mirroring `order_wrap_host_can_expose_then_export` (including the canonical-name half) is exactly right, and the TLA coupling is safe since the whole pass is disabled when anything is TLA-tainted. The output shape matches what Rollup emits for this case, so I'm on board with the direction.

There's one correctness gap I'd like to see handled before this lands, plus a few smaller things.

### `export * from` an external module breaks the extraction

`dynamic_entry_supports_namespace_extraction` checks `resolved_exports` for a `then`, but an `export *` chain that reaches an *external* module contributes names we can't see at build time, and those bypass the check entirely:

```js
// main.js
export const app = import('./app.js');

// app.js — purely dynamic entry; shared.js gets inlined by the already-loaded pass
import { compute } from './shared.js';
export const own = compute(21) + 1;
export const done = import('./lazy.js').then((m) => m.lazyValue);
export * from 'ext'; // external
```

The merged chunk comes out as:

```js
// dist/app.js
export * from "ext";
// ...
var app_exports = /* @__PURE__ */ __exportAll({ done: () => done, own: () => own });
// ...
export { done, compute as n, own, app_exports as t };

// dist/main.js
const app = import("./app.js").then((n) => n.t);
```

Two problems fall out of this, both verified at runtime:

1. **The external names are silently dropped.** Entry-level external stars only exist as the chunk-level `export *`, and the rewritten importer never sees the chunk namespace — it reads `.t`, and `app_exports` carries no `__reExport` for the external. So `(await import('./app.js')).<anything from ext>` is `undefined` where the unbundled module (and the unmerged output) had the value. This hits every external-star entry; nothing exotic needed.

2. **If the external supplies a callable `then`, the whole import is hijacked.** The chunk namespace becomes thenable at runtime, so `import("./app.js")` assimilates through it and the `(n) => n.t` callback receives whatever `then` produced instead of the namespace — the awaited result is `undefined`. This is exactly the class `internal-docs/code-splitting/design.md` ("Thenable chunk namespaces") says must refuse the collapse — "whatever an `export * from` chain reaching an external module supplies at runtime" — and the equivalent decision in `order_wrapping.rs` guards it with `entry_level_external_module_idx.is_empty()`.

The fix I'd suggest for this PR is to make `dynamic_entry_supports_namespace_extraction` also bail when the entry's direct-or-transitive `export *` chain reaches an external module. One caveat: `chunk.entry_level_external_module_idx` isn't populated yet at this point (that happens later, in `finalize_chunk_plan` via `find_entry_level_external_module`), so the check needs its own small walk over the `IsExportStar` import records — the BFS in `find_entry_level_external_module` is the template to copy.

(Longer term we could keep the merge even here: forward the external star through the synthetic namespace with `__reExport` and stop emitting the dead chunk-level `export *` / signature exports on these purely-dynamic chunks — that would also make the `then` refusal unnecessary. But that touches entry-chunk export rendering across formats, so refusing extraction is the right scope for this PR.)

Here are the two regression fixtures, one per facet. They assert the runtime semantics rather than the chunk layout, so they stay valid whichever fix shape you choose — happy to push them onto this branch.

<details>
<summary><code>crates/rolldown/tests/rolldown/optimization/chunk_merging/already_loaded_external_star_no_extraction</code> — external names must survive</summary>

**`_config.json`**

```json
{
  "_comment": "`app.js` contains `export * from 'ext-star'` with `ext-star` external. Entry-level external stars are published as a chunk-level `export *`, which only the chunk's own namespace exposes — a `.then((n) => n.<ns>)` extraction bypasses it, so the external module's names would silently vanish from the `import()` namespace. The already-loaded analysis must therefore not group `shared.js` into the dynamic entry chunk: it stays in its own chunk and `import('./app.js')` resolves to the chunk namespace, which forwards the external names.",
  "config": {
    "input": [
      {
        "name": "main",
        "import": "./main.js"
      }
    ],
    "external": ["ext-star"]
  }
}
```

**`main.js`**

```js
export const app = import('./app.js');
```

**`app.js`**

```js
import { compute } from './shared.js';

export const own = compute(21) + 1;
export const done = import('./lazy.js').then((m) => m.lazyValue);

// The external module's names are only known at runtime; they reach the
// `import()` namespace through the chunk-level `export *`.
export * from 'ext-star';
```

**`lazy.js`**

```js
import { compute } from './shared.js';

export const lazyValue = compute(42);
```

**`shared.js`**

```js
export function compute(x) {
  return x * 2;
}
```

**`node_modules/ext-star/package.json`**

```json
{
  "name": "ext-star",
  "type": "module",
  "exports": "./index.js"
}
```

**`node_modules/ext-star/index.js`**

```js
export const marker = 'from-ext';
```

**`_test.mjs`**

```js
import { readdirSync } from 'node:fs';
import { join } from 'node:path';

const { app } = await import('./dist/main.js');
const appNs = await app;

// The `import()` namespace must include the external star re-export's names
// alongside `app.js`'s own exports, exactly like the unbundled module. A
// `.then((n) => n.<ns>)` extraction would return the synthetic namespace,
// which cannot carry names only known at runtime.
assert.strictEqual(appNs.marker, 'from-ext');
assert.strictEqual(appNs.own, 43);
assert.strictEqual(await appNs.done, 84);

const jsFiles = readdirSync(join(import.meta.dirname, 'dist')).filter((file) =>
  file.endsWith('.js'),
);

// The semantic assertions above are the contract; the chunk layout is not. One
// valid fix refuses extraction for entries whose `export *` chain reaches an
// external module (shared.js stays in its own chunk: main + app + lazy +
// shared), another forwards the external names through the synthetic namespace.
// assert.strictEqual(jsFiles.length, 4, `Expected 4 chunks but got: ${jsFiles.join(', ')}`);
```

Currently fails with:

```
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
actual: undefined
expected: 'from-ext'
```

</details>

<details>
<summary><code>crates/rolldown/tests/rolldown/optimization/chunk_merging/already_loaded_external_star_then_no_extraction</code> — a runtime `then` hijacks the rewritten import</summary>

**`_config.json`**

```json
{
  "_comment": "`app.js` contains `export * from 'ext-then'` with `ext-then` external, so its runtime namespace can carry names unknowable at build time — here a callable `then`, which makes the chunk namespace thenable (see internal-docs/code-splitting/design.md, \"Thenable chunk namespaces\"). The `.then((n) => n.<ns>)` namespace extraction would be hijacked by that `then` and observe the wrong value, so the already-loaded analysis must not group `shared.js` into the dynamic entry chunk: it stays in its own chunk and `import('./app.js')` assimilates through the external `then` exactly like the unbundled module.",
  "config": {
    "input": [
      {
        "name": "main",
        "import": "./main.js"
      }
    ],
    "external": ["ext-then"]
  }
}
```

**`main.js`**

```js
export const app = import('./app.js');
```

**`app.js`**

```js
import { compute } from './shared.js';

export const own = compute(21) + 1;
export const done = import('./lazy.js').then((m) => m.lazyValue);

// The external module supplies a callable `then` at runtime, so this star
// re-export makes the namespace thenable in a way no static check can see.
export * from 'ext-then';
```

**`lazy.js`**

```js
import { compute } from './shared.js';

export const lazyValue = compute(42);
```

**`shared.js`**

```js
export function compute(x) {
  return x * 2;
}
```

**`node_modules/ext-then/package.json`**

```json
{
  "name": "ext-then",
  "type": "module",
  "exports": "./index.js"
}
```

**`node_modules/ext-then/index.js`**

```js
// A callable `then` export makes any namespace carrying it thenable: promise
// resolution assimilates it, so `import()` settles with what `then` produces
// instead of the namespace itself.
export function then(resolve) {
  resolve({ hijacked: true });
}
```

**`_test.mjs`**

```js
import { readdirSync } from 'node:fs';
import { join } from 'node:path';

const { app } = await import('./dist/main.js');
const resolved = await app;

// `app.js`'s namespace carries the external star re-export's callable `then`,
// so `import('./app.js')` must assimilate through it — exactly like the
// unbundled module — and settle with the value that `then` produces. A
// `.then((n) => n.<ns>)` extraction would instead receive that value as `n`
// and read `.<ns>` off it, yielding `undefined`.
assert.deepEqual(resolved, { hijacked: true });

const jsFiles = readdirSync(join(import.meta.dirname, 'dist')).filter((file) =>
  file.endsWith('.js'),
);

// The semantic assertion above is the contract; the chunk layout is not. One
// valid fix refuses extraction for entries whose `export *` chain reaches an
// external module (shared.js stays in its own chunk: main + app + lazy +
// shared), another preserves the thenable through the synthetic namespace.
// assert.strictEqual(jsFiles.length, 4, `Expected 4 chunks but got: ${jsFiles.join(', ')}`);
```

Currently fails with:

```
AssertionError [ERR_ASSERTION]: Expected values to be loosely deep-equal:
actual: undefined
expected: { hijacked: true }
```

</details>

### Smaller things

- The three new fixtures all exercise plain ESM entries. The CJS and wrapped-ESM arms of `include_simulated_facade_namespace` are only reached via the facade-elimination caller today, so a CJS dynamic entry (and ideally a wrapped-ESM one) going through `apply_dynamic_entry_namespace_extractions` would be cheap insurance.
- `dynamic_dominator_with_exports_no_merge` now snapshots the merge happening, so the name pins the opposite of what the test does. Maybe rename it, or at least add a `_comment` to its `_config.json`.
- The "Thenable chunk namespaces" section in `internal-docs/code-splitting/design.md` lists the guard sites; this PR adds a third decision point that should be recorded there. The updated paragraph in `implementation.md` could also mention when extraction is refused (`then` export, `preserveModules`, `minifyInternalExports` off), since those gates are load-bearing.
- Typo in `dynamic_already_loaded.rs`: "If a module is exports a `then`".

IWANABETHATGUY:
I think the overall direction and implementation of #10526 are solid. There are still a few minor issues, though. For reference, #10533 takes a more incremental approach by splitting the changes up and addressing a few regressions at the same time.

nicolo-ribaudo:
wait, i'm working on the other feedback still.

nicolo-ribaudo:
Done :) in a82b38827c99be54d5112c7c6826cf497b00d931

IWANABETHATGUY:
Thanks again for the work here — I re-reviewed the branch and ran a detailed side-by-side against an alternative implementation. I don't think we should land it in its current shape. One blocking finding, one secondary, and a proposal for how to converge.

**Blocking: external star re-exports are dropped from the extracted namespace.**

For a dynamic entry with an entry-level `export * from '<external>'`, the merge is still applied and the call site rewrites to `.then((n) => n.t)`, but `t` is built by `__exportAll({...})` from statically-known exports only — everything the external supplies at runtime silently vanishes from the `import()` result:

```js
// app.js (dynamic entry whose shared dep gets inlined)
export * from 'node:path';
export const done = import('./lazy.js').then((m) => m.value);

// main.js
const app = await import('./app.js');
typeof app.sep; // expected 'string' — actual: 'undefined' on this branch
```

The chunk still emits `export * from "node:path"` at the chunk level, but the extraction callback discards the chunk namespace (where those names live) and returns the synthetic one. The same shape also breaks an external-supplied callable `then`: promise assimilation should settle with the value `then` forwards, but the callback reads `.t` off the assimilated value and observes `undefined`. Both fail deterministically on this branch. Rollup handles these by building the extracted namespace with `_mergeNamespaces(...)` — a *runtime* merge — and by not emitting the chunk-level star for extracted entries, so the chunk namespace never becomes thenable.

**Secondary: extraction cost on the partial-usage path.**

For this issue's own repro the `import()` usage is statically known (`.then((app) => app.done)`), so no namespace hiding is needed at all — the reads provably can't observe the extra exports. This branch pays the machinery anyway: the namespace object, the `__exportAll`/`__defProp` helpers in the chunk, and an extra `.then((n) => n.t)` microtask hop. Measured on the repro: 871 B total vs 436 B achievable without extraction (unminified).

**Proposal.**

I've put up a stacked alternative that addresses both: #10532 (merge when partial usage proves the extra exports unobservable — no extraction machinery, leaner than Rollup on this graph), #10560 (extraction for opaque usage via the existing `common_chunk_exported_facade_chunk_namespace` + `rewrite_dynamic_import_for_merged_entry` path — output **byte-identical** to this branch where both act, and it passes this PR's `already_loaded_namespace_extraction` and `already_loaded_then_export_no_extraction` fixtures unchanged), and #10559 (external stars forwarded through a runtime `__reExport` merge, verified against Rollup 4.62.3 on these exact graphs; executable reproductions of the cases above live there as `already_loaded_external_star{,_then}_extraction`).

To be clear, the direction here was right — the two implementations converge on identical output wherever this branch is correct. And the `simulated_facade_inclusion.rs` refactor is genuinely valuable; I'd love to land that as a follow-up on top of the stack rather than folded into the behavior change. Would appreciate your eyes on the stack.

<details>
<summary><b>Full comparison data</b> — per-test chunk counts &amp; correctness, emitted bytes, implementation shape, guard coverage (every cell measured by running the fixture under the named implementation)</summary>

### Setup

- **Stack**: #10532 (partial-usage fast path) ← #10560 (namespace extraction) ← #10559 (external-star runtime merge), heads `cc6f710ee` / `5a98e3550` / `628adedcf`, rebased onto `main` @ `78d6f144d`.
- **This PR**: head `13693e08`.
- Rollup column: rollup 4.62.3 on the same module graphs. All runs from clean `dist/`, deterministic across repeats.

### Per-test: chunk count and correctness

| # | Graph shape | Rollup | Stack | This PR |
|---|---|---|---|---|
| 1 | `issues/10263` — the issue repro, partial usage (`.then(m => m.done)`) | 3 ✅ | **3 ✅ (lean)** | 3 ✅ (extraction) |
| 2 | `already_loaded_namespace_extraction` — `export const app = import(...)`, opaque | 3 ✅ | 3 ✅ | 3 ✅ |
| 3 | `already_loaded_then_export_no_extraction` — entry statically exports callable `then` | — | 4 ✅ | 4 ✅ |
| 4 | `dynamic_dominator_with_exports_no_merge` — rest-destructured namespace | 3 ✅ | 3 ✅ | 3 ✅ |
| 5 | `Object.keys(ns)` observation | 3 ✅ | 3 ✅ | 3 ✅ |
| 6 | reads `ns.missing` (undefined in source semantics) | 3 ✅ | 3 ✅ | 3 ✅ |
| 7 | static `then` export (stack's own refusal fixture) | — | 4 ✅ | 4 ✅ |
| 8 | `export * from 'node:path'`, reads runtime-only `ns.sep` | 3 ✅ | **3 ✅** | 3 ❌ `sep` → `undefined` |
| 9 | `export * from '<external>'`, reads a runtime-only named export | 3 ✅ | **3 ✅** | 3 ❌ value → `undefined` |
| 10 | `export * from '<external>'` where the external supplies a callable `then` | 3 ✅ | **3 ✅** | 3 ❌ hijack value → `undefined` |
| | **union suite** | | **10/10** | **7/10** |

Rows 2–3 are **byte-identical outputs** between the stack and this PR. Rows 8–10 are one bug class: the extraction namespace enumerates only build-time-known exports, so runtime-supplied names vanish. Under the stack, #10560 alone keeps rows 8–10 at 4 chunks (correct, refused); #10559 lifts them to 3 with the runtime merge.

### Emitted bytes for the issue's own graph (unminified)

| Chunk | Stack (tier-1 lean) | This PR |
|---|---|---|
| `app.js` | **210 B** | 628 B (incl. `__defProp`/`__exportAll` helpers hidden in snapshots as `// HIDDEN [\0rolldown/runtime.js]`) |
| `lazy.js` | 121 B | 121 B |
| `main.js` | 105 B | 122 B (`.then(n => n.t)` hop) |
| **total** | **436 B** | **871 B (2.0×)** |

Where extraction is genuinely needed (opaque usage), stack and PR outputs are byte-identical — the delta exists only where the stack's tier-1 proves extraction unnecessary.

### Implementation shape

| | Stack (all 3 PRs) | This PR |
|---|---|---|
| Rust source diff | **+310 / −18** across 2 files | +345 / −181 across 5 files |
| New modules | none | `simulated_facade_inclusion.rs` (+141) |
| `chunk_optimizer.rs` | untouched | ~230 lines churned (helper-extraction refactor) |
| Machinery | reuses simulated-facade subsystem as-is | refactors it into a shared helper first |
| Known duplication | ~40 lines of `IncludeContext` construction (unifiable later by exactly this PR's refactor) | — |
| Snapshot churn outside intended fixtures | zero (across ~2000 fixtures, every run) | not measured |

### Guard-boundary coverage

| Boundary | Stack fixture | This PR |
|---|---|---|
| Partial usage → lean merge | `issues/10263` | — (always extracts) |
| Opaque namespace → extraction | `already_loaded_opaque_namespace_extraction` | `already_loaded_namespace_extraction` |
| Unexported-name read → extraction, `undefined` preserved | `already_loaded_unexported_read_extraction` | — |
| Static `then` export → refuse | `already_loaded_then_export_keeps_split` | `already_loaded_then_export_no_extraction` |
| CJS `export *` → refuse | `already_loaded_cjs_star_keeps_split` | — |
| External star → forward via runtime merge | `already_loaded_external_star_extraction` | — (broken, untested) |
| External star supplying runtime `then` | `already_loaded_external_star_then_extraction` | — (broken, untested) |

</details>

nicolo-ribaudo:
> **Blocking: external star re-exports are dropped from the extracted namespace.**
> 
> For a dynamic entry with an entry-level `export * from '<external>'`, the merge is still applied and the call site rewrites to `.then((n) => n.t)`, but `t` is built by `__exportAll({...})` from statically-known exports only — everything the external supplies at runtime silently vanishes from the `import()` result:
> 
> ```js
> // app.js (dynamic entry whose shared dep gets inlined)
> export * from 'node:path';
> export const done = import('./lazy.js').then((m) => m.value);
> 
> // main.js
> const app = await import('./app.js');
> typeof app.sep; // expected 'string' — actual: 'undefined' on this branch
> ```
> 
> The chunk still emits `export * from "node:path"` at the chunk level, but the extraction callback discards the chunk namespace (where those names live) and returns the synthetic one. The same shape also breaks an external-supplied callable `then`: promise assimilation should settle with the value `then` forwards, but the callback reads `.t` off the assimilated value and observes `undefined`. Both fail deterministically on this branch. Rollup handles these by building the extracted namespace with `_mergeNamespaces(...)` — a _runtime_ merge — and by not emitting the chunk-level star for extracted entries, so the chunk namespace never becomes thenable.

I don't think this is correct? I added a test case that shows that in that case `typeof app.sep` is a string, in d43e580 (**EDIT**: 20d3cc6 now that I rebased).

nicolo-ribaudo:
> **Secondary: extraction cost on the partial-usage path.**
> 
> For this issue's own repro the `import()` usage is statically known (`.then((app) => app.done)`), so no namespace hiding is needed at all — the reads provably can't observe the extra exports. This branch pays the machinery anyway: the namespace object, the `__exportAll`/`__defProp` helpers in the chunk, and an extra `.then((n) => n.t)` microtask hop. Measured on the repro: 871 B total vs 436 B achievable without extraction (unminified).

Good suggestion, done in 73d0bfd4808f805678b788e6b1d12b30afc0d0f7 (**EDIT**: afcb666 now that I rebased).

IWANABETHATGUY (approved):
Thanks!

PR #10529, opened 2026-07-30
fix(binding): stop reporting the hotUpdate hook for parallel JS plugins

### LINKED ISSUE #10499: Parallel JS plugins: `hotUpdate` sets the hook-usage bit but is never forwarded
Found in review of #10305 — the note without an inline anchor in [this review](https://github.com/rolldown/rolldown/pull/10305#pullrequestreview-4762608947).

**The gap.** A plugin created with `defineParallelPlugin` whose worker implementation declares `hotUpdate` switches the engine's per-file hook chain on, but the hook itself never runs:

- `ParallelJsPlugin::register_hook_usage` delegates to the first worker plugin's real usage bits (`crates/rolldown_binding/src/options/plugin/parallel_js_plugin.rs`), and the JS side sets the `hotUpdate` bit when the implementation declares the hook (`packages/rolldown/src/plugin/generated/hook-usage.ts`).
- The engine gates the per-file chain on that bit (`has_hot_update_hook()` in `crates/rolldown/src/hmr/hmr_stage.rs`).
- But `ParallelJsPlugin` has no `hot_update` override, so the call falls through to the `Plugin` trait's default no-op. The workers never see it.

Net effect: the engine pays for the chain on every file change, and the plugin author silently gets nothing. This is the same declared-but-never-forwarded pattern that #10345 fixed for `closeWatcher`, `watchChange`, `renderStart`, and the other missing hooks — `hotUpdate` landed after that sweep (in #10305) and is the one hook left out.

```mermaid
flowchart LR
    W["worker plugin declares hotUpdate"] --> B["usage bit set — engine runs the per-file chain"]
    B --> P["ParallelJsPlugin has no hot_update override"] --> N["trait default no-op — hook silently skipped"]
```

Minimal example:

```js
// impl.js — loaded on each worker thread
export default defineParallelPluginImplementation(() => ({
  name: 'my-parallel',
  transform(code) {
    // runs on workers as expected
  },
  hotUpdate({ file }) {
    // never runs — but declaring it still enables the engine chain
  },
}))
```

**Fix options** (from the review):

1. Mask the `hotUpdate` bit in `ParallelJsPlugin::register_hook_usage`, so the engine honestly reports "no hook" and skips the chain.
2. Or document that parallel plugins do not support `hotUpdate`.

Actually forwarding the hook to a worker is possible but not needed today: `hotUpdate` is an internal dev-only hook, and its only consumer (Vite's bundled-dev adapter) runs on the main thread.

Refs: #10305 (the hook), #10345 (the earlier forwarding sweep), #10491 (separate `hotUpdate` follow-up about asset entries).

Nic-Polumeyv:
`ParallelJsPlugin` reports the first worker plugin's hook usage verbatim, including `hotUpdate`, which it has no override for, so the call reaches the trait's no-op default. Mask the bit so the engine is not told about a hook this plugin cannot answer.

Fixes #10499.

### DIFF
--- a/crates/rolldown_binding/src/options/plugin/parallel_js_plugin.rs
+++ b/crates/rolldown_binding/src/options/plugin/parallel_js_plugin.rs
@@ -18,6 +18,11 @@ use crate::worker_manager::WorkerManager;
 use super::BindingPluginOptions;
 use super::JsPlugin;
 
+/// Hooks a worker plugin can declare but this plugin has no override for, so the call would
+/// reach the `Plugin` trait's no-op default. Reporting them would only make the engine run the
+/// hook's chain for a plugin that cannot answer it.
+const UNFORWARDED_HOOKS: HookUsage = HookUsage::HotUpdate;
+
 #[derive(Debug)]
 #[cfg_attr(target_family = "wasm", allow(unused))]
 pub struct ParallelJsPlugin {
@@ -407,6 +412,6 @@ impl Plugin for ParallelJsPlugin {
   }
 
   fn register_hook_usage(&self) -> HookUsage {
-    Plugin::register_hook_usage(self.first_plugin())
+    Plugin::register_hook_usage(self.first_plugin()) & !UNFORWARDED_HOOKS
   }
 }


Nic-Polumeyv:
Closing this because of #10530.

PR #10532, opened 2026-07-30
fix(code-splitting): excessive splitting of static+dynamic shared deps

### LINKED ISSUE #10263: [Bug]: Excessive automatic code splitting when importing a module both statically and dynamically

IWANABETHATGUY:
## What is this PR solving?

Fixes #10263 — a module imported both statically and dynamically splits into its own chunk even when it is guaranteed to be already loaded.

For the issue's graph (`main.js` → `import('./app.js')`; `app.js` and `lazy.js` both statically import `vendor.js`; `lazy.js` is only dynamically imported from `app.js`), Rolldown emitted 4 chunks where Rollup emits 3. The already-loaded pass (`dynamic_already_loaded.rs`) correctly reduces `vendor`'s bits to `{app}`, but the namespace guard then rejected the merge because `vendor`'s exports are not part of `app`'s signature — the extra chunk export would leak onto the `import('./app.js')` namespace.

## How

Add one acceptance condition to that guard: the extra exports are provably **unobservable**. For a pure dynamic entry, when every `import()` of it reads a statically known name set (`DynamicImportExportsUsage::Partial`) and each used name resolves to one of the entry's non-ambiguous exports, no recorded read can reach a merged binding:

- dynamic-entry export retention is driven by the same usage set, so the used names always stay exported under their own names, and
- generated names for the merged atom's exports are deconflicted against the entry's export names on both naming paths (minified base54 and the `ConflictResolver`, which processes the entry module first via `deconflict_order_key`).

Reads of names the entry does not export (source semantics `undefined`) keep the guard, since a generated name could otherwise shadow them. Opaque namespace uses (`Object.keys`, passing the namespace around) are `Complete` and keep the old behavior — the stacked PRs above this one handle those.

Output for the issue's repro (3 chunks, and leaner than Rollup's — no namespace object, no runtime helper, no extra promise hop):

```js
// app.js
function f0(x) { return x * 2 + 1; }
console.log(f0(1));
const done = import("./lazy.js").then(({ fn }) => { fn(); });
export { done, f0 as t };
// main.js
const done = import("./app.js").then((app) => app.done);
```

## Alternatives explored

Rollup accepts the reduction unconditionally and repairs the namespace at render time (`generateFacades` + `.then(n => n.a)` rewrites). That machinery is layered on top of this PR as #10560 / #10559; this PR keeps the cases that need no repair at all maximally lean.

## Tests

`crates/rolldown/tests/rolldown/issues/10263/` reproduces the issue graph and asserts 3 chunks plus runtime behavior (`_test.mjs` executes the output). Full Rust suite passes with no other snapshot changes.

---
Stack: **#10532 (this)** ← #10560 ← #10559


IWANABETHATGUY:
Closing in favour of #10526, which has since implemented the same partial-usage lean path — its `issues/10263` output is now byte-identical to this branch (436 B). Branch preserved locally at `cc6f710ee`.

PR #10544, opened 2026-07-30
fix(code-splitting): materialize consumer-local barrel routing when the order-wrap plan is empty

### LINKED ISSUE #10515: strictExecutionOrder: pre-chunk consumer-local barrel routing without matching lowering emits a startup-crashing CJS chunk cycle
### Summary

With `strictExecutionOrder`, the pre-chunk consumer-local barrel probe can make a routing decision that final lowering never materializes. Chunk placement then splits a CJS re-export barrel's leaves per consumer while the emitted barrel keeps its monolithic interop wrapper, producing chunks that `require()` each other in a cycle. Loading either entry crashes at startup:

```
TypeError: require_barrel.init_barrel is not a function
```

Regression introduced by #10488 (part of the #10294 strict-order work). Bisected to exactly that merge:

| Commit | Result |
| --- | --- |
| `f5a53205e` (main, merge parent) | pass |
| `9fef0099f` (main, squash-merge of #10488) | **crash** |

Both wrap-all and `experimental.onDemandWrapping` are affected.

### Reproduction

Seven one/two-line modules. Two CJS entries each `require()` their own ESM forwarder into one shared re-export-only barrel over two CJS leaves; output format `cjs`:

```js
// entry-a.cjs
const { a } = require('./forwarder-a.mjs');
module.exports = { a };

// entry-b.cjs
const { b } = require('./forwarder-b.mjs');
module.exports = { b };

// forwarder-a.mjs
export { a } from './barrel.js';

// forwarder-b.mjs
export { b } from './barrel.js';

// barrel.js
export { a } from './a.cjs';
export { b } from './b.cjs';

// a.cjs
exports.a = 'a';

// b.cjs
exports.b = 'b';
```

Options: `input: [entry-a.cjs, entry-b.cjs]`, `format: 'cjs'`, strict execution order enabled (no other non-default options; no `sideEffects` annotations needed).

As a rolldown Rust fixture (validated form), drop the files above plus this `_config.json` and `_test.mjs` into `crates/rolldown/tests/rolldown/function/experimental/strict_execution_order/cjs_reexport_barrel_probe_plan_divergence/` and run `cargo test -p rolldown --test integration probe_plan_divergence`:

```json
{
  "configName": "wrap-all",
  "config": {
    "input": [
      { "name": "entry-a", "import": "./entry-a.cjs" },
      { "name": "entry-b", "import": "./entry-b.cjs" }
    ],
    "strictExecutionOrder": true,
    "format": "cjs"
  },
  "configVariants": [{ "_configName": "on-demand", "onDemandWrapping": true }]
}
```

```js
// _test.mjs

const entryA = await import('./dist/entry-a.js');
assert.strictEqual(entryA.a, 'a');

const entryB = await import('./dist/entry-b.js');
assert.strictEqual(entryB.b, 'b');
```

### Observed output

The barrel chunk keeps the monolithic interop wrapper but its leaves were placed into the two entry chunks, so every chunk requires the others:

```js
// dist/barrel.js
const require_rolldown_runtime = require("./rolldown-runtime.js");
const require_entry_a = require("./entry-a.js");
const require_entry_b = require("./entry-b.js");
var import_a, import_b;
var init_barrel = require_rolldown_runtime.__esmMin((() => {
	import_a = require_entry_a.require_a();
	import_b = require_entry_b.require_b();
}));
```

`entry-a.js` hosts `require_a` (and requires `barrel.js`), `entry-b.js` hosts `require_b`. Requiring `entry-a.js` therefore loads `barrel.js`, which loads `entry-b.js`, whose body calls `require_barrel.init_barrel()` while `barrel.js` is still mid-load — `init_barrel` is not exported yet, so it throws. Loading one entry also executes the other entry's module body along the way.

On `f5a53205e` both leaves are placed in the shared barrel chunk and both entries load correctly.

### Mechanism

- `generate_chunks` builds a pre-chunk probe (`pre_chunk_order_state`) that runs `consumer_local_reexport_plan` unconditionally and routes entry reachability per consumer (`determine_reachable_modules_for_entry`), so `a.cjs` gets only entry-a's bit and `b.cjs` only entry-b's.
- In this graph every reachable ESM module is already `WrapKind::Esm` (the CJS entries `require()` the forwarders, interop-wrapping everything), so the order-wrap plan is empty. `apply_order_wraps` takes its `plan.is_empty()` short-circuit (`order_wrapping.rs:432`) and never reaches `lower_order_state` — the consumer-local plan is never applied and no per-record carriers exist. The short-circuit's comment ("safe only because an empty plan makes each one a no-op") no longer holds: `consumer_local_reexport_plan` eligibility is plan-independent.
- Result: placement assumed per-record carriers; emission produced the monolithic barrel. The probe also cannot be rescued by the on-demand force-add (`order_analysis.rs`, `all_at_risk.extend(carrier importers)`) because the interop-wrapped barrel is never order-wrap-eligible, which is why both strict modes fail.

Minimality notes (each verified): one shared forwarder instead of two makes routing union the demand (no split, passes); collapsing the two leaves into one file gives the leaf both entry bits (placed beside the barrel, passes); any plain `WrapKind::None` ESM module in the graph makes the plan non-empty, lowering runs, and the output is correct.

### Suggested direction

Run the consumer-local lowering even when the ordinary order-wrap plan is empty (move the `consumer_local_reexport_plan` application out of the `plan.is_empty()` short-circuit), or gate the pre-chunk probe on lowering actually running. The former preserves the #10488 optimization for require-heavy graphs, which are exactly its target.

### Related but distinct

The same seven-file shape with default **esm** output already fails on pre-merge main (`entryA.a === undefined`, silently — no chunk split involved, so it is a different, older bug and stays broken on both sides of the #10488 merge). Filing separately; fixing this issue will not fix that one.

hyfdev:
With `strictExecutionOrder`, a CJS multi-entry build whose entries `require()` ESM forwarders into one shared re-export-only barrel over CJS leaves emits chunks that `require()` each other in a cycle and crash at startup:

```
TypeError: require_barrel.init_barrel is not a function
```

Fixes #10515. Regression from #10488, shipped in v1.2.1.

## Problem

Consumer-local re-export routing is computed in three places: the pre-chunk placement probe, the on-demand projector probes, and real lowering. The probes run `consumer_local_reexport_plan` unconditionally, and its eligibility never reads the order-wrap plan — but real lowering sits behind the `plan.is_empty()` short-circuit in `apply_order_wraps`.

When every reachable ESM module is already interop-wrapped (the CJS entries `require()` the forwarders), the order-wrap plan is empty. Chunk placement has already routed each barrel leaf to only its consuming entry's chunk, while emission keeps the barrel's monolithic interop body referencing the per-entry `require_*` wrappers — so the emitted chunks require each other in a startup cycle.

The on-demand carrier force-add cannot rescue this: an interop-wrapped barrel never passes `is_order_wrap_eligible`, so the final plan can stay empty even when the wrap-all plan is not. That also rules out fixing this on the probe side — the probe runs before chunking, while the on-demand plan is only computed from the chunk graph afterwards.

## Fix

Compute the consumer-local plan before the short-circuit and take the fast path only when both the order-wrap plan and the consumer-local plan are empty. Real lowering then materializes exactly the routes placement consumed, in both strict modes, and `lower_order_state` receives the precomputed plan instead of recomputing it.

One exception keeps an existing contract: with code splitting disabled and at most one entry, placement has a single reachability bit, so the probe's routes moved nothing, and an inert on-demand plan keeps its promise of flag-off-equivalent output (pinned by `cjs_reexport_barrel_code_splitting_disabled` from #10514). The disabled option alone is not a sufficient key — only dynamic entries are dropped in disabled mode, so a plugin-emitted chunk still adds a second entry whose per-entry placement consumes the probe's routes — hence the exception is keyed on entry count as well.

## Test

- `cjs_reexport_barrel_probe_plan_divergence`: the seven-file reproduction from #10515, wrap-all + on-demand variants. Crashes on current `main`, passes with the fix; the barrel becomes an empty routing waypoint and each entry initializes only its own leaf through a per-record memoized carrier.
- `cjs_reexport_barrel_on_demand_inert_plan`: adds a plain-ESM third entry so the wrap-all plan is non-empty while the on-demand plan stays empty — the shape a probe-side gate could not fix. On current `main` the wrap-all variant passes and the on-demand variant crashes; both pass with the fix.
- No existing snapshot changes — every previously covered configuration emits byte-identical output.


IWANABETHATGUY (commented):
Nice write-up in the description — I traced the three-way split (probe / projector / real
lowering) and the fix reads right to me. Couple of things I checked independently and they
hold up: `consumer_local_reexport_plan` never reads `input.plan`, so the probe and lowering
really do reach the same decision, and hoisting the call above the wrapper loop in
`lower_order_state` is behaviorally a no-op since that loop only touches
`insert_order_wrapper` / `set_reexport_init_transparent`. I also convinced myself the
exception is earning its keep: in the on-demand variant of
`cjs_reexport_barrel_code_splitting_disabled` the barrel is `WrapKind::Esm`, so the plan is
empty but it still owns a carrier — pull the guard and that snapshot flips.

None of the below blocks merging as far as I'm concerned — the fix is right and this is a
shipped crash, so landing it promptly is worth something on its own. A few things I'd like to
see follow, the first one ideally as a commit on this branch:

- **The `entries.len() <= 1` half of the guard isn't covered by anything.** The description
  makes a good case that it's load-bearing (emitted chunk → second entry → per-entry placement
  consumes the routes), but all five strict-order fixtures with `codeSplitting: false` are
  single-entry, so deleting that clause passes CI while quietly restoring the #10515 cycle for
  multi-entry non-splitting builds. That's exactly the kind of clause someone simplifies away
  in six months. Cheapest fix I can see is a `codeSplitting: false` config variant on
  `cjs_reexport_barrel_probe_plan_divergence` — by your own reasoning that build still has to
  route, so it should fail today with the clause removed and pass with it.

- **The new `_test.mjs` files only assert values.** The thing that actually makes this fix
  interesting — entry-a initializes `a.cjs` and never touches `b.cjs` — lives only in the
  snapshot right now, so a future snapshot regression sails past the test. The sibling fixture
  you're citing asserts both the `__events` ordering and the structural difference between
  variants (`assert.doesNotMatch(main, /init_pure_barrel_cjs/)`); pushing to
  `globalThis.__events` from the two leaves and asserting per-entry lists would pin the real
  guarantee and match what the rest of the suite does.

- **The empty-plan path isn't cheap anymore.** `synchronous_cycle_modules` (tarjan over the
  whole module graph) plus the `consumer_local_reexport_plan` fixpoint now run on a path that
  used to return immediately, and in on-demand mode an empty plan is the common healthy case.
  The exception branch is the annoying one: it builds the SCC and the whole
  `OrderLoweringInput` and then throws both away. An early return keeps that case exactly as
  cheap as it was:

      if plan.is_empty() && code_splitting_disabled && self.link_output.entries.len() <= 1 {
        // …existing fast path…
      }
      // only now build cyclic_modules + input and compute the plan

  (CodSpeed being flat doesn't say much here — none of the benchmarks turn on
  `strictExecutionOrder`.)

- **Follow-up, not for this PR:** the "all three phases must agree" invariant is currently held
  up by a doc comment. The plan is pure over `ModuleIdx`/`SymbolRef` values that survive
  chunking, so `pre_chunk_order_state` could compute it once and hand it to both the projector
  and `apply_order_wraps` — that makes the invariant structural instead of documented and drops
  two redundant fixpoints and two redundant SCC passes. Worth an issue; this is the second bug
  that's come out of the three copies drifting.

- **Also follow-up:** in both new snapshots `barrel.js` is an entire emitted chunk holding
  `var init_barrel = __esmMin(() => {})` and its `defineProperty` export, and each entry
  carries a bare `require("./barrel.js")` that does nothing. Correct, just an empty file plus a
  startup require per entry. Collapsing a barrel that routing has emptied out seems worth
  tracking separately.

Nits, take or leave:

- `ConsumerLocalReexportPlan::is_empty` — carriers only get pushed alongside their owner
  module, so `carriers.is_empty()` is already implied by `modules.is_empty()`. Harmless, but
  the `&&` hints at a state that can't happen.
- `self.link_output.entries.len()` counts distinct entry *modules*, while `generate_chunks`
  deliberately uses the flattened entry-*point* count and has a comment saying so. Unique
  modules is the right quantity here (duplicate entry points on one module produce identical
  bit sets), but a half-line note would save the next person the round trip I just made.
- The `internal-docs/code-splitting/implementation.md` addition tacks another ~700 characters
  onto a paragraph that was already pretty dense. Content's right — maybe break the invariant
  out into its own short paragraph?

IWANABETHATGUY (approved):

hyfdev:
Addressed in 73b525e9d and 60121af16, with two deviations worth flagging:

- **Coverage for the `entries.len() <= 1` clause** — the suggested `codeSplitting: false` variant on `cjs_reexport_barrel_probe_plan_divergence` can't be expressed: that fixture has two `input`s, and disabled splitting with multiple inputs is rejected at option validation. The reachable shape is the emitted-chunk one, so the coverage landed as `disabled_splitting_emitted_entry` in the strict invariants suite: one input plus a plugin-emitted second entry over the same seven-file graph, `codeSplitting: false`, both strict modes, executed with Node. Verified it earns its keep: deleting the entry-count clause makes it crash with the #10515 cycle (`require_barrel.__commonJSMin is not a function`); the build really does split (`entry-a.js`, `chunks/entry-b.js`, `chunks/barrel.js`).

- **Behavioral pins in the new `_test.mjs` files** — tried `__events` first and it can't pin this graph: observable side effects in the leaves flip their carriers to eager by design, after which every consumer correctly initializes both leaves (the run showed `['a','b']` on entry-a — the spec answer for effectful leaves, and incidentally a live check that the eager-carrier path works under an empty plan). The fixtures now pin the routing structurally instead — each leaf's body may appear only in its consuming entry's chunk, and the barrel chunk may not retain any monolithic leaf init — with a comment explaining why events can't work here. The events approach stays meaningful only with side-effect-free leaves, which the existing sibling fixtures already cover.

- Fast path restructured as sketched: the single-entry disabled case returns before the SCC scan and the fixpoint; the shared tail is `apply_inert_order_wrap_plan`. Nits all taken. Agreed on both follow-ups — happy to file the issues; the shared-computation one should also fold in the third copy in `analyze_execution_order`'s `placement_state`.

hyfdev:
### Merge activity

* **Jul 31, 9:10 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Jul 31, 10:02 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Jul 31, 10:10 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10554, opened 2026-07-30
fix: reduce finalization memory usage

ericclemmons:
**tl;dr – Bound chunk-finalization memory and return unused mimalloc pages after builds.**

`finalize_assets` currently materializes the transitive dependency closure for every output chunk at the same time. On output graphs with many shared chunks, this makes the retained dependency data grow quadratically. A Cloudflare Docs production build with 19,035 chunks exceeds an 8 GB cgroup while constructing these closures.

This calculates each chunk's transitive closure only while computing that chunk's hash, preserving the existing deterministic traversal and hash semantics without retaining all closures simultaneously. ~~It also forces a mimalloc collection after successful N-API `generate` and `write` calls so downstream build stages can reuse pages freed by Rolldown.~~

With the Workers Builds reproduction (`8,000,000,000` bytes, 8 CPUs), the previously OOM-killed build completes all 8,734 pages and 1,635 cold image optimizations in 5m53s. Peak cgroup usage is 7.78 GB and the container exits 0 with `OOMKilled=false`.

---

References: https://github.com/rolldown/rolldown/issues/9330


crates/rolldown_binding/src/binding_bundler.rs
```diff
@@ -80,7 +80,9 @@ impl BindingBundler {
         return Ok(napi::Either::A(BindingErrors::new(vec![error])));
       }
 
-      Ok(napi::Either::B(bundle_output.assets.into()))
+      let outputs = bundle_output.assets.into();
+      crate::collect_allocator();
```
shulaoda: Could you help me confirm that what's the main part reducing the memory, `mi_collect(true)` or the logic change in `finalize_chunks`? I think `mi_collect(true)` only do a deep clean on the thread it's running, and for other threads, it's a cross-thread memory free, so it shouldn't have a big impact.

shulaoda: Just remove the `collect_allocator` related logic and check whether it still resolve your problem

ericclemmons: Resolved in [698e645](https://github.com/rolldown/rolldown/pull/10554/commits/698e645dc99e99727d7cb2897b35ba456cdd360a)

shulaoda:
Thank you so much! It's a big improvement! 🚀

I verified it locally with the real Cloudflare Docs build (8,785 pages, ~19,000 hashed chunks). Peak memory of the whole build dropped by ~21%, from ~10 GB to ~7.9 GB. The bundling phase itself dropped by ~35% and build time stayed the same.

shulaoda (approved):
LGTM! I just removed unnecessary unit test cases and unused `drain_deferred_drops` helper function.

PR #10558, opened 2026-07-31
perf(plugin): clock the build only when it is being measured

IWANABETHATGUY:
Sits on #10509.

## What

`BuildTimings` took an `Instant` on every build, whether or not anyone would read it.

Before #10520 the two clocks lived on the hook timing collector, which only existed when timing was on — so the `Option<Instant>` it handed back *was* the gate. Giving the clocks their own type dropped that, and `bundle.rs` went to a bare `std::time::Instant::now()` in three places.

The gate goes back, this time owned by the type that needs it:

- `BuildTimings::new(measured)` — decided once, where the driver is built.
- `start()` returns `Option<Instant>`: `None` when the build is not being measured.
- `record_total` / `record_link_stage` take what `start()` gave and do nothing with `None`.

The call sites read as they did before the split.

## What decides it

`options.plugin_timings`. The JavaScript side registers that callback only when `measureTimings && checks.pluginTimings !== false`, so its presence is already how this side knows a report is coming. The two halves cannot disagree about whether a build is measured, because a report needs both of them.

## Tests

Two added: a build that is not measured hands back no instant and stores nothing; a measured one does start a clock. The existing gate tests are unchanged.

## Not in scope

The threshold and the ratio, and what any measured build reports. If this PR changes what a build with `checks.pluginTimings` on reports, it is wrong.

### DIFF
--- a/crates/rolldown/src/bundle/bundle.rs
+++ b/crates/rolldown/src/bundle/bundle.rs
@@ -43,7 +43,7 @@ impl<Fs: FileSystem + Clone + 'static> Bundle<Fs> {
   #[tracing::instrument(level = "debug", skip_all, parent = &self.bundle_span)]
   /// This method intentionally get the ownership of `self` to show that the method cannot be called multiple times.
   pub async fn write(mut self) -> BuildResult<BundleOutput> {
-    let start = std::time::Instant::now();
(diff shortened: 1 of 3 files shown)


shulaoda (approved):

shulaoda:
@codex review

PR #10559, opened 2026-07-31
feat(code-splitting): forward external star re-exports through extracted namespaces

IWANABETHATGUY:
## What is this PR solving?

Part of #10263 (stacked on #10560). Final Rollup-parity gap for the already-loaded inlining: a dynamic entry with an entry-level `export * from <external>` still refused namespace extraction (`has_dynamic_exports`), keeping one more chunk than Rollup for such graphs.

Rollup handles this with a runtime merge (verified against Rollup 4.62.3 on these exact graphs): the extracted namespace is built by `_mergeNamespaces({...static}, [externalNs])`, so names only known at runtime — including a callable `then` supplied by the external — survive, and the chunk itself carries no `export * from <external>`, keeping the chunk namespace non-thenable.

## How

Two small changes express the same design with existing machinery:

- `dynamic_entry_namespace_extraction_allowed` accepts an entry whose dynamic exports are explained entirely by its own **direct** external star records (`dynamic_exports_explained_by_direct_external_stars`). A CJS `export *` or a star chain that only reaches an external transitively keeps the refusal.
- `find_entry_level_external_module` skips the entry-level flatten for extraction-registered entries. That one skip does double duty: the materialized namespace keeps the star records as runtime `__reExport` merges (Rolldown's `_mergeNamespaces` equivalent), and the suppressed chunk-level `export *` keeps the chunk namespace non-thenable, so an external-supplied `then` assimilates exactly where the unbundled module would — after the `.then(n => n.<alias>)` callback returns the merged namespace.

The namespace statement's self-declared symbols (an external star record's namespace binding, consumed by its own `__reExport` call) now survive the getter narrowing.

Emitted shape:

```js
// app.js — no chunk-level `export * from 'ext-then'`
var app_exports = /* @__PURE__ */ __exportAll({ done: () => done });
import * as import_ext_then from "ext-then";
__reExport(app_exports, import_ext_then);
export { done, f0 as n, app_exports as t };
// main.js
const app = import("./app.js").then((n) => n.t);
```

## Tests

- `already_loaded_external_star_extraction` — `node:path`'s `sep` (runtime-only name) observable through the extracted namespace; 3 chunks
- `already_loaded_external_star_then_extraction` — the external's callable `then` hijacks assimilation with `{hijacked: true}`, matching the unbundled module and Rollup; 3 chunks

Both execute the emitted output. Full Rust suite passes.

---
Stack: #10532 ← #10560 ← **#10559 (this)**


IWANABETHATGUY:
Closing for now — we're merging #10526 first. This layer's external-star runtime `__reExport` merge (forwarding runtime-only names, incl. a callable `then`, through the extracted namespace instead of refusing the merge) is the one capability #10526 doesn't have, so it'll come back as a follow-up on top of it. Branch preserved locally at `628adedcf`.

PR #10560, opened 2026-07-31
feat(code-splitting): namespace extraction for already-loaded dynamic entry merges

IWANABETHATGUY:
## What is this PR solving?

Part of #10263 (stacked on #10532; supersedes #10533, which could not be reopened after its branch was amended).

#10532 merges an already-loaded shared dependency only when partial `import()` usage proves the extra exports unobservable. When the namespace is observed opaquely (`export const app = import('./app.js')`, `Object.keys(ns)`, rest-destructuring) or a read targets an unexported name, Rolldown still split — Rollup emits 3 chunks there via its `generateFacades` fallback (`Chunk.ts` includes `module.namespace` in the chunk, exports it, and call sites extract it with `.then(n => n.a)`).

## How

Rolldown already owns that exact machinery for eliminated facade chunks: `common_chunk_exported_facade_chunk_namespace`, the narrowed simulated namespace, the cross-chunk namespace export, and the finalizer's `rewrite_dynamic_import_for_merged_entry` (`import('./chunk.js').then(n => n.<alias>)`, all formats and wrap kinds). This PR opts the surviving dynamic entry chunk into it from the already-loaded pass: register the entry in the map, narrow its namespace statement to link-retained getters, include the namespace/wrapper symbols (`SymbolIncludeReason::SimulatedFacadeChunk`), and record the `__exportAll` demand before standalone runtime extraction so placement stays normal.

The #10532 fast path still wins when provable, so the issue's own graph keeps the leaner no-extraction output.

Extraction is refused when the simulated namespace cannot reproduce the entry's interface:
- a statically known `then` export — the chunk namespace would become thenable and `import()` would assimilate before the extraction callback runs (`already_loaded_then_export_keeps_split`);
- dynamic exports, e.g. a CJS `export *` (`already_loaded_cjs_star_keeps_split`) — the external-star case is then lifted by #10559 on top of this PR.

## Tests

- `already_loaded_opaque_namespace_extraction` — `Object.keys` observes exactly the entry's exports; 3 chunks
- `already_loaded_unexported_read_extraction` — `ns.missing` stays `undefined` through the extracted namespace; 3 chunks
- `already_loaded_then_export_keeps_split`, `already_loaded_cjs_star_keeps_split` — refusals pinned; 4 chunks with source-faithful runtime behavior
- `dynamic_dominator_with_exports_no_merge` — now inlines `sharedDynamic` into `dynamic1` (snapshot-only change; its runtime assertions are unchanged)

All `_test.mjs` files execute the emitted output. Full Rust suite passes.

---
Stack: #10532 ← **#10560 (this)** ← #10559


IWANABETHATGUY:
Closing in favour of #10526, which now implements the same namespace-extraction fallback and the same `has_dynamic_exports` guard. Verified equivalent: its output for the extraction fixtures is byte-identical to this branch, and its full suite is green. Branch preserved locally at `5a98e3550`.

PR #10564, opened 2026-07-31
fix(transform): transformer gate skips newer syntax that oxc can lower

sekyungk:
## Summary

- Run the Oxc transformer for JavaScript files when the target lacks Explicit Resource Management (`using` declarations).
- Simplify `should_transform_js()`. Its `Normal`/`Raw` arms always returned the same result. Now it checks `self.target` directly.
- Add an es2024 regression fixture.

## Problem

`should_transform_js()` only checked whether the target needed Unicode Sets RegExp transformed. Targets such as Chrome 133 support Unicode Sets RegExp but not Explicit Resource Management, so the transformer never ran and `using` declarations were emitted unchanged. To fix this, it now also checks whether the target needs Explicit Resource Management lowered.

Along the way, `should_transform_js()` itself is cleaned up. The env flags that the `Normal` variant used to read are computed from the same `target` by oxc's `EnvOptions::from(EngineTargets)` via `has_feature()`, so they always agree with the `Raw` variant's `self.target.has_feature()` checks by construction (tsconfig merging never touches `target`). The per-variant branching is replaced with a single expression over `self.target`.

The new es2024 fixture verifies that `using` is lowered through the default build path.

## AI usage

Drafted with Codex and double-checked with Claude. I reviewed and verified all changes myself.


IWANABETHATGUY:

IWANABETHATGUY:
Nice find — this is a real bug, and the reasoning in the description holds up. I checked the chain you assert: `BundlerTransformOptions` has no `env` field, `TryFrom` derives `env` from `options.target` via `EnvOptions::from_target` (which is literally `EngineTargets::from_target(s).map(Self::from)`), `prepare_build_context.rs` builds `TransformOptions.target` from the same string, and `tsconfig_merge.rs` only ever reads `compilerOptions.target` for `useDefineForClassFields`. So the two variants really do agree by construction — the unification is sound, not just plausible.

I also confirmed your fixture is a genuine regression test: reverting only `should_transform_js` to the old body flips the snapshot back to raw `using resource = acquireResource();`.

Since maintainer edits are on, I pushed one commit (000d6212d) rather than round-tripping review comments. Happy to drop or rework any of it — details below.

## 1. Derive the gate from every lowerable feature

The two-feature check is correct *today*, but only by accident of ordering: unicode-sets is ES2024, so for pure ES targets it subsumes every older lowering, and ERM is the newest thing oxc can lower. `EnvOptions` has ~20 flags and the gate named 2 of them, so the next oxc release that gains a lowering silently reopens exactly this bug — and nothing in the type system or the test suite would catch it.

It now derives from `LOWERABLE_ES_FEATURES`, the full set oxc's `impl From<EngineTargets> for EnvOptions` turns into transform flags. The ES2025 RegExp features stay out deliberately: oxc knows them but has no transform option for them, so listing them would buy a pass that cannot change anything.

The guard against this rotting again is `tests::env_lowers_anything`, which destructures `EnvOptions` and every ES-edition group exhaustively — when an oxc upgrade adds a group, as ES2026 once was, it stops compiling. One gap I could not close: `RegExpOptions` lives in a private oxc module, so it cannot be named in a pattern and its flags are read field by field. A brand-new RegExp lowering would still slip past. Noted in place.

Verified the test bites: dropping `ES2026ExplicitResourceManagement` from the list fails on `es2024`.

## 2. Precompute it

Two reasons. The widened list walks oxc's static feature table once per entry — a `OnceLock` load, a hash into the feature map, then a map lookup per engine inside it — and `PreProcessEcmaAst` asks this once per JS module. Separately, on `main` the `Normal` arm was a single `bool` field read (`opts.env.regexp.set_notation`), and this PR turned it into two `has_feature` calls; small, but it was a step backwards on the hot path.

It is now a `bool` computed in `new`/`new_raw`/`Default`, so the per-module cost is a field read again and widening the list costs nothing at runtime.

## 3. Two more fixtures

**One correction to something I initially got wrong:** `TsConfig::default()` is `Auto(true)`, so your es2024 fixture is already exercising the **`Raw`** variant — `Normal` was the untested one. Added `tsconfig: false` to cover it.

Also added `chrome133`, which is the case your own description leads with but that nothing tested. It is worth pinning separately: engine targets do not reduce to a single ES version, so they take a different path through `has_feature`, and it is the shape a real Vite/browserslist config produces. I confirmed it lowers correctly.

Both are genuine regression tests — with the pre-PR gate restored, all three fixtures emit raw `using`.

## Verification

- `cargo test --workspace`: 1952 passed. The 5 failures are environmental in a fresh worktree (4 need `node_modules`, 1 needs the `test262` submodule), not snapshot mismatches — so the widened list shifted no existing output.
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- --deny warnings`, `cargo check --workspace --all-targets` all clean.
- Only 4 in-repo fixtures set `transform.target` (es6/es2015/es2018/es2021), all already below the old gate, which is why nothing else needed a snapshot update.

## Two notes, neither blocking

**The description has an inverted sentence.** "Check whether the target **supports** Explicit Resource Management" — `has_feature` returns `true` when the feature is *not* supported (needs transformation); its own doc comment says so. The code is right, the prose reads backwards. Only you can edit the description, so flagging rather than fixing.

**An upstream gap worth knowing about:** `target: "safari18"` does *not* lower `using`. oxc's `ES2026ExplicitResourceManagement` entry lists only `{Chrome 134, Node 24, Firefox 141, Edge 134, Es 2026}` — no Safari, iOS, or Samsung — so `has_feature` falls through to the `Es` default and reports it supported. Safari did not ship ERM until 26.0. Out of scope here, since `EnvOptions` is built from the same call and the transformer would not lower it either way, but it means the fix does not reach those targets. Probably worth an oxc issue.

IWANABETHATGUY (approved):

sekyungk:
@IWANABETHATGUY

I appreciate your sincere comments and ongoing feedback.

I just revised the sentences. I apologize for not addressing the meaning clearly.
As this commit has significantly changed the entire implementation of this PR,  would it be better if I update the PR description to relate the final implement?
Regarding this project, I'm fully open to whatever you prefer. Just feel free to let me know.

I was also aware of the fragility of the feature enumeration — I went with minimal hardcoding because oxc has no API to ask whether a target needs any lowering.
Keeping the list but enforcing updates at compile time with env_lowers_anything - I think I should learn this.

On the Safari gap: I had run into the same thing while verifying the oxc compat data — the ES2026ExplicitResourceManagement entry has no Safari/iOS, and it's still missing on oxc's latest main.
I'll open a separate PR to oxc for it.

It took time as I'm not get used to express something in English.
Thank you.

sekyungk:
@IWANABETHATGUY  https://github.com/oxc-project/oxc/issues/25155 I just checked the PR you opened on oxc. Seems I was a bit late this time. Thanks a lot.

IWANABETHATGUY:
> @IWANABETHATGUY
> 
> I appreciate your sincere comments and ongoing feedback.
> 
> I just revised the sentences. I apologize for not addressing the meaning clearly. As this commit has significantly changed the entire implementation of this PR, would it be better if I update the PR description to relate the final implement? Regarding this project, I'm fully open to whatever you prefer. Just feel free to let me know.
> 
> I was also aware of the fragility of the feature enumeration — I went with minimal hardcoding because oxc has no API to ask whether a target needs any lowering. Keeping the list but enforcing updates at compile time with env_lowers_anything - I think I should learn this.
> 
> On the Safari gap: I had run into the same thing while verifying the oxc compat data — the ES2026ExplicitResourceManagement entry has no Safari/iOS, and it's still missing on oxc's latest main. I'll open a separate PR to oxc for it.
> 
> It took time as I'm not get used to express something in English. Thank you.

And don't read the extra commits as a knock on the PR. You found a real bug, diagnosed the cause correctly, and your instincts were right on both counts — that the enumeration was fragile, and that oxc's compat data was missing Safari. That's the hard part. Nice work.

PR #10567, opened 2026-07-31
fix: initialize re-exported wrapped ESM modules at the entry chunk

### LINKED ISSUE #10543: [Bug]: Dep optimizer: wrapped-ESM init_* never emitted when a facade entry only re-exports from a shared chunk (MUI StyledEngineProvider)

A remaining variant of #9502 (fixed by #9717 in 1.1.2). In Vite's dep optimizer, when an optimized-dep **entry** (here `@mui/material/styles`) only **re-exports** bindings from a shared split chunk, and the underlying module is wrapped as lazy ESM (`__esmMin`), the corresponding `init_*()` call is emitted **nowhere in the graph** — the shared chunk defines and exports `init_StyledEngineProvider`, but no chunk (including the facade entry) imports or calls it, so the wrapper that assigns `import_react = __toESM(require_react())` never runs.

Reproduced on rolldown 1.1.5 (vite 8.1.5) and rolldown 1.2.1 (vite 8.2.0-beta.0).

Minimized to three imports — removing any one of them makes the bug disappear:

```jsx
// src/main.jsx
import { StyledEngineProvider } from "@mui/material/styles";
import "@mui/icons-material/Stop";
import "@mui/lab";
import { createRoot } from "react-dom/client";

createRoot(document.getElementById("root")).render(
  <StyledEngineProvider injectFirst>
    <h1>it rendered</h1>
  </StyledEngineProvider>
);
```

```html
<!-- index.html -->
<!doctype html>
<html>
  <body>
    <div id="root"></div>
    <script type="module" src="/src/main.jsx"></script>
  </body>
</html>
```

```json
// package.json (dependency mix matters: @mui/material v6 + v5-era icons/lab,
// which nest their own @mui/material v5 copy — install with --legacy-peer-deps)
{
  "name": "rolldown-dep-optimizer-dangling-init-repro",
  "private": true,
  "type": "module",
  "dependencies": {
    "@emotion/react": "^11.7.1",
    "@emotion/styled": "^11.6.0",
    "@mui/icons-material": "^5.10.3",
    "@mui/lab": "^5.0.0-alpha.97",
    "@mui/material": "^6.4.0",
    "react": "^18.2.0",
    "react-dom": "^18.2.0"
  },
  "devDependencies": { "vite": "^8.1.5" }
}
```

Steps:

```bash
npm install --legacy-peer-deps
npx vite --force   # open the page → white screen + useMemo TypeError
# or headless:
npx vite optimize --force
grep -rl "init_StyledEngineProvider = __esmMin" node_modules/.vite/deps/   # defined
grep -rl "init_StyledEngineProvider()" node_modules/.vite/deps/            # never called → bug
```

A trivial project importing only `@mui/material/styles` + `@mui/material` does **not** reproduce — `StyledEngineProvider` is then emitted unwrapped. The icon + lab entries push the optimizer into the lazy-wrapping path, and the init call is then dropped.

**Workaround** — forcing the whole pre-bundle into a single chunk removes the cross-chunk boundary; the same repro then renders fine:

```js
// vite.config.js
export default {
  optimizeDeps: {
    rolldownOptions: {
      output: { codeSplitting: { groups: [{ name: "deps", test: /./ }] } },
    },
  },
};
```

### What is expected?

Every module wrapped as `__esmMin` has its `init_*()` invoked (directly or transitively) before its exported bindings are used. A facade entry that re-exports a wrapped module's bindings should import and call its init. The app renders "it rendered".

The dev server white-screens. The wrapper that assigns `import_react$1` never runs, so rendering `StyledEngineProvider` (imported from `@mui/material/styles`) crashes:

```
TypeError: Cannot read properties of undefined (reading 'useMemo')
    at StyledEngineProvider (node_modules/.vite/deps/styles-BcPZjlv2.js:34:31)
```

The generated chunks show the dangling init — defined and exported, but called nowhere:

```js
// node_modules/.vite/deps/styles-*.js (shared chunk)
var init_StyledEngineProvider = __esmMin((() => {
    import_react$1 = /* @__PURE__ */ __toESM(require_react());
    ...
}));
...
export { ..., StyledEngineProvider as v, init_StyledEngineProvider as y };
```

```bash
$ grep -rl "init_StyledEngineProvider = __esmMin" node_modules/.vite/deps/
node_modules/.vite/deps/styles-BcPZjlv2.js
$ grep -rl "init_StyledEngineProvider()" node_modules/.vite/deps/
(no matches)
```

The `@mui_material_styles.js` facade entry imports `StyledEngineProvider as v` but not `init_StyledEngineProvider as y`.

### System Info

```Shell
System:
  OS: macOS 26.5.2
  CPU: (12) arm64 Apple M4 Pro
  Memory: 199.53 MB / 24.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 22.22.2 - ~/.nvm/versions/node/v22.22.2/bin/node
  npm: 10.9.7 - ~/.nvm/versions/node/v22.22.2/bin/npm
Browsers:
  Chrome: 150.0.7871.187
  Safari: 26.5.2
npmPackages:
  vite: ^8.1.5 => 8.1.5 (rolldown 1.1.5; also reproduced with vite 8.2.0-beta.0 / rolldown 1.2.1)
```

### Any additional comments?

Hit this upgrading a large React dashboard (MUI 6 + some v5-era MUI packages) from Vite 7 to 8 — production build is fine, only the dev server breaks. Related earlier fixes in the same family: #9502 / #9717 (fixed in 1.1.2 — this variant survives it), #8950, #9515. Downstream symptom thread: vitejs/vite#22499 (the `useMemo` reports there match this).

Happy to test a nightly/patch build against the real-world app if useful.

hyfdev:
Fixes https://github.com/rolldown/rolldown/issues/10543.

The issue is that after wrapping, rolldown should output the entry chunk with a `init_xxx` call to ensure the execution of the wrapped module.

```js
// @mui_material_styles.js — before
import { …, v as StyledEngineProvider } from "./styles-BcPZjlv2.js";
export { StyledEngineProvider, … };   // nothing ever runs the module body that assigns `v`

// after
import { …, v as StyledEngineProvider, y as init_StyledEngineProvider } from "./styles-6TFHfymC.js";
init_StyledEngineProvider();          // evaluates the wrapped module first
export { StyledEngineProvider, … };
```

--edited

While this PR does fix the issue, I feel that the complexity around "wrappers" are increases significantly, we need to find a way to abstract them.

### DIFF
--- a/crates/rolldown/src/utils/chunk/render_chunk_exports.rs
+++ b/crates/rolldown/src/utils/chunk/render_chunk_exports.rs
@@ -13,7 +13,7 @@ use rolldown_utils::{
 };
 use rustc_hash::FxHashSet;
 
-use crate::esm_init_obligations::WrappedEsmInitTarget;
+use crate::esm_init_obligations::{WrappedEsmInitTarget, collect_entry_reexported_wrapper_inits};
 use crate::{stages::link_stage::LinkStageOutput, types::generator::GenerateContext};
 
 pub fn render_wrapped_entry_chunk(
@@ -83,6 +83,8 @@ pub fn render_wrapped_entry_chunk(
       return Some(rendered);
     }
 
+    let reexport_init_calls = render_entry_reexported_wrapper_init_calls(ctx, entry_id, entry_meta);
+
     match ctx.esm_init_target(entry_id) {
       Some(target) => {
         let wrapper_ref_name = ctx.finalized_string_pattern_for_symbol_ref(
@@ -90,17 +92,93 @@ pub fn render_wrapped_entry_chunk(
           ctx.chunk_idx,
           &ctx.chunk.canonical_names,
         );
-        if target.tla_tainted {
-          Some(concat_string!("await ", wrapper_ref_name, "();"))
+        let own_init_call = if target.tla_tainted {
+          concat_string!("await ", wrapper_ref_name, "();")
         } else {
-          Some(concat_string!(wrapper_ref_name, "();"))
+          concat_string!(wrapper_ref_name, "();")
+        };
+        match reexport_init_calls {
+          Some(mut calls) => {
+            calls.push_str(&own_init_call);
+            Some(calls)
+          }
+          None => Some(own_init_call),
         }
       }
-      None => None,
+      None => reexport_init_calls,
     }
   } else {
     None
   }
+}
+
+/// Off-strict, an entry can re-export bindings owned by an ESM-wrapped module hosted in another
+/// chunk while every statement that could call that module's `init_*` was tree-shaken (a pure
+/// re-export barrel chain resolves bindings symbol-to-symbol, keeping the barrels' forwarding
+/// statements excluded). Cross-chunk registration still imports each such wrapper — the non-strict
+/// arm of `add_depended_symbol_with_wrapped_esm_init` pairs every depended binding with its
+/// wrapper — but nothing ever calls it, so the wrapped module never evaluates and its bindings
+/// read as `undefined` at runtime (issue #10543).
+///
+/// The entry chunk owns these calls: it already imports every such wrapper (no new chunk-graph
+/// edges, hence no new cyclic-evaluation hazards for eager top-level init calls in shared
+/// chunks). ESM evaluates dependencies before the importer's body, so an entry hosting its own
+/// body gets the calls as a body prelude from the finalizer
+/// (`entry_reexported_wrapper_init_prelude`); this chunk-tail path serves the remaining shapes,
+/// where the tail still precedes the entry body: a facade chunk hosts no body at all, and a
+/// wrapped entry's body only runs inside the wrapper invoked right after these calls.
+fn render_entry_reexported_wrapper_init_calls(
+  ctx: &GenerateContext<'_>,
+  entry_id: ModuleIdx,
+  entry_meta: &crate::types::linking_metadata::LinkingMetadata,
+) -> Option<String> {
+  if ctx.options.is_strict_execution_order_enabled() {
+    // Strict execution order routes entry initialization through order-wrap lowering
+    // (consumer-local targets and entry facades) instead.
+    return None;
+  }
+  let entry_hosted_here = ctx.chunk_graph.module_to_chunk[entry_id] == Some(ctx.chunk_idx);
+  let entry_is_wrapped = ctx.esm_init_target(entry_id).is_some();
+  if entry_hosted_here && !entry_is_wrapped {
+    return None;
+  }
+  let inits = collect_entry_reexported_wrapper_inits(
+    entry_id,
+    entry_meta,
+    &ctx.link_output.metas,
+    &ctx.link_output.module_table.modules,
+    &ctx.link_output.symbol_db,
+    Some(&ctx.chunk.canonical_names),
+  );
+  let mut rendered = String::new();
+  for init in inits {
+    // An in-chunk wrapped entry was wrapped together with its whole static graph (wrap
+    // propagation), so a same-chunk owner's init is already covered inside the chunk — by a
+    // statement-position call or the excluded-statement metadata's same-chunk arm — and calling
+    // it again here would be a pure duplicate. A facade chunk gets no such guarantee for the
+    // modules it hosts, so only the in-chunk wrapped entry filters.
+    if entry_hosted_here && ctx.chunk_graph.module_to_chunk[init.owner] == Some(ctx.chunk_idx) {
+      continue;
+    }
+    let wrapper_name = ctx.finalized_string_pattern_for_symbol_ref(
+      init.wrapper_ref,
+      ctx.chunk_idx,
+      &ctx.chunk.canonical_names,
+    );
+    if init.tla_tainted {
+      // Defensive parity with `wrapped_esm_init_call_expr`'s `await_if_tla`; believed
+      // unreachable today. Non-`esm` formats reject top-level await at scan time
+      // (`AstScanner::handle_top_level_await`), so this cannot produce an `await` inside a
+      // plain iife/umd/cjs factory. Under `esm`, every off-strict `WrapKind::Esm` cause is
+      // blocked from combining with TLA: a `require` reaching a TLA subtree is a build error
+      // (`REQUIRE_TLA`), and the dynamic-import-with-splitting-disabled cause is single-chunk,
+      // where a statement-position or metadata init always covers the module first.
+      writeln!(rendered, "await {wrapper_name}();").expect("writing to a string cannot fail");
+    } else {
+      writeln!(rendered, "{wrapper_name}();").expect("writing to a string cannot fail");
+    }
+  }
+  (!rendered.is_empty()).then_some(rendered)
 }
 
 #[expect(clippy::too_many_lines)] // Dispatches over every output format and export mode inline.
--- a/crates/rolldown/src/esm_init_obligations.rs
+++ b/crates/rolldown/src/esm_init_obligations.rs
@@ -33,10 +33,12 @@
 //! Purpose contracts are deliberately *not* identical, and each divergence is encoded (and
 //! justified) on [`ObligationPurpose`] rather than re-derived at call sites.
 
+use oxc_str::CompactStr;
 use rolldown_common::{
-  ChunkIdx, ConstExportMeta, ExportsKind, ImportKind, ImportRecordIdx, ImportRecordMeta,
-  IndexModules, InlineConstMode, Module, ModuleIdx, NormalModule, ResolvedImportRecord, Specifier,
-  SymbolOrMemberExprRef, SymbolRef, SymbolRefDb, WrapKind,
+  ChunkIdx, ConcatenateWrappedModuleKind, ConstExportMeta, ExportsKind, ImportKind,
+  ImportRecordIdx, ImportRecordMeta, IndexModules, InlineConstMode, Module, ModuleIdx,
+  NormalModule, ResolvedImportRecord, Specifier, SymbolOrMemberExprRef, SymbolRef, SymbolRefDb,
+  WrapKind,
 };
 use rustc_hash::{FxHashMap, FxHashSet};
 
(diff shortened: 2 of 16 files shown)


shulaoda:
@codex review

crates/rolldown/src/utils/chunk/render_chunk_exports.rs
```diff
@@ -83,26 +85,112 @@ pub fn render_wrapped_entry_chunk(
...
+  entry_meta: &crate::types::linking_metadata::LinkingMetadata,
+) -> Option<String> {
+  if ctx.options.is_strict_execution_order_enabled() {
+    // Strict execution order routes entry initialization through order-wrap lowering
+    // (consumer-local targets and entry facades) instead.
+    return None;
+  }
+  let mut rendered = String::new();
+  let mut seen_wrappers = FxHashSet::default();
+  // Mirror the non-strict registration enumeration in `collect_depended_symbols`, so every call
```
IWANABETHATGUY: This comment is doing a lot of work. The two enumerations live in different files and different stages, with four matching conditions each, and if they ever drift the failure is silent: the `canonical_names` guard below turns "registration no longer adds this wrapper" into "no call emitted", which is #10543 again, with nothing red to show for it.

Could we extract the walk so there's only one copy? Something like `entry_reexported_esm_wrappers(link_output, entry_id) -> impl Iterator<Item = SymbolRef>` that both registration and this consume. Then the invariant is a fact about the code rather than a promise in a comment.

crates/rolldown/src/utils/chunk/render_chunk_exports.rs
```diff
@@ -83,26 +85,112 @@ pub fn render_wrapped_entry_chunk(
...
+  // rendered here is backed by a registered import (or a local declaration).
+  for resolved_export in entry_meta
+    .resolved_exports
+    .iter()
+    .sorted_unstable_by_key(|(name, _)| *name)
```
IWANABETHATGUY: This sorts the entry's whole export map on every entry chunk of every build, including the overwhelmingly common case where nothing qualifies and the loop emits nothing. Filtering first and sorting the (usually empty) survivors gives identical output — filtering preserves relative order, so filter-then-sort and sort-then-filter agree under a total key order — and turns O(exports · log exports) into O(qualifying · log qualifying).

CodSpeed says no change, but I doubt the benchmarks have a barrel entry with a few thousand exports, which is precisely the shape this bug shows up in (MUI). Cheap enough to just do.

crates/rolldown/src/utils/chunk/render_chunk_exports.rs
```diff
@@ -83,26 +85,112 @@ pub fn render_wrapped_entry_chunk(
...
+    }
+    if !seen_wrappers.insert(canonical_wrapper_ref) {
+      continue;
+    }
+    let wrapper_name = ctx.finalized_string_pattern_for_symbol_ref(
+      wrapper_ref,
+      ctx.chunk_idx,
+      &ctx.chunk.canonical_names,
+    );
+    if owner_meta.is_tla_or_contains_tla_dependency {
```
IWANABETHATGUY: This branch has no test, and I'm not sure it's safe in every format `render_wrapped_entry_chunk` serves. `iife.rs` and `umd.rs` build a plain (non-async) factory, so a top-level `await` there is a `SyntaxError`, and I couldn't find any format/TLA validation in the crate that would head it off.

It's not new — the existing `own_init_call` just above has the same shape — but this widens it to entries with no own init target, and off-strict `WrapKind::Esm` + TLA does look reachable: `determine_module_exports_kind.rs:65-72` wraps a dynamically-imported ESM module as `Esm` when code splitting is disabled, which is exactly iife/umd/cjs.

Either a cell on `wrapped_esm_init_interop/excluded_reexport_tla` that exercises it, or a comment explaining why the combination can't arise. Right now a reader can't tell whether this line is load-bearing or dead.

IWANABETHATGUY (commented):
Nice writeup, and the "no new chunk-graph edges" argument holds up — I traced the enumeration against `collect_depended_symbols` (`compute_cross_chunk_links.rs:636-651`) and the non-strict arm of `add_depended_symbol_with_wrapped_esm_init` (`:758-769`), and every call this renders is backed by an import that already exists on main. Also confirmed the fixture red→green locally (`AssertionError: undefined == 'v18'` without the change), and the full integration suite comes back with zero snapshot churn beyond the three you committed.

One thing I'd want before merge: `internal-docs/code-splitting/design.md` needs updating in this same change. The "Trigger placement" table opens with *"Every site that can run a wrapped module, in one place"*, and this adds one that isn't listed — entry-chunk tail, off-strict, for re-exported wrapped modules. A new row would do it.

While you're in there, the rejected alternative from your PR description deserves to land in that doc too. "Let the metadata collector keep cross-chunk targets" is the obvious first thing anyone will try — I'd have guessed the same, on the reasoning that the resulting chunk cycle is harmless because calls only happen inside wrappers post-evaluation. Your finding that off-strict output is full of *eager* top-level init calls, so the cycle inverts evaluation order and you get `init_identifier is not a function`, is exactly the kind of thing that's expensive to rediscover. Right now it only exists in a PR description.

On the ordering point Codex raised: I couldn't make it bite. Three attempts — entry body reading the re-exported binding, namespace access, and entry body observing a global the wrapped module sets — all self-heal, because referencing the module (or the module having any side effect at all) re-includes the forwarding chain that carries `init_w()`. The shape only exists when the wrapped module is side-effect-free *and* unreferenced by the entry body, and referencing it is precisely what fixes it. Not a regression either way, since main never calls it at all. Might still be worth a sentence in the doc comment on why the tail rather than the prologue — prologue is more ESM-faithful, but runs a cross-chunk wrapper body before this chunk's own module bodies, which is unsafe if the two chunks form a cycle. That's the non-obvious half of the decision and it's currently unrecorded.

crates/rolldown/tests/esbuild/splitting/edge_case_issue2793_without_splitting/artifacts.snap
```diff
@@ -29,6 +29,8 @@ var init_edge_case_issue2793_without_splitting = __esmMin((() => {
 	init_b();
 }));
 //#endregion
+init_a();
```
IWANABETHATGUY: All three snapshot deltas in this PR are same-chunk owners that were already initialized in-chunk, so these are pure duplicates. They're inert — `__esmMin` is once-guarded, and duplicate init calls already occur on main (`basic_wrapped_esm` shipped two `init_foo()` before this PR) — but they're bytes in every entry chunk that re-exports from a wrapped module, which isn't rare.

Filtering to `module_to_chunk[owner] != Some(chunk_idx)` would erase all three deltas and still fix #10543. I'm assuming you kept same-chunk deliberately, since an unwrapped barrel gets no metadata at all and a fully tree-shaken same-chunk chain would otherwise stay uncovered — if that's the reason, could it go in the doc comment on `render_entry_reexported_wrapper_init_calls`? As it stands the snapshot noise reads as accidental rather than as the point.

crates/rolldown/src/utils/chunk/render_chunk_exports.rs
```diff
@@ -83,26 +85,112 @@ pub fn render_wrapped_entry_chunk(
       return Some(rendered);
     }
 
+    let reexport_init_calls = render_entry_reexported_wrapper_init_calls(ctx, entry_id, entry_meta);
```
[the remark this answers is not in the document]
hyfdev: Fixed in ae687e7c8. The calls now precede the entry module's body: an in-chunk unwrapped entry gets them as a finalizer-emitted body prelude at the after-imports position (deduplicated against statement-position inits, ordered by module execution order); the chunk tail only remains for the two shapes where it already precedes the entry body — a facade chunk hosting no body, and a wrapped entry whose body runs inside the wrapper invoked after these calls. The fixture now pins the ordering with an entry-body probe observing the wrapped module's side effect — red with the tail placement, green with the prelude.

crates/rolldown/src/utils/chunk/render_chunk_exports.rs
hyfdev: Done in 764324802 — the walk now lives once as `collect_entry_reexported_wrapper_inits` in `esm_init_obligations.rs`. Registration consumes it with `canonical_names: None` (it runs before chunk names exist and is what makes a wrapper reachable in the first place); both emission sites consume it with the chunk's assigned names. Registration-covers-emission is now a property of the shared function rather than a promise in a comment.

crates/rolldown/src/utils/chunk/render_chunk_exports.rs
hyfdev: Done in 764324802 — one filtering pass over the export map, then a name sort of only the qualifying survivors (kept so the first-export-wins dedup stays deterministic under hash-order iteration), then the execution-order sort of the deduped wrappers.

crates/rolldown/src/utils/chunk/render_chunk_exports.rs
hyfdev: I tried to build that cell and could not reach the branch — the combination is blocked upstream twice. Non-esm formats reject top-level await at scan time (`AstScanner::handle_top_level_await`), so no `await` can land in an iife/umd factory. Under `esm`, adding TLA to the regression fixture's subtree turns the wrap-causing `require` into a `REQUIRE_TLA` build error (`This require call is not allowed because the transitive dependency "w.js" contains a top-level await`), and the other off-strict wrap cause you cite — dynamic import with splitting disabled — is single-chunk, where a statement-position or metadata init always covers the module before the entry emission would.

764324802 documents this at the `await` arm and on the `tla_tainted` field. The branch stays for parity with `wrapped_esm_init_call_expr`'s `await_if_tla` (which the body-prelude path shares, and whose await is load-bearing — `excluded_reexport_tla`).

crates/rolldown/tests/esbuild/splitting/edge_case_issue2793_without_splitting/artifacts.snap
hyfdev: Same-chunk was indeed deliberate for the shapes with no other call site — that rationale now lives on the collector's doc comment. And since your review the duplicates themselves are gone: the body-prelude commit's dedup removed the `basic_wrapped_esm` one, `ambiguous_star_reexport`'s call is now the correctly-placed before-body init of a cross-chunk owner, and 764324802 filters same-chunk owners at the tail for in-chunk wrapped entries (wrap propagation wrapped their whole static graph, so those are always covered in-chunk) — which restores this snapshot to main byte-for-byte.

IWANABETHATGUY (approved):

hyfdev:
### Merge activity

* **Aug 3, 5:44 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Aug 3, 5:44 AM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Aug 3, 5:50 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10577, opened 2026-08-01
docs: correct stale `buildStart` comment in `ScanStageCache::merge`

shulaoda:
The comment claimed partial scans skip `buildStart`. That is only true for HMR patch scans, which bypass `ScanStage` entirely and cannot re-discover emitted entries at all. `ScanStage` partial scans call `build_start` unconditionally, and the fallback is still needed there because the emitting hook may not re-run.


IWANABETHATGUY (approved):

crates/rolldown/src/types/scan_stage_cache.rs
```diff
@@ -275,9 +275,10 @@ impl ScanStageCache {
         user_defined_entry_modules.insert(idx);
       }
     }
-    // Entries emitted via `emitFile(type: 'chunk')` are only discovered by
-    // full scans (partial scans skip `buildStart`); their rows persist in
-    // `entry_points`, so keep their modules flagged as entries too.
+    // Entries emitted via `emitFile(type: 'chunk')` may not be re-discovered
+    // by a partial scan. HMR patch scans skip `buildStart` entirely, and in
+    // `ScanStage` partial scans the emitting hook may not re-run. Their rows
+    // persist in `entry_points`, so keep their modules flagged as entries too.
```
IWANABETHATGUY:
```suggestion
    // Entries emitted via `emitFile(type: 'chunk')` may not be re-discovered by a
    // partial scan. The HMR stage's partial scans bypass `ScanStage`, so `buildStart`
    // never runs there. `ScanStage` partial scans do run it, but a plugin may emit
    // from `resolveId`/`load`/`transform`/`moduleParsed`, and those only re-run for
    // re-scanned modules. Their rows persist in `entry_points`, so keep their
    // modules flagged as entries too.
```
`compile_lazy_entry` also calls `merge` without going through `ScanStage`, so it's not only patch scans. I think naming the hooks is more useful than "may not re-run" here.

h-a-n-a (approved):

PR #10579, opened 2026-08-02
fix(dev): force `cleanDir` off in dev mode

shulaoda:
Dev mode should force `cleanDir` option off


crates/rolldown/src/utils/prepare_build_context.rs
```diff
@@ -280,6 +282,7 @@ pub fn prepare_build_context(
     // treeshaking, so it must be disabled as well.
     raw_treeshake = TreeshakeOptions::Boolean(false);
     experimental.lazy_barrel = Some(false);
+    clean_dir = false;
```
h-a-n-a, line 287:
```suggestion
    // Dev rebuilds write only the changed chunks, so cleaning the output
    // directory would delete chunks the browser can still ask for.
    clean_dir = false;
```

h-a-n-a (approved):

PR #9969, opened 2026-06-25
test(dev): reject unknown lazy compile modules

sapphi-red:
This PR adds a test to ensure lazy compile requests to modules that are not found to be rejected. This is to avoid a malicious actor to send a lazy compile request to a module that is not included in the module graph and receive a chunk generated from it, potentially containing sensitive values. This is similar to what [Vite's `server.fs.strict` tries to block](https://v8.vite.dev/config/server-options#server-fs-strict).

PR #10580, opened 2026-08-02
fix(dev): keep `?rolldown-lazy` proxy ids out of user `resolveId` hooks

btea:
## Summary

Fixes vitejs/vite#23124 — re-resolving a lazy proxy id (`?rolldown-lazy=1`) no longer leaks to user `resolveId` hooks, and known proxy ids are claimed by the lazy compilation plugin itself.

## Why

A plugin virtual module reached through a dynamic import becomes a lazy entry whose proxy id carries the internal `?rolldown-lazy=1` query. When that stub id is re-resolved (e.g. entry resolution for a lazy compilation request), virtual-module plugins no longer recognize their own id with the query appended, and nothing else can resolve a virtual id → `UNRESOLVED_ENTRY` on first page load.

#10426 covered `load` / `transform` / `transform_ast` / `module_parsed` but not `resolve_id`; Vite's own workaround (#22778) likewise only wraps `transform`.

## What changed

- `LazyCompilationPlugin::resolve_id`: claim proxy ids present in `lazy_entries` for any import kind. Unknown proxy ids still fall through and stay unresolvable (cf. the #9969 gate).
- `PluginDriver`: extend the `should_skip_plugin_for_lazy_proxy` skip to `resolve_id`, so user plugins never see the internal query; the lazy plugin itself still runs to claim the id. Matching is on the specifier only — the fetched proxy's `import($MODULE_ID)` (importer = proxy id) still resolves the real module through user plugins normally.
- Regression test in `packages/rolldown/tests/dev/dev-lazy-compile.test.ts`: a virtual module behind a lazy proxy — asserts user `resolveId` hooks never see the proxy id, the proxy id re-resolves to itself, and the real module still compiles into the lazy chunk.
- Update `internal-docs/lazy-compilation/implementation.md` with the new `resolve_id` contract.

## Validation

- `pnpm --filter rolldown run build-native:debug`
- `pnpm --filter rolldown-tests exec vitest run dev/dev-lazy-compile.test.ts` (8/8)
- Negative control: with the Rust changes stashed and a full rebuild, the new test fails (`this.resolve(proxyId)` returns `null`)
- E2E with the issue's repro (vite-bundleddev-virtual-lazy-repro, local vite + this rolldown via pnpm overrides):
  - Baseline (npm rolldown 1.2.1): server logs `[UNRESOLVED_ENTRY] Cannot resolve entry module \0virtual:lazy-me?rolldown-lazy=1`, page shows `ERR Failed to fetch dynamically imported module`
  - This change: first lazy compile succeeds, page renders `hello from a virtual module`, zero server/console errors

Note: the repro's plugin must also recognize its own `\0`-prefixed id in `resolveId` (the canonical virtual-module pattern, which real-world plugins like @embroider/vite follow). Without it, the fetched proxy's import of the *real* module id can't resolve — a separate behavior difference from classic Vite dev (rolldown re-resolves the real id), better handled as a follow-up.

## AI disclosure

This PR was developed with AI assistance. All generated changes were reviewed, tested, and adjusted before submission.

### DIFF
--- a/crates/rolldown_plugin/src/plugin_driver/build_hooks.rs
+++ b/crates/rolldown_plugin/src/plugin_driver/build_hooks.rs
@@ -99,6 +99,14 @@ impl PluginDriver {
       if skipped_plugins.has_bit(plugin_idx) {
         continue;
       }
+      if should_skip_plugin_for_lazy_proxy(
+        self.should_skip_user_plugins_for_lazy_proxy_modules,
+        args.specifier,
+        plugin_idx,
+        self.lazy_compilation_plugin_idx,
+      ) {
+        continue;
+      }
       let ret = async {
         trace_action!(action::HookResolveIdCallStart {
           action: "HookResolveIdCallStart",
--- a/crates/rolldown_plugin_lazy_compilation/src/lazy_compilation_plugin.rs
+++ b/crates/rolldown_plugin_lazy_compilation/src/lazy_compilation_plugin.rs
@@ -79,6 +79,20 @@ impl Plugin for LazyCompilationPlugin {
     ctx: &rolldown_plugin::PluginContext,
     args: &rolldown_plugin::HookResolveIdArgs<'_>,
   ) -> rolldown_plugin::HookResolveIdReturn {
+    // Re-resolution of a known proxy id (e.g. the dev server resolving the stub id to
+    // serve a lazy compilation request) is claimed here; unknown proxy ids fall through
+    // and stay unresolvable (cf. the unknown-lazy-id rejection in
+    // `HmrStage::compile_lazy_entry`).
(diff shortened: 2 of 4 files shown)


h-a-n-a (approved):

hyfdev:
@sapphi-red cc

This is not a blocker review comment.

Similar to runtime module. This is adding some privileges to some internal modules. I wonder if there is better way to do this.

One key thing is that I think the current plugin system is only designed for normal users, while we need something that's for internal and can hide from users, so we don't need to expose something internal to outside users.

sapphi-red:
Ah, maybe we can have `isInternal` in the ModuleInfo 🤔

hyfdev:
> Ah, maybe we can have `isInternal` in the ModuleInfo 🤔

I wrote some thoughts at https://github.com/rolldown/backlog/issues/1

Not urgent, but we could use the place to discuss as a start.

PR #10594, opened 2026-08-03
feat(code-splitting): widen namespace extraction for already-loaded merges

IWANABETHATGUY:
`dynamic_entry_supports_namespace_extraction` refused three cases it does not need to, each costing an extra chunk:

- `preserveModules`. This is dead code, since `generate_chunks` never calls `split_chunks` there.
- `!minifyInternalExports`. This is the default for non-ESM output without minify. So a cjs, iife or umd build lost namespace extraction unless it was minified.
- a declaring symbol named `then`. Only the export name can make a namespace thenable, and `const then = 1; export { then as ready }` puts `ready` in the namespace.

refs #10263

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/dynamic_already_loaded.rs
+++ b/crates/rolldown/src/stages/generate_stage/dynamic_already_loaded.rs
@@ -212,19 +212,23 @@ impl GenerateStage<'_> {
   /// entry's namespace object from its chunk and rewriting every dynamic importer to
   /// `.then((n) => n.<ns>)` (see `rewrite_dynamic_import_for_merged_entry`).
   fn dynamic_entry_supports_namespace_extraction(&self, entry_module_idx: ModuleIdx) -> bool {
-    if self.options.preserve_modules || !self.options.minify_internal_exports {
-      return false;
-    }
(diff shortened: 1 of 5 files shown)


hyfdev (approved):
Why this is a `perf`?

IWANABETHATGUY:
> Why this is a `perf`?

**Perf** the effect of chunk optimization. Using the test snapshot as an example, the number of chunks is reduced from 4 to 3. I guess both `feat` and `perf` make sense?

hyfdev:
> > Why this is a `perf`?
> 
> **Perf** is the chunk optimization. Using the test snapshot as an example, the number of chunks is reduced from 4 to 3. I guess both `feat` and `perf` make sense?

I think `feat` is better. I expect `perf` or `refactor` usually won't change bundler's behavior but improving the performance or code quality.

IWANABETHATGUY:
> > > Why this is a `perf`?
> > 
> > 
> > **Perf** is the chunk optimization. Using the test snapshot as an example, the number of chunks is reduced from 4 to 3. I guess both `feat` and `perf` make sense?
> 
> I think `feat` is better. I expect `perf` or `refactor` usually won't change bundler's behavior but improving the performance or code quality.

updated

PR #7164, opened 2025-11-20
fix: data race in `FileEmitter::emit_file` causing duplicate asset emissions

shulaoda:
closes #7163

PR #10605, opened 2026-08-03
fix(file-emitter): derive asset reference ids from content, not emit order

jjb:
## What is this PR solving?

Emitted assets get a different reference id on every build of identical input, which makes builds non-reproducible.

`FileEmitter::assign_reference_id` hashes the name it is given, but falls back to hashing a shared `AtomicUsize` counter when passed `None`:

```rust
pub fn assign_reference_id(&self, filename: Option<ArcStr>) -> ArcStr {
    xxhash_base64_url(
      filename
        .unwrap_or_else(|| {
          self.base_reference_id.fetch_add(1, Ordering::Relaxed).to_string().into()
        })
        .as_bytes(),
    )
```

The asset-dedup branch of `emit_file` passes `None`, so any asset emitted without an explicit `fileName` receives `xxhash(counter++)`. Assets are emitted concurrently while modules are processed, so which counter value each asset receives depends on the order the `emit_file` calls happen to arrive. Two builds of byte-identical input therefore emit **the same set of reference ids permuted across different assets**.

This is observable output rather than an internal detail, because an asset plugin writes the id into the module's own source text — `__ROLLDOWN_ASSET__#<referenceId>` in `rolldown_plugin_asset_module`, `__VITE_ASSET__<referenceId>__` in Vite — and that text is what lands in the sourcemap's `sourcesContent`. `.js.map` files therefore differ between builds of the same commit even when every `.js` file is byte-identical. It surfaced for us as a CI job that diffs the built frontend between two images and requires them to match: it fails stochastically, with five or six sourcemaps differing and zero JS files differing. Every one of the 14 differing reference ids in that build was exactly `xxhash_base64_url(String(k))` for k = 2..13.

## Reproduction (no Vite involved)

`npm i rolldown@1.2.2 && node repro.mjs`

```js
import { rolldown } from 'rolldown'

const N = 12
const build = async () => {
  const ids = {}
  const plugin = {
    name: 'emit-assets',
    resolveId(id) { if (id === 'entry' || id.startsWith('asset-')) return '\0' + id },
    load(id) {
      if (id === '\0entry')
        return Array.from({ length: N }, (_, i) => `import a${i} from 'asset-${i}'`).join('\n') +
          `\nexport default [${Array.from({ length: N }, (_, i) => `a${i}`)}]`
      if (id.startsWith('\0asset-')) {
        const n = Number(id.slice('\0asset-'.length))
        // exactly what Vite's asset plugin does: emit with `name`, no `fileName`
        ids[`asset-${n}`] = this.emitFile({ type: 'asset', name: `asset-${n}.txt`, source: `payload ${n}` })
        return 'export default "x"'
      }
    },
  }
  const bundle = await rolldown({ input: 'entry', plugins: [plugin] })
  await bundle.generate({ format: 'es' })
  await bundle.close()
  return JSON.stringify(Object.entries(ids).sort())
}

const seen = new Set()
for (let i = 0; i < 8; i++) seen.add(await build())
console.log(`${seen.size} distinct asset -> referenceId mappings across 8 builds`)
```

`rolldown@1.2.2` prints **8**. The same script against `rollup` prints **1** — Rollup derives reference ids from the emitted file's identity (`fileName || name || type`) and re-hashes only to break collisions. Passing an explicit `fileName` to `emitFile` makes rolldown stable too, which isolates the failing branch.

## History

Not a recent regression. da183aab ("feat: support PluginContext emit asset", #1349, 2024-06-05) added the counter fallback, but wrote `fetch_add(0, ...)`, so the counter never advanced and *every* asset emitted without a `fileName` received `xxhash("0")` — deterministic, but all such assets collided on one id. cfc16e80 ("feat: support emit binary asset", #1359, 2024-06-06) changed it to `fetch_add(1, ...)`, fixing the collision and introducing the order dependence in the same line. It has been latent ever since because nothing surfaced the id to users.

An earlier restructuring, d18df9b7 ("fix: data race in `FileEmitter::emit_file` causing duplicate asset emissions", #7164), is where the literal `assign_reference_id(None)` call appears, but it did not change behaviour: before it the same call passed `file.file_name.clone()`, which was still `None` at that point because `generate_file_name` runs afterwards.

## The fix

Use the source hash that `emit_file` already computes a few lines above as the id base, instead of the counter.

Dedup semantics are unchanged: assets with identical content already share one reference id via the `Occupied` branch, and a content-derived id is equal exactly when the content is equal. The id stays 22 characters, so the `REFERENCE_ID_LEN` invariant that `import.meta.ROLLDOWN_FILE_URL_<referenceId>_<urlId>` parsing depends on still holds.

Content-derived ids begin with a digit about ten times in sixty-four, where counter-derived ids reached that only after four unnamed assets. #10623 establishes that this is fine and adjusts the fixtures that assumed otherwise, which is why this PR is stacked on it.

## Tests

`asset_reference_ids_do_not_depend_on_emit_order` emits the same assets forward and reversed and asserts the assignment matches. Without the fix it fails, showing the permutation directly:

```
  left: [(0, "VRAku6fjghkApIISiBWPzg"), (1, "WPGYjwIlzWVNM1bYhOc83w"), ... (7, "$oUiuAXR7gZMKRQU56HW2w")]
 right: [(0, "$oUiuAXR7gZMKRQU56HW2w"), (1, "DUq8YoikZLXEejpNZ1KQvg"), ... (7, "VRAku6fjghkApIISiBWPzg")]
```

The asset snapshot moves with the change. Only the intermediate `__ROLLDOWN_ASSET__#` placeholder differs; the emitted filename on that same line is untouched, since asset filenames were already derived from the source hash.

## Parts that may need reviewer attention

- **No collision handling.** `assign_reference_id` does not check whether an id is already taken, for any of its callers; Rollup loops until the id is unique. A content-derived id could in principle collide with a filename-derived one. This is pre-existing rather than introduced here, but the fix does add a second hash domain, so you may want collision resolution added.
- **`emit_chunk` has the same latent issue.** It passes `chunk.name`, which is `Option`, so unnamed emitted chunks still take the counter path and remain order-dependent. I left that out of scope since I have no reproduction for it; happy to extend the PR if you would like it covered.
- **Chosen alternative.** Hashing the sanitized name instead of the content would also be stable, but it would break the dedup invariant, since two assets with identical content and different names must keep sharing one reference id.

<sub>Disclosure: this patch, the diagnosis and the reproduction were produced with Claude Code, then verified by running the tests and the reproduction described above.</sub>


jjb:
@sapphi-red here's the PR that surfaced the problem in #10623

jjb:
@sapphi-red any thoughts on this one?

shulaoda:
> @sapphi-red any thoughts on this one?

Could you resolve the conflicts? I will take a look before next release.

PR #10618, opened 2026-08-04
perf(chunk-naming): build minified export names without allocating

IWANABETHATGUY:
`generate_minified_names` allocated a `Vec<u8>` per candidate name, but `CompactStr` inlines anything up to 24 bytes and these names are at most 6. Built in a stack array instead.


IWANABETHATGUY:
@Dunqing, Could oxc expose this function from `oxc_mangler`? That way, we could share the same code instead of duplicating it.

hyfdev (approved):

PR #10623, opened 2026-08-04
test(resolve-file-url): allow a reference id to start with a digit

jjb:
## What is this PR solving?

The seven `plugin/resolve-file-url` fixtures assert that the reference id passed to the hook is usable as a standalone JS identifier, but nothing in rolldown actually requires that. This is already reachable on `main`. Counter-derived ids are `xxhash` of an incrementing integer, and the fourth is `8HSUfh7cJHNFxCkInNzNXw`, so a build only has to emit four unnamed assets. The fixtures do not catch it because each emits a single asset and gets `xxhash("0")`, which happens to start with a letter — so today they pass by luck of the alphabet rather than by design. Reference ids are hashes, so roughly ten in sixty-four begin with a digit, and such an id is perfectly usable:

- The id is only ever interpolated **after a prefix** — `import.meta.ROLLDOWN_FILE_URL_<referenceId>`. The property name therefore starts with `R`, and digits are legal everywhere after the first character of an `IdentifierName`. The documented plugin pattern in `docs/apis/plugin-api/file-urls.md` interpolates it the same way.
- `strip_file_url_prefix` splits the optional `_<urlId>` suffix off by `REFERENCE_ID_LEN` rather than scanning for a separator, precisely because an id may contain `_` and `$`. A digit changes nothing there.
- The two plugins that put an id into generated code, `rolldown_plugin_asset_module` and `rolldown_plugin_copy_module`, embed it inside a string literal.
- Rollup, whose `ROLLUP_FILE_URL_` form is supported as an alias, generates ids with `getHash64(referenceId).slice(0, 8).replaceAll('-', '$')` — the same `-` to `$` remap, and likewise no digit guard.

## Where the assertion came from

Nothing suggests the stricter pattern was a deliberate guarantee. It arrived in #10291. It seems like the regex was made to match existing fixtures, and maybe @sapphi-red assumed that portion needed to be a valid JS identifier.

Two later PRs, #10296 and #10222, then copied the line into more fixtures as they added cases.

## What changed

- **The pattern is stated once.** `REFERENCE_ID_REGEX` is exported from `rolldown-tests/utils` and used by all seven fixtures, instead of the literal being copied into each.
- **The contract is covered on ids the emitter really produces.** `emitted_reference_ids_have_the_expected_format` emits a batch through `emit_file` and asserts both halves: every id contains only identifier-legal characters, and at least one begins with a digit.
- **One comment is corrected.** `assign_reference_id` said the id "needs to be a valid identifier", which justifies remapping `-` but reads as forbidding a leading digit.

## This was discovered while working on #10605

#10605 makes asset reference ids content-derived so builds are reproducible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2683,6 +2683,7 @@ dependencies = [
  "oxc_resolver",
  "oxc_sourcemap",
  "oxc_str",
+ "regex",
  "rolldown_ecmascript",
  "rolldown_error",
  "rolldown_fs",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2683,7 +2683,6 @@ dependencies = [
  "oxc_resolver",
  "oxc_sourcemap",
  "oxc_str",
- "regex",
  "rolldown_ecmascript",
  "rolldown_error",
  "rolldown_fs",
--- a/crates/rolldown_common/Cargo.toml
+++ b/crates/rolldown_common/Cargo.toml
@@ -47,6 +47,7 @@ num-bigint = { workspace = true }
 
 [dev-dependencies]
 insta = { workspace = true }
+regex = { workspace = true }
 
 [features]
 default = ["serde"]
--- a/crates/rolldown_common/Cargo.toml
+++ b/crates/rolldown_common/Cargo.toml
@@ -47,7 +47,6 @@ num-bigint = { workspace = true }
 
 [dev-dependencies]
 insta = { workspace = true }
-regex = { workspace = true }
 
 [features]
 default = ["serde"]
--- a/crates/rolldown_common/src/file_emitter.rs
+++ b/crates/rolldown_common/src/file_emitter.rs
@@ -462,7 +462,6 @@ pub type SharedFileEmitter = Arc<FileEmitter>;
 #[cfg(test)]
 mod tests {
   use super::*;
-  use regex::Regex;
 
   /// Reference ids are the base64url encoding of a 128-bit xxhash (with `-` remapped to `$`),
   /// which is always 22 characters. The `import.meta.ROLLDOWN_FILE_URL_<referenceId>_<urlId>`
--- a/crates/rolldown_common/src/file_emitter.rs
+++ b/crates/rolldown_common/src/file_emitter.rs
@@ -486,46 +486,4 @@ mod tests {
       assert_eq!(emitter.assign_reference_id(Some(ArcStr::from(name))).len(), 22, "name={name:?}");
     }
   }
-
-  /// Generates a batch of ids and checks the format. Reference ids must contain only characters allowed
-  /// in JS identifiers, but they do not need to conform to the form of a JS identifier (e.g. leading number is allowed)
-  #[test]
-  fn emitted_reference_ids_have_the_expected_format() {
-    let emitter = FileEmitter::new(Arc::new(NormalizedBundlerOptions::default()));
-
-    let reference_ids = (0..64)
-      .map(|n| {
-        let name = format!("asset-{n}.txt");
-        emitter
-          .emit_file(
-            EmittedAsset {
-              name: Some(name.clone()),
-              source: StrOrBytes::Str(format!("payload for asset {n}")),
-              ..Default::default()
-            },
-            Some(FilenameTemplate::new(
-              "assets/[name]-[hash]".to_string(),
-              "output.assetFileNames",
-            )),
-            Some(ArcStr::from(name)),
-          )
-          .unwrap()
-      })
-      .collect::<Vec<_>>();
-
-    assert!(
-      reference_ids.iter().any(|id| id.starts_with(|c: char| c.is_ascii_digit())),
-      "expected at least one reference id to start with a digit, got {reference_ids:?}"
-    );
-
-    // equivalent of `REFERENCE_ID_REGEX` in `packages/rolldown/tests/src/utils.ts`
-    // `\w` in JS is ASCII-only, in rust it is unicode, so we write it out here
-    let reference_id_regex = Regex::new(r"^[A-Za-z0-9_$]+$").unwrap();
-    for id in &reference_ids {
-      assert!(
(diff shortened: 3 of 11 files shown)


jjb:
ping @sapphi-red

crates/rolldown_common/src/file_emitter.rs
```diff
@@ -485,4 +486,52 @@ mod tests {
...
+
+    assert!(
+      reference_ids.iter().any(|id| id.starts_with(|c: char| c.is_ascii_digit())),
+      "expected at least one reference id to start with a digit, got {reference_ids:?}"
+    );
+
+    // `REFERENCE_ID_REGEX` in `packages/rolldown/tests/src/utils.ts` states the same
+    // contract as `/^[$\w]+$/`. Written out here rather than as `\w`, which this crate
+    // builds Unicode-aware via `unicode-perl` and would also accept letters and digits
+    // from other scripts, unlike JavaScript's ASCII-only `\w`.
```
jjb:
```suggestion
    // equivalent of `REFERENCE_ID_REGEX` in `packages/rolldown/tests/src/utils.ts`
    // `\w` in JS is ASCII-only, in rust it is unicode, so we write it out here
```

crates/rolldown_common/src/file_emitter.rs
```diff
@@ -485,4 +486,48 @@ mod tests {
       assert_eq!(emitter.assign_reference_id(Some(ArcStr::from(name))).len(), 22, "name={name:?}");
     }
   }
+
+  /// Checks the format of a batch of real ids. Reference ids need only contain characters
+  /// an identifier can contain, since they are always interpolated after a prefix that
+  /// supplies the first one, so some of them start with a digit and nothing may assume
+  /// otherwise.
```
jjb:
```suggestion
  /// Checks the format of a batch of real ids. Reference ids must contain only characters allowed
  // in JS identifiers. They do not need to conform to the form of a JS identifier. (e.g. leading number is allowed)
```

crates/rolldown_common/src/file_emitter.rs
```diff
@@ -485,4 +486,46 @@ mod tests {
       assert_eq!(emitter.assign_reference_id(Some(ArcStr::from(name))).len(), 22, "name={name:?}");
     }
   }
+
+  /// Checks the format of a batch of real ids. Reference ids must contain only characters allowed
```
jjb:
```suggestion
  // Checks the format of a batch of real ids. Reference ids must contain only characters allowed
```

crates/rolldown_common/src/file_emitter.rs
```diff
@@ -485,4 +486,46 @@ mod tests {
       assert_eq!(emitter.assign_reference_id(Some(ArcStr::from(name))).len(), 22, "name={name:?}");
     }
   }
+
+  // Generates a batch of ids and checks the format. Reference ids must contain only characters allowed
+  // in JS identifiers, but they do not need to conform to the form of a JS identifier (e.g. leading number is allowed)
```
jjb:
```suggestion
  /// Generates a batch of ids and checks the format. Reference ids must contain only characters allowed
  /// in JS identifiers, but they do not need to conform to the form of a JS identifier (e.g. leading number is allowed)
```

crates/rolldown_common/src/file_emitter.rs
```diff
@@ -485,4 +486,46 @@ mod tests {
...
+    // equivalent of `REFERENCE_ID_REGEX` in `packages/rolldown/tests/src/utils.ts`
+    // `\w` in JS is ASCII-only, in rust it is unicode, so we write it out here
+    let reference_id_regex = Regex::new(r"^[A-Za-z0-9_$]+$").unwrap();
+    for id in &reference_ids {
+      assert!(
+        reference_id_regex.is_match(id),
+        "reference id does not match {reference_id_regex}: {id}"
+      );
+    }
+  }
```
sapphi-red:
```suggestion
```
I think we cannot have this because all the generated reference id may not start with a digit in the future.

packages/rolldown/tests/src/utils.ts
```diff
@@ -8,6 +8,18 @@ import type {
...
 
+/**
+ * @description
+ * Matches a reference id returned by `PluginContext.emitFile`.
+ *
+ * A reference id is only guaranteed to be made of characters that an identifier can
+ * contain. It is *not* guaranteed to be usable as a standalone identifier: ids are hashes,
+ * so they may begin with a digit. That is fine, because an id is always interpolated after
+ * a prefix which supplies the identifier's first character, as in
+ * `import.meta.ROLLDOWN_FILE_URL_<referenceId>`.
```
sapphi-red:
```suggestion
 * A reference id is guaranteed to be made of characters that an identifier can
 * contain.
```
I think we don't need to clarify about the first character. I simply overlooked the code in the review.

crates/rolldown_common/src/file_emitter.rs
```diff
@@ -462,6 +462,7 @@ pub type SharedFileEmitter = Arc<FileEmitter>;
 #[cfg(test)]
 mod tests {
   use super::*;
+  use regex::Regex;
```
sapphi-red:
```suggestion
```

crates/rolldown_common/Cargo.toml
```diff
@@ -47,6 +47,7 @@ num-bigint = { workspace = true }
 
 [dev-dependencies]
 insta = { workspace = true }
+regex = { workspace = true }
```
sapphi-red:
```suggestion
```

Cargo.lock
```diff
@@ -2683,6 +2683,7 @@ dependencies = [
  "oxc_resolver",
  "oxc_sourcemap",
  "oxc_str",
+ "regex",
```
sapphi-red:
```suggestion
```

sapphi-red (approved):

jjb:
🎉

PR #10636, opened 2026-08-05
docs(plugin-api): document sourcemap differences from Rollup

### LINKED ISSUE #10635: enabling sourcemap with classic worker using importScripts fails
Context:

The new workbox-build fork at Vite PWA org. will add support to build classic and esm service workers from ESM service worker source code. workbox-build will run both builds in parallel (the SW types for the build are classic, module and classic-and-module).

Since Rolldown cannot be used to build the classic (iife) SW with code splitting ( [StackBlitz with classic webworker with code splitting error]( https://stackblitz.com/edit/vitejs-vite-7fufakck?file=vite.config.ts) ), the new workbox-build adds a Rolldown plugin to "transform" the classic ESM build output to iife format including all its chunks, it just works.

The issue:

The issue with Rolldown or Vite 8 and the classic build is when we enable the sourcemap, then, the corresponding Rolldown build fails. I don't have the exact version of Rolldown breaking the build with sourcemap enabled, I started using beta versions, but I can fix Rolldown version to check it.

I need to check also if the issue is when using Vite+ from `@voidzero-dev/vite-plus-core@0.2.7`. My monorepo is working fine with Vite 8.1.5 and Rolldown 1.1.2, maybe there is a regression on new versions (run `pnpm why rolldown` and `pnpm why vite` at `examples/unplugin-pwa-vite8-ts`).

The Rolldown plugin and the logic at workbox-build can be found here https://github.com/userquin/composable-vite-pwa/blob/main/packages/workbox/build/src/build/builder/rolldown-plugin.ts  and the [prepareSWChunks module](https://github.com/userquin/composable-vite-pwa/blob/main/packages/workbox/build/src/build/builder/prepare-sw-chunks.ts#L65).

<details>
<summary>Rolldown error when enabling sourcemap with the classic SW build</summary>

Image
</details>

IWANABETHATGUY:
Rollup does not check the sourcemap a plugin returns. Rolldown checks every index, so merging a new map into `chunk.map` with `Object.assign` fails the build here and not there. The docs now cover that, and show how to transform a chunk in `renderChunk` instead.

closed https://github.com/rolldown/rolldown/issues/10635


docs/apis/plugin-api.md
```diff
@@ -282,3 +282,90 @@ In Rollup, the [`options`](/reference/Interface.Plugin#options) hook is called o
...
+    // ...your transform...
+    if (!s.hasChanged()) continue;
+
+    // `hires: true` is required, or the composed map comes back empty.
+    const step = s.generateMap({ source: chunk.fileName, hires: true });
+    chunk.code = s.toString();
+
+    // Assign the composed map, do not spread it. `toString()` lives on its
+    // prototype, and spreading would leave you with `[object Object]`.
+    chunk.map = remapping([step, chunk.map], () => null);
```
userquin: I guess you should add `@jridgewell/remapping` dependency somewhere

IWANABETHATGUY: nice catch!

docs/apis/plugin-api.md
```diff
@@ -282,3 +282,97 @@ In Rollup, the [`options`](/reference/Interface.Plugin#options) hook is called o
...
+Rollup keeps sourcemaps in JavaScript for the whole build, so it never checks a plugin's map as a whole. What it does with a bad index depends on the field. A mapping that points at a source the map does not list is dropped. A mapping that points at a missing name is kept, and only the name goes. Nothing is logged either way. Rollup does not always survive a bad map, though. A five-field mapping with no `names` array throws.
+
+Rolldown converts each plugin map into a Rust value, and that step checks every index. The build fails if a mapping points at a source or a name the map does not list, or if `x_google_ignoreList` holds an index past the end of `sources`:
+
+```
+Failed to convert json sourcemap to struct
+Reference to non-existing source at position 1
+```
+
+So a plugin that has always worked with Rollup can still fail here. The map was wrong in both cases. Only Rolldown says so.
```
sapphi-red: I think we can make this simpler.

````suggestion
Rollup does not validate the sourcemap passed to it and silently ignores it. Rolldown validates the sourcemap when converting to the internal representation. This can cause an error if an invalid sourcemap is passed. For example:

```
Failed to convert json sourcemap to struct
Reference to non-existing source at position 1
```
````

File: docs/apis/plugin-api.md [hunk 3]
```diff
@@ -282,3 +282,97 @@ In Rollup, the [`options`](/reference/Interface.Plugin#options) hook is called o
...
+renderChunk(code) {
+  const s = new MagicString(code);
+  s.prepend('/* banner */\n');
+  // Rolldown composes this map into the chunk's own, so the emitted `.map`
+  // follows the new line numbers and the ignore list is rebuilt for you.
+  return { code: s.toString(), map: s.generateMap({ hires: true }) };
+}
+```
+
+:::
```
sapphi-red: I don't think this is "usual" and can be removed.

docs/apis/plugin-api.md
```diff
@@ -282,3 +282,97 @@ In Rollup, the [`options`](/reference/Interface.Plugin#options) hook is called o
...
+
+### Editing a Chunk's Sourcemap
```
sapphi-red: I think the main point of this explanation is not the "Rollup / Rolldown difference".

I think this should live in https://rolldown.rs/apis/plugin-api/transformations (I think we don't need to mention about the validation in this section)

docs/apis/plugin-api.md
```diff
@@ -282,3 +282,97 @@ In Rollup, the [`options`](/reference/Interface.Plugin#options) hook is called o
...
+generateBundle(options, bundle) {
+  for (const chunk of Object.values(bundle)) {
+    if (chunk.type !== 'chunk' || !chunk.map) continue;
+
+    const s = new MagicString(chunk.code);
+    // ...your transform...
+    if (!s.hasChanged()) continue;
+
+    // `hires: true` is required, or the composed map comes back empty.
+    const step = s.generateMap({ source: chunk.fileName, hires: true });
```
sapphi-red:
```suggestion
    // `hires: 'boundary'` is required, or the composed map comes back empty.
    const step = s.generateMap({ source: chunk.fileName, hires: 'boundary' });
```
I think we should recommend `hires: 'boundary'`.

docs/apis/plugin-api.md
```diff
@@ -282,3 +282,97 @@ In Rollup, the [`options`](/reference/Interface.Plugin#options) hook is called o
...
+
+If you have to transform a chunk here, compose the maps and write the asset yourself:
+
+```js
+import remapping from '@jridgewell/remapping';
+import MagicString from 'magic-string';
+
+generateBundle(options, bundle) {
+  for (const chunk of Object.values(bundle)) {
+    if (chunk.type !== 'chunk' || !chunk.map) continue;
```
sapphi-red:
```suggestion
    if (chunk.type !== 'chunk') continue;
```
The transform should happen even if the map doesn't exist.

docs/apis/plugin-api.md
```diff
@@ -282,3 +282,97 @@ In Rollup, the [`options`](/reference/Interface.Plugin#options) hook is called o
...
+    // ...your transform...
+    if (!s.hasChanged()) continue;
+
+    // `hires: true` is required, or the composed map comes back empty.
+    const step = s.generateMap({ source: chunk.fileName, hires: true });
+    chunk.code = s.toString();
+
+    // Assign the composed map, do not spread it. `toString()` lives on its
+    // prototype, and spreading would leave you with `[object Object]`.
+    chunk.map = remapping([step, chunk.map], () => null);
```
sapphi-red:
```suggestion
    if (chunk.map) {
      chunk.map = remapping([step, chunk.map], () => null);
    }
```

docs/apis/plugin-api.md
```diff
@@ -282,3 +282,97 @@ In Rollup, the [`options`](/reference/Interface.Plugin#options) hook is called o
...
+
+    // The emitted file comes from this asset, not from `chunk.map`.
+    const asset = bundle[`${chunk.fileName}.map`];
+    if (asset) asset.source = chunk.map.toString();
+  }
+}
+```
+
+This cannot fix the hash. `generateBundle` runs after hashing, so the emitted filename still carries the hash of the untransformed code.
```
sapphi-red: I think we should start the section with something like "We discourage transforming in `generateBundle` for .... That said, ...".

docs/apis/plugin-api.md
IWANABETHATGUY: Accepted. I will tweak the wording about Rollup sourcemap validation, since Rollup does validate the sourcemap in some scenarios.

docs/apis/plugin-api.md
```diff
@@ -282,3 +282,12 @@ In Rollup, the [`options`](/reference/Interface.Plugin#options) hook is called o
 In Rollup, certain hooks like [`writeBundle`](/reference/Interface.FunctionPluginHooks#writebundle) are "parallel" by default, meaning they run concurrently across multiple plugins. This requires plugins to explicitly set `sequential: true` if they need their hooks to run one after another.
 
 In Rolldown, the [`writeBundle`](/reference/Interface.FunctionPluginHooks#writebundle) hook is already sequential by default, so plugins do not need to specify `sequential: true` for this hook.
+
+### Sourcemap Validation
+
+Rollup does not check a plugin's sourcemap against its own `sources` and `names`. A mapping that points at a missing source is dropped. A mapping that points at a missing name is kept without the name. Rolldown checks every index while converting the map to the internal representation. So a map that Rollup accepts can fail the build here. For example:
```
sapphi-red:
```suggestion
Rollup does not check a plugin's sourcemap against its own `sources` and `names`. A mapping that points at a missing source is dropped. A mapping that points at a missing name is kept without the name. Rolldown checks every index while converting the map to the internal representation. So an invalid map that Rollup accepts can fail the build here. For example:
```
I think we should clarify that those are invalid.

docs/apis/plugin-api/transformations.md
```diff
@@ -19,3 +19,48 @@ return {
   map: null,
 };
 ```
+
+## Transforming a Chunk
+
+Transform a chunk in [`renderChunk`](/reference/Interface.Plugin#renderchunk) and return the new map with it. Rolldown composes that map with the chunk's own, rebuilds `x_google_ignoreList`, and hashes the transformed code:
```
sapphi-red: To align with the other wordings in the docs:
```suggestion
To transform a chunk, you can use [`renderChunk`](/reference/Interface.Plugin#renderchunk). If you return the sourcemap for the transform you applied, Rolldown composes that map with the previous transforms.
```

Is there a reason to mention `x_google_ignoreList`?

docs/apis/plugin-api/transformations.md
```diff
@@ -19,3 +19,48 @@ return {
...
+
+```js
+import MagicString from 'magic-string';
+
+renderChunk(code) {
+  const s = new MagicString(code);
+  s.prepend('/* banner */\n');
+  return { code: s.toString(), map: s.generateMap({ hires: 'boundary' }) };
+}
```
sapphi-red:
```suggestion

export default function myPlugin() {
  return {
    name: 'example',
    renderChunk(code) {
      const s = new MagicString(code);
      s.prepend('/* banner */\n');
      return { code: s.toString(), map: s.generateMap({ hires: 'boundary' }) };
    }
  }
}
```
Let's make structurally correct so that it's not confusing.

docs/apis/plugin-api/transformations.md
```diff
@@ -19,3 +19,48 @@ return {
...
+      // Assign the composed map, do not spread it. `toString()` lives on its
+      // prototype, and spreading would leave you with `[object Object]`.
+      chunk.map = remapping([step, chunk.map], () => null);
+
+      // The emitted file comes from this asset, not from `chunk.map`.
+      const asset = bundle[`${chunk.fileName}.map`];
+      if (asset) asset.source = chunk.map.toString();
+    }
+  }
+}
```
sapphi-red:
```suggestion
export default function myPlugin() {
  return {
    name: 'example',
    generateBundle(options, bundle) {
      for (const chunk of Object.values(bundle)) {
        if (chunk.type !== 'chunk') continue;
    
        const s = new MagicString(chunk.code);
        // ...your transform...
        if (!s.hasChanged()) continue;
    
        // A low-resolution map can compose down to nothing, so keep the mappings at the boundaries.
        const step = s.generateMap({ source: chunk.fileName, hires: 'boundary' });
        chunk.code = s.toString();
    
        if (chunk.map) {
          // Assign the composed map, do not spread it. `toString()` lives on its
          // prototype, and spreading would leave you with `[object Object]`.
          chunk.map = remapping([step, chunk.map], () => null);
    
          // The emitted file comes from this asset, not from `chunk.map`.
          const asset = bundle[`${chunk.fileName}.map`];
          if (asset) asset.source = chunk.map.toString();
        }
      }
    }
  }
}
```

docs/apis/plugin-api/transformations.md
```diff
@@ -19,3 +19,48 @@ return {
...
+    // ...your transform...
+    if (!s.hasChanged()) continue;
+
+    // A low-resolution map can compose down to nothing, so keep the boundaries.
+    const step = s.generateMap({ source: chunk.fileName, hires: 'boundary' });
+    chunk.code = s.toString();
+
+    if (chunk.map) {
+      // Assign the composed map, do not spread it. `toString()` lives on its
+      // prototype, and spreading would leave you with `[object Object]`.
```
sapphi-red: If we want to mention about the "spread", I think we should clarify with the example. But I personally think we can just say `// compose the sourcemap` here.

docs/apis/plugin-api/transformations.md
IWANABETHATGUY: Because `sourcemapIgnoreList` defaults to `/node_modules/` (bindingify-output-options.ts:96), so `x_google_ignoreList` is set on most builds without the plugin author doing anything. That's to say, this is not an edge case, so I prefer to point it out specifically.

sapphi-red: I see. Then maybe:
```suggestion
To transform a chunk, you can use [`renderChunk`](/reference/Interface.Plugin#renderchunk). If you return the sourcemap for the transform you applied, Rolldown composes that map with the previous transforms and rebuilds `x_google_ignoreList` field based on the options.
```

sapphi-red (approved):
Thanks!

PR #10637, opened 2026-08-06
fix(dev): flush re-emitted assets when their content changes

btea:
The `FileEmitter` is reused for the whole DevEngine lifetime and `emitted_files` is never cleared in dev mode. Since the reference id of an asset emitted with an explicit `fileName` is derived from that name, re-emitting the asset with changed content in a later build was silently dropped after the first flush: `add_additional_files` skipped it forever. This broke consumers that serve emitted assets in dev, e.g. Vite's bundled dev server missing updated html entry assets (vitejs/vite#23179).

Record the source hash of every flushed asset and mark the file dirty when it is re-emitted with different content, so it is flushed again. Assets re-emitted with unchanged content stay deduplicated, and the filename conflict check still only runs on the first flush.

## AI disclosure

This PR was developed with AI assistance. All generated changes were reviewed, tested, and adjusted before submission.

### DIFF
--- a/crates/rolldown_common/src/file_emitter.rs
+++ b/crates/rolldown_common/src/file_emitter.rs
@@ -87,6 +87,10 @@ pub struct FileEmitter {
   options: Arc<NormalizedBundlerOptions>,
   /// Mark the files that have been emitted to bundle.
   emitted_files: FxDashSet<ArcStr>,
+  /// Source hashes of the files that have been flushed to a bundle, keyed by reference id.
+  /// The emitter is reused across dev-mode rebuilds, so a file that is re-emitted with
+  /// different content must be flushed again (see `insert_new_file`).
+  emitted_file_source_hashes: FxDashMap<ArcStr, ArcStr>,
   emitted_chunks: FxDashMap<ArcStr, ArcStr>,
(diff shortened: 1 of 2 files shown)


crates/rolldown_common/src/file_emitter.rs
```diff
@@ -329,6 +334,16 @@ impl FileEmitter {
     sanitized_file_name: Option<ArcStr>,
   ) -> anyhow::Result<()> {
     self.generate_file_name(file, hash, asset_filename_template, sanitized_file_name)?;
+    // The emitter is reused across dev-mode rebuilds. If this asset was already flushed
+    // to a bundle but is now re-emitted with different content, mark it dirty so that
+    // `add_additional_files` flushes it again instead of dropping the update.
+    if self
+      .emitted_file_source_hashes
+      .get(&reference_id)
+      .is_some_and(|flushed_hash| flushed_hash.as_str() != hash.as_str())
```
h-a-n-a:
```suggestion
      .is_some_and(|flushed_hash| &*flushed_hash != hash)
```

We can stick to the original compare method here for better performance in potential.

h-a-n-a (approved):
LGTM. Thanks! Hand over to you @shulaoda . You're the assignee.

PR #10638, opened 2026-08-06
fix(binding): illegal instruction crash on Android arm64

### LINKED ISSUE #6342: [Bug]: Illegal instruction on android arm64

- Do `npm install rolldown`
- Run `npx rolldown`

### What is expected?

Show help screen (_maybe_), it's normal.

```console
~/p $ npx rolldown
Illegal instruction        npx rolldown
```

### System Info

```Shell
# I can't do that
```

### Any additional comments?

I use `@rolldown/binding-android-arm64` binding package.

shulaoda:
Fixes #6342.

On `aarch64-linux-android`, mimalloc was built with arch optimization enabled, because the `no_opt_arch` feature was only applied under `cfg(all(target_os = "linux", target_arch = "aarch64"))` and Android has its own `target_os`. The optimized build emits ARMv8.1 instructions, so devices with ARMv8.0 CPUs (such as Cortex-A53) crash with `Illegal instruction` (SIGILL) at startup.

This PR extends the cfg to aarch64 on Linux, Android, and FreeBSD, so mimalloc is always built with `no_opt_arch` there. FreeBSD arm64 previously matched none of the mimalloc cfg blocks, so it is also covered now.


GourangaDasSamrat:
Hi @shulaoda,

Thanks for the fix — as mentioned earlier, I tested this pkg.pr.new build on my Android ARM64 device (Cortex-A53 / MediaTek MT6762V) and it completely resolves the `Illegal instruction` (SIGILL) crash. Vite + React dev server now starts and runs without issues.

Since this is still a draft PR, could you share if there's an estimated timeline for:

1. Marking this as ready for review / getting it approved and merged into `main`
2. Cutting a new npm release that includes this fix

A lot of us on Termux/Android are currently blocked on this for running any Vite-based project (React, Vue, etc.), so any update on the timeline would be really helpful. Happy to help with further testing if needed. Thanks again for the quick turnaround on this!

shulaoda:
@GourangaDasSamrat Hey! I’ll merge this before the next release, which is scheduled for Wednesday.

IWANABETHATGUY (approved):

h-a-n-a (approved):

PR #10655, opened 2026-08-09
fix(renamer): rename nested `require`/`__filename`/`__dirname` bindings in CJS output

### LINKED ISSUE #10654: [Bug]: CJS output shadows the injected `require` when a module declares its own `var require`

https://github.com/marcoroth/rolldown-cjs-require-shadow-repro

```sh
git clone https://github.com/marcoroth/rolldown-cjs-require-shadow-repro
cd rolldown-cjs-require-shadow-repro
npm install
npm start     # exits 1 — prints the emitted line, then the TypeError
```

`src/entry.js`:

```js
export async function init() {
  const { createRequire } = await import("node:module")
  var require = createRequire(import.meta.url)

  const path = require("node:path")
  const os = require("node:os")

  return path.sep + os.EOL
}
```

`rolldown.config.mjs`:

```js
import { defineConfig } from "rolldown"

export default defineConfig({
  input: "./src/entry.js",
  output: { file: "./dist/entry.cjs", format: "cjs" },
  platform: "node",
})
```

### What is expected?

`import.meta.url` is rewritten for CJS output using a `require("url")` call. When the module already declares its own `require` binding in that scope, the injected call should reference the real CommonJS `require`, either by emitting it before the shadowing declaration or by renaming the local binding.

Rollup handles this by deconflicting the local to `require$1`:

```js
var require$1 = createRequire(require('u' + 'rl').pathToFileURL(__filename).href);
```

The injected `require("url")` is emitted *inside the initializer of the shadowing declaration*:

```js
async function init() {
	const { createRequire } = await import("node:module");
	var require = createRequire(require("url").pathToFileURL(__filename).href);
	const path = require("node:path");
	const os = require("node:os");
	return path.sep + os.EOL;
}
```

`var require` is hoisted to the top of the function, so the `require("url")` in its own initializer resolves to the local binding, which is still `undefined` at that point. The module throws `TypeError: require is not a function` on the first call.

The ESM output is unaffected, since `import.meta.url` needs no rewriting there.

### System Info

```Shell
System:
    OS: macOS 15.7.7
    CPU: (16) arm64 Apple M4 Max
    Memory: 3.98 GB / 64.00 GB
    Shell: 4.0.2 - /opt/homebrew/bin/fish
  Binaries:
    Node: 24.16.0 - /Users/marcoroth/.local/share/mise/installs/node/24.16.0/bin/node
    Yarn: 1.22.22 - /Users/marcoroth/.local/share/mise/installs/node/24.16.0/bin/yarn
    npm: 11.13.0 - /Users/marcoroth/.local/share/mise/installs/node/24.16.0/bin/npm
    pnpm: 11.21.0 - /Users/marcoroth/.local/share/mise/installs/node/24.16.0/bin/pnpm
    Deno: 2.9.0 - /opt/homebrew/bin/deno
    Watchman: 2026.01.12.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 150.0.7871.187
    Safari: 26.5
```

### Any additional comments?

This is not a hand-written pattern. Emscripten emits exactly this shape in its generated glue when built with `-s EXPORT_ES6=1 -s ENVIRONMENT='node'`:

```js
if (ENVIRONMENT_IS_NODE) {
  const { createRequire } = await import("node:module");
  var require = createRequire(import.meta.url)
}
```

and later reads that binding for `require("node:fs")`, `require("node:path")` and `require("node:url")`, so it stays a variable. Emscripten used to allow opting out with `-s USE_ES6_IMPORT_META=0`, but current versions reject it with `Disabling is no longer supported`, so the shape cannot be avoided on the producing side.

We hit this migrating a monorepo from rollup to rolldown. Every CommonJS bundle carrying the emscripten module, four CLIs and a language server, threw on load. We are working around it with a `transform` plugin that renames the binding where the generated file enters the graph, which we would like to drop once this is fixed.

marcoroth:
Fixes #10654 

CommonJS output rewrites `import.meta.url` to `require("url").pathToFileURL(__filename).href`. The injected identifier is built as a bare, unbound identifier in `try_rewrite_import_meta_prop_expr`, so it is invisible to the renamer. When the module declares its own `require` in an enclosing scope, that binding is hoisted and captures the injected call inside its own initializer:

```js
// input
function init() {
  var require = createRequire(import.meta.url);
  const path = require("node:path");
  const os = require("node:os");
  return path.sep + os.EOL;
}
```

```js
// output before this PR — throws `require is not a function` on the first call
var require = createRequire(require("url").pathToFileURL(__filename).href);
```

The renamer already reserves `require` for CommonJS output, but only for root-scope symbols, so a nested binding is left alone.

#### The change

Adds `rename_bindings_shadowing_cjs_require`, a fifth nested-scope pass alongside the existing ones. It is modelled on the wrapper-params pass, which solves the same shape of problem for `exports` and `module`, and runs only for `OutputFormat::Cjs`.

Output now matches rollup, which deconflicts the local the same way:

```js
var require$1 = createRequire(require("url").pathToFileURL(__filename).href);
```

#### Alternatives considered

Hoisting the polyfill to a chunk-level variable also works and is immune to shadowing, but it changes the emitted shape for every `import.meta.url` user rather than only the shadowing case.

Making the injected identifier a canonical, symbol-backed reference would be the more general fix, since the same bare `require` identifier is emitted for external imports too. Renaming the shadowing binding is the smaller change and follows a pattern already established in this file.

#### Notes for reviewers

The pass renames any nested binding named `require` under CommonJS output, not only those in scopes that actually enclose an injected call. That is deliberately broader than the reference-precise passes above it, because several rewrites emit a bare `require(...)`, not just the `import.meta.url` polyfill. Renaming a local is semantically transparent, and in practice it is inert: the full suite produced no snapshot changes outside the new test. Happy to narrow it to reference-precise if preferred.

### DIFF
--- a/crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
+++ b/crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
@@ -380,6 +380,11 @@ fn rename_shadowing_symbols_in_nested_scopes<'a>(
       output_format,
       OutputFormat::Iife | OutputFormat::Umd | OutputFormat::Cjs
     ));
+
+    if matches!(output_format, OutputFormat::Cjs) {
+      ctx.rename_bindings_shadowing_cjs_require();
+    }
+
     ctx.rename_cjs_locals_shadowing_referenced_chunk_bindings();
   }
 }
--- a/crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
+++ b/crates/rolldown/src/utils/chunk/deconflict_chunk_symbols.rs
@@ -382,7 +382,7 @@ fn rename_shadowing_symbols_in_nested_scopes<'a>(
     ));
 
     if matches!(output_format, OutputFormat::Cjs) {
-      ctx.rename_bindings_shadowing_cjs_require();
+      ctx.rename_bindings_shadowing_cjs_ambient_names();
     }
 
     ctx.rename_cjs_locals_shadowing_referenced_chunk_bindings();
--- a/crates/rolldown/src/utils/renamer.rs
+++ b/crates/rolldown/src/utils/renamer.rs
@@ -480,6 +480,41 @@ impl NestedScopeRenamer<'_, '_> {
     }
   }
 
+  /// Rename nested bindings that would shadow the ambient `require` of CommonJS output.
+  ///
+  /// Several rewrites emit a bare `require(...)` call into the module body: external imports, and
+  /// the `import.meta.url` polyfill (`require("url").pathToFileURL(__filename).href`). Those calls
+  /// mean the CommonJS `require`, so a nested binding of the same name must not capture them.
+  ///
+  /// The binding is hoisted, so it shadows even an injected call inside its own initializer:
+  ///
+  /// ```js
+  /// // input, the shape emscripten emits with `-s EXPORT_ES6=1 -s ENVIRONMENT='node'`
+  /// function init() {
+  ///   var require = createRequire(import.meta.url);
+  ///   return require("node:path").sep;
+  /// }
+  /// ```
+  ///
+  /// Without renaming, the polyfill resolves to the still-undefined local and the module throws
+  /// `require is not a function` on first call:
+  ///
+  /// ```js
+  /// var require = createRequire(require("url").pathToFileURL(__filename).href);
+  /// ```
+  pub fn rename_bindings_shadowing_cjs_require(&mut self) {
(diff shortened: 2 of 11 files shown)


Dunqing (approved):
Thank you for working on this fix! The implementation looks good and cleaner. I pushed a comment to cover `import.meta.filename` and `import.meta.dirname`, which are transformed to `__filename` and `__dirname`!

marcoroth:
Thanks for the merge, @Dunqing and @IWANABETHATGUY! 🙏🏼

PR #10660, opened 2026-08-10
fix(code-splitting): stop negated sideEffects globs retaining modules

JoviDeCroock:
Package sideEffects patterns form a positive allowlist, but fast_glob interprets a leading exclamation mark as a negation. Passing those patterns through made each one match nearly every unrelated module and prevented affected packages from tree-shaking.

Ignore leading-bang string and array patterns before matching, and cover the behavior with a generic unit and integration regression.

> [!NOTE]
> This reproduction was created by a human, implemented by GPT5.6-sol and reviewed by a human

### DIFF
--- a/crates/rolldown_common/src/types/package_json.rs
+++ b/crates/rolldown_common/src/types/package_json.rs
@@ -49,21 +49,12 @@ impl PackageJson {
   pub fn check_side_effects_for(&self, module_path: &str) -> Option<bool> {
     let side_effects = self.side_effects.as_ref()?;
     // Is it necessary to convert module_path to relative path?
-    // `sideEffects` is a positive allowlist, so a leading-`!` entry never marks a module as
-    // side-effectful. Handing it to `fast_glob` would instead match nearly every path that
-    // does not match the remainder of the pattern.
     match side_effects {
       SideEffects::Bool(s) => Some(*s),
-      SideEffects::String(pattern) => Some(
-        !pattern.starts_with('!')
-          && glob_match_with_normalized_pattern(pattern.as_str(), module_path),
-      ),
-      SideEffects::Array(patterns) => Some(
-        patterns
-          .iter()
-          .filter(|pattern| !pattern.starts_with('!'))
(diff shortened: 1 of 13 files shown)


crates/rolldown_common/src/types/package_json.rs
```diff
@@ -49,12 +49,63 @@ impl PackageJson {
...
+      SideEffects::String(pattern) => Some(
+        !pattern.starts_with('!')
+          && glob_match_with_normalized_pattern(pattern.as_str(), module_path),
+      ),
+      SideEffects::Array(patterns) => Some(
+        patterns
+          .iter()
+          .filter(|pattern| !pattern.starts_with('!'))
+          .any(|pattern| glob_match_with_normalized_pattern(pattern, module_path)),
+      ),
```
IWANABETHATGUY: `!foo.js` is a legal path on POSIX. Rollup (with `@rollup/plugin-node-resolve`) matches it literally when `!foo.js` is an item in the `sideEffects` array. So I prefer to escape the leading `!` in `glob_match_with_normalized_pattern` instead of dropping the pattern.

```suggestion
    match side_effects {
      SideEffects::Bool(s) => Some(*s),
      SideEffects::String(p) => Some(glob_match_with_normalized_pattern(p.as_str(), module_path)),
      SideEffects::Array(pats) => {
        Some(pats.iter().any(|p| glob_match_with_normalized_pattern(p.as_str(), module_path)))
      }
```

and then in `side_effects.rs`, before the `fast_glob::glob_match` call:

```rs
  // `fast_glob` reads a leading `!` as a negation, but `sideEffects` is a positive allowlist.
  if normalized_glob.starts_with('!') {
    normalized_glob.insert(0, '\\');
  }
```

JoviDeCroock: Implemented and added a test, thank you for the review

IWANABETHATGUY (approved):

IWANABETHATGUY (approved):

PR #10678, opened 2026-08-15
perf: enable compiler cache via `module.enableCompileCache()`

btea:
https://beta.docs.nodejs.org/module#moduleenablecompilecacheoptions

```
hyperfine --warmup 5 --runs 50 'node bin/cli.mjs --version'  # without this PR
 Time (mean ± σ):      50.4 ms ±   2.1 ms    [User: 48.4 ms, System: 7.6 ms]
 Range (min … max):    47.4 ms …  58.3 ms    50 runs

hyperfine --warmup 5 --runs 50 'node bin/cli.mjs --version'  # with this PR
 Time (mean ± σ):      47.6 ms ±   1.0 ms    [User: 45.8 ms, System: 7.6 ms]
 Range (min … max):    46.2 ms …  51.0 ms    50 runs
```


shulaoda (approved):
Thanks for the contribution.

Vite does the same thing, and later added `module.flushCompileCache()` after 10s. The compile cache is only written to disk when the process exits. For `--watch`, the process stays alive, so the cache may never be flushed. Vite also found that flushing in a SIGINT handler can hang the process.

We may want the same flush for Rolldown `--watch`. Something like:

```js
import module from 'node:module';
try {
  module.enableCompileCache?.();
  setTimeout(() => {
    try {
      module.flushCompileCache?.();
    } catch {}
  }, 10 * 1000).unref();
} catch {}
await import('../dist/cli.mjs');
```

Also, the benefit for Rolldown is likely small. This API only speeds up JS parse/compile. Most of Rolldown is the native Rust addon, which this cache does not speed up.

btea:
Yes, for Rolldown the overall benefits are extremely limited.

PR #10698, opened 2026-08-18
feat(code-splitting): PoC of `codeSplitting.inlineCommonChunks`, not for merge

hyfdev:
POC only for exploring


hyfdev:
Dropped as this is a one rough attempt

PR #10699, opened 2026-08-18
ci(wasi): retry the Node Test step to absorb the shared-dlmalloc flake

hyfdev:
I noticed wasi job sometimes fails. After investigation, I tracked it down to https://github.com/rolldown/rolldown/issues/10697.

The fundamental fix isn't easy and require upstream works. So Adding a retry mechanism to make things easier for now.

PR #10700, opened 2026-08-18
test(dev): expect a hot update across a circular import

hyfdev:
Fix dev test ci error found in https://github.com/rolldown/rolldown/pull/10699

The dev-server harness builds whatever `vitejs/vite` `rolldown-canary` rebased onto `main` gives it, so this arrived without any rolldown change: the last green run is 06:12Z and the first red one is 07:39Z, with vitejs/vite#23259 landing at 07:26Z between them. Every `Node Dev Server Test` job has been red since, on every branch and every Node version, including `main`.

### The playground

```
main.js  →  a.js  →  b.js  ⇄  c.js
   ↑                    b and c import each other
   └── only main calls import.meta.hot.accept()
```

### What changed upstream

Editing `c.js` makes the HMR client walk up the importers looking for a module that accepts updates:

```
c.js → imported by b; b does not accept, keep going
b.js → imported by a and by c
         via a: a → main, and main accepts ✓
         via c: that is where the walk started, it loops back
```

It used to drop the whole walk the moment it looped back — including the `main` boundary it had already found — and reload the page:

```
full reload needed: circular import chain between `b.js` and `c.js`
```

vitejs/vite#23259 makes it skip the path it already walked and keep the boundary it found, so the edit is applied through `main` and the page stays. Vite's unbundled dev has behaved this way since vitejs/vite#14867; bundled dev was the outlier.

### Why the test broke

`hmr-accept-outside-circular` pinned the old behavior: wait for that log line, then assert the page reloaded (a marker planted on `window` is gone). Neither holds now — the log is deleted, so the wait times out with `Timeout waiting for browser logs`, and the marker survives.

The spec now asserts the hot update: the edit renders and the marker survives. The marker also carries the original guarantee, which the log used to carry — if the walk missed `main`'s boundary entirely, the page would reload and wipe it, so a "content is fresh" assertion alone cannot pass by accident.

Validation: the full browser suite passes locally against Vite `main` at c0f2fc607, which contains the change — 104 passed / 12 skipped, exactly CI's counts with the one failure flipped. Not verified: the new assertion has not been run against a pre-#23259 Vite.

### DIFF
--- a/packages/test-dev-server/tests/playground/hmr-accept-outside-circular/__tests__/hmr-accept-outside-circular.spec.ts
+++ b/packages/test-dev-server/tests/playground/hmr-accept-outside-circular/__tests__/hmr-accept-outside-circular.spec.ts
@@ -1,5 +1,5 @@
 import { describe, expect, test } from 'vitest';
-import { editFile, page, untilBrowserLogAfter, waitForBuildStable } from '~utils';
+import { editFile, page, waitForBuildStable } from '~utils';
 
 // Port of the removed Rust fixture
 // `crates/rolldown/tests/rolldown/topics/hmr/accept-outside-circular`: same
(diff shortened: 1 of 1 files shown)


h-a-n-a (approved):

hyfdev:
### Merge activity

* **Aug 18, 1:59 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Aug 18, 1:59 PM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Aug 18, 2:08 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10703, opened 2026-08-18
feat(binding): allocation-tracking global allocator behind the `tracking_allocator` feature

IWANABETHATGUY:
This PR adds a `tracking_allocator` cargo feature to the binding so that a build can report live and peak Rust-side memory to JS, exposed as `getNativeMemoryStats` / `resetNativeMemoryStats`. Enabling it shifts timing by at most ~2.3%; a stock build is unchanged.


hyfdev (approved):

h-a-n-a (approved):

shulaoda (approved):

crates/rolldown_binding/src/lib.rs
```diff
@@ -29,17 +29,32 @@ use napi_derive::napi;
...
+static ALLOC: rolldown_tracking_allocator::TrackingAllocator =
+  rolldown_tracking_allocator::TrackingAllocator;
+
 pub mod binding_bundler;
 pub mod binding_dev_engine;
 pub mod binding_dev_options;
 pub mod binding_watcher_bundler;
 pub mod classic_bundler;
 mod generated;
+pub mod native_memory;
```
shulaoda, line 57: Could we also add a cfg `default_global_allocator` feature gate?

IWANABETHATGUY: gate the native_memory module itself out — is not reasonable, for two concrete reasons:

1. napi codegen generates from the compiled crate, so a cfg-gated module changes the export surface per feature set. binding.cjs and binding.d.cts would differ between builds, and the generated-code check in CI would fail for whichever build did not produce the committed files.
2. The stacked bench PR imports getNativeMemoryStats from rolldown/experimental unconditionally and branches on null. A vanishing export turns that graceful branch into an import error on every stock binding.

PR #10707, opened 2026-08-18
chore(deps): upgrade oxc to 0.146.0

Boshen:
Rolldown needs Oxc 0.146.0 to pick up the isolated-declarations fix for unresolved parameter types while retaining the 0.145.0 property-mangler migration.

### Regression fix

Oxc 0.146.0 includes [oxc-project/oxc#25909](https://github.com/oxc-project/oxc/pull/25909), which preserves unresolved parameter types during isolated declaration generation. This removes the declaration-generation failure observed in Rolldown's Vite compatibility job.

The Rust workspace dependencies, `Cargo.lock`, and npm catalog packages (`@oxc-project/runtime`, `@oxc-project/types`, `oxc-parser`, and `oxc-transform`) now use 0.146.0.

### Breaking changes

Both come from oxc's new property mangler:

- `oxc::minifier::MinifierOptions` gained a required `mangle_properties: Option<ManglePropertiesOptions>` field. Set to `None` in all three initializers in `minify_options.rs` — the bundler does not plumb property mangling through.
- `oxc_minify_napi::MinifyOptions` gained `mangleProps`, which leaked into rolldown's public `MinifyOptions` (`Omit<BindingMinifyOptions, 'module' | 'sourcemap'>`) and broke the `IsSchemaSubType` assertion in `validator.ts`. Added `'mangleProps'` to the `Omit`, since `normalize_binding_options` only reads `mangle`/`compress`/`codegen` and would silently drop the option.

### Snapshot churn

- **Class bindings are now preserved by DCE** — `(class { static { foo } })` becomes `var A_keep = class { static { foo } };`. This accounts for most of the changed snapshots.
- **`@__KEY__` / `#__KEY__` annotations are now emitted** by codegen (`mangle_props_key_comment*`, `preserve_key_comment`) — groundwork for the new property mangler.
- **Better template-literal DCE** — `dce_template_literal` drops `var alsoKeep;` and `${alsoKeep}`, which is provably `undefined` and so has no `ToString` side effect.

The esbuild compatibility metric is unchanged: `stats.md` stays at 688 passed / 87.42%, and every `summary/*.md` regenerated byte-identical, so none of these codegen changes move a case across the pass/fail line.

### Transitive updates

`oxc-browserslist` 3.0.11 → 5.0.1, `oxc_sourcemap` 8.1.0 → 8.1.2, `oxc-miette` 4.0.0 dropped, `lazy-regex` 3.6.1 added (pulled in by the property mangler).

Verified after rebasing onto current `main`: the full Rust workspace suite, parser and transformer package smoke tests, native Rolldown debug binding build, Vite monorepo build and declaration generation, plus 934 unit, 1,128 serve, 690 bundled-serve, and 923 build tests.


crates/rolldown/tests/esbuild/dce/dce_template_literal/artifacts.snap
```diff
@@ -7,10 +7,8 @@ source: crates/rolldown_testing/src/integration_test.rs
 
 ```js
 //#region entry.js
-var alsoKeep;
```
Boshen, line 10: @Dunqing

crates/rolldown/tests/esbuild/dce/dce_of_decorators/artifacts.snap
```diff
@@ -12,25 +12,25 @@ const fn = () => {
 };
 //#endregion
 //#region keep-these.js
-(@fn class {});
-(class {
+var Class = @fn class {};
```
IWANABETHATGUY, line 15: The previous output looks correct, but not a blocker.

shulaoda: @Dunqing cc

crates/rolldown/tests/esbuild/dce/dce_template_literal/artifacts.snap
Dunqing: Just confirmed that this is correct since `alsoKeep` is a variable without initialization and has no side effects, so `alsoKeep` can be deleted safely.

crates/rolldown/tests/esbuild/dce/dce_of_decorators/artifacts.snap
Dunqing: This is correct because @sapphi-red fixed the pre-existing issue to retain the name for the class expression. Rolldown runs DCE with `keepNames`, so the name is preserved.

shulaoda (approved):

IWANABETHATGUY (approved):

Dunqing:
We need to wait for a new Oxc release to fix the failed CI job.

crates/rolldown/tests/esbuild/dce/dce_of_decorators/artifacts.snap
sapphi-red: Yes, linking the PR here for reference: https://github.com/oxc-project/oxc/pull/25584

crates/rolldown/tests/esbuild/default/mangle_props_key_comment/artifacts.snap
```diff
@@ -8,9 +8,19 @@ source: crates/rolldown_testing/src/integration_test.rs
...
+	`_mangleThis`
+);
+x._mangleThisToo(
+	/* #__KEY__ */
+	"_mangleThisToo",
+	/* #__KEY__ */
+	`_mangleThisToo`
+);
+x._someKey = /* #__KEY__ */ "_someKey" in y;
 x([`foo._mangleThis = bar._mangleThisToo`, `foo.notMangled = bar.notMangledEither`]);
```
sapphi-red, line 24: (not a blocker) Maybe we need to avoid inlining in DCE when `/* #__KEY__ */` is specified on the string.

input:
```js
x([
    `foo.${/* @__KEY__ */ '_mangleThis'} = bar.${/* @__KEY__ */ '_mangleThisToo'}`,
    `foo.${/* @__KEY__ */ 'notMangled'} = bar.${/* @__KEY__ */ 'notMangledEither'}`,
])
```

sapphi-red: slightly related to this, maybe we should port these `/* @__KEY__ */` related test cases to oxc repo, if equivalent doesn't exist

sapphi-red (approved):

crates/rolldown/tests/esbuild/default/mangle_props_key_comment/artifacts.snap
sapphi-red: @Dunqing

Dunqing: Ok, i will create a issue for this

PR #10714, opened 2026-08-19
feat(dev): invalidate query-suffixed modules when their file changes

h-a-n-a:
This resolves a behavior mismatch issue in Rolldown that query variants are not invalidated when a query variant module whose file changes is not changed.

For example, 

```js
import './index.html?html-proxy&index=0'
import './index.html?html-proxy&index=1'
```

Both modules above should be invalidated when `index.html` changes, whereas only `index.html` is invalidated now.

Details: https://github.com/vitejs/vite/pull/23194#discussion_r3802163794


h-a-n-a:
* **#10714** <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10714?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rolldown/rolldown/10714?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rolldown?ref=merge-queue-instructions-in-stack-comment&prId=10714).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/rolldown/src/hmr/hmr_stage.rs
```diff
@@ -140,6 +141,20 @@ impl<'a, Fs: FileSystem + Clone + 'static> HmrStage<'a, Fs> {
...
       affected_modules.extend(transform_dep_modules);
+      // Modules addressed as `<changed file>?query` (an html proxy module, a
+      // `foo.vue?vue&type=...` sub-module) read their content from the changed file, so
+      // they are re-fetched with it; the exact-id lookup above cannot see them and would
+      // keep serving their cached copies.
+      if let Some(variant_idxs) = self.cache.module_idxs_by_clean_path.get(&changed_file_path) {
+        let mut query_variant_modules = variant_idxs
+          .iter()
+          .copied()
+          .filter(|module_idx| self.module_table().modules[*module_idx].as_normal().is_some())
```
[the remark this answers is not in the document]
h-a-n-a: Fixed

crates/rolldown/src/types/scan_stage_cache.rs
```diff
@@ -37,6 +39,9 @@ pub struct ScanStageCache {
   pub user_defined_entry: FxHashSet<ModuleId>,
   // Usage: Map file path emitted by watcher to corresponding module index
   pub module_idx_by_abs_path: FxHashMap<ArcStr, ModuleIdx>,
+  /// Clean path (module id with `?query`/`#fragment` stripped) -> modules addressed as
+  /// `<clean path>?query`.
+  pub module_idxs_by_clean_path: FxHashMap<ArcStr, FxIndexSet<ModuleIdx>>,
```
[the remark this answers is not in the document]
h-a-n-a: Fixed

hyfdev (approved):

packages/test-dev-server/tests/playground/hmr-query-variant/dev.config.mjs
```diff
@@ -0,0 +1,46 @@
...
+        return `${cleanPath}?upper`;
+      }
+      return null;
+    },
+    load(id) {
+      if (!id.endsWith('?upper')) {
+        return null;
+      }
+      const cleanPath = id.slice(0, -'?upper'.length);
+      const code = nodeFs.readFileSync(cleanPath, 'utf-8');
```
sapphi-red, line 26: Since this module depends on `cleanPath`, shouldn't this plugin call `this.addWatchFile`? In that case, wouldn't Rolldown work fine?

h-a-n-a: I assume this is a builtin behavior that the bundler would auto-listen to the file path without the query the way Vite html plugin and Vite vue plugin do. It might never occur to them that a module with `foo.vue?vue&type=script` has to listen `foo.vue` again.

sapphi-red: `this.addWatchFile` wasn't needed in unbundled-dev because unbundled-dev doesn't cache the module transforms. The equivalent would be to disable the cache completely, but I don't think we should do that.
The middle-ground would be to auto-watch these files like this PR. But that will cause problems like https://github.com/rollup/rollup/issues/5013 (for example, even if a plugin reads `/home/project/a/foo/bar` for id `/foo?bar`, `/foo` will be watched incorrectly, and there's no way to avoid that from the plugin). Even if we introduce this behavior, I think it should be implemented to build watch mode as well for consistency.
The plugins need to update the code, but I think it's needed. The plugins that didn't call `this.addWatchFile` was broken in build watch mode so it had to fix anyways.

h-a-n-a: That's interesting. Putting the watch behavior aside, a change to the file whose filename is also linked to a module whose module id is combining which with a query or a fragment should be invalidated. This is how it is handled in Vite today ([ref](https://github.com/vitejs/vite/blob/de1111ab0be00879b404e7ed3b2a80e264edddc1/packages/vite/src/node/server/moduleGraph.ts#L343-L389)), which just looks like how auto-watch handles these files (even though it's not auto-watch in practice. It's just invalidating the affiliated modules altogether when that file changes). 

> But that will cause problems like https://github.com/rollup/rollup/issues/5013 (for example, even if a plugin reads /home/project/a/foo/bar for id /foo?bar, /foo will be watched incorrectly, and there's no way to avoid that from the plugin).

This might not be a big issue I presume. Since in the [source code](https://github.com/vitejs/vite/blob/de1111ab0be00879b404e7ed3b2a80e264edddc1/packages/vite/src/node/server/moduleGraph.ts#L368), only `filename` is expected in the `resolvedId`, which in this example, it's `/foo` and it will end up in `fileToModulesMap` as a key even though it's useless because there won't be function call with `/foo` in the changed set. In additional to that, bundled-dev Rolldown only watches files that come from disk read (without taking other manual way into account) and unbundled-dev Vite only watches project root recursively by default, making `/foo` a dead key in the map that will not be watched. It might not add `/foo` to the watch list in the first place (?, I presume).

sapphi-red: Hmm, but I think the plugin would still need to call `this.addWatchFile` if the file is a virtual module.

For example, assume a virtual module with id=`/foo/my-virtual.vue`, which uses code from `/foo/bar.vue` and `/foo/baz.vue`. Module with id=`/foo/my-virtual.vue?vue&script` is the script tag part of `/foo/my-virtual.vue`. In this case, this PR will add filepath=`/foo/my-virtual.vue` as the dependency of `/foo/my-virtual.vue?vue&script`. But the dependency needed here is `/foo/bar.vue` and `/foo/baz.vue`, which is the dependency of module id=`/foo/my-virtual.vue`. So the plugin would still need to call `this.addWatchFile`.

I think what is expected here is to register the dependencies of the module without the query as the dependency, or to have a concept of dependency between modules (if we want to register the relationships automatically). That said, these behaviors are different from unbundled dev... We can implement this behavior in unbundled dev though.

Having `.file` on the module graph node always felt something wrong as it's not always true that there's one-to-one mapping of module -> file.

h-a-n-a: > In this case, this PR will add filepath=/foo/my-virtual.vue as the dependency of /foo/my-virtual.vue?vue&script. But the dependency needed here is /foo/bar.vue and /foo/baz.vue, which is the dependency of module id=/foo/my-virtual.vue. So the plugin would still need to call this.addWatchFile.

I agree with you. In this virtual module case, it definitely requires the plugin to call `this.addWatchFile`. However, for non-virtual module cases, should we make the implicit contract that this PR supports or do you prefer letting plugins write their own watching logic? AFAIK, If the answer is the latter one, plugin hooks for sub-module cases like `foo.vue?vue&type=script` would need to call `addWatchFile` to `foo.vue` explicitly.

sapphi-red: Hmm, let's add this feature in both bundle-dev & build with a opt-in flag (`legacy.autoDependencyRegistrationForModulesWithQuery`) and enable that in Vite by default. Later on, we can start emitting warnings for it and remove it.

I guess we eventually should add `this.load()` with an option that registers the relationship between modules. IIRC `this.load` will not set up that relationship now, right?

h-a-n-a: _(The following content is written with AI help for structure and wording and the thoughts are mine)_

To summarize the discussion so far, there are two separate problems, and they exist in unbundled-dev, bundled-dev, and build watch alike:

**Problem 1 — deriving a file path from the module id (`cleanPath`) is unsound.**

The clean path is only *accidentally* correct: it works when the id's path part really is the source file (`foo.vue?vue&type=script` → `foo.vue`), and it is wrong when the base is a virtual id (`/my-virtual.vue?vue&type=script` → a path that exists nowhere). The engine cannot tell the two apart from the id alone.

**Problem 2 — a module→*file* dependency is the wrong shape for the virtual case.**

- Non-virtual base: linking the sub-module to the base *file* happens to be correct. This (combined with problem 1) is exactly what unbundled-dev does today.
- Virtual base: `/my-virtual.vue` is assembled from `bar.vue` and `baz.vue`. Its sub-module `/my-virtual.vue?vue&type=script` needs to follow **the base module's file dependencies plus its own** — a single file link to the (nonexistent) base path cannot express that. What the sub-module really depends on is the base *module*, not any file.

**Consequence of 1 + 2:** watching a nonexistent path is harmless but useless, and invalidation keyed on it never fires — the failure is silent. So the fix should not be a better file-path guess. It should change *what* the derived dependency points at.

## Proposal: module-first auto dependency, with a file fallback

I like the direction of your suggestion a lot. The proposal below is really just that insight taken one step further: if the module relationship is the right primitive, the engine can derive it automatically for the query convention, and then the flag has nothing left to deprecate.

So, instead of the `legacy.` flag plus a later `this.load` dependency API, I think we can make the derivation itself sound. For a query module `X?q` with `cleanPath = X`:

1. **If `X` is a module in the graph** — add an invalidation edge `X?q → module X`. The sub-module then follows whatever the base module depends on (its file, plus everything the base registered via `addWatchFile`). This is the "register the dependencies of the module without the query" behavior you described, derived automatically.
2. **Otherwise, if `X` exists on the filesystem** — add a plain file dependency `X?q → file X`. This covers ids like `foo.svg?url` where the base is never a module of its own.
3. **Otherwise** — add nothing. The plugin must register its dependencies explicitly. No dead edge is created, and nothing pretends the module is covered.

The prerequisite: a virtual module id should never point to a path that exists on the filesystem. `\0`- and `virtual:`-prefixed ids satisfy this by construction. The rule only leans on the convention for path-shaped virtual ids, and the module check runs first, so a virtual base module always shadows a same-named file.

```mermaid
flowchart TD
    A["query module X?q<br/>cleanPath = X"] --> B{"X is a module<br/>in the graph?"}
    B -->|yes| C["invalidation edge<br/>X?q depends on module X"]
    B -->|no| D{"X exists<br/>on disk?"}
    D -->|yes| E["file dependency<br/>X?q depends on file X"]
    D -->|no| F["no edge<br/>plugin registers explicitly"]
    C --> G["a change to any file dep of module X<br/>invalidates X and then X?q transitively"]
```

### The cases

| case | example | `X` is a module | `X` on disk | derived dependency | result |
| --- | --- | --- | --- | --- | --- |
| A | `foo.vue?vue&type=script` | yes | yes | edge to module `foo.vue` | works, same as today's unbundled-dev |
| B | `/my-virtual.vue?vue&type=script` | yes (virtual) | no | edge to the virtual base module | works once the base registers `bar.vue` / `baz.vue` — better than unbundled-dev today, and the plugin registers them once on the base, not per sub-module |
| C | `foo.svg?url` | no | yes | file dependency on `foo.svg` | works, same as this PR today |
| D | `/foo?bar` reading `/home/project/a/foo/bar` | no | no | nothing | honest failure — explicit registration needed, and no wrong path is ever watched (the rollup/rollup#5013 case) |

sapphi-red: > 1. **If `X` is a module in the graph** — add an invalidation edge `X?q → module X`. The sub-module then follows whatever the base module depends on (its file, plus everything the base registered via `addWatchFile`). This is the "register the dependencies of the module without the query" behavior you described, derived automatically.
> 2. **Otherwise, if `X` exists on the filesystem** — add a plain file dependency `X?q → file X`. This covers ids like `foo.svg?url` where the base is never a module of its own.

The problem with these are that there's no way to remove these relationships. For example, even if the plugin knows that`/my-virtual.vue?vue&type=script` only depends on `bar.vue` part, the change in `baz.vue` will trigger update for `/my-virtual.vue?vue&type=script`.

h-a-n-a: AFAIK, removing the relationship is an optimization technique. Bundler should work even if this optimization is not available. In Vite today, editing `foo.vue` where only the script changed still invalidates `foo.vue`, `foo.vue?vue&type=script` and other sub-modules in that Vue component (by calling `getModulesByFile` and invalidates the whole `fileToModulesMap` set) and plugin-vue uses `hotUpdate` hook to pinpoint the fine-grained update to each module.

> even if the plugin knows that/my-virtual.vue?vue&type=script only depends on bar.vue part, the change in baz.vue will trigger update for /my-virtual.vue?vue&type=script.

This makes me think how Vue plugin knows that a module is depending only on one of the dependencies of its main module, which is the `my-virtual.vue`. 

I feel like it's something related to how the current Vue module diffs are handled, which is by diffing the descriptor between the prev build and the current one. This should be the same in this scenario where the change to `baz.vue` should be filtered with the `hotUpdate` hook so it will not rerun the sub-module? Otherwise, it will hurt the performance a little bit.

Out of the box, I think we should be as compatible as possible by default?

sapphi-red: > > even if the plugin knows that/my-virtual.vue?vue&type=script only depends on bar.vue part, the change in baz.vue will trigger update for /my-virtual.vue?vue&type=script.
> 
> This makes me think how Vue plugin knows that a module is depending only on one of the dependencies of its main module, which is the `my-virtual.vue`.

Ah, I think this example wasn't a good one. I think a better example would be `Foo.vue` & `Foo.vue?raw` and when `Foo.vue` has a `<script setup src="./foo.js">`. In this case, `Foo.vue` should rebuild for the change in `Foo.vue` and `foo.js`, while `Foo.vue?raw` should only rebuild for the change in `Foo.vue`.

> Out of the box, I think we should be as compatible as possible by default?

That's the purpose of `legacy.` flag. While I think it's ok to keep the `1. If X is a module in the graph` behavior (as long as it's documented), I think the behavior of `2.if X exists on the filesystem` should be removed. The plugin would still need to watch `X` or the `X` needs to be part of the module graph, as only the relationship is setup.

h-a-n-a: > In this case, Foo.vue should rebuild for the change in Foo.vue and foo.js, while Foo.vue?raw should only rebuild for the change in Foo.vue.

That makes sense. By default, changes to file `foo.js` would affect both `Foo.vue` module and `Foo.vue?raw` module (transitively dependent, as `X` is a module in the graph too), causing `Foo.vue?raw` to rebuild when it's not necessary. To make the update precise, we need to ignore `Foo.vue?raw` in `hotUpdate` hook if `ctx.file` is `foo.js`.

> I think the behavior of 2.if X exists on the filesystem should be removed. The plugin would still need to watch X or the X needs to be part of the module graph, as only the relationship is setup.

I was thinking about adding the `X` to the watch list (`transform_dependencies`) in this case so that the users does not need to setup the relationship again. Are you saying that this behavior should only be effective with the `legacy` flag?

h-a-n-a: _(The following content is written with AI help for structure and wording and the thoughts are mine)_

For context, `this.addWatchFile` in today's unbundled-dev Vite does much less than one would expect. It affects only a very small scope:

1. **It adds the file to chokidar** ([`pluginContainer.ts#L867-L873`](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/server/pluginContainer.ts#L867-L873)). Inside the project root this is a no-op, because the root is already watched recursively. It only matters for files outside the root.
2. **It records the file in a per-module set** (`ctx._addedImports`, stored in a private WeakMap, [`pluginContainer.ts#L182`](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/server/pluginContainer.ts#L182)). But nothing in the HMR pipeline reads this set. It creates no invalidation link.

So in unbundled dev, calling `this.addWatchFile(file)` alone does **not** invalidate the calling module when `file` changes. Invalidation comes from exactly one index: `fileToModulesMap`, keyed by `cleanUrl(id)` at insert time ([`moduleGraph.ts#L368`](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/server/moduleGraph.ts#L368)). A plugin that wants "this file feeds my module" has to wire it manually. The only plugin in Vite core that does this is the CSS plugin: it reads its own `_addedImports` and plants `createFileOnlyEntry(file)` nodes into `fileToModulesMap`, then registers them as imports ([`css.ts#L1275-L1294`](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/plugins/css.ts#L1275-L1294)). Every other plugin's `addWatchFile` call is effectively "make sure an event fires", nothing more.

This is why "the plugin should just call `this.addWatchFile`" was never the working contract in dev: in unbundled dev the call does almost nothing, so plugins had no reason to rely on it there — it only carried weight in build watch mode. Rolldown's `addWatchFile` is actually the stronger one: it registers a per-module `transform_dependencies` entry that the HMR stage consumes directly to expand the affected set, so the file→module half of the chain already works in Rolldown today. The missing half is only module→sub-module — which is what the derived edge adds.

### Comparison with my proposal

Splitting invalidation into its two halves — file→module ("which module does a changed file feed") and module→sub-module ("which query variants follow that module") — the three worlds look like this:

|  | unbundled-dev Vite | Rolldown today | Rolldown + proposal |
| --- | --- | --- | --- |
| file→module (`addWatchFile`) | inert — no invalidation link, CSS plugin hand-wires its own | works — per-module `transform_dependencies`, consumed by the HMR stage | same as today (unchanged) |
| module→sub-module (query variants) | works by accident — `cleanUrl(id)` keys all variants under one file | missing — exact-id match only, variants stay stale | works — derived edge `X?q → module X` (rule 1) |
| virtual base (`/my-virtual.vue` from `bar.vue`/`baz.vue`) | broken — `addWatchFile` links nothing, plugin must hand-write a `hotUpdate` hook or the sub-modules stay stale | half-works — base is invalidated via `transform_dependencies`, but its output is unchanged so the update collapses to no-op and the edit is lost | works — `bar.vue` → base (via `transform_dependencies`) → sub-modules (via the edge), with `addWatchFile` called once on the base |

sapphi-red: > To make the update precise, we need to ignore `Foo.vue?raw` in `hotUpdate` hook if `ctx.file` is `foo.js`.

For this the plugin that handles `?raw` would need to filter out the deps that was added for the plugin that handles `.vue`. That feels tricky to do, as the plugin would need to add `hotUpdate` (and `shouldTransformCachedModule` for watch mode?) for ideal behavior.

> I was thinking about adding the `X` to the watch list (`transform_dependencies`) in this case so that the users does not need to setup the relationship again. Are you saying that this behavior should only be effective with the `legacy` flag?

Even if it's in the watch list, the plugin would need to call `this.addWatchFile` so that the watcher watches that file. So I think that behavior alone won't help.

> 2\. **It records the file in a per-module set** (`ctx._addedImports`, stored in a private WeakMap, [`pluginContainer.ts#L182`](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/server/pluginContainer.ts#L182)). But nothing in the HMR pipeline reads this set. It creates no invalidation link.

IIRC this part creates the relationship in the module graph.
https://github.com/vitejs/vite/blob/7bdad37a4a772936d24e0116ac32a1f09fd8a055/packages/vite/src/node/plugins/importAnalysis.ts#L840-L874

h-a-n-a: > For this the plugin that handles ?raw would need to filter out the deps that was added for the plugin that handles .vue. That feels tricky to do, as the plugin would need to add hotUpdate (and shouldTransformCachedModule for watch mode?) for ideal behavior.

This does feel tricky to support. The possible way to do it seems to be for each plugin to return a map of dependencies, for example:

```js
transform_dependencies: { "self": ["Foo.vue"], "transitive": ["Foo.vue", "foo.js"] } // in Foo.vue metadata
``` 

But this makes a lot harder for plugin authors to maintain and it's not backward compatible. Or we can by default split the `self` and `transitive`, using the file to the module as `self` and anything else as `transitive`. so that the `?raw` plugin can leverage `self` for `this.addWatchFile` automatically.

sapphi-red: If we don't have this automatic relationship, it would be much simpler. The `?raw` plugin will not call anything and the `.vue` plugin will call the dependencies with `this.addWatchFile` and the virtual module plugin will call `this.addWatchFile` in the load hook. The `.vue?vue&type=script` can call `this.addWatchFile` by checking the values `.vue` plugin handled (it is because both of these are handled by the same plugin), or it can use `this.load` if we add the feature I suggested in https://github.com/rolldown/rolldown/pull/10714#discussion_r3820769224.

PR #10732, opened 2026-08-20
perf: deduplicate regress codegen

Boshen:
`regress` marks its public construction and matching entry points `#[inline]`, so each crate that calls them emits its own copy of the parser and backtracking executor. This change funnels the binding and replace plugin through `rolldown_utils::HybridRegex` and keeps construction plus iterator advancement behind `#[inline(never)]` boundaries, so those large bodies are emitted once instead of once per caller.

This reduces the macOS arm64 release binding by 129.25 KiB (0.78%).


hyfdev (approved):

hyfdev:
### Merge activity

* **Aug 20, 1:13 PM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Aug 20, 1:13 PM UTC**: `hyfdev` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Aug 20, 1:22 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10739, opened 2026-08-22
fix: treat module-scope `await using` as top-level await

MarshallOfSound:
A module-scope `await using` declaration awaits when the module body completes, but the AST scanner only records top-level await for `AwaitExpression` and `for await` — an `await using` has no `AwaitExpression` node, so `visit_variable_declaration` never flagged it.

```js
// dep.js
await using conn = open();
export { conn };
```

Before this change such a module was not treated as TLA: `format: 'cjs'`/`'iife'` emitted it without the "Top-level await is not supported" error, `require('./dep.js')` didn't report `REQUIRE_TLA`, and under `strictExecutionOrder` it was wrapped in a non-async `__esmMin(() => { await using … })` (a syntax error). Now `VariableDeclarationKind::AwaitUsing` at a valid TLA scope goes through `handle_top_level_await` like the other two forms.

Tests: `errors/tla_await_using_cjs` (format error), `errors/require_tla_await_using` (`REQUIRE_TLA`), `topics/tla/await_using` (executes; async init wrapper under both strict-execution-order variants). All three fail without the change.

AI-assisted (Claude Code); reviewed and tested by me.

### DIFF
--- a/crates/rolldown/src/ast_scanner/impl_visit.rs
+++ b/crates/rolldown/src/ast_scanner/impl_visit.rs
@@ -335,6 +335,11 @@ impl<'me, 'ast: 'me> VisitJs<'ast> for AstScanner<'me, 'ast> {
   }
 
   fn visit_variable_declaration(&mut self, decl: &ast::VariableDeclaration<'ast>) {
+    // `await using x = …` awaits when the enclosing scope is exited, without any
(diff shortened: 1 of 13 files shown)


shulaoda (approved):
LGTM. Thanks for the contribution!

PR #10740, opened 2026-08-22
perf: evaluate codeSplitting `test`/`name` functions before the assignment loop

MarshallOfSound:
Function-valued `codeSplitting.groups[].test` / `.name` are JS callbacks, and `build_module_groups` awaited them one (module × group) pair at a time inside the assignment loop — one main-thread round trip each. On a large app with a few function groups that is ~10⁵ sequential hops between link and chunk rendering.

This evaluates them before the loop and caches the results: calls are dispatched in bounded batches, one callback function at a time (every module for group 0's `test`, then group 1's, …, then the `name` callbacks likewise), each in the existing `stable_id` order. A given JS function's queued calls run in dispatch order, so a stateful callback still sees a deterministic sequence; different functions are never in flight together. The set of calls, group membership and output are unchanged.

Synthetic project (20k leaf modules shared across 200 dynamic entries, 6 groups with function `test`+`name`, 95,545 callback invocations), second `generate()` on a warm graph, output byte-identical:

| | `generate()` |
|---|---|
| main | 2.30 s |
| this PR | 0.44 s |

On the real app this came from (~20k modules, 3 function groups) it removes ~3 s from a production build.

<details><summary>fixture</summary>

```js
// gen.mjs — writes src/ and entry.js
import fs from 'node:fs';
const N = 20000, E = 200; fs.mkdirSync('src', { recursive: true }); let entry = '';
for (let i = 0; i < N; i++) fs.writeFileSync(`src/m${i}.js`, `export const v${i} = ${i};\n`);
for (let e = 0; e < E; e++) { let s = ''; for (let k = 0; k < 300; k++) { const i = (e*7919 + k*104729) % N; s += `export { v${i} } from "./m${i}.js";\n`; } fs.writeFileSync(`src/e${e}.js`, s); entry += `export const p${e} = () => import("./src/e${e}.js");\n`; }
fs.writeFileSync('entry.js', entry);
// bench.mjs
const groups = Array.from({ length: 6 }, (_, g) => ({ name: (id) => 'grp' + g + '-' + (id.length % 7), test: (id) => id.endsWith(g + '.js'), priority: g }));
const b = await rolldown({ input: './entry.js' }); await b.generate({});
console.time('generate'); await b.generate({ codeSplitting: { groups } }); console.timeEnd('generate');
```
</details>

AI-assisted (Claude Code); reviewed and benchmarked by me.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/manual_code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/manual_code_splitting.rs
@@ -115,6 +115,8 @@ impl ManualSplitter<'_> {
   async fn build_module_groups(
     &self,
   ) -> BuildResult<(IndexVec<ModuleGroupIdx, ModuleGroup>, Vec<ModuleGroup>)> {
+    /// Max JS callback invocations in flight at once.
+    const CALLBACK_BATCH: usize = 4096;
     let metas = &self.link_output.metas;
     let mut module_groups: IndexVec<ModuleGroupIdx, ModuleGroup> = IndexVec::default();
     let mut group_idx_by_id: FxHashMap<ModuleGroupId, ModuleGroupIdx> = FxHashMap::default();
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
Superseded by #10745

PR #10743, opened 2026-08-22
fix: model a facade's host import in the fold cycle check

### LINKED ISSUE #10734: [Bug]: Dynamic entry facade chunk exports an undeclared namespace binding, producing invalid ESM

```bash
git clone https://github.com/ulrichstark/rolldown-undeclared-namespace-export-repro
cd rolldown-undeclared-namespace-export-repro
npm install && npm run repro
```

### What is expected?

Every emitted chunk is valid ESM: each name in a chunk's `export {}` list is declared or imported within that chunk.

The build exits 0 and emits no warnings, but the facade chunk for the dynamic entry is not valid ESM. It fails `node --check` on its own, with no linking or execution involved:

```console
$ node --check out/service-B4l0TYVy.js
export { entryValue, shared as n, chainA as r, routes, service_exports as t, useChainB, useShared };
                                                       ^^^^^^^^^^^^^^^^^^^^
SyntaxError: Export 'service_exports' is not defined in module
```

This is an ES early error, not a resolution or interop issue: every name in a module's `ExportedBindings` must appear in its `VarDeclaredNames` or `LexicallyDeclaredNames`. `service_exports` is never declared or imported — the identifier occurs exactly once in the entire output, in that export list.

Here is the whole chunk:

```js
import "./rolldown-runtime-D7D4PA-g.js";
import { i as useShared, n as routes, r as useChainB, t as entryValue } from "./service-Z1Wy1UDZ.js";
//#endregion
//#region src/chain-a.mjs
const chainA = "chainAleafAlib";
//#endregion
//#region src/shared.mjs
const shared = "shared" + chainA;
//#endregion
export { entryValue, shared as n, chainA as r, routes, service_exports as t, useChainB, useShared };
```

Three details suggest the export entry and the runtime dependency were registered while the declaration statement itself was dropped:

- `src/service.mjs` is split across two chunks, both named `service-*`: `service-B4l0TYVy.js` (the facade above) and `service-Z1Wy1UDZ.js` (the module body).
- The first line is a **bare, unbound** side-effect import of the runtime chunk, whose only export is `__exportAll as t`. The dependency edge on the helper survived; the binding and the declaration did not.
- The chunk opens with an orphaned `//#endregion` that has no matching `//#region`.

The generated entry expects that namespace to exist:

```js
// out/index.mjs
const loadService = () => import("./service-B4l0TYVy.js").then((n) => n.t);
```

So `t` is meant to be the namespace object for `src/service.mjs`, and nothing in the chunk defines it. Importing the entry therefore throws before any user code runs:

```
SyntaxError: Export 'service_exports' is not defined in module
```

Because nothing fails at build time, for a server-shaped bundle the first symptom is the first request.

### System Info

```Shell
System:
    OS: macOS 26.6.2
    CPU: (18) arm64 Apple M5 Pro
    Memory: 191.75 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 26.1.0 - /Users/ulrichstark/.nvm/versions/node/v26.1.0/bin/node
    npm: 11.13.0 - /Users/ulrichstark/.nvm/versions/node/v26.1.0/bin/npm
  Browsers:
    Chrome: 151.0.7922.170
    Safari: 26.6.2
  npmPackages:
    rolldown: 1.2.5 => 1.2.5
```

### Downstream reports

Found via a TanStack Start + Nitro production build, where the same defect surfaces as `Export 'ssr_exports' is not defined in module` and 500s every route on a build that exited 0:

- nitrojs/nitro#4533 — *Production build emits invalid `ssr_exports` reference in bundled SSR chunk (TanStack Start, any preset)*
- TanStack/router#8031 — *Start + Nitro on Vite 8.2: SSR service chunk re-exports an undeclared namespace (`ssr_exports`) → every request 500s while `vite build` exits 0*

Both are open. The affected chunk there is `.output/server/_ssr/ssr.mjs`, with an export list of the same shape:

```js
export { getResponseHeaders as a, ssr_exports as c, createServerEntry, server_default as default, ... }
```

Vite 8.2.x ships Rolldown 1.2.2+, which is how this reaches Start/Nitro users.

### When it started

Introduced in **1.2.2**. `npm run matrix` in the repro builds the same input against eight versions:

| rolldown | chunks | result |
|---|---|---|
| 1.1.0, 1.1.5, 1.2.0, 1.2.1 | 5 | clean |
| 1.2.2, 1.2.3, 1.2.4, 1.2.5 | 7 | `Export 'service_exports' is not defined in module` |

The chunk count changing from 5 to 7 at the same boundary suggests the regression is in chunk splitting rather than in codegen alone. 1.2.5 is the current `latest` on npm, so there is no released version with the fix.

ulrichstark:
`optimize_dynamic_entry_bits` can clear a dynamic entry's own bit from the atom holding its entry module, leaving that entry chunk as a facade whose module lives in a shared chunk. `compute_chunk_imports` makes the facade import the host chunk — to run the module and re-export its interface — but no module inside the facade imports the entry module, so the atom graph had no way to express the edge.

The fold's cycle check therefore treated a facade as having zero static in-degree and merged atoms into it that the host chunk statically depends on, emitting a chunk cycle that evaluates in the wrong order. Importing the dynamic entry failed before any user code ran:

    SyntaxError: Export 'service_exports' is not defined in module

`reduced_atom_graph_has_static_cycle` now adds the matching edge for every entry bit whose entry module is hosted elsewhere. The `cycle` graph only ever over-approximates, so a redundant edge costs at most a missed merge. That edge was already part of the prediction
`debug_assert_module_level_static_import_prediction` reconciles against emission; only the fold's own graph was missing it.

This also re-grounds the hosting-atom attribution of entry-export service edges, which used to rest on facades having zero static in-degree: a facade now depends on its host, so every service edge emission hangs on the facade stays reachable as facade -> host -> target.

Fixes #10734


hyfdev:
LGTM. It fixed https://github.com/rolldown/rolldown/issues/10734.

hyfdev (approved):

hyfdev:
Thank you!

PR #10745, opened 2026-08-23
perf: batch `codeSplitting` group `test` / `name` calls

IWANABETHATGUY:
Function-valued `codeSplitting.groups[].test` and `.name` are JS callbacks, and each call crossed the napi boundary on its own, so a build with M modules and G function groups paid M * G crossings before it rendered a chunk.

The JS side now hands the binding one shim per callback that takes the whole id array and loops in JS, so a group crosses at most twice. A group's calls are contiguous now instead of interleaved with the other groups', and a group's whole `test` pass runs before its first `name` call, but each callback still sees its own ids in `stable_id` order.

<details><summary>benchmark</summary>

20k leaf modules shared across 200 dynamic entries, 6 groups with function `test` + `name`. Best of 11 `generate()` calls, release builds of all three, byte-identical output.

| | `generate()` | the 86,856 `test` calls alone |
|---|---|---|
| main | 1243 ms | 803 ms |
| #10740 | 412 ms | 57 ms |
| this PR | 361 ms | 18 ms |

The second column comes from timing the same app again with the six `test` functions replaced by regexes that match identically. The output stays byte-identical, so the only difference is that those calls never leave Rust. #10740 overlaps one crossing per call with `try_join_all`, which leaves 656 ns each; batching the payload leaves 207 ns.

What is left is mostly not the user's code — replaying those same 86,856 calls in plain JS costs 0.7 ms, so ~95% of the 18 ms is marshalling one id string per module per group. Sending the id list once for all groups would cut that further.

Benchmark produced with Claude Code.

</details>

This and #10740 solve the same problem and cannot both land.

### DIFF
--- a/crates/rolldown/src/stages/generate_stage/manual_code_splitting.rs
+++ b/crates/rolldown/src/stages/generate_stage/manual_code_splitting.rs
@@ -1,4 +1,5 @@
 use std::{
+  borrow::Cow,
   cmp::{Ordering, Reverse},
   collections::BinaryHeap,
   path::Path,
@@ -10,7 +11,8 @@ use itertools::Itertools;
 use oxc_index::IndexVec;
 use rolldown_common::{
   Chunk, ChunkKind, ChunkingContext, EntryPoint, ManualCodeSplittingOptions, MatchGroup,
-  MatchGroupTest, Module, ModuleIdx, ModuleTable, ModuleTagBitSet, ModuleTagRegistry, NormalModule,
+  MatchGroupName, MatchGroupTest, Module, ModuleIdx, ModuleTable, ModuleTagBitSet,
+  ModuleTagRegistry, NormalModule,
 };
 use rolldown_error::BuildResult;
 use rolldown_plugin::SharedPluginDriver;
@@ -112,8 +114,9 @@ impl ManualSplitter<'_> {
     Ok(())
   }
 
-  /// Whether `module` clears the group's module-level constraints: required tags, the module size
-  /// bounds, and the share count.
+  /// Runs the checks that sit between a group's `test` call and its `name` call.
+  ///
+  /// They read data that only Rust holds. That is why `test` and `name` need two calls.
   fn passes_static_filters(
     &self,
     match_group_index: usize,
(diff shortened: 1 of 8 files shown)


crates/rolldown/src/stages/generate_stage/manual_code_splitting.rs
```diff
@@ -178,33 +181,70 @@ impl ManualSplitter<'_> {
...
     let ctx = ChunkingContext::new(Arc::clone(&self.plugin_driver.module_infos));
 
-    for normal_module in sorted_normal_modules {
-      for (match_group_index, match_group) in self.match_groups.iter().copied().enumerate() {
-        let is_matched = match &match_group.test {
-          None => true,
-          Some(MatchGroupTest::Regex(reg)) => reg.matches(&normal_module.id),
-          Some(MatchGroupTest::Function(func)) => {
-            func(&normal_module.id).await?.unwrap_or_default()
+    for (match_group_index, match_group) in self.match_groups.iter().copied().enumerate() {
```
hyfdev, line 196: The batching direction makes sense, but this loop inversion changes observable behavior for valid configurations.

The following findings were summarized by AI, but I have reviewed them and agree with the AI's assessment.

Before this PR, callbacks run module-major:

`test(a) → name(a) → test(b) → name(b)`

After this PR, they run test-batch-first:

`test(a) → test(b) → name(a) → name(b)`

A stateful but well-typed pair such as `test(id) { current = id; return true }` and `name() { return current }` therefore changes its chunk assignments: stable Rolldown can produce separate names for `a` and `b`, while this branch assigns both modules the last name.

The new documentation accurately describes the changed order, but it is migration guidance rather than compatibility. Since `main` must remain compatible with the latest stable release, could we either preserve the old observable call order while batching, or first get an explicit decision that this callback interleaving is allowed to change?

If the new order is accepted deliberately, please also add a fixture that asserts both the callback sequence and the resulting group assignments. Currently no fixture combines function-valued `test` and `name`.

hyfdev: While AI does make sense, I personally don't want to give such a promise to the contract, even there might be a way that we could do the batching and keep the order still.

hyfdev (approved):
LGTM other than https://github.com/rolldown/rolldown/pull/10745#discussion_r3841806069 needs quick conclusion.

crates/rolldown/src/stages/generate_stage/manual_code_splitting.rs
hyfdev: Conclusion: Let's ignore this behavior change, as we don't find users rely on this order behaviors. Will reopen this, if there are users report related issues.

PR #10753, opened 2026-08-24
fix: name the module when a `codeSplitting` group callback returns a wrong type

IWANABETHATGUY:
One call classifies or names a whole batch, so napi validates the array and reports its type rather than the offending element's. A plain JS config that returned a string from `test` produced `The function returned object, but expected object`.

Both shims now check each result before the batch crosses, so the error names the module and the type it expected. `test` crosses as a byte array, which also drops a per-element conversion. `name` still crosses as strings.

refs #10745

### DIFF
--- a/packages/rolldown/src/utils/bindingify-output-options.ts
+++ b/packages/rolldown/src/utils/bindingify-output-options.ts
@@ -404,14 +404,23 @@ function bindingifyCodeSplitting(
  * Wraps a per-id `test` in the batched shim that the binding expects.
  *
  * The loop runs in JS so that a group makes one napi crossing, not one per module.
+ *
+ * The result is a `Uint8Array`, which crosses as a buffer instead of one tagged value per id.
  */
-function batchTest(
-  test: CodeSplittingTestFunction,
-): (ids: string[]) => ReturnType<CodeSplittingTestFunction>[] {
+function batchTest(test: CodeSplittingTestFunction): (ids: string[]) => Uint8Array {
   return (ids) => {
-    const results: ReturnType<CodeSplittingTestFunction>[] = [];
+    const results = new Uint8Array(ids.length);
     for (let index = 0; index < ids.length; index++) {
-      results.push(test(ids[index]));
+      const result = test(ids[index]);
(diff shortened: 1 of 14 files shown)


packages/rolldown/src/utils/bindingify-output-options.ts
```diff
@@ -404,14 +404,16 @@ function bindingifyCodeSplitting(
...
-function batchTest(
-  test: CodeSplittingTestFunction,
-): (ids: string[]) => ReturnType<CodeSplittingTestFunction>[] {
+function batchTest(test: CodeSplittingTestFunction): (ids: string[]) => Uint8Array {
   return (ids) => {
-    const results: ReturnType<CodeSplittingTestFunction>[] = [];
+    const results = new Uint8Array(ids.length);
     for (let index = 0; index < ids.length; index++) {
-      results.push(test(ids[index]));
+      results[index] = test(ids[index]) ? 1 : 0;
```
hyfdev: Could we preserve runtime type validation here instead of coercing every result by truthiness?

The following findings were summarized by AI, but I have reviewed them and agree with the AI's assessment.

`CodeSplittingTestFunction` only permits `boolean | undefined | void`, and stable Rolldown rejects other values. With this code, strings, numbers, objects, and even `Promise.resolve(false)` are accepted; the Promise is truthy, so it captures the module. A rejected Promise can also let the build finish and only surface later as an unhandled rejection. This can turn a configuration mistake into silently incorrect chunks.

We can keep the `Uint8Array` transport while validating each result and identifying its module:

```ts
const result = test(ids[index]);
if (result != null && typeof result !== 'boolean') {
  throw new TypeError(/* include ids[index] */);
}
results[index] = result === true ? 1 : 0;
```

The current non-boolean acceptance fixture should then become an error fixture that asserts both the invalid type and the offending module id.

hyfdev:
Overall. LGTM

hyfdev (approved):

packages/rolldown/src/utils/bindingify-output-options.ts
IWANABETHATGUY: Added the check and turned the fixture into an error case. 👍

PR #10797, opened 2026-08-31
fix: derive asset reference ids from content hash for deterministic builds

MarshallOfSound:
### Problem

Assets emitted without an explicit `fileName` get `reference_id = xxhash(<emission counter>)`. The counter's value depends on cross-thread module processing order, so identical builds produce different reference ids. The ids appear in pre-render chunk code (e.g. Vite's `__VITE_ASSET__<id>__` placeholders), which makes source maps (`sourcesContent` keeps the tokens) and anything hashed over pre-render code (e.g. Sentry's injected debug ids, which then change final content-hashed chunk filenames) nondeterministic across identical builds.

### Fix

Remove the emission counter; `assign_reference_id` now requires a seed:

- deduplicated assets seed from their content hash (already computed for the dedupe map, which is keyed on it, so ids stay unique per asset);
- chunks seed from `name`, falling back to the module `id` when unnamed — duplicate unnamed emits of one id then share a reference id, matching how duplicate named emits already behave;
- explicit-`fileName` files and prebuilt chunks seed from the file name, as before.

Id format is unchanged. Bundle output is unchanged — only ids become reproducible, as the updated snapshot shows (the `__ROLLDOWN_ASSET__#<id>` token changes, the rendered URL beside it does not).

### Verification

- New test: `asset_reference_ids_are_independent_of_emission_order`.
- Downstream check: a large Vite 8 app that renamed ~35 chunks per identical-source CI build (via Sentry debug ids over asset placeholders) now builds byte-for-byte reproducibly, source maps included.

### DIFF
--- a/crates/rolldown/tests/rolldown/function/module_types/asset/artifacts.snap
+++ b/crates/rolldown/tests/rolldown/function/module_types/asset/artifacts.snap
@@ -35,7 +35,7 @@ export { main_default as default };
 - ../foo.txt
 (0:0) "module." --> (27:1) "module."
 (0:7) "exports = " --> (27:8) "exports = "
-(0:17) "\"__ROLLDOWN_ASSET__#VRAku6fjghkApIISiBWPzg\"" --> (27:18) "\"./assets/foo-Cctl21F7.txt\";\n"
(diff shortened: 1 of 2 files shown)


yhx-12243:
Not only this one, https://github.com/rolldown/rolldown/blob/main/crates/rolldown_common/src/file_emitter.rs#L145 will also cause this issue, for example, `chunk.name` can be None.

PR #10799, opened 2026-08-31
fix: synchronize sourcemap assets after generateBundle mutations

### LINKED ISSUE #10676: generateBundle: `chunk.map` mutations never reach the emitted .map asset (`chunk.code` mutations do)
rollup lets a `generateBundle` hook mutate `chunk.code` and `chunk.map` together, and writes what you mutated. rolldown applies the `code` mutation but the emitted `.map` **asset** keeps the old mappings — the map file is materialized before `generateBundle` runs (`render_chunk_to_assets.rs` pushes it as an `Output::Asset`), and `apply_changes` / `update_output_chunk` only update the chunk's own `map` field, never that asset.

so any plugin that prepends a line to a chunk and compensates the map ships a map where every mapping is one line early. that's not hypothetical: `@tsdown/css`'s `css.inject` does exactly this, which misaligns the sourcemap of every css-injecting library built with tsdown (fix on that side incoming).

## repro

rolldown 1.2.4 (latest):

```js
import { writeFileSync } from 'node:fs';

writeFileSync('entry.js', 'export const value = 42;\n');

  input: './entry.js',
  plugins: [
    {
      name: 'mutator',
      generateBundle(_options, bundle) {
        for (const chunk of Object.values(bundle)) {
          if (chunk.type !== 'chunk') continue;
          // prepend one line, compensate the map (rollup semantics)
          chunk.code = `import './style.css';\n${chunk.code}`;
          if (chunk.map) chunk.map.mappings = `;${chunk.map.mappings}`;
        }
      },
    },
  ],
});

const { output } = await bundle.generate({ format: 'es', sourcemap: true });
const chunk = output.find((o) => o.type === 'chunk');
const mapAsset = output.find((o) => o.fileName.endsWith('.map'));

console.log('chunk.code starts with :', JSON.stringify(chunk.code.split('\n')[0]));
console.log('chunk.map mappings     :', chunk.map.mappings);
console.log('emitted .map mappings  :', JSON.parse(mapAsset.source).mappings);
```

```
chunk.code starts with : "import './style.css';"
chunk.map mappings     : ;;AAAA,MAAa,QAAQ
emitted .map mappings  : ;AAAA,MAAa,QAAQ
```

the mutated chunk.map and the emitted `.map` asset disagree; the written file is the stale one.

## related wrinkle

mutating *only* `chunk.map` (without also assigning `code`) doesn't even mark the chunk as changed — the mutable-chunk proxy's set trap never fires for nested mutation (`chunk.map` is a cached getter result), so `collectChangedBundle` skips the chunk entirely.

## suggested fix

`apply_changes` knows the chunk's `sourcemap_filename`, so it could refresh the sibling sourcemap asset when a change carries a `map`. happy to attempt a PR if that direction sounds right — or if the stance is "mutate in renderChunk instead" (per #10636), a loud warning when a changed chunk has a diverging map asset would at least stop the silent misalignment.

---

*Disclosure: this investigation and write-up are AI-assisted exploration (Claude Fable 5 via Claude Code), directed and reviewed by me.*

sonofmagic:
`generateBundle` can mutate both `chunk.code` and `chunk.map`. Rolldown applies those chunk changes, but the materialized `.map` asset is created before the hook and can retain stale mappings.

This PR keeps sourcemap assets synchronized with successfully updated chunks in the native binding layer.

## What changed

- Refresh the asset identified by the chunk's stable `sourcemapFileName` after a chunk mutation.
- Keep explicit asset updates, renames, and deletions authoritative over automatic synchronization.
- Never recreate an asset that a plugin deleted.
- Lazily allocate tracking state and skip the synchronization pass when no changed chunk has a sourcemap.

The JavaScript bridge still transfers complete changed chunk snapshots. Deep tracking of nested `chunk.map` mutations is intentionally out of scope for this focused fix.

Fixes #10676

Reproduction: https://github.com/icebreaker-demo/rolldown-generate-bundle-sourcemap-repro

## Validation

- `cargo check --ignore-rust-version -p rolldown_binding`
- `cargo fmt --all -- --check`
- `pnpm exec vp lint packages/rolldown/tests/fixtures/plugin/generate-bundle-sourcemap-asset-sync/_config.ts packages/rolldown/tests/fixtures/plugin/generate-bundle-sourcemap-asset-sync-no-map/_config.ts`
- `pnpm exec vitest run fixtures-concurrent.test.ts -t 'plugin/generate-bundle' --reporter verbose --hideSkippedTests`

Codspeed is used to detect regressions in the native output-application path. The fix adds no extra JS-to-Rust calls and avoids allocating or traversing synchronization state when it is not needed.

AI disclosure: AI tools were used to assist with repository exploration and drafting. I reviewed and verified the implementation and tests.

### DIFF
--- a/crates/rolldown_binding/src/types/binding_outputs.rs
+++ b/crates/rolldown_binding/src/types/binding_outputs.rs
@@ -73,8 +73,8 @@ impl JsChangedOutputs {
     outputs: &mut Vec<rolldown_common::Output>,
   ) -> anyhow::Result<()> {
     let mut result = Ok(());
-    let mut sourcemap_updates = HashMap::<String, String, FxBuildHasher>::default();
-    let mut explicitly_updated_assets = HashSet::<String, FxBuildHasher>::default();
+    let mut sourcemap_updates = None;
+    let mut explicitly_updated_assets = None;
     if !self.deleted.is_empty() || !self.changes.is_empty() {
       outputs.retain_mut(|output| {
         if result.is_err() {
@@ -91,11 +91,15 @@ impl JsChangedOutputs {
               } else if let (Some(filename), Some(map)) =
                 (&old_chunk.sourcemap_filename, &old_chunk.map)
               {
-                sourcemap_updates.insert(filename.clone(), map.to_json_string());
+                sourcemap_updates
+                  .get_or_insert_with(HashMap::default)
+                  .insert(filename.clone(), map.to_json_string());
               }
             }
             (v @ rolldown_common::Output::Asset(_), Either::B(asset)) => {
-              explicitly_updated_assets.insert(asset.filename.clone());
+              explicitly_updated_assets
+                .get_or_insert_with(HashSet::default)
+                .insert(asset.filename.clone());
               *v = rolldown_common::Output::Asset(Arc::new(asset.into()));
             }
             _ => {}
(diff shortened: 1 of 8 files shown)


hyfdev:
IMO, I don't think it's a bug, but let's discuss in https://github.com/rolldown/rolldown/issues/10676#issuecomment-5503945020

PR #10809, opened 2026-09-01
chore(deps): upgrade oxc to 0.148.0

### LINKED ISSUE #10722: [Bug]: DCE removes a variable that is still read from a default parameter value

Full reproduction and description: **_https://github.com/AleksandrIlinykh/rolldown-dce-repro_**

Run:
```sh
npm install
npm run build && npm run check   # exits 1 when the bug is there
npm run cases                    # every case, broken and correct, in one table
```

The first case is two modules and needs nothing but rolldown:

```js
// trigger.js
export let n = 1
export function bump() { n += 1 }

// victim.js
export function factory(get) {
  const read = function (a = n.IS_BLOCK) { return a }
  var n = get(81937)
  return read
}

// entry.js
import { n, bump } from './trigger.js'
import { factory } from './victim.js'
```

### What is expected?

`var n = get(81937)` is kept, and `n.IS_BLOCK` resolves to that local variable, the way it does
before bundling.

The read inside the default parameter value resolves to `n` from `trigger.js`. The local `n` is then
seen as unused and its declaration is removed, while the read stays:

```js
function factory(get) {
	const read = function(a = n.IS_BLOCK) {   // now reads the other n, which holds 1
		return a;
	};
	get(81937);                               // "var n =" is gone
	return read;
}
```

Four things have to come together: another module declares the same name, that module comes first in
the import order, the variable is declared below the function that reads it, and the read is a
default parameter value. Take any one away and the output is correct.

Each of them is a directory in the repository with its own modules, entry and build result, and the
name says what it produces: `01-broken-baseline`, `02-broken-let-instead-of-var`,
`03-broken-variable-also-used-normally`, `04-broken-local-function-declaration`,
`05-broken-notistack-and-jodit`, `06-correct-imports-swapped`,
`07-correct-trigger-variable-renamed`, `08-correct-victim-variable-renamed`,
`09-correct-var-declared-before-reader`, `10-correct-read-in-function-body`,
`11-correct-no-trigger-module`.

Two of them are worth a look. Case 03 keeps the variable in use by returning it, so it cannot
be called unused, and the declaration still disappears, which suggests the removal is a consequence
of the wrong resolution rather than the cause. Case 04 uses a local function declaration instead of a
variable and is checked by running the output: the built code returns `"THE OTHER ONE"` where the
same source without a bundler returns `"the local one"`.

| `output.minify` | small case | real packages |
| --------------- | ---------- | ------------- |
| `'dce-only'` | broken | broken |
| `true` | correct | broken |
| `false` | correct | correct |

With `true` the mangler renames bindings, and in the small case that happens to separate the two
names again. In a large chunk it does not.

Through Vite this means `build.minify: 'oxc'` (the default) and `build.minify: false` are both
affected, since Vite maps the second one to `'dce-only'`. Only `'esbuild'` and `'terser'` map to
`output.minify: false`.

### System Info

```Shell
System:
    OS: Windows 11 10.0.26200
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11850H @ 2.50GHz
  Binaries:
    Node: 24.14.0
    npm: 11.9.0
  npmPackages:
    rolldown: 1.2.5 => 1.2.5


Also reproduced on 1.1.5.
```

### Any additional comments?

Found in a production application and hard to trace. The damaged module was `getBlockWrapper` from
`jodit`, the first thing an editor's Enter handler calls. Jodit swallows the exception in its own
event system and the handler has already cancelled the browser default by then, so Enter did nothing
at all and the console stayed empty.

#9204 looks like the same defect. It was closed by the reporter without a reproduction, as one
library colliding with a second copy of itself. Case 04 is that shape and reproduces on the current
release.

Not reproduced by `minifySync` from `rolldown/utils` on the same input, by the vendor file alone in
any minify mode, or by putting it next to unrelated packages in a large chunk. It needs the name
collision.

Best regards.

### LINKED ISSUE #10779: [Bug]: transform.define treats a function-body var as visible in a default parameter

Repository: https://github.com/why-reproductions-are-required/vite-dead-dynamic-import-repro

Run the isolated Rolldown reproduction:

```sh
git clone https://github.com/why-reproductions-are-required/vite-dead-dynamic-import-repro.git
cd vite-dead-dynamic-import-repro
pnpm install --frozen-lockfile
pnpm dlx rolldown@1.2.5 src/define-only.js --platform browser --dir /tmp/rolldown-define --transform.define 'typeof window:"undefined"'
```

The source uses a global `window` reference in a default parameter. The function body has a separate `var window` binding:

```js
export function load(
  value = typeof window !== "undefined" ? import("browser-only-package/browser") : null,
) {
  var window;
  return value;
}
```

### Version change that exposed the failure

The vinext integration test has this package boundary:

| Vite+ | Vite | Rolldown | Result |
| --- | --- | --- | --- |
| `0.2.7` | `8.1.5` | `1.2.0` | Pass |
| `0.2.8` | `8.2.0` | `1.2.2` | Fail |

Direct Rolldown builds narrow the output change to `1.2.0` versus `1.2.1`:

| Rolldown | Output for `src/server.js` |
| --- | --- |
| `1.2.0` | Removes `load()` and the dynamic import |
| `1.2.1` through `1.2.5` | Keeps `load()` and emits a browser chunk |

Rolldown `1.2.0` produces:

```js
const value = "server";
export { value };
```

Rolldown `1.2.1` produces:

```js
function load(value = typeof window !== "undefined" ? import("./browser-eoGTUZR-.js") : null) {
  var window;
  return value;
}
load();
const value = "server";
export { value };
```

Two behaviors cause this difference:

1. Rolldown `1.2.0` removes `load()` because Oxc classifies the call as side-effect-free. [Oxc PR #24791](https://github.com/oxc-project/oxc/pull/24791) corrected dead-code elimination for default-parameter side effects. [Rolldown PR #10497](https://github.com/rolldown/rolldown/pull/10497) updated Oxc to `0.142.0`, which contains that correction. Rolldown `1.2.1` includes the update and keeps the call.
2. The retained call exposes a separate native `define` scope problem. Rolldown does not replace `typeof window` because the function body declares `var window`.

The Oxc change is not the native `define` regression. The same `define` behavior occurs with Rolldown `1.2.0` when an export keeps the function in the bundle. The `src/define-only.js` reproduction isolates that behavior.

### What is expected?

The `define` transform replaces `typeof window` with `"undefined"`. Rolldown removes the false branch and does not emit a browser chunk.

JavaScript evaluates the default parameter outside the function body environment. The body declaration does not bind the `window` reference in the parameter. The ECMAScript algorithm describes the separate environments: https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-functiondeclarationinstantiation

Rolldown `1.2.5` leaves `typeof window` unchanged and emits a browser chunk. The server output contains:

```js
function load(value = typeof window !== "undefined" ? import("./browser-eoGTUZR-.js") : null) {
  var window;
  return value;
}
```

The transform appears to treat the function-body `var window` declaration as visible in the default parameter.

### System Info

```Shell
System:
  OS: macOS 26.2
  CPU: (16) arm64 Apple M4 Max
  Memory: 1.03 GB / 48.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 24.19.0
  npm: 11.17.0
  pnpm: 10.30.1
Browsers:
  Chrome: 151.0.7922.174
  Firefox: 151.0.2
  Safari: 26.2
Rolldown: 1.2.5
```

### Any additional comments?

Can a maintainer confirm whether `transform.define` should follow the separate default-parameter scope here?

camc314:
## Summary

- Upgrade Oxc Rust dependencies to 0.148.0
- Upgrade @oxc-project/runtime, @oxc-project/types, oxc-parser, and oxc-transform JavaScript packages to 0.148.0.

### oxc-project/oxc#26099

Oxc now resolves default-parameter references before visiting function-body bindings. Rolldown can apply global define replacements in default parameters correctly, so issue 10779 no longer expects execution failure.

The associated snapshot removes six unused body-local declarations. This matches JavaScript parameter-scope semantics.

The same fix resolves #10722: on the merge base, its reproduction binds the default-parameter read to another module's identically named export and drops the still-read local `var` declaration; on this branch the declaration is kept and the bundle executes correctly.

Commit: ac404dededadb65e5b8544d22edfbea53be3f787

### oxc-project/oxc#26172

Oxc now removes switch cases whose conditions are statically unreachable. The affected fixture places unresolved require calls inside dead cases; those cases are eliminated before Rolldown resolves their imports.

The snapshot drops 195 lines covering 15 UNRESOLVED_IMPORT diagnostics from unreachable known-value, default, continue, return, and fallthrough cases. The remaining FAIL-x diagnostic is reachable and remains unchanged.

Commit: 36a4bee93bc54481cb54168a4351fdd26a368695

fixes #10779
fixes #10722


crates/rolldown/tests/esbuild/dce/dead_code_inside_unused_cases/artifacts.snap
```diff
@@ -5,201 +5,6 @@ source: crates/rolldown_testing/src/integration_test.rs
 
 ## UNRESOLVED_IMPORT
 
-```text
```
camc314, line 8: this feels strange to me that DCE/minification happens before resolving imports 🤷

Dunqing:
@codex link the issues fixed by this upgrade in the PR description.

crates/rolldown/tests/esbuild/dce/dead_code_inside_unused_cases/artifacts.snap
Dunqing: Looks fine to me. Since these `require` calls can never be reached, reporting resolution errors for them would only make the build fail unnecessarily.

Dunqing (approved):

IWANABETHATGUY (approved):

crates/rolldown/tests/esbuild/dce/dead_code_inside_unused_cases/artifacts.snap
hyfdev: Expected behavior.

PR #10821, opened 2026-09-04
feat(docs): add wrangler

mdong1909:


docs/public/_redirects
sapphi-red, line 10: We need to port the redirect here:
https://github.com/rolldown/rolldown/blob/26b4c6e56c5553d72a910d0f73d60fa331c0ed88/docs/void.json#L7-L92
or
https://github.com/rolldown/rolldown/blob/26b4c6e56c5553d72a910d0f73d60fa331c0ed88/docs/netlify.toml#L14-L170
(both should have the same behavior)

docs/package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "dev": "vitepress dev",
     "build": "vitepress build",
+    "build:cloudflare": "node .vitepress/scripts/generate-reference.ts && vitepress build",
```
sapphi-red:
```suggestion
    "build:cloudflare": "pnpm generate && pnpm build",
```

mdong1909: done!

docs/public/_redirects
mdong1909: Thanks for catching this. updated and did spot checks

sapphi-red (approved):
LGTM

hyfdev (approved):
Good job!

PR #10825, opened 2026-09-05
fix: tree-shake expression statements that coerce known constants

### LINKED ISSUE #10817: [Bug]: variables do not get tree-shaken in some cases

https://repl.rolldown.rs/#eNptkE1ugzAQha8ysiqRVMSQRVsJ1FWv4UUcMJElGEfGtImQ796x+Slp6401733jN+ORNawYmTVtW5sv5JXBRl+464OKrPjPSVlFhu6uxjoYoVaNRvURbfDQWNOBWPsEKwUKVLdIEyuHNt5rz24UCKDxOriCGjXW6kYxgqVBN4OLRoQAGmM7GTnVdzMCkGXQadTNvQBnB/VXTepKHQy29yRYngC/p7mYT9e8ZeG1/rVmr+gH6mXBhGdU9k6i6+lNgaEyreKtuewmdF/SE4oVYSLKWfFN1oM25c0fFZ0l8x2OeZ6Xwk4xLqw1OEX6aw7PMxXtH0SSm0g6yabvTOLpaZT+tBGrR3EzNE39SSMdec7zw1k5yV/emP8G4rO7tQ==

### What is expected?

Unused variables/expressions should be removed from the resulting bundle.
```
//#region constants.ts
const second = 1e3;

//#endregion
//#region index.ts
console.log(second);

//#endregion
```

Rolldown removes unused variables, but leaves behind useless expressions.
```
//#region constants.ts
const second = 1e3;
60 * second;
const a = "aaaa";
`${a}`;
`${a}`;

//#endregion
//#region index.ts
console.log(second);

//#endregion
```

### System Info

```Shell
System:
    OS: Windows 11 10.0.26200
    CPU: (32) x64 AMD Ryzen 9 9950X3D 16-Core Processor          
    Memory: 35.15 GB / 63.58 GB
  Binaries:
    Node: 26.8.1 - C:\Program Files\nodejs\node.EXE
    npm: 12.0.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 152.0.7977.76
    Edge: Chromium (152.0.4191.53)
    Firefox: 155.0 - C:\Program Files\Mozilla Firefox\firefox.exe
```

### Any additional comments?

Additionally:
- if I remove the variable declarations, but leave behind their expressions, the expressions will not be removed in the output either
  ```js
  export const second = 1000;
  60 * second;
  
  const a = 'aaaa';
  `${a}`;
  `${a}`;
  ```
- removing one of the ``` `${a}` ``` DOES remove all the strings
  - input:
    ```js
    export const second = 1000;
    60 * second;
    
    const a = 'aaaa';
    `${a}`;
    // `${a}`;
    ```
  - output:
    ```js
    //#region constants.ts
    const second = 1e3;
    60 * second;
    
    //#endregion
    //#region index.ts
    console.log(second);
    
    //#endregion
    ```
- enabling `minify: true` or `minify: 'dce-only'` does not help, surprisingly
- with `minify: true`:
  - changing `1000` to `100` or `1e2` DOES remove the `60 * second` expression
    - changing `1000` to `1e3` does not help
  - changing `'aaaa'` to `'aaa'` DOES remove all the strings
  - `minify: 'dce-only'` does not help with the above
  - input:
    ```js
    export const second = 100;
    60 * second;
    
    const a = 'aaa';
    `${a}`;
    `${a}`;
    ```
  - output:
    ```js
    console.log(100);
    ```

IWANABETHATGUY:
Expression statements like `60 * second;` or `` `${a}`; `` were kept after the unused `const` was removed, because the side-effect analyzer did not know the type of the constant and treated the coercion as possibly throwing. A `const` initialized with a template literal (`` const a = `` ``, the follow-up case in the issue) was not recorded as a constant at all, so its coercions were kept even with `minify: 'dce-only'`.

Object-valued constants are still kept, and the imported-constant pass only runs when `optimization.inlineConst` is enabled because it reuses the constant map collected for inlining.

fixes #10817


crates/rolldown/src/stages/link_stage/tree_shaking/determine_side_effects.rs
```diff
@@ -28,6 +38,38 @@ impl LinkStage<'_> {
...
+  /// side-effect-only imports, and `include_statements` emits it empty.
+  ///
+  /// Every `Analyzed` verdict resets to its scan-time value. A relaxed module recomputes it from
+  /// its scan-time statements only, as `lazy_check_side_effects` did, so appended bundler
+  /// statements do not count.
+  pub fn recompute_analyzed_side_effects(&mut self) {
+    if self.relaxed_side_effect_modules.is_empty() {
+      return;
+    }
+    for module in self.module_table.modules.iter_mut().filter_map(Module::as_normal_mut) {
```
hyfdev-agent, line 55: **P1: Multi-entry import cycles can lose side effects.**

The [side-effect recomputation](https://github.com/rolldown/rolldown/blob/f8bde773ee072f09ae8b205f12d5c80244b23785/crates/rolldown/src/stages/link_stage/tree_shaking/determine_side_effects.rs#L55) resets module states before traversing their dependencies. A module in a cycle is cached as side-effect-free too early, and that verdict is not updated when a dependency is later found to have side effects. This removes an import that another entry needs. In the reproduction, base prints `effect → main`, while HEAD prints only `main`.

crates/rolldown/src/ast_scanner/mod.rs
```diff
@@ -1118,9 +1118,14 @@ impl<'me, 'ast: 'me> AstScanner<'me, 'ast> {
...
 
   pub fn add_constant_symbol(&mut self, symbol_id: SymbolId, value: ConstExportMeta) {
+    let scoping = self.result.symbol_ref_db.scoping();
     let is_mutated = !self.result.symbol_ref_db.is_facade_symbol(symbol_id)
-      && self.result.symbol_ref_db.scoping().symbol_is_mutated(symbol_id);
-    if is_mutated {
+      && scoping.symbol_is_mutated(symbol_id);
+    // Direct `eval` can assign a `let`, a `var`, or an `exports` property without a write
+    // reference, so only a `const` is stable in such a module.
+    let eval_can_assign = scoping.root_scope_flags().contains_direct_eval()
```
hyfdev-agent, line 1126: **P2: Multi-pass constant inlining can use a value from before a direct eval mutation.**

The [new eval check](https://github.com/rolldown/rolldown/blob/f8bde773ee072f09ae8b205f12d5c80244b23785/crates/rolldown/src/ast_scanner/mod.rs#L1126) only covers the scan stage; the link stage adds the old value back to the constant map. With `inlineConst: { mode: "smart", pass: 3 }`, let `eval` change `value` from `'aaaa'` to `'bbbbb'`. The importing module's `` `${value.length}` `` is incorrectly inlined as `"4"`; both the original program and base print `"5"`.

hyfdev-agent (commented):
**Two regressions are confirmed at head `f8bde773`; I recommend fixing them before merging.**

Both issues were reproduced by comparing base and head. All 108 Rust unit tests and the four new fixtures passed; the existing tests do not cover these two cases.

PR #10841, opened 2026-09-08
perf(vite-reporter): reuse emitted sourcemap sizes

hyfdev:
Builds with large JSON modules can allocate a large temporary string just to print a sourcemap's size. This PR reuses the byte length of the emitted sourcemap instead.

## Problem

With sourcemaps and build reporting enabled, the reporter calls `chunk.map.to_json_string().len()` even though the sourcemap has already been serialized into an output asset.

## Fix

For external and hidden maps, size reporting uses the final asset content after `generateBundle` hooks. A local filename-to-size index matches each chunk to its sourcemap asset. Reported sizes therefore reflect asset-only plugin edits. Inline maps and missing map assets keep the existing serialization path. Emitted files are unchanged.

## Validation

The benchmark imports a synthetic 250 MB JSON object containing records into a minified Vite build with sourcemaps enabled and gzip reporting disabled. The table shows medians of three runs per variant using release tracking builds on macOS arm64:

| Metric | Baseline | This PR | Change |
| --- | ---: | ---: | ---: |
| Peak live Rust allocations | 4.20 GiB | 3.49 GiB | -16.94% |
| Peak process RSS | 5.52 GiB | 5.52 GiB | +0.09% |

The approximately 17% reduction is in outstanding Rust allocations, including spare capacity, not process RSS.

Tests cover external, hidden and inline maps; disabled maps; custom filenames; inline filename collisions; and modified or missing assets. The modified string and `Uint8Array` asset cases fail on the baseline and pass with this change.


sapphi-red (approved):
Makes sense to me. It even feels slightly more correct.

IWANABETHATGUY (approved):

shulaoda (approved):

shulaoda:
### Merge activity

* **Sep 14, 5:17 AM UTC**: The merge label 'graphite: merge-when-ready' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown) once it meets the requirements.
* **Sep 14, 5:17 AM UTC**: `shulaoda` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).
* **Sep 14, 5:34 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=rolldown&repo=rolldown).

PR #10855, opened 2026-09-10
fix(minify): preserve annotation comments when minified (#10854)

### LINKED ISSUE #10854: [Bug]: `comments.annotation: true` does not preserve `#__PURE__` under `minify: true` (legal comments in the same output are preserved)

```
npm i rolldown@1.2.8
```

`input.js`

```js
/*! legal comment */
export const T = /*#__PURE__*/ Object.freeze({ a: 1 });
export const f = (n) => n + 1;
```

`build.mjs`

```js
import { rolldown } from "rolldown";
import { readFileSync } from "node:fs";

for (const minify of [false, true]) {
  const bundle = await rolldown({ input: "input.js" });
  await bundle.write({
    file: `out-${minify}.mjs`,
    format: "esm",
    minify,
    comments: { legal: true, annotation: true },
  });
  console.log(`--- minify: ${minify} ---`);
  console.log(readFileSync(`out-${minify}.mjs`, "utf8").trim());
}
```

```
node build.mjs
```

### What is expected?

With `comments: { legal: true, annotation: true }`, both comment categories survive minification, so a published library keeps the `#__PURE__` annotations its consumers' bundlers need in order to drop the value.

The legal comment is preserved and the annotation is dropped, in the same output:

```
--- minify: false ---
//#region input.js
/*! legal comment */
const T = /*#__PURE__*/ Object.freeze({ a: 1 });
const f = (n) => n + 1;

//#endregion
export { T, f };
--- minify: true ---
/*! legal comment */
const e=Object.freeze({a:1}),t=e=>e+1;export{e as T,t as f};
```

The surviving `/*! legal comment */` is the control: it shows `comments` is read and honoured in that same run, so this is not "minify strips every comment". `comments.legal` works; `comments.annotation` does not.

Measured across the option surface — the `legal` column is the liveness control in every row:

| output options (all with `minify: true`) | `/*! legal */` | `#__PURE__` |
| --- | --- | --- |
| *(default comments)* | kept | **dropped** |
| `comments: true` | kept | **dropped** |
| `comments: false` | dropped | dropped |
| `comments: { legal: true, annotation: true }` | kept | **dropped** |
| `comments: { legal: false, annotation: true }` | dropped | **dropped** |

Identical on **1.2.7** and **1.2.8**. Both annotation spellings behave the same — `/*#__PURE__*/` and `/*@__PURE__*/` — so this is not about the spelling `CommentsOptions.annotation` documents. At `minify: false` the annotation is preserved.

### System Info

```Shell
  System:
    OS: macOS 26.6.2
    CPU: (12) arm64 Apple M3 Pro
    Memory: 11.17 GB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.18.1 - /opt/homebrew/bin/node
    npm: 11.16.0 - /opt/homebrew/bin/npm
    pnpm: 11.9.0
    bun: 1.3.14 - /opt/homebrew/bin/bun
  npmPackages:
    rolldown: ^1.2.8 => 1.2.8
```

### Any additional comments?

**Why this matters for library builds.** A library bundled with `minify: true` ships the minified chunk, and the consumer's bundler is the one that decides whether a frozen top-level constant can be dropped. With the annotation gone from the published file, that decision can no longer be made, and the constant is retained in every consumer bundle.

Measured on a real package (23-package monorepo, built by `tsdown` 0.23, which resolves rolldown 1.2.7): a consumer importing **one small symbol** from a subpath pays 7.5–10× what the symbol costs, because the chunk it lands in opens with a comma chain of frozen tables that nothing may drop. Adding `/*#__PURE__*/` in initializer position measures −1 417 B raw / −485 B gzip on those entries when the annotation reaches the consumer, and 0 when it does not. Corroborating: across that repo's entire `dist`, `__PURE__` occurs exactly twice, both in the one package not minified by rolldown.

Related but not the same: #10068 (heuristic DCE for pure top-level call/new initializers) and #9943 (`NO_SIDE_EFFECTS` on non-function initializers) both concern whether rolldown itself acts on such initializers. This report is narrower — it is about the annotation reaching the *output*, so someone else's bundler can act on it.

justonemorenight:
### Summary

Fixes #10854

When `minify: true`, `remove_whitespace` is set to `true`. In `crates/rolldown/src/stages/generate_stage/minify_chunks.rs`, `CommentOptions` previously gated `annotation` comments behind `options.comments.annotation && !remove_whitespace`.

As a result, pure/annotation comments (`/*#__PURE__*/`, `/*@__NO_SIDE_EFFECTS__*/`, etc.) were unconditionally stripped whenever minification was enabled, even when `comments.annotation: true` (or the default `comments: true`) was configured. This prevented downstream bundlers from performing dead-code elimination on published library chunks.

### Changes

- In `minify_chunks.rs`, pass `options.comments.annotation` directly to `CommentOptions.annotation` (consistent with how `legal` comments are handled).
- Added an integration test fixture `preserve_annotations_when_minified` verifying comment retention across `minify: true` and multiple `comments` option variants (`legal: true, annotation: true`, `legal: false, annotation: true`, `legal: true, annotation: false`, and `comments: false`).


babamottos-bit (approved):

babamottos-bit (approved):

IWANABETHATGUY:
Hi @babamottos-bit, thanks for reviewing! Just a quick heads-up: since write permissions are needed to approve PRs in this repo, submitting approvals doesn't affect the merge status, but it does send out extra notifications. Feel free to leave any feedback directly in regular comments instead. Thanks!

hyfdev (approved):
Thank you!

IWANABETHATGUY (approved):

PR #10865, opened 2026-09-12
fix: preserve direct eval CommonJS bindings

### LINKED ISSUE #10826: Function parameters are renamed inside a function containing a direct eval, breaking the eval'd code
### Environment

- rolldown 1.2.7 (also reproduced through Vite 8.2.2, which bundles with rolldown)
- Node 22.23.2, macOS 15

### Reproduction

`lib.cjs` (a webpack-style module registry, the shape webpack's `eval` devtool produces; this exact pattern ships inside `@nasaworldwind/worldwind`, which bundles jszip that way):

```js
var modules = {
  './a.js': function (module, exports, __webpack_require__) {
    eval("exports.answer = 42;");
  },
};
function load(id) {
  var module = { exports: {} };
  modules[id].call(module.exports, module, module.exports, load);
  return module.exports;
}
module.exports = { answer: load('./a.js').answer };
```

`entry.js`:

```js
import lib from './lib.cjs';
console.log('answer:', lib.answer);
```

```sh
node entry.js                                   # answer: 42
rolldown entry.js --format esm --file out.js
node out.js                                     # answer: undefined
```

### What happens

The parameters of the function that contains the direct `eval` are renamed in the output (`module` -> `module$1`, `exports` -> `exports$1`), but the code inside the `eval` string still refers to `exports`, so it now resolves to the outer CommonJS wrapper's `exports` object instead of the module's own. The property lands on the wrong object and the module's export is `undefined`.

The same happens with `minify: false`, so it is the scope-hoisting/CJS transform rather than the minifier.

### Expected

Direct `eval` makes every binding in the enclosing scopes observable by name, so a function containing a direct `eval` (and its enclosing scopes) must keep its original identifiers. esbuild and Rollup both keep the names in this case and the bundled code prints 42.

### Real-world impact

Any Vite 8 project that imports `@nasaworldwind/worldwind` fails at runtime after `vite build` with `utils.inherits is not a function` (jszip's `exports.inherits` is assigned to the wrong object). Dev mode, which pre-bundles with esbuild, works. I worked around it in [worldwind-kit](https://github.com/qwertymuzaffar/worldwind-ui/blob/main/packages/kit/src/uneval.ts) by rewriting `eval("...")` calls into inline functions before bundling.

dexhunter:
## What is this PR solving?

When a CommonJS module uses direct `eval` inside a webpack-style wrapper, the evaluated code resolves the wrapper's original `module` and `exports` bindings. Rolldown was renaming those bindings during deconfliction, so the generated bundle could update a different binding and return the wrong value.

This keeps the `module` and `exports` wrapper names stable for modules that contain direct eval while preserving the existing collision handling for other wrapper parameters.

Fixes #10826

## What other alternatives have you explored?

The change is limited to the two CommonJS wrapper names that direct eval resolves. External factory parameters and modules without wrapper parameters continue through the existing collision checks.

## Validation

The focused issue fixture and no-eval control were run against base `2ef10c689f73174ee8163df5bcf1485fa5ee7833` and candidate `37dfc18c02cd9a89935015ccf3ef4ef3bde47d21` with the pinned Rust 1.98.1 and Node 22.23.1 toolchains. The base produced `42` for the direct reference and `undefined` for the bundled direct-eval case. The candidate produced `42` for both; its no-eval controls continued to produce `42` in both direct and bundled forms.

Candidate `cargo clippy --workspace --all-targets -- --deny warnings` and `cargo check --workspace --all-features --all-targets --locked` passed. The bounded candidate test command `cargo test --workspace --exclude rolldown_binding --locked` also passed: 2,415 tests passed, 0 failed, and 70 were ignored.

An earlier candidate attempt reused a stale build artifact; its failed result is retained in the trajectory. The changed crate was then rebuilt and the candidate checks passed.

This change targets correctness; no performance improvement or full WorldWind application coverage is claimed.

AI assistance was used for implementation and evaluation, with a separate agent reviewing the final change and results.

Weco Observe: https://dashboard.weco.ai/share/tQh2JcFvaT_4C6VBOw7nMnsJyROYA2s2

### DIFF
--- a/crates/rolldown/src/utils/renamer.rs
+++ b/crates/rolldown/src/utils/renamer.rs
@@ -469,10 +469,17 @@ impl NestedScopeRenamer<'_, '_> {
       return;
     }
 
+    // Direct eval resolves CJS wrapper bindings by their original source names. Preserve only
+    // those two names in an eval-bearing CJS module; external IIFE/UMD factory names still need
+    // the ordinary collision protection below.
+    let preserve_cjs_wrapper_names = is_cjs_wrapped && self.module.meta.has_eval();
+
     // Skip root scope (index 0), check nested scopes only
     for (_, bindings) in self.scoping.iter_bindings().skip(1) {
       for (&name, symbol_id) in bindings {
-        if wrapper_param_names.contains(name.into()) {
+        if wrapper_param_names.contains(name.into())
+          && !(preserve_cjs_wrapper_names && matches!(name.as_str(), "exports" | "module"))
+        {
           let symbol_ref = (self.module_idx, *symbol_id).into();
           self.renamer.register_nested_scope_symbols(symbol_ref, name.as_str());
         }
--- a/crates/rolldown/tests/rolldown/issues/10826_direct_eval_cjs/_config.json
+++ b/crates/rolldown/tests/rolldown/issues/10826_direct_eval_cjs/_config.json
@@ -1,0 +1,13 @@
+{
+  "config": {
+    "input": [
+      { "name": "main", "import": "./main.js" },
+      { "name": "control", "import": "./control.js" }
+    ],
+    "format": "esm",
+    "platform": "node"
+  },
+  "expectExecuted": true,
+  "writeToDisk": true,
+  "snapshot": false
+}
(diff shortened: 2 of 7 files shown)


IWANABETHATGUY:

IWANABETHATGUY (approved):

PR #10879, opened 2026-09-14
fix: __vitePreload inlined awaited import should be transformed with correct syntax

smeng9:
Fixes downstream https://github.com/vitejs/vite/issues/23485


sapphi-red:
I think we should skip wrapping with `__vitePreload` like v7 if the specifier of the dynamic import is not a static string because Vite / Rolldown won't know what to preload anyway.

If we were to wrap, we have to keep the sideeffect outside of the `__vitePreload`. The output would have to be like
```js
let _temp = await resolveSpecifier(spec)
return await __vitePreload(() => import(_temp), []);
```
. Generating this is a bit tricky.

Issue #5269, opened 2025-07-14
[Task]: Add test to ensure stable export name order when `output.minifyInternalExports` is enabled


sapphi-red:
### Reproduction link or steps

1. Clone https://github.com/notKamui/miniverso
2. `git switch rolldown`
3. Comment out this line (https://github.com/notKamui/miniverso/blob/aa845cb1545eedf9a9ffdd48e80b8cb82d35a83e/vite.config.ts#L20) (tailwind is unstable as well)
4. `rm -rf src/routes/_authed src/routes/_authed.tsx src/routes/auth`
5. `pnpm i`
6. Run `pnpm build` multiple times
### What is expected?

The generated file contains the same content even if it's built multiple times.

### What is actually happening?

The generated file sometimes contains a different content.

### System Info

```Shell
System:
    OS: Linux 6.6 Ubuntu 24.04.2 LTS 24.04.2 LTS (Noble Numbat)
    CPU: (24) x64 AMD Ryzen 9 5900X 12-Core Processor
    Memory: 11.29 GB / 23.47 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.16.0 - ~/.local/share/asdf/data/installs/nodejs/22.16.0/bin/node
    npm: 10.9.2 - ~/.local/share/asdf/data/plugins/nodejs/shims/npm
    pnpm: 10.12.4 - ~/.local/share/asdf/data/installs/nodejs/22.16.0/bin/pnpm
    bun: 1.2.17 - ~/.local/share/asdf/data/shims/bun

Rolldown https://github.com/rolldown/rolldown/tree/a3a4228c16289091eccb66451ce47785f7ecadbe
```

### Any additional comments?

I found that the entry order of this loop is not stable.
https://github.com/rolldown/rolldown/blob/a3a4228c16289091eccb66451ce47785f7ecadbe/crates/rolldown/src/stages/generate_stage/compute_cross_chunk_links.rs#L460-L466
When there's multiple symbols that has the same owner module, the order of them were not stable.
I added `println!("export_ref: {:?}", chunk_export);` in this loop and got the following outputs:
<details>

```
export_ref: SymbolRef { owner: 88, symbol: SymbolId(4) }
export_ref: SymbolRef { owner: 193, symbol: SymbolId(1) }
export_ref: SymbolRef { owner: 193, symbol: SymbolId(3) }
[...]

PR #5281, opened 2025-07-15
fix(minify-internal-exports): ensure minifying internal exports stably

hyf0:
Part of  https://github.com/rolldown/rolldown/issues/5269.

PR #10882, opened 2026-09-15
fix: stabilize same-named cross-chunk exports

danieljvdm:
Cross-chunk exports can receive different aliases when asynchronous module loading changes discovery order. A generated module namespace and a local binding can share a name, such as `server_exports`, so `(Reverse(exec_order), name)` does not fully order them.

Use the symbol’s per-module ID as the final tie-breaker while preserving the existing execution-order and name priority. This closes the remaining same-name case from #5269.

AI-assisted implementation and test development with Codex.


danieljvdm:
The regression uses six small JavaScript inputs and public `load` / `moduleParsed` hooks to force opposite child-discovery orders. Both parent loads arrive before one is released; parsing its child releases the other. Each generated program must preserve the static and dynamic exported values and source-map attachment, then both schedules must produce identical filenames and bytes. There are no sleeps or probabilistic retries.

Against the unchanged, locally built native base `c3c32ec74ef295895e7d640490f9cf29cf748a5a`:

```text
× cross-chunk export names are stable across module load orders 49ms
AssertionError: expected [ 'consumer-Yyk3GD0H.js', …(6) ]
  to deeply equal [ 'consumer-Cp5lGn7g.js', …(6) ]
Test Files  1 failed (1)
Tests       1 failed | 1 skipped (2)
```

All scheduling, runtime and map assertions passed before the exact-files failure. With the `SymbolId` tie-breaker:

```text
✓ cross-chunk export names are stable across module load orders 58ms
Test Files  1 passed (1)
Tests       1 passed | 1 skipped (2)
```

The skipped case was excluded by the named invocation. The final formatted file passed both cases. Completed local validation:

| Scope | Result |
| --- | --- |
| Native binding, JS glue and declaration typecheck | Passed |
| Final hash-stability file | 2 passed |
| Full behavior suite, default concurrent forks | 25 passed across 8 files |
| Rust deconfliction integration fixtures, including default extended variants | 28 passed; 1 pre-existing ignored `reserved_names` case |
| Rust minified-export fixture | 1 passed |
| Existing conflict-resolver unit tests | 8 passed |
| Owning all-feature library Clippy, Rust/TS formatting, configured type-aware TS lint | Passed |
| Required generator | Passed; generated source unchanged |

A separate, single cancellation proof held both real parent loads and called public Vitest `cancelCurrentRun`. Both loads rejected through the real test signal; bundle close completed before cleanup restored the fixture global and removed its directory. Public cancellation settled **7.790ms** after the cancellation request. The process exited in **0.4966345s** with the API driver's intentional status **130**, no surviving owned process, and unchanged source/config/native artifact hashes. This verifies API cancellation and cleanup, not OS signal propagation or ordinary regression success. The initial offline result-parser error was corrected against retained raw evidence without rerunning the experiment.

The qualified red also exited naturally, with no surviving process. Environment: local macOS, Rust 1.98.1, pnpm 11.24.0 and Vitest 4.1.11, using the locally built debug native backend. These are correctness and lifetime observations, not a speed comparison. Full workspace or cross-platform CI is not claimed.

The history is specific: [#5208](https://github.com/rolldown/rolldown/commit/0e46adc5b4b8a26554185b02d67eb3ec858f3418) introduced execution-order-only allocation. [Issue #5269](https://github.com/rolldown/rolldown/issues/5269) reported unstable same-owner ordering; [#5281](https://github.com/rolldown/rolldown/commit/8ef00f36fd5abfd82743706f7b615375d2353c31) added names but left distinct same-named symbols tied. [#9831](https://github.com/rolldown/rolldown/commit/8a0652cdc349284b6d5d92f692c22286a4bab2d3) later centralized that existing key. Reachable exported owners already have unique deterministic execution orders, so a per-owner symbol ID resolves this remaining tie without depending on asynchronous module indices.

The initial missing-CMake failure and fixture/output-directory collision were retained and excluded from the qualified red/green comparison. The corrected regression was unchanged between red and green; its later formatting changed layout only.

Subsequent hosted confirmation: [upstream CI passed](https://github.com/rolldown/rolldown/actions/runs/34925855787) on exact PR commit `b0dff2073025a7af20d487fa383a184f1af25f31`. All 16 selected jobs succeeded. The logs show four successful executions of this regression: the Ubuntu Node 20, 22 and 24 suites, plus WASI. The full Ubuntu Cargo job, Browser, Vite integration, and Node/dev-server validation also passed. Seven Windows/macOS jobs were skipped by the existing workflow selection; this run does not establish those platforms.

Issue #10844, opened 2026-09-08
inlineDynamicImports + a dynamic import() inside an import cycle wraps the whole cycle as non-treeshakeable, even modules unrelated to the dynamic import


jjeff:
With `output.inlineDynamicImports: true`, rolldown wraps a large fraction of an app's module graph as non-treeshakeable lazy `__esmMin` wrappers whenever the graph contains an import cycle plus a dynamic `import()` of a module inside that cycle. Inside a wrapped region, a namespace import (`import { m } from './messages.js'; m.foo()`) becomes a real runtime object instead of being shaken per-property, so every property the namespace exports ships even when only a handful are ever read.

## Repro (observed in a real app, ~2,600 wrapped modules)

Using `vite build` (rolldown 1.2.4) in lib mode with `output.inlineDynamicImports: true` on an app whose main entry has:

- a large mutual-import cycle across ~30 modules, and
- at least one `await import('<module inside the cycle>')` cycle-breaker

nearly every module reachable from that cycle gets wrapped, including modules with no direct relationship to the dynamic import. Concretely, in our app:

| Experiment | `__esmMin(` wrapper count |
|---|---|
| Entry as-is | 2,604 |
| Same entry with the one `await import()` rewritten to a static import | 567 |
| 2-file synthetic cycle, no dynamic import | 0 |
| Same synthetic cycle + one `import()` of a cycle member | 1,562 |

`inlineDynamicImports: false` (code-splitting) does not help — the same modules wrap across chunks. `experimental.onDemandWrapping: true` does not help either. `output.strictExecutionOrder` was already off.

## Impact

A namespace-imported module whose export surface is much larger than what any one entry point uses (a generated i18n message table, a barrel re-export) ships in full once wrapped, even though direct/static per-property imports of the same module tree-shake correctly outside a wrapped region. In our case this added several hundred KB of dead code to a packaged Electron `main.js` (unused localization strings, and a browser SDK barrel pulled in through a similar namespace-import path) — see spaceagetv/missioncontrol#5693 for the full case study and our interim workaround (a generated barrel that only imports the properties the entry point's real module graph references).

## Ask

Either scope the wrapping to only the modules that actually sit on a cycle-breaking dynamic-import path (rather than the whole reachable cycle), or make wrapped-region namespace imports still shakeable per-property when nothing in the reachable code dynamically re-exports the whole namespace object.

Happy to share a minimal repro repo if useful — the 2-file synthetic cycle above already reproduces it outside our app.

[...]

Issue #10843, opened 2026-09-08
inlineDynamicImports: a dynamic import() of an import-cycle member disables per-property tree-shaking of an unrelated namespace re-export elsewhere in the bundle


jjeff:
## Summary

With `output.inlineDynamicImports: true`, a single `await import()` of a module that is itself part of an unrelated import cycle causes rolldown to wrap a large portion of the bundle's modules as non-treeshakeable lazy `__esmMin` initializers. Inside a wrapped region, `import { m } from './messages.js'; m.someKey()` (where `messages.js` is `export * as m from './messages-index.js'`, itself `export *`-ing hundreds of small modules) stops being tree-shaken per-property — every message module ships, not just the ones actually called through `m`.

Found while investigating why a packaged Electron main-process bundle (built via `@electron-forge/plugin-vite` in lib/cjs mode, `inlineDynamicImports: true` because the output has to be a single file for `bytenode` to compile) carried all ~1,900 compiled i18n message modules instead of the ~1,000 the main process actually calls, plus the full `@sentry/browser` barrel pulled in through one similarly-shaped `require('@sentry/browser')` inside a wrapped module.

## Environment

- rolldown: 1.2.4
- Also present via vite: 8.2.1 (`@electron-forge/plugin-vite`'s lib-mode config, `format: 'cjs'`)
- Node: (whatever CI/dev uses — not version-specific in our testing)

## What we've established (from the real codebase, not synthetic)

| Experiment (unminified build, `__esmMin(` count) | Wrappers | Live message-namespace properties retained |
|---|---|---|
| Real main entry as-is | 2,604 | 1,554 / ~1,900 |
| 2-file synthetic cycle, no dynamic import | 0 | 2 (fully tree-shaken) |
| Same cycle + one `import()` of a cycle member | 1,562 | 1,558 (essentially everything) |
| A large real module (`StateHouse.ts`) as-is | 2,439 | 1,421 |
| Same module, its own `await import()` calls rewritten to static imports | 567 | 147 |
| Same real entry with `output.inlineDynamicImports: false` (code-split) | 2,601 across chunks | 1,558 |
| Same real entry with `experimental.onDemandWrapping: true` | 2,604 | 1,554 |

[...]

Issue #10697, opened 2026-08-18
WASI throws "memory access out of bounds" when concurrent builds call JS plugin hooks


hyfdev:
## Symptom / behavior in Rolldown

On the threaded WASI binding (`wasm32-wasip1-threads`), a build randomly traps with `RuntimeError: memory access out of bounds` (stack entirely inside wasm). This is what makes the `wasi / WASI` CI job flaky — e.g. run 32091017370, `fixtures-concurrent.test.ts > external/regexp`. The native binding runs the same suite and never shows it.

In Rolldown it shows up when two or more builds run concurrently in one process with a JS plugin hook (e.g. `transform`) active — the wasi test suite's concurrent fixtures with plugins are the usual trigger. That combination is only *how Rolldown reaches the underlying conditions* (concurrent allocation across wasm instances while memory grows); it is not the root cause, and neither the plugin hook nor the concurrency at the build level is fundamentally required — see the root cause below.

It surfaces in three shapes, all the same underlying fault:
1. a random test rejects with `RuntimeError: memory access out of bounds`;
2. a test that asserts on an error message gets the OOB text instead (e.g. `expected 'memory access out of bounds' to contain 'Invalid substitution'`) — the trap was caught by the plugin driver and turned into a build diagnostic;
3. mass 60s timeouts / a hung suite — a crashed worker thread is dropped silently (`@emnapi/wasi-threads`' `handleError` only calls `terminateAllThreads()` when the message contains `RuntimeError`/`unreachable`; a bare `memory access out of bounds` worker error doesn't match), so in-flight builds wait forever.

## Root cause

[...]

PR #10898, opened 2026-09-16
fix(finalizer): mark no-op `init_*()` from `export *` as pure

L4XB:
## What this solves

`ref #10844` (partial — see "Scope" below).

#9678 made the finalizer annotate a wrapped-ESM `init_*()` call `/* @__PURE__ */` when the
callee's `__esm` closure is provably empty, so `minify: 'dce-only'` can drop the call, then the
now-unreferenced wrapper declaration, then the runtime helper. Its summary describes setting
`pure = true` "at the three emission sites".

There is a fourth. The `export * from 'path'` branch in `ScopeHoistingFinalizer` builds its init
call inline with `ast::Expression::new_call_expression` instead of going through
`wrapped_esm_init_call_expr`, so it never consults `init_is_noop` and never sets the flag:

```rust
// crates/rolldown/src/module_finalizers/mod.rs, `export * from 'path'` branch
let mut init_expr =
  ast::Expression::new_call_expression(SPAN, wrapper_ref, None, [], false, self);
```

Because the emitted call is opaque to DCE, and because the call keeps the wrapper alive, the
empty `var init_x = __esmMin(() => {})` declaration survives with it. A barrel that re-exports
N function-only modules therefore emits N dead wrappers plus N dead calls whenever anything
forces the barrel to be wrapped.

That "anything" is easy to hit: `wrap_module_recursively` seeds from the dynamically imported
module and then recurses over **every** import record, so under `codeSplitting: false` a single
`import()` of an import-cycle member wraps the whole graph reachable from the cycle — barrel and
message modules included. That is the situation #10844 reports.

## The change

Route the `export *` emission through the same annotation the other sites already use. One call
expression, one file:

```rust
let mut init_expr = ast::Expression::new_call_expression_with_pure(
  SPAN, wrapper_ref, None, [], false,
  self.ctx.final_esm_init_metadata.init_is_noop(importee.idx),
  self,
);
```

Target resolution and the TLA `await` branch are untouched; this only adds the annotation.

## Effect

New fixture `crates/rolldown/tests/rolldown/issues/10844` — a two-module cycle, a dynamic
`import()` of a cycle member, and an `export * as m` barrel over two function-only modules,
under `codeSplitting: false`.

Before:

```js
//#region src/m1.js
function m1() { return "m1"; }
var init_m1 = __esmMin((() => {}));
//#endregion
//#region src/m2.js
function m2() { return "m2"; }
var init_m2 = __esmMin((() => {}));
//#endregion
//#region src/messages-index.js
var init_messages_index = __esmMin((() => {
	init_m1();
	init_m2();
}));
//#endregion
```

After — the annotations let DCE drop the calls, which leaves both wrappers unreferenced, which
empties the barrel's own closure:

```js
//#region src/m1.js
function m1() { return "m1"; }
//#endregion
//#region src/m2.js
function m2() { return "m2"; }
//#endregion
//#region src/messages-index.js
var init_messages_index = __esmMin((() => {}));
//#endregion
```

Scaled to the same shape with 200 message modules (local fixture, not committed), bundling
through the same Rust integration harness:

| | `main.js` bytes | `init_*` wrappers | empty wrappers |
|---|---|---|---|
| `5b4746e` | 18,446 | 204 | 200 |
| this branch | 8,265 | 4 | 1 |

## Scope — what this does not do

- **It does not narrow the wrapped region.** `wrap_module_recursively` still wraps the full
  transitive closure of import records from any seed, which is the other half of #10844's ask.
  That is a faithful port of esbuild's `recursivelyWrapDependencies` ("All dependencies must
  also be wrapped"), and on the fixture above esbuild wraps 336 modules to rolldown's 335 — so
  narrowing it is a deliberate divergence that wants its own design discussion, not a drive-by.
- **It does not change tree-shaking.** I could not reproduce the per-property loss reported in
  #10843 in any configuration I tried (esm and cjs output, 1 and 130 distinct-key importers,
  barrel inside and outside the wrapped region, `onDemandWrapping` on and off): rolldown
  retained exactly the properties actually read, 130 of 200, in every one. esbuild retained
  200 of 200 in all of them. The residual cost in those measurements was the wrapper
  scaffolding this PR removes, not retained message bodies.
- `init_messages_index` above stays as an empty wrapper because `init_is_noop` is a source-AST
  predicate and `export … from` statements make it return `false` before lowering. That is the
  documented conservative direction and it is O(1) per barrel; the per-message wrappers, which
  are the part that scales, are gone.

## Point for reviewers

The `await` branch immediately below the changed call. Marking a call pure and then awaiting it
would be wrong, but the two are mutually exclusive: `init_is_noop` only holds when every
top-level statement is a hoisted function declaration or a source-less export clause, so such a
module has no import records and cannot carry a TLA dependency. Worth a second opinion.

## Tests

- New: `crates/rolldown/tests/rolldown/issues/10844` — snapshotted and executed by the harness;
  its assertions (`hub() === 'm1m2b'`, same through the dynamic import) pass on both sides, so
  it discriminates on the snapshot, not on a crash.
- `cargo test -p rolldown`: 2004 passed / 5 failed on this branch, and the identical 2004 / 5
  on `5b4746e`. The five are pre-existing and environmental — `test262_module_code` (submodule
  not checked out), `npm_packages__util_deprecate` and three `cjs_module_lexer_compat` fixtures
  (node deps not installed). I verified each fails at the merge base too.
- No other snapshot in the suite changed.
- `cargo fmt --all --check` and `cargo clippy -p rolldown --all-targets` are clean.

File and line references are against `main` @ `5b4746e442989d770c606ce08d2737e6aafbd25d`.


L4XB:
Note on the red check: `Node Validation` failed in step `Build @rolldown/browser`, and the failure is the known WASI heap trap, not this branch.

Job: https://github.com/rolldown/rolldown/actions/runs/35105558080/job/104826688050

```
$ node --enable-source-maps --import @oxc-node/core/register -C dev ./build.ts
(node:3966) ExperimentalWarning: WASI is an experimental feature and might change at any time

Error: memory access out of bounds
    at napi_create_error (.../@emnapi/core/dist/emnapi-core.js:4591:23)
    at rolldown_binding.wasm.<napi::error::JsError>::into_value (wasm://wasm/rolldown_binding.wasm-21eaf0fe:wasm-function[7410])
    ...
error: recipe `build-browser` failed on line 231 with exit code 1
```

The stack is entirely inside the wasm binding and the trap surfaced as a build diagnostic — shape 2 of the three shapes in #10697, and exactly what the comment above the `Node Test` step in `reusable-wasi.yml` describes ("multiple wasm instances share one dlmalloc heap and corrupt it under concurrent allocation while the memory grows — a 'memory access out of bounds' trap that has nothing to do with the code under test").

Three measurements pin it to the flake rather than to this PR:

- **Same commit, same workflow run, identical command, green.** `browser / Browser` runs `prepare-fixture.mjs` → `pnpm run --filter rolldown build-browser-pkg:debug`, which is what `just build-browser` invokes; it passed and packed `rolldown-browser.tgz` (6.92 MB). `wasi / WASI` also passed, including `Build Rolldown with WASI Binding`, which drives the same wasm-based `build-node`.
- **The same step fails on branches with no Rust changes at all.** `renovate/vite+` @ `ba5e0f363` (pnpm-lock.yaml only, job 102352459167) and `renovate/rust-crates` @ `0a193a356` (Cargo.lock/Cargo.toml only, job 101527486775) both died in `Build @rolldown/browser` with the identical `Error: memory access out of bounds`. Three hits in eleven days, on three unrelated diffs.
- **This diff cannot cause an allocator trap.** It changes one call expression in `crates/rolldown/src/module_finalizers/mod.rs` (plus one fixture), reusing the `init_is_noop` predicate already consulted at two other emission sites — no new recursion, indexing or unsafe. `Rust Validation`, `cargo-test`, all six `node-test` / `node-dev-server-test` jobs, `Vite Test Ubuntu`, `Repo Validation` and `autofix` are green on the same head.

Could someone re-run `Node Validation`? `ci.yml` triggers only on `opened` / `synchronize`, so I can't restart it from my side without pushing a no-op commit. Worth knowing that steps 9-16 of that job (`Build @rolldown/debug`, the `git diff --exit-code` codegen check, `vp lint`, `Node Type Test`, publint, knip, workspace check) were skipped when step 8 died, so they have not yet run on this head.

Separately, and not something I'd fold into this PR: `reusable-wasi.yml` already wraps its `Node Test` step in the documented 3-attempt retry for this exact trap, while `Build @rolldown/browser` in `ci.yml` has no guard. Happy to send that as its own one-line CI PR if you want it.

PR #10901, opened 2026-09-16
feat(rolldown_plugin_esm_external_require): prefer 'module.exports' over a namespace spread

isker:
When `esmExternalRequire` handles a require of an external module that is itself CJS and that assigns a function to `module.exports`, the current behavior of spreading said module in the generated facade breaks: the spread makes the module not callable.

Teach `esmExternalRequire` to use the NodeJS `'module.exports'` convention before trying to do that spread, so that any module providing it can thread such a callable value through. Update the documentation, which was already outdated (it mentions no spread at all), to indicate the current behavior.

More background about our use case: we have a large monorepo build that invokes rollup, and we are trying to convert it to rolldown. Rollup is invoked many times granularly over different parts of the source code in this repo, to incrementally generate a large number of chunks that are dynamically linked together at runtime; in each invocation, modules from other invocations are externalized. Some of these chunks are third party dependencies that are distributed only as CJS. Some of those chunks require other such externalized CJS chunks. And some of these dynamically-linked facades involve `module.exports = function() ...`, thus the described issue.

We are already generating facade modules for any CJS chunks that ensure that the resulting chunks export all the necessary CJS symbols. We could update this facade to include a `'module.exports'` export, and have it work in concert with this new `esmExternalRequire` support. Without this support, we will have to roll our own version of `esmExternalRequire` that does not break this kind of callable interop.

AI disclosure: I implemented this change using Amp. You can see the thread that produced it [here](https://ampcode.com/threads/T-01a0aad3-e6f6-704e-abfd-29a445bed4c2). I understand and take responsibility for this code.

That being said, I am sure you have greater expertise than I do on CJS/ESM interop, among other things. **I am not sure this is the right design** (I am not sure why we are even spreading in the first place), only that it would support my use case.


isker:
@shulaoda sorry for the ping, but I saw you were assigned. Can you take a look?

shulaoda:
> @shulaoda sorry for the ping, but I saw you were assigned. Can you take a look?

Thanks for the contribution, I will take a look before today release.

shulaoda (approved):
Thanks for the contribution!

PR #10910, opened 2026-09-17
fix(module-loader): assign ModuleIdx in a stable order so builds are reproducible

### LINKED ISSUE #10909: [Bug]: ModuleIdx is assigned in task-completion order, so identical builds produce different output
### Reproduction

https://github.com/y0ngha/rolldown-nondeterministic-output

Zip snapshot: https://github.com/y0ngha/rolldown-nondeterministic-output/releases/download/v1/rolldown-nondeterministic-output.zip

```sh
git clone https://github.com/y0ngha/rolldown-nondeterministic-output && cd rolldown-nondeterministic-output
pnpm install

# A) no "sideEffects" in package.json
node -e "const f='package.json',j=require('./'+f);delete j.sideEffects;require('fs').writeFileSync(f,JSON.stringify(j,null,2))"
./run.sh 12
#   12 main-<hashA>.js        <- one output, 12 times

# B) "sideEffects": ["*.css","*.scss"]
node -e "const f='package.json',j=require('./'+f);j.sideEffects=['*.css','*.scss'];require('fs').writeFileSync(f,JSON.stringify(j,null,2))"
./run.sh 12
#    7 main-<hashA>.js
#    5 main-<hashB>.js        <- same input, two different outputs
```

`run.sh` runs `rm -rf dist .vite && vite build` N times and counts the output filenames.
Nothing else changes between runs - same files, same lockfile, same machine.

Reproduced on **rolldown 1.0.3** (vite 8.0.16) and **1.2.8** (vite 8.3.0).

`sideEffects` is the switch that makes this visible in the reduced case, not the cause.

### Confirming the cause without a patch

When the native binding loads, rolldown builds its own tokio runtime
(`crates/rolldown_binding/src/lib.rs`) with two pools: `ROLLDOWN_WORKER_THREADS` (async workers,
default `num_cpus::get_physical() * 3 / 2`) and `ROLLDOWN_MAX_BLOCKING_THREADS` (the
`spawn_blocking` pool, default 4). Completions from both land in the same `ModuleLoaderMsg`
channel, and `ModuleIdx` is handed out in arrival order. Pinning both pools to one thread
removes the race, and the output stops moving:

| 12 builds of the same input | distinct outputs |
| --- | --- |
| defaults | 7 : 5 |
| `ROLLDOWN_WORKER_THREADS=1` | 9 : 3 |
| `ROLLDOWN_MAX_BLOCKING_THREADS=1` | 6 : 6 |
| **both set to `1`** | **12 / 12 identical** |

Pinning one pool is not enough, because the other keeps racing. This is a diagnostic rather
than a workaround, since a single-threaded build gives up the parallelism.

### Why there is no REPL link

Two reasons, both in the source:

- That runtime block is `#[cfg(not(target_family = "wasm"))]`, and `rolldown_utils/src/rayon.rs`
  replaces rayon with a sequential shim under `target_family = "wasm"`. The browser build is
  therefore always in the "both pools at 1" state above, where this bug does not appear.
- `@rolldown/browser` has no npm resolution, and the ~430-module subgraph that flips is mostly
  dependency code, so the reproduction does not fit a REPL file map.

A REPL link would show a clean, deterministic bundle and read as "not reproducible". The
repository is a plain `pnpm install` plus `vite build` in a loop instead, with no native
toolchain or patched binary needed to see the split.

---

### What is expected?

Two builds of the same input produce byte-identical output.

The builds differ in **which modules survive tree-shaking**, not only in hashes. Chunk
metadata from `generateBundle` over 8 builds of the reduced case:

```
modules per build: [436, 436, 5, 5, 436, 5, 436, 5]
```

One variant keeps a ~430-module subgraph (an api-client chain and its vendor dependencies);
the other drops it and keeps ~5. Both builds succeed.

### Root cause

`ModuleIdx` is handed out while handling task-completion messages, which arrive in whatever
order parallel module tasks happen to finish:

```rust
// crates/rolldown/src/module_loader/module_loader.rs
while self.remaining > 0 {
    let Some(msg) = self.rx.recv().await else { break };   // parallel completion order
    ModuleLoaderMsg::NormalModuleDone(..) => {
        ... try_spawn_new_task(..) -> alloc_ecma_module_idx()   // index assigned here
    }
}
```

Link-stage passes then walk the module table by index, so that timing reaches the output.
Two we traced end to end:

**`tree_shaking/determine_side_effects.rs`** - a re-entrant visit (`SideEffectCache::Visited`)
returns the module's not-yet-finalized verdict, so the module the walk starts from decides the
result for shared or circular dependencies.

**`determine_module_exports_kind.rs`** - an `ExportsKind::None` importee is promoted by
whichever importer is visited first (`import` → `Esm`, `require` → `CommonJs`). The existing
comment says as much:

```rust
// the "earlier importer's promotion is observed by later importers" semantics.
```

That `exports_kind` decides the `__toESM(mod, isNodeMode)` flag, so emitted code flips between:

```js
var i = n(d(), 1), h = n(x(), 1)
var i = n(d()),    h = n(x())
```

`isNodeMode` is not cosmetic - it selects
`isNodeMode || !mod || !mod.__esModule ? wrap-as-default : keep`, so an unstable value can
change what `import X from 'cjs-pkg'` resolves to at runtime.

### Fixing individual passes does not help

| attempt | our app (5759 files, 3 builds) |
| --- | --- |
| stock 1.2.8 | 848 / 896 / 875 files changed |
| sort those two passes by `stable_id` | 852 / 964 / 972 |
| #10893 (worklist side-effect propagation) | 908 / 919 / 955 |

#10893 is worth calling out since it rewrites one of the affected passes. It fixes a
**correctness** bug (a module in a cycle marked side-effect free before the effect is found),
not this one: with it applied on top of v1.2.8 the reduced repro still splits. A pass
can compute a correct verdict and the build can still differ, because the walk it is handed
depends on task timing - and it does not touch `determine_module_exports_kind`, which produces
the unstable `__toESM` flag. The two are complementary and apply cleanly together.

(Measured on our app, where a run takes ~40s and the sample is stable. The reduced repro
shows the same behaviour but with a smaller sample per run.)

Every pass consumes the same nondeterministic index, so fixing one moves the symptom instead
of removing it.

### Ordering index assignment does

Handling completions one wave at a time, ordered by module id, while tasks keep running in
parallel:

| | reduced repro | our app |
| --- | --- | --- |
| with that change | **12 / 12 identical** | **4 / 4 builds byte-identical (5761 files)** |

Both symptoms disappear together - the tree-shaking flip and the `__toESM` flag flip - which
is what identifies index assignment as the shared cause. Draft PR: https://github.com/rolldown/rolldown/pull/10910

### Impact

In our production app (~4600 chunks, vite 8.0.16), **~25% of chunk filenames changed on every
clean build of the same commit** (1191/4603). A single flipped module changes one chunk's
bytes, which shifts entries in Vite's `__vitePreload` dependency arrays, which changes each
importer - so a handful of unstable modules fan out to more than a thousand renamed chunks.
No CDN or browser keeps a warm copy across deployments.

We ruled out our own setup first: removing panda (styled-system), the design system, redux,
every custom vite plugin, a build-timestamp `define`, and cutting a 577-module import cycle -
cumulatively - left the variation unchanged (1169 → 1222 files changed).

---

### System Info

```
System:
  OS: macOS 26.2
  CPU: (10) arm64 Apple M5
  Memory: 32.00 GB
Binaries:
  Node: 22.19.0
  pnpm: 11.24.0
npmPackages:
  rolldown: 1.0.3 (via vite 8.0.16), 1.2.8 (via vite 8.3.0)
```

### Any additional comments?

**AI usage disclosure.** Per the contribution guide's AI Usage Policy: I used Claude Code while
investigating this. The elimination passes, the build-to-build diffing and the reduction of the
reproduction were run as scripted experiments with its help, and it drafted this report. Every
number quoted here comes from a build I ran and checked myself, the root cause was confirmed by
reading the source and by instrumenting a local debug build, and the linked fix was built and
measured before being written up. I take responsibility for the content.

The reduced repository preserves the module-graph shape of the original app - 32 source
files, 35 runtime import edges, **no cycles**. Internal names were mechanically renamed and
i18n data replaced with placeholders. Synthetic graphs of the same shape did not reproduce
it, which is why this is a reduction of a real app rather than a generated fixture.

y0ngha:
Fixes #10909.

`ModuleIdx` is handed out while handling task-completion messages, and those arrive in
whatever order parallel module tasks happen to finish:

```rust
while self.remaining > 0 {
    let Some(msg) = self.rx.recv().await else { break };
    ModuleLoaderMsg::NormalModuleDone(..) => { ... try_spawn_new_task(..) }  // index assigned
}
```

Link-stage passes walk the module table by index, so that timing reaches the output. Two
symptoms traced end to end, both order-dependent walks over that table:

**A. Tree-shaking flips.** `tree_shaking/determine_side_effects.rs` re-enters a module that is
still being visited (`SideEffectCache::Visited`) and returns its not-yet-finalized verdict, so
for a shared or circular dependency the module the walk *starts from* decides the answer. In
the reproduction this keeps or drops a ~430-module subgraph:

```
modules per build: [436, 436, 5, 5, 436, 5, 436, 5]
```

`package.json`'s `sideEffects` is what makes this visible, and it is worth being explicit that
it is a switch rather than a cause: declaring it routes modules onto the `Analyzed(false)`
recursive branch where the re-entrancy lives. Without it that branch is never taken, so the
build looks stable while the underlying ordering is just as unstable.

**B. The `__toESM` interop flag flips.** `determine_module_exports_kind.rs` promotes an
`ExportsKind::None` importee according to whichever importer is visited first (`import` -> Esm,
`require` -> CommonJs); the existing comment calls this "the earlier importer's promotion is
observed by later importers". That kind decides the second argument of `__toESM(mod,
isNodeMode)`, so emitted code alternates between:

```js
var i = n(d(), 1), h = n(x(), 1)
var i = n(d()),    h = n(x())
```

This one is not cosmetic. `isNodeMode` selects
`isNodeMode || !mod || !mod.__esModule ? wrap-as-default : keep`, so an unstable value can
change what `import X from 'cjs-pkg'` resolves to at runtime, not just the hash.

The two sit in different passes and have different-looking symptoms, which is why fixing either
pass alone only moves the problem. They share one input: the order the module table is in.

This handles completions **one wave at a time, ordered by module id**. Tasks still run in
parallel; only the point where indices are handed out is ordered.

```rust
let mut wave: VecDeque<ModuleLoaderMsg> = VecDeque::new();
while self.remaining > 0 || !wave.is_empty() {
  if wave.is_empty() {
    let wave_size = self.remaining;
    let mut batch = Vec::with_capacity(wave_size as usize);
    for _ in 0..wave_size {
      let Some(msg) = self.rx.recv().await else { break };
      batch.push(msg);
    }
    batch.sort_by_cached_key(module_loader_msg_order_key);
    wave.extend(batch);
  }
  let Some(msg) = wave.pop_front() else { break };
  match msg { /* unchanged */ }
}
```

A wave waits for a *count* of messages rather than for any particular one, which matters for
liveness - see "Alternatives considered" below.

36 insertions, 2 deletions. The existing `match` body is untouched.

## Results

| | reduced repro (12 builds) | my app (5761 files, 3-4 builds) |
| --- | --- | --- |
| stock | splits 8 : 4; the ratio varies per run | 848 / 896 / 875 files changed |
| sorting those two passes by `stable_id` | still splits | 852 / 964 / 972 |
| #10893 | still splits | 908 / 919 / 955 |
| **this change** | **12 / 12 identical** | **0 across 4 builds** |

Reproduction: https://github.com/y0ngha/rolldown-nondeterministic-output

## Relationship to #10893

#10893 touches `determine_side_effects.rs`, which is one of the passes affected here, so it is
worth stating how the two differ. They fix different things and do not overlap:

| | #10893 | this PR |
| --- | --- | --- |
| file | `tree_shaking/determine_side_effects.rs` | `module_loader/module_loader.rs` |
| problem | **correctness** - a module in a cycle is marked side-effect free before the effect is found, so an import path is dropped and `main.js` logs `['main']` instead of `['effect', 'main']` | **reproducibility** - the same input yields different output between runs |
| mechanism | replaces the recursive `SideEffectCache` with worklist propagation | orders `ModuleIdx` assignment in the module loader |
| scope | that one pass | the index every link-stage pass consumes |

Because #10893 removes the `SideEffectCache::Visited` re-entry, I expected it to also remove
the ordering dependency in that pass. It does not fix the reported nondeterminism:

```
with #10893 applied on top of v1.2.8, the reduced repro still splits
our app, 3 builds:                                               908 / 919 / 955 files changed
```

That is the measurement that pointed away from individual passes: the pass can compute a
correct verdict and the build can still differ between runs, because the walk it is handed
depends on task timing. The same applies to the `determine_module_exports_kind` path, which
#10893 does not touch at all and which produces the unstable `__toESM(mod, isNodeMode)` flag.

The two changes are complementary - #10893 makes the side-effect verdict right, this PR makes
the input to every pass stable. This branch is based on v1.2.8 rather than on
#10893, and I measured #10893 on its own. Stacking them is clean: cherry-picking this commit
onto #10893's head applies without conflict and `cargo check -p rolldown` passes.

## Alternatives considered

- **Fix each pass** - stable-sort its walk, or make it order-independent. Measured above: no
  effect, since every pass consumes the same nondeterministic index.
- **Release each completion as soon as no smaller index is still in flight** (a conservative
  ordering instead of a barrier). I built this hoping to avoid the barrier's cost, but it
  **deadlocks**, so I am not proposing it and did not benchmark it on an idle machine.
  `PluginContext::load` sends `FetchModule` and then awaits the module it asked for from inside
  a hook, so the calling task stays in flight while it waits. The module it asked for is
  discovered later and therefore gets a *higher* index, and an ordering rule that refuses to
  handle a higher index before a lower one in flight makes the two wait on each other forever.
  It reproduces on rolldown's own `build-node` step. The wave barrier avoids this because it
  waits for a *count* of messages, not for a particular one, so a later index can still be
  handled in the same wave.
- **Re-map `ModuleIdx` after loading** (load in arrival order, permute at the end). The only
  shape that keeps the pipeline completely, and it has no ordering rule to deadlock on, but
  every `ModuleIdx` in the scan output has to move, including `SymbolRef { owner: ModuleIdx,
  symbol }` and the `IndexVec<ModuleIdx, Option<SymbolRefDbForModule>>` behind `symbol_ref_db`,
  plus entry points, import records' `resolved_module`, the stmt-info and AST side tables and
  the incremental cache's `module_id_to_idx`. Much larger and easy to get subtly wrong.
- **Serialize module loading** - deterministic, but gives up parallelism.

## Tests

`crates/rolldown/tests/rolldown/function/deterministic_module_idx` scans the same fixture twice
with its two importers finishing in opposite orders (a plugin staggers their `load`) and
compares the module table. Which importer finishes first is what decides how the two importers'
children are interleaved, so without a stable order the table comes out permuted:

```
left:  [runtime, main.js, p.js, q.js, p1.js, p2.js, p3.js, q1.js, q2.js, q3.js]
right: [runtime, main.js, p.js, q.js, q1.js, q2.js, q3.js, p1.js, p2.js, p3.js]
```

The test fails on `main` without the loader change and passes with it. It asserts on the module
table rather than on emitted output, because whether an index permutation reaches the output
depends on the graph, while the permutation itself is the thing being fixed.

`cargo test -p rolldown`: 2000 passed, no snapshot changes. (`test262_module_code` fails in my
checkout because the test262 submodule is not cloned; unrelated.)

`internal-docs/linking/module-execution-order/implementation.md` documents stable `ModuleIdx`
as a precondition for the order guarantee that doc makes.

## Benchmarks

`cargo bench -p bench` on an idle M5 laptop. Both binaries are built up front and then run
alternately, one process at a time, so no recompile lands between the two columns.

`bundle@rome_ts`, three pairs:

| pair | stock | this change | change |
| --- | --- | --- | --- |
| 1 | 49.55 ms | 50.96 ms | +2.9% (p = 0.05) |
| 2 | 48.49 ms | 52.79 ms | +8.9% |
| 3 | 48.90 ms | 52.01 ms | +6.4% |
| **mean** | **48.98 ms** | **51.92 ms** | **+6.0%** |

Stock spans 48.49-49.55 ms and the patched build spans 50.96-52.79 ms, so both are steady to
within a few percent.

An earlier revision of this description reported +15.1% here, with a claim that the barrier also
widened the run-to-run spread to 24%. That was wrong: those runs were taken while other work was
running on the same machine. Re-measured with the machine idle, the spread is normal and the
cost is the ~6% above. I have corrected it rather than leave the worse number standing.

`bundle@threejs`, `bundle@threejs-sourcemap` and `bundle@multi-duplicated-top-level-symbol`
stayed within noise. The `sourcemap/*` group cannot be affected by this change at all - it never
touches the module loader - and it still moved across runs, which is the floor for what is
measurable here.

## For reviewers

A wave waits for every in-flight completion before handling any of them, so it ends on its
slowest task and the next round of module tasks is spawned later than it could be. Parsing and
transforming are untouched - they run in the same parallel tasks as before, and the loop that
handles completions was already single-threaded. The cost is pipeline depth: ~6% on `rome_ts`,
nothing measurable elsewhere.

If the barrier is not acceptable, the re-map option under "Alternatives considered" is the one
I would take next; it keeps the pipeline completely. It is a much larger change, so I would
rather agree on the direction before writing it.

## AI usage disclosure

Per the contribution guide's AI Usage Policy: I used Claude Code to investigate the bug and to
draft this PR. The change is small and I understand it, the measurements in the table are from
builds I ran and verified myself, and I read the surrounding module-loader code before and after
changing it.
I take responsibility for this contribution.


y0ngha:
Marking this ready for review. One check is red and I do not think it is mine:

`wasi / WASI` fails on `fixtures-concurrent.test.ts > builtin-plugin/import-glob/tsconfig-paths-multiple-targets`
(969 passed, 1 failed). The workflow retried it three times and then labelled it itself:

```
wasi Node Test attempt 3 failed or hung; no more retries (known flake, see #10697)
```

#10697 is open and describes WASI throwing "memory access out of bounds" when concurrent builds
call JS plugin hooks, which this change does not touch. Everything else on the run is green.

Happy to rebase or re-run if you would rather see it pass first.

hyfdev:
Thanks for your effort. I checked the PR and I think https://github.com/rolldown/rolldown/pull/10918 is in a more correct direction.

y0ngha:
> 수고해주셔서 감사합니다. PR을 확인해 보니 #10918 이 더 올바른 방향인 것 같습니다.

Thank you for confirming. 
Since you've already researched and worked on Repl, there probably won't be any issues, but I'll apply #10918 and check if the error disappears after applying it to my app.

y0ngha:
> Thanks for your effort. I checked the PR and I think #10918 is in a more correct direction.

Hi @hyfdev 

I also tested it on my app, and it was confirmed that the randomness issue was resolved.
I will close this PR.
Thank you to the rolldown team for their prompt response.

PR #10918, opened 2026-09-19
fix: apply `package.json#sideEffects` to ids a plugin returns as a string

### LINKED ISSUE #10909: [Bug]: ModuleIdx is assigned in task-completion order, so identical builds produce different output

IWANABETHATGUY:
A `resolveId` hook that returns a bare id string carries no `packageJsonPath`, so the module got neither its `package.json#sideEffects` policy nor its module format, while the same file reached through a relative import got both. The module loader keeps whichever `ResolvedId` arrives first, so the same input produced two different bundles.

The linked reproduction now builds the same 5-module output on 15 of 15 builds, where stock splits 6:6 over 12 builds between 5 and 437 modules.

close #10909


crates/rolldown_plugin/src/utils/resolve_id_with_plugins.rs
```diff
@@ -48,18 +48,34 @@ pub fn infer_module_def_format(
...
 fn resolved_id_from_hook_output<Fs: FileSystem>(
   resolver: &Resolver<Fs>,
   r: HookResolveIdOutput,
 ) -> anyhow::Result<ResolvedId> {
-  let package_json = r
-    .package_json_path
-    .as_ref()
-    .map(|p| resolver.try_get_package_json_or_create(p.as_path()))
-    .transpose()?;
+  let id = ModuleId::new(r.id);
```
hyfdev, line 60: This feels suspcious.

crates/rolldown_resolver/src/resolver.rs
```diff
@@ -54,6 +57,20 @@ impl<Fs: FileSystem + Clone + 'static> Resolver<Fs> {
     let require_resolver = default_resolver.clone_with_options(config.require_options);
     let css_resolver = default_resolver.clone_with_options(config.css_options);
     let new_url_resolver = default_resolver.clone_with_options(config.new_url_options);
+    let exact_path_resolver = default_resolver.clone_with_options(OxcResolveOptions {
```
hyfdev, line 60: This feels suspcious too.

crates/rolldown_plugin/src/utils/resolve_id_with_plugins.rs
IWANABETHATGUY: A `resolveId` hook that returns a string gives no `package.json`. The internal resolver attaches one to every id it resolves. Example:

```
node_modules/fmt/package.json   { "sideEffects": false }
node_modules/fmt/index.js       globalThis.fmt = 1
src/a.js                        import 'fmt'    // internal resolver
src/b.js                        import '@fmt'   // a plugin returns the path of index.js as a string
```

For `a.js`, the `ResolvedId` has the `package.json`, so rolldown drops `fmt`. For `b.js`, the `ResolvedId` has no `package.json`, so rolldown keeps `fmt`. The module loader keeps the first `ResolvedId` per id (https://github.com/rolldown/rolldown/blob/ca66f9cea2791e392aa00c417075b47b5a7c5c3d/crates/rolldown/src/module_loader/module_loader.rs#L320-L326). So one build drops `fmt` and the next build keeps it (#10909).

This lookup only attaches the `package.json` and the module format. The id from the plugin does not change. JS plugins return a string, so a rule that plugins must set `packageJsonPath` does not help them.

crates/rolldown_resolver/src/resolver.rs
IWANABETHATGUY: oxc_resolver has no public function that returns the `package.json` and the module format for an existing path. Its resolution functions are `resolve`, `resolve_file`, and `resolve_with_context`, and each one runs the full pipeline. So this resolver runs the full pipeline, but every option that rewrites the path is off.

Those options must be off because the id from the plugin is final. This example sets `resolve.alias: { '<root>/alias-src/index.js': '<root>/alias-target/index.js' }`:

- A plugin returns `<root>/alias-src/index.js`. Its `package.json` has `"sideEffects": true`.
- The default resolver applies the alias to that path. It returns the `package.json` of `alias-target`. That file has `"sideEffects": false`.
- rolldown then drops a module that has side effects.

`tsconfig` paths, `extensions`, and `extension_alias` can rewrite the path too.

I can request a lookup function in oxc_resolver. It takes a path and returns only the `package.json` and the module format. Then we can remove this resolver and the option list. I think that is work for a follow-up PR.

PR #10933, opened 2026-09-20
fix: replace `rolldown-custom/ban-expect-assertions` with builtin `no-restricted-properties`

isker:
Replace the custom `rolldown-custom/ban-expect-assertions` JS lint with a builtin rule `no-restricted-properties`. (This rule did not exist when the JS lint was introduced.) This allows rolldown's pre-commit hook to run in low-memory environments.

Specifically, Oxlint allocates 4GB of virtual memory per-thread in the presence of JS lints. Relevant code:
https://github.com/oxc-project/oxc/blob/12cb5473519327bd37780f2e42bf1eade98542b6/crates/oxc_allocator/src/pool/fixed_size.rs#L31-L38
https://github.com/oxc-project/oxc/blob/12cb5473519327bd37780f2e42bf1eade98542b6/crates/oxc_allocator/src/pool/fixed_size.rs#L377-L380
https://github.com/oxc-project/oxc/blob/12cb5473519327bd37780f2e42bf1eade98542b6/crates/oxc_allocator/src/fixed_size.rs#L19
https://github.com/oxc-project/oxc/blob/12cb5473519327bd37780f2e42bf1eade98542b6/crates/oxc_allocator/src/arena/fixed_size/unix.rs

If the environment cannot allocate 4GB, the pre-commit hook fails. [I'm using such an environment](https://ampcode.com/docs/orbs) to contribute to rolldown today. I understand that there can't be a guarantee of no JS lints (or other things that require more memory) in the future, but this particular one seems unnecessary and easy to replace. (Note that this new configuration is not strictly equivalent, because it fails on any usage of `expect.assertions`, not just calls as the current JS lint does. That being said, I think the distinction does not matter in this particular case.)

AI disclosure: this fix was generated by Amp. You can see details in [the thread](https://ampcode.com/threads/T-01a0bf99-54e7-703e-b8fb-07420e9de4c2). I understand and take responsibility for the changes.


shulaoda (approved):
Thanks for the contribution!

PR #10929, opened 2026-09-20
fix(docs): restore Algolia search

isker:
Search is broken on the site currently:
image

This seems to fix it, as the comment describes.

AI disclosure: this entire fix was implemented with Amp. You can see the [thread](https://ampcode.com/threads/T-01a0bf7a-b23b-7437-81c2-a1b103c08818), which includes a screenshot demonstrating the fix: 
image

PR #10934, opened 2026-09-21
fix(docs): type-check Algolia `indices` option

shulaoda:
Fix https://github.com/rolldown/rolldown/actions/runs/35522778246/job/106109809314


docs/.vitepress/config.ts
```diff
@@ -279,9 +279,10 @@ const config = defineConfig({
         appId: process.env.ALGOLIA_APP_ID || '',
         apiKey: process.env.ALGOLIA_API_KEY || '',
         // DocSearch 5 requires `indices`; `indexName` remains for VitePress compatibility.
+        // VitePress types only declare `indexName`, so the extra field is asserted.
         indexName: 'rolldown',
         indices: ['rolldown'],
-      } satisfies DefaultTheme.AlgoliaSearchOptions & { indices: string[] },
```
hyfdev, line 284: Why not fix the code?

hyfdev (approved):

docs/.vitepress/config.ts
shulaoda: DocSearch 5 needs `indices` filed, but VitePress doesn't have it by default and still checks `indexName`.

isker:
Sorry for landing the bad types. CI was somehow green in my PR but things did fail to typecheck locally.

docs/.vitepress/config.ts
shulaoda: It's related to https://github.com/rolldown/rolldown/pull/10929

PR #10944, opened 2026-09-22
feat(watch): support directories in addWatchFile

shulaoda:
Watch directories passed to `this.addWatchFile` recursively, so a change below them triggers a rebuild or an HMR update.


crates/rolldown_dev/src/bundle_coordinator.rs
h-a-n-a, line 177: Are we still using kqueue on macOS?

h-a-n-a:
We can add a generated file check just like what [Rollup](https://github.com/rollup/rollup/blob/c28de6fa616ff6525802996a6b480b57d1e7b0e1/src/watch/watch.ts#L279-L281) did. Not a blocker.

crates/rolldown_dev/src/bundle_coordinator.rs
shulaoda: I think it's a wrong comment, we seem to use fs event.
Image

h-a-n-a: Let's update that comment.

shulaoda: I’ll leave the comments for now and clean up the inaccurate ones in a later PR so this one stays focused.

h-a-n-a (approved):

crates/rolldown_dev/src/bundle_coordinator.rs
h-a-n-a: Maybe leave the conversation open next time. So reviewers don't have to un-collapse it and see your reply. I will resolve it or you can resolve it after the PR is approved.

PR #10945, opened 2026-09-22
feat(watch): watch missing addWatchFile paths

shulaoda:
Watch paths passed to `this.addWatchFile` that do not exist yet, as long as their parent directory does.


packages/rolldown/tests/watch/watch.test.ts
```diff
@@ -841,6 +841,54 @@ test.concurrent(
   },
 );
 
+test.concurrent(
+  'PluginContext addWatchFile with a missing path',
```
h-a-n-a, line 845: Can we try turn on the Rollup test as well? https://github.com/rolldown/rolldown/blob/1551cf8a09f39ae680e7c6d20f9c3b35ff9b0d71/packages/rollup-tests/test/watch/index.js#L1618

crates/rolldown_fs_watcher/src/notify/mod.rs
```diff
@@ -66,11 +66,19 @@ impl<'me> NotifyPathsMutAdapter<'me> {
...
+      Ok(_) => true,
+      Err(error) if error.kind() == std::io::ErrorKind::NotFound => {
+        path.parent().is_some_and(Path::exists)
+      }
+      Err(_) => false,
+    };
+    let result = if is_watchable {
+      self.0.add(path, WatchMode { recursive_mode, target_mode: TargetMode::TrackPath })
+    } else {
+      Err(::notify::Error::path_not_found().add_path(path.to_path_buf()))
```
h-a-n-a: Looks not right. Any other metadata error is reported as not found here as well.

shulaoda: Yes, maybe we should resolve this inside notfiy-rs

packages/rolldown/tests/watch/watch.test.ts
shulaoda: When everything is ready, I will try to reopen some rollup tests, currently it also has some work to be done

crates/rolldown_watcher/src/watch_task.rs
```diff
@@ -246,9 +246,6 @@ impl WatchTask {
         continue;
       }
       let path = watch_path.as_path();
-      if !path.exists() {
```
h-a-n-a, line 249: Now that the parent-exists check in `rolldown_fs_watcher` is gone, a path whose parent is also missing reaches notify. Before this PR such a path was skipped silently.

Suggest keeping a skip here:

```rust
if !path.exists() && !path.parent().is_some_and(Path::exists) {
  continue;
}
```

It is retried next build, since the path never enters `watched_files`.

_Reviewed with Claude Fable 5.1._

shulaoda: That's why I said "maybe we should resolve it inside notify-rs" (https://github.com/rolldown/rolldown/pull/10945#discussion_r4069154987).

This is notify's responsibility: under `TrackPath` it tracks a missing path through its parent directory, so a path with `!path.exists() && !path.parent().is_some_and(Path::exists)` is already an error there. On inotify `add` returns it right away. What is left is making the batching backends return it from `add` as well instead of failing the whole `commit`, and that is what I'll do in notify, so rolldown does not need a check of its own.

h-a-n-a: I'm talking about maintaining the same behavior that a parent is missing as it is now if we haven't implement this on the notify-rs side.

shulaoda: got it, we just leave it here for now.

PR #10946, opened 2026-09-22
docs(watch): document watch paths and path identity

shulaoda:
Document `WatchPath` and directory and missing watch paths in the internal docs.


internal-docs/module-id/implementation.md
```diff
@@ -28,7 +28,7 @@ However, this is **only used in downstream/output contexts**, not in the core mo
 - `relativeId.ts` — computing relative import paths
 - `MetaProperty.ts` — import.meta relative paths
 
-Plugin APIs like `addWatchFile()` do **no normalization** — they trust the caller to provide a path consistent with the module ID convention.
+Plugin APIs like `addWatchFile()` do **no normalization** — they trust the caller to provide a path consistent with the module ID convention. Rolldown normalizes these paths itself where it compares them with changed paths, see `WatchPath` in [watch-mode](../watch-mode/implementation.md#path-identity).
```
h-a-n-a: This sentence describes Rolldown but sits inside "How Rollup Does It". Please move it to "How Rolldown Does It Today" below, next to the `add_watch_file` row of the table, so the Rollup section stays about Rollup.

_Reviewed with Claude Fable 5.1._

internal-docs/watch-mode/implementation.md
```diff
@@ -305,19 +305,27 @@ All watch-related hooks are **blocking** — the coordinator awaits their comple
...
 ## File Watching
 
-- After each build, `bundler.watch_files()` returns the current set.
-- `WatchTask::update_watch_files()` diffs against the current set — new files are added to the per-task `FsWatcher`.
-- `include`/`exclude` patterns filter which files are watched (via `pattern_filter`).
-- Files are watched **non-recursively** (individual file watches).
-- Batch operations: `fs_watcher.paths_mut()` returns a guard for batching adds, committed via `.commit()`.
+- After each build, `bundler.watch_files()` returns the current set: the modules read from disk, their tsconfig files, and the raw strings plugins passed to `this.addWatchFile`.
+- `WatchTask::update_watch_files()` diffs against the current set — new paths are added to the per-task `FsWatcher`.
+- `include`/`exclude` patterns filter which watch paths are registered (via `pattern_filter`). They are not applied to what lies below a watched directory yet.
```
h-a-n-a: "yet" makes this read as a gap to fix. Rollup behaves the same way: its filter runs only when a path is registered (`src/watch/watch.ts:266`), and changes under a watched directory are never filtered. Suggest: "They apply to the registered path only, not to files below a watched directory. This matches Rollup."

_Reviewed with Claude Fable 5.1._

internal-docs/watch-mode/implementation.md
```diff
@@ -305,19 +305,27 @@ All watch-related hooks are **blocking** — the coordinator awaits their comple
...
+- Batch operations: `fs_watcher.paths_mut()` returns a guard for batching adds, committed via `.commit()`. A path only enters `watched_files` after the commit succeeded, so a skipped path is tried again with the next build.
+
+### Watch Paths
+
+A watch path can be a **file**, a **directory** or **missing**, like in Rollup. Bundled dev follows the same rules.
+
+- Every path is registered with `RecursiveMode::Recursive` and `TargetMode::TrackPath`. notify ignores the recursive mode for a file. `TrackPath` follows the path: a missing path is watched once it is created, and a path that changes between a file and a directory stays watched.
+- notify tracks a missing path through its parent directory. Without the parent inotify and kqueue fail, and one failing path makes the batching backends drop the whole batch, so `rolldown_fs_watcher` rejects such a path itself.
```
h-a-n-a: Nit: Windows fails too when the parent is missing, not only inotify and kqueue. "Without the parent the backends fail" would be safer than naming two.

_Reviewed with Claude Fable 5.1._

h-a-n-a (commented):
The new Watch Paths and Path Identity sections match the code, including the batching claim (kqueue, Windows, and poll all commit a batch at once). Two more leftovers in `internal-docs/module-id/implementation.md` are outside the diff, so they are listed here:

- Line 86, table row "Watch file comparison": still says `ArcStr` eq with a `#[cfg(windows)]` backslash fallback, marked "Fragile". #10943 replaced the set with `WatchPath` and removed that fallback. Suggest: `WatchPath (Arc<Path>)` / absolute + lexically normalized at insert, `Path` component eq / handles separators, `.`, trailing slash; not `..` or symlinks.
- Line 127, Unresolved Questions, "Should the watch file set use PathBuf instead of ArcStr?": answered by #10943. Please remove it, or replace it with a one-line note: "Decided: the watch set stores `WatchPath`, see watch-mode/implementation.md#path-identity."

_Reviewed with Claude Fable 5.1._

